PR #4, opened 2018-10-09
Cleanup i18n translations

emmenko:
#### Summary

image

#### Description

I pushed the translations already to tx:

image


emmenko:
Update: the last commit https://github.com/commercetools/merchant-center-application-kit/pull/4/commits/227538f56b48996964a25f69cc5971ac2471be68 requires to pass `applicationMessages` as a prop to `<ApplicationShell>`. The idea is that the application that has its own messages would inject them into the AppShell which will then merge them with the own ones and the ones from UI Kit.

tdeekens:
I appreciate the effort but does that mean we will have three projects two of which have < 60 translations? I did not know this was a plan nor discussed anywhere. Did people raise this to you as being a concern that we can't manage those 120 translations by hand?

Quick q: how do we pull these?

tdeekens:
I agree with that. Lets move discussions to the ui-kit repo. My main qs are asked there. I approve this in the mean time.

emmenko:
Alright, I've updated the translations in Tx, we're good to go now! 🚀

PR #10, opened 2018-10-11
Playground: remove unnecessary files/config, integrate it in dev workflow

emmenko:
Cleans up the playground folder (imported from https://github.com/commercetools/merchant-center-starter-app - #9 ).

Now we can use that app to test/develop some features in the app-kit packages.

Dependencies are installed as part of `yarn install` and `yarn start` will start the dev server of the playground app.


.travis.yml
```diff
@@ -13,4 +13,6 @@ cache:
   directories:
     - node_modules
 
-script: yarn run jest --projects jest.*.config.js --maxWorkers=4 --reporters jest-silent-reporter
+script:
+  - yarn run jest --projects jest.*.config.js --maxWorkers=4 --reporters jest-silent-reporter
+  - yarn playground:build
```
emmenko: This might be useful to catch some possible problems with the packages.

As a follow up, we might also think about running some e2e tests...not a high prio though

playground/package.json
```diff
@@ -2,74 +2,36 @@
...
     "now-build": "NODE_ENV=production npm run build",
-    "now-start": "NODE_ENV=production npm run start:prod",
-    "lint": "jest --projects jest.eslint.config.js",
-    "format": "npm run format:js && npm run format:md",
-    "format:js": "prettier --write 'src/**/*.js'",
-    "format:md": "prettier --parser markdown --write '*.md,src/**/*.md'",
-    "test": "jest --config jest.test.config.js",
-    "test:watch": "jest --config jest.test.config.js --watch",
-    "ci": "jest --projects jest.*.config.js"
+    "now-start": "NODE_ENV=production npm run start:prod"
```
emmenko: In theory, we can install `now` to automatically build/deploy the app. It won't be useful for branch deployments though and for checking the UI because of CORS restrictions if you open the browser at a domain that is not whitelisted.

For now I'll keep it without but we might consider looking into that workflow in the future.

playground/package.json
```diff
@@ -2,74 +2,36 @@
...
-    "react-dom": "16.5.1",
+    "react": "16.5.2",
+    "react-dom": "16.5.2",
+    "react-intl": "2.7.1",
     "react-loadable": "5.5.0",
     "react-redux": "5.0.7",
     "react-router": "4.3.1",
-    "react-router-dom": "4.3.1"
+    "react-router-dom": "4.3.1",
+    "react-select": "1.2.1"
```
dferber90: Unrelated: Oh, I wasn't aware that we're using `react-select` here! We probably need to upgrade it to v2 at some point.

emmenko: Yeah, actually for the playground app it's because of the import

```js
import 'react-select/dist/react-select.css';
```

Also the AppShell still uses `v1`

PR #12, opened 2018-10-15
fix(navbar): adds Scrollable and FixedMenu

lufego:
#### Summary

The "absolute" positioned items in the NavBar, covered other items when the menu got expanded in smaller height screens

#### Description

This PR adds a FixedMenu and a ScrollableMenu so **if the screen gets smaller** it will add a scrolling bar inside the items which gets rendered inside the `ScrollableMenu` section

#### Cross-Browser

* [x] Chrome
* [x] Firefox
* [ ] Safari - couldn't test on safari locally since I can't login in this browse, issue needs to be reported
* [ ] (IE/Edge)

#### Tests
* Components
  * [x] Rendered elements (also different states)

#### Technical debt & future

The NavBar is being refactored, so this is a quick fix for the problem

**Before**
bug-1

**After**
bug-2

**SUPPORT LINK**
https://jira.commercetools.com/browse/SUPPORT-3996

### DIFF
--- a/packages/application-shell/src/components/navbar/navbar.js
+++ b/packages/application-shell/src/components/navbar/navbar.js
@@ -97,7 +97,6 @@ export const MenuItem = props => (
       [styles.item__active]: props.isActive,
       [styles['item_menu-collapsed']]: !props.isMenuOpen,
       [styles['item__no-submenu']]: !props.hasSubmenu,
-      [styles['item--bottom']]: props.isBottomItem,
     })}
     onClick={props.onClick}
     onMouseEnter={props.onMouseEnter}
@@ -343,9 +342,10 @@ export class DataMenu extends React.PureComponent {
     return Boolean(match);
   };
 
-  handleToggleItem = index => {
-    if (this.state.activeItemIndex !== index)
-      this.setState({ activeItemIndex: index });
+  handleToggleItem = (menuType, index) => {
+    const activeItem = `${menuType}-${index}`;
+    if (this.state.activeItemIndex !== activeItem)
+      this.setState({ activeItemIndex: activeItem });
   };
 
   shouldCloseMenuFly = event => {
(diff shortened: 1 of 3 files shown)


packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -379,111 +379,131 @@ export class DataMenu extends React.PureComponent {
...
+            hasSubmenu={hasSubmenu}
+            isActive={isActive}
+            isMenuOpen={this.state.isMenuOpen}
+            onClick={() => {
+              this.handleToggleItem(menuType, index);
+            }}
+            onMouseEnter={
+              this.state.isMenuOpen
+                ? null
+                : () => this.handleToggleItem(menuType, index)
```
islam3zzat: Nitpick: Can we early-return when `this.state.isMenuOpen` is `true` in `handleToggleItem`, instead of  conditionally adding this prop

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -379,111 +379,131 @@ export class DataMenu extends React.PureComponent {
...
+            }
+            onMouseLeave={
+              this.state.isMenuOpen ? null : this.shouldCloseMenuFly
```
islam3zzat: here too

packages/application-shell/src/components/navbar/navbar.js
dferber90: I would say conditionally adding the prop might be preferable as we don't keep a `mouseEnter` event listener around in case the menu is open already.

packages/application-shell/src/components/navbar/navbar.js
lufego: This is old code 😅based on what Filip told me, navbar will be completely reworked on soon

montezume:
Might make sense to bump the version, create a release and then update MC so that we have this in the MC?

montezume:
@mariabarrena I opened a PR for a fix to that https://github.com/commercetools/merchant-center-application-kit/pull/15

PR #15, opened 2018-10-17
fix(nav-bar): widen expanded navbar to fit spanish support text

montezume:
#### Summary

Right now, when in Spanish mode, you get this behaviour because the active item isn't wide enough. (because the text is Support - in english) instead of just support

bug

#### Approach

Make active item wider so that this doesn't happen.

fix

cc @mariabarrena


mariabarrena:
Thanks for taking care of this! We needed to say that support is in English only and that's why the menu item is way longer for Spanish. I would expect the same to happen in all other not-English languages.

PR #18, opened 2018-10-17
feat(quick-access): support opening quick access when link has focus

dferber90:
Quick Access not opening was reported as a bug:

image

So far it was not possible to open Quick Access by pressing `f` while a link had focus.

This PR enables that by exempting links from the early return we are doing. This early return is what stops Quick Access from opening while a user types into a form input for example.

I think it's a fair assumption to open Butler while a link has focus.

### DIFF
--- a/packages/application-shell/src/components/quick-access/index.js
+++ b/packages/application-shell/src/components/quick-access/index.js
@@ -47,7 +47,9 @@ export default class QuickAccessTrigger extends React.Component {
       // react-modal uses it for example. We want to treat those elements
       // similar to document.body.
       // See https://stackoverflow.com/a/32912224
-      event.target.getAttribute('tabindex') !== '-1'
+      event.target.getAttribute('tabindex') !== '-1' &&
+      // Do not prevent Quick Access from opening when a link has focus
+      event.target.nodeName !== 'A'
     )
       return;
 


packages/application-shell/src/components/quick-access/index.js
```diff
@@ -47,7 +47,9 @@ export default class QuickAccessTrigger extends React.Component {
       // react-modal uses it for example. We want to treat those elements
       // similar to document.body.
       // See https://stackoverflow.com/a/32912224
-      event.target.getAttribute('tabindex') !== '-1'
+      event.target.getAttribute('tabindex') !== '-1' &&
+      // Do not prevent Quick Access from opening when a link has focus
+      event.target.nodeName !== 'A'
```
emmenko: Can you point to the documentation about this info?

tdeekens:
@dferber90 and @emmenko I'd like to release a version of app-kit packages. Would you want this in that release? Anything missing here?

emmenko:
Let's leave it for another release and wait for Dom's feedback. I don't think it's super crucial to merge this.

@tdeekens go ahead with the release

dferber90:
@tdeekens Thanks for reminding me! Yes, I'd love to have this in.

packages/application-shell/src/components/quick-access/index.js
dferber90: What do you mean? Which Info?

emmenko: `nodeName !== 'A'`

dferber90: This is the regular DOM. You can do `document.querySelector('div').nodeName` and it will return `DIV`. Are you talking about a link to MDN? I wouldn't add that as it's the DOM itself.

The documentation is at https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeName

emmenko: Ah ok, I thought it was something else. Nvm then

dferber90: No worries, I had a suspicion it's just a misunderstanding :)

PR #20, opened 2018-10-18
chore(app-shell): define peer dependencies

### LINKED ISSUE #17: Make ui-kit a peerDep (at least in application-shell)
Problem

> After speaking to @malcolm we noticed something which wasn’t obvious at first. As `application-kit` and all apps consume `ui-kit` there is a bit of a hassle when releasing new ui-kit versions: When there is a new ui-kit release, we need to upgrade app-kit to the new version. Then we need to release the new app-kit and then we need to upgrade app-kit and ui-kit in all applications. Quite some work we didn’t really expect :/

> Let me try to explain the problem a bit more:
>
> The packages use the app-shell. They also use ui-kit.
>
> The app-shell uses the ui-kit.
>
> If we release a new version of the ui-kit and update the applications only, then they will end up with both versions of the ui-kit in the application - as the app uses the new ui-kit version while the  app-shell still uses the (pinned) old one.

Possible solution after talking to @emmenko:

> I think it would work if we made ui-kit a peerDep in app-shell and specify a range. Then we only have the problem on breaking releases which is already better!

The task would be to make `ui-kit` a peer dep in `application-shell` (and possibly in other packages) and to specify a version range instead of a pinned version.

emmenko:
Closes #17 

I also included other dependencies as peer dependencies, like we do in the ui-kit. WDYT?

### DIFF
--- a/packages/application-shell/package.json
+++ b/packages/application-shell/package.json
@@ -37,7 +37,6 @@
     "@commercetools-frontend/sdk": "1.0.0-rc.0",
     "@commercetools-frontend/sentry": "1.0.0-rc.0",
     "@commercetools-frontend/storage": "1.0.0-rc.0",
-    "@commercetools-frontend/ui-kit": "2.0.0-rc.11",
     "@commercetools-frontend/url-utils": "1.0.0-rc.0",
     "@flopflip/launchdarkly-adapter": "2.4.1",
     "@flopflip/memory-adapter": "1.1.0",
@@ -73,16 +72,10 @@
     "prop-types": "15.6.2",
     "query-string": "5.1.0",
     "raven-js": "3.27.0",
-    "react": "16.5.2",
-    "react-apollo": "2.2.4",
-    "react-dom": "16.5.2",
-    "react-intl": "2.7.1",
     "react-loadable": "5.5.0",
     "react-modal": "3.6.1",
     "react-redux": "5.0.7",
     "react-required-if": "1.0.3",
-    "react-router": "4.3.1",
-    "react-router-dom": "4.3.1",
     "react-select": "1.2.1",
     "react-testing-library": "5.2.0",
     "recompose": "0.26.0",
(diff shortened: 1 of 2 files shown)


emmenko:
image

packages/application-shell/package.json
```diff
@@ -94,6 +87,22 @@
...
+    "react": "16.5.2",
+    "react-apollo": "2.2.4",
+    "react-dom": "16.5.2",
+    "react-intl": "2.7.1",
+    "react-router": "4.3.1",
+    "react-router-dom": "4.3.1",
     "wait-for-observables": "1.0.3"
+  },
+  "peerDependencies": {
+    "@commercetools-frontend/ui-kit": ">=2 <3 || >=2.0.0-rc.1 <3",
```
tdeekens: My brain.

packages/application-shell/package.json
```diff
@@ -94,6 +87,22 @@
...
+    "react": ">=16",
```
tdeekens: Might want to have a guard against the biggest version too. This is like claiming: I work with anything from the future which may be odd.

packages/application-shell/package.json
emmenko: 😅 I added a screenshot with some `semver` tests in case you want to help your brain 😉

packages/application-shell/package.json
emmenko: Should we do that for every dep? Same then would apply to the ui-kit...

tdeekens: Not sure. Personally I would only claim the library to be working until the next major release. Then, given that comes out, test with the next RC and depending on that update the range.

dferber90: Yeah, I'd do it for all deps - including ui-kit. We should also stop with the `rc` releases and just do regular ones. We're past the `rc` phase of the migration and back to normal development there anyways.

emmenko: Alright, updated

PR #22, opened 2018-10-22
[application-shell]: Allow passing project key as variable on operation

tdeekens:
#### Summary

The project key is read from the url in a project related appliation context.
This holds for most applications like `application-categories`, `application-discounts` etc. However, the `application-account` does not run with the project key being part of the url. As a result we allow passing the project key as a variable on the operation allowing it to be the fallback.

### DIFF
--- a/packages/application-shell/src/apollo-links/header-link.js
+++ b/packages/application-shell/src/apollo-links/header-link.js
@@ -13,9 +13,18 @@ const headerLink = new ApolloLink((operation, forward) => {
       `GraphQL target "${target}" is missing or is not supported`
     );
 
-  const projectKey = selectProjectKeyFromUrl();
-  // NOTE: keep header names with capital letters to avoid possible conflicts
-  // or problems with nginx.
+  /**
+   * NOTE:
+   *   The project key is read from the url in a project related appliation context.
+   *   This holds for most applications like `application-categories`, `application-discounts` etc.
+   *   However, the `application-account` does not run with the project key being part of the url.
+   *   As a result we allow passing the project key as a variable on the operation allowing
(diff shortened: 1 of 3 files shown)


packages/application-shell/src/apollo-links/header-link.js
```diff
@@ -13,9 +13,18 @@ const headerLink = new ApolloLink((operation, forward) => {
...
+  /**
+   * NOTE:
+   *   The project key is read from the url in a project related appliation context.
+   *   This holds for most applications like `application-categories`, `application-discounts` etc.
+   *   However, the `application-account` does not run with the project key being part of the url.
+   *   As a result we allow passing the project key as a variable on the operation allowing
+   *   it to be the fallback.
+   */
+  const projectKey =
+    selectProjectKeyFromUrl() || operation.variables.projectKey;
```
emmenko: Should we do it the other way around? Prefer the `variables` option over the URL if given? Not 100% sure about it though...

dferber90: I would have expected the one from `variables` to take precedence as well.

montezume: I would go the other way around too

tdeekens: I had it like that before too. Wasn't sure either. I'll change it.

emmenko:
> Should/could we release this with the other standing PRs?

Sure thing. Would you like to do the honors or should I do that?

tdeekens:
I'd rather wait a bit for GitHub to settle. I can merge and publish. Sure.

PR #35, opened 2018-10-31
feat(connectors): introduce <ApplicationStateProvider> 📣

emmenko:
#### Summary

To make it easier to access "application state" information, instead of having multiple little connectors, we can have a single connector (provider/consumer pair) that exposes all the necessary information that are part of the application state (e.g. user, project, environment).

#### Description

One of the advantages of this connector is to have an easier setup for writing integration tests with react-testing-library.

NOTE: I dropped the other connectors, as I felt migrating them takes little effort. There will be some follow up tasks for each team then to start using this connector and replace e.g. (`FetchUser`, etc).

~~PS: I'll leave some inline comments to explain some things (going for lunch now though).~~


packages/application-shell-connectors/src/components/application-state/application-state.js
```diff
@@ -0,0 +1,98 @@
...
+import moment from 'moment-timezone';
+import omit from 'lodash.omit';
+
+const { Provider, Consumer } = React.createContext({});
+
+const defaultTimeZone = moment.tz.guess() || 'Etc/UTC';
+
+// Expose only certain fields as some of them are only meant to
+// be used internally in the AppShell
+const mapUserPropsToApplicationState = user => ({
```
emmenko: Should we memoize this function?

packages/application-shell-connectors/src/components/application-state/application-state.js
```diff
@@ -0,0 +1,98 @@
...
+  id: user.id,
+  email: user.email,
+  firstName: user.firstName,
+  lastName: user.lastName,
+  // NOTE: this is an alias for the original field `user.language`.
+  // This field is used for the application intl locale though, therefore
+  // we might want to expose it with a more explicit name to avoid
+  // confusion.
+  applicationLocale: user.language,
```
emmenko: I feel like exposing a more explicit field would make it easier to understand what it does.

packages/application-shell-connectors/src/components/application-state/application-state.js
```diff
@@ -0,0 +1,98 @@
...
+const mapProjectPropsToApplicationState = (project, projectDataLocale) => ({
+  key: project.key,
+  version: project.version,
+  name: project.name,
+  countries: project.countries,
+  currencies: project.currencies,
+  languages: project.languages,
+  permissions: omit(project.permissions, ['__typename']),
+  // Additional fields that do not belong directly to a project
+  ...(projectDataLocale ? { dataLocale: projectDataLocale } : {}),
```
emmenko: The "project data locale" is available within the `project` object.

image

packages/application-shell-connectors/src/components/application-state/application-state.js
```diff
@@ -0,0 +1,98 @@
...
+});
+
+const ApplicationStateProvider = props => (
+  <Provider
+    value={{
+      environment: props.environment,
+      ...(props.user
```
emmenko: As you will see in the `AppShell`, the `ApplicationStateProvider` will be rendered multiple times, depending on which objects are available.
For example, the `environment` will always be available, even if `user` or `project` are not. Or the `project` might not be available but the `user` is.

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -97,241 +96,253 @@ export const RestrictedApplication = props => (
...
-                      messages,
-                      i18n[language],
-                      props.applicationMessages[language]
-                    ),
-                  })}
-            >
-              <SetupFlopFlipProvider
-                user={user}
-                defaultFlags={props.defaultFeatureFlags}
+        <ApplicationStateProvider user={user} environment={props.environment}>
```
emmenko: As explained before, at this point we only know the `user` and the `environment`.

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -97,241 +96,253 @@ export const RestrictedApplication = props => (
...
-            <ConfigureIntlProvider
-              // We do not want to pass the language as long as the locale data
-              // is not loaded.
-              {...(isLoadingLocaleData
-                ? {}
-                : {
-                    // We need to pass the value as `undefined` in case the user
-                    // has no `timeZone` defined so the `defaultProps` in the
-                    // `<Context.Provider>` kick in.
-                    timeZone: user && user.timeZone ? user.timeZone : undefined,
```
emmenko: This is not necessary anymore as it will be available in the new provider

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -97,241 +96,253 @@ export const RestrictedApplication = props => (
...
+                        i18n[language],
+                        props.applicationMessages[language]
+                      ),
+                    })}
               >
-                <React.Fragment>
-                  {/* NOTE: the requests in flight loader will render a loading
+                <SetupFlopFlipProvider
+                  user={user}
+                  appEnv={props.environment.env}
```
emmenko: We can now simply pass it as a prop instead of connecting from inside the component

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -97,241 +96,253 @@ export const RestrictedApplication = props => (
...
+                                    <ProjectDataLocale
+                                      locales={project.languages}
+                                    >
+                                      {({ locale, setProjectDataLocale }) => (
+                                        <ApplicationStateProvider
+                                          user={user}
+                                          project={project}
+                                          projectDataLocale={locale}
+                                          environment={props.environment}
+                                        >
```
emmenko: At this point, we know also the `projectKey`, so we put all the info in the provider again

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -97,241 +96,253 @@ export const RestrictedApplication = props => (
...
+                                  isLoadingProject
+                                )
+                                  return <LoadingNavBar />;
+
+                                return (
+                                  <ApplicationStateProvider
+                                    user={user}
+                                    project={project}
+                                    environment={props.environment}
+                                  >
```
emmenko: For the `<Navbar>`, we do not need to know the `projectDataLocale`

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -97,241 +96,253 @@ export const RestrictedApplication = props => (
...
+                                    <ReconfigureFlopFlip
+                                      user={getFlopflipReconfiguration(
+                                        routerProps.match.params.projectKey
+                                      )}
+                                    />
+                                    <ProjectContainer
+                                      user={user}
+                                      match={routerProps.match}
+                                      location={routerProps.location}
+                                      environment={props.environment}
```
emmenko: We pass the `environment` here so that we can pass it to the provider rendered within the `<ProjectContainer>`

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -382,7 +393,7 @@ UnrestrictedApplication.displayName = 'UnrestrictedApplication';
 export default class ApplicationShell extends React.Component {
   static displayName = 'ApplicationShell';
   static propTypes = {
-    configuration: PropTypes.object.isRequired,
```
emmenko: I renamed this as `configuration` is a bit too generic and maybe even confusing

packages/application-shell-connectors/src/components/application-state/application-state.js
tdeekens: I'd like to hear @dferber90's thoughts cause he's put some 💭 into the naming of languages across the app.

packages/application-shell-connectors/src/components/application-state/application-state.js
tdeekens: Once we thought `resourceLocale` might be a good name. Not sure that's still true.

packages/application-shell-connectors/src/components/application-state/application-state.js
```diff
@@ -0,0 +1,98 @@
...
+  currencies: project.currencies,
+  languages: project.languages,
+  permissions: omit(project.permissions, ['__typename']),
+  // Additional fields that do not belong directly to a project
+  ...(projectDataLocale ? { dataLocale: projectDataLocale } : {}),
+});
+
+const ApplicationStateProvider = props => (
+  <Provider
+    value={{
```
tdeekens: I would suggest to memoize this. It's often said to be a good practice. Also this whole spreading down there makes me feel a bit unease given how far up the component tree we are.

packages/application-shell-connectors/src/components/application-state/application-state.js
tdeekens: > depending on which objects are available.

Until they become available (by other components) and are passed or never will. Given e.g. we never have a project in one app.

packages/application-shell-connectors/src/components/application-state/application-state.js
```diff
@@ -0,0 +1,98 @@
...
+
+const GetApplicationState = props => (
+  <Consumer>{applicationState => props.render(applicationState)}</Consumer>
+);
+GetApplicationState.displayName = 'GetApplicationState';
+GetApplicationState.propTypes = {
+  render: PropTypes.func.isRequired,
+};
+
+const withApplicationState = (propKey = 'applicationState') => Component => {
```
dferber90: I wonder whether we should give the option to rename the prop. It might be easier to remove this. Then the application code becomes more consistent. If anybody really needs to rename the prop for some reason, they can use `renameProp` from `recompose`.

Further, if we give the option to rename now, it becomes more complicated later to add a `connect`-style `mapStateToProps` pattern - which I'd prefer to have in case we really need it.

Something like

```js
withApplicationState(
  applicationState => ({ userId: applicationState.user.id }),
)(SomeComponent)
```

I'm not saying we should build this now. My point is to remove the ability to rename the prop so that 1) consumers are more consistent and 2) it becomes easier to add appropriate functionality later on once we learned more about the use-cases for `withApplicationState`.

packages/application-shell-connectors/src/components/application-state/README.md
```diff
@@ -0,0 +1,31 @@
+# Connectors for `applicationState`
+
+This component exposes getters/setters to set and retrieve the `applicationState`.
+
+> It uses the new React Context API, so you need to provide the `ApplicationStateProvider` up in the tree. This is done within the AppShell!
+
+Additionally it provides a HOC to inject the `applicationState`.
+
+Use this component to access about the `user`, `project` and application `environment`.
```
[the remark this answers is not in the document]
dferber90: *Update*: After reading the code I now get that it's "just" `window.app`. A comment here explaining that it holds that value - and why one would use the consumer instead of directly accessing `window.app` might be useful. I'd find it useful if the comment could also explain which values `window.app` contains and what they're used for.

packages/application-shell-connectors/src/components/application-state/application-state.js
emmenko: Good point, I'll refactor that 👌

packages/application-shell-connectors/src/components/application-state/application-state.js
emmenko: Alright, I'm fine with `user.locale`. I though that naming it `applicationLocale` would make it more explicit.

packages/application-shell-connectors/src/components/application-state/README.md
```diff
@@ -0,0 +1,31 @@
...
+<ApplicationStateProvider
+  user={{...}}
+  project={{...}}
+  projectDataLocale="en"
+  environment={{...}}
+>
+  <div>
+    {/* ... */}
+    {/* In the application specific code */}
+    <GetApplicationState
```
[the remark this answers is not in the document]
emmenko: I personally like more the `with` for HOCs (because we are **injecting** something) and `Get` for components, because it's a getter after all.

But yeah, it's a subtle difference. If it's important to settle on a name, we should then write an ADR or something for future implementations. WDYT?

packages/application-shell-connectors/src/components/application-state/README.md
```diff
@@ -0,0 +1,31 @@
+# Connectors for `applicationState`
+
+This component exposes getters/setters to set and retrieve the `applicationState`.
+
+> It uses the new React Context API, so you need to provide the `ApplicationStateProvider` up in the tree. This is done within the AppShell!
```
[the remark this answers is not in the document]
emmenko: Good point!

packages/application-shell-connectors/src/components/application-state/README.md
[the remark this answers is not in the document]
emmenko: Yeah, I can list some of the values, however, `window.app` can contain application specific values which does not make sense to document.

packages/application-shell-connectors/src/components/application-state/application-state.js
```diff
@@ -0,0 +1,98 @@
...
+  id: user.id,
+  email: user.email,
+  firstName: user.firstName,
+  lastName: user.lastName,
+  // NOTE: this is an alias for the original field `user.language`.
+  // This field is used for the application intl locale though, therefore
+  // we might want to expose it with a more explicit name to avoid
+  // confusion.
+  applicationLocale: user.language,
+  numberFormat: user.numberFormat,
```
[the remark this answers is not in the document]
emmenko: Ah true, I think we deprecated this. Good catch!

packages/application-shell-connectors/src/components/application-state/application-state.js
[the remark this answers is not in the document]
emmenko: Sure, it sounds reasonable to me 👌

packages/application-shell-connectors/src/components/application-state/application-state.js
```diff
@@ -0,0 +1,98 @@
...
+        ? { user: mapUserPropsToApplicationState(props.user) }
+        : {}),
+      ...(props.project
+        ? {
+            project: mapProjectPropsToApplicationState(
+              props.project,
+              props.projectDataLocale
+            ),
+          }
+        : {}),
```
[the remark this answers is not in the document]
emmenko: Alright, I wasn't 100% sure about that. I can also document that values might be `null`.

packages/application-shell-connectors/src/components/application-state/README.md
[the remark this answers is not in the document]
tdeekens: I don't think it's crucial. Just wanted to mention it. Various libraries go different about it anyway: `injectIntl`, `injectFeatureToggle`, `withRouter`. I think we can agree that `with*` is good for the HoC. `injectIntl`, if I recall correctly, renders a `WithIntl`. I am fine with the `Get*` just haven't seen it anywhere else (as in other libraries) and was curious as to why we deviate.

packages/application-shell-connectors/src/components/application-state/application-state.js
```diff
@@ -0,0 +1,98 @@
...
+);
+ApplicationStateProvider.displayName = 'ApplicationStateProvider';
+ApplicationStateProvider.propTypes = {
+  user: PropTypes.object,
+  // NOTE: project (and dataLocale) might be eventually undefined because
+  // in some views (e.g. accounts) we are not in a project context anymore.
+  project: PropTypes.object,
+  projectDataLocale: PropTypes.string,
+  // This is the environment configuration coming from `windows.app`
+  environment: PropTypes.object.isRequired,
```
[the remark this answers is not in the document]
emmenko: Like I said, this is app specific. There are some common keys though that I can document 👌

packages/application-shell-connectors/src/components/application-state/application-state.js
emmenko: Ah yeah good point. Actually I would allow to pass a "map" function like we do in `FetchUser` and `FetchProject`. Thanks for reminding me that 👌

emmenko:
@tdeekens @dferber90 thanks a lot for the first round of reviews! I'll apply some changes and ping you again when it's ready for another review.

---

> Lets refactor the app to this in a team of 1-2 [...]

@tdeekens the idea was to only migrate the "deprecated" connectors first (since they have been removed) once we release this.
The migration for other parts (`globalAppState`, `FetchProject`, etc) can be done e.g. by each team separately.

> Not sure this was discussed: we remove our `Fetch{User,Project}` and encapsulate this into the provider?

We could do that but I think it would complicate things a bit. For example, we don't expose every field from the `user` and `project` objects but only a subset of them (hence the mapping functions). However, some of those fields that are not exposed (e.g. `launchdarklyTrackingId`, project `suspension`, etc) are used within the AppShell and therefore accessible directly from the `FetchUser` and `FetchProject` components.

So given that, do we still want to include everything into the Provider? We then loose the ability to only expose useful fields...or do you have other suggestions?

tdeekens:
> So given that, do we still want to include everything into the Provider?

I forgot about that part, sorry. I think it's worthy to "only" expose meaningful fields down the application through the provider. It's also fine by me to split between a provider/consumer pair and a fetching component.

packages/application-shell-connectors/src/components/application-state/README.md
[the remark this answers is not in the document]
emmenko: Well yeah, I would say it's kind of our own convention. I think as long as it makes sense and it's clear for the consumers on what it does, it's fine to use it.

Also, those `inject` functions have been there for a long time, before all those new patterns started to come out (e.g. render props, provider/consumer, etc).

But again, this is just my personal taste, I'm fine also if we agree on something more "standard".

emmenko:
Ok, but we can still have the Fetch components within the provider. Only some parts of the AppShell will need to access `Fetch{User,Project}` directly to get the "internal" fields.
Also, we can deprecate the usage of `Fetch{User,Project}` from outside the AppShell and eventually do not expose those anymore and only use them within the AppShell.

emmenko:
Alright, I think this can be reviewed again. In the last commit (https://github.com/commercetools/merchant-center-application-kit/pull/35/commits/9d28a348868cb3e74bb65250dcfba274e9ada618) I also deprecated the usage of `Fetch{User,Project}` components from outside the AppShell.

Let me know if I need to do more changes. Thanks 🙏

packages/application-shell/src/from-core/deprecate-component/deprecate-component.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
```
tdeekens: I wouldn't mind not having `from-core` in the path. We can duplicate this without feeling "bad" I guess 😄.

emmenko: 😇

Where should I put it then? It's not really an AppShell specific component, it can also be used by other packages...

tdeekens: Oh sorry. I wasn't clear. Keep it in this repo. I just found the `from-core` path strange. I'd put it where it's needed to into the `applications-shell` package. I just got confused by having a new package called `from-core`.

packages/application-shell-connectors/src/components/application-state/README.md
```diff
@@ -0,0 +1,74 @@
...
+```js
+withApplicationState()(MyComponent);
+```
+
+...or pass a mapping function as the first argument to return custom shape of the injected props
+
+```js
+withApplicationState(applicationState => ({
+  projectKey: applicationState.project && applicationState.project.key,
+  userEmail: applicationState.user && applicationState.user.email,
```
tdeekens: Nitpick: `applicationState.user?.email` so others use optional chaining and do not copy this. Not incredibly important obviously.

packages/application-shell-connectors/src/components/application-state/application-state.js
```diff
@@ -0,0 +1,134 @@
...
+    key: PropTypes.string.isRequired,
+    version: PropTypes.number,
+    name: PropTypes.string.isRequired,
+    countries: PropTypes.arrayOf(PropTypes.string).isRequired,
+    currencies: PropTypes.arrayOf(PropTypes.string).isRequired,
+    languages: PropTypes.arrayOf(PropTypes.string).isRequired,
+    permissions: PropTypes.object.isRequired,
+    // ...plus other fields that we don't want to expose
+  }),
+  projectDataLocale: PropTypes.string,
```
tdeekens: Is this still correct? I thought it was `project.dataLocale`?

packages/application-shell/src/from-core/deprecate-component/deprecate-component.js
emmenko: Ah ok. Well, the thing with those components is that they are copy-pasted from core. So in case we need to change something there, we should also replicate the changes in core (and the other way around). I think until we find a more maintainable solution (e.g. some components into ui-kit, other components somewhere else) we should keep it like this so we know which components need to be migrated somewhere.

packages/application-shell-connectors/src/components/application-state/README.md
emmenko: I always forget that we can now use it 🤦‍♂️ Pardon me

packages/application-shell-connectors/src/components/application-state/application-state.js
emmenko: This is the prop that we pass to the provider. The application state will contain `project.dataLocale` though...I know, it's a bit confusing.

packages/application-shell-connectors/src/components/application-state/README.md
dferber90: I'm not even sure I would add it since this documentation is for outsiders as well, who might not have the proposal transform.

emmenko: Also a good point. The proposal is also enabled in our babel preset though...

I don't have a strong opinion on this, you guys can decide

tdeekens: I dont mind. i preaume ppls building UIKit would use our preset at a minimum base often times.

emmenko:
@dferber90 any last wishes? How would you like to proceed next?

emmenko:
Alright, I'm going to merge this so we can iterate on this

packages/application-shell-connectors/src/components/application-state/README.md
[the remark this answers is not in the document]
dferber90: Sure :) My intention was to document that it contains `window.app` and why one would use `environment` over just accessing `window.app` directly - and to mention that the values can be custom as well as mentioning the known ones.

packages/application-shell/src/from-core/deprecate-component/deprecate-component.js
dferber90: I like that we are documenting that they are duplicates from `core`. However I would add a comment to the file which explains this instead of using the folder hierarchy to communicate it. This allows us to place the components in the folder which fits them best and makes it even more obvious that they're duplicates.

packages/application-shell-connectors/src/components/application-state/application-state.js
dferber90: I also got confused by this at first but it's okay to keep it around.

What I think we should do is to enable storing the `dataLocale` per project. Right now there is only one `dataLocale` and it is set for all projects - even when the projects might use different languages. So `localStorage` contains `{ dataLocale: 'en' }`.

Instead, I'd store one `dataLocale` for each project in `localStorage` like

`{ dataLocaleByProjectKey: { 'test-with-big-data-44': 'en', 'random-project-2': 'de' } }`.

This is a new feature of course, but it would also help to clean up this code as we'd be passing `dataLocaleByProjectKey` to `ApplicationStateProvider`.

But this would needs some user-research first to determine whether having a per-project data-locale setting is something users actually want.

dferber90:
@emmenko I appreciate the urge to move forward, but 22 hours to review a PR as big and essential as this one is pretty rushed 😢

adnasa:
what the hell, I missed out on all the fun... _6 days later_

packages/application-shell/src/components/application-shell/application-shell.js
emmenko: Ah, nice catch!!

PR #40, opened 2018-11-02
Add bundling (rollup)

montezume:
So far we are shipping untranspiled/unbundled code. This meant that our consumers had to have the right webpack loaders in place in order to consume our different packages.

This PR makes it so that the application-kit browser code ships as standalone libraries, which can be consumed by any project which has the peerDependencies like react and react-dom installed.

### Usage

A new command was added:
- `yarn build`: New command to build the 14 packages.

### Consumers

Consumers should only use our libraries by using named exports from the main export:

```js
import { Avatar } from '@commercetools-frontend/application-shell'
```

The main export will serve all components. We won't support diving / proxying for exports.

### CSS

We are not creating external CSS files at the moment. All CSS (global and modular) will be bundled in the built applications ESM or CJS file.

### TODO

* Work more with code splitting, as I had to (for now) remove the code split in `QuickAccess`
* Test in MC-FE


.gitignore
```diff
@@ -67,8 +67,9 @@ typings/
 dist
 public
 
+**/.DS_Store
```
montezume: This was missing

packages/application-shell/src/components/quick-access/index.js
```diff
@@ -1,13 +1,14 @@
...
+// import ButlerContainer from './butler-container';
 import * as gtm from '../../utils/gtm';
 import trackingEvents from './tracking-events';
-
-const QuickAccessModal = Loadable({
-  loader: () => import('./quick-access' /* webpackChunkName: "quick-access" */),
-  loading: ButlerContainer,
-});
+import QuickAccessModal from './quick-access';
+// TODO: make this work in different environments.
```
montezume: one way to deal with this would be to set an env variable, and change the path based on that.

the current issue is that `./quick-access` doesn't exist in the context of `dist`

packages/i18n/load-i18n/load-i18n.js
```diff
@@ -9,16 +9,16 @@ export default function loadI18n(lang) {
 
   const reactIntlChunkImport = import(/* webpackChunkName: "react-intl-data-[request]" */
   /* webpackMode: "lazy" */
-  `./locale-data/${lang}`);
+  `../locale-data/${lang}`);
```
montezume: just wanted to change the directory structure so that in both cases, data was one folder down.

packages/l10n/country-information/country-information.js
```diff
@@ -16,7 +16,7 @@ const getCountriesForLocale = (locale, cb) => {
   // The files are named like "country-data-en-json.chunk.js" after compilation
   // https://webpack.js.org/api/module-methods/#import-
   import(/* webpackChunkName: "country-data-[request]" */
-  `./data/countries/${supportedLocale}.json`)
+  `../data/countries/${supportedLocale}.json`)
```
montezume: these "webpackChunkNames" are most likely removed by rollup. We should look into a way to preserve them.

packages/l10n/register-moment-locales.js
```diff
@@ -1,22 +0,0 @@
-// We need to import the moment locales (language + country) for each
-// supported language (en, de).
```
montezume: This file isn't used anymore.

packages/l10n/country-information/country-information.js
tdeekens: What if we just preserve comments?

packages/l10n/index.js
```diff
@@ -0,0 +1,13 @@
...
+export { timeZonesShape, withTimeZones } from './time-zone-information';
+export {
+  default as extractLanguageFromLocale,
+} from './utils/extract-language-from-locale';
+export {
+  default as getSupportedLanguage,
+} from './utils/get-supported-language';
+export {
+  default as getSymbolFromCurrency,
+} from './utils/get-symbol-from-currency';
```
tdeekens: I like that we generally move to a root export. Has been easier to work with as the location of a module inside of a package should be implementation detail to it. Ultimately it means less breaking changes.

packages/l10n/package.json
```diff
@@ -19,7 +21,12 @@
     "access": "public"
   },
   "scripts": {
-    "generate-data": "node ./scripts/generate-l10n-data.js"
+    "generate-data": "node ./scripts/generate-l10n-data.js",
+    "prebuild": "rimraf dist/**",
+    "build": "cross-env npm run build:es && npm run build:cjs",
+    "build:watch": "cross-env npm run build:es -- -w",
+    "build:cjs": "cross-env NODE_ENV=development rollup -c ../../rollup.config.js  -f cjs ./index.js -o dist/l10n.cjs.js",
```
tdeekens: Nice that we keep the config simple and do the rest with CLI options.

rollup.config.js
```diff
@@ -0,0 +1,99 @@
...
+    name: pkg.name,
+    sourcemap: true,
+    globals: {
+      react: 'React',
+      redux: 'redux',
+      'react-redux': 'react-redux',
+    },
+  },
+  external: deps.concat(peerDeps).concat(deps),
+  plugins: [
```
tdeekens: Can you elaborate where if we the babel.macro where not and why?

montezume: You mean adding [graphql-tag.macro](https://github.com/leoasis/graphql-tag.macro)? Yeah, we could do that. 👍

montezume: The babel config still need work

emmenko:
First if all, thanks for looking into this topic!

I’d like to focus on one point though: code splitting. I think removing it is not an option,I think we can all agree on that.
And maybe we are approaching this topic in the wrong way.
For instance, is Rollup the right tool for the AppShell package? It might be for most of the other packages but for the AppShell maybe it’s not (E.g.: if code splitting breaks).
Would that work for example if we simply use babel to ship transpired code? When we build the apps, Webpack can then do code splitting and bundle the rest of the code.

Let’s talk about this first, as I feel this is one of the most crucial points.

tdeekens:
I value this PR, the experiment and effort a lot. I think we all agree that building this using Rollup would be nice. Let's get a bit more strategic towards the code splitting issue then. 

1. We seem to have about 18-20 split points
2. Most of those are around login and certain project status pages (suspended, expired) 
3. Most heavy split points could be: country, language and time zone information

Given that I think keeping code splitting is a good goal. A couple of months ago I read that rollup actually supports it through `experimentalCodeSplitting` and `experimentalDynamicImport`. I guess things are still behind those config options [reference](https://medium.com/rollup/rollup-now-has-code-splitting-and-we-need-your-help-46defd901c82). Given this does not work (I presume it has been tired) we can use some clever env variable injection or some babel or rollup rewiring (e.g. [this](https://github.com/speedskater/babel-plugin-rewire).

Generally I think using the same bundling process across repos: UIKit and mc-app-kit could have benefits also if it's not 100% strictly needed. Additionally I feel that this pattern of main index.js with re-exports from any library (not strictly related to this) has benefits as it encapsulates a library nicely. I've enjoyed using UIKit with it far more as I don't have to remember the implementation detail of the library where its modules are. This is/can obviously a separate thing to tackle before or with bundling in whatever form.

> Note: I just saw @montezume responded after I submitted.

emmenko:
The important thing after all is that we can make it work 😜

I’m just not sure if the process is “correct”:

```
transpile (rollup) -> split bundles -> pkg -> webpack -> split bundles
```

As opposed to

```
transpile (???) -> pkg -> webpack -> split bundles
```

Maybe I’m simply not understanding something but that’s where I’m confused at the moment. 🙃

—-

PS: as for the main exports, I obviously agree of it being a useful change 👌

emmenko:
Thanks for the explanation. This stuff is still a bit fuzzy to me, so I'm raising those questions because I'm obviously not fully understanding how this works.

The important part is that we have code splitting properly work 😇

Alright, so how should we proceed here?

PS: should we maybe separate the "main exports" refactoring from the bundling config in separate PRs? /cc @montezume

montezume:
@emmenko that sounds good to me. I'd like to start with

1) #41  
2) PR for the "main exports refactoring"
3) come back to rollup

tdeekens:
>  This stuff is still a bit fuzzy to me, so I'm raising those questions because I'm obviously not fully understanding how this works.

Glad it helped and no worries. It's super cool that we're all learning more and more about bundling and rollup together. I find it neat that rollup becomes our defacto bundler: Node.JS SDK, UIKit and _maybe_ app-kit. That means knowledge is really transferable.

packages/l10n/country-information/country-information.js
montezume: Good news, rollup doesn't remove them 💯

rollup.config.js
```diff
@@ -0,0 +1,104 @@
...
+const postcssDiscardComments = require('postcss-discard-comments');
+const postcssImport = require('postcss-import');
+const postcssPresetEnv = require('postcss-preset-env');
+const postcssReporter = require('postcss-reporter');
+
+const { pkg } = readPkgUp.sync({
+  cwd: fs.realpathSync(process.cwd()),
+});
+
+const format = process.env.npm_lifecycle_event.split(':')[1];
```
dferber90: Could you add a comment to explain what this is?

tdeekens: Sorry. Wrong button. 

When bundling for CJS we do not want to generate named exports in the JSON plugin. There seems to be no other way to read the CLI arguments passed to Rollup. This uses the npm lifecycle event which is the run script‘s name to deduct what format we are building. In that it is a convention to deduct it from the name. Ive also not found another way and did this on flopflip too. Maybe a comment would help in addition we could throw if we cant deduct a known format here to avoid hard to trace bundling errors later.

package.json
```diff
@@ -22,7 +22,9 @@
     "test": "jest --config jest.test.config.js",
     "test:watch": "jest --config jest.test.config.js --watch",
     "playground:start": "cd playground; yarn start",
-    "playground:build": "cd playground; yarn build"
+    "playground:build": "cd playground; yarn build:playground",
```
tdeekens: We can use yarn‘s build in —cwd if I am not mistaken. Alternatively pushd and popd could be nice just to be able to go back again. Not even sure we ever have to. 😇

packages/application-shell/package.json
```diff
@@ -18,12 +18,19 @@
...
+  "main": "dist/cjs/index.js",
+  "module": "dist/es/index.js",
   "engines": {
     "node": ">=8",
     "npm": ">=5"
   },
+  "scripts": {
+    "prebuild": "rimraf dist/**",
+    "build": "cross-env npm run build:es && npm run build:cjs",
+    "build:watch": "cross-env npm run build:es -- -w",
```
tdeekens: Maybe this should be build:es:watch. Not super sure if we ever want to build anything but es in watchmode but ar least it would be clear what the output was.

rollup.config.js
```diff
@@ -0,0 +1,104 @@
...
+};
+
+const config = {
+  output: {
+    name: pkg.name,
+    sourcemap: true,
+    globals: {
+      react: 'React',
+      redux: 'redux',
+      'react-redux': 'react-redux',
```
tdeekens: Why do they need to be globals if we exclude deps from being bundled after all. I think I knew why once but can‘t recall.

.travis.yml
```diff
@@ -14,6 +14,7 @@ cache:
     - node_modules
 
 script:
+  - yarn build
```
emmenko: I would run this after the tests/linters

package.json
emmenko: True, I just double checked. It's possible to use `--cwd` 💯

packages/actions-global/package.json
```diff
@@ -18,8 +18,15 @@
   "publishConfig": {
     "access": "public"
   },
-  "main": "./index.js",
-  "module": "./index.js",
+  "main": "dist/actions-global.cjs.js",
+  "module": "dist/actions-global.es.js",
+  "scripts": {
+    "prebuild": "rimraf dist/**",
+    "build": "cross-env npm run build:es && npm run build:cjs",
```
emmenko: Remove `cross-env`

packages/actions-global/package.json
```diff
@@ -18,8 +18,15 @@
...
+    "build:watch": "cross-env npm run build:es -- -w",
+    "build:cjs": "cross-env NODE_ENV=production rollup -c ../../rollup.config.js  -f cjs ./index.js -o dist/actions-global.cjs.js",
+    "build:es": "cross-env NODE_ENV=production rollup -c ../../rollup.config.js  -f es ./index.js -o dist/actions-global.es.js"
+  },
```
emmenko: We should also provide the `file` option to determine which files gets published. We also do it in the ui-kit (https://github.com/commercetools/ui-kit/blob/8250ebca39ec2a85d8d0d1839013b16cb28b0001/package.json#L18-L29)

packages/application-shell-connectors/package.json
```diff
@@ -18,8 +18,15 @@
...
-  "module": "./src/index.js",
+  "scripts": {
+    "prebuild": "rimraf dist/**",
+    "build": "cross-env npm run build:es && npm run build:cjs",
+    "build:watch": "cross-env npm run build:es -- -w",
+    "build:cjs": "cross-env NODE_ENV=production rollup -c ../../rollup.config.js  -f cjs ./src/index.js -o dist/application-shell-connectors.cjs.js",
+    "build:es": "cross-env NODE_ENV=production rollup -c ../../rollup.config.js  -f es ./src/index.js -o dist/application-shell-connectors.es.js"
+  },
+  "main": "./dist/application-shell-connectors.cjs.js",
+  "module": "./dist/application-shell-connectors.es.js",
```
emmenko: Nitpick: can we keep the same order for the keys in all package jsons? E.g. `main` and `module` come after `publishConfig` 😇 🙏

packages/application-shell/package.json
```diff
@@ -18,12 +18,20 @@
   "publishConfig": {
     "access": "public"
   },
-  "main": "./src/index.js",
-  "module": "./src/index.js",
+  "main": "dist/cjs/index.js",
+  "module": "dist/es/index.js",
```
emmenko: Shouldn't this be `dist/application-shell.cjs.js` and `dist/application-shell.es.js`

packages/i18n/package.json
```diff
@@ -18,13 +18,20 @@
...
+  },
   "dependencies": {
-    "@commercetools-frontend/i18n": "1.0.0-rc.3",
-    "@commercetools-frontend/sentry": "1.0.0-rc.3",
+    "@commercetools-frontend/sentry": "1.0.0-rc.2",
     "moment": "2.22.2",
     "prop-types": "15.6.2",
-    "react": "16.6.0"
+    "react": "16.6.0",
+    "react-intl": "2.7.1"
```
emmenko: Uh, those should be defined as peer dependencies. I guess I missed that last time #34 🙈

packages/l10n/README.md
```diff
@@ -28,6 +28,8 @@ withCurrencies(ownProps => ownProps.locale)(Component);
 // { "EUR": { "label": "Euro", "symbol": "€" } }
 ```
 
+- `@commercetools-frontend/l10n`
```
emmenko: 🤔

playground/package.json
```diff
@@ -4,9 +4,9 @@
   "description": "A starter example for building a Merchant Center application",
   "private": true,
   "scripts": {
-    "build": "mc-scripts build",
+    "build:playground": "mc-scripts build",
```
emmenko: Hmm why did you rename this? 🤔

rollup.config.js
```diff
@@ -0,0 +1,104 @@
+const fs = require('fs');
+const babel = require('rollup-plugin-babel');
+const readPkgUp = require('read-pkg-up');
+const resolve = require('rollup-plugin-node-resolve');
+const json = require('rollup-plugin-json');
+const commonjs = require('rollup-plugin-commonjs');
+const postcss = require('rollup-plugin-postcss');
+const peerDeps = require('rollup-plugin-peer-deps-external');
+const getBabelPreset = require('@commercetools-frontend/babel-preset-mc-app');
```
emmenko: Nitpick: move it after/before all the "rollup" requires, just to order the imports and have then more readable 😇

playground/package.json
emmenko: I see...well, on CI we build the playground anyway. So maybe it's not a bad idea to build it together with the other packages?

emmenko: PS: you can always "exclude" some packages from running lerna commands

emmenko: So that would be `lerna run build --ignore "merchant-center-starter-app"`

dferber90: I would keep `build` for building the package. Building an "example" like playground is better suited as its own command imho.

package.json
```diff
@@ -21,8 +21,8 @@
     "release": "lerna publish --exact --preid rc",
     "test": "jest --config jest.test.config.js",
     "test:watch": "jest --config jest.test.config.js --watch",
-    "playground:start": "cd playground; yarn start",
-    "playground:build": "cd playground; yarn build:playground",
+    "playground:start": "yarn --cwd playground start",
+    "playground:build": "yarn --cwd playground; yarn build:playground",
```
tdeekens: Nitpick: can't we just `yarn build:playround --cwd playground`?

emmenko: This should be `yarn build --cwd playground`

playground/package.json
emmenko: I would like to keep `build`, because this is meant to be also an example on how to setup a project. I provided a couple of options on how to deal with this:
- ignore the playground package from the `lerna run build` command
- let it also build this package

dferber90: Ah, sorry I got confused. I thought `playground` is nested inside `application-shell`! I'm on board then :)

package.json
montezume: Actually it should be `yarn  --cwd playground build`

playground/package.json
montezume: Done in [637015263b0534fe1d3d047008032cf93574e67e](https://github.com/commercetools/merchant-center-application-kit/pull/40/commits/637015263b0534fe1d3d047008032cf93574e67e)

packages/application-shell/package.json
montezume: Have updated this, now both cjs and es live in the same directory.

packages/application-shell-connectors/package.json
montezume: Yeah went through and did this. Also added "files"

.travis.yml
montezume: Some of the tests rely on the bundle though? All of the tests in `playground` rely on it, and they are being ran as well. [Here](https://github.com/commercetools/merchant-center-application-kit/blob/master/playground/src/components/channels-list/channels-list.spec.js) for example.

If this runs without the bundle, you get errors like 

```
    Cannot find module '@commercetools-frontend/application-shell' from 'entry-point.spec.js'
```
We could split the test run in two I guess?

emmenko: Hmm what do you mean? Nothing should rely on the bundle...I'm confused

montezume: The `playground` relies on the bundle right? So it stands to reason that any test in `playground`, requires the bundle to be built.

package.json
```diff
@@ -21,8 +21,10 @@
     "release": "lerna publish --exact --preid rc",
     "test": "jest --config jest.test.config.js",
     "test:watch": "jest --config jest.test.config.js --watch",
-    "playground:start": "cd playground; yarn start",
-    "playground:build": "cd playground; yarn build"
+    "playground:start": "yarn --cwd playground start",
+    "playground:build": "yarn --cwd playground build",
+    "build": "NODE_ENV=production lerna run build --ignore 'merchant-center-starter-app'",
+    "build:es:watch": "NODE_ENV=development lerna run --parallel build:es:watch"
```
emmenko: I guess we should also ignore the playground here

packages/application-shell-connectors/index.js
```diff
@@ -0,0 +1 @@
+export * from './src';
```
emmenko: Nitpick: should we maybe add a comment to those files explaining the reason for it?

packages/application-shell/package.json
```diff
@@ -18,11 +18,21 @@
...
+  "files": [
+    "dist",
+    "package.json",
+    "LICENSE",
+    "README.md"
+  ],
+  "scripts": {
+    "prebuild": "rimraf dist/**",
+    "build": "npm run build:es && npm run build:cjs",
+    "build:cjs": "cross-env NODE_ENV=production rollup -c ../../rollup.config.js  -f cjs ./index.js --dir ./dist --chunkFileNames application-shell-[name]-[hash].cjs.js --entryFileNames application-shell-[name].cjs.js --experimentalCodeSplitting",
```
emmenko: 💯

packages/application-shell/package.json
```diff
@@ -18,11 +18,21 @@
...
+    "build:es": "cross-env NODE_ENV=production rollup -c ../../rollup.config.js  -f es ./index.js --dir ./dist --chunkFileNames application-shell-[name]-[hash].es.js --entryFileNames application-shell-[name].es.js --experimentalCodeSplitting",
+    "build:es:watch": "npm run build:es -- -w",
+    "test:build:es": "cross-env NODE_ENV=development rollup -c ../../rollup.config.js  -f es ./src/test-utils.js --o ./dist/test-utils/index.js"
```
emmenko: Is this used somewhere? 🤔

packages/i18n/package.json
```diff
@@ -18,13 +18,28 @@
...
+  },
   "dependencies": {
-    "@commercetools-frontend/i18n": "1.0.0-rc.3",
     "@commercetools-frontend/sentry": "1.0.0-rc.3",
     "moment": "2.22.2",
-    "prop-types": "15.6.2",
-    "react": "16.6.0"
+    "prop-types": "15.6.2"
+  },
+  "peerDependencies": {
```
emmenko: 💯

playground/webpack.config.dev.js
```diff
@@ -5,7 +5,6 @@ const distPath = path.resolve(__dirname, 'dist');
 const entryPoint = path.resolve(__dirname, 'src/index.js');
 const sourceFolders = [
   path.resolve(__dirname, 'src'),
-  path.resolve(__dirname, '../packages'),
```
emmenko: 🎉

rollup.config.js
emmenko: Are those comments still relevant?

montezume:
@emmenko the postcss ones? Yeah, they are until we move to the next version of UI-Kit.

packages/application-shell/package.json
montezume: @dferber90 and I decided that it didn't make sense to add `src/test-utils.js` to the main export. Now users will access it in their code like 
```import testUtils from '@commercetools-frontend/applicatoin-shell/test-utils```

packages/application-shell-connectors/index.js
montezume: good idea 👍

packages/application-shell/package.json
montezume: Ah, but it should be added to the main build as well. WOops

dferber90: This is the same approach as `react-apollo` uses btw.

dferber90: I would rename this script to `build:test-utils:es` because the current name kinda looks like this would run some tests.

dferber90: I would add this command to `yarn build`  as well, so that the full package gets built with `yarn build`.

emmenko: > Ah, but it should be added to the main build as well

Yeah I meant that 👌

montezume: done

PR #50, opened 2018-11-05
chore: move to named exports

montezume:
#### Summary

Instead of importing from `@commercetools-frontend/x/something` we only import from `@commercetools-frontend/x` and use named exports.

Also, removed the mix of default and named exports in `application-shell`. [ref](https://github.com/rollup/rollup/issues/1961).

This will help us on our transition towards shipping bundled code.

(This will be a breaking change due to the removing default ApplicationShell export)


packages/application-shell/src/index.js
```diff
@@ -3,7 +3,8 @@
  */
 import './public-path';
 
-export { default } from './components/application-shell';
```
emmenko: Can we keep it the `default`? I consider this to be the **main** component of the package and the rest are additional/complementary stuff that gets exported, so there is like a distinction of which component is more important to use.

But it's not super important if there are good reasons to refactor it...

packages/application-shell/src/index.js
```diff
@@ -3,7 +3,8 @@
...
+export { default as ApplicationShell } from './components/application-shell';
+export { default as apolloClient } from './configure-apollo';
```
emmenko: Was this exposed on purpose?

packages/i18n/async-locale-data/async-locale-data.spec.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { shallow } from 'enzyme';
 import { reportErrorToSentry } from '@commercetools-frontend/sentry';
-import loadI18n from '@commercetools-frontend/i18n/load-i18n';
```
emmenko: Thanks, not sure what happened here 🤦‍♂️

packages/application-shell/src/index.js
montezume: I guess we could keep both for now. But in general, it's not a great idea to expose both a default and named exports from a single module

https://github.com/rollup/rollup/issues/1961

packages/application-shell/src/index.js
montezume: yes, it's used in the discounts app.

packages/application-shell/src/index.js
emmenko: Ah I see, it's mostly for CJS support. Alright, it's fine by me then 💯

tdeekens: I would not reccomend that either. I also do not mind a full and painful refactor across all apps with this maybe even not being backwards compat. It should be a do-able effort and a long standing refactor passed onto teams might be in sum a bigger effort than a at once version.

emmenko:
Maybe we should have waited before merging this. It's a breaking change and I was thinking of releasing this together with the rollup bundle to only have one breaking release.

I guess we won't do any more releases in between...

PR #57, opened 2018-11-06
chore: don't mix named and default exports on code splitting

montezume:
#### Summary

Currently, we have a few places where a new chunk contains mixed exports. This isn't great for CJS builds.

This will help remove some rollup warnings and shouldn't have any effects on consumers, as they would have to be directly importing from these code split components to be effected.


packages/application-shell/src/components/quick-access/quick-access.js
```diff
@@ -23,7 +23,7 @@ const containsMatchByProductKey = data => Boolean(data.productByKey);
 const containsMatchByVariantKey = data => Boolean(data.productByVariantKey);
 const containsMatchByVariantSku = data => Boolean(data.productByVariantSku);
 
-export class QuickAccess extends React.Component {
+class QuickAccess extends React.Component {
```
montezume: wasn't imported anywhere.

PR #58, opened 2018-11-06
chore: refactor i18n for better code splitting

montezume:
#### Summary

We can't use dynamic code splitting in rollup like we were in webpack. This is more explicit. The downside, is we can't define something like "supportedLanguages" in one place. The upside, is we don't need to care about the location of the files, as rollup will split on them.

I'm going to do the same thing in l10n as preparation for the rollup PR as well.

### DIFF
--- a/packages/i18n/README.md
+++ b/packages/i18n/README.md
@@ -5,6 +5,7 @@ Supported languages:
 
 - `en`
 - `de`
+- `es`
 
 ## Install
 
--- a/packages/i18n/load-i18n.js
+++ b/packages/i18n/load-i18n.js
@@ -2,57 +2,36 @@ import { addLocaleData } from 'react-intl';
 import { getMatchingMomentCode } from './utils';
 
 const getReactIntlChunkImport = lang => {
-  let chunkImport;
   switch (lang) {
-    case 'en':
-      chunkImport = import(/* webpackChunkName: "react-intl-data-en" */ 'react-intl/locale-data/en');
-      break;
     case 'de':
-      chunkImport = import(/* webpackChunkName: "react-intl-data-de" */ 'react-intl/locale-data/de');
-      break;
+      return import(/* webpackChunkName: "react-intl-data-de" */ 'react-intl/locale-data/de');
     case 'es':
-      chunkImport = import(/* webpackChunkName: "react-intl-data-es" */ 'react-intl/locale-data/es');
-      break;
+      return import(/* webpackChunkName: "react-intl-data-es" */ 'react-intl/locale-data/es');
     default:
-      chunkImport = import(/* webpackChunkName: "react-intl-data-en" */ 'react-intl/locale-data/en');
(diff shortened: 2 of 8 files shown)


packages/i18n/README.md
```diff
@@ -5,6 +5,7 @@ Supported languages:
 
 - `en`
 - `de`
+- `es`
```
montezume: 😆

packages/i18n/load-i18n.js
```diff
@@ -1,24 +1,73 @@
...
+      chunkImport = import(/* webpackChunkName: "react-intl-localized-strings-de" */ './data/de.json');
+      break;
+    case 'es':
+      chunkImport = import(/* webpackChunkName: "react-intl-localized-strings-es" */ './data/es.json');
+      break;
+    default:
+      chunkImport = import(/* webpackChunkName: "react-intl-localized-strings-en" */ './data/en.json');
+  }
+  return chunkImport;
+};
```
tdeekens: This is totally fine. It‘s one wall of code. Things to change are easily spotted snd should not change every other week after all.

montezume: Yeah... it's not the prettiest. Once all of the language related changes have been made we will need to update the readme docs on how to add new languages. But like you said, it's not happening everyday 🤷‍♂️. Updating a few files is small work compared to translating thousands of strings 😆

packages/i18n/load-i18n.js
```diff
@@ -1,24 +1,73 @@
...
+      chunkImport = import(/* webpackChunkName: "react-intl-localized-strings-en" */ './data/en.json');
+      break;
+    case 'de':
+      chunkImport = import(/* webpackChunkName: "react-intl-localized-strings-de" */ './data/de.json');
+      break;
+    case 'es':
+      chunkImport = import(/* webpackChunkName: "react-intl-localized-strings-es" */ './data/es.json');
+      break;
+    default:
+      chunkImport = import(/* webpackChunkName: "react-intl-localized-strings-en" */ './data/en.json');
```
dferber90: Lazy question: Do we preserve these comments? They are useful for webpack, so there is value in preserving them.

montezume: Yes - here is the output in my current rollup branch

```js

var getReactIntlChunkImport = function getReactIntlChunkImport(lang) {
  var reactIntlChunkImport;

  switch (lang) {
    case 'en':
      reactIntlChunkImport = import(
      /* webpackChunkName: "react-intl-data-en" */
      'react-intl/locale-data/en');
      break;
   ...etc
```

packages/i18n/load-i18n.js
```diff
@@ -1,24 +1,73 @@
 import { addLocaleData } from 'react-intl';
 import { getMatchingMomentCode } from './utils';
 
+const getReactIntlChunkImport = lang => {
+  let chunkImport;
+  switch (lang) {
+    case 'en':
```
emmenko: Nitpick: it being also the default option, we can omit this case

PR #59, opened 2018-11-06
chore: rework code splitting for l10n

montezume:
#### Summary

Same idea as in #58.


packages/l10n/country-information.js
```diff
@@ -3,6 +3,25 @@ import createL10NInjector from './create-l10n-injector';
 import getSupportedLanguage from './utils/get-supported-language';
 import extractLanguageFromLocale from './utils/extract-language-from-locale';
 
+const getImportChunk = lang => {
+  let importChunk;
+  switch (lang) {
+    case 'en':
+      importChunk = import(/* webpackChunkName: "country-data-en" */ './data/countries/en.json');
```
emmenko: Same nitpick as the other PR 🤪

PS: you can also `return import(...)` directly instead of assigning a local var

PR #60, opened 2018-11-06
Add ApplicationState to test-utils

dferber90:
This enhances the capabilities of our `test-utils`.

The application-kit exports `render` (and a bunch of other `react-testing-library` functions) from `@commercetools-frontend/application-shell/src/test-utils`. The `test-utils` file contains a `render` function which can be used to render any component as-if it was rendered by `AppShell`. The `render` function sets up the context providers for the component-under-test. It accepts some options which influence what that context looks like.

The context includes:
- `IntlProvider`
- `ApolloProvider`
- `ConfigureFlopFlip`
- `ApplicationStateProvider`
- `Router`
- permissions

Tests in applications can now test their components using this `render` function like so:

```jsx
// The component under test

// A dumb component which prints the users first and last name
const Username = () => (
  <GetApplicationState
    render={applicationState =>
      [
        applicationState.user.firstName,
        applicationState.user.lastName,
      ].join(' ')
    }
  />
);
export default Username
```

```jsx
// a test
import { render } from '@commercetools-frontend/application-shell/src/test-utils'
import Username from './username'

describe('Username', () => {
  it('should print the user name', () => {
    const { container } = render(<Username />);
    expect(container).toHaveTextContent('Sheldon Cooper');
  });

  // it is also possible to overwrite the default values
  it('should be possible to overwrite', () => {
    const { container } = render(<Username />, {
      user: { firstName: 'Leonard', lastName: 'Hofstadter' }
    });
    expect(container).toHaveTextContent('Leonard Hofstadter');
  });

  // it is even possible to access the defaults because the "render" method returns them
  it('should be possible to overwrite', () => {
    const { container, user } = render(<Username />, {
      user: { firstName: 'Leonard', lastName: 'Hofstadter' }
    });
    expect(container).toHaveTextContent(user.firstName + ' ' + user.lastName);
  });
});
```

This also works for feature flags, permissions, projects and so on. See the `test-utils.spec.js` file of this PR to find out which providers `test-utils` influences.

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -1,5 +1,6 @@
----
-extends: "@commercetools-frontend/eslint-config-mc-app"
+extends: '@commercetools-frontend/eslint-config-mc-app'
 settings:
   react:
     version: '^16.0'
+  import/ignore:
+    - test-utils.js$
--- a/packages/application-shell/src/test-utils.js
+++ b/packages/application-shell/src/test-utils.js
@@ -31,7 +31,7 @@ afterEach(memoryAdapter.reset);
 // Inspired by
 // https://github.com/kentcdodds/react-testing-library-course/blob/2a5b1560656790bb1d9c055fba3845780b2c2c97/src/__tests__/react-router-03.js
 // eslint-disable-next-line import/prefer-default-export
-export const render = (
+const render = (
   ui,
   {
     // react-intl
@@ -68,7 +68,7 @@ export const render = (
 // Test setup for rendering with Redux
 // We expose a sophisticated function because we plan to get rid of Redux
 // Use this function only when your test actually needs Redux
-export const renderWithRedux = (
+const renderWithRedux = (
   ui,
   {
     // Consuemrs of renderWithRedux can use
@@ -84,3 +84,12 @@ export const renderWithRedux = (
   // this to test implementation details).
   store,
 });
+
+// re-export everything
+export * from 'react-testing-library';
+
+export {
+  // override render method of react-testing-library
+  render,
+  renderWithRedux,
+};
--- a/packages/application-shell/src/test-utils.js
+++ b/packages/application-shell/src/test-utils.js
@@ -12,6 +12,7 @@ import { ConfigureFlopFlip } from '@flopflip/react-broadcast';
 import { MockedProvider as ApolloMockProvider } from 'react-apollo/test-utils';
 import memoryAdapter from '@flopflip/memory-adapter';
 import { Provider as StoreProvider } from 'react-redux';
+import { ApplicationStateProvider } from '@commercetools-frontend/application-shell-connectors';
 import { createReduxStore } from './configure-store';
 
 // Reset memoryAdapter after each test, so that the next test accepts the
@@ -20,17 +21,52 @@ import { createReduxStore } from './configure-store';
 // location is better for it.
 afterEach(memoryAdapter.reset);
 
-// This function renders any component within the application context.
+// These default values get merged with the values provided by the test from
+// the call to "render"
+const defaultUser = {
+  id: 'user-id-1',
+  email: 'sheldon.cooper@caltech.edu',
+  firstName: 'Sheldon',
+  lastName: 'Cooper',
+  language: 'en',
+  timeZone: 'Etc/UTC',
+};
+
+const defaultEnvironment = {
+  frontendHost: 'localhost:3001',
+  mcApiUrl: 'https://mc-api.commercetools.com',
+  location: 'eu',
+  env: 'production',
+  cdnUrl: 'http://localhost:3001',
+  servedByProxy: false,
+};
+
+const defaultProject = {
+  key: 'test-with-big-data',
+  version: 43,
+  name: 'Test with big data',
+  countries: ['de', 'en'],
+  currencies: ['EUR', 'GBP'],
+  languages: ['de', 'en-GB'],
+  permissions: { canManageProject: true },
+};
+
+// Allow consumers of `render` to extend the defaults by passing an object
+// or to completely omit the value by passing `null`
+const mergeOptional = (defaultValue, value) =>
+  value === null ? undefined : { ...defaultValue, ...value };
+
+// This function renders any component within the application context, as if it
+// was rendered inside <ApplicationShell />.
 // The context is not completely set up yet, some things are missing:
 //   - Tracking on context
-//   - Project information
+//   - react-intl's information from addLocaleData
 //   - possibly more that I'm not aware of right now
 //
 //  We can add these things as we go and when we need them.
 
 // Inspired by
 // https://github.com/kentcdodds/react-testing-library-course/blob/2a5b1560656790bb1d9c055fba3845780b2c2c97/src/__tests__/react-router-03.js
-// eslint-disable-next-line import/prefer-default-export
 const render = (
   ui,
   {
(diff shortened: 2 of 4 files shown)


.eslintrc.yaml
```diff
@@ -1,5 +1,6 @@
----
-extends: "@commercetools-frontend/eslint-config-mc-app"
+extends: '@commercetools-frontend/eslint-config-mc-app'
 settings:
   react:
     version: '^16.0'
+  import/ignore:
+    - test-utils.js$
```
dferber90: We do `export * from 'react-testing-library'` in `test-utils.js`, which confuses the `import/named` rule. It complains when doing `import { fireEvent } from '<xy>/test-utils'` as `fireEvent` is one of the forwarded exports.

To avoid a warning in every consumer, we simply ignore `test-utils`. This is something all consumers of `application-kit` (more specifically, `@commercetools-frontend/application-shell/src/test-utils`) which have `import/named` enabled need to add to their ESLint setup as well.

packages/application-shell/src/test-utils.spec.js
```diff
@@ -0,0 +1,258 @@
...
+// This is a bit different from our usual tests, as we are testing our testing
+// tools here instead of actual components.
+import React from 'react';
+import { injectIntl } from 'react-intl';
+import { Query } from 'react-apollo';
+import gql from 'graphql-tag';
+import { injectFeatureToggle } from '@flopflip/react-broadcast';
+import { Switch, Route } from 'react-router';
+import { render, wait } from './test-utils';
+import { GetApplicationState } from '../../application-shell-connectors';
```
dferber90: This file shows all the different things `render` can influence when testing any component. We now have the majority of things covered 🎈

packages/application-shell/src/test-utils.spec.js
```diff
@@ -0,0 +1,277 @@
+// This file tests our test-utils file #inception
```
dferber90: This file shows how consumers of  `application-kit` could use `render` from `test-utils` to test different aspects of a feature.

packages/application-shell/src/test-utils.js
```diff
@@ -45,30 +81,53 @@ export const render = (
...
+          </ConfigureFlopFlip>
+        </ApolloMockProvider>
+      </IntlProvider>,
+      renderOptions
+    ),
+    // adding `history` to the returned utilities to allow us
+    // to reference it in our tests (just try to avoid using
+    // this to test implementation details).
+    history,
+    // Adding user, project & environment so tests know about the merge results
```
emmenko: 💯

.eslintrc.yaml
dferber90: Then every single test which imports anything other than `render` or `renderWithRedux` from `test-utils` would have to use `eslint-ignore`. 

Aside from that, which upside do you see in using `eslint-ignore` over disabling the rule only for that import specifically?

packages/application-shell/src/test-utils.js
```diff
@@ -20,18 +21,53 @@ import { createReduxStore } from './configure-store';
...
+  lastName: 'Cooper',
+  language: 'en',
+  timeZone: 'Etc/UTC',
+};
+
+const defaultEnvironment = {
+  frontendHost: 'localhost:3001',
+  mcApiUrl: 'https://mc-api.commercetools.com',
+  location: 'eu',
+  env: 'production',
```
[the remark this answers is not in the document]
dferber90: The default values don't matter if I'm not mistaken. All requests to the outside are stubbed - and the tokens used to make the requests aren't valid either. Is there a concern I'm missing?

I prefer having the test-setup close to the production setup so that the tests are closer to real conditions. I would even like to use a real value for `frontendHost` but I didn't know what that would look like. `mc.commercetools.com`?

packages/application-shell/src/test-utils.js
```diff
@@ -20,18 +21,53 @@ import { createReduxStore } from './configure-store';
...
+  cdnUrl: 'http://localhost:3001',
+  servedByProxy: false,
+};
+
+const defaultProject = {
+  key: 'test-with-big-data',
```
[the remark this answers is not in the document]
dferber90: The key of our test-project is `test-with-big-data-44` so this is already a test key.

Which other key would you prefer?

packages/application-shell/src/test-utils.js
```diff
@@ -20,18 +21,53 @@ import { createReduxStore } from './configure-store';
...
+  version: 43,
```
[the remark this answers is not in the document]
dferber90: This `version` is the one which updates after every change to the project settings (like languages, currencies, countries), right? I don't mind the number we use but I'm also to lazy too change it now. It doesn't matter to me.

dferber90:
We want to wait for #40 before merging this PR.

.eslintrc.yaml
tdeekens: That's a valid point. I also like the honesty of having it centralised in an ESLint config. We have spread ignoring to export a default export throughout the app locally. Maybe we should also add that here given we don't seem to mind.

packages/application-shell/src/test-utils.js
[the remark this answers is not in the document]
tdeekens: Silly argument for using production: we want to test how the user perceives it as similar as possible. He will be on a production environment. Even if this might not change anything :).

packages/application-shell/src/test-utils.js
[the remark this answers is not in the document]
tdeekens: I don't mind. Our E2E project is `e2e-testaccount-39` maybe this could be `integration-testaccount-39`.

packages/application-shell/src/test-utils.js
```diff
@@ -45,30 +81,53 @@ export const render = (
...
-  // to reference it in our tests (just try to avoid using
-  // this to test implementation details).
-  history,
-});
+) => {
+  const mergedUser = mergeOptional(defaultUser, user);
+  const mergedProject = mergeOptional(defaultProject, project);
+  const mergedEnvironment = mergeOptional(defaultEnvironment, environment);
+  return {
+    ...rtlRender(
```
tdeekens: Never seen this before 🌠.

packages/application-shell/src/test-utils.spec.js
```diff
@@ -0,0 +1,277 @@
...
+describe('ApplicationState', () => {
+  describe('user', () => {
+    const TestComponent = () => (
+      <GetApplicationState
+        render={applicationState =>
+          [
+            applicationState.user.firstName,
+            applicationState.user.lastName,
+          ].join(' ')
+        }
```
tdeekens: I was way to slow grasping this but it's nice.

.eslintrc.yaml
dferber90: Yeah, we can do another PR to improve our ESLint rules. I'd want to disable [arrow-body-style](https://eslint.org/docs/rules/arrow-body-style) and [prefer-default-export](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/prefer-default-export.md). They both don't add much value and I find myself often changing back-and-forth between `return` and no return when debugging.

packages/application-shell/src/test-utils.js
dferber90: This is just `render` from `react-testing-library`, which I've renamed in the import. The `rtl` is short for `react-testing-library`. I found this pattern on the `react-testling-library` README and it's awesome.

dferber90:
Merged now as I'm pairing with @montezume on the rollup bundling and we want this one in for it.

PR #62, opened 2018-11-07
fix: circular dependency fix

montezume:
2e5a77bc32a7eaae461f16e900e91d24b9fb25fce5d1dfec57f422e5f3f0b842

#### Summary

Currently, in our `header-link`, we are using apolloClient to read query to get the user id. This causes a circular dependency that looks something like

```(!) Circular dependency: src/configure-apollo.js -> src/apollo-links/index.js -> src/apollo-links/header-link.js -> src/utils/index.js -> src/utils/select-user-id/index.js -> src/utils/select-user-id/select-user-id.js -> src/configure-apollo.js```

#### Approach

Instead of importing `apolloClient` in `select-user-id`, we pass in the cache, which we can get inside of the headerLink using `getContext()`. Then we execute getQuery on the cache. Also added two missing deps to `application-shell/package.json`

### DIFF
--- a/packages/application-shell/package.json
+++ b/packages/application-shell/package.json
@@ -25,6 +25,7 @@
     "npm": ">=5"
   },
   "dependencies": {
+    "@babel/runtime": "7.1.2",
     "@commercetools-frontend/actions-global": "1.0.0-rc.3",
     "@commercetools-frontend/application-shell-connectors": "1.0.0-rc.3",
     "@commercetools-frontend/browser-history": "1.0.0-rc.3",
@@ -86,7 +87,8 @@
     "redux-thunk": "2.3.0",
     "reselect": "3.0.1",
     "unfetch": "3.1.1",
-    "uuid": "3.3.2"
+    "uuid": "3.3.2",
+    "warning": "4.0.2"
   },
   "devDependencies": {
     "@commercetools-frontend/ui-kit": "2.0.0-rc.11",
--- a/packages/application-shell/src/utils/get-correlation-id/get-correlation-id.js
+++ b/packages/application-shell/src/utils/get-correlation-id/get-correlation-id.js
@@ -1,8 +1,8 @@
 import uuid from 'uuid/v4';
 import selectProjectKeyFromUrl from '../select-project-key-from-url';
-import selectUserId from '../select-user-id';
 
-export default cache =>
-  ['mc', selectProjectKeyFromUrl(), selectUserId(cache), uuid()]
+export default function getCorrelationId({ userId } = {}) {
+  return ['mc', selectProjectKeyFromUrl(), userId, uuid()]
     .filter(Boolean)
     .join('/');
+}
--- a/packages/application-shell/src/utils/get-correlation-id/get-correlation-id.js
+++ b/packages/application-shell/src/utils/get-correlation-id/get-correlation-id.js
@@ -2,7 +2,7 @@ import uuid from 'uuid/v4';
 import selectProjectKeyFromUrl from '../select-project-key-from-url';
 import selectUserId from '../select-user-id';
 
-export default () =>
-  ['mc', selectProjectKeyFromUrl(), selectUserId(), uuid()]
+export default cache =>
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

File: packages/application-shell/package.json [hunk 1]
```diff
@@ -86,7 +87,8 @@
     "redux-thunk": "2.3.0",
     "reselect": "3.0.1",
     "unfetch": "3.1.1",
-    "uuid": "3.3.2"
+    "uuid": "3.3.2",
+    "warning": "4.0.2"
```
montezume: this is used [here](https://github.com/commercetools/merchant-center-application-kit/blob/4bdb1197f26d03f3ce77ed33d5d7b47db93e4d41/packages/application-shell/src/from-core/deprecate-component/deprecate-component.js)

packages/application-shell/package.json
```diff
@@ -25,6 +25,7 @@
     "npm": ">=5"
   },
   "dependencies": {
+    "@babel/runtime": "7.1.2",
```
montezume: this is used [here](https://github.com/commercetools/merchant-center-application-kit/blob/7719428fefffa2fbf24128aec7ae8212e96d06f2/packages/application-shell/src/components/quick-access/quick-access.js#L119)

packages/application-shell/src/utils/select-user-id/select-user-id.js
```diff
@@ -1,9 +1,8 @@
-import apolloClient from '../../configure-apollo';
 import UserIdQuery from './select-user-id.graphql';
 
-export default () => {
+export default cache => {
   try {
-    const queryResult = apolloClient.readQuery({
+    const queryResult = cache.readQuery({
```
montezume: this is the fix

packages/application-shell/package.json
emmenko: What happens without it?

montezume: you get a warning with the rollup build saying unresolved dep

montezume: the compiled code looks like

```
...
      var _subCommands = _asyncToGenerator(
      /*#__PURE__*/
      _regeneratorRuntime.mark(function _callee() {

```

emmenko: Hmm ok 🤷‍♂️

packages/application-shell/src/utils/get-correlation-id/get-correlation-id.js
```diff
@@ -2,7 +2,7 @@ import uuid from 'uuid/v4';
 import selectProjectKeyFromUrl from '../select-project-key-from-url';
 import selectUserId from '../select-user-id';
 
-export default () =>
-  ['mc', selectProjectKeyFromUrl(), selectUserId(), uuid()]
+export default cache =>
```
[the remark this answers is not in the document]
tdeekens: I would dare to counter argue: a function should receive only what it needs. This function only needs the cache and not the whole context?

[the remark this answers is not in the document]
emmenko: I'm not saying to pass the context, only the `cache` but as a named argument. It's easier then to pass more options if needed.

Not super important though...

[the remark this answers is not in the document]
emmenko: Oh right, this is also used there...🤔

montezume:
I'm not sure how to proceed here, since createReduxStore needs access to `getCorrelationId` as well, and in our entry points, we have the <StoreProvider> wrapping the <ApplicationShell>.

emmenko:
Can we try with the approach I suggested?

image

montezume:
I'm not really familiar with the SDK. I'm not sure what you mean by when using the SDK

You mean not using the middleware at all? Somehow passing userId on every sdk request?

emmenko:
Yeah, let me push a commit later

montezume:
I'll close this and move the dep updates into #40

emmenko:
?? The PR was branched off this one. You could have just merged that here 🙃

montezume:
Okay, I can do that still. 🤔

PR #71, opened 2018-11-08
QuickAccess: allow opening in new tab by cmd+click or cmd+enter

dferber90:
This adds a feature requested by @antonboyko to QuickAccess which allows opening the result in a new tab by pressing either <kbd>cmd+click</kbd> or by using <kbd>cmd+enter</kbd> when on the selected result. On windows the combination is <kbd>ctrl+click</kbd> or <kbd>ctrl+enter</kbd>.

This also gets rid of the behaviour where Privacy and Help were opened in a new tab/window by default.

### DIFF
--- a/packages/application-shell/src/components/quick-access/butler/butler.js
+++ b/packages/application-shell/src/components/quick-access/butler/butler.js
@@ -25,6 +25,25 @@ const isCloseCombo = event =>
   !event.altKey &&
   !event.shiftKey;
 
+const getPlatform = () => {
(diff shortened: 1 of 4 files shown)


emmenko:
Remember to update the changelog also ;)

dferber90:
Isn't the changelog for developers? It feels a bit weird to mention user-facing features of the app-shell in the changelog. Nothing changed for consumers of app-kit.

montezume:
Yeah, I would only put things in the changelog that could change consumer code.

PR #73, opened 2018-11-08
feat(app-context): add forward-compatible useApplicationContext 🎉

emmenko:
I couldn't resist 😜 (https://github.com/commercetools/merchant-center-application-kit/pull/72#discussion_r231988805)

### DIFF
--- a/packages/application-shell-connectors/src/components/application-context/application-context.js
+++ b/packages/application-shell-connectors/src/components/application-context/application-context.js
@@ -5,7 +5,7 @@ import { defaultMemoize } from 'reselect';
 import moment from 'moment-timezone';
 import omit from 'lodash.omit';
 
-const { Provider, Consumer } = React.createContext({});
+const Context = React.createContext({});
 
 const defaultTimeZone = moment.tz.guess() || 'Etc/UTC';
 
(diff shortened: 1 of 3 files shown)


packages/application-shell-connectors/src/components/application-context/application-context.js
```diff
@@ -134,9 +136,22 @@ const withApplicationContext = mapApplicationContextToProps => Component => {
   return WrappedComponent;
 };
 
+// eslint-disable-next-line import/no-mutable-exports
+let useApplicationContext;
```
tdeekens: Instead of exporting undefined we could export a function throwing given there is no useContext with e.g: This is meant to be a forward compatible React hook to consume the application context. React hooks, more specifically useContext, do not seem to be available yet on the installed version of React.“.

dferber90: We can also get around `// eslint-disable-next-line import/no-mutable-exports` by doing

```js
const useApplicationContext = React.useContext
  ? mapApplicationContextToProps => {}
  : () => throw new Error('React hooks do not seem to be available yet on the installed version of React.') 
```

PR #74, opened 2018-11-09
Update dependencies

tdeekens:
#### Summary

This pull request updates the dependencies which I suspect not cause much trouble.

#### Description

Given the a new release around the corner I felt updating the dependencies with it might be good. This does not include: any `postcss` dependencies (I recommend doing them separately) and `recompose` which also caused us trouble in the `0.30.0` version.


packages/application-shell/package.json
```diff
@@ -63,7 +63,7 @@
     "classnames": "2.2.6",
     "common-tags": "1.8.0",
     "debounce-async": "0.0.2",
-    "downshift": "2.2.3",
+    "downshift": "3.1.5",
```
tdeekens: Double checked. Nothing we use here affects this.

packages/url-utils/package.json
```diff
@@ -18,8 +18,8 @@
   "publishConfig": {
     "access": "public"
   },
-  "main": "dist/url-utils.js",
-  "module": "dist/url-utils.js",
+  "main": "./dist/url-utils.cjs.js",
```
montezume: thanks 👍

tdeekens:
Yay.

PR #77, opened 2018-11-09
refactor(assets): remove named exports

montezume:
After some soul searching and discussion with @dferber90 and @tdeekens, providing base64 encoded SVGs  doesn't really provide much value for our users, and complicates testing.

It makes more sense to leave it up to the consumer of the bundle how to deal with SVGs and other images. This way, they can decide whether or not to base64, or to just copy the images.


montezume:
@dferber90 Okay I'll give that a try, thanks for the tip 😄

montezume:
@dferber90 can you re-review now?

montezume:
@dferber90 I did put this in the README.md though

PR #78, opened 2018-11-09
refactor(navbar): remnove `CAN_VIEW_PRODUCTS` feature toggle

adnasa:
#### Summary

Let's clean up these feature toggles.
This was introduced in conjunction with issues that we faced with multi-cloud. I'd like to release this upon merge. any objections?


adnasa:
Will have this released by lunch.

adnasa:
aaah...
maybe I shouldn't jump the gun on releasing due to breaking changes introduced since last

PR #87, opened 2018-11-12
chore: update deps

montezume:
From my look this morning, all of the updates seem small.
* Closes #81 
* Closes #82 
* Closes #83 
* Closes #84 
* Closes #85 
* Closes #86


packages/mc-scripts/package.json
```diff
@@ -42,7 +42,7 @@
     "graphql-tag": "2.10.0",
     "html-webpack-plugin": "3.2.0",
     "json-loader": "0.5.7",
-    "mini-css-extract-plugin": "0.4.0",
+    "mini-css-extract-plugin": "0.4.4",
```
tdeekens: I am not 100% sure but didn't we have a problem with this one?

montezume: I checked the release log and it seems pretty inconsequential. I don't think it will cause us any problems 🤷‍♂️

dferber90: I remember there were some problems around the order of CSS files, which lead to specificity issues in the MC. I'm not sure whether this was the source of it or not. I might be confusing it.

montezume: @dferber90 I mistakenly blamed problems with our "external" css on order of css files, but the actual problem was we were using a CSS minifier that was trying to normalize z-index. My assumption was wrong about it being due to the order / precedence. @tdeekens

tdeekens: Oki. Thanks.

emmenko: See here: https://github.com/commercetools/merchant-center-frontend/blob/ecea0566aad24b436e69987bdce499c385e3816d/scripts/throw-if-pinned-dependency-version-changes.js#L23-L26

emmenko: We should probably create that check script also in this repo.

However, maybe the new versions fixed this problem. So maybe we can check it again.

PR #89, opened 2018-11-12
refactor: use React.lazy and React.Suspense within the AppShell

emmenko:
To get us ready for **_the future_** 🤫

In this PR I replaced `react-loadable` with [`React.lazy`](https://reactjs.org/docs/code-splitting.html#reactlazy) and `<React.Suspense>` (for handling the loading state). Errors are now caught more specifically using Error Boundaries.

With this change, we can simply use `React.lazy` to load the split points in the applications.

PS: we had to upgrade the React Router to `4.0.0-beta` to [avoid prop types warnings](https://github.com/ReactTraining/react-router/blob/5ccb241cd59eae7a15de31c7babff6619378be39/packages/react-router/modules/Route.js#L131-L137) because of the `React.lazy` returning an object instead of a function (see AppShell snapshot diff).


playground/src/components/entry-point/entry-point.js
```diff
@@ -1,12 +1,10 @@
 import React from 'react';
 import { Provider as StoreProvider } from 'react-redux';
-import Loadable from 'react-loadable';
 import {
   ApplicationShell,
   reduxStore,
   setupGlobalErrorListener,
   RouteCatchAll,
-  AsyncChunkLoader,
```
emmenko: I haven't removed this component yet because is still used by the applications. We might deprecate is at some point and then remove it.

packages/application-shell/src/components/quick-access/index.js
```diff
@@ -82,7 +109,9 @@ export default class QuickAccessTrigger extends React.Component {
 
   render() {
     return this.state.isVisible ? (
-      <QuickAccessModal {...this.props} onClose={this.close} />
+      <QuickAccessContainer>
+        <QuickAccessModal {...this.props} onClose={this.close} />
+      </QuickAccessContainer>
```
emmenko: @dferber90 please check this part if it's still ok

packages/application-shell/src/components/quick-access/butler-container/butler-container.js
```diff
@@ -1,41 +1,14 @@
...
-          'timedOut',
-          'pastDelay',
-          'isLoading',
-          'retry',
-        ])}
-      />
-    );
-  }
-}
+const ButlerContainer = React.memo(props => (
```
emmenko: ...why not 😁

packages/application-shell/src/components/project-container/project-container.js
```diff
@@ -108,42 +116,38 @@ export class ProjectContainer extends React.Component {
...
-              <ProjectSuspended
-                isTemporary={
-                  project.suspension.reason ===
-                  SUSPENSION_REASONS.TEMPORARY_MAINTENANCE
-                }
-              />
-            );
-          if (project.expiry && project.expiry.isActive)
-            return <ProjectExpired />;
+      <React.Suspense fallback={<ApplicationLoader />}>
```
emmenko: This diff is because of the indentation for adding `<React.Suspense>` to the tree

packages/application-shell/src/components/project-container/project-container.js
```diff
@@ -108,42 +116,38 @@ export class ProjectContainer extends React.Component {
...
-                  user={this.props.user}
-                  project={project}
-                  projectDataLocale={locale}
-                  environment={this.props.environment}
-                >
-                  <React.Fragment>
-                    {!isNil(project.expiry.daysLeft) &&
-                      project.expiry.daysLeft <= minDaysToDisplayNotification &&
-                      project.expiry.daysLeft >=
-                        maxDaysToDisplayNotification && (
```
emmenko: I moved this logic to a function, it makes it a bit easier to read

packages/application-shell/src/components/quick-access/butler-container/butler-container.js
dferber90: Not too sure, but wouldn't this prevent updates within `Butler` like `shouldComponentUpdate` does?

emmenko: Tho component is only used for the loading state

packages/application-shell/package.json
```diff
@@ -117,8 +116,8 @@
     "react-apollo": ">=2 <3",
     "react-dom": ">=16 <17",
     "react-intl": ">=2 <3",
-    "react-router": ">=4 <5",
-    "react-router-dom": ">=4 <5"
+    "react-router": ">=4 <5 || >=4.4.0-beta.0",
+    "react-router-dom": ">=4 <5 || >=4.4.0-beta.0"
```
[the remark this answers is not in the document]
tdeekens: I actually never tired and dont know if the -beta.0 things are part of „official“ semver and hence of the underlying matchers in yarn and npm. Is somebody else sure or tried? I am asking to make sure that saying > beta.0 actually permits beta.6 being installed.

tdeekens:
For instance https://github.com/smooth-code/loadable-components builds on lazy nicely. We may not need the options now but we could claim that we want to replace react-loadable with that across all apps. Just wondering what your thoughts are.

packages/application-shell/package.json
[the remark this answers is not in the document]
emmenko: Yes, we are doing the same for the uikit.
I can put an upper bound limit, no problem.

emmenko:
I would be careful to use library that do a lot of abstraction. Also, you would still need to use `<React.Suspense>` yourself right? So at the end you're just replacing `React.lazy`.

I would stick to the React APIs for now, we know what we get from them.

packages/application-shell/src/components/quick-access/butler-container/butler-container.js
emmenko: Oh shit, I see now it's also used in the `<Butler>`. Sorry my mistake, will remove the memoization

emmenko: Alright, I ended up removing the intermediate `<ButlerContainer>` as it wasn't really necessary anymore.

https://github.com/commercetools/merchant-center-application-kit/pull/89/commits/c5e70b4a863360b4d13c2598d5e7f31d75143bed

emmenko:
Alright, anything else holding this off? Can I get some approvals?

PR #93, opened 2018-11-13
fix(application-shell): expose test-utils

dferber90:
We were exposing `test-utils` from the `dist` folder by accident which is inconvenient for consumers, and not what we explain in our docs.

After this PR importing `test-utils` will work as documented:
```diff
- import { render } from '@commercetools-frontend/application-shell/dist/test-utils';
+ import { render } from '@commercetools-frontend/application-shell/test-utils'; 
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,6 +67,7 @@ typings/
 # Build
 dist
 public
+packages/application-shell/test-utils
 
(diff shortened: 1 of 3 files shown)


montezume:
Looks good. Do you want to update the changelog in this PR though?

PR #96, opened 2018-11-13
fix: cjs for test-utils instead of es

montezume:
Build cjs version of test utils in app-shell instead of es version.


montezume:
@dferber90 One step at a time 😄

PR #98, opened 2018-11-13
refactor(login-sso): avoid query params in authorize url, use session storage and state nonce

emmenko:
Long story short: redirecting to the IdP `/authorize` endpoint with the `redirect_url` with some query parameters does not guarantee that those params are sent back.
Since we rely on the `organizationId` to authorize the user, we need to be able to "get it" back when the IdP redirects to our `/login/sso/callback` route.

Following [the tips in here](https://auth0.com/docs/protocols/oauth2/oauth-state#how-to-use-the-parameter-to-restore-application-state) I refactored the code a bit to store the `organizationId` as a "state" object within the `nonce` in the ~~`localStorage`~~ `sessionStorage`.

This will result in the `redirect_url` to be only e.g. `https://mc.commercetools.com/login/sso/callback` and the saved nonce will contain the needed state:

```js
// sessionStorage
"nonce_123": {
  "organizationId": "org-123"
}
```

---

Unrelated thought: we should probably consider moving the login pages outside of the AppShell and into e.g. the "fallback" application. It will allow to do changes to those pages without having to touch the AppShell.

### DIFF
--- a/packages/application-shell/src/components/login-sso-callback/login-sso-callback.js
+++ b/packages/application-shell/src/components/login-sso-callback/login-sso-callback.js
@@ -14,9 +14,6 @@ export class LoginSSOCallback extends React.PureComponent {
   static displayName = 'LoginSSOCallback';
   static propTypes = {
     location: PropTypes.shape({
-      query: PropTypes.shape({
-        organizationId: PropTypes.string.isRequired,
-      }).isRequired,
       hash: PropTypes.string.isRequired,
     }).isRequired,
     redirectTo: PropTypes.func.isRequired,
@@ -31,16 +28,20 @@ export class LoginSSOCallback extends React.PureComponent {
     const fragments = qs.parse(this.props.location.hash.substring(1));
     const idToken = fragments.id_token;
     const decodedIdToken = jwtDecode(idToken);
-    const nonce = storage.get(STORAGE_KEYS.NONCE);
-    storage.remove(STORAGE_KEYS.NONCE);
+    const sessionState = storage.get(
+      `${STORAGE_KEYS.NONCE}_${decodedIdToken.nonce}`,
+      { storage: window.sessionStorage }
+    );
+    // Clear the nonce, we don't need it anymore
+    storage.remove(STORAGE_KEYS.NONCE, { storage: window.sessionStorage });
 
-    if (decodedIdToken.nonce !== nonce) {
+    if (!sessionState) {
       this.setAuthenticationFailed(true);
     } else {
       this.props
         .requestAccessToken({
           idToken,
-          organization: this.props.location.query.organizationId,
+          organization: sessionState.organizationId,
         })
         .then(payload => {
           // Set a flag to indicate that the user has been authenticated
--- a/packages/application-shell/src/components/login-sso-callback/login-sso-callback.js
+++ b/packages/application-shell/src/components/login-sso-callback/login-sso-callback.js
@@ -28,12 +28,10 @@ export class LoginSSOCallback extends React.PureComponent {
     const fragments = qs.parse(this.props.location.hash.substring(1));
     const idToken = fragments.id_token;
     const decodedIdToken = jwtDecode(idToken);
-    const sessionState = storage.get(
-      `${STORAGE_KEYS.NONCE}_${decodedIdToken.nonce}`,
-      { storage: window.sessionStorage }
-    );
+    const nonceKey = `${STORAGE_KEYS.NONCE}_${decodedIdToken.nonce}`;
+    const sessionState = window.sessionStorage.getItem(nonceKey);
     // Clear the nonce, we don't need it anymore
-    storage.remove(STORAGE_KEYS.NONCE, { storage: window.sessionStorage });
+    window.sessionStorage.removeItem(nonceKey);
 
     if (!sessionState) {
       this.setAuthenticationFailed(true);
--- a/packages/application-shell/src/components/login-sso/login-sso.js
+++ b/packages/application-shell/src/components/login-sso/login-sso.js
@@ -12,7 +12,6 @@ import {
 } from '@commercetools-frontend/url-utils';
 import { actions as sdkActions } from '@commercetools-frontend/sdk';
 import { connect } from 'react-redux';
-import * as storage from '@commercetools-frontend/storage';
 import { Notification } from '@commercetools-frontend/react-notifications';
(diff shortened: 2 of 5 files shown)


packages/application-shell/src/components/login-sso-callback/login-sso-callback.js
```diff
@@ -31,16 +28,20 @@ export class LoginSSOCallback extends React.PureComponent {
     const fragments = qs.parse(this.props.location.hash.substring(1));
     const idToken = fragments.id_token;
     const decodedIdToken = jwtDecode(idToken);
-    const nonce = storage.get(STORAGE_KEYS.NONCE);
-    storage.remove(STORAGE_KEYS.NONCE);
+    const sessionState = storage.get(
+      `${STORAGE_KEYS.NONCE}_${decodedIdToken.nonce}`,
+      { storage: window.sessionStorage }
```
emmenko: I'm not 100% happy with this...I'm not sure having a `storage` package is also really necessary.

packages/application-shell/src/components/login-sso-callback/login-sso-callback.js
```diff
@@ -31,16 +28,20 @@ export class LoginSSOCallback extends React.PureComponent {
...
 
-    if (decodedIdToken.nonce !== nonce) {
+    if (!sessionState) {
       this.setAuthenticationFailed(true);
     } else {
       this.props
         .requestAccessToken({
           idToken,
-          organization: this.props.location.query.organizationId,
+          organization: sessionState.organizationId,
```
emmenko: This is basically the goal of the fix/refactoring ℹ️

packages/application-shell/src/components/login-sso-callback/login-sso-callback.js
```diff
@@ -31,16 +28,20 @@ export class LoginSSOCallback extends React.PureComponent {
     const fragments = qs.parse(this.props.location.hash.substring(1));
     const idToken = fragments.id_token;
     const decodedIdToken = jwtDecode(idToken);
-    const nonce = storage.get(STORAGE_KEYS.NONCE);
-    storage.remove(STORAGE_KEYS.NONCE);
+    const sessionState = storage.get(
+      `${STORAGE_KEYS.NONCE}_${decodedIdToken.nonce}`,
```
emmenko: This implicitly validates that the `nonce` is valid ℹ️

packages/application-shell/src/components/login-sso/login-sso.js
```diff
@@ -67,26 +75,26 @@ export class LoginSSO extends React.PureComponent {
...
-          });
           const params = querystring.stringify({
             scope: 'openid email profile',
             response_type: 'id_token',
             client_id: authProvider.clientId,
             redirect_uri: trimLeadingAndTrailingSlashes(
-              joinPaths(
-                this.props.originUrl,
-                this.props.match.url,
-                `callback?${redirectUriParams}`
```
emmenko: And this was what originated the problem

packages/application-shell/src/components/login-sso-callback/login-sso-callback.js
```diff
@@ -31,16 +28,20 @@ export class LoginSSOCallback extends React.PureComponent {
...
+      { storage: window.sessionStorage }
+    );
+    // Clear the nonce, we don't need it anymore
+    storage.remove(STORAGE_KEYS.NONCE, { storage: window.sessionStorage });
```
tdeekens: I always felt wrapping a native API only makes sense if:

1. Eases testing eg mocks to be used
2. It is needed and easy to swap the storage wihout having to change the call-site

Neither holds for us. Jest has great mocking plugins for local storage etc. Plus we do not really want to or need to swap the engine (as in strategy pattern) here. 

Anyway I am fine with keeping thigs obviously. However we can ask how much we need the package.

packages/application-shell/src/components/login-sso/login-sso.js
```diff
@@ -32,9 +32,17 @@ export const getMessageKeyForError = error => {
...
+function generateAndCacheNonceWithState(state) {
   const nonce = uuid();
-  storage.put(STORAGE_KEYS.NONCE, nonce);
+  // We store additional information within the given `nonce`
+  // to then retrieve it later when the IdP redirects back
+  // to our application. The URL will contain the `nonce` within
+  // the id_token and, once validated, we can retrieve and use
+  // the state object.
+  // https://auth0.com/docs/protocols/oauth2/oauth-state#how-to-use-the-parameter-to-restore-application-state
+  storage.put(`${STORAGE_KEYS.NONCE}_${nonce}`, state, {
```
tdeekens: Nice. The nounsr is important here. Is there any way we should clean up old entries? Like once you set a new one here all with a certain pattern are pruned?

packages/storage/index.js
```diff
@@ -1,4 +1,8 @@
-export const put = (key, value) => window.localStorage.setItem(key, value);
-export const get = key => window.localStorage.getItem(key);
-export const remove = key => window.localStorage.removeItem(key);
-export const clear = () => window.localStorage.clear();
+export const put = (key, value, { storage = window.localStorage } = {}) =>
```
tdeekens: Dont we have a storage package somewhere? Maybe then we actually do not need that and should:

1, Extend our package we have
2. Just operate with those APIs straight away

packages/application-shell/src/components/login-sso/login-sso.js
emmenko: The entry is cleaned when the login callback is rendered. Also, since I put it now in `sessionStorage`, is automatically cleaned up when the browser is closed.
I think it's enough...

packages/application-shell/src/components/login-sso-callback/login-sso-callback.js
emmenko: Yeah. Are you referring to this package? https://www.npmjs.com/package/jest-localstorage-mock?

tdeekens: Yes and I think it could be easily extended or duped for session storage. 

Which is why I‘d wish us go use

1. Browser APIs directly instead of mapping and renaming or
2 Build on our existing core storage package instead of adding another abstraction here which may just be confusing as to which one exists for what reason

emmenko:
So we ended up simply using `window.sessionStorage` without any abstractions on top, as it's fairly easy to test it now with jest. I guess we can also start deprecating the `storage` package maybe (as follow up).

PR #102, opened 2018-11-15
refactor(quick-access): correctly apply permissions to all commands and queries

emmenko:
Went through with @dferber90 together.


dferber90:
I adapted the tests to the changed GraphQL query variables and added subcommand filtering in places where we forgot it. Should be good to go now 🏃 

🚀🚀🚀

PR #103, opened 2018-11-15
fix(i18n): exports to not include default key

emmenko:
Same problem we had with the ui-kit when we switched to rollup. 

The messages were exposed in an object `{ default: { /* messages */ } }`.

With this change, the messages returned is correct

screen shot 2018-11-15 at 16 49 05


emmenko:
Thanks, I'll do a release later

PR #107, opened 2018-11-17
feat(application-shell-connectors): expose ownerId

### LINKED ISSUE #106: Expose project owner (organizationId) from ApplicationContext
**Describe the bug**
The `<ApplicationContext>` currently does not expose the `owner` of the `project` (a.k.a. the `organizationId`).

**Expected behavior**
The `project` exposed by the application context should contain a fiel `ownerId`.

montezume:
Fixes #106


packages/application-shell-connectors/src/components/application-context/application-context.js
```diff
@@ -38,6 +38,7 @@ const mapProjectToApplicationContextProject = project => {
     countries: project.countries,
     currencies: project.currencies,
     languages: project.languages,
+    ownerId: project.owner.id,
```
tdeekens: Only "concern" why flattening it is that the GraphQL query on `owner` resolves to an organization ideally. So the `FetchProject` should/could be able to fetch also other properties. From `name` to `teams` (then `members`) or just `version`. Flattening it out fights a bit against (even though non breaking) to expose those properties at some point. That's why maybe leaving it `owner.id` might be worthwhile for extensibility stake. What do y'all think?

tdeekens:
I don't mind either way. Lets go with `ownerId` first I think. If we need more fields we can just add `owner.*` and deprecate `ownerId`.

packages/application-shell-connectors/src/components/application-context/application-context.js
emmenko: If we need to fetch those info about the org, I’d rather have a separate `organization` query.

So I’d just keep the reference on the project.
Mostly because you wouldn’t need the org info when working on an app.

So yeah, let’s go with that and if we have need to expose more stuff we can deprecate it

tdeekens: > If we need to fetch those info about the org, I’d rather have a separate organization query.

Sorry, why separate or what does it mean? A field `organization`?

Anyway, as you said, I also do not think we'll ever need more fields.

emmenko: No I simply meant a separate graphql query `organization(id: string)` via a `FetchOrganization` or something.

But anyway, it’s only speculation for now 😉

PR #109, opened 2018-11-18
Update dependencies (incl. and importantly flopflip)

tdeekens:
#### Summary

This pull request is a simple update of dependencies but is required to be able to use the latest version of floppi di flip in the mc.

@montezume knows about the issue. Updating flopflip to latest in the mc triggers showing app-shell toggled elements while hiding in app toggled components.

After debugging flopflip for an hour I noticed that this is no mainly an issue with the library. It's cause app-kit has a strict version requirement. So has the mc. Both are installed and we run two versions of flopflip.

This feels undesired for one for flopflip but for any dependency.

Updating flopflip here and keeping it in sync (updating it in the mc) will work (debugged locally). 

A follow-up and take away from this could be:

1. Should be loosen version requirements here so yarn/npm can resolve more freely

I would like the updated version of flopflip as it removed recompose and we won't have two versions of it bundled (likely).


montezume:
can you please a branch deploy? 😄

packages/application-shell-connectors/package.json
```diff
@@ -42,8 +42,8 @@
     "recompose": "0.26.0"
   },
   "devDependencies": {
-    "react": "16.6.1",
-    "react-apollo": "2.2.4"
+    "react": "16.6.3",
+    "react-apollo": "2.3.1"
```
tdeekens: Server side rendering improvements and compatibility things with hooks.

Generally this would suffer from the same issue when not updating in the mc. Yarn here resolves to 2.3.1 while in the mc to 2.2.4 resulting in respective `node_modules` and likely double bundling.

packages/sdk/package.json
```diff
@@ -36,19 +36,19 @@
...
     "@commercetools-frontend/constants": "2.0.0",
-    "@commercetools/api-request-builder": "4.1.0",
-    "@commercetools/sdk-client": "1.5.9",
+    "@commercetools/api-request-builder": "4.1.1",
+    "@commercetools/sdk-client": "1.5.10",
     "@commercetools/sdk-middleware-correlation-id": "1.0.1",
-    "@commercetools/sdk-middleware-http": "4.0.7",
-    "@commercetools/sdk-middleware-user-agent": "1.2.10",
+    "@commercetools/sdk-middleware-http": "4.0.8",
+    "@commercetools/sdk-middleware-user-agent": "1.2.11",
```
tdeekens: This are already likely double bundled I presume due to how webpack will resolve.

packages/mc-scripts/package.json
```diff
@@ -22,18 +22,18 @@
...
-    "@svgr/webpack": "4.0.2",
+    "@svgr/webpack": "4.0.3",
     "babel-core": "7.0.0-bridge.0",
     "babel-loader": "8.0.4",
     "babel-plugin-react-intl": "3.0.1",
     "babel-polyfill": "6.26.0",
     "browserslist": "4.3.4",
     "chalk": "2.4.1",
-    "clean-webpack-plugin": "0.1.19",
+    "clean-webpack-plugin": "1.0.0",
```
tdeekens: Nothing changed. I checked. Just first major.

emmenko:
Q: have you tried using yarn `resolutions` in the mc to pull the right version?

emmenko:
Another thought: have FlopFlip only be installed in AppShell and expose it through the AppShell, so we only have one place to manage the deps.

```js
import { ToggleFeature } from '@commercetools-frontend/application-shell';
```

We might also apply this to things like `react-intl` and other "similar" libraries, although peer dependencies should be as good as well. Making `flopflip` a peer deps feels a bit weird though, since we only use it within our team...don't know, just some thoughts I had while reflecting about this sort of issues.

tdeekens:
I generally wanted to keep a discussion focussed on an issue. Now that we started... I will explain why it's __not__ flopflip related only.

> Q: have you tried using yarn resolutions in the mc to pull the right version?

That would work but I don't feel that being a sustainable solution in any case. Not for flopflip nor for any other dependency. It would overwrite asked for dependency versions with unclear side effects.

> Another thought: have FlopFlip only be installed in AppShell and expose it 

Interesting idea. However, it is __not__ only related to flopflop. Side thought: any case where:

```txt
app
   node_modules
     A
        node_modules
           B@0.0.x
     C
        node_modules
           B@0.0.y
```

is undesired for almost any web application given B is not a small utility. You would likely always want:

```txt
app
   node_modules
     A
	 B@0.0.x
     C
```

Given app-kit pins dependencies this outcome is highly unlikely as soon as versions in consumers slightly diverge and overlap in any consuming app.

> We might also apply this to things like react-intl and other "similar" libraries

Back to your suggestion: yes that would be an option. We would have to list which libraries are affected. flopflip, react-intl and react-apollo are the ones which come too mind. However, I took the detour above cause we likely do that for all as it affects any little, small dependency from lodash to literally anything.  

> Making flopflip a peer deps feels a bit weird though

I would also not do that. 

My suggestion would be to likely loosening the dependency ranges in app-kit so the consuming app has freedom to resolve to hopefully one version. 

From your comments I am not sure if you do not accept this dependency update PR altogether. I would like to move forward with it to homogenise versions in the mc as we already diverged in all Node.JS SDK deps resulting in double bundling already I believe.

tdeekens:
The the commit 5dea37a I loosened up the dependencies I suspect to be shared which we however want to resolve to be hoisted with a minor/patch range. How would you feel about that? I feel it's safe middle ground as:

1. We still have the lockfile here
2. app-kit doesn't 100% care what they resolve to as it often only sets up

The cons are:

1. Lockfile pruning might mean different versions

Let me know if you want to include this in this PR or have done/discussed separately.

emmenko:
Yes I agree, in general we need to make it safe and easy to update. We should also document properly what would be the expectations and the trade offs of the dependency management.

I think it would make sense to tackle this problem separately from this PR, which only aims of “fixing” a problem with the flopflip version.

tdeekens:
> I think it would make sense to tackle this problem separately from this PR

Okay. I reverted my previous commit. I will squash merge so it's not on `master` but remains here for illustration.

I would like to tackle flopflip _and_ other deps in this PR to not release too many versions of app kit.

We can next week discuss and document this properly.

tdeekens:
Thanks for the feedback.

PR #116, opened 2018-11-21
feat(i18n): add backwards compatible async messages

montezume:
#### Summary

When <AsyncLocaleData> is given the prop `fetchApplicationMessages`, of type function, and that function is invoked with a language, which results in a promise, then `AsyncLocaleData` will resolve that promise and invoke `children` with its result. Else, it will just invoke children with the property `applicationMessages`, which is passed into it from app-shell.

### DIFF
--- a/packages/application-shell/src/components/application-shell/application-shell.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.js
@@ -75,8 +75,12 @@ export const RestrictedApplication = props => (
         // user's browser to attempt to match the language for the correct translations.
         const userLocale = getBrowserLanguage(window);
         return (
-          <AsyncLocaleData locale={userLocale}>
-            {({ language, messages }) => {
+          <AsyncLocaleData
+            locale={userLocale}
+            fetchApplicationMessages={props.fetchApplicationMessages}
+            applicationMessages={props.applicationMessages}
+          >
+            {({ language, messages, applicationMessages }) => {
               reportErrorToSentry(error, {});
               return (
                 <ConfigureIntlProvider
@@ -84,7 +88,7 @@ export const RestrictedApplication = props => (
                   messages={mergeMessages(
                     messages,
                     i18n[language],
-                    props.applicationMessages[language]
+                    applicationMessages
                   )}
                 >
                   <ErrorApologizer />
@@ -104,8 +108,17 @@ export const RestrictedApplication = props => (
             Therefore, as long as there is no locale, the children should consider using the
             `isLoading` prop to decide what to render.
           */}
-          <AsyncLocaleData locale={user && user.language}>
-            {({ isLoading: isLoadingLocaleData, language, messages }) => (
+          <AsyncLocaleData
+            locale={user && user.language}
+            applicationMessages={props.applicationMessages}
+            fetchApplicationMessages={props.fetchApplicationMessages}
+          >
+            {({
+              isLoading: isLoadingLocaleData,
+              language,
+              messages,
+              applicationMessages,
+            }) => (
               <ConfigureIntlProvider
                 // We do not want to pass the language as long as the locale data
                 // is not loaded.
@@ -116,7 +129,7 @@ export const RestrictedApplication = props => (
                       messages: mergeMessages(
                         messages,
                         i18n[language],
-                        props.applicationMessages[language]
+                        applicationMessages
                       ),
                     })}
               >
@@ -345,7 +358,8 @@ RestrictedApplication.propTypes = {
   environment: PropTypes.object.isRequired,
   defaultFeatureFlags: PropTypes.object,
   render: PropTypes.func.isRequired,
-  applicationMessages: PropTypes.object.isRequired,
+  applicationMessages: PropTypes.object,
+  fetchApplicationMessages: PropTypes.func,
   INTERNAL__isApplicationFallback: PropTypes.bool.isRequired,
 };
 
@@ -403,7 +417,8 @@ export default class ApplicationShell extends React.Component {
     ),
     render: PropTypes.func.isRequired,
     onRegisterErrorListeners: PropTypes.func.isRequired,
-    applicationMessages: PropTypes.object.isRequired,
+    applicationMessages: PropTypes.object,
+    fetchApplicationMessages: PropTypes.func,
     // Internal usage only, does not need to be documented
     INTERNAL__isApplicationFallback: PropTypes.bool,
   };
@@ -448,6 +463,9 @@ export default class ApplicationShell extends React.Component {
                                 applicationMessages={
                                   this.props.applicationMessages
                                 }
+                                fetchApplicationMessages={
+                                  this.props.fetchApplicationMessages
+                                }
                                 INTERNAL__isApplicationFallback={
                                   this.props.INTERNAL__isApplicationFallback
                                 }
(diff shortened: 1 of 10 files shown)


packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -345,7 +358,8 @@ RestrictedApplication.propTypes = {
   environment: PropTypes.object.isRequired,
   defaultFeatureFlags: PropTypes.object,
   render: PropTypes.func.isRequired,
-  applicationMessages: PropTypes.object.isRequired,
+  applicationMessages: PropTypes.object,
+  fetchApplicationMessages: PropTypes.func,
```
montezume: maybe oneOf or something? I'm not sure how to make a propType be one or the other

packages/i18n/async-locale-data/async-locale-data.js
```diff
@@ -35,12 +42,28 @@ class AsyncLocaleData extends React.Component {
...
+    }
+
+    Promise.all(promises).then(
+      response => {
         this.setState({
           isLoading: false,
           language,
-          messages: data,
+          messages: response[0],
+          applicationMessages: response[1]
```
montezume: if no promise result, then just use the passed in applicationMessages

packages/application-shell/src/components/application-shell/application-shell.js
tdeekens: You can do

```js
applicationMessages: PropTypes.oneOfType([
    PropTypes.func,
    PropTypes.instanceOf(Whatever)
])
```

or any other type combination.

montezume: yeah that makes sense 👍

montezume: sort of weird though because applicationMessages needs to be a function, that receives a language, and returns a promise 🤔

packages/i18n/async-locale-data/async-locale-data.js
```diff
@@ -35,12 +42,28 @@ class AsyncLocaleData extends React.Component {
...
+        ? this.props.fetchApplicationMessages(language)
+        : null;
+
+    let promises = [loadI18n(language)];
+
+    if (isPromiseLike(applicationMessagePromise)) {
+      promises = promises.concat([applicationMessagePromise]);
+    }
+
+    Promise.all(promises).then(
```
tdeekens: Maybe it eases the eye if we'd do a 

```js
const messageFetchingPromises = do {
  if(isPromiseLike(applicationMessagePromise)) 
    [loadI18n(language), applicationMessagePromise];
  else
    [loadI18n(language)]
}
const [messages, applicationMessages] = await Promise.all(messageFetchingPromises);

this.setState({...})
```

packages/application-shell/src/components/application-shell/application-shell.js
tdeekens: Yeah it's a `applicationMessages(language: Language) => Promise<ApplicationMessage>` as far as I understand.

packages/i18n/async-locale-data/async-locale-data.js
```diff
@@ -1,11 +1,15 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { reportErrorToSentry } from '@commercetools-frontend/sentry';
+import isFunction from 'lodash.isfunction';
```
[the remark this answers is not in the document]
montezume: removed it 👍

packages/application-shell/README.md
```diff
@@ -65,13 +65,13 @@ ReactDOM.render(<EntryPoint />, document.getElementById('root'));
...
-| `onRegisterErrorListeners` | `func`   |    ✅    | -       | A callback function to setup global event listeners, called when the `ApplicationShell` is mounted                                            |
+| Props                      | Type               | Required | Default | Description                                                                                                                                                                                                       |
+| -------------------------- | ------------------ | :------: | ------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `applicationMessages`      | `object` or `func` |    ✅    | -       | Either an object containing all the translated messages per locale (`{ "en": { "Welcome": "Welcome" }, "de": { "Welcome": "Wilkommen" }}`), or a function that returns a promise that resolves to such an object. |
```
montezume: this was outdated

packages/i18n/package.json
```diff
@@ -34,6 +34,7 @@
     "build:es:watch": "npm run build:es -- -w"
   },
   "dependencies": {
+    "@babel/runtime": "7.1.5",
```
montezume: if you use an async function without having this as a dependency, then rollup will complain. we don't bundle dependencies with rollup anyways, so this won't increase bundle size

montezume:
Okay I addressed the PR comments and tried to update the documentation in `ApplicationShell`. I don't think we should add documentation for `AsyncLocaleData` itself, as it's not used externally - it's only used inside of AppShell, not inside of the apps directly.

packages/application-shell/README.md
emmenko: I would document how this works. Maybe add a section below about `## Loading i18n messages with code splitting`

packages/i18n/async-locale-data/async-locale-data.js
```diff
@@ -33,18 +39,38 @@ class AsyncLocaleData extends React.Component {
...
+    try {
+      const [messages, applicationMessages] = await Promise.all(
+        messageFetchingPromises
+      );
+      this.setState({
+        isLoading: false,
+        language,
+        messages: mergeMessages(
+          messages,
+          applicationMessages || this.props.applicationMessages[language]
```
emmenko: Use `this.props.applicationMessages && this.props.applicationMessages[language]` since the prop is optional and could run into runtime errors

playground/src/components/entry-point/entry-point.js
```diff
@@ -11,6 +11,19 @@ import {
 import { Sdk } from '@commercetools-frontend/sdk';
 import * as globalActions from '@commercetools-frontend/actions-global';
 import { Redirect, Route, Switch } from 'react-router-dom';
+// import applicationMessages from '../../i18n';
+
+const fetchApplicationMessages = lang =>
+  new Promise((resolve, reject) =>
+    import(`../../i18n/data/${lang}.json` /* webpackChunkName: "application-messages-[request]" */).then(
```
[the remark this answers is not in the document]
tdeekens: Insane.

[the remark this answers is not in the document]
montezume: 🤷‍♂️. We had this [request] thing before we did the split out anyways.

packages/i18n/async-locale-data/async-locale-data.js
montezume: 👍

montezume:
@emmenko documented [here](https://github.com/commercetools/merchant-center-application-kit/pull/116/commits/9f05fabe6667b49dd20d2b002c8c897730c6e8d3)

packages/application-shell/README.md
```diff
@@ -63,6 +63,34 @@ EntryPoint.displayName = 'EntryPoint';
 ReactDOM.render(<EntryPoint />, document.getElementById('root'));
 ```
 
+## Loading i18n messages with code splitting
+
+```js
+// define a function that accepts a language, and returns a promise.
+const applicationMessages = lang =>
```
emmenko: Nitpick: we can maybe name this variable `loadApplicationMessagesForLanguage`, to make it clear what it does

montezume:
anything from your side @tdeekens ?

packages/i18n/async-locale-data/async-locale-data.js
montezume: applicationMessages is defined as being a required prop in the docs though, and the current propType is object.isRequired? Should we change it to optional?

PR #117, opened 2018-11-21
feat(i18n): fix code splitting for mc-app-kit messages

montezume:
#### Summary

Even though we were using `import()`  for the react-intl-localized-strings used by the app-kit, it wasn't resulting in any code splitting because
* we weren't using `experimentalCodeSplitting` in the rollup config for this package
* we were exporting the json directly from index.js

This PR resolves both of those and updates the README for the i18n package.

I'm not sure I would consider this a breaking change, because the app-kit messages are only really supposed to be used inside of the app-kit 🤔.

If we want to keep the direct exports of the i18n files, we should look into using multiple entry points here... But I don't think it's necessary. WDYT?

### DIFF
--- a/packages/i18n/README.md
+++ b/packages/i18n/README.md
@@ -1,9 +1,10 @@
 # @commercetools-frontend/i18n
 
-MC messages and locale data (moment and react-intl).
-Supported languages:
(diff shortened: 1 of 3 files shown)


emmenko:
I'm updating the README, hold on a sec

PR #120, opened 2018-11-22
WIP: Add peerDeps while adding ranges

tdeekens:
#### Summary

This pull request is meant to drive an internal discussion how to deal with shared dependencies.


packages/application-shell/package.json
```diff
@@ -51,56 +51,49 @@
     "@commercetools-frontend/sentry": "2.1.1",
     "@commercetools-frontend/storage": "2.0.0",
     "@commercetools-frontend/url-utils": "2.0.0",
-    "@flopflip/launchdarkly-adapter": "2.4.4",
-    "@flopflip/memory-adapter": "1.1.0",
-    "@flopflip/react-broadcast": "7.2.3",
```
tdeekens: This is only configured here so I don't think we need to have it as a ranged dep here. Maybe the adapter can as that should not be shared.

packages/application-shell/package.json
```diff
@@ -51,56 +51,49 @@
...
     "apollo-cache-inmemory": "1.3.10",
-    "apollo-client": "2.4.6",
```
tdeekens: Just configured here never really used so no real dep.

packages/application-shell/package.json
```diff
@@ -51,56 +51,49 @@
...
     "apollo-link-error": "1.1.1",
     "apollo-link-http": "1.5.5",
     "apollo-link-logger": "1.2.3",
     "apollo-link-retry": "2.2.5",
-    "classnames": "2.2.6",
-    "common-tags": "1.8.0",
-    "debounce-async": "0.0.2",
-    "downshift": "3.1.5",
-    "fast-equals": "1.6.1",
-    "formik": "1.3.2",
```
tdeekens: The above are all shared and used across both so both lib/app have a requirement towards them.

packages/application-shell/package.json
tdeekens: Which one do you mean? I wrote the adapter does not have to be a peer dep as it is only used internally. 

Re-exporting is the other option suggested on the issue yes. I was wondering mostly which packages we want (if) to reexport and which we want as peer dep and which as a wider range. 

I opened this PR hoping to fuel getting such feedback. What are the concrete pros and cons you see in re-exporting? 

I think we have two proposals in how we could (if) re-export. Through props or through ESM.

emmenko: Given our experience, we usually need the exports from `@flopflip/react-broadcast` when working on applications, right?

What about my other suggestion of having an extra package?

tdeekens: I‘d be interested in contrasting approaches from props and cons. 

My main concern about packages is, as mentioned earlier, this is not only about flopflip but also about other dependencies. Having 1-n packages each re-exporting some library feels really odd and boilerplate heavy. 

We could obviously just re-export from the app-shell so you can import { ToggleFeature, branchOnFeatureToggle } from app-shell. That would however be a breaking change and delutes where the dependency originates from. 

Peer deps might feel semantically a bit wrong I agree and leaks the problem to apps. However other libs are also a peer dep like react-router etc. 

What‘s your line of argument weighting things?

emmenko: > his is not only about flopflip but also about other dependencies. 

Right, maybe we list first what dependencies are "affected" by this?

About the peer deps, I would do that only for dependencies that are usually used when you develop an app (e.g. react, etc).
I think flopflip is a special case, because we only use it internally at CT, so we can also treat it as a special case (hence the idea with the package).

tdeekens: > maybe we list first what dependencies are "affected" by this?

😂 It‘s what I opened this PR for and wrote once or twice again already we‘d have to declare. 🤙🏻

I know what other peer deps are and why they are declared as such. Flopflip is also used when you develop an app some what. 

Still I‘d like to dicuss pros and cons for some scenarios to decide on how we proceed.

tdeekens:
Didn‘t create a meaningful dicussion around objective ways of to solve this as I hoped it could. As a result closing.

PR #121, opened 2018-11-22
fix: app crashing when project is not found

emmenko:
We found out that when the app was requesting a project that does not exist (or the user does not have access to), the app would crash.

image

The problem was that in some places we were assuming that the `project` was defined, which might not always be. Now, when the project is not found, we render the app correctly and show the 404 page.

image

### DIFF
--- a/packages/application-shell/src/components/application-shell/application-shell.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.js
@@ -222,11 +222,12 @@ export const RestrictedApplication = props => (
                             <FetchProject projectKey={projectKeyFromUrl}>
                               {({ isLoading: isLoadingProject, project }) => {
                                 // Render the loading navbar as long as all the data
-                                // hasn't been loaded.
+                                // hasn't been loaded, or if the project does not exist.
                                 if (
                                   isLoadingUser ||
                                   isLoadingLocaleData ||
-                                  isLoadingProject
+                                  isLoadingProject ||
+                                  !project
                                 )
                                   return <LoadingNavBar />;
 
--- a/packages/application-shell/src/components/quick-access/create-commands.js
+++ b/packages/application-shell/src/components/quick-access/create-commands.js
@@ -316,6 +316,7 @@ export default ({
         ].filter(Boolean),
       },
     project &&
+      project.languages &&
       project.languages.length > 1 && {
         id: 'action/set-resource-language',
         text: intl.formatMessage(messages.setResourceLanguage),
--- a/packages/application-shell/src/components/quick-access/index.js
+++ b/packages/application-shell/src/components/quick-access/index.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import Loadable from 'react-loadable';
+import { reportErrorToSentry } from '@commercetools-frontend/sentry';
 import ButlerContainer from './butler-container';
 import * as gtm from '../../utils/gtm';
 import trackingEvents from './tracking-events';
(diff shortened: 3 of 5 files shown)


packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -222,11 +222,12 @@ export const RestrictedApplication = props => (
...
                               {({ isLoading: isLoadingProject, project }) => {
                                 // Render the loading navbar as long as all the data
-                                // hasn't been loaded.
+                                // hasn't been loaded, or if the project does not exist.
                                 if (
                                   isLoadingUser ||
                                   isLoadingLocaleData ||
-                                  isLoadingProject
+                                  isLoadingProject ||
+                                  !project
```
emmenko: This is to avoid rendering the navbar. If we don't do that, the query to the settings api to fetch the custom app links will fail with `401`, because the `projectKey` is missing, causing the app to log the user out.
So we only render the navbar if there is a project

packages/application-shell/src/components/quick-access/index.js
```diff
@@ -20,7 +21,21 @@ export default class QuickAccessTrigger extends React.Component {
     document.removeEventListener('keydown', this.handler);
   }
 
-  state = { isVisible: false };
+  // If for some reason an error is thrown, catch it and do not render QuickAccess
+  // to avoid crashing the entire app.
+  componentDidCatch(error, errorInfo) {
```
emmenko: This is just to prevent possible errors from quickaccess to crash the app

packages/permissions/src/utils/has-permissions.js
```diff
@@ -36,7 +36,10 @@ const hasManageProjectPermission = actualPermissions =>
 //     'ViewProducts'
 // - actualPermissions:
 //     { canViewProducts: true, canManageOrders: false }
-export const hasPermission = (demandedPermission, actualPermissions) =>
+// NOTE: in case the `actualPermissions` are not defined, fall back to an empty object.
+// This might be the case when the permissions for a user/project could not be loaded
+// (e.g. project not found).
+export const hasPermission = (demandedPermission, actualPermissions = {}) =>
```
emmenko: This is the actual fix for the thrown error

packages/application-shell/src/components/application-shell/application-shell.js
qmateub: First of all sorry for point something that was not added in this PR but would make sense to extract this condition into a variable for the sake of readability `canRenderNavBar` or something like that? 😋

packages/application-shell/src/components/quick-access/create-commands.js
```diff
@@ -316,6 +316,7 @@ export default ({
         ].filter(Boolean),
       },
     project &&
+      project.languages &&
```
qmateub: Could we also use directly the conditional chain? something like `project?.languages?.length > 1 `

packages/application-shell/src/components/quick-access/index.js
qmateub: 👌

packages/application-shell/src/components/quick-access/create-commands.js
emmenko: I think we said we prefer not to use this syntax in the public repos

qmateub: OMG 🤦‍♂️ My bad, though we were in the mc repo 🙏

PR #124, opened 2018-11-23
chore: sync en.json -> core.json after pulling messages from tx

emmenko:
Just to sync `en.json` -> `core.json` after pulling new messages from tx


adnasa:
> Just to sync en.json -> core.json after pulling new messages from tx

For more context to others, we do this to ensure that new translators has up-to-date original message for better reference

adnasa:
btw, shall we do a sync now before merge or?

dferber90:
> Just to sync en.json -> core.json after pulling new messages from tx

Sidenote: Sometimes when testing it's useful to use the `core.json` messages to ensure tests don't break because translations change. Having the sync between messages prevents us from doing that. I'm not too concerned with it, but wanted to point it out so you're aware.

emmenko:
> Sometimes when testing it's useful to use the `core.json` messages to ensure tests don't break because translations change.

I think we usually mock `intl.formatMessage` to return the message id "as the translated message", so we just assert on ids.
Is that what you're talking about or something different?

dferber90:
Mocking is the case for `enzyme` tests, but not when using `react-testing-library` (or our `test-utils`, where the `IntProvider` is fed the `core.json` messages). Maybe it doesn't matter as much for app-kit. It is more important for the MC itself though.

emmenko:
> Mocking is the case for enzyme tests, but not when using react-testing-library

Yeah I see ok.

> It is more important for the MC itself though.

We also have the same sync happening there. Probably we should not rely too much on the translation messages.

Thanks for pointing it out though 💯

PR #126, opened 2018-11-23
docs(changelog): for v2.3.3 [skip ci]

emmenko:


emmenko:
Putting on hold until we include #135 and #136

tdeekens:
Makes sense. I am super eager to learn how well auto-publishing to `@next` works on UIKit. Given it proves valuable we can do the same here maybe and just move tags every two weeks or so having kind of like a release train.

emmenko:
Closing in favor of #138

PR #135, opened 2018-11-26
feat(quick-access): respect feature toggles

dferber90:
Respects feature-toggles for QuickAccess.

I did not go through the feature-toggles in depth. Let me know in case I forgot to hide some Quick Access links to routes which are hidden by your team's feature toggles 🙇


dferber90:
The tests were failing because they were running without feature toggles. Fixed it now 👍

PR #136, opened 2018-11-26
chore: update deps

emmenko:
Closes #129
Closes #130 
Closes #131 
Closes #132 
Closes #133 
Closes #134


packages/application-shell/package.json
```diff
@@ -118,7 +118,8 @@
     "react-dom": ">=16 <17",
     "react-intl": ">=2 <3",
     "react-router": ">=4 <5 || >=4.4.0-beta.0 <5",
-    "react-router-dom": ">=4 <5 || >=4.4.0-beta.0 <5"
+    "react-router-dom": ">=4 <5 || >=4.4.0-beta.0 <5",
+    "react-testing-library": ">=5 || <6"
```
emmenko: I moved this as a peer dep because in `jest-preset-mc-app` it was already defined as such.

packages/application-shell/package.json
```diff
@@ -54,13 +54,13 @@
     "@flopflip/launchdarkly-adapter": "2.4.4",
     "@flopflip/memory-adapter": "1.1.0",
     "@flopflip/react-broadcast": "7.2.3",
-    "apollo-cache-inmemory": "1.3.10",
-    "apollo-client": "2.4.6",
-    "apollo-link": "1.2.3",
-    "apollo-link-error": "1.1.1",
-    "apollo-link-http": "1.5.5",
+    "apollo-cache-inmemory": "1.3.11",
+    "apollo-client": "2.4.7",
```
tdeekens: ⚠️ just we don't forget to update it in the mc-fe too otherwise we have two of them.

packages/application-shell/package.json
tdeekens: Should solve us not being able to update in the mc-fe.

packages/application-shell/package.json
emmenko: Actually this is never used in the MC directly:

image

PR #142, opened 2018-11-27
feat(app-shell/navbar): add menu item for custom apps

emmenko:


packages/i18n/data/core.json
```diff
@@ -59,6 +59,7 @@
   "NavBar.Categories.list": "Categories list",
   "NavBar.Categories.search": "Categories search",
   "NavBar.Categories.title": "Categories",
+  "NavBar.CustomApplications.title": "Custom Applications",
```
dferber90: Do you wanna add it to Quick Access as well (behind a feature-toggle)? Just in case it didn't cross your mind.

emmenko: Ah right, good idea

PR #143, opened 2018-11-28
feat(test-utils): run tests agains actual ApolloProvider

pa3:
#### Summary

I am about to try new way of mocking Graphql: one which will mock http requests, and not relay on `<MockProvider />`. To do this, I need:

1.  a way to render Component under test with production Apollo setup
2. force `render` method from `test-utils` to re-create Apollo Client before each test, so that tests will not share the same Apollo Cache.

#### Description

Our current approach of mocking Graphql is utilising `<MockProvider/>` helper from `react-apollo`, but that has a bunch of drawbacks:

1. Tests are too tightly coupled to the implementation of a Component being tested, because
 * test have to import Graphql queries used by those Components
 * appropriate responses are matched based on query variables values, which is implementation detail of a Component under test.
2. `<MockProvider/>` does not do a good job in letting us know, when it wasn't able to find appropriate response for a given request. That makes keeping fixtures up to date with query variable values hard and frustrating. 
3. Tests do not give enough confidence, as they always return same response no matter what was the shape of a query fired by a Component. This way it's quite easy to end up with false-positive results.  
4. Our setup of ApolloClient is not being exercised in a test harness.

#### Solution

I will combine our actual Graphql schema, `factory-girl`-like approach and http-mocking to solve all that drawbacks. For now I gonna experiment with that in MC repo, and once I will be happy with the result I will spin off a PR in this repo.

But to start experimenting I need to changes in app-kit introduced in this PR.

### DIFF
--- a/packages/application-shell/src/test-utils/test-utils.js
+++ b/packages/application-shell/src/test-utils/test-utils.js
@@ -4,7 +4,9 @@
 // import { render } from ""@commercetools-frontend/application-shell/test-utils"
 // and then use it together with react-testing-library.
 import React from 'react';
+import PropTypes from 'prop-types';
 import { Router } from 'react-router-dom';
+import { ApolloProvider } from 'react-apollo';
 import { render as rtlRender } from 'react-testing-library';
 import { createMemoryHistory } from 'history';
 import { IntlProvider } from 'react-intl';
@@ -14,6 +16,7 @@ import memoryAdapter from '@flopflip/memory-adapter';
 import { Provider as StoreProvider } from 'react-redux';
 import { ApplicationContextProvider } from '@commercetools-frontend/application-shell-connectors';
 import { createReduxStore } from '../configure-store';
+import { createApolloClient } from '../configure-apollo';
 
 // Reset memoryAdapter after each test, so that the next test accepts the
 // defaultFlags param.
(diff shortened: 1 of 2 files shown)


packages/application-shell/src/test-utils/test-utils.js
```diff
@@ -60,6 +62,16 @@ const defaultPermissions = { canManageProject: true };
 const mergeOptional = (defaultValue, value) =>
   value === null ? undefined : { ...defaultValue, ...value };
 
+const ApolloProvider = props =>
+  props.mocks ? (
+    <ApolloMockProvider {...props} />
+  ) : (
+    <ApolloActualProvider client={createApolloClient()}>
```
emmenko: So this creates a new apollo client each time it (re)renders?

packages/application-shell/src/test-utils/test-utils.js
```diff
@@ -60,6 +62,16 @@ const defaultPermissions = { canManageProject: true };
...
+      {props.children}
+    </ApolloActualProvider>
+  );
+ApolloProvider.displayName = 'ApolloProvider';
```
dferber90: @pa3 How do you feel about creating a separate `render` function to keep the experimental stuff out of our actual `render` fn used in tests?

packages/application-shell/src/test-utils/test-utils.js
pa3: Exactly.

packages/application-shell/src/test-utils/test-utils.js
pa3: That can be done, but I am also happy with the semantics: "you didn't gave us mocks, you probably don't want to mock Apollo".

dferber90: From what I understand you're still experimenting with this new approach. Adding it to the "official" `render` function makes it seem as-if it is a recommended way already. That's why I would prefer a separate function for it as we might want to change/remove it after the experimentation.

packages/application-shell/src/test-utils/test-utils.js
adnasa: I did not know that `createApolloClient` returns a closure.
Because in `configure-apollo`, you exported `createApolloClient` already being invoked.

```js
export default createApolloClient();
```

packages/application-shell/src/test-utils/test-utils.js
pa3: @dferber90 Updated this PR to have additional `experimentalRender` function.

PR #144, opened 2018-11-28
Fix double dependency bundling

tdeekens:
#### Summary

This pull request re-exports crucial dependencies and adds ranges to commonly shared utility deps.

#### Description

More information can be found on an internal chapter issue.


packages/application-shell/package.json
```diff
@@ -73,18 +73,17 @@
...
     "jwt-decode": "2.2.0",
-    "lodash.camelcase": "4.3.0",
-    "lodash.flatmap": "4.5.0",
-    "lodash.has": "4.5.2",
-    "lodash.isnil": "4.0.0",
-    "lodash.last": "3.0.0",
-    "lodash.omit": "4.5.0",
-    "lodash.once": "4.1.1",
-    "lodash.snakecase": "4.1.1",
+    "lodash.camelcase": "^4.3.0",
```
montezume: so this leads to no more double bundling of lodash libs?

packages/application-shell/package.json
```diff
@@ -102,13 +101,14 @@
   },
   "devDependencies": {
     "@commercetools-frontend/ui-kit": "4.0.0",
-    "react": "16.6.3",
+    "prop-types": "15.6.2",
```
montezume: why is this necessary?

packages/application-shell/src/index.js
```diff
@@ -27,3 +27,14 @@ export {
 } from './utils/setup-global-error-listener';
 export { selectUserId, selectProjectKeyFromUrl } from './utils';
 export { default as AsyncChunkLoader } from './components/async-chunk-loader';
+
+/**
+ * NOTE:
+ *    These are library re-exports which are shared between app-kit and
+ *    MC apps. Not having them re-exported here often leads to bundling
+ *    them twice given app-kit and any MC app require different versions.
+ *    Something they never should anyways and causes breakage.
```
montezume: 👍 for the comment

packages/application-shell/package.json
tdeekens: That's what I am trying to solve. Given they update easily in diverging ways here and on the mc-fe I don't see harm in having a range given how mature the library is too.

packages/application-shell/package.json
tdeekens: I found it super odd to have it as a dependency. Prop-types to me are always a devDep?

dferber90: My understanding is that as `prop-types` are imported into the production code, they should be regular dependencies (just like lodash etc which is also imported).

Unless we use a compile step to remove the imports - but even then nothing would import the prop-types and that would be fine to have them as a regular dependency as well.

tdeekens: Ok. Reverted in last commit.

tdeekens:
> Should we also adjust the renovate config?

The renovate config has pinOnlyDevDependencies so it would not up PRs to bin the normal deps again. Am I missing something.

emmenko:
> The renovate config has pinOnlyDevDependencies so it would not up PRs to bin the normal deps again. 

Alright, was just asking as you know more about that suff 😇

emmenko:
Ok thanks.

We just need to be careful then when updating deps now 🤓

tdeekens:
> We just need to be careful then when updating deps now 🤓

Yes updating to any breaking change of flopflip here will need to be taken into account when updating app-kit in apps. Given that I am the maintainer of the library I think we're ~safe~unsafe.

emmenko:
I meant for the range version updates. From time to time we had problems with patch/minor versions breaking things...remember? 😅

pa3:
Cool!

packages/application-shell/src/index.js
```diff
@@ -27,3 +27,14 @@ export {
...
+ *
+ *    More information can be found here: https://github.com/commercetools/fe-chapter-notes/issues/61
```
adnasa: so we link to our FE-chapter on OSS? 🤔

tdeekens: Why not. It will show a 404 anyway. It doesn't contain a ticket number of anything. I wouldn't worry too much about it.

PR #147, opened 2018-11-28
chore: update ui-kit

montezume:
So @qmateub can have the `Avatar`

### DIFF
--- a/playground/package.json
+++ b/playground/package.json
@@ -20,7 +20,7 @@
     "@commercetools-frontend/mc-http-server": "3.0.0",
     "@commercetools-frontend/permissions": "3.0.0",
     "@commercetools-frontend/sdk": "3.0.0",
-    "@commercetools-frontend/ui-kit": "2.0.0-rc.11",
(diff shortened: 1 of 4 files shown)


playground/package.json
```diff
@@ -20,7 +20,7 @@
     "@commercetools-frontend/mc-http-server": "3.0.0",
     "@commercetools-frontend/permissions": "3.0.0",
     "@commercetools-frontend/sdk": "3.0.0",
-    "@commercetools-frontend/ui-kit": "2.0.0-rc.11",
+    "@commercetools-frontend/ui-kit": "4.0.2",
```
dferber90: @montezume This is quite the jump here. Did you try the playground out?

montezume: yessir. It doesn't even use ui-kit. Maybe I should just remove it?

montezume: yeah I tried it out. We are only using ~4 components, and none of those had changes

  LoadingSpinner, Table, Spacings, Constraints, Text

PR #149, opened 2018-11-28
refactor(navbar): use IconSwitcher to statically render icons based on icon names

emmenko:
As discussed, this removes the `import { Icons } from '@commercetools-frontend/ui-kit'` and uses a component to statically render an icon (explicitly imported) based on the icon name as string.

### DIFF
--- a/packages/application-shell/src/components/navbar/navbar.js
+++ b/packages/application-shell/src/components/navbar/navbar.js
@@ -68,38 +68,38 @@ import messages from './messages';
 // in which case we don't have a clue on the icons used.
 // A possible solution to that would be to import the SVG using
 // a `<img src>`, given that the SVG are available from a public URL.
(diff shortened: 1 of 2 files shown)


packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -8,7 +8,24 @@ import { ToggleFeature, injectFeatureToggle } from '@flopflip/react-broadcast';
...
+  CartIcon,
+  BoxIcon,
+  GearIcon,
+  SupportIcon,
+  HeartIcon,
+  PaperclipIcon,
+  PluginIcon,
+  RocketIcon,
+  StarIcon,
+  ConnectedSquareIcon,
```
montezume: I wonder if we should dynamically import these 🤔 

only half joking 😆

PR #151, opened 2018-11-29
[MCD-1509] Refactor dropdown user settings menu

qmateub:
#### Summary

Ticket: https://jira.commercetools.com/browse/MCD-1509
Design: https://projects.invisionapp.com/d/main#/projects/prototypes/16095086

#### Description

This PR contains the changes to refactor the user settings menu to the new designs including:

* add downshift for the outer click functionality
* Address minor design issues (border, inside padding, separators, user information)
* Hide projects, organizations and profile options if we are in account section.
* Pass email from `app-bar` so we can use it

| When in the MC | When in account section  |  
|:-:|---|
| image | image |

### DIFF
--- a/packages/application-shell/src/components/user-settings-menu/user-settings-menu.js
+++ b/packages/application-shell/src/components/user-settings-menu/user-settings-menu.js
@@ -1,8 +1,9 @@
 import PropTypes from 'prop-types';
 import React from 'react';
-import { Link } from 'react-router-dom';
+import { Link, withRouter, matchPath } from 'react-router-dom';
 import { FormattedMessage } from 'react-intl';
 import classnames from 'classnames';
+import Downshift from 'downshift';
 import { ToggleFeature } from '@flopflip/react-broadcast';
 import {
   withMouseOverState,
@@ -12,7 +13,6 @@ import {
   Avatar,
 } from '@commercetools-frontend/ui-kit';
 import { LOGOUT_REASONS } from '@commercetools-frontend/constants';
-import MenuStateContainer from '../menu-state-container';
 import Card from '../../from-core/card';
 import { MCSupportFormURL } from '../../constants';
 import { PROJECTS_LIST, ORGANIZATIONS_LIST } from '../../feature-toggles';
@@ -40,6 +40,7 @@ UserAvatar.propTypes = {
   gravatarHash: PropTypes.string.isRequired,
   firstName: PropTypes.string,
   lastName: PropTypes.string,
+  email: PropTypes.string,
   // Injected
   handleMouseOver: PropTypes.func.isRequired,
   handleMouseOut: PropTypes.func.isRequired,
@@ -47,21 +48,32 @@ UserAvatar.propTypes = {
 };
 const UserAvatarWithHoverState = withMouseOverState(UserAvatar);
 
-export default class UserSettingsMenu extends React.PureComponent {
+export class UserSettingsMenu extends React.PureComponent {
   static displayName = 'UserSettingsMenu';
 
   static propTypes = {
     firstName: PropTypes.string,
     lastName: PropTypes.string,
+    email: PropTypes.string,
     gravatarHash: PropTypes.string.isRequired,
+
+    // withRouter
+    location: PropTypes.shape({
+      pathname: PropTypes.string.isRequired,
+    }).isRequired,
   };
 
   render() {
+    const isAccountPath = Boolean(
+      matchPath(this.props.location.pathname, {
+        path: '/account',
+      })
+    );
     return (
       <div className={styles.container} data-test="user-settings-menu">
-        <MenuStateContainer>
+        <Downshift>
           {({ isOpen, toggleMenu }) => (
-            <div>
+            <div onBlur={toggleMenu}>
               <div
                 className={styles['settings-container']}
                 onClick={toggleMenu}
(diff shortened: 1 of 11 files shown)


packages/application-shell/src/components/user-settings-menu/user-settings-menu.js
```diff
@@ -40,28 +40,40 @@ UserAvatar.propTypes = {
...
+
+    // withRouter
+    location: PropTypes.shape({
+      pathname: PropTypes.string.isRequired,
+    }).isRequired,
   };
 
   render() {
+    const isAccountPath = Boolean(
+      matchPath(this.props.location.pathname, {
```
montezume: I would rather if this is a prop that's passed in.

Something like "isFullMenu"

packages/application-shell/src/components/user-settings-menu/user-settings-menu.js
```diff
@@ -74,43 +86,65 @@ export default class UserSettingsMenu extends React.PureComponent {
...
+                            {[this.props.firstName, this.props.lastName]
+                              .join(' ')
+                              .trim()}
+                          </Text.Body>
+                          <Text.Body truncate>{this.props.email}</Text.Body>
+                        </Spacings.Stack>
+                      </Spacings.Inline>
+                    </Spacings.Inset>
+                  </div>
+                  {!isAccountPath && (
```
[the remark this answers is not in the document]
tdeekens: It's a design decision. Feel free to contact them, they assured us it's wanted.

packages/application-shell/src/components/user-settings-menu/user-settings-menu.js
qmateub: Yeah, I thought that too but the problem is how many levels we would need to forward that prop down the tree 🤔

qmateub: @montezume If you take a look at the tree we can not pass such prop from the parents without using withRouter at some point. I think then this is a valid solution 😅

montezume: Can't you pass it from ApplicationShell -> AppBar -> UserSettingsMenu?

packages/application-shell/src/components/user-settings-menu/user-settings-menu.js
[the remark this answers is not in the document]
emmenko: I thought so yes, I'll ask them

packages/application-shell/src/components/user-settings-menu/user-settings-menu.js
montezume: Hm, isn't it also shown during loading though? So it would be undefined then? Maybe the initial idea was better.

tdeekens: Is this still an open conversation open for improvement?

packages/application-shell/src/components/user-settings-menu/user-settings-menu.js
```diff
@@ -40,28 +40,40 @@ UserAvatar.propTypes = {
...
   static propTypes = {
     firstName: PropTypes.string,
     lastName: PropTypes.string,
+    email: PropTypes.string,
     gravatarHash: PropTypes.string.isRequired,
+
+    // withRouter
+    location: PropTypes.shape({
+      pathname: PropTypes.string.isRequired,
+    }).isRequired,
```
tdeekens: You can also pass the location from the top not having to `withRouter` injecting a ton of other props?

packages/application-shell/src/components/user-settings-menu/user-settings-menu.mod.css
```diff
@@ -8,12 +8,12 @@
 
 .menu {
   position: absolute;
-  border: 1px var(--color-gray-60) solid;
+  border: 1px var(--color-green-40) solid;
   box-shadow: var(--shadow-7-first), var(--shadow-7-second);
-  width: 200px;
+  width: 315px;
```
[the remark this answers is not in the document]
tdeekens: I like a variable for this on top of the file sorry. Why however can't the menu be "simple" and just grow with content? /cc @Shecki

packages/application-shell/src/components/user-settings-menu/user-settings-menu.mod.css
```diff
@@ -35,7 +34,6 @@
   top: -8px;
```
tdeekens: Can we change those to normal spacings while we're at it?

packages/application-shell/src/components/user-settings-menu/user-settings-menu.js
```diff
@@ -40,28 +40,40 @@ UserAvatar.propTypes = {
...
+        path: '/account',
+      })
+    );
     return (
       <div className={styles.container} data-test="user-settings-menu">
-        <MenuStateContainer>
+        <Downshift>
           {({ isOpen, toggleMenu }) => (
-            <div>
+            <div onBlur={toggleMenu}>
```
tdeekens: Why are we not spreading the downshift props? That's what I meant a day ago to have better accessibility (e.g. click outside, keyboard navigation):

```js
<Downshift
    onChange={selection => alert(`You selected ${selection.value}`)}
    itemToString={item => (item ? item.value : '')}
  >
    {({
      getInputProps,
      getItemProps,
      getLabelProps,
      getMenuProps,
      isOpen,
      inputValue,
      highlightedIndex,
      selectedItem,
    }) => (
      <div>
        <label {...getLabelProps()}>Enter a fruit</label>
        <input {...getInputProps()} />
        <ul {...getMenuProps()}>
          {isOpen
            ? items
                .filter(item => !inputValue || item.value.includes(inputValue))
                .map((item, index) => (
                  <li
                    {...getItemProps({
                      key: item.value,
                      index,
                      item,
                      style: {
                        backgroundColor:
                          highlightedIndex === index ? 'lightgray' : 'white',
                        fontWeight: selectedItem === item ? 'bold' : 'normal',
                      },
                    })}
                  >
                    {item.value}
                  </li>
                ))
            : null}
        </ul>
      </div>
    )}
  </Downshift>,
```

packages/application-shell/src/components/user-settings-menu/user-settings-menu.js
qmateub: Yeah I can do that, I asked you guys what you think https://github.com/commercetools/merchant-center-application-kit/pull/151#discussion_r237447918

tdeekens: Sorry didn't see that.

packages/application-shell/src/components/user-settings-menu/user-settings-menu.js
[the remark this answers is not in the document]
emmenko: Short update here, just as FYI.

I spoke to @Shecki yesterday and he explained the motivation from a UX perspective. What is worrying me though is that this decision comes from a "_We think that it's best..._" without any kind of user research.

Now, I don't want raise a big fuss for such a small change. However, what's bothering me is that we still seem to make those decisions purely based on our assumptions **without proper user research or user testing**. For example, we could have used feature toggling here to target different user groups to test with or without the links and see what works better. Only then we can "safely" make a decision what's the better usability.

We all want to provide a good and usable product but we also need to do it properly and with the right tools.

Cheers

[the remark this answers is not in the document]
qmateub: Couldn't agree more 👍 Thanks for following the discussion with design and let us know @emmenko 🙏

qmateub:
FYI Guys,

I'm trying to improve this with downshift, that's why you don't see movement around 😋

qmateub:
@emmenko @tdeekens @montezume 

Okay, so after acquiring some experience with `downshift` I think I manage to improve this component a little bit in terms of functionality&accesibility. Let me share what I found:

| We want arrows navigation | We want tabs navigation | We want the 2 worlds |
|---|---|---|
| For achieving this we will need to render a fake `input` in the background and after the menu gets opened you can just go down with the keyboard arrows  | We need to use a `button` in the background so the component is tabbable after the project switcher  | I think I can say that this cannot be done only with `downshift` or at least I was not able to achieve it 😓 |

I looked at several popular sites on how this is done, and for the most of them the tabs seems that is the most used approach (google, ebay, GH), and that's the one I took (Obviously this will need some validation from UX).

So in general, the improvements are:

🍰  Now the user menu is tabable so you can tab and press enter to show the menu.
🍭  All the menu options can be tabbed.
🍰 After you select one option or click ENTER in one option the menu closes out.
🍰 If you click outside the menu automatically closes out.
🍰  If you press ESC while in the menu button (not in an item), the menu automatically closes out.

Let me know your thoughts on this, as I said, it's the first time working with this library 🙏

qmateub:
I've also improved the `location` thingy in 515c14f. Please @tdeekens verify that this is what you were suggestion and I hope that this make sense for the application-shell @emmenko 😅

packages/application-shell/src/components/user-settings-menu/user-settings-menu.mod.css
```diff
@@ -24,18 +33,16 @@
...
   border-style: solid;
   border-width: 0 6px 6px;
-  border-color: transparent transparent var(--color-white) transparent;
   z-index: 1;
 }
 
 .menu::after {
   content: '';
   position: absolute;
   right: 28px;
```
tdeekens: `--menu-offset-right: 28px;` 🤷‍♂️

packages/application-shell/src/components/user-settings-menu/user-settings-menu.js
```diff
@@ -40,120 +40,184 @@ UserAvatar.propTypes = {
...
+                      </React.Fragment>
+                    )}
+                    {
+                      // FIXME: added as urgent request for GDPR changes
+                      // Should be properly added on the CTP-1209 task
+                    }
+                    <a
+                      href={`https://commercetools.com/privacy`}
+                      target="_blank"
+                      onClick={toggleMenu}
```
montezume: how come we need to toggleMenu here?

also can you add rel="noopener noreferrer"

tdeekens: Some dark corner of my brain tells
me we had it and removed it. Can not remember the reason nor if it was good Maybe dig the blame a bit to see and if there is nothing change it 😇

montezume: @tdeekens yeah since the links are "internal", there isn't a security concern here 😄.

packages/application-shell/src/components/user-settings-menu/user-settings-menu.js
```diff
@@ -162,10 +162,6 @@ export class UserSettingsMenu extends React.PureComponent {
                         </ToggleFeature>
                       </React.Fragment>
                     )}
-                    {
-                      // FIXME: added as urgent request for GDPR changes
-                      // Should be properly added on the CTP-1209 task
-                    }
```
qmateub: @tdeekens

PR #152, opened 2018-11-29
Fix flopflip being bundled (remove rollup-plugin-peer-deps-external)

tdeekens:
#### Summary

This pull requests fixes flopflip (and potentially other) deps to be added to an exclusion list so they're bundled.

#### Description

Given we want to achieve the above we have to remove rollup-plugin-peer-deps-external and do the work ourselves. It's not much work as one of the aims of the module seems to be to auto-add things like `lodash/flatMap` if it finds `lodash`.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -49,12 +49,24 @@ const externalDependencies = pkgDependencies
       !dependenciesRequiringToBeBundled.includes(potentiallyExternalDependency)
   );
 
+/**
+ * Note:
+ *    This function is copied from rollup-plugin-peer-deps-external.
+ *    Reference: https://github.com/Updater/rollup-plugin-peer-deps-external/blob/master/src/get-modules-matcher.js
+ *    Without it rollup has more warnings about unresolved dependencies and unused external imports.
+ */
+const createExternalsFromDependencies = modulesNames => {
+  const moduleRegExp = module => new RegExp(`^${module}(\\/.+)*$`);
+  const regexps = modulesNames.map(moduleRegExp);
+  return id => regexps.some(regexp => regexp.test(id));
+};
+
 const config = {
   output: {
     name: pkg.name,
     sourcemap: true,
   },
-  external: externalDependencies,
+  external: createExternalsFromDependencies(externalDependencies),
   plugins: [
     babel({
       exclude: '**/node_modules/**',
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -7,7 +7,6 @@ const resolve = require('rollup-plugin-node-resolve');
 const json = require('rollup-plugin-json');
 const commonjs = require('rollup-plugin-commonjs');
 const postcss = require('rollup-plugin-postcss');
-const peerDeps = require('rollup-plugin-peer-deps-external');
 const builtins = require('rollup-plugin-node-builtins');
 const babelPluginImportGraphQL = require('babel-plugin-import-graphql');
 const postcssCustomProperties = require('postcss-custom-properties');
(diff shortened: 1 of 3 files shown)


rollup.config.js
```diff
@@ -27,15 +26,36 @@ const isCJS = format === 'cjs';
...
+const pkgOptionalDependencies = Object.keys(pkg.optionalDependencies || {});
+
+const externalDependencies = pkgDependencies
+  .concat(pkgPeerDependencies)
+  .concat(pkgOptionalDependencies)
+  .filter(
+    potentiallyExternalDependency =>
+      !dependenciesRequiringToBeBundled.includes(potentiallyExternalDependency)
+  );
+
```
tdeekens: I hope the naming makes sense. I don't mind having it here straight in the config to make it explicit.

tdeekens:
This results in a dist file which contains

```js
exports.ToggleFeature = __chunk_2.toggleFeature;
exports.injectFeatureToggle = __chunk_2.injectFeatureToggle$1;
exports.injectFeatureToggles = __chunk_2.injectFeatureToggles$1;
exports.branchOnFeatureToggle = __chunk_2.branchOnFeatureToggle$1;
exports.ConfigureFlopFlip = __chunk_2.Configure;
exports.useFeatureToggle = __chunk_2.useFeatureToggle;
exports.version = __chunk_2.version;
exports.ReconfigureFlopFlip = __chunk_2.reconfigureAdapter;
exports.FetchUser = __chunk_6.DeprecatedFetchUser;
...
```

whereas before it was mapping over flopflip exports cause it wasn't sure if they'd be there or not.

tdeekens:
The rollup-plugin-peer-deps-external has some internal module matching we can just copy over. 

Last commit does that. With it we get a build output of:

```txt
./index.js → ./dist...
(!) Unresolved dependencies
https://rollupjs.org/guide/en#warning-treating-module-as-external-dependency
create-react-context (imported by ../../node_modules/@flopflip/react-broadcast/dist/@flopflip-react-broadcast.es.js, ../../node_modules/@flopflip/react/dist/@flopflip-react.es.js)
lodash.flowright (imported by ../../node_modules/@flopflip/react-broadcast/dist/@flopflip-react-broadcast.es.js, ../../node_modules/@flopflip/react/dist/@flopflip-react.es.js)
react-fast-compare (imported by ../../node_modules/@flopflip/react/dist/@flopflip-react.es.js)
react-is (imported by ../../node_modules/@flopflip/react/dist/@flopflip-react.es.js)
lodash.intersection (imported by ../../node_modules/@flopflip/react/dist/@flopflip-react.es.js)
(!) Unused external imports
isValidElementType imported from external module 'react-is' but never used
```

I am not sure that can be solved here and I'd rather have to take to flopflip maybe.

tdeekens:
> Cool, now I guess we need to add the following deps create-react-context, lodash.flowright, react-fast-compare, react-is, and lodash.intersection

Maybe I am not 100% sure we have to though to achieve what we want. All those utility libs can be picked up by the application level bundler. It only gets odd again if apps and app-kit depend on a diverging version of create-react-context. However that will also work as the polyfill will not bite.

Given we add them we can read flopflip's package.json. However that truly feels a bit odd. I can't think of a better option however.

montezume:
Yeah, if we aren't going to bundle them then we need to include them in the externals array. 🤔 I would like to avoid any rollup warnings because having benign warnings like this make it really easy to have real dangerous warnings be introduced later.

tdeekens:
> Yeah, if we aren't going to bundle them then we need to include them in the externals array.

No cause they will be resolved by yarn added to the node_modules on a respective level and when webpack runs added to the bundle.

montezume:
@tdeekens Yes, but I would like to avoid rollup warnings.

emmenko:
Turns out we don't need this after all, everything works just fine 😇

Tobi can explain it a bit better

PR #155, opened 2018-11-29
feat(test-utils): wrap rendered component with React.Suspense, include modals container, document renderWithRedux

emmenko:


packages/application-shell/src/test-utils/test-utils.js
```diff
@@ -116,20 +120,28 @@ const render = (
...
+          <ApolloProviderComponent mocks={mocks} addTypename={addTypename}>
+            <ConfigureFlopFlip adapter={adpater} defaultFlags={flags}>
+              <ApplicationContextProvider
+                user={mergedUser}
+                project={mergedProject && { ...mergedProject, permissions }}
+                environment={mergedEnvironment}
+                projectDataLocale={dataLocale}
+              >
+                <Router history={history}>
+                  <React.Suspense fallback={<LoadingFallback />}>
```
emmenko: This is new now, so apps can safely use `React.lazy`

packages/application-shell/src/test-utils/test-utils.js
```diff
@@ -116,20 +120,28 @@ const render = (
...
+                    {ui}
+                  </React.Suspense>
+                </Router>
+              </ApplicationContextProvider>
+            </ConfigureFlopFlip>
+          </ApolloProviderComponent>
+        </IntlProvider>
+        {/* This container is used to render modals, so we need to make it available in tests */}
+        <div id={PORTALS_CONTAINER_ID} />
```
emmenko: And this is also new, which is required to render the modals etc

packages/application-shell/src/components/quick-access/quick-access.spec.js
```diff
@@ -409,7 +408,7 @@ describe('QuickAccess', () => {
     );
 
     // should close quick access
-    expect(container).toBeEmpty();
+    expect(queryByTestId('quick-access')).not.toBeInTheDocument();
   });
 
   it('should cycle through the history', async () => {
```
emmenko: @dferber90 this test apparently is currently failing on `master` (at least locally for me).

image

Can you check if that works for you? In travis it also seem to pass...

emmenko:
@dferber90 in the last commit (https://github.com/commercetools/merchant-center-application-kit/pull/155/commits/389348500d796dab78d70510d83d53e558e75279) I changed the assertion of quick access to check for a non existing element.

PS: we need to use a query function as mentioned in the docs

image

packages/application-shell/src/components/quick-access/quick-access.spec.js
dferber90: Weird that it would fail locally but work on CI. I believe we are no longer able to use `toBeEmpty` since we always add the modal container now.

emmenko: Is it failing locally for you as well?

> I believe we are no longer able to use toBeEmpty since we always add the modal container now

Correct, but I think it's better to not rely on an empty container.

packages/application-shell/src/test-utils/test-utils.js
dferber90: I've been thinking about this some more. I'm not sure our current solution is the best one. It seems a bit weird to always provide the portal container (or maybe I'm just not used to it) - thereby messing with the actual container DOM for every test case. All our other setup functions don't influence the container's DOM tree itself.

The way it's done in `react-testing-library`'s course is that they expose a `Modal` which you can render into your tests: see [example](https://github.com/kentcdodds/react-testing-library-course/blob/master/src/__tests__/portals.js#L11-L13). We could expose that `Modal` component from our test-utils. This is nice because it's declarative and doesn't mess with the test cases in which we don't need a modal.

But maybe it's a better idea to always render the portal container as we're doing now. It's nice that devs don't need to worry about the modal at all if we always put it in place. I'm really not sure, and both approaches obviously work. I'm torn between them. It's up to you which way to go.

Maybe let's proceed with what we have here, and in case it gets annoying/confusing we can try the other approach.

emmenko: I agree with your thoughts in general, however in our case the AppShell always renders it and therefore I think it should be provided by default as well since most of our views use the modal 😊

dferber90: I'm totally aware of that :)

What's giving me discomfort is that we call `render(<Test />)`, but the container will contain not just the output of `Test`. I think this can be quite confusing for first-timers. I dislike this disconnect. But maybe it's the best decision to take. Anyways, let's proceed as-is 🚴

.travis.yml
```diff
@@ -22,10 +22,14 @@ cache:
     - node_modules
 
 before_install:
+  # Install transifex
   - curl https://bootstrap.pypa.io/get-pip.py | sudo -H python3.6
   - (pip -V && pip3 -V && pip3.6 -V) | uniq
   - python3.6 -m pip install --user transifex-client
   - export PATH="$HOME/.local/bin:$PATH"
+  # Install yarn
+  - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.12.3
```
emmenko: Let's use the latest version of yarn to avoid weird CI inconsistencies

emmenko:
I really don't understand what's going on here.

For me this test is failing locally, even on `master`!!

image

On CI on `master` everything is good, however in this branch things are failing with this weird error:

image

What the f*** is going on??! 🤯

emmenko:
> On CI on `master` everything is good, however in this branch things are failing with this weird error:

Ok Dom pointed out that on CI tests fail because we don't allow `console.log` (I forgot about it).
But I don't see any log locally (probably because jest swallows them) 🙃

emmenko:
Alright, CI passes again. Dom says on his laptop tests are also green. So probably it's something with my setup.

We can merge this now then.

PR #159, opened 2018-11-30
chore(http-server): update favicon images for different resolutions

emmenko:
This has been bothering me for a while. I took now the favicons from our docs website.

image


tdeekens:
Will release this with the permissions for you.

emmenko:
I would also like to include #155 if we manage to find out what's wrong with the tests

PR #161, opened 2018-11-30
Remove Quick Access feature toggle

dferber90:
🎉 Now that Quick Access is out, we can remove the feature toggle for it.

### DIFF
--- a/packages/application-shell/src/components/application-shell/application-shell.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.js
@@ -2,11 +2,11 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Router, Redirect, Route, Switch } from 'react-router-dom';
 import { ApolloProvider } from 'react-apollo';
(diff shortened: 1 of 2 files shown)


dferber90:
@tdeekens Yeah, I was thinking that we could also let this PR sit for a week before merging... but on the other hand, it has worked flawlessly so far. Not sure, we can also play it safe.

tdeekens:
I've talked to the REWE team yesterday and they've used it for weeks and like it. It depends You call. If you add the "On hold" label to this we can wait. Otherwise I'll include it in the release later :)

dferber90:
Let's skip this release then just in case and ship it with the next one (as there were some modifications before we released to respect permissions/feature-toggles which haven't been out there as long). I didn't expect us to release so soon again 🚄

dferber90:
I rebased the branch and I'd be ready to merge this if you're on board 🚢 

cc @emmenko

PR #164, opened 2018-11-30
WIP: Add view permissions to constants and add them to menu

tdeekens:
#### Summary

This pull request adds the new permissions (all still defaulting to true) to the permissions constants and integrates them into the menu.

#### Description

We need the `*List` permissions as we want to enable the following use cases

> You can definitely have a customer service representative that takes orders, but shouldn’t be able to access the entire order collection

It (but we do not want to break things) becomes a bit of a mix of permissions. That however to me is fine. It's up to the UI and system we will build if e.g. 

Given you select `viewOrdersList` we actually also save `viewOrders` (as the API scope)


tdeekens:
I want to make sure the v4 release is in and the defaulted permissions are loaded correctly before merging this next week. Also PO should clarify these settings applied here to be correct.

Lastly, APIs have been deployed through to prod to support the new names of permissions.

tdeekens:
I am clarifying the implemented permissions with the product department in JIRA now. Until they're approved from there this is on hold.

tdeekens:
I have doubts this can work as intended. Given [this](https://github.com/commercetools/merchant-center-application-kit/blob/master/packages/application-shell/src/components/navbar/navbar.js#L264) I presume having one permissions which just defaults to `true` will effectively "over-rule" all other permissions are the time. Meaning that nothing can be disabled at any time with this.

tdeekens:
I would like some feedback on three approaches out of this

1. Split permissions passed in into `OAuthScopes` (OR matched) and `viewPermissions` (AND matched)
2. Add a feature toggle to the permission application
   - Needs a follow up migration removing OAuth scope permissions from the view
3. Expand the syntax being passable to `RestrictedByPermissions`

```js
<RestrictedByPermissions
   permissions={{
     and: [[permissions.ViewOrdersList], [permissions.ManageOrders, permissions.A]]
   }}
>

// or with or

<RestrictedByPermissions
   permissions={{
     or: [[permissions.ViewOrdersList], [permissions.ManageOrders, permissions.A]]
   }}
>
```

The API would be

1. Given you pass an array you can pass `shouldMatchAllPermissions`
2. Given you pass an object all outer are joined like `shouldMatchAllPermissions` but the inner are joined by the key

/cc @emmenko @montezume

emmenko:
I'm a bit confused. Can you give a bit more context?

tdeekens:
To my understanding the permissions in the menu are joined in an `or` fashion determined by the `shouldMatchSomePermissions` prop being set/passed to `RestrictedByPermissions` to be `true`. That means given we pass e.g. a `canViewProductsList` (defaulted to `true`) we will loose any ability to hide the according menu item.

If my understanding in that is not mistaken this would break any ability in MC to hide any menu receiving one of the defaulted permissions.

Given that I am trying to find a way out, if I am understanding to current setup correctly. One suggestion is to allow a slightly more sophisticated DSL like suggested.

emmenko:
I see, right it will break it.

Hmm so how should it be eventually? Are the new permissions supposed to take over the API scopes?

tdeekens:
Eventually possibly but I don't think we have enough use cases to really 100% ensure that this would be possible. Given also that they will co-exist until we get them into the Core WS we might have to find a way for them to live side by side.

Hence my suggestion. Do you see another way or any specific reason against the DSL enhancement?

tdeekens:
I think instead of having this complexity in the mc-fe or having `$and` and `$or` we could map in the mc-be. So we map from OAuth scopes to the new permissions`

Given you have manageProject you get: `viewProjectsList`, `addProduct`
Given you have viewOrders you get: `viewOrdersList` but not `addOrder`

This means we need a static mapping of OAuth scope permissions to these new permissions in the mc-be in case in team permissions are not present yet.

tdeekens:
Will close this as from my side this is neither actionable nor seems the name of actions and visibility settings defined (which can be problematic).

PR #166, opened 2018-11-30
fix(l10n): prefer loading default, fall back to normal import (esm vs cjs)

emmenko:
Caused by #154 

## Problem

This is what l10n loads

image

To explain it a bit further: the bundled JSONs look like this

image

The `default` export includes all the locales (languages + country)

image

But the named exports are only for "top-level" locales (language tag).

The CJS includes all locales in one object

image

## Solution

We check first if the loaded module has a `default` property, in which case we pick that (ESM bundle). If not, we simply load the module as-is (CJS bundle).

### DIFF
--- a/packages/i18n/load-i18n.js
+++ b/packages/i18n/load-i18n.js
@@ -46,7 +46,12 @@ export default function loadI18n(lang) {
   const localeDataPromises = [reactIntlChunkImport, momentChunkImport];
 
   return Promise.all(localeDataPromises).then(response => {
-    addLocaleData([...response[0].default]);
-    return getLocalizedStringsChunkImport(lang).then(result => result.default);
+    // Prefer loading `default` (for ESM bundles) and
+    // fall back to normal import (for CJS bundles).
+    const loadedData = response[0].default || response[0];
+    addLocaleData([...loadedData]);
+    return getLocalizedStringsChunkImport(lang).then(
+      result => result.default || result
+    );
   });
 }
--- a/packages/l10n/country-information.js
+++ b/packages/l10n/country-information.js
@@ -27,7 +27,9 @@ const getCountriesForLocale = (locale, cb) => {
   // The files are named like "country-data-en-json.chunk.js" after compilation
   // https://webpack.js.org/api/module-methods/#import-
   getImportChunk(supportedLocale)
-    .then(countries => cb(null, countries))
+    // Prefer loading `default` (for ESM bundles) and
+    // fall back to normal import (for CJS bundles).
+    .then(countries => cb(null, countries.default || countries))
     .catch(error => cb(error));
 };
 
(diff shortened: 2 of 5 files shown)


emmenko:
I think the "problem" relies here: https://github.com/rollup/rollup-pluginutils/blob/fe8b70a981114becb856dfbc0f1099b6a255ce17/src/dataToEsm.js#L58

```js
// this check
key === makeLegalIdentifier(key)

// if key is e.g. `en-US`, the check fails because `makeLegalIdentifier` will strip the `-` resulting in `enUS`
```

It also makes sense because variable names cannot contain hyphens, so having `en-US` as a named export will never work.

emmenko:
@montezume was there any particular reason for using the `namedExports` option? Otherwise we can simply set it to `false` as it might only bring confusion...

WDYT?

montezume:
@emmenko yeah might make sense to set it to false as well 👍 
Thanks for the info about the invalid exports

packages/i18n/load-i18n.js
```diff
@@ -46,7 +46,12 @@ export default function loadI18n(lang) {
   const localeDataPromises = [reactIntlChunkImport, momentChunkImport];
 
   return Promise.all(localeDataPromises).then(response => {
-    addLocaleData([...response[0].default]);
-    return getLocalizedStringsChunkImport(lang).then(result => result.default);
+    // Prefer loading `default` (for ESM bundles) and
+    // fall back to normal import (for CJS bundles).
+    const loadedData = response[0].default || response[0];
```
tdeekens: The module format sniff.

packages/i18n/load-i18n.js
```diff
@@ -46,7 +46,12 @@ export default function loadI18n(lang) {
...
+    addLocaleData([...loadedData]);
```
tdeekens: Why do we spread and not just pass loadedData?

tdeekens:
Disabling it will also reduce bundle size right? If my brain is not frozen 😀

packages/i18n/load-i18n.js
emmenko: I think it was copy-pasted from the l10n implementation where we spread more than one. We can make it simpler sure 😊

emmenko:
I guess the difference is very minimal, but yeah 😅

emmenko:
Ok I guess we can switch the option to be `false`

tdeekens:
I am obviously fine with it either way but given we are here and likely wont come back here for quite some time.

packages/i18n/load-i18n.js
```diff
@@ -26,11 +26,11 @@ const getMomentChunkImport = momentLocaleCode => {
 const getLocalizedStringsChunkImport = lang => {
   switch (lang) {
     case 'de':
-      return import(/* webpackChunkName: "react-intl-localized-strings-de" */ './data/de.json');
+      return import(/* webpackChunkName: "i18n-app-kit-locale-de" */ './data/de.json');
```
montezume: good name 👍

PR #175, opened 2018-12-03
fix(test-utils): improve test setup

emmenko:
While I was implementing tests for an application, I encountered some problems and missing things that can be provided by the `test-utils` out of the box.

I recommend a review-per-commit.

### DIFF
--- a/packages/application-shell/src/from-core/modal-container/modal-container.js
+++ b/packages/application-shell/src/from-core/modal-container/modal-container.js
@@ -16,13 +16,13 @@ import { PORTALS_CONTAINER_ID } from '@commercetools-frontend/constants';
 // instead of a specific element that will be cleaned up, resulting in
 // console errors (even though the test passes). We only need to to this in
 // test environment.
-const getParentSelector = () =>
+const parentSelector = () =>
   process.env.NODE_ENV === 'test'
     ? document.body
     : document.querySelector(`#${PORTALS_CONTAINER_ID}`);
 
 const ModalContainer = props => (
-  <Modal {...props} parentSelector={getParentSelector} ariaHideApp={false}>
+  <Modal {...props} parentSelector={parentSelector} ariaHideApp={false}>
     {props.children}
   </Modal>
 );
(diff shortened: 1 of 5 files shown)


packages/application-shell/src/from-core/modal-container/modal-container.js
```diff
@@ -3,12 +3,26 @@ import React from 'react';
...
+// element, used to render the modal content.
+// Apparently though, when the <Modal> component unmounts between tests, it
+// tries to remove the child elements within the parent selector
+// ("portals-container") which may have been cleaned up by another test,
+// resulting in the <Modal> to throw an NotFoundError.
+// To solve this, we can simply use "document.body" as the parent selector
+// instead of a specific element that will be cleaned up, resulting in
+// console errors (even though the test passes). We only need to to this in
+// test environment.
+const getParentSelector = () =>
```
dferber90: I would call this function `parentSelector` as `parentSelector={parentSelector}` makes more sense to me than `parentSelector={getParentSelector}`.

packages/application-shell/src/test-utils/test-utils.js
```diff
@@ -160,29 +160,33 @@ const render = (
...
   {
-    // Consumers of renderWithRedux can use
-    //   { store: createReduxStore({ requestsInFlight: null, .. }) }
-    // to pass an initial state to Redux.
-    store = createReduxStore(),
+    // Pass an initial state that will be used to rehydrate the redux store.
+    // Possible state values are:
+    // - requestsInFlight: []
+    // - notifications: []
+    preloadedState,
```
dferber90: With this change, we gain a simpler API to write a test, but we lose the ability to pass in an arbitrary store. Which means we can't replace the middleware for our tests anymore. I'm not sure whether we should already limit the API that much. Did we learn that nobody ever needs to pass in a configured store?

emmenko: I thought about it but I don't know when we would be needing to replace the `store`. But I can revert this change if you prefer so.

dferber90: I generally like to keep the API open until the use-cases are clear. Then we can restrict the API to force people in the right direction. I can't make a good call here at the moment.

Are all stores configured the same for all applications? If yes, then it makes sense to configure the store in a similar way for our tests and we might not need to enable injecting middleware. It's still an assumption though.

I'd keep it open for now until we've seen more cases using it. If they all end up being the same (using preloaded state only), we can limit the API.

emmenko: I reverted the change

PR #177, opened 2018-12-03
chore(playground): remove ui-kit materials transpilation

montezume:
#### Summary

Since `ui-kit` v4.0.0, we no longer access material variables by @importing them in our css files. So we no longer need to pass this folder to our webpack configs.

#### Approach

Stop passing `ui-kit` materials folder to webpack configs in playground.


montezume:
sorry @importing 😆

PR #182, opened 2018-12-05
MCD-1509(user-settings-menu): address UX review

qmateub:
Only two points for the UX review.

No hiding sections in the menu when in the account part.

Change item text to "Organizations & Teams"


packages/application-shell/src/components/user-settings-menu/user-settings-menu.js
```diff
@@ -92,76 +91,55 @@ export class UserSettingsMenu extends React.PureComponent {
...
-                              {[this.props.firstName, this.props.lastName]
-                                .join(' ')
-                                .trim()}
-                            </Text.Body>
-                            <Text.Body truncate>{this.props.email}</Text.Body>
-                          </div>
-                        </Spacings.Inline>
-                      </Spacings.Inset>
-                    </div>
-                    {!this.props.isAccountPath && (
```
qmateub: This is finally gone @emmenko 🎉

PR #183, opened 2018-12-05
MCD-1523 : Refactor CT logo to be 🏠button

qmateub:
#### Summary

Ticket: https://jira.commercetools.com/browse/MCD-1523

#### Description

UX really wants that our CT logo becomes the home button for the MC so does not matter in where part you at if you click the button you will get redirected to the dashbard 👍 

I've also added a new shape for the `LoadingPlaceholder` called `square` for the logo. The sizes are quite similar to the `img` tag ones so they both fit 😋

### DIFF
--- a/packages/application-shell/src/components/app-bar/app-bar.js
+++ b/packages/application-shell/src/components/app-bar/app-bar.js
@@ -19,7 +19,7 @@ const AppBar = props => {
       <Spacings.Inline>
         <div className={styles.logo}>
           {!props.user ? (
-            <LoadingPlaceholder shape="square" size="l" />
+            <img src={LogoSVG} className={styles['logo-img']} alt="Logo" />
           ) : (
             <Link
               to={`/${previouslyUsedProjectKey ||
--- a/packages/application-shell/src/components/app-bar/app-bar.js
+++ b/packages/application-shell/src/components/app-bar/app-bar.js
@@ -23,7 +23,7 @@ const AppBar = props => {
           ) : (
             <Link
               to={`/${previouslyUsedProjectKey ||
-                props.user.defaultProjectKey}/dashboard`}
+                props.user.defaultProjectKey}`}
             >
               <img src={LogoSVG} className={styles['logo-img']} alt="Logo" />
             </Link>
--- a/packages/application-shell/src/components/app-bar/app-bar.js
+++ b/packages/application-shell/src/components/app-bar/app-bar.js
@@ -12,79 +12,93 @@ import { REQUESTS_IN_FLIGHT_LOADER_DOM_ID } from '../requests-in-flight-loader/c
 import messages from './messages';
 import styles from './app-bar.mod.css';
 
-const AppBar = props => (
-  <div className={styles['app-bar']} data-test="top-navigation">
-    <Spacings.Inline>
-      <div className={styles.logo}>
-        <img src={LogoSVG} className={styles['logo-img']} alt="Logo" />
-      </div>
-
-      <div
-        id={REQUESTS_IN_FLIGHT_LOADER_DOM_ID}
-        className={styles['loader-container']}
-      />
-    </Spacings.Inline>
-
-    <div className={styles.navigation}>
-      <Spacings.Inline scale="m" alignItems="center">
-        <div className={styles.switchers}>
-          <Spacings.Inline alignItems="center">
-            {/* This node is used by a react portal */}
-            <div id="locale-switcher" />
-            {do {
-              const previouslyUsedProjectKey = selectProjectKeyFromLocalStorage();
-              if (!props.user) {
-                <LoadingPlaceholder shape="rect" size="s" />;
-              }
-              // The `<ProjectSwitcher>` should be rendered only if the
-              // user is fetched and the user has projects while the app runs in an project context.
-              else if (props.user.projects.total > 0 && props.projectKeyFromUrl)
-                <ProjectSwitcher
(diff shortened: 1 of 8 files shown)


packages/application-shell/src/components/app-bar/app-bar.js
```diff
@@ -12,79 +12,93 @@ import { REQUESTS_IN_FLIGHT_LOADER_DOM_ID } from '../requests-in-flight-loader/c
...
+  return (
+    <div className={styles['app-bar']} data-test="top-navigation">
+      <Spacings.Inline>
+        <div className={styles.logo}>
+          {!props.user ? (
+            <LoadingPlaceholder shape="square" size="l" />
+          ) : (
+            <Link
+              to={`/${previouslyUsedProjectKey ||
+                props.user.defaultProjectKey}/dashboard`}
```
qmateub: That's why I need the LoadingPlaceholder now, because we need info from the user which can be loading state 👍

packages/application-shell/src/components/app-bar/app-bar.js
```diff
@@ -12,79 +12,93 @@ import { REQUESTS_IN_FLIGHT_LOADER_DOM_ID } from '../requests-in-flight-loader/c
...
+      </Spacings.Inline>
 
-    <div className={styles.navigation}>
-      <Spacings.Inline scale="m" alignItems="center">
-        <div className={styles.switchers}>
-          <Spacings.Inline alignItems="center">
-            {/* This node is used by a react portal */}
-            <div id="locale-switcher" />
-            {do {
-              const previouslyUsedProjectKey = selectProjectKeyFromLocalStorage();
```
qmateub: I just pull this up since it's reused now

packages/application-shell/src/components/app-bar/app-bar.js
emmenko: Omit the `dashboard`, the redirect happens automatically then

packages/application-shell/src/components/loading-placeholder/loading-placeholder.mod.css
```diff
@@ -8,6 +8,10 @@
   --rect-width-l: 450px;
   --rect-width-xl: 600px;
   --rect-height: 32px;
+  --square-size-s: 8px;
+  --square-size-m: 16px;
+  --square-size-l: 32px;
+  --square-size-xl: 64px;
```
qmateub: Those are the sizes for the new shape 👍

packages/application-shell/src/components/app-bar/app-bar.js
tdeekens: Can we not redirect to the dashboard but only the root?

packages/application-shell/src/components/app-bar/app-bar.js
```diff
@@ -12,79 +12,93 @@ import { REQUESTS_IN_FLIGHT_LOADER_DOM_ID } from '../requests-in-flight-loader/c
...
-        <img src={LogoSVG} className={styles['logo-img']} alt="Logo" />
-      </div>
+const AppBar = props => {
+  const previouslyUsedProjectKey = selectProjectKeyFromLocalStorage();
+  return (
+    <div className={styles['app-bar']} data-test="top-navigation">
+      <Spacings.Inline>
+        <div className={styles.logo}>
+          {!props.user ? (
+            <LoadingPlaceholder shape="square" size="l" />
```
emmenko: Can't we always render the logo? I find it weird to have a placeholder for the logo

packages/application-shell/src/components/app-bar/app-bar.js
tdeekens: Yup. What I said above @qmateub :)

packages/application-shell/src/components/app-bar/app-bar.js
montezume: Yeah I also find this weird

qmateub: Yeah but did you guys read my comment here https://github.com/commercetools/merchant-center-application-kit/pull/183#discussion_r239092501 ?

We are using user information which can be loading, that's why this rendering case 🤔

emmenko: The "loading" is just for the link, not for the logo. Just render the image twice and we're fine. No loading spinner

qmateub: So true 🤦‍♂️

qmateub: Any objections on keeping the `square` shape @emmenko or should I remove it?

qmateub:
Thanks for the quick feedback guys @tdeekens @montezume @emmenko 🙏 

I've removed the new shape and the usage of the loading placeholder by just rendering the image 👍

packages/i18n/data/en.json
```diff
@@ -156,7 +156,7 @@
   "QuickAccess.useProject": "Switch to project \"{projectName}\"",
   "TopNavigation.labelLoading": "Processing...",
   "UserSettingsMenu.logout": "Logout",
-  "UserSettingsMenu.manageOrganizations": "Manage Organizations",
+  "UserSettingsMenu.manageOrganizations": "",
```
montezume: how come this is empty?

qmateub: Because you the one in core is the correct one 👍

PR #185, opened 2018-12-06
fix(app-shell): show <ProjectExpired> also when daysLeft are less than 0

emmenko:
I noticed that projects with a trial expired based on `daysLeft` should be shown the `<ProjectExpired>` page

### DIFF
--- a/packages/application-shell/src/components/project-container/project-container.js
+++ b/packages/application-shell/src/components/project-container/project-container.js
@@ -132,7 +132,10 @@ export class ProjectContainer extends React.Component {
                   }
                 />
               );
-            if (project.expiry && project.expiry.isActive)
+            if (
+              project.expiry &&
+              (project.expiry.isActive || project.expiry.daysLeft < 0)
+            )
               return <ProjectExpired />;
 
             return (
--- a/packages/application-shell/src/components/project-switcher/project-switcher.js
+++ b/packages/application-shell/src/components/project-switcher/project-switcher.js
@@ -135,12 +135,14 @@ export class ProjectSwitcher extends React.PureComponent {
         className={classnames(styles['item-text-main'], {
           [styles['item-text-disabled']]:
             (project.suspension && project.suspension.isActive) ||
-            (project.expiry && project.expiry.isActive),
+            (project.expiry &&
+              (project.expiry.isActive || project.expiry.daysLeft < 0)),
         })}
       >
         {project.name}
         {((project.suspension && project.suspension.isActive) ||
-          (project.expiry && project.expiry.isActive)) && (
+          (project.expiry &&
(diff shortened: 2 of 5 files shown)


packages/application-shell/src/components/project-container/project-container.js
```diff
@@ -132,7 +132,10 @@ export class ProjectContainer extends React.Component {
                   }
                 />
               );
-            if (project.expiry && project.expiry.isActive)
+            if (
+              project.expiry &&
```
tdeekens: I am annoying I know but I write it as a suggestion

```suggestion
project.expiry?.isActive && project.expiry?.daysLeft < 0
```

Didn't even know that `daysLeft` goes into negative numbers.

emmenko: > I am annoying I know but I write it as a suggestion

I thought we didn't want to use that in app-kit / ui-kit

> Didn't even know that daysLeft goes into negative numbers.

Something that Luis did.

emmenko:
I found also other places where it was broken. For example, now the project listed in the dropdown have are correctly disabled

image

packages/application-shell/src/components/project-container/project-container.js
tdeekens: Oh okay. I remember vaguely.

packages/application-shell/src/components/project-switcher/project-switcher.js
```diff
@@ -135,12 +135,14 @@ export class ProjectSwitcher extends React.PureComponent {
...
-            (project.expiry && project.expiry.isActive),
+            (project.expiry &&
+              (project.expiry.isActive || project.expiry.daysLeft < 0)),
         })}
       >
         {project.name}
         {((project.suspension && project.suspension.isActive) ||
-          (project.expiry && project.expiry.isActive)) && (
+          (project.expiry &&
+            (project.expiry.isActive || project.expiry.daysLeft < 0))) && (
```
tdeekens: Given this spreads so much it might have been a good idea to have `expiry.hasPassed` but I don't mind. Just a bit odd with the negative number.

tdeekens:
Before expired was just the `Boolean` of `project.expired` now `isActive` is. So the semantics didn't change just moved. Maybe changing `isActive` to not only be `isActive: Boolean(obj.expired)` but `isActive: Boolean(obj.expired) && daysLeft < 0` would simplify things? Not sure what that change in semantics means though.

emmenko:
You mean in graphql?

tdeekens:
Yeah in out resolvers.

emmenko:
Good idea, maybe we don’t even need this PR then. I’ll look into it

emmenko:
So the fix is now implemented in the graphql API, we don't need this anymore.

PR #192, opened 2018-12-10
Update dependencies

tdeekens:
#### Summary

- Closes #191
- Closes #190
- Closes #189
- Closes #188
- Closes #187
- Closes #186


package.json
```diff
@@ -38,30 +38,30 @@
...
     "cross-env": "5.2.0",
-    "dotenv": "6.1.0",
+    "dotenv": "6.2.0",
     "enzyme": "3.7.0",
-    "eslint": "5.9.0",
+    "eslint": "5.10.0",
     "husky": "1.2.0",
     "jest": "23.6.0",
-    "jest-dom": "2.1.1",
+    "jest-dom": "3.0.0",
```
tdeekens: Fix toHaveTextContent to be case-sensitive

packages/application-shell/package.json
```diff
@@ -88,7 +88,7 @@
     "query-string": "5.1.0",
     "raven-js": "3.27.0",
     "react-modal": "3.6.1",
-    "react-redux": "5.1.1",
+    "react-redux": "6.0.0",
```
tdeekens: Removed legacy context. As far as I recall from Twitter.

packages/react-notifications/package.json
```diff
@@ -44,7 +44,7 @@
     "moment": "2.22.2",
     "moment-timezone": "0.5.23",
     "prop-types": "15.6.2",
-    "react-redux": "5.1.1",
+    "react-redux": "6.0.0",
```
tdeekens: Given this we should/have to require also to update in the mc-fe as I would assume the new/old context otherwise would not pair too well together?

montezume: I think it would make more sense to get the update working to 6.0.0 in MC-fe before bumping this here?

tdeekens: Is that even possible? The store is defined here, right? Maybe I am failing to grasp something.

montezume: What I am concerned about is the behaviour changes around dispatching actions in constructors / componentWillMount.
https://github.com/reduxjs/react-redux/releases

emmenko: It says it’s actually an improvement because now reads are consistent. I don’t think we should have a problem with that

montezume: Yeah, but are we sure that but we aren't relying on the existing behavior? 🤔

tdeekens: Aren‘t we mixing topics again. 

1. Behavior might have changed but we are unsure if affected
2. Path to upgrade. First here or first mc-fe

To understanding 2. should be here first and then mc-fe as the consumer. Otherwise my understanding would we that we would define stores here on legacy context and consume on the updated context. That sounds wrong. Is that understanding false?
To my understaning for 1. we can not definately tell, correct. Is there a path we see to get more certainty?

montezume: In terms of merged PRs, the path to upgrade should happen as you described, here first and then in MC-FE.

In terms of testing this PR, I think it would be worthwhile to do the old copy and paste trick, and upgrade MC-FE redux to 6.0.0, and make sure everything seems fine?

tdeekens: Sure. Seems reasonable.

tdeekens: So I tried to get it to work and failed. I would recommend us taking a `react-redux` update separately from any other dep for now. I removed the update in the last commit. Would the PR be fine as it stands then? /cc @montezume and @qmateub

emmenko: Could you mention what failed exactly and what problems did you encounter?

tdeekens: I would like to take that discussion elsewhere on an issue and move on with this. It was basically something that store was not configured correctly.

emmenko: Alright thanks

emmenko: The upgrade to v6 is now taken care in here: #197

PR #194, opened 2018-12-12
Expose new permissions

qmateub:
#### Summary

This is related to https://github.com/commercetools/merchant-center-application-kit/pull/164

Ticket: https://jira.commercetools.com/browse/MCD-1516

#### Description

So we want to starting working on hide/show action button depending on the new permissions. As Tobi  and me discussed this is kind of a bet since this is not 100% definitive but we definitively need something to start with.

I just picked the part from Tobi's PR where we are exposing the new permissions 👍


packages/application-shell/src/components/fetch-project/fetch-project.graphql
```diff
@@ -31,8 +31,12 @@ query ProjectQuery($projectKey: String!) {
       canAddCustomers
       canAddCustomerGroups
       canAddDiscountCodes
+      canAddProductDiscounts
+      canAddCartDiscounts
```
qmateub: So apparently those two are wrong since they are not in our BE schema 
image

cc @tdeekens

qmateub: Is this a leftlover? Should I create quick PR in the BE for adding them? 🤔

tdeekens: Feel free to yes.

qmateub:
> Please align with Ida. This is really hard to change after we start using it.

Yep, not going to go ahead with this without checking it with her 👍

tdeekens:
This is not mergeable as it stands. The mc-be needs to expose those permissions first and needs to be deployed across envs. Just making sure this doesn't merge-slip in.

PR #195, opened 2018-12-12
Add core-ws graphql target to constants

tdeekens:
#### Summary

A new GraphQL target is there. We need to proxy to it.

### DIFF
--- a/packages/constants/index.js
+++ b/packages/constants/index.js
@@ -35,5 +35,5 @@ export const GRAPHQL_TARGETS = {
   DASHBOARD_SERVICE: 'dashboard',
   PIM_INDEXER: 'pim-indexer',
   SETTINGS_SERVICE: 'settings',
-  CORE_WS: 'core-ws',
+  ADMINISTRATION_SERVICE: 'administration',
 };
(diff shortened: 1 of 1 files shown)


packages/constants/index.js
```diff
@@ -35,4 +35,5 @@ export const GRAPHQL_TARGETS = {
   DASHBOARD_SERVICE: 'dashboard',
   PIM_INDEXER: 'pim-indexer',
   SETTINGS_SERVICE: 'settings',
+  CORE_WS: 'core-ws',
```
tdeekens: Thought about the name literally ten minutes. `core` just sounds really silly. We can also name it something completely different but I don't know how much sense that makes.

qmateub: Can't think on another name since this is the one that I've been reading this past weeks. So to me it's fine 😅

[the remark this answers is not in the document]
tdeekens: It's not only users. 

Maybe `Administration` or `Account`?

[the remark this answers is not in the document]
tdeekens: How about: a2a6026.

packages/constants/index.js
```diff
@@ -35,4 +35,5 @@ export const GRAPHQL_TARGETS = {
   DASHBOARD_SERVICE: 'dashboard',
   PIM_INDEXER: 'pim-indexer',
   SETTINGS_SERVICE: 'settings',
+  ADMINISTRATION_SERVICE: 'core-ws',
```
emmenko: You mean `ADMINISTRATION_SERVICE: 'administration'` ?

tdeekens: Et voila.

tdeekens:
Sure thing! Thanks for being persistent on the review!

Issue #68, opened 2018-11-08
Add notifications to ApplicationState


dferber90:
Now that we have the beautiful `ApplicationStateProvider` which serves as the-one-true-intersection-point between applications and and the AppShell, I think it might be a good idea to enhance it even more.

One of the things which is currently still inconvenient is showing notifications in the MC. I think we could move this to the `ApplicationStateProvider` as well.

Right now dispatching notifications works like this

```jsx
import { connect } from 'react-redux';
import * as globalActions from '@commercetools-frontend/actions-global';

// your component
const SomeComponent = props => (
  <button
    onClick={() =>
      props.showNotification({
        kind: "success",
        domain: DOMAINS.SIDE,
        text: "hello"
      })
    }
  >
    Show Notification
  </button>
);

export default connect(
  null,
  { showNotification: globalActions.showNotification }
)(SomeComponent);
```

However this could get as simple as

```jsx
const SomeComponent = props => (
  <GetApplicationState
    render={applicationState => (
      <button
        onClick={() =>
          applicationState.showNotification({
            kind: "success",
            domain: DOMAINS.SIDE,
            text: props.intl.formatMessage(messages.lineItemUpdated)
          })
        }
      >
        Show Notification
      </button>
    )}
  />
);
```

This would allow us to also test notifications dispatched by components more easily.

```jsx
import SomeComponent from './some-component'
import { render, fireEvent } from '@commercetools-frontend/application-kit/test-utils'

it("should dispatch a notification when clicked", () => {
  const { getByText, notifications } = render(<SomeComponent />);
  fireEvent.click(getByText("Show Notification"));
  expect(notifications).toContainEqual({
    kind: "success",
    domain: DOMAINS.SIDE,
    text: "hello"
  });
});
```

dferber90:
One thing which is kinda weird is that the `applicationState` now isn't just *state*  anymore, but it evolves to more than that. It is now the intersection of the application and the application shell. It provides the API to exchange information between the two. It is no longer just a getter, but also a dispatcher/setter.

Given the new capabilities it is weird to call `applicationState.showNotification`, because a state is usually something which gets read only. It is also a bit weird that the component enabling this is called `GetApplicationState` when we are not really getting a state, but we only want to use it to emit an event (show a notification).

Should we maybe rename `applicationState` and `GetApplicationState`?

I can see more usage-as-setters coming in the future:
- setting the `dataLocale`?
- dealing with *view settings* (not sure about this one)?

[...]

PR #197, opened 2018-12-12
Upgrade to react-redux v6, drop legacy plugin stuff 🔥

### LINKED ISSUE #193: Remove/deprecate local store provider (legacy plugins)
Now that we have migrated the apps to not use `globalAppState`, we can finally get rid of the `LocalStoreProvider`. With that, there are also other things that we can/should remove that are not necessary anymore with the current structure (e.g. plugin concept, reducers, local middlewares).

I'll be start working on that this week.

emmenko:
Closes #193

## Summary (BREAKING CHANGES)

- upgrade to [`react-redux@6`](https://github.com/reduxjs/react-redux/releases/tag/v6.0.0)
  - uses new `React.Context` API
  - ...and because of that it can now support nested providers!! 🎉
- drop legacy plugin stuff
  - `<LocalStoreProvider>`
  - `<InjectReducer>` (_changed a bit to `<InjectReducers>`, see below_)
  - a bunch of middlewares to scope the actions to local/global
- use `lodash-es` and update Rollup config (source: https://github.com/jaredpalmer/formik/commit/3588f4ae641b4a794d6e7d071ab81a2b86efc537)

With the new react-redux API, it's now possible to use nested Providers. What we can do is to have a main `<Provider>` within the AppShell, which contains the "internal store" for things like notifications, sdk, etc).
Each application can then define its own `<Provider>` with its own `React.Context` and `store`. At the end, it's just a matter of **connecting to the correct store**.

~~Have a look at the playground changes to see how the setup got simpler.~~

Update: for now we don't promote using nested providers (see 
https://github.com/commercetools/merchant-center-application-kit/pull/197#discussion_r244685548).

## Upgrade migration

- In the application `<EntryPoint>` the `setupGlobalErrorListener` can be called without arguments (the store is implicitly imported)
- In the application `<EntryPoint>` the `<ApplicationShell>` does not need to be wrapped with the redux `<Provider>` anymore (the provider is defined within the AppShell)
- In the application `<EntryPoint>` the `onRegisterErrorListeners` now accepts a named argument `dispatch`
- the `reduxStore` is no longer exported from the AppShell
- applications that use the `<InjectReducer>` need to migrate to use `<InjectReducers>`. This is similar to the previous `<InjectReducers>` but the API is a bit different
  - the component is now called `<InjectReducers>` instead of `<InjectReducer>` (plural vs singular). The reason being a different API and because we now pass an **object of reducers**, instead of **combining the reducers** first and pass the final reducer
  - the passed **reducers object** is now namespaced by `id` in the `store.injectedReducers` (for checking purposes) but is spread when creating the actual store reducer. This is necessary because now we don't have the "LocalStoreProvider" that maps the namespaced store to the given reducer name. It also allows to inject reducers multiple times (the user still needs to be careful with naming conflicts).
```js
// before
const reducer = combineReducers({ /* app reducers */ });
<InjectReducer name="application-products" reducer={reducer} />

// after
const reducers = { /* app reducers */ };
<InjectReducers id="application-products" reducers={reducers} />
```
- `react-redux` and `redux` are now peer dependencies and need to be explicitly installed from each application
- previously we were passing some tracking stuff into a `mcng` context. Now this uses the new Context API
```js
import { GtmContext } from '@commercetools-frontend/application-shell';

class MyComponent extends React.Component {
  static contextType = GtmContext;

  // this.context.track
  // this.context.getHierarchy
}
```


packages/application-shell/src/components/gtm-booter/gtm-booter.js
```diff
@@ -3,9 +3,9 @@ import PropTypes from 'prop-types';
 import * as gtm from '../../utils/gtm';
 import defaultTrackingEventWhitelist from '../../tracking-whitelist';
 
-const mcngShape = PropTypes.shape({
-  track: PropTypes.func.isRequired,
-  getHierarchy: PropTypes.func.isRequired,
+export const GtmContext = React.createContext({
```
dferber90: This was originally planned to contain "everything" an application might need. Now that we established `ApplicationContext` we don't need this anymore and the renaming is the right thing to do 👍 

Nice to see that we finally ended up where we planned to go over a year a go 🌈

emmenko:
I'll enhance the playground a bit to play around with the redux stores, so we can test things out.

Thanks for the early feedback, happy to see that we all agree with the direction we're going here 🎉

playground/src/components/state-machines-list/state-machines-list.spec.js
```diff
@@ -0,0 +1,84 @@
...
+          result: {
+            total: 2,
+            count: 2,
+            results: [
+              { key: 'sm1', name: { en: 'SM 1' } },
+              { name: { de: 'SM 2' } },
+            ],
+          },
+        });
+    });
```
tdeekens: We should be able to use the new native `renderProp` which has a slightly different syntax in `renderProp(...props)()`. Maybe worth trying to use so we don't have to migrate this later. I plan to do so on my train ride home.

emmenko:
@dferber90 @tdeekens would you mind doing another round of review?
I've refactored the playground app to make use of the "nested" redux store and show how to use it.

I've update the PR description to include the latest changes.

Let me know if you have any questions or if something is not clear.

emmenko:
Update: there are some weird problems with the `QuickAccess` tests related to the sdk mocks and _maybe_ the new redux version.

@dferber90 and I are investigating this...

rollup.config.js
```diff
@@ -24,7 +24,7 @@ const postcssReporter = require('postcss-reporter');
 const { pkg } = readPkgUp.sync({
   cwd: fs.realpathSync(process.cwd()),
 });
-
+const importReplacements = [{ original: 'lodash', replacement: 'lodash-es' }];
```
tdeekens: Is this something we want todo in the app too?

examples/starter/src/store.js
```diff
@@ -0,0 +1,7 @@
+/* eslint-disable import/prefer-default-export */
+import React from 'react';
+
+// Usage:
+//   <Provider store context={ApplicationStoreContext} />
+//   <ConnectedComponent context={ApplicationStoreContext} />
+export const ApplicationStoreContext = React.createContext(null);
```
tdeekens: I guess this is now "clashing" with the create-mc-app (and hence CT docs)? I am getting a worried by us starting dependent changes in parallel without finishing the prior.

emmenko:
1. Yeah sorry about that, it came up during development, as I found a PR (which I linked somewhere) from Formik from one of the contributors of rollup. I can open a separate PR if you’d like me to

2. We could have 2 consecutive breaking change release but I’m not sure it brings much value. The upgrade path here is clear and does not require too much work. Moreover, the “nested” provider is not strictly necessary, it’s more of opt-in in case apps want to separate their store from the AppShell

3. Yes I’m aware of that, I already planned to do it. That’s also why I’d like to get this released soon so that when most people come back from the holidays, they’ll find the latest version. This of course includes the new create-mc-app package.

examples/starter/src/store.js
emmenko: I haven’t updated this branch after merging create-mc-app. Once I do, changes are properly reflected there as well

rollup.config.js
emmenko: I don’t see why not 😉

tdeekens:
Okay. 

Regarding 1. I don’t mind. Now that it is here we can keep it here. Still, would be good in the future to not mix thing to keep reviews easy and also the release notes pointing to one PR per topic. 
For 2. and 3. we then release this soon and just migrate all apps for teams? I fear coordination taking more time than carrying out the work? Also cause we have the UIKit breaking release already (maybe we time such things better in the future also due zo potential regressions). Maybe aim end of next week if this passes review early? Is that realistic (I am back Thursday)?

For the lodash work on the MC: lets just open an issue to keep track of it also across repos.

emmenko:
Sure, I’m back on the 2nd. I think the uikit breaking release can be kept independent from this, there should not be any conflicts.

Let’s coordinate next week anyway 😉 Thanks for the review!!

packages/application-shell/src/components/gtm-booter/gtm-booter.js
```diff
@@ -19,20 +19,6 @@ class GtmBooter extends React.Component {
...
-    mcng: mcngShape.isRequired,
-  };
-  getChildContext() {
-    return {
-      mcng: {
-        track: gtm.track,
-        getHierarchy: gtm.getHierarchy,
-      },
-    };
-  }
```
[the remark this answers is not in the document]
emmenko: No it's not directly related. It's not much to be refactored, I'll take care of that.

playground/src/components/state-machines-list-connector/state-machines-list-connector.js
```diff
@@ -1,19 +1,25 @@
...
 import PropTypes from 'prop-types';
-import { Sdk } from '@commercetools-frontend/sdk';
+import { connect } from 'react-redux';
+import { createSdkGet } from '@commercetools-frontend/sdk';
+import { ApplicationStoreContext } from '../../store';
 import * as actions from './actions';
 
-class ChannelsListConnector extends React.Component {
-  static displayName = 'ChannelsListConnector';
+const SdkGet = createSdkGet({ context: ApplicationStoreContext });
```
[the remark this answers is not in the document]
emmenko: Thanks for the feedback! Yes, having some abstractions might be useful, we just need to find the more suitable one. For now I would not add any abstraction and try to get a feeling on how things can be improved.
Using a nested store is not strictly necessary, if you don't provide one, the default one defined within the AppShell will be used. Therefore, people can start "adopting" the nested provider whenever and however they like. If we then add an abstraction, it should be easy to "upgrade" as it's something opt-in.

---

As for the suggestions you made, I think having static methods would not really work, because we need to differentiate between context of AppShell and nested context.

However, maybe we can invert the contexts, so the AppShell uses a custom one and apps can use the default redux store context. Let me look into that...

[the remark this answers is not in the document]
emmenko: > However, maybe we can invert the contexts, so the AppShell uses a custom one and apps can use the default redux store context. Let me look into that...

This won't work, mostly because of the notifications and the action creators. If we were to put that into a custom store, each "bound" notification action creator would need to connect to the specific store context, which would make the upgrade a huge pain in the ass.

On a related note, if dispatching notifications is much more straight-forward (see #68), we can internally manage that without leaking the implementation to the consumer.

I would say that for now we keep it as is. The nested provider is **not a requirement**, so apps do not need to do that for now. Maybe when we do #68, we can refactor the internals of the AppShell to use a different store context without requiring apps to do that. So better to not promote much the fact that the apps can use a nested provider.

I'll refactor the playground to keep it without, to avoid confusion. It was anyway for testing purposes to make sure things are still working.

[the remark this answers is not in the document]
emmenko: Done, see https://github.com/commercetools/merchant-center-application-kit/pull/197/commits/bc35bc583f9311184f961931445d16477fa51bff

emmenko:
Update: I've added again a `<InjectReducers>` to allow apps to pass their own reducers and being injected to the default redux `store`.

This is mostly for "backwards compatibility" in the sense of not requiring apps to use a custom redux store/context. The API for the `<InjectReducers>` changed a bit though:

- the component is now called `<InjectReducers>` instead of `<InjectReducer>` (plural vs singular). The reason being a different API and because we now pass an **object of reducers**, instead of **combining the reducers** first and pass the final reducer
- the passed **reducers object** is now namespaced by `id` in the `store.injectedReducers` (for checking purposes) but is spread when creating the actual store reducer. This is necessary because now we don't have the "LocalStoreProvider" that maps the namespaced store to the given reducer name. It also allows to inject reducers multiple times (the user still needs to be careful with naming conflicts).

In other words, apps need little changes to keep working as-is when they upgrade to this version. Furthermore, they have the option to use a nested redux store/context ([not recommended yet but will simplify things in the future](https://github.com/commercetools/merchant-center-application-kit/pull/197#discussion_r244685548)).

packages/actions-global/global.js
```diff
@@ -66,11 +65,11 @@ export function showNotification(notification, meta = {}) {
   if (!isNumber(dismissAfter))
     dismissAfter = notification.kind === 'success' ? 5000 : 0;
 
-  return toGlobal(addNotification(notification, { ...meta, dismissAfter }));
```
qmateub: Finally 🙏

packages/application-shell/src/components/inject-reducers/inject-reducers.spec.js
```diff
@@ -0,0 +1,52 @@
...
+      </InjectReducers>
+    );
+
+    await waitForElement(() => getByText(/The count is 0/i));
+
+    fireEvent.click(getByTestId('increment'));
+
+    await waitForElement(() => getByText(/The count is 1/i));
+  });
+});
```
emmenko: I'm still amazed by how much more robust tests can be if we don't test implementation details 🤩

packages/application-shell/src/components/inject-reducers/inject-reducers.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { ReactReduxContext } from 'react-redux';
+
+export class InjectReducers extends React.PureComponent {
+  static displayName = 'InjectReducers';
+  static propTypes = {
+    id: PropTypes.string.isRequired,
+    reducers: PropTypes.objectOf(PropTypes.func).isRequired,
+    shouldCleanUpWhenUnmounts: PropTypes.bool,
```
montezume: shouldCleanUpOnUnmount?

packages/application-shell/src/components/inject-reducers/inject-reducers.spec.js
```diff
@@ -0,0 +1,52 @@
...
+    case 'INCREMENT_COUNT':
+      return state + 1;
+    default:
+      return state;
+  }
+};
+const reducers = { counter: counterReducer };
+
+describe('injecting reducers', () => {
+  it('should inject reducers and render count from connected store', async () => {
```
montezume: nice use of RTL 🎉

packages/application-shell/src/test-utils/README.md
```diff
@@ -119,6 +120,7 @@ This section describes the methods exported by `@commercetools-frontend/applicat
...
 | `options.addTypename` | `Boolean`     | Apollo                                                                                                                                                                                             | If queries are lacking `__typename` (which happens when mocking) it’s important to pass `addTypename: false`, which is the default. See [`MockedProvider.addTypename`](https://www.apollographql.com/docs/guides/testing-react-components.html#addTypename) for more information.                                                         |
+| `options.sdkMocks`    | `Array`       | Redux                                                                                                                                                                                              | Allows to mock requests made with `@commercetools-frontend/sdk` (Redux). A mock contains an object with `{ action, response }`, where `action` represents the Redux action from the `@commercetools-frontend/sdk` package and `response` represents the network JSON response.                                                            |
```
montezume: Allows `mocking` requests ... A mock contains an object with `the shape` ...

packages/application-shell/src/test-utils/test-utils.js
```diff
@@ -199,11 +194,11 @@ const renderWithRedux = (
 ) => {
   invariant(
     !(store && storeState),
-    'test-utils: You provided store and storeState. Provide one of them only.'
+    'test-utils: You provided both `store` and `storeState`. Please provide only one of them.'
```
montezume: nice english fix 🎉

emmenko:
I'll merge this tomorrow morning (Thursday) so that @tdeekens can have a final look.

packages/application-shell/src/components/inject-reducers/inject-reducers.spec.js
tdeekens: Pretty good validation 😇

packages/sdk/src/components/sdk-get/sdk-get.js
```diff
@@ -125,7 +125,12 @@ export class SdkGet extends React.Component {
   }
 }
 
-export default connect(
-  null,
-  dispatch => ({ dispatch })
-)(SdkGet);
+export default function createSdkGet(connectOptions = {}) {
```
tdeekens: Do we still need this? Disregard comment if so.

emmenko: Not necessarily, but I felt like keeping it as it's mostly an internal change and can already allow consumers of the package to pass a custom context. It's also not documented so we might use it to experiment with things.

However, I can also revert it if we don't want to "give this out" just yet.

tdeekens: I personally would not give this out. Even an undocumented breaking change or an public API is a breaking change. If we want to play with something like this maybe we should do so from a canary release not from the latest dist-tag. I am also not sure if we would name the API like that. Just my two cents though.

tdeekens: I personally would not give this out. Even an undocumented breaking change or an public API is a breaking change. If we want to play with something like this maybe we should do so from a canary release not from the latest dist-tag. I am also not sure if we would name the API like that. Just my two cents though.

emmenko:
Alright, let's finally do this!! 🚀

PR #199, opened 2018-12-13
chore: update ui-kit to v5

montezume:
#### Summary

Update ui-kit to version [v5.0.0](https://github.com/commercetools/ui-kit/releases/tag/v5.0.0). App kit isn't affected by the breaking changes. 🎉 🇨🇦 

Decided to make this WIP as the upgrade in the front end is quite involved. Better to get it done there before merging this, or else we will be deploy-blocked.


emmenko:
@montezume @tdeekens should we also include this in the `v5` release?

tdeekens:
I think so @montezume and @dferber90 should say.

montezume:
yeah why not 🎉

emmenko:
I'm looking into the failing build, because the Table prints a warning (making CI fail)...

emmenko:
Alright, I fixed it 
https://github.com/commercetools/merchant-center-application-kit/pull/199/commits/5226f8b6473e03e66893922022f9dec618ac1635. It's a bit hacky but it does its job for now. This also shows that we need to have a better developer friendlier `<Table>`, as it's also impossible to test it with RTL at the moment.

montezume:
@emmenko just use puppeteer and dom testing library to test it :trollface:

PR #200, opened 2018-12-13
feat(examples/starter): add example starter setup

emmenko:
This PR adds a new project setup in the `examples/starter` folder, so that people can copy-paste and get started with the bare minimum setup.


examples/starter/src/components/entry-point/entry-point.spec.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { shallow } from 'enzyme';
```
montezume: could we use react-testing-library instead? I would leave out enzyme entirely.

examples/starter/webpack.config.prod.js
```diff
@@ -0,0 +1,12 @@
+const path = require('path');
+const createWebpackConfigForProduction = require('@commercetools-frontend/mc-scripts/config/create-webpack-config-for-production');
```
montezume: also feels really opinionated. Since we are shipping compiled code, there's no reason why they need to use our webpack config right? maybe they want to use parcel 🤷‍♂️

tdeekens: Yeah but it is like the easy way to get started. So I think it is fine showing that they can use our config too.

examples/starter/src/components/entry-point/entry-point.spec.js
emmenko: I did (`main-view.spec.js`), but I also wanted to have a simple example of using shallow rendering.

tdeekens:
> separate repo

Historically we had 
https://github.com/commercetools/merchant-center-starter-app. I guess it is just a bit simpler to have one area to point people to and for us to maintain. But I agree the copy and paste is less convienient than a git clone.

examples/starter/webpack.config.prod.js
emmenko: It should be possible yes. However, this is meant to be a starter example that you just copy-paste and get started. We can also provide a separate one with parcel, or document it how to use it.
I would take it one step at a time ;)

emmenko:
> Are you sure we dont forget or can we test this somehow? Almost any CLI option could be forgotten.

It's more or less already done in CI: we install the package, lint, test and build it. Or did you mean something else?

emmenko:
> n general it looks cool, but are we sure this is the best place for it? I think it would be better as a seperate repository that they could just clone, rather than copy pasting from this examples folder.

I would keep it here because it's easier to test and update it.

> Also, I think it's a bit overopinionated and doesn't need to have the linting, prettier, webpack config, etc.

It's supposed to be a starter setup, that you just copy-paste it and run it. If we leave out those extra things you would probably have to add them anyway. In the future maybe we can have our CLI bootstrap such a project.
Also, if you don't want to use something, it's easier to remove stuff then having to add it.

tdeekens:
1. Too much boilerplate and tooling

I understand the argument but I think we're kind of applying/suggesting a our own and a default (almost industry standard) tooling setup. So it should we fine as they can remove things. It's not like CRA where you have to eject (yet).

2. Testing and consistency with app

Yes, I think building and testing is one big aspect of it! How do we ensure however that the `env.json` stays consistent with our real world setup? It won't change too much I agree. Maybe it's fine.

montezume:
1. I would just ship an example app using CRA and stop trying to reinvent the wheel. By using our webpack setup, there is so many opinions made about the setup.

montezume:
Also, by providing a webpack config we can make it more explicit about what we are actually doing. I would imagine that the most relevant part of our webpack config to consumers would be 

```js
                postcssCustomMediaQueries({
                  importFrom: require.resolve(
                    '@commercetools-frontend/ui-kit/materials/media-queries.mod.css'
                  ),
                }),
                postcssCustomProperties({
                  preserve: false,
                  importFrom: require.resolve(
                    '@commercetools-frontend/ui-kit/materials/custom-properties.css'
                  ),
                }),
```

so if they copy pasted this setup, and then wanted to move to their own webpack setup, they would probably be a bit stumped as to why their customProperties aren't working anymore 😆

examples/starter/README.md
```diff
@@ -0,0 +1,28 @@
+<p align="center">
+  <a href="https://commercetools.com/">
+    <img alt="commercetools logo" src="http://cdn.rawgit.com/commercetools/press-kit/master/PNG/72DPI/CT%20logo%20chrom%20black%20horizontal%20RGB%2072dpi.png">
+  </a>
+  <b>Merchant Center Examples: Starter.</b>
+</p>
+
+This example contains the bare minimum setup to get you started with the development of the Merchant Center Application.
```
celeste-horgan:
```suggestion
This example application gives you the minimum setup to develop and launch a working Merchant Center application.
```

examples/starter/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Installation
+
+Checkout the github repository and copy-paste the `examples/starter` folder into your new project.
+
```
celeste-horgan:
```suggestion

To do so, run the following commands:

```

examples/starter/README.md
```diff
@@ -0,0 +1,28 @@
...
+```bash
+$ git checkout git@github.com:commercetools/merchant-center-application-kit.git
+
+$ mkdir my-new-custom-application-project
+$ cp -R merchant-center-application-kit/examples/starter/* my-new-custom-application-project
+
```
celeste-horgan:
```suggestion

Next, run the following command to initialize the yarn workspace:

```

examples/starter/README.md
```diff
@@ -0,0 +1,28 @@
...
+$ cd my-new-custom-application-project
+$ yarn
+```
+
+## Start the development server
+
```
celeste-horgan:
```suggestion

Run the following command to start the development server and launch the application:

```

emmenko:
> I would just ship an example app using CRA and stop trying to reinvent the wheel

That's currently not possible. Our `mc-scripts` have some specific logic to e.g. load the `env.json` or `csp.json`, an html template, etc. Trying to start an app without those things will simply fail.

Nevertheless, it might be a goal to improve our tooling to be less dependent from our own setup.

emmenko:
I’ve been thinking about this and in general it would be cool to have this “boilerplate tooling setup” like jest configs, husky, prettier etc into a package (kind of like `kcd-scripts`) that we can re-use across all our projects.

Then this example can be simpler. For now I would keep it like this and as we improve our tooling we can remove things and make it simpler.

montezume:
Can we hold off on merging this? I’m concerned about how opinionated the scripts are. We can talk about it tomorrow

emmenko:
This needs to be merged in by latest tomorrow, as it's referenced by the custom apps documentation

montezume:
I’m concerned that if a user wants to integrate this into their setup, or use their own webpack config, it will be confusing and we don’t provide any example of that. But yeah, my concern is noted. Merge if it’s urgent

emmenko:
> I’m concerned that if a user wants to integrate this into their setup

Which setup? The example is just to get started from scratch.

As for the other points, yes we can and will provide better examples but for now this is what we need to get people started.

montezume:
My flights about to take off so i cant argue about this, but I have already given several comments over the last few days about my concerns.

emmenko:
Let's talk about it when you're back, I think we want the same thing at the end 😉

Have a good flight! ✈️

tdeekens:
Please @montezume don't feel overruled by this. I really value this kind of persistent feedback! And always continue to do so! I also don't think it would have happened differently if you were not on a flight back. The "need" culminating into having it like this precedes this PR.

emmenko:
Yes I think we all want the same thing at the end. I'm also the first one not 100% happy with the current solution but given the time and deadline that's the "optimal" solution so far.

We will have different follow ups to improve our tooling and we can then decide how to manage those examples.

tdeekens:
> We will have different follow ups to improve our tooling and we can then decide how to manage those examples.

Yes, a really interesting aspect of this is how much things will decrease our ability to move our own infra later on. E.g.: we will never be able to remove the header based schema stitching from the mc-be likely.

montezume:
Okay, thanks for the explanation. Let's get this merged then as is, and when I have some time I'll create some issues with example code of how we can de-opinoniate the setup.

tdeekens:
Thanks Malcolm. Really appreciated.

emmenko:
I started an issue and wrote down some of the topics that we noticed. #214 

Let's maybe gather some ideas and have some meetings to discuss how we want to proceed there.

The document is meant to be updated with things we decide and find along the way. Feel free to update and add new topics if you have more ideas.

PR #209, opened 2018-12-17
Add pim search to quick access

dferber90:
pim-search

This PR
- adds support for pim-search to Quick Access.
- also adds support of mocking `sdk` requests made through the sdk middleware to `test-utils` - similar to how `ApolloMockProvider` lets us mock GraphQL requests.

I might eventually split the PR up into the two separate parts, but I left it in one PR while gathering feedback for now.

### DIFF
--- a/packages/application-shell/src/components/quick-access/index.js
+++ b/packages/application-shell/src/components/quick-access/index.js
@@ -45,6 +45,10 @@ class QuickAccessContainer extends React.Component {
 export class QuickAccessTrigger extends React.Component {
   static displayName = 'QuickAccessTrigger';
 
+  static propTypes = {
+    getPimSearchStatusForProject: PropTypes.func.isRequired,
+  };
+
   componentDidMount() {
     document.addEventListener('keydown', this.handler);
     if (this.props.project) {
@@ -138,7 +142,7 @@ export class QuickAccessTrigger extends React.Component {
   // Otherwise we'd need to
   // - ensure the response we receive belongs to the current project
   // - refetch the pim search info when the project key changes
-  updatePimSearchInfo = () => {
+  updatePimSearchInfo = async () => {
     // skip when there is no project
     if (!this.props.project) return;
 
@@ -150,41 +154,8 @@ export class QuickAccessTrigger extends React.Component {
     // skip checking when user can't view products anyways
     if (!canViewProducts) return;
 
-    this.props
-      .dispatch(
-        // TODO this should be sdkActions.head()
-        // and then we should check whether the response code is
-        // - 200 meaning the project is indexed
-        // - 404 meaning the project is not indexed
-        //
-        // But there is a problem in tne node-sdk client as it tries to
-        // .json()-parse the response to HEAD requests which results in an
-        // error, so we send a regular request for now and limit to no results
-        // instead to keep the payload minimal
-        sdkActions.post({
-          uri: `/proxy/pim-search/${this.props.project.key}/search/products`,
-          payload: {
-            query: {
-              fullText: {
-                field: 'name',
-                language: this.props.projectDataLocale,
-                value: '',
-              },
-            },
-            limit: 0,
-            offset: 0,
-          },
-        })
-      )
-      .then(
-        () => true,
-        // project is not using pim-indexer when response error code is 404,
-        // but we treat all errors as non-indexed as a safe guard
-        () => false
-      )
-      .then(isProjectIndexed => {
-        this.setState({ isProjectIndexed });
-      });
+    const isProjectIndexed = await this.props.getPimSearchStatusForProject();
+    this.setState({ isProjectIndexed });
   };
 
   render() {
@@ -202,5 +173,37 @@ export class QuickAccessTrigger extends React.Component {
 
 export default connect(
   null,
-  dispatch => ({ dispatch })
+  (dispatch, props) => ({
+    getPimSearchStatusForProject: () =>
+      dispatch(
+        // TODO this should be sdkActions.head()
+        // and then we should check whether the response code is
+        // - 200 meaning the project is indexed
+        // - 404 meaning the project is not indexed
+        //
+        // But there is a problem in tne node-sdk client as it tries to
+        // .json()-parse the response to HEAD requests which results in an
+        // error, so we send a regular request for now and limit to no results
+        // instead to keep the payload minimal
+        sdkActions.post({
+          uri: `/proxy/pim-search/${props.project.key}/search/products`,
+          payload: {
+            query: {
+              fullText: {
+                field: 'name',
+                language: props.projectDataLocale,
+                value: 'availability-check',
+              },
+            },
+            limit: 0,
+            offset: 0,
+          },
+        })
+      ).then(
+        () => true,
+        // project is not using pim-indexer when response error code is 404,
+        // but we treat all errors as non-indexed as a safe guard
+        () => false
+      ),
+  })
 )(QuickAccessTrigger);
--- a/packages/application-shell/src/components/quick-access/index.js
+++ b/packages/application-shell/src/components/quick-access/index.js
@@ -61,8 +61,16 @@ export class QuickAccessTrigger extends React.Component {
   state = {
     isVisible: false,
     hasError: false,
+    // We store the information of whether a project is indexed by pim-indexer,
+    // to avoid having to refetch that information every time Quick Access is
+    // opened. We can't move the information to the quick-access.js component
+    // as that component unmounts and would lose its state.
+    //
+    // We need to know whether a project is indexed by pim-indexer to know
+    // whether we should query pim-search or whether we can skip that request.
+    //
     // We don't need to update this information when the project key changes,
-    // as changing a project always results in a full reload anyways.
+    // as changing a project always results in a full page reload anyways.
     //
     // undefined: we did not check yet
     // true: the project is indexed by pim-indexer
--- a/packages/application-shell/src/components/quick-access/index.js
+++ b/packages/application-shell/src/components/quick-access/index.js
@@ -1,17 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { connect } from 'react-redux';
-import { actions as sdkActions } from '@commercetools-frontend/sdk';
 import { reportErrorToSentry } from '@commercetools-frontend/sentry';
-import {
-  hasSomePermissions,
-  permissions,
-} from '@commercetools-frontend/permissions';
 import * as gtm from '../../utils/gtm';
 import trackingEvents from './tracking-events';
 import ButlerContainer from './butler-container';
 
-const QuickAccessModal = React.lazy(() =>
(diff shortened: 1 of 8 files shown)


packages/application-shell/src/components/quick-access/index.js
```diff
@@ -36,11 +42,18 @@ class QuickAccessContainer extends React.Component {
...
 
+  static propTypes = {
+    getPimSearchStatusForProject: PropTypes.func.isRequired,
+  };
+
   componentDidMount() {
     document.addEventListener('keydown', this.handler);
+    if (this.props.project) {
+      this.updatePimSearchInfo();
+    }
```
dferber90: We check whether PIM-Search is enabled first to avoid making requests to pim-search for projects which don't use pim-search yet.

packages/application-shell/src/components/quick-access/index.js
```diff
@@ -121,11 +137,73 @@ export default class QuickAccessTrigger extends React.Component {
...
+      dispatch(
+        // TODO this should be sdkActions.head()
+        // and then we should check whether the response code is
+        // - 200 meaning the project is indexed
+        // - 404 meaning the project is not indexed
+        //
+        // But there is a problem in tne node-sdk client as it tries to
+        // .json()-parse the response to HEAD requests which results in an
+        // error, so we send a regular request for now and limit to no results
+        // instead to keep the payload minimal
```
dferber90: I could not get `head` requests to work due to something inside of `node-sdk` which tries to parse a nonexistent response body. We'd need to dig deeper in case we want to solve this. However things seem to work quite fine while making `POST` requests as a workaround.

packages/application-shell/src/components/quick-access/quick-access.js
```diff
@@ -92,17 +100,29 @@ class QuickAccess extends React.Component {
...
-        // Pass conditional arguments to disable some of the queries
-        canViewProducts: hasSomePermissions(
-          [permissions.ViewProducts, permissions.ManageProducts],
-          this.props.project.permissions
-        ),
-      }).then(data => {
+  getProjectCommands = debounce(async searchText => {
+    const idsOfProductsMatchingSearchText = this.props.isProjectIndexed
+      ? await this.props.pimSearchProductIds(searchText)
+      : [];
```
dferber90: Here we only search products in case the project is indexed by pim-indexer.

packages/application-shell/src/components/quick-access/quick-access.js
```diff
@@ -283,5 +327,36 @@ export default compose(
     'canViewDashboard',
     'canViewDiscounts',
     'customApplications',
-  ])
+  ]),
+  connect(
+    null,
+    (dispatch, props) => ({
+      pimSearchProductIds: searchText =>
```
dferber90: This is where we query pim-search using full-text search.

packages/application-shell/src/components/quick-access/quick-access.spec.js
```diff
@@ -1045,4 +1258,102 @@ describe('QuickAccess', () => {
...
+              productByVariantKey: null,
+            },
+          },
+        },
+      ];
+      const { getByTestId, getByText } = renderWithRedux(
+        <QuickAccess {...createTestProps()} />,
+        {
+          mocks,
+          sdkMocks: [
```
dferber90: This is the best example of how the new `sdkMocks` option in `renderWithRedux` can be used to mock sdk responses.

packages/application-shell/src/components/quick-access/index.js
adnasa: funny because I ran into the exact same problem
ref: https://github.com/commercetools/nodejs/issues/1022

packages/application-shell/src/components/quick-access/index.js
```diff
@@ -61,6 +74,9 @@ export default class QuickAccessTrigger extends React.Component {
   state = {
     isVisible: false,
     hasError: false,
+    // We don't need to update this information when the project key changes,
+    // as changing a project always results in a full reload anyways.
+    isProjectIndexed: false,
```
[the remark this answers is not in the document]
dferber90: The pim index applies to the whole project. I'm confused. Which name would you suggest?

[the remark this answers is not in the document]
emmenko: For example, `isProjectPimIndexed`.

packages/application-shell/src/components/quick-access/index.js
dferber90: > Can we maybe check this only when the user starts searching for products specifically?

This is not possible. We do not know what the user is searching for. They just enter a text and we find anything that could match. So we always need to search products.

> I'm just wondering if we need to do this 

We use this information to avoid making requests when a user searches. If the project is not indexed then we don't search products using pim-search at all and don't make a request again. We check it once upfront so that we already know whether we need to query pim-search when the user actually starts typing.

In the beginning I loaded the information when the search component itself mounts, but that felt wrong as it would lead to a new request every time quick-access is opened again (as the component loses state when it unmounts).

> I'm not too familiar with the internals, so I can't say for sure if that would be possible...

I think we can get the best of both worlds. We can store the information of whether to search or not in the state of `QuickAccessTrigger` (where it is now), and we can load that state (in case it was not loaded yet) when `QuickAccess` mounts.

This means we:
- load the information only once (per application)
- load the information only when the user activates quick-access instead of always

Thanks for pointing it out 👍

packages/application-shell/src/components/quick-access/index.js
dferber90: The refactor should be quite manageable and it doesn't seem like there's any action going on upstream. I'd rather get PIM Search support in right away and avoid this branch going stale.

What you want to do once `nodejs` supports `HEAD` requests is to change this to `sdkActions.head`, remove the `payload` and adapt all tests (the `createPimAvailabilityCheckSdkMock` function). You also need to add the `sdkActions.head` function to the `'@commercetools-frontend/sdk` package.

packages/application-shell/src/components/quick-access/quick-access.js
```diff
@@ -92,17 +100,29 @@ class QuickAccess extends React.Component {
...
+      [permissions.ViewProducts, permissions.ManageProducts],
+      this.props.project.permissions
+    );
+
+    return this.query(QuickAccessQuery, {
+      searchText: sanitize(searchText),
+      target: GRAPHQL_TARGETS.COMMERCETOOLS_PLATFORM,
+      // Pass conditional arguments to disable some of the queries
+      canViewProducts,
+      productsWhereClause: `id in (${idsOfProductsMatchingSearchText
```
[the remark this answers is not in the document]
dferber90: When the list is empty, then `includeProductsByIds` is `false` and the request is not executed. So this is not a problem.

Sidenote: Even if it was executed then `id in ()` actually works (I had this for a bit while working on this PR).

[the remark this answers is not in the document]
emmenko: Cool, just making sure 😉

packages/application-shell/src/components/quick-access/index.js
emmenko: Sure, it's fine by me 👌

packages/application-shell/src/components/quick-access/quick-access.js
dferber90: I want to keep them close to the code and don't see an upside in that. In the beginning I even had

```
connect(
  null,
  (dispatch) => ({ dispatch })
)
```

which allowed me to "inline" the `sdkActions.post` call to where it's actually being triggered. I liked that as it kept things together. I ultimately moved it down with this expression on my face 😐 as it's "the way it's supposed to be" and because it felt kinda weird to just pass `dispatch`.

packages/application-shell/src/test-utils/test-utils.js
```diff
@@ -163,26 +165,114 @@ const renderWithRedux = (
     // Consumers of renderWithRedux can use
     //   { store: createReduxStore({ requestsInFlight: null, .. }) }
     // to pass an initial state to Redux.
-    store = createReduxStore(),
+    store = undefined,
```
dferber90: Now that I slept on it, maybe we should just remove `store` completely? Maybe it's the right time to tighten the API? Not sure..

packages/application-shell/src/components/quick-access/index.js
dferber90: There is no good caching in `sdk`. Moving the request to `cDM` of `QuickAccess` seems ideal as that's exactly when we need the information. By letting the parent handle the state we avoid making the request twice and we control the "caching" ourselves. I'm sold on that approach as it's simple and minimal effort.

packages/application-shell/src/test-utils/test-utils.js
dferber90: I wouldn't add a note to say that it "might be deprecated." I would only add such a note in case we are certain that we're going to get rid of it. If we need to keep it around, then it's here to stay anyways. No consumers in the MC are using it at the moment.

My gut feeling is that we can remove it - but then again I also didn't see the `sdk` mocking stuff coming up.

packages/application-shell/src/components/quick-access/index.js
emmenko: I'd like to have at least a code comment explaining those trade-offs.

dferber90: I can add that once it's implemented 👍

dferber90: Done in the latest commit 40d576e598d31c9740e9791d2ab2abd7fbe4c8e8

packages/application-shell/src/components/quick-access/index.js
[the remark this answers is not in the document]
dferber90: Done 👍

packages/application-shell/src/components/quick-access/index.js
```diff
@@ -61,6 +61,21 @@ export default class QuickAccessTrigger extends React.Component {
...
+    // We need to know whether a project is indexed by pim-indexer to know
+    // whether we should query pim-search or whether we can skip that request.
+    //
+    // We don't need to update this information when the project key changes,
+    // as changing a project always results in a full page reload anyways.
+    //
+    // undefined: we did not check yet
+    // true: the project is indexed by pim-indexer
+    // false: the project is not indexed by pim-indexer
+    isProjectPimIndexed: undefined,
```
pa3: While `undefined` semantics of booleans in JS is kind of legit, I personally don't like this three-state-booleans, for exactly this reason: you have to add comments all around this places, so that people reading will take `undefined` into consideration.

dferber90: Yeah that bothered me as well while writing that. I can create an enum for it instead to make it clearer:

```
const pimIndexerStates = 
  UNCHECKED: 'UNCHECKED',
  INDEXED: 'INDEXED',
  NOT_INDEXED: 'NOT_INDEXED',
}
```

PR #213, opened 2018-12-18
Add `allAppliedPermissions` and remove `permissions`

tdeekens:
#### Summary

This pull request adds the not yet implemented (@qmateub is on it) `allAppliedPermissions` field.

#### Description

For the underlying system nothing should change. We map back the `allAppliedPermissions` of shape `[ { name: string, value: boolean } ]` into `{ [name: string]: boolean }`. 

This all in all should decouple app-kit from any app in the future in regards to permissions as anything the mc-be decides to flush through will be available. The only left-over dependency is constants.

### DIFF
--- a/packages/application-shell/src/components/fetch-project/fetch-project.js
+++ b/packages/application-shell/src/components/fetch-project/fetch-project.js
@@ -12,6 +12,26 @@ import { GRAPHQL_TARGETS } from '@commercetools-frontend/constants';
 import deprecateComponent from '../../from-core/deprecate-component';
 import ProjectQuery from './fetch-project.graphql';
 
+/**
+ * NOTE:
+ *
+ * Permissions are being fetched though the `allAppliedPermissions` which
+ * is an array of `{ name: string, value: boolean }`. This gives more
+ * flexibility to introduce new permissions to apps without having to release
+ * the merchant-center-app-kit.
+ *
+ * The application below however expects permissions to be of the shape
+ * `[name: string]: boolean` which is what the shape above is mapped into here.
+ */
+export const mapAppliedPermissionsToPermissions = appliedPermissions =>
+  appliedPermissions.reduce(
+    (transfromedPermissions, appliedPermission) => ({
+      ...transfromedPermissions,
+      [appliedPermission.name]: appliedPermission.value,
+    }),
+    {}
+  );
+
 const graphqlOptions = {
   alias: 'withProject',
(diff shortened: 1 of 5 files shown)


tdeekens:
To be decided. 

1. Have the components log a warning when any permission is to be toggled upon which wasnt fetched
2. Deal with it for now as there are more important things
3. Pass any string and not use them

I would likely go somewhere between 1. and 2.

emmenko:
Is anything else missing from your side or can we finally review it? Asking because of the "WIP"...

tdeekens:
I went a bit back and forth with the approach but decided to solve it on the level of `fetch-project` to not have breaking changes anywhere else in the application.

I want the query to actually be available before removing the WIP status (so nobody merges it by accident). Feel free to leave an approval given the code you see here.

packages/application-shell/src/components/fetch-project/fetch-project.js
```diff
@@ -12,6 +12,26 @@ import { GRAPHQL_TARGETS } from '@commercetools-frontend/constants';
...
+ *
+ * Permissions are being fetched though the `allAppliedPermissions` which
+ * is an array of `{ name: string, value: boolean }`. This gives more
+ * flexibility to introduce new permissions to apps without having to release
+ * the merchant-center-app-kit.
+ *
+ * The application below however expects permissions to be of the shape
+ * `[name: string]: boolean` which is what the shape above is mapped into here.
+ */
+export const mapAppliedPermissionsToPermissions = appliedPermissions =>
```
emmenko: Would it make sense to move this function into the `permissions` package instead?

tdeekens: True. Maybe it makes sense to expose it too.

emmenko: Sure, was just a suggestion 👌

tdeekens: I value it! It's a really cool one. Maybe I'll add a note to make sure that we remind ourselves why this ain't moved yet.

qmateub:
@tdeekens PR in the MC-BE merged. We need to deploy that and then we can use it here 🎉

tdeekens:
Sure. I‘ll test a built version of this later locally and merge it then.

tdeekens:
Tested with a local build all fine.

PR #216, opened 2018-12-20
Fix test utils and quick access

dferber90:


packages/application-shell/src/components/quick-access/quick-access.js
```diff
@@ -134,9 +134,10 @@ class QuickAccess extends React.Component {
 
   getProjectCommands = debounce(
     async searchText => {
-      const idsOfProductsMatchingSearchText = this.props.pimIndexerState
-        ? await this.props.pimSearchProductIds(searchText)
-        : [];
+      const idsOfProductsMatchingSearchText =
+        this.props.pimIndexerState === pimIndexerStates.INDEXED
```
dferber90: I forgot to add this after introducing the `pimIndexerStates` enum.

packages/application-shell/src/test-utils/test-utils.js
```diff
@@ -242,10 +245,10 @@ const renderWithRedux = (
             )}`
           );
 
-        const { response, error } = clonedSdkMocks[index];
+        const { response, error } = sdkMocksStack[index];
 
         // remove result as each result is only mocked once
-        sdkMocks.splice(index, 1);
+        sdkMocksStack.splice(index, 1);
```
dferber90: I was mutating the user-passed thing instead of the clone. This resulted in the error not showing the `sdkMocks` provided by the user but by showing the mocks at that point in time instead which made debugging a bit harder.

PR #217, opened 2018-12-20
Refactor `Authorized` in permissions pkg to not rely on constants

tdeekens:
#### Summary

After the mc-fe now can work with `allAppliedPermissions` as a more flexible `Array<AppliedPermission>` we should also improve the behaviour of `Authorized`. 

#### Details

`Authorized` currently used the constants to determine in its prop-types if the permission is known. This is a bit inflexible:

1. Introducing new permissions requires an app-kit release
2. Tight coupling and duplication of consumer and producer

Give that we can loosen the contract a bit maybe by _validating known (actual) and required (demanded)_ permissions against one another. Given we find one which we do not know about we give a prop-type warning.

Side note: I find the naming of `demanded` and `actual` a bit misleading cause it doesn't provide an inverse: what's the opposite of actual?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,9 +33,9 @@ before_install:
 
 script:
   - set -e
+  - yarn run jest --projects jest.*.config.js --maxWorkers=4 --reporters jest-silent-reporter
   - yarn build
   - yarn playground:build
-  - yarn run jest --projects jest.*.config.js --maxWorkers=4 --reporters jest-silent-reporter
   - set +e
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,9 +33,9 @@ before_install:
 
 script:
   - set -e
-  - yarn run jest --projects jest.*.config.js --maxWorkers=4 --reporters jest-silent-reporter
   - yarn build
   - yarn playground:build
+  - yarn run jest --projects jest.*.config.js --maxWorkers=4 --reporters jest-silent-reporter
   - set +e
 
 after_success:
--- a/packages/application-shell/src/test-utils/test-utils.spec.js
+++ b/packages/application-shell/src/test-utils/test-utils.spec.js
@@ -169,9 +169,7 @@ describe('ApplicationContext', () => {
 
   describe('permissions', () => {
     const TestComponent = () => (
-      <RestrictedByPermissions
-        permissions={[{ mode: 'manage', resource: 'products' }]}
-      >
+      <RestrictedByPermissions permissions={['ManageProducts']}>
         {({ isAuthorized }) => (isAuthorized ? 'Authorized' : 'Not allowed')}
       </RestrictedByPermissions>
     );
--- a/packages/permissions/src/components/authorized/authorized.js
+++ b/packages/permissions/src/components/authorized/authorized.js
@@ -1,24 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { wrapDisplayName } from 'recompose';
-import { defaultMemoize } from 'reselect';
-import warning from 'warning';
-import camelCase from 'lodash.camelcase';
-import upperFirst from 'lodash.upperfirst';
 import { ApplicationContext } from '@commercetools-frontend/application-shell-connectors';
 import {
   hasSomePermissions,
   hasEveryPermissions,
   getUnconfiguredPermissions,
 } from '../../utils/has-permissions';
-
-const ensurePermissionsKeyShape = defaultMemoize(demandedPermissions =>
-  demandedPermissions.map(permission =>
-    typeof permission === 'string'
-      ? permission
-      : upperFirst(camelCase(`${permission.mode} ${permission.resource}`))
-  )
-);
 
 class Authorized extends React.Component {
   static displayName = 'Authorized';
@@ -47,10 +35,7 @@ class Authorized extends React.Component {
       }
 
       // Legacy shape of permissions
-      return PropTypes.shape({
-        mode: PropTypes.string.isRequired,
-        resource: PropTypes.string.isRequired,
-      });
+      return null;
     },
     actualPermissions: PropTypes.objectOf(PropTypes.bool).isRequired,
     render: PropTypes.func,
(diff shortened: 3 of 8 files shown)


tdeekens:
Note, that this now also removes the legacy shape support of permissions and there by is a breaking change. I can't find that shape anywhere in our mc-fe code anymore. Correct me if I am wrong. Otherwise I would also clean that up when upgrading to this.

.travis.yml
```diff
@@ -33,9 +33,9 @@ before_install:
 
 script:
   - set -e
-  - yarn run jest --projects jest.*.config.js --maxWorkers=4 --reporters jest-silent-reporter
   - yarn build
   - yarn playground:build
+  - yarn run jest --projects jest.*.config.js --maxWorkers=4 --reporters jest-silent-reporter
```
tdeekens: I think testing should occur after building. Without tests even failed.

packages/application-shell/src/test-utils/test-utils.spec.js
```diff
@@ -169,9 +169,7 @@ describe('ApplicationContext', () => {
 
   describe('permissions', () => {
     const TestComponent = () => (
-      <RestrictedByPermissions
-        permissions={[{ mode: 'manage', resource: 'products' }]}
-      >
```
tdeekens: This was deprecated already and just changed to the new format.

packages/permissions/src/components/authorized/authorized.js
```diff
@@ -1,66 +1,59 @@
...
 import { wrapDisplayName } from 'recompose';
-import { defaultMemoize } from 'reselect';
-import warning from 'warning';
-import camelCase from 'lodash.camelcase';
-import upperFirst from 'lodash.upperfirst';
 import { ApplicationContext } from '@commercetools-frontend/application-shell-connectors';
 import {
   hasSomePermissions,
   hasEveryPermissions,
+  getUnconfiguredPermissions,
```
tdeekens: The name unconfigured comes from the fact that there is no opposite to actual (from `actualPermissions`).

packages/permissions/src/utils/has-permissions.js
```diff
@@ -69,3 +70,17 @@ export const hasSomePermissions = (demandedPermissions, actualPermissions) =>
...
+//     ['ViewProducts', 'ManageOrders']
+// - actualPermissions:
+//     { canViewProducts: true, canManageOrders: false }
+export const getUnconfiguredPermissions = (
+  demandedPermissions,
+  actualPermissions
+) =>
+  demandedPermissions.filter(demandedPermission =>
+    isNil(actualPermissions[toCanCase(demandedPermission)])
+  );
```
tdeekens: Feels like this belongs here actually.

packages/permissions/src/utils/has-permissions.spec.js
```diff
@@ -104,3 +105,23 @@ describe('hasSomePermissions', () => {
     ).toBe(false);
   });
 });
+
+describe('getUnconfiguredPermissions', () => {
+  describe('given all permissions are configured (passed as `actualPermissions`', () => {
```
qmateub: Typo, your forgot the closing `)`

packages/permissions/src/utils/has-permissions.spec.js
```diff
@@ -104,3 +105,23 @@ describe('hasSomePermissions', () => {
...
+      expect(
+        getUnconfiguredPermissions(['ManageOrders'], { canManageOrders: true })
+      ).toHaveLength(0);
+    });
+  });
+
+  describe('given some permissions are not configured (not passed as `actualPermissions`', () => {
+    it('should return unconfigured permissions', () => {
+      expect(
+        getUnconfiguredPermissions(['ManageOrders', 'ViewStars'], {
```
qmateub: ⭐️

packages/permissions/src/components/authorized/authorized.js
emmenko: How about simply `getInvalidPermissions`?

packages/permissions/src/components/authorized/authorized.js
```diff
@@ -1,66 +1,59 @@
...
+     */
+    demandedPermissions(props, propName, componentName) {
+      const namesOfNonConfiguredPermissions = getUnconfiguredPermissions(
+        props.demandedPermissions,
+        props.actualPermissions
+      );
+
+      if (namesOfNonConfiguredPermissions.length > 0) {
+        return new Error(
+          `Invalid prop \`${propName}\` supplied to \`${componentName}\`. The permission(s) ${namesOfNonConfiguredPermissions.join(
```
emmenko: Nitpick: `namesOfNonConfiguredPermissions.toString()`

tdeekens:
@emmenko anything from your end? Just FYI: next steps are removing the dependency/usage in all apps towards the permissions constants. So they have their own, just like feature flags. Then we can again drop the constants here. I would like to, if possible, have that in one release of app-kit.

packages/permissions/src/components/authorized/authorized.js
tdeekens: Gives, `permissionA,permissionB` over `permissionA permissionB`. I don't mind I changed it to `toString`.

packages/permissions/src/utils/has-permissions.js
```diff
@@ -69,3 +70,14 @@ export const hasSomePermissions = (demandedPermissions, actualPermissions) =>
...
     hasPermission(permission, actualPermissions)
   );
+
+// Returns an Array<String> of unconfigured (not passed as `demandedPermissions`) permissions.
+// The shapes of the arguments are:
+// - demandedPermissions:
+//     ['ViewProducts', 'ManageOrders']
+// - actualPermissions:
+//     { canViewProducts: true, canManageOrders: false }
+export const getInvalidPermissions = (demandedPermissions, actualPermissions) =>
```
tdeekens: @emmenko referring to (you commented on an outdated ref I can't respond on):

> How about simply getInvalidPermissions?

Sure, I don't mind. Invalid can entail more than unconfigured to it may be more flexible for future needs.

tdeekens:
@emmenko 

> We put it before building on purpose, to avoid tests relying on build outputs.
> What's not working exactly?

Let me try to find the build... [here it is](https://travis-ci.org/commercetools/merchant-center-application-kit/builds/470558656). I tried again to revert the order back to the original [here](https://travis-ci.org/commercetools/merchant-center-application-kit/builds/472710352?utm_source=github_status&utm_medium=notification) - same result. I agree that build output before tests can be noisy but I always felt that building before testing is "good" as dependencies may depend among one another and should consume a fresh build.

emmenko:
Let me have a look first why the tests fail.

If we do decide to build first before testing, should we only do it on CI or also when we develop locally?

emmenko:
Btw: #197 has also breaking changes. Should we plan to do one major release including both PRs or should we do two separate major releases?

tdeekens:
I would suggest also when developing localif we decide toto keep things in sync. If I never build before developing or forget to I would either get unexpected results or a failing (due to no build) development experience?

Coordination for the breaking release is totally fine.

tdeekens:
I am on it and have a suspicion. To be updated.

tdeekens:
Opened this issue https://github.com/commercetools/merchant-center-application-kit/issues/228 to track the non prop-type removal.

tdeekens:
Sorry to bother but anything left here? /cc @emmenko

Issue #214, opened 2018-12-18
[Tooling] Future improvements


emmenko:
While implementing a starter example #200 , we bumped into some issues that we would like to address in the near future. Those can be grouped into different topics:

## 👶 Easy to install starter example
The objective here is to have a basic example (like the one we have in `examples/starter`) that can be installed / bootstrapped with one single command.

I guess the classic example for that would be [CRA](https://facebook.github.io/create-react-app/).

```bash
$ npx create-mc-app my-app
```

Ideally we could create **templates** that you can install using `create-react-app`

```bash
$ npx create-react-app my-app \
  --scripts-version=@commercetools-frontend/mc-scripts \
  --template-starter 
```

For that to work we just need to restructure a bit the `mc-scripts` package and start supporting templates.

## 🧘‍♀️ Loosen dependency tooling
At the moment to develop an app, our own tooling is required, such as `mc-scripts`, `mc-http-server`. While this is per-se not a problem, it however forces users to use our opinionated environment like Webpack.

Ideally, people can choose whatever bundler or setup they like in order to develop the applications.

> There are currently some problems with using different bundlers (e.g. parcel) as they don't fully support yet `core-js@3`, which causes several problems on production bundles.

#### Actions
_Checklist of what needs to be done to achieve this goal_
- [x] add script to compile html #670 

## 🚀 Deploy everywhere
At the moment, we recommend to use our `mc-http-server` as it comes with a couple of built-in features like injecting runtime stuff into the `index.html.template`, provides security headers, etc.

Until now, I found two use cases that won't work out of the box with our current setup:
- **static website like Netlify**: we need to provide some scripts or something that would help generating the necessary setup
- **serverless functions like Now (v2)**: they run on Lambda functions, which means we can't just use a docker container but we need to upload specific files for each function

#### Actions
_Checklist of what needs to be done to achieve this goal_
- [x] can deploy to Now v2 #670 
  - [ ] documentation example
- [ ] can deploy to netlify
  - [ ] documentation example

Freyert:
We've been exploring using Google Storage/S3 buckets as these are fairly common for customers to have easy access to when they want to deploy a custom app. Most of them deploy the static assets in S3 or Google Storage today, but the `mc-http-server` is still required.

It seems like it should be possible from looking at the `now.sh` the static generator creates. The only problem is that S3/Google Storage can't take wild cards and route them to a single file (nor really route files in general).

[...]

PR #226, opened 2018-12-24
feat(create-mc-app): add new package create-mc-app to bootstrap applications

emmenko:
Refs: #214 

Inspired by `create-react-app` (but kept much simpler).

Now to bootstrap a new application you simply need to:

```bash
$ npm install --global @commercetools-frontend/create-mc-app
$ create-mc-app my-new-custom-application-project --template starter

# or

$ npx @commercetools-frontend/create-mc-app my-new-custom-application-project --template starter
```

NOTE: this is just a first step on improving our tooling, there are many other improvements that we can do.


emmenko:
@tdeekens @montezume sorry to bother, have you had a chance to look at this?

packages/create-mc-app/bin/create-mc-app.js
```diff
@@ -0,0 +1,111 @@
...
+const path = require('path');
+const fs = require('fs');
+const os = require('os');
+const execSync = require('child_process').execSync;
+const mri = require('mri');
+
+const flags = mri(process.argv.slice(2), { alias: { help: ['h'] } });
+const commands = flags._;
+
+const availableTemplates = ['starter'];
```
tdeekens: Just curious: what are further templates? Given we have this option we already have some in the back of our heads?

packages/create-mc-app/bin/create-mc-app.js
```diff
@@ -0,0 +1,111 @@
...
+const templateName = flags.template || 'starter';
+if (!availableTemplates.includes(templateName)) {
+  throw new Error(
+    `The provided template name "${templateName}" does not exist. Available templates are "${availableTemplates.toString()}". Make sure you are also using the latest version of "@commercetools-frontend/create-mc-app".`
+  );
+}
+
+const projectDirectoryPath = path.resolve(projectDirectoryName);
+
+if (fs.existsSync(projectDirectoryPath)) {
```
tdeekens: Sorry this is me again: `throwIfProjectDirectoryExists`.

packages/create-mc-app/bin/create-mc-app.js
```diff
@@ -0,0 +1,111 @@
...
+  process.exit(0);
+}
+
+const projectDirectoryName = commands[0];
+
+if (!projectDirectoryName) {
+  throw new Error('Missing required argument "<project-directory>"');
+}
+const templateName = flags.template || 'starter';
+if (!availableTemplates.includes(templateName)) {
```
tdeekens: `throwIfTemplateIsNotSupported`

packages/create-mc-app/bin/create-mc-app.js
```diff
@@ -0,0 +1,111 @@
...
+
+// Change the package name based on the given project directory name
+const appPackageJson = JSON.parse(
+  fs.readFileSync(path.join(projectDirectoryPath, 'package.json'), {
+    encoding: 'utf8',
+  })
+);
+const updatedAppPackageJson = Object.assign({}, appPackageJson, {
+  name: projectDirectoryName.toLowerCase().replace(/_/gi, '-'),
+});
```
tdeekens: Maybe a comment would help: 

```
// Given the package name is derived from the `projectDirectoryName`
// the latter needs to be sanitised to have a ensure a valid package name.
// The `projectDirectoryName` should not have restrictions (e.g. no `_`)
// as a result the package name potentially needs to be altered when derived.
```

packages/create-mc-app/bin/create-mc-app.js
```diff
@@ -0,0 +1,111 @@
...
+fs.writeFileSync(
+  path.join(projectDirectoryPath, 'package.json'),
+  JSON.stringify(updatedAppPackageJson, null, 2) + os.EOL,
+  { encoding: 'utf8' }
+);
+
+// Install the dependencies
+try {
+  execSync(`${useYarn ? 'yarn' : 'npm'} install`);
```
tdeekens: Maybe instead of `useYarn` we can have `getPackageManager` which just returns `yarn` or `npm`? Then we'd just have

```js
const packageManager = getPackageManager();

//...

execSync(`${packageManager} install`);
```

packages/create-mc-app/bin/create-mc-app.js
emmenko: Nothing specific at the moment but I can image having different templates focusing on different parts. We can also have templates for e.g. Typescript, Reason, Parcel, etc.

packages/create-mc-app/bin/create-mc-app.js
emmenko: Sure, I mostly took it from CRA. I'll change it

packages/create-mc-app/bin/create-mc-app.js
tdeekens: Got it.

emmenko:
@montezume I’d like to get your feedback as well before merging this, thanks!

PR #229, opened 2018-12-28
chore(rollup): assign babel dev env when running rollup

### LINKED ISSUE #228: Babel and rollup remove prop-types in production builds
**Describe the bug**

All `packages/*` are being built with Rollup while Babel gets a say in what transforms to apply. Whenever we build for production we remove prop-types from the artefact(s). This implies that no other consumer (always consuming a production build) receives prop-type warnings using any package.

**To Reproduce**

Given you build any or all packages
And you inspect the build output
Then no prop-types are part of any component

**Root cause**

The Babel config [here](https://github.com/commercetools/merchant-center-application-kit/blob/master/packages/babel-preset-mc-app/index.js) is being run in "production mode" which triggers this condition:

```js
isEnvProduction && [
  // Remove PropTypes from production build
  require('babel-plugin-transform-react-remove-prop-types').default,
  {
    removeImport: true,
  },
],
```

to remove prop-types. This is triggered by Rollup builds by e.g. the permissions package with

```json
"build:es": "cross-env NODE_ENV=production rollup -c ../../rollup.config.js  -f es ./index.js -o ./dist/permissions.es.js",
```

The respective Babel config is shared between Rollup (library builds) and Webpack (application builds).

**The crux**

Not run library builds (all packages) in "production mode". The `isEnvProduction` in Babel seems to trigger:

1. Removal or prop-types
2. babel-preset-env building for the n-1 browser versions
   - This may sound like "oh wow, that's a modern target" but includes older IEs I believe

Given we do not want to run Babel again on our built assets when installed in apps (as they're in `node_modules`) we actually want to run those Babel transforms. 

Maybe we want to remove prop-types removal from our default Babel config altogether and apply it additionally in our webpack config [here](https://github.com/commercetools/merchant-center-application-kit/blob/master/packages/mc-scripts/config/create-webpack-config-for-production.js#L247) only for production which is where prop-types should be removed.

**Expected behavior**

Likely prop-types should be part of the build artefact and then should be removed by the respective "application-level" build system. Moreover, prop-types are environment dependent. They're often intended on a staging environment (just for informative reasons) while being suppressed on production.

**Additional context**

This came up during #217 where tests failed with an error being thrown while the error disappeared when a build steps were run before testing. Then tests passed as prop-types are then removed. 

Sorry for the wall but I wanted to document this. /cc @montezume and @emmenko as this seems up your interests.

emmenko:
Fixes #228 

This is a simple proposal to fix the prop-types removal when running rollup builds.


montezume:
Can you check what other libraries do? Since we don't run babel on this code from the MC-FE, wouldn't the prop types be left then in production?

emmenko:
Might be related: https://github.com/facebook/create-react-app/issues/209

What we could do maybe is to use the option `mode: wrap`: https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types#mode

This would ship the bundle with the `process.env.NODE_ENV !== "production"` condition, which then webpack should pick up when building the production application bundle.

What do you think?

montezume:
@emmenko that's exactly what I see in the `downshift` library. If it's good enough for KCD 🤔

emmenko:
Alright, this should work now.

Ref (https://github.com/kentcdodds/kcd-scripts/blob/master/src/config/babelrc.js#L70-L71)

montezume:
Seems to be an open issue with `unsafe-wrap`.
https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types/issues/175
I tried "wrap" on ui-kit and it built with no errors.

packages/babel-preset-mc-app/index.js
```diff
@@ -10,6 +10,7 @@ module.exports = function getBabePresetConfigForMcApp() {
   const isEnvDevelopment = env === 'development';
   const isEnvProduction = env === 'production';
   const isEnvTest = env === 'test';
+  const isRollup = process.env.BUILD_ROLLUP === true;
```
tdeekens: Build tool sniffing feels a bit odd but I can not think of anything else other than solving the issue somewhere entirely different.

packages/babel-preset-mc-app/index.js
```diff
@@ -102,9 +103,10 @@ module.exports = function getBabePresetConfigForMcApp() {
       isEnvProduction && [
         // Remove PropTypes from production build
         require('babel-plugin-transform-react-remove-prop-types').default,
-        {
-          removeImport: true,
-        },
+        // In case of rollup bundles, we want to keep the prop types but wrap
+        // them into a `process.env.NODE_ENV !== "production"` so that when
+        // building the final application bundles, those codes parts can be removed.
+        isRollup ? { mode: 'wrap' } : { removeImport: true },
```
tdeekens: Given that prop-types were never be made available for consumers this is not breaking. Just for completeness, otherwise it would as so,e may have built components by reading prop-types.

PR #240, opened 2019-01-02
test(app-shell): convert LoginLocked test to RTL

emmenko:
image

Snapshots on new year are counterproductive 😅


packages/application-shell/src/components/login-locked/login-locked.spec.js
```diff
@@ -1,46 +1,10 @@
...
-      });
-      it('should match snapshot', () => {
-        expect(wrapper).toMatchSnapshot();
-      });
-    });
-  });
+it('renders maintenance page for locked account', async () => {
+  const { getByText, queryByText } = render(<LockedAccount />);
+  await waitForElement(() => getByText(/Account Locked/i));
+  expect(queryByText(/reset your password/i)).toBeDefined();
```
montezume: nitpick: just wondering why you are using queryByText here? It's mostly only used to test for things not to be there?

how about

   expect(getByText(/reset your password/i)).toBeInTheDocument();

emmenko: As far as I understand, `getByText` throws an error if it does not find the element, therefore it's usually used within a e.g. `waitForElement` call.

The `queryByText` does not throw, so we can use a normal `expect`.

Correct me if I'm wrong though... 🙃

montezume: ah I'm not really familiar with waitForElement. good call 👍

PR #241, opened 2019-01-02
docs(changelog): for release v5.0.0

emmenko:
In preparation for the upcoming `v5.0.0` release.

~~_NOTE: wait for #197 to be merged first_~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,62 @@
+## [5.0.0](https://github.com/commercetools/merchant-center-application-kit/compare/v4.4.0...v5.0.0) (2019-01-03)
+
+## BREAKING CHANGES 💣
+
+We dropped some legacy stuff related to Redux. Please refer to the PR [#197](https://github.com/commercetools/merchant-center-application-kit/pull/197) to get more detailed info and to follow the upgrade migration guide.
+
+Furthermore, we dropped support to the following deprecated things:
+
+- format of permissions as objects (`[{ mode: 'manage', resource: 'products' }]`) from the `@commercetools-frontend/permissions` package
+- components to access user and project data `<FetchUser>` and `<FetchProject>` from the `@commercetools-frontend/application-shell` package
+
+#### 🐛 Type: Bug
+
+- `application-shell`
+  - [#216](https://github.com/commercetools/merchant-center-application-kit/pull/216) Fix test utils and quick access ([@dferber90](https://github.com/dferber90))
+- `application-shell`
+  - [#211](https://github.com/commercetools/merchant-center-application-kit/pull/211) fix(login): reset password dialog redirect, refactor to use formik ([@emmenko](https://github.com/emmenko))
+
+#### 💥 Type: Breaking Change
+
+- `application-shell`, `permissions`, `sdk`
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,48 @@
+## [5.0.0](https://github.com/commercetools/merchant-center-application-kit/compare/v4.4.0...v5.0.0) (2019-01-03)
+
+## BREAKING CHANGES 💣
+
+When using the `@commercetools-frontend/permissions` package, we no longer rely on a static list of permission scopes (e.g. `[permissions.ViewProducts]`), instead we expect the list of `demandedPermissions` to be a **String** (e.g. `['ViewProducts']`). This is necessary as we need more flexibility in defining and using new permissions.
```
emmenko: @tdeekens @qmateub please check if that's correct and/or if I missed something.

qmateub: Looks correct to me 👍

tdeekens: It is correct. Just maybe a bit strangely phrased. TL;DR

1. Instead of hard-coding supported permissions (by constants from the prrmissions package) we now validate required permissions to actually be configured to gain more flexibility
2. The legacy (and long time deprecated) format of permissions with mode and resource is no longer supported

The change is two fold. The fact that we do not validate/require any passed permissions to be part of the constants is not a breaking change (likely) as consumers of app-kit do not set this up and the mc-be still exposes the same permissions. The removal of the legacy format however is. Even though it was introduced before app-kit was made public.

emmenko: Thanks, I'll update the text

emmenko:
Maybe we can also drop the deprecated `FetchUser` and `FetchProject` exports in this release. I'll open up a separate PR.

CHANGELOG.md
```diff
@@ -1,3 +1,59 @@
...
+  - [#216](https://github.com/commercetools/merchant-center-application-kit/pull/216) Fix test utils and quick access ([@dferber90](https://github.com/dferber90))
+- `application-shell`
+  - [#211](https://github.com/commercetools/merchant-center-application-kit/pull/211) fix(login): reset password dialog redirect, refactor to use formik ([@emmenko](https://github.com/emmenko))
+
+#### 💥 Type: Breaking Change
+
+- `application-shell`, `permissions`, `sdk`
+  - [#242](https://github.com/commercetools/merchant-center-application-kit/pull/242) chore: drop deprecated components ([@emmenko](https://github.com/emmenko))
+- `application-shell`, `permissions`
+  - [#217](https://github.com/commercetools/merchant-center-application-kit/pull/217) Refactor `Authorized` in permissions pkg to not rely on constants ([@tdeekens](https://github.com/tdeekens))
```
tdeekens: This is very unlikely breaking but rather a feature. Anyway better safe than sorry.

emmenko: Ah because it had the "breaking change" label...it's also listed in "Refactoring". Should I remove this point from "breaking change"?

tdeekens: Yeah I think it makes sense. Sorry it was two PRs in one. Should have done two. We really should get better at small PRs at times :). Also the lodash refactor could be a separate entry.

emmenko: Yes agree 👌

> Also the lodash refactor could be a separate entry

Ah good point, we can add an entry for that yes.

PR #243, opened 2019-01-02
chore: publish canary releases on CI for master branch

emmenko:
This is more of an idea, as I've been seeing many libraries doing this recently.

The idea is that every time the `master` branch is built on CI, Lerna will publish a new canary version to the `canary` dist channel. The version should look like this: `1.0.0-canary.0+{sha}`.

What are the benefits from doing that?
This makes it easier to test early if changes are properly working, before cutting a normal release.

What do you think? Should we give it a try?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,5 +39,5 @@ script:
   - set +e
 
 after_success:
-  - if [ "$TRAVIS_BRANCH" = "master" ]; then yarn release-canary; fi
+  - if [ "$TRAVIS_BRANCH" = "master" ] && [[ ! "$TRAVIS_COMMIT_MESSAGE" =~ \[skip\ publish\] ]]; then yarn release-canary; fi
   - ./scripts/push_translations.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,4 +39,5 @@ script:
   - set +e
 
 after_success:
+  - if [ "$TRAVIS_BRANCH" = "master" ]; then yarn release-canary; fi
   - ./scripts/push_translations.sh
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -150,6 +150,6 @@ The command will promote the version published on `next` to the `latest` npm dis
 
 On `master` branch, we automatically publish **canary** releases from CI to the `canary` distribution channel, _after_ the build runs successfully.
 
-Canary releases are useful to test early changes that are not yet released. Usually, they run for each Pull Request merged, as we squash the commits into one.
+Canary releases are useful to test early changes that should not be released yet to `next` or `latest`. They are automatically triggered and released after a Pull Request merged into `master`, unless the commit message contains `[skip publish]`.
 
-NOTE that canary releases won't create git tags and version bump commits.
+Note that canary releases __will not create git tags and version bump commits__.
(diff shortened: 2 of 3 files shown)


emmenko:
> Can/should we document this somewhere?

Sure, we can document it if we want to go ahead with this.

> How will it deduct semver versions? When at 2.0.0 and `feat: xxx` is present it goes to `2.1.0-alpha-0`

I'm not 100% sure, I suppose something like `2.1.0-canary.0+{sha}`...we need to try and see what happens.

montezume:
@tdeekens I think it's difficult to enforce a squashing strategy to make something like that work with multiple developers... We looked into using semantic release for ui-kit and decided it was more trouble than benefit.

emmenko:
I think it's safe to give this a try and see how it works (also for understanding more how canary releases are cut, etc). If we see that we don't use it much, we can always drop it.

For example, for ui-kit (in case we also set it up there) it can be useful to use new components that are not being released yet. We had the situation several times where we had to "copy-paste" things because they were not available in ui-kit yet.

tdeekens:
Yes. I agree. Last question: [does this](https://github.com/commercetools/merchant-center-application-kit/blob/master/package.json#L22) promote a canary release on next to latest when I run it? I likely do not want it but only to refer to stable releases (not canary on next).

emmenko:
The promotion only works for next tags to latest. The canary releases do not affect that, which is fine I think.

tdeekens:
> The promotion only works for next tags to latest. The canary releases do not affect that, which is fine I think.

Ah got it. It's on another dist-tag.

emmenko:
Alright, I'll add some documentation around this.

CONTRIBUTING.md
```diff
@@ -136,10 +138,18 @@ All releases go the the `next` distribution channel. This gives us a chance to t
...
 ```
 
-The command will promote the version published on `next` to the `latest` dist-tag for each package.
+The command will promote the version published on `next` to the `latest` npm dist-tag, for each package.
+
+## Canary releases
+
+On `master` branch, we automatically publish **canary** releases from CI to the `canary` distribution channel, _after_ the build runs successfully.
+
+Canary releases are useful to test early changes that are not yet released. Usually, they run for each Pull Request merged, as we squash the commits into one.
```
tdeekens:
```suggestion
 Canary releases are useful to test early changes that should not yet be released to `next` or `latest`. They are automatically triggered and released after a pull request merged into `master`. Please prefer squash merging for a clean commit history on `master`.
```

CONTRIBUTING.md
```diff
@@ -136,10 +138,18 @@ All releases go the the `next` distribution channel. This gives us a chance to t
...
+
+NOTE that canary releases won't create git tags and version bump commits.
```
tdeekens:
```suggestion
 Note that canary releases __will not create git tags and version bump commits__.
```

.travis.yml
```diff
@@ -39,4 +39,5 @@ script:
   - set +e
 
 after_success:
+  - if [ "$TRAVIS_BRANCH" = "master" ] && [[ ! "$TRAVIS_COMMIT_MESSAGE" =~ \[skip\ publish\] ]]; then yarn release-canary; fi
```
emmenko: I tested this locally:

image

emmenko:
> Should we make squash merging the only option on app-kit with this?

It's always been like this 😉

emmenko:
Are we good to go with this then?

tdeekens:
Yup. Lets give it a shot.

emmenko:
Cool, I'll do the bug fix release first and then merge this.

PR #244, opened 2019-01-03
chore: include create-mc-app templates in yarn workspaces

emmenko:
We include the template packages in yarn workspaces so that lerna can apply the version changes when we do a new release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
   },
   "workspaces": [
     "packages/*",
+    "packages/create-mc-app/templates/*",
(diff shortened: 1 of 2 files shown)


emmenko:
Or can we maybe define this as a workspace project? 🤔
I’ll try that later then

emmenko:
Ok that seems to do the trick. If I try to publish, the versions of the packages in the template starter get updated as well (but not published because it's a private package).

screenshot 2019-01-03 at 16 09 49

PR #245, opened 2019-01-03
refactor(application-shell): permission prop-type form navbar

tdeekens:
#### Summary

This removes the constant based prop-type testing from the navbar in the application shell. 

#### Description

The prop-type check should be performed by `RestrictedByPermission` (and is) based on the configured permissions. Not at any given layer of the component tree which just forwards props and pretends to have knowledge and/or a say in what permissions it should allow forwarding without even doing anything with them.

### DIFF
--- a/packages/application-shell/src/components/navbar/navbar.js
+++ b/packages/application-shell/src/components/navbar/navbar.js
@@ -32,10 +32,7 @@ import {
   GRAPHQL_TARGETS,
   NO_VALUE_FALLBACK,
 } from '@commercetools-frontend/constants';
-import {
-  RestrictedByPermissions,
-  permissions,
-} from '@commercetools-frontend/permissions';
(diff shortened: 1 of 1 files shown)


packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -325,9 +320,7 @@ export class DataMenu extends React.PureComponent {
             ),
             uriPath: PropTypes.string.isRequired,
             featureToggle: PropTypes.string,
-            permissions: PropTypes.arrayOf(
-              PropTypes.oneOf(Object.keys(permissions))
-            ),
```
tdeekens: This is the same prop-type now elsewhere until it his `RestrictedByPermission` which then gives a prop-type warning given the permission is not configured with a new component tree in the stack.

PR #247, opened 2019-01-04
fix(rollup): rename lodash import to lodash-es only for es bundle

emmenko:
We need to run the babel plugin to transform the import `lodash` to `lodash-es` only when bundling ES modules, not CJS 🙈

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -24,74 +24,81 @@ const postcssReporter = require('postcss-reporter');
 const { pkg } = readPkgUp.sync({
   cwd: fs.realpathSync(process.cwd()),
 });
-const importReplacements = [{ original: 'lodash', replacement: 'lodash-es' }];
 const babelOptions = getBabelPreset();
 
-const config = {
-  output: {
-    name: pkg.name,
-    sourcemap: true,
-  },
-  plugins: [
-    peerDeps({
-      includeDependencies: true,
-    }),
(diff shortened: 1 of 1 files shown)


rollup.config.js
```diff
@@ -27,71 +27,77 @@ const { pkg } = readPkgUp.sync({
...
+      peerDeps({
+        includeDependencies: true,
+      }),
+      babel({
+        runtimeHelpers: true,
+        ...babelOptions,
+        plugins: [
+          babelPluginImportGraphQL.default,
+          ...babelOptions.plugins,
+          isFormatEs && [
```
emmenko: This is the fix

rollup.config.js
```diff
@@ -27,71 +27,77 @@ const { pkg } = readPkgUp.sync({
...
-            '@commercetools-frontend/ui-kit/materials/custom-properties.css'
-          ),
-        }),
-        postcssColorModFunction(),
-        postcssReporter(),
-      ],
-    }),
-  ],
+const createConfig = cliArgs => {
+  const isFormatEs = cliArgs.format === 'es';
```
emmenko: To access the CLI args, we need to convert the config into a function

rollup.config.js
montezume: ah this makes sense, since we aren't bundling deps 👍

PR #251, opened 2019-01-04
Add `menuVisibilities` and `allAppliedMenuVisibilities`

tdeekens:
#### Summary

This pull request adds the notion of visibility overwrites to the menu.

#### Description

Visibility overwrites are a concept in the broader topic of permissions (more detail on internal PRs). A visibility overwrite currently targets the menu and provides the ability to hide any menu items independent of permissions. 

In this PR the following is added:

1. `allAppliedVisibilityOverwrites` are being fetched on the project query through the mc-be (not yet merged there)
2. `allAppliedVisibilityOverwrites` are mapped into the same shape is permissions and flushed through the application context as `visibilityOverwrites`
3. The `<NavBar />` `config.js` and rendering logic is adjusted to take in visibility overwrites
   - Any menu item (in `config.js`) can have a `visibilityOverwrite` which is then being checked
   - Given all submenus of a main menu are hidden then the main menu is automatically being hidden

Please have a first look at the implementation and implications of this and give some feedback.


packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -247,39 +248,55 @@ export const MenuItemDivider = () => (
...
-    <ToggleFeature flag={props.featureToggle}>
-      {permissionsWrapper}
-    </ToggleFeature>
-  ) : (
-    permissionsWrapper
-  );
-};
-ToggledWithPermissions.displayName = 'ToggledWithPermissions';
-ToggledWithPermissions.propTypes = {
+export const RestrictedMenuItem = props => (
```
tdeekens: I renamed this but this whole `<NavBar />` starts getting a bit worried. I didn't dare to touch too much as it looks like it would fall apart. I think it's a prime candidate to refactoring from code and to use rtl but I am not sure how much design would like to change things too. Hence I just keep things as is.

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -247,39 +248,55 @@ export const MenuItemDivider = () => (
...
+      // NOTE: Custom application are activated/deactivated while their
+      // visibility is not controlled via a visibiility overwrite.
+      if (
+        props.visibilityOverwrites &&
+        props.visibilityOverwrites.every(
+          visibilityOverwrite =>
+            configuredVisibilityOverwrites[visibilityOverwrite] === true
+        )
+      )
+        return null;
```
tdeekens: This logic will eventually move into a pkg in another PR depending on what else needs to be hidden by visibility overwrites. It is plural as the main menu aggregates all overwrites of all its submenus and if all are true we hide. Any submenu just has one overwrite (see below). A submenus overwrite can be hardcoded in the config or later be CoC where it's "just" the key which is helpful for e.g. custom apps.

packages/application-shell/src/test-utils/test-utils.js
```diff
@@ -107,6 +108,7 @@ const render = (
     user,
     project,
     permissions = defaultPermissions,
+    visibilityOverwrites = defaultVisibilityOverwrites,
```
tdeekens: Just so we can write rtl tests with them later in the app.

packages/application-shell/src/components/navbar/navbar.js
qmateub: That's a good point. If you take a look at the whole file (at least for me) is pretty hard to follow because there is something that points to the top of the file another thing that uses stuff from the bottom so ...

I become a super fan of declarative stuff and here gives the feeling of us trying to automate the rendering somehow. This component is "finite" at some point. I got for example that the custom application is more dynamic but the rest I don't think so.

```
Menu
 MenuEntry (products)
   MenuItem (add product)
   MenuItem (list)
 MenuEntry (customers)
   MenuItem (add customer)
````

Surely worths a refactor 😋

Comment (emmenko) [reply packages/application-shell/src/components/navbar/navbar.js]: A rewrite of the navbar is already planned, I just need to finalise some things with Sven first before we start the refactoring.

**Comment by tdeekens** [comment]
This now no longer depends on visibility overwrites but menu visibilities. Further more values are (yet) not flushed down the app context.

File: packages/application-shell/src/components/navbar/navbar.js [hunk 4]
```diff
@@ -445,11 +460,17 @@ export class DataMenu extends React.PureComponent {
   renderMenu = (menu, menuType, index) => {
     const isActive = this.state.activeItemIndex === `${menuType}-${index}`;
     const hasSubmenu = Boolean(menu.submenu) && menu.submenu.length > 0;
+    const namesOfMenuVisibilitiesOfAllSubmenus = hasSubmenu
+      ? menu.submenu.map(submenu => submenu.menuVisibility).filter(Boolean)
```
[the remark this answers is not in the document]
tdeekens: Done in d68932e

tdeekens:
I will leave this as WIP until I release the API and tested it against staging.

tdeekens:
@qmateub can you review this? It's only pending the release of uikit@6 in the mc-fe cause app-kit now already is on v6.

tdeekens:
Tested around a bit. Will merge tomorrow.

PR #252, opened 2019-01-04
fix(app-shell): do not render GtmContext.Provider, so default value is used

emmenko:
Caused by #197 

This was a tricky one 🧐


emmenko:
> I honestly didn't know that. 

Me neither

> That then applies to many other context providers too?

I think this is the only one where we simply want to use the default values.

---

I pushed a commit to include the context in the test-utils https://github.com/commercetools/merchant-center-application-kit/pull/252/commits/39e4d805897b76ce49326151b864e6b4c8fce08a

PR #264, opened 2019-01-07
fix(create-mc-app): publish application templates as separate packages

emmenko:
#### Summary

In the `create-mc-app` we currently have the app templates included in the package. 
In #244 we defined the `templates` folder to be used as a workspace, so that lerna can update the dependency versions of the template app when we do releases.

However, the `create-mc-app` might not get updated if none of its dependencies are updated, in which case installing the template will result in outdated dependencies.

#### Description

So to amend this problem, we can move the templates into separate workspaces (outside of the `create-mc-app` package) and require the templates as normal dependencies, so that when we release, the package versions are correctly updated.


application-templates/starter/package.json
```diff
@@ -1,8 +1,23 @@
 {
-  "name": "merchant-center-application-template-starter",
-  "version": "5.0.3",
+  "name": "@commercetools-frontend/application-template-starter",
+  "version": "0.0.0",
```
emmenko: This will be correctly bumped once we do the next release

packages/create-mc-app/bin/create-mc-app.js
```diff
@@ -106,7 +112,7 @@ fs.writeFileSync(
 
 // Install the dependencies
 try {
-  execSync(`${packageManager} install`);
+  execSync(`${packageManager} install`, { stdio: 'inherit' });
```
emmenko: This allows to see the output of the command when the child process executes.

packages/create-mc-app/bin/create-mc-app.js
```diff
@@ -71,12 +71,18 @@ const packageManager = useYarn ? 'yarn' : 'npm';
 // TODO: we could check for min yarn/npm versions
 // See https://github.com/facebook/create-react-app/blob/0f4781e8507249ce29a9ac1409fece67c1a53c38/packages/create-react-app/createReactApp.js#L225-L254
 
-const templatePath = path.join(__dirname, '../templates', templateName);
+const templatePackageName = `@commercetools-frontend/application-template-${templateName}`;
+const modulePathForPackage = require.resolve
+  .paths(templatePackageName)
```
emmenko: The `require.resolve.paths` returns a list of location of `node_modules` that are checked by the node module resolution. For example:

image

packages/create-mc-app/bin/create-mc-app.js
```diff
@@ -71,12 +71,18 @@ const packageManager = useYarn ? 'yarn' : 'npm';
...
+  .find(modulesPath =>
+    fs.existsSync(path.join(modulesPath, templatePackageName))
```
emmenko: We only want to know in which `node_modules` folder the template package got installed.

If we were to use `require.resolve('@commercetools-frontend/application-template-starter')` it will fail because the function resolves the path to the module **entry point**. However, we are only interested in the **location** of the package.

package.json
```diff
@@ -30,8 +30,8 @@
     "build:es:watch": "NODE_ENV=development lerna run --no-private --parallel build:es:watch"
   },
   "workspaces": [
+    "application-templates/*",
```
tdeekens: Quick q: why not `packages/application-templates` instead of opening a whole new workspace?

emmenko: I find it weird to mix it with "normal" packages

tdeekens: I find it weird to make a whole new workspace just for one package. That makes me feel that this is actually something which does not really belong into the repo after all :).

emmenko: Why not? What is the problem of having different workspaces? It's just a way or organising things. The template is not really a library package but an app. We do the same in our apps repo.

The big advantage in having the template in the monorepo is that dep versions are automatically updated as we make changes. If we move it to a separate repo we need to manage those deps manually.

emmenko:
Actually I just realised something: the template package now is treated as a normal node module, so its deps are also installed when you install `create-mc-app`. I guess it's not too bad but also kind of a weird side effect 🙃

emmenko:
Let me think a bit further if there are other solutions... 🤔

tdeekens:
> Actually I just realised something: the template package now is treated as a normal node module, so its deps are also installed when you install create-mc-app. I guess it's not too bad but also kind of a weird side effect 🙃

I don't think there is much we can do about that cause that's how a package in any workspace is handled?

emmenko:
I opened another PR with a different approach: #265

emmenko:
Closing in favour of #265

PR #265, opened 2019-01-07
refactor(create-mc-app): how to install template

emmenko:
Alternative approach for #264 

#### Summary

In the `create-mc-app` we currently have the app templates included in the package. 
In #244 we defined the `templates` folder to be used as a workspace, so that lerna can update the dependency versions of the template app when we do releases.

However, the `create-mc-app` might not get updated if none of its dependencies are updated, in which case installing the template will result in outdated dependencies.

#### Description

To amend for this problem, we can move the templates into separate workspaces (outside of the `create-mc-app` package).
This way, we ensure that the templates get their dependencies updated when we create releases.
We also decouple the templates from the `create-mc-app` package, so that we don't need to constantly update it.

The `create-mc-app` script now accepts an optional `--template-version` argument where you can specify a specific version of a template that you wish to install.
The value can be a normal semver version or a git branch, as it will be used for git cloning.
Example of valid values:

```
5.0.0
v5.0.0
5.0.0-beta.0
master
my-branch-name
```

What happens then is that we _shallow clone_ the repo for the specific "version" and we attempt to copy the template into the folder name passed to the `create-mc-app` script.
The repo gets cloned into the "OS" tmp folders, so it gets cleaned up by the system.


.travis.yml
```diff
@@ -33,9 +33,10 @@ before_install:
 
 script:
   - set -e
-  - yarn run jest --projects jest.*.config.js --maxWorkers=4 --reporters jest-silent-reporter
-  - yarn build
-  - yarn playground:build
+  - ./scripts/install_app_templates.sh
+  # - yarn run jest --projects jest.*.config.js --maxWorkers=4 --reporters jest-silent-reporter
```
emmenko: Temporary disabled for testing the installation. Will change later

packages/create-mc-app/bin/create-mc-app.js
```diff
@@ -1,122 +1,17 @@
...
-);
-
-// Install the dependencies
 try {
-  execSync(`${packageManager} install`);
+  const options = steps.parseArguments();
+  steps.downloadTemplate(options);
+  steps.updatePackageInfo(options);
+  steps.installDependencies(options);
+  steps.success(options);
```
emmenko: Now it should be easier to follow the steps

packages/create-mc-app/src/steps/download-template.js
```diff
@@ -0,0 +1,64 @@
...
+  );
+
+  throwIfTemplateVersionDoesNotExist(
+    templateName,
+    clonedRepositoryPath,
+    tagOrBranchVersion
+  );
+
+  try {
+    execSync(`mv ${templateFolderPath} ${projectDirectoryPath}`);
```
emmenko: FYI: this changed to `mv` (before it was `cp -R`)

packages/create-mc-app/src/validations.js
```diff
@@ -0,0 +1,56 @@
+const fs = require('fs');
```
tdeekens: I really enjoy the split by concern. Validation, steps... reallx vlear 👌

package.json
```diff
@@ -30,8 +30,8 @@
     "build:es:watch": "NODE_ENV=development lerna run --no-private --parallel build:es:watch"
   },
   "workspaces": [
+    "application-templates/*",
```
tdeekens: It is still a pretty bold assumption of ours that we will ever have more. But it would be awesome 😁

packages/create-mc-app/src/steps/download-template.js
```diff
@@ -0,0 +1,64 @@
...
+module.exports = function downloadTemplateFromRemoteRepository({
+  projectDirectoryPath,
+  templateName,
+  tagOrBranchVersion,
+}) {
+  console.log(
+    `==> Downloading template "${templateName}" for version "${tagOrBranchVersion}"\n`
+  );
+
+  const tmpDir = os.tmpdir();
```
tdeekens: 👌

.travis.yml
```diff
@@ -36,6 +36,7 @@ script:
   - yarn run jest --projects jest.*.config.js --maxWorkers=4 --reporters jest-silent-reporter
   - yarn build
   - yarn playground:build
+  - ./scripts/install_app_templates.sh
```
tdeekens: Nice quick test.

packages/create-mc-app/src/steps/success.js
```diff
@@ -0,0 +1,17 @@
...
+  console.log(`Success! 🎉`);
+  console.log(
+    `The Merchant Center application ${projectDirectoryName} was successfully bootstrapped based on the ${templateName} template.`
+  );
+  console.log();
+  console.log(`To get started:`);
+  console.log(`  $ cd ${projectDirectoryName}`);
+  console.log(`  $ ${shouldUseYarn ? 'yarn' : 'npm'} start`);
+  console.log();
+  console.log(`For more info, have a look at the README. Enjoy 🚀`);
```
tdeekens: Nice effort to get the output looking sharp. I always wonder if something like this is fun to play with e.g. signale
https://github.com/klaussinani/signale.

package.json
emmenko: I plan to 😜

packages/create-mc-app/src/steps/success.js
emmenko: Yeah I also wanted to look into something like this. I didn't know that library, I'll check it out thanks! 💯

emmenko:
Alright, I'll merge the loggers improvements and then we can ship this! Thanks a lot for the feedback! 💯

PR #266, opened 2019-01-07
feat(create-mc-app): use fancy logger

emmenko:
Branch experiment for fancy loggers using https://github.com/klaussinani/signale

screenshot 2019-01-07 at 22 42 26

### DIFF
--- a/packages/create-mc-app/src/logger.js
+++ b/packages/create-mc-app/src/logger.js
@@ -1,9 +1,13 @@
-const signale = require('signale');
+const { Signale } = require('signale');
 
-signale.config({
-  displayBadge: false,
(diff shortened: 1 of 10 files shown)


packages/create-mc-app/src/logger.js
```diff
@@ -1,9 +1,13 @@
-const signale = require('signale');
+const { Signale } = require('signale');
 
-signale.config({
-  displayBadge: false,
+module.exports = new Signale({
+  config: { displayBadge: false, displayLabel: false },
```
emmenko: I find it a bit more clean without these, as we already have the "scope" next to the log

image

PR #269, opened 2019-01-08
chore: update ui-kit to v6

montezume:
Update UI-Kit to [v6.0.0](https://www.npmjs.com/package/@commercetools-frontend/ui-kit/v/6.0.0)

* Delete PasswordField
* Fix tokens import
* Moved SubHeadline to use h4 instead of h3 (as h3 isn't supposed to be an option for that)


packages/application-shell/src/components/login-sso/__snapshots__/login-sso.spec.js.snap
```diff
@@ -9,7 +9,7 @@ exports[`rendering <Formik> form elements should match layout 1`] = `
     scale="m"
   >
     <TextSubheadline
-      elementType="h3"
+      elementType="h4"
```
montezume: h3 isn't supposed to be part of subheadline

montezume:
It's nice to have proptype warnings again 💯

tdeekens:
> It's nice to have proptype warnings again 💯

That was really a bad hiccup :)

emmenko:
We can merge this here, there is no problem with that. We could also do a new release here (but I would wait for that).

PR #270, opened 2019-01-08
fix: template starter missing dependency

emmenko:
I used the canary build to try the new installation approach of the template for `create-mc-app`.
The template is correctly installed but the app does not start because of the missing `redux` dependency.

So now I changed the installation script on CI to install the app outside of the app-kit repo (to avoid using hoisted dependencies) and we also run tests and build, just to make sure that everything works out of the box.


emmenko:
Works!

image

PR #272, opened 2019-01-10
feat(babel-preset-mc-app): disable console on production

tdeekens:
#### Summary

This pull request adds a babel plugin to remove console.* statements on production.

#### Description

Given we are building for production we can remove console log statements. It's not strictly needed but also can not hurt much.

There are other alternatives to do this:

1. During uglification
    - Decided against it as it create a lock-in to a minifier
2. Overwriting `console.log = () => {}`
    - It's not really a good practice to overwrite built-ins
    - It also affects code which is not "ours"

Given that I went with a simple babel plugin applied only when building for prod. Moreover the babel plugin allows only overwriting specific loggers (so if needed later we can keep `error` again etc).

### DIFF
--- a/packages/mc-scripts/config/create-webpack-config-for-production.js
+++ b/packages/mc-scripts/config/create-webpack-config-for-production.js
@@ -401,7 +401,7 @@ module.exports = ({
           },
         ],
       },
-      // Process JS with Babel.
+      // Process application JavaScript with Babel.
       {
         test: /\.js$/,
         use: [
@@ -425,6 +425,37 @@ module.exports = ({
           },
         ],
         include: sourceFolders,
+      },
+      /**
+       * NOTE:
+       *    Some dependencies may use `console.*` to log (e.g. Apollo). These log statements
+       *    should be removed for production builds. This could also be achieved using `UglifyJS`.
+       *    However, the fact that also `prop-types` (from dependencies in `node_modules`)
+       *    should be stripped from production builds requires a separate configuration
+       *    for the `babel-loader` including files from `node_modules` while removing
+       *    the mentioned `prop-types` and console statements.
+       */
+      {
+        test: /\.js$/,
+        use: [
+          require.resolve('thread-loader'),
+          {
+            loader: require.resolve('babel-loader'),
+            options: {
+              babelrc: false,
+              plugins: [
+                require.resolve('babel-plugin-transform-remove-console'),
+                [
+                  require.resolve(
+                    'babel-plugin-transform-react-remove-prop-types'
+                  ),
+                  { removeImport: true },
+                ],
+              ],
+            },
+          },
+        ],
+        include: /node_modules/,
       },
       // Allow to import `*.graphql` SDL files.
       {
--- a/packages/mc-scripts/config/create-webpack-config-for-production.js
+++ b/packages/mc-scripts/config/create-webpack-config-for-production.js
@@ -96,19 +96,8 @@ const defaultToggleFlags = {
  * The function requires the file path to the related application
  * "entry point".
  */
-module.exports = ({
-  distPath,
-  entryPoint,
-  sourceFolders,
-  toggleFlags = defaultToggleFlags,
-}) => {
-  // NOTE: Without properly defauling passing one flag (e.g. `enabledVendorOptimizations`)
-  // would disable other flags defaulted to `true` (e.g. `enableExtractCss`).
-  const defaultedToggleFlags = {
-    ...defaultToggleFlags,
-    ...toggleFlags,
-  };
-
+module.exports = ({ distPath, entryPoint, sourceFolders, toggleFlags }) => {
+  const mergedToggleFlags = { ...defaultToggleFlags, ...toggleFlags };
   return {
     // Don't attempt to continue if there are any errors.
     bail: true,
@@ -126,7 +115,7 @@ module.exports = ({
     optimization: {
       minimizer: [
         new UglifyJsPlugin(uglifyConfig),
-        defaultedToggleFlags.enableExtractCss &&
+        mergedToggleFlags.enableExtractCss &&
           new OptimizeCSSAssetsPlugin(optimizeCSSConfig),
       ].filter(Boolean),
       // Automatically split vendor and commons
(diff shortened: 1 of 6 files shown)


montezume:
You can use the `isRollup` variable that @emmenko introduced

tdeekens:
> You can use the isRollup variable that @emmenko introduced

True. Did that in the last commit.

montezume:
I would like to try it locally first

tdeekens:
> I pulled the branch, ran yarn build to build the libraries, then yarn build in playground, and started the playground in production mode, and the console logs are still there from LoggedInUserQuery, etc.

I struggle to follow the sentiment of this comment as I assume you know why we see those: they're coming from libs in `node_modules`. Are you suggesting to just use the minifier, correct?

montezume:
Okay, then maybe I misunderstand what we are trying to achieve with this PR. I thought the idea was to disable all console logs in production. If the idea is to only disable console logs coming from `application-products`, or whichever application is running, then yeah, this achieves that. But, it doesn't remove console logs from our other apps.

montezume:
Thanks for bringing this up though, I noticed that for the same reason as we are having trouble with this, we aren't removing prop-types from ours (or external) libraries!

tdeekens:
I tried to achieve first that log-statements from our app's are gone. So e.g. your sdk middleware. The fact that other libraries still log is a bit of a different concern to me. However, in 677b9f1 I added dropping log statements to uglify-js. This should get rid of most. However, I'd like to keep things in both places given we move to another uglifier we still retain control over our application level log statements.

tdeekens:
> Okay, I'm fine with merging this and then we can look into the separate babel config for vendors later.

I gave it a shot in 4b1751b. This is roughly what we would aim to achieve, right?

tdeekens:
@montezume sourceFolders are now excluded (even though only `node_modules` were included before).

packages/mc-scripts/config/create-webpack-config-for-production.js
```diff
@@ -437,6 +437,8 @@ module.exports = ({
        */
       {
         test: /\.js$/,
+        include: /node_modules/,
+        exclude: sourceFolders,
```
montezume: maybe we should include this in our docs. what would happen if a user passed "node_modules" as part of their sourceFolders 🤔

montezume: hm, maybe we can just document it. or provide a way to override this behaviour. maybe "optimizeVendorJavascript" can be a setting?

tdeekens: Sure. I can do that tomorrow.

tdeekens:
Alright, in the last commits 

1. I added a `enabledVendorOptimizations` flag we can pass to out webpack config
2. I simplified the concatenation a bit (to just `filter(Boolean)` cause it made head go 💥
3. I more safely defaulted toggles
   - This is a bit of a "classic" with default arguments and option objects. When having `{ a: true, b: false }` and defaulting to them. Then passing ` { b: true }` you set `a` silently to `undefined` and hence `false`. That's rarely what you want.

@emmenko and @montezume please have another look. Cheers!

packages/mc-scripts/config/create-webpack-config-for-production.js
```diff
@@ -100,91 +101,84 @@ module.exports = ({
   entryPoint,
   sourceFolders,
   toggleFlags = defaultToggleFlags,
-}) => ({
-  // Don't attempt to continue if there are any errors.
-  bail: true,
+}) => {
+  // NOTE: Without properly defauling passing one flag (e.g. `enabledVendorOptimizations`)
+  // would disable other flags defaulted to `true` (e.g. `enableExtractCss`).
+  const defaultedToggleFlags = {
```
emmenko: Nitpick: `default` and `defaulted` are very similar and it's easier to mistype them. I got myself confused for a couple of seconds. For example, in `test-utils` we name those variables `mergedXxx`. So maybe `mergedToggleFlags`?

packages/mc-scripts/config/create-webpack-config-for-production.js
```diff
@@ -100,91 +101,84 @@ module.exports = ({
...
+    ...defaultToggleFlags,
+    ...toggleFlags,
+  };
```
emmenko: I think we can simplify this like:

```js
const mergedToggleFlags = Object.assign({}, defaultToggleFlags, toggleFlags);
```

...and remove the default named parameter "toggleFlags ~~= defaultToggleFlags~~,"

...and we don't need the comment anymore.

packages/mc-scripts/config/create-webpack-config-for-production.js
tdeekens: Sure. I don't mind. It's just how I name them generally so I am used to the naming. Will change.

packages/mc-scripts/config/create-webpack-config-for-production.js
tdeekens: Prettier will format it to spreading syntax generally. Will see what happens if the arg is not defaulted to an object.

tdeekens:
@emmenko please check 35402cf

packages/mc-scripts/config/create-webpack-config-for-production.js
emmenko: Naming is hard 😇

tdeekens: It's sometimes the only fun thing left in programming 😄

packages/mc-scripts/config/create-webpack-config-for-production.js
```diff
@@ -208,248 +191,283 @@ module.exports = ({
...
+                // directory for faster rebuilds.
+                cacheDirectory: true,
+                highlightCode: true,
+              },
+            },
+          ],
+          include: sourceFolders,
+        },
+        /**
+         * NOTE:
```
montezume: 💯

montezume:
Looks great now, thanks

emmenko:
@tdeekens would you like to merge and update the changelog PR?

PR #274, opened 2019-01-10
chore(app-shell): remove unused components 🚮

emmenko:


tdeekens:
Should also hit the release I think.

PR #275, opened 2019-01-12
fix(mc-scripts): add dynamic import plugin to fix vendor optimizations

tdeekens:
#### Summary

This pull request adds the dynamic import plugin to the list of plugins applied in the "vendor optimizations" run. Without it fails as dynamic imports occur in our own app-shell.

I tested this by npm linking in the package and we can build apps again as a result. Given that the previous version (even though not on latest) is broken I will go ahead and merge/release this to unblock the mc-fe.


packages/mc-scripts/package.json
```diff
@@ -64,9 +65,9 @@
     "thread-loader": "2.1.1",
     "uglifyjs-webpack-plugin": "1.2.7",
     "url-loader": "1.1.2",
-    "webpack": "4.28.3",
     "webpack-bundle-analyzer": "3.0.3",
     "webpack-dev-server": "3.1.14",
+    "webpack": "4.28.3",
     "webpackbar": "3.1.4"
```
tdeekens: Just ordered them.

packages/mc-scripts/config/create-webpack-config-for-production.js
```diff
@@ -431,6 +432,7 @@ module.exports = ({ distPath, entryPoint, sourceFolders, toggleFlags }) => {
               options: {
                 babelrc: false,
                 plugins: [
+                  require.resolve('@babel/plugin-syntax-dynamic-import'),
```
tdeekens: The fix.

PR #277, opened 2019-01-13
chore(eslint-config-mc-app): use only peer deps

emmenko:
Our `eslint-config-mc-app` had hardcoded dependencies, including e.g. `eslint`, which made it confusing to update. For instance, having different versions of `eslint` installed, causes it to break.

Instead, the config should only define **peerDependencies**!

There were also a couple of dependencies not used by the config.


packages/eslint-config-mc-app/package.json
```diff
@@ -20,22 +20,18 @@
...
   "main": "./index.js",
   "module": "./index.js",
-  "dependencies": {
-    "babel-eslint": "10.0.1",
-    "eslint": "5.12.0",
-    "eslint-config-airbnb-base": "13.1.0",
-    "eslint-config-prettier": "3.3.0",
-    "eslint-formatter-pretty": "2.0.0",
-    "eslint-plugin-babel": "5.3.0",
-    "eslint-plugin-cypress": "2.2.0",
```
emmenko: This wasn't used by the config

packages/eslint-config-mc-app/package.json
```diff
@@ -20,22 +20,18 @@
   },
   "main": "./index.js",
   "module": "./index.js",
-  "dependencies": {
-    "babel-eslint": "10.0.1",
-    "eslint": "5.12.0",
-    "eslint-config-airbnb-base": "13.1.0",
-    "eslint-config-prettier": "3.3.0",
-    "eslint-formatter-pretty": "2.0.0",
```
emmenko: This is only used by the jest runner

packages/eslint-config-mc-app/package.json
```diff
@@ -20,22 +20,18 @@
...
-    "eslint-plugin-jest": "22.1.2",
-    "eslint-plugin-jsx-a11y": "6.1.2",
-    "eslint-plugin-prefer-object-spread": "1.2.1",
-    "eslint-plugin-prettier": "3.0.1",
-    "eslint-plugin-react": "7.12.3"
-  },
+  "dependencies": {},
   "peerDependencies": {
-    "eslint": ">= 5.3"
+    "babel-eslint": "10.x",
```
emmenko: I found this way to determine a version range, I think it's simpler.

screenshot 2019-01-13 at 16 53 55

emmenko: We can change the other places where we define peer deps to use the same format.

tdeekens:
Will merge to get a canary build (to test in app) and to draft the changelog. Hope that's fine.

PR #279, opened 2019-01-13
docs(changelog): for v6.0.0 [skip ci]

tdeekens:


emmenko:
Actually it's fine, I got confused. I think we can go ahead with this.

PR #286, opened 2019-01-14
refactor: disable loggers based on runtime configuration

emmenko:
Follow up of #272 

I've been giving it some more thoughts and I think we can solve this problem a bit differently, which is why I'm proposing this new approach.

In general, if we think about it, libraries usually don't log anything. So there is no need to run the babel transform in every node module.
On the other hand, the AppShell is like a mix of library and application code, which is why it logs some stuff (apollo, redux).

At the end, we want to disable logging on production BUT **when we need it, we should be able to see those logs**.

Therefore, we can these applications logs are enabled based on some runtime configuration:
- `process.env.DEBUG === 'true'`
- app is built for development (not test, not production)
- if the app is built for production and there is a `debug=true` query parameter in the URL

I like the fact that we can also enable this on runtime via a query parameter. This is useful for example when working on support tickets with the customer, to temporary enable debug logs.

Let me know what you think.

### DIFF
--- a/playground/package.json
+++ b/playground/package.json
@@ -6,6 +6,7 @@
   "scripts": {
     "build": "mc-scripts build",
     "start": "mc-scripts start",
+    "start:prod:local": "NODE_ENV=production mc-http-server --csp=$(pwd)/csp.json --config=$(pwd)/env.json --use-local-assets",
     "i18n:build": "mc-scripts extract-intl --output-path=$(pwd)/src/i18n/data 'src/**/!(*.spec).js' --build-translations",
     "deploy": "npm run build; rm dist/assets/*.js.map; npm run deploy:eu; npm run deploy:us",
     "deploy:eu": "now --local-config=now.eu.json --docker; now alias --local-config=now.eu.json",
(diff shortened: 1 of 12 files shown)


montezume:
```
In general, if we think about it, libraries usually don't log anything. So there is no need to run the babel transform in every node module.
```

I'm not sure how this relates to this PR, since it doesn't remove running babel on every node_module.

emmenko:
> since it doesn't remove running babel on every node_module.

There is still prop types transform, which I didn't touch, hence the loader is still there. But we can also remove that as well.

emmenko:
In the last commit (https://github.com/commercetools/merchant-center-application-kit/pull/286/commits/d628d5320c4074427847cebb972f855d23470745) I dropped support for the `enabledVendorOptimizations`.

playground/package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "build": "mc-scripts build",
     "start": "mc-scripts start",
+    "start:prod:local": "NODE_ENV=production mc-http-server --csp=$(pwd)/csp.json --config=$(pwd)/env.json --use-local-assets",
```
montezume: thanks for adding this!

tdeekens:
I guess the fact that we're still on `@next` means we don't need to be a breaking release again cause the breaking change never made it to `@latest`.

PR #288, opened 2019-01-14
chore: make pegjs an optional dependency

emmenko:
This has been bothered me for a while.

The `pegjs` dependency is only used by the discounts app, therefore it should not be a default of webpack/jest.

We can define it as an `optionalDependencies`. The webpack loader is only defined then if the module can be found.

### DIFF
--- a/packages/jest-preset-mc-app/jest-preset.js
+++ b/packages/jest-preset-mc-app/jest-preset.js
@@ -36,6 +36,7 @@ module.exports = {
   transform: {
     '^.+\\.js$': resolveRelativePath('./transform-babel-jest.js'),
     '^.+\\.graphql$': 'jest-transform-graphql',
+    // NOTE: this is an optional dependency, we can keep the transform config here though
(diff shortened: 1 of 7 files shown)


emmenko:
I thought about that and there are different approaches to that. And we can easily complicate things a lot more.

I definitely want to look into that but I was planning to do that another time. This is a simple fix that I think it's good enough for now.

emmenko:
By bothered I mean seeing the peer deps warning of pegjs not being installed.

> by allowing users to extend the config and loaders

Yes, like I said I definitely want to get there but it requires more time and effort, including decisions about the approach to take. Therefore we should plan this properly.
I would say this is a "painless" quick fix for the time being.

However, since this is not really an actual issue or something, if you want to skip this I'm totally fine with it. Feel free to close then.

montezume:
If it's not a big priority, maybe we can take some time to look into this over the next week

emmenko:
Let's have a chat about this first, I also have other ideas in mind. Thanks for the feedback.

PR #289, opened 2019-01-14
External menu configurations

### LINKED ISSUE #37: [Proposal] External/Remote Navbar Config
## Motivation

Up until now we’ve been dealing with a static configuration for the navbar links, defined within the [AppShell](https://github.com/commercetools/merchant-center-application-kit/blob/9f2f12c33b5976e14d239b96f0c3c9aed5a51987/packages/application-shell/src/components/navbar/config.js#L236-L245).
This allowed us to share the same config between all the applications, including custom applications.

However, having all the config for each application in one place has some disadvantages:
* Changes need to be released in app-kit, even while you are developing some new features (not released yet) that require a menu link
* All apps need to upgrade to the latest version to have the latest changes, including custom apps

I would say that those are becoming some very annoying problems that we should start dealing with.

Ideally, the AppShell should simply not have any hardcoded config and fetch it from a “common” place. This will allow to decouple the releasing of new changes in the Navbar from the AppShell and furthermore to ensure that all applications will get the same config on runtime, no matter which version of the AppShell is installed on each application.

## So what is this “common” config and where does it come from?

I think that for starters, each application (the ones managed by us) should contain the Navbar config for their own routes. Then, during deployment, the config will be pushed to this “common” place, that contains all the configs of all applications.
This has the advantage of having the configs co-located to the application and being decoupled from the AppShell (as mentioned). Furthermore, changes to the config will be reflected immediately to all applications once deployed together with the application.

The main question I have is around the “common” place.
Eventually, the Navbar component will need to fetch this configs from an API endpoint or something.
I suppose we can have this config defined in our mc-settings API. Managing the updates to the config from the deployment though, might require some effort to sync the updates.
Another option might be to push the config to our storage bucket, as a static file. I think the problem here would be to properly sync/merge the configs from the different apps deployments.

I think this is the only open question we have to proceed with this approach.

Let me know what do you think and if you have concerns or suggestions about it.

---

Update: the final proposal is described here https://github.com/commercetools/merchant-center-application-kit/issues/37#issuecomment-444500693

emmenko:
Closes #37 

The menu configuration for each application has been moved outside of the AppShell and is available as a [GraphQL endpoint](https://github.com/commercetools/merchant-center-frontend/pull/6116) on the MC domain (_not the MC API_).

Each _internal_ application (managed by CT) can now define their own menu config, which is then uploaded to a storage service so that the GraphQL server can aggregate those configs are expose them.

**Custom applications still need to manage their settings separately, however they don't need to update the app-kit dependencies every time we change the menu links**. 🎉

For the "account" links, it works the same way.

screenshot 2019-01-14 at 18 17 49

---

TODOS:

- [x] for local development, allow to load the application menu config


packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -189,33 +187,14 @@ export class MenuItemLink extends React.PureComponent {
...
     children: PropTypes.node.isRequired,
     useFullRedirectsForLinks: PropTypes.bool.isRequired,
   };
   static defaultProps = {
     exactMatch: false,
-    tracking: {},
   };
   redirectTo = targetUrl => window.location.replace(targetUrl);
   render() {
-    if (this.props.externalLink) {
```
emmenko: This case was only used by the support link, which is now rendered separately. Hence the cleanup here

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -317,23 +294,19 @@ export class DataMenu extends React.PureComponent {
     data: PropTypes.arrayOf(
       PropTypes.shape({
         key: PropTypes.string.isRequired,
-        labelKey: PropTypes.string,
```
emmenko: All labels are now coming from `labelAllLocales`

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -481,144 +451,154 @@ export class DataMenu extends React.PureComponent {
...
-                </div>
-                <div className={styles.title}>{this.renderLabel(menu)}</div>
+            <div className={styles['item-icon-text']}>
+              <div className={styles.icon}>
+                <IconSwitcher
+                  iconName={menu.icon}
+                  size="scale"
+                  theme={getIconTheme(
+                    isActive || this.isMainMenuRouteActive(menu.uriPath),
+                    menu.shouldRenderDivider
```
emmenko: This determines the "theme" of the icon now

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -481,144 +451,154 @@ export class DataMenu extends React.PureComponent {
...
-              this.handleToggleItem(menuType, index);
-            }}
-            onMouseEnter={
-              this.state.isMenuOpen
-                ? null
-                : () => this.handleToggleItem(menuType, index)
-            }
-            onMouseLeave={
-              this.state.isMenuOpen ? null : this.shouldCloseMenuFly
+        <MenuItem
```
emmenko: Nothing much changed here, mostly indentation

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -481,144 +451,154 @@ export class DataMenu extends React.PureComponent {
...
+                  </RestrictedMenuItem>
+                ))
+              : null}
+          </MenuGroup>
+        </MenuItem>
       </RestrictedMenuItem>
     );
   };
 
+  renderSupportLink = () => {
```
emmenko: This is now rendered separately

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -481,144 +451,154 @@ export class DataMenu extends React.PureComponent {
...
-        <ScrollableMenu>
-          {this.props.data
-            .filter(menu => this.bottomMenuItems.indexOf(menu.key))
-            .map((menu, index) => this.renderMenu(menu, 'scrollable', index))}
-        </ScrollableMenu>
-        <FixedMenu>
-          {this.props.data
-            .filter(menu => !this.bottomMenuItems.indexOf(menu.key))
-            .map((menu, index) => this.renderMenu(menu, 'fixed', index))}
+        <div className={styles['scrollable-menu']}>
```
emmenko: No need for a react component here, keep it simple

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -481,144 +451,154 @@ export class DataMenu extends React.PureComponent {
...
+          {this.props.data.map((menu, index) => (
+            <React.Fragment key={index}>
+              {menu.shouldRenderDivider && <MenuItemDivider />}
```
emmenko: The divider is now rendered based on the flag `shouldRenderDivider`

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -481,144 +451,154 @@ export class DataMenu extends React.PureComponent {
...
-export const FixedMenu = props => (
-  <div className={styles['fixed-menu']}>{props.children}</div>
-);
-FixedMenu.displayName = 'FixedMenu';
-FixedMenu.propTypes = {
-  children: PropTypes.node,
-};
-
-export const NavBarLayout = props => (
+export const NavBarLayout = React.forwardRef((props, ref) => (
```
emmenko: NOTE that we use `React.forwardRef`, as documented here: https://reactjs.org/docs/forwarding-refs.html

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -643,30 +628,28 @@ export class NavBar extends React.PureComponent {
...
     }),
-
-    // injectFeatureToggle
-    areProjectExtensionsEnabled: PropTypes.bool.isRequired,
   };
 
-  getNode = node => {
-    this.node = node;
-  };
+  ref = React.createRef();
```
emmenko: 🎉

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -643,30 +628,28 @@ export class NavBar extends React.PureComponent {
...
-            this.props.projectExtensionsQuery.projectExtension
-              ? defaultNavigationItems.concat(
-                  this.props.projectExtensionsQuery.projectExtension.applications.map(
-                    app => app.navbarMenu
-                  )
-                )
-              : defaultNavigationItems
-          }
+          rootNode={this.ref.current}
+          data={navbarMenu.concat(customAppsMenu)}
```
emmenko: Here we merge the internal apps with the custom apps config

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -693,9 +675,16 @@ export default compose(
           : null,
     };
   }),
+  graphql(FetchApplicationsMenu, {
+    name: 'applicationsMenuQuery',
+    options: () => ({
+      // Pass a different apollo client here, as we are connecting to a different API
+      client: frontendClient,
```
emmenko: See changes in `configure-apollo.js` for more info

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -705,25 +694,30 @@ export default compose(
...
-          </MenuItem>
-        ))}
-      </React.Fragment>
-    </MenuGroup>
-  </NavBarLayout>
-);
-LoadingNavBar.displayName = 'LoadingNavBar';
+export class LoadingNavBar extends React.Component {
+  static displayName = 'LoadingNavBar';
+  ref = React.createRef();
```
emmenko: Even though we don't use it here, the `NavBarLayout` expects a `ref`

packages/application-shell/src/components/navbar/navbar.spec.js
```diff
@@ -1,13 +1,14 @@
-import { shallow } from 'enzyme';
 import React from 'react';
+import { shallow } from 'enzyme';
```
emmenko: FYI: when we re-implement the navbar, we can drop these tests and write RTL tests.

packages/application-shell/src/components/user-settings-menu/user-settings-menu.js
```diff
@@ -62,136 +67,193 @@ function stateReducer(state, changes) {
...
+  return NO_VALUE_FALLBACK;
+};
 
-  static propTypes = {
-    firstName: PropTypes.string,
-    lastName: PropTypes.string,
-    email: PropTypes.string,
-    gravatarHash: PropTypes.string.isRequired,
-  };
+const UserSettingsMenuBody = props => {
```
emmenko: This component is basically rendered only when the user opens the menu, therefore we can connect to fetch the menu config on this component, hence the split of components

packages/application-shell/src/components/user-settings-menu/user-settings-menu.js
```diff
@@ -62,136 +67,193 @@ function stateReducer(state, changes) {
...
-                </Card>
-              )}
+              <Text.Body isBold>
+                {[props.firstName, props.lastName].join(' ').trim()}
+              </Text.Body>
+              <Text.Body truncate>{props.email}</Text.Body>
             </div>
+          </Spacings.Inline>
+        </Spacings.Inset>
+        {menuLinks.map(menu => (
```
emmenko: This is basically the only thing that changed

packages/application-shell/src/configure-apollo.js
```diff
@@ -65,8 +43,35 @@ const fragmentMatcher = new IntrospectionFragmentMatcher({
   introspectionQueryResultData,
 });
 
-export const createApolloClient = () =>
-  new ApolloClient({
+export const createApolloClientForBackendApi = () => {
+  const httpLink = createHttpLink({
+    uri: `${window.app.mcApiUrl}/graphql`,
```
emmenko: This client connects to the MC API. The other one connects to the new graphql API in the frontend (proxy).

I couldn't come up with better names, so I went with `{backend,frontend}Client`

packages/application-shell/src/configure-apollo.js
```diff
@@ -83,5 +88,55 @@ export const createApolloClient = () =>
       },
     }),
   });
+};
+
+export const createApolloClientForFrontendApi = () => {
+  const httpLink = createHttpLink({
+    // uri: `${window.location.origin}/graphql`,
+    uri: `https://mc-6116.gcp.escemo.com/graphql`,
```
emmenko: Reminder that I need to remove this

packages/application-shell/src/components/navbar/devonly-menu-loader.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { wrapDisplayName } from 'recompose';
+
+const noop = () => Component => Component;
+
+const devonlyMenuLoader =
+  process.env.NODE_ENV !== 'development'
```
emmenko: Dealing with loading the `menu.json` config for development was/is tricky to do, not much for the implementation but for the **approach**.

For instance, on production the application uses the `https://mc.commercetools.com/api/graphql` endpoint to load the configs through the Proxy server. **In development though, there is no proxy running**.
Furthermore, you want to load the `menu.json` of the application you're **developing from**.
In fact, you can't use the links to the other apps because they lead to 404 pages.

With that in mind, we need to **disable** the graphql queries to the proxy endpoint, since the URL is defined as `${window.location.origin}/api/graphql`, which in development mode points to `http://localhost:3001/api/graphql`.
I thought about having a fake graphql endpoint in the webpack dev server, but there are some disadvantages with that:
- it's tight to webpack (e.g. if someone wants to develop an app using something different, there is a problem)
- we have duplication of the graphql schema in two places (app-kit + mc proxy)

I also tried using a custom apollo link to write to the cache with based on the `menu.json`, so that no network request is made since apollo will try to read from the cache. However, besides the implementation being a bit complex, it seems that apollo makes the request anyway, probably because the cache is not being filled in time.

This leaves us with the final approach I took, which is maybe not the ideal solution but it's good enough. One of the important things to keep in mind was that the setup should be as simple as possible for the user:

```js
DEV_ONLY__getNavbarMenuConfig={() => import('../../../menu.json')}
```

I chose to explicitly mark the prop with `DEV_ONLY_` to make it clear that this won't be used in production.

Little downside: I suppose webpack will create a chunk for the `menu.json`, since it finds `import()` split point. We could e.g. wrap the call with a `process.env.NODE_ENV` so that uglily will remove it, but I'm not sure if that makes things simpler.

Please let me know what do you think about this approach, about naming etc. I'm open to suggestions 😇

montezume: yeah maybe wrapping the call would be a good idea? but another unused chunk in prod doesn't really make any difference 🤔

packages/application-shell/src/components/navbar/devonly-menu-loader.js
```diff
@@ -0,0 +1,32 @@
...
+    ? noop
+    : (getLoaderFn, mapMenuToProps) => Component => {
+        class WrappedComponent extends React.Component {
+          static displayName = wrapDisplayName(
```
tdeekens: Nitpick: maybe not use recompose here as it's almost deprecated and we could aim to get rid of it in the app-shell maybe?

packages/i18n/data/core.json
```diff
@@ -40,6 +40,7 @@
   "LoginForm.forgotPassword": "Forgot password?",
   "LoginForm.forgotPasswordTitle": "We are redirecting you to the Forgot Password form in {countdown} seconds.",
   "LoginForm.InvalidCredentials": "Sorry, but the login details entered below are not in our system. Please enter your correct email address and password below.",
+  "LoginForm.invalidEmail": "Please enter a valid Email address.",
```
montezume: one day we will have translations that don't have German capitalization rules in English 😆

packages/application-shell/src/components/user-settings-menu/user-settings-menu.js
```diff
@@ -62,136 +68,207 @@ function stateReducer(state, changes) {
...
+      // We can assume here that the navbar already fetched the data, since this
+      // component gets rendered only when the user opens the menu
+      fetchPolicy: 'cache-only',
+      context: {
+        uri: `${window.location.origin}/api/graphql`,
+      },
+    }),
+  }),
+  devonlyMenuLoader(
+    ownProps => ownProps.DEV_ONLY__getAppbarMenuConfig,
```
tdeekens: Should we also call these `fetchAppbarMenuConfig` to indicate they fetch it?

packages/application-shell/src/components/navbar/devonly-menu-loader.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { wrapDisplayName } from 'recompose';
+
+const noop = () => Component => Component;
+
+const devonlyMenuLoader =
+  process.env.NODE_ENV !== 'development'
+    ? noop
+    : (getLoaderFn, mapMenuToProps) => Component => {
```
tdeekens: ~getLoaderFn~createLoader?

packages/application-shell/src/components/user-settings-menu/user-settings-menu.js
```diff
@@ -62,136 +68,207 @@ function stateReducer(state, changes) {
...
+const ConnectedUserSettingsMenuBody = compose(
+  graphql(FetchApplicationsMenu, {
+    name: 'applicationsMenuQuery',
+    skip: ownProps => !!ownProps.DEV_ONLY__getMenuConfig,
+    options: () => ({
+      // We can assume here that the navbar already fetched the data, since this
+      // component gets rendered only when the user opens the menu
+      fetchPolicy: 'cache-only',
+      context: {
+        uri: `${window.location.origin}/api/graphql`,
```
tdeekens: Should this maybe be an environment variable? `process.env.MC_PORXY_API_URL` or something?

packages/application-shell/src/components/navbar/devonly-menu-loader.js
tdeekens: Could we also just have a HoC which reads the process.env to alternate the fetching behaviour? So instead of having a `withDevOnlyMenuLoader` and some `graphql` (Apollo) application we have a `withMenuQuery` and it alternates between fetching from the .json and through Apollo based on the env or a specific prop?

tdeekens: Yeah, I think @montezume and I mean the same. Wrap the call in the HoC itself.

emmenko: Sure, I can try that

application-templates/starter/src/components/entry-point/entry-point.js
```diff
@@ -8,17 +8,10 @@ import { Sdk } from '@commercetools-frontend/sdk';
...
-      },
-      error => {
-        reject(error);
-      }
-    )
-  );
+const loadApplicationMessagesForLanguage = async lang => {
+  const messages = await import(`../../i18n/data/${lang}.json` /* webpackChunkName: "application-messages-[request]" */);
+  return messages.default;
+};
```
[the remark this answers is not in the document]
emmenko: Yes

packages/application-shell/src/components/navbar/devonly-menu-loader.js
emmenko: I didn’t know that. What is the alternative then? We usually use it for `compose`

packages/application-shell/src/components/user-settings-menu/user-settings-menu.js
emmenko: I would keep this as default, we could allow to optionally overwrite it if necessary.

packages/application-shell/src/components/navbar/devonly-menu-loader.js
tdeekens: Sorry, that came across wrong. I didn't mean to say "let's drop it" or that we decided. It was more of a suggestion: could the app-shell work without and if so what would be have to remove or custom build. Keep it for now. Should be a separate thing anyway. Sorry for the confusion.

packages/application-shell/src/components/user-settings-menu/user-settings-menu.js
tdeekens: Okay.

packages/application-shell/src/components/navbar/devonly-menu-loader.js
emmenko: No worries, I understand your point. I was just asking what should we use to `compose`. Should we just copy the implementation and have it as an internal function?

tdeekens: `flowRight` from lodash I suppose (the renamed compose to flowRight).

1. I would have these HoCs as little utils, yes
2. For the mc-fe it's almost unrealistic and not worthwhile getting rid of recompose

emmenko: FYI: https://www.apollographql.com/docs/react/api/react-apollo.html#compose

packages/application-shell/src/components/with-applications-menu/with-applications-menu.js
```diff
@@ -0,0 +1,93 @@
...
+          render={({ environment }) => (
+            <Query
+              query={FetchApplicationsMenu}
+              fetchPolicy={queryOptions.fetchPolicy || 'cache-first'}
+              context={{
+                // Allow to overwrite the API url from `env.json`
+                uri: `${environment.mcProxyApiUrl ||
+                  defaultApiUrl}/api/graphql`,
+              }}
+              onError={props.handleActionError}
```
emmenko: FYI: this is not necessary anymore when we use #292

emmenko:
I think this is good to go now 🚀

PS: I'm not sure if this is considered a breaking change or just a feature 🤔

PR #292, opened 2019-01-15
feat: add handleApolloErrors HOC to automatically dispatch query errors

emmenko:
As I was developing #289 I noticed that handling apollo query errors is not very much straightforward, especially in our case that we want to dispatch a notification error and not render an error within the component.

So I decided to experiment a bit and came up with the `handleApolloErrors` HOC. The implementation should be documented and tested well enough, so please try to read through and see if it's clear enough.

### DIFF
--- a/packages/application-shell/src/components/handle-apollo-errors/README.md
+++ b/packages/application-shell/src/components/handle-apollo-errors/README.md
@@ -1,0 +1,21 @@
+# `handleApolloErrors`
+
+Handles `ApolloError`s from GraphQL queries to dispatch notification errors.
+
+## Motivation
+
+When using the `graphql` HOC to send queries, it's a bit tricky to handle possible errors. For instance, the _connected_ component will need to check the `error` prop within the graphql query object and decide what to do. Most of the time, we simply want to dispatch a page notification to show an error message, some other times we might want to handle the error within that component.
+
+In the first case, we use this HOC to dispatch the notification errors for us.
+
+## Usage
+
+> This HOC should be composed **after** the `graphql` query components, to be able to access the query props.
+
+```js
+compose(
+  graphql(MyQuery, { name: 'user' }),
+  graphql(MyQuery, { name: 'project' }),
+  handleApolloErrors(['user', 'project'])
+);
+```
--- a/packages/application-shell/src/components/handle-apollo-errors/handle-apollo-errors.js
+++ b/packages/application-shell/src/components/handle-apollo-errors/handle-apollo-errors.js
@@ -28,11 +28,12 @@ function handleApolloErrors(queryResultNames) {
           // If an error has not been dispatched yet, we keep track of it and we dispatch it.
           const queryResult = this.props[name];
           if (queryResult) {
-            const dispatchedError = this.dispatchedErrors.get(name);
-            if (dispatchedError && !queryResult.error) {
+            const hasPreviouslyDispatchedError = Boolean(this.dispatchedErrors.get(name));
+            const hasQueryErrored = Boolean(queryResult.error);
+            if (hasPreviouslyDispatchedError && !hasQueryErrored) {
               this.dispatchedErrors.delete(name);
             }
-            if (!dispatchedError && queryResult.error) {
+            if (!hasPreviouslyDispatchedError && hasQueryErrored) {
               this.dispatchedErrors.set(name, queryResult.error);
               this.context.store.dispatch(handleActionError(queryResult.error));
             }
--- a/packages/application-shell/src/components/handle-apollo-errors/handle-apollo-errors.js
+++ b/packages/application-shell/src/components/handle-apollo-errors/handle-apollo-errors.js
@@ -28,7 +28,9 @@ function handleApolloErrors(queryResultNames) {
           // If an error has not been dispatched yet, we keep track of it and we dispatch it.
           const queryResult = this.props[name];
           if (queryResult) {
-            const hasPreviouslyDispatchedError = Boolean(this.dispatchedErrors.get(name));
+            const hasPreviouslyDispatchedError = Boolean(
+              this.dispatchedErrors.get(name)
+            );
             const hasQueryErrored = Boolean(queryResult.error);
             if (hasPreviouslyDispatchedError && !hasQueryErrored) {
               this.dispatchedErrors.delete(name);
--- a/packages/application-shell/src/components/handle-apollo-errors/handle-apollo-errors.js
+++ b/packages/application-shell/src/components/handle-apollo-errors/handle-apollo-errors.js
@@ -1,0 +1,41 @@
+import React from 'react';
+import { wrapDisplayName } from 'recompose';
+import { ReactReduxContext } from 'react-redux';
+import { handleActionError } from '@commercetools-frontend/actions-global';
+
+function handleApolloErrors(queryResultNames) {
+  return Component => {
+    class WrappedComponent extends React.Component {
+      static displayName = wrapDisplayName(Component, 'handleApolloErrors');
+      static contextType = ReactReduxContext;
+      dispatchedErrors = new Map();
+      componentDidUpdate() {
+        queryResultNames.forEach(name => {
+          // To avoid infinite loops (dispatch -> store updates -> connected renders -> didUpdate -> dispatch)
+          // we keep track of the errors that have been dispatched.
+          // If an error has been dispatched already, we don't do anything else.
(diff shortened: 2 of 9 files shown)


packages/application-shell/src/components/handle-apollo-errors/handle-apollo-errors.js
```diff
@@ -0,0 +1,50 @@
...
+        queryResultNames.forEach(name => {
+          // To avoid infinite loops (dispatch -> store updates -> connected renders -> didUpdate -> dispatch)
+          // we keep track of the errors that have been dispatched.
+          // If an error has been dispatched already, we don't do anything else.
+          // If an error has been dispatched already but the query does not contain an error anymore,
+          // we remove the entry from the dispatchedErrors map.
+          // If an error has not been dispatched yet, we keep track of it and we dispatch it.
+          const queryResult = this.props[name];
+          if (queryResult) {
+            const dispatchedError = this.dispatchedErrors.get(name);
```
tdeekens:
```suggestion
            const hasPreviouslyDispatchedError = Boolean(this.dispatchedErrors.get(name));
```

tdeekens: `const hasQueryErrored = Boolean(queryResult.error)`

packages/application-shell/src/components/handle-apollo-errors/handle-apollo-errors.js
```diff
@@ -0,0 +1,50 @@
...
+            if (dispatchedError && !queryResult.error) {
```
tdeekens:
```suggestion
            if (hasPreviouslyDispatchedError && !hasQueryErrored) {
```

packages/application-shell/src/components/handle-apollo-errors/handle-apollo-errors.js
```diff
@@ -0,0 +1,50 @@
...
+              this.dispatchedErrors.delete(name);
+            }
+            if (!dispatchedError && queryResult.error) {
```
tdeekens:
```suggestion
            if (!hasPreviouslyDispatchedError && hasQueryErrored) {
```

packages/application-shell/src/components/handle-apollo-errors/handle-apollo-errors.spec.js
```diff
@@ -0,0 +1,90 @@
...
+      </React.Fragment>
+    );
+  }
+}
+
+describe('given the graphql queries for "user"', () => {
+  beforeAll(() => {
+    console.error = jest.fn();
+  });
+  const Connected = handleApolloErrors(['user'])(Test);
```
tdeekens:
```suggestion
  const WithApolloErrorHandling = handleApolloErrors(['user'])(Test);
```

packages/application-shell/src/components/handle-apollo-errors/handle-apollo-errors.spec.js
```diff
@@ -0,0 +1,90 @@
...
+      const { getByText, getByTestId } = renderWithRedux(
+        <QueryController name="user" queryResult={{ data: { ok: true } }}>
+          <Connected />
+        </QueryController>
+      );
+      await waitForElement(() => getByText('loading'));
+
+      fireEvent.click(getByTestId('user'));
+
+      await waitForElement(() => getByText(/Status: true/i));
```
tdeekens: Nice testing.

packages/application-shell/src/components/handle-apollo-errors/handle-apollo-errors.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { wrapDisplayName } from 'recompose';
+import { ReactReduxContext } from 'react-redux';
+import { handleActionError } from '@commercetools-frontend/actions-global';
+
+function handleApolloErrors(queryResultNames) {
```
tdeekens: Maybe `withApollorErrorHandling`?

tdeekens: I am not sure Apollo calls these `queryResultNames` maybe just `queryNames`?

packages/application-shell/src/components/handle-apollo-errors/handle-apollo-errors.js
```diff
@@ -0,0 +1,50 @@
...
+  return Component => {
+    class WrappedComponent extends React.Component {
+      static displayName = wrapDisplayName(Component, 'handleApolloErrors');
+      static propTypes = {
```
tdeekens: I don't think we need to have an object and then spread? Can't we just `static propTypes = queryNames.reduce...` or was this for general extensibility of the prop-types?

packages/application-shell/src/components/handle-apollo-errors/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Motivation
+
+When using the `graphql` HOC to send queries, it's a bit tricky to handle possible errors. For instance, the _connected_ component will need to check the `error` prop within the graphql query object and decide what to do. Most of the time, we simply want to dispatch a page notification to show an error message, some other times we might want to handle the error within that component.
+
+In the first case, we use this HOC to dispatch the notification errors for us.
+
+## Usage
+
+> This HOC should be composed **after** the `graphql` query components, to be able to access the query props.
```
montezume: 👍

packages/application-shell/src/components/handle-apollo-errors/handle-apollo-errors.js
emmenko: `withApollorErrorHandling` usually implies that the HOC injects props into the wrapped component, but in this case we only have a side effect.

montezume: i also like `withApolloErrorHandling` or even `injectApolloErrorHandling`

emmenko: > I am not sure Apollo calls these `queryResultNames` maybe just `queryNames`?

They call it `QueryResult`, see https://www.apollographql.com/docs/react/api/react-apollo.html#query-render-prop

montezume:
Maybe just come to an agreement on the naming and then 🚀

packages/application-shell/src/components/handle-apollo-errors/handle-apollo-errors.js
emmenko: Yeah for making it less confusing if you need to extend it. I'm not sure if it would be the case here, so I guess we can remove the spread.

packages/application-shell/src/components/handle-apollo-errors/handle-apollo-errors.spec.js
```diff
@@ -0,0 +1,90 @@
...
+      await waitForElement(() =>
+        getByText(/Sorry, but there seems to be something wrong./i)
+      );
+    });
+  });
+  describe('when no query fails', () => {
+    it('should not dispatch any notification error', async () => {
+      const { getByText, getByTestId } = renderWithRedux(
+        <QueryController name="user" queryResult={{ data: { ok: true } }}>
+          <Connected />
```
emmenko:
```suggestion
          <WithApolloErrorHandling />
```

packages/application-shell/src/components/handle-apollo-errors/handle-apollo-errors.spec.js
```diff
@@ -0,0 +1,90 @@
...
+  beforeAll(() => {
+    console.error = jest.fn();
+  });
+  const Connected = handleApolloErrors(['user'])(Test);
+  describe('when the "user" query fails', () => {
+    it('should dispatch notification error', async () => {
+      const error = new Error('Oops');
+      const { getByText, getByTestId } = renderWithRedux(
+        <QueryController name="user" queryResult={{ data: null, error }}>
+          <Connected />
```
emmenko:
```suggestion
          <WithApolloErrorHandling />
```

packages/application-shell/src/components/handle-apollo-errors/handle-apollo-errors.js
tdeekens: I agree that `with` and `inject` should be reserved for component which add an additional prop (inject) into a component. Recompose also has a bunch of components `getContext` or `defaultProps` etc which are named without. I think `handleApolloErrors` is a good name after all. Sorry for the confusion.

PR #314, opened 2019-01-24
Refactor to not `ReconfigureFlopflip` but pass `projectKey` from url

tdeekens:
#### Summary

This pull request refactors the app-shell and the setup-flopflip components to not reconfigure flopflip but to pass the projectKey upon first render.

#### Description

I received feedback about flashings of feature toggles in some parts of the app. Furthermore, in certain navigation scenarios an upcoming feature to not show up.

After spending a reasonable amount of time again to understand how our app renders I suggest to simplify the setup:

- Use the `selectProjectKeyFromUrl` which is used in app-shell
- Pass the projectKey from the app-shell to the setup-flopflip component
- To not `ReconfigureFlopflip` at all

Another option would have been to remove the memoization on `ReconfigureFlopflip` to make sure there is no reconfiguration which is "invalidated" right after due to the the app-shell re-rendering `ConfigureFlopflip`.

I hope this simplification makes sense.


packages/application-shell/package.json
```diff
@@ -51,9 +51,9 @@
     "@commercetools-frontend/sentry": "7.0.0",
     "@commercetools-frontend/storage": "7.0.0",
     "@commercetools-frontend/url-utils": "7.0.0",
-    "@flopflip/launchdarkly-adapter": "2.4.5",
-    "@flopflip/memory-adapter": "1.1.0",
-    "@flopflip/react-broadcast": "7.2.4",
+    "@flopflip/launchdarkly-adapter": "2.4.6",
+    "@flopflip/memory-adapter": "1.1.1",
+    "@flopflip/react-broadcast": "7.2.5",
```
tdeekens: Mostly internal dep updates and some simplifications in how we handle adapter state. Nothing which affects the apps.

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -126,6 +123,7 @@ export const RestrictedApplication = props => (
               >
                 <SetupFlopFlipProvider
                   user={user}
+                  projectKey={projectKeyFromUrl}
```
tdeekens: This makes it a whole lot simpler.

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -319,11 +317,6 @@ export const RestrictedApplication = props => (
                                 path="/:projectKey"
                                 render={routerProps => (
                                   <React.Fragment>
-                                    <ReconfigureFlopFlip
-                                      user={getFlopflipReconfiguration(
-                                        routerProps.match.params.projectKey
-                                      )}
-                                    />
```
tdeekens: Cause this component and `ConfigureFlopflip` rendered by `SetupFlopFlipProvider` end up in a "fight" overwriting each others `adapterArgs`. There are props to influence how flopflip can merge args but I don't think it's worth having this complexity.

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.js
```diff
@@ -12,13 +12,10 @@ const ldClientSideIdProduction = '5979d95f6040390cd07b5e01';
 // one above only for `production` environment.
 const ldClientSideIdStaging = '5979d95f6040390cd07b5e00';
 
-export const getFlopflipReconfiguration = defaultMemoize(projectKey => ({
```
tdeekens: This is part of the problem. By memoizing here, the `ReconfigureFlopflip` never gets to invoke `componentDidUpdate` which makes it in some cases loose the fight for the latest props against `ConfigureFlopflip`.

PR #316, opened 2019-01-25
Ml move media queries

montezume:
#### Summary

Moves `PageNotFound` and `ServicePageResponseLayout` to a new package called `application-components`. 

Moves `media-queries` from `ui-kit` to inside of `application-components`.

Refactors the tests to use `react-testing-library` and kill ~~useless~~ snapshots.


package.json
```diff
@@ -76,7 +76,7 @@
     "rollup-plugin-node-builtins": "2.1.2",
     "rollup-plugin-node-resolve": "4.0.0",
     "rollup-plugin-peer-deps-external": "2.2.0",
-    "rollup-plugin-postcss": "1.6.3",
+    "rollup-plugin-postcss": "2.0.0",
```
montezume: sick of seeing that warning. this release allows us to use rollup > 1 without a warning 👍

packages/application-components/src/components/service-page-response-layout/service-page-response-layout.spec.js
```diff
@@ -0,0 +1,65 @@
...
+      const { getByText } = render(<ServicePageResponseLayout {...props} />);
+      expect(getByText(props.title)).toBeInTheDocument();
+      expect(getByText(props.paragraph1)).toBeInTheDocument();
+      expect(getByText(props.paragraph2)).toBeInTheDocument();
+    });
+  });
+  describe('with both paragraphs and a body content in between', () => {
+    beforeEach(() => {
+      props = createTestProps({
+        paragraph2: 'title 2',
```
montezume: before we were passing in divs, which aren't allowed descendants of paragraph or header tags.

packages/application-components/materials/media-queries.css
```diff
@@ -0,0 +1,13 @@
...
+@custom-media --screen-max-biggerdesktop (max-width: 1279px); /* md */
+@custom-media --screen-max-giantdesktop (max-width: 1679px); /* lg */
+@custom-media --screen-max-jumbodesktop (max-width: 1919px); /* xlg */
+
+/* MIN-WIDTH */
+@custom-media --screen-min-mobile (min-width: 768px); /* xsm */
+@custom-media --screen-min-desktop (min-width: 1024px); /* sm */
+@custom-media --screen-min-biggerdesktop (min-width: 1280px); /* md */
+@custom-media --screen-min-giantdesktop (min-width: 1680px); /* lg */
+@custom-media --screen-min-jumbodesktop (min-width: 1920px); /* xlg */
```
emmenko: Q: should we provide this in other formats as well, like we do with the custom properties? E.g. json, js...

packages/application-components/package.json
```diff
@@ -0,0 +1,57 @@
...
+  "license": "MIT",
+  "private": false,
+  "publishConfig": {
+    "access": "public"
+  },
+  "main": "dist/application-components.cjs.js",
+  "module": "dist/application-components.es.js",
+  "files": [
+    "dist",
+    "materials",
```
emmenko: Nitpick: should we specify the single files, like in the ui-kit?

packages/application-components/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "build": "npm run build:es && npm run build:cjs",
+    "build:cjs": "cross-env NODE_ENV=production rollup -c ../../rollup.config.js  -f cjs ./index.js -o dist/application-components.cjs.js",
+    "build:es": "cross-env NODE_ENV=production rollup -c ../../rollup.config.js  -f es ./index.js -o dist/application-components.es.js",
+    "build:es:watch": "npm run build:es -- -w"
+  },
+  "dependencies": {
+    "prop-types": "15.6.2"
+  },
+  "devDependencies": {
+    "@commercetools-frontend/application-shell": "7.0.0",
```
emmenko: I'm not sure I like this, it introduces a circular dependency. I don't think we need the test utils from app-shell to be honest, we can use RTL directly because there is not much to setup.

I would add a `test-utils` file in this package with only the setup needed for the components (e.g. react-intl etc).

packages/application-components/materials/media-queries.css
montezume: @emmenko good point, if we do that, we should actually use the `breakpoints` that are exposed in `ui-kit` custom properties as well. I'll look into this 🎉

packages/application-components/package.json
montezume: yep, good call

emmenko:
We can do the media queries improvements as a follow up. Let's get this merged then! 🚀

package.json
tdeekens: Nice!

packages/application-components/src/components/page-not-found/page-not-found.spec.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { render } from '../../../test-utils';
+import PageNotFound from './page-not-found';
+import messages from './messages';
+
+describe('rendering', () => {
+  it('should render help desk link', () => {
+    const { getByText } = render(<PageNotFound />);
+    expect(getByText(messages.helpDesk.defaultMessage)).toBeInTheDocument();
```
tdeekens: 💯

packages/application-components/test-utils/test-utils.js
```diff
@@ -0,0 +1,20 @@
...
+
+const customRender = (node, { locale = 'en', ...rtlOptions } = {}) => ({
+  ...render(<IntlProvider locale={locale}>{node}</IntlProvider>, rtlOptions),
+});
+
+// re-export everything
+export * from 'react-testing-library';
+
+// override render method
+export { customRender as render };
```
tdeekens: 👍

PR #328, opened 2019-01-30
Enable session storage along with local storage

amine-benselim:
### Summary

- add session storage to the storage module
- adapt `storage` usage and update failing tests

#### Description

this aims to unify the use of storage inside the merchant center app kit and also next in the merchant center frontend. 

since I needed to work with `sessionStorage` on mc-frontend, and after seeing it's been used previously directly from `window.sessionStorage`, I thought it is better to add it to the `storage` module to have a better and more unified developer experience.

### DIFF
--- a/packages/application-shell/src/components/application-shell/application-shell.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.js
@@ -7,7 +7,7 @@ import {
   joinPaths,
   trimLeadingAndTrailingSlashes,
 } from '@commercetools-frontend/url-utils';
-import * as storage from '@commercetools-frontend/storage';
+import { localStorage } from '@commercetools-frontend/storage';
 import { DOMAINS, LOGOUT_REASONS } from '@commercetools-frontend/constants';
 import history from '@commercetools-frontend/browser-history';
 import {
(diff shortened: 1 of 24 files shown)


amine-benselim:
> Hey, thanks for the PR.
> 
> However, this would be a breaking change because the API of the package changed and thus we would need to migrate all imports. I don't think it's worth it to be honest.
> We actually had this discussion some time ago and decided not to change it.
> 
> In fact, the `storage` package should probably be deprecated, and simply use `window.{local,session}Storage`.
> One of the reasons we had the package in the first place was to make it easier to mock it, but you can simply do it with `jest` now.

I also agree with you that we don't essentially need a package for this since it does not add or alter any functionality of `Storage`.

but I think having a package to wrap localStorage and not sessionStorage is wrong (nothing special about localStorage), so maybe removing the package is a good idea.

montezume:
Might make sense just to mark the current package as deprecated and then in the next next release, dump it. 🤔 

AFAIK there's actually only a few uses of `@commercetools-frontend/storage` so it shouldn't be heruclean to refactor it.

amine-benselim:
> Might make sense just to mark the current package as deprecated and then in the next next release, dump it. 🤔
> 
> AFAIK there's actually only a few uses of `@commercetools-frontend/storage` so it shouldn't be heruclean to refactor it.

yes I think this is the right thing to do. I will refactor `merchant-center-frontend` to remove it while working on my current task. so it would only be used locally relative to this project.

emmenko:
Good, we can close this then. Thanks!

PR #329, opened 2019-01-31
fix(application-components): name file correctly

emmenko:
🤦‍♂️

The file was not in the published package because of the wrong name


montezume:
😆

PR #332, opened 2019-02-01
Add graceful shutdown and isolated Prometheus metrics server to mc-http-server

tdeekens:
#### Summary

This pull request adds graceful shutdown capabilities to the mc-http-server and isolates the server exposing the Prometheus metrics. Due to the fact the `GET /metrics` is not available anymore this should be considered a breaking change.


emmenko:
Q: why is it a breaking change?

emmenko:
I think we only use that endpoint internally, so I wouldn't worry much about that. A minor version bump should be enough.

PR #340, opened 2019-02-04
chore: update deps

tdeekens:
- Closes #339
- Closes #338
- Closes #337
- Closes #336
- Closes #335
- Closes #334


tdeekens:
Nothing interesting here. We can't update anything to jest@24 yet due to their fixed but not released but of swallowing some config options in presets.

PR #342, opened 2019-02-04
Add placeholder to quick-access input

tdeekens:
#### Summary

This pull request adds a placeholder text to the input of quick access.

cleanshot 2019-02-04 at 14 48 49 2x

### DIFF
--- a/packages/application-shell/src/components/quick-access/butler/butler.js
+++ b/packages/application-shell/src/components/quick-access/butler/butler.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Fuse from 'fuse.js';
-import { FormattedMessage } from 'react-intl';
+import { FormattedMessage, injectIntl } from 'react-intl';
 import last from 'lodash/last';
 import classnames from 'classnames';
 import { SearchIcon } from '@commercetools-frontend/ui-kit';
@@ -44,7 +44,7 @@ const hasNewWindowModifier = event => {
   }
 };
 
-export default class Butler extends React.Component {
+export class Butler extends React.Component {
   static displayName = 'Butler';
 
   static propTypes = {
@@ -63,6 +63,10 @@ export default class Butler extends React.Component {
     getNextCommands: PropTypes.func.isRequired,
     executeCommand: PropTypes.func.isRequired,
     onClose: PropTypes.func.isRequired,
+    // injectIntl
+    intl: PropTypes.shape({
+      formatMessage: PropTypes.func.isRequired,
+    }).isRequired,
   };
 
   state = {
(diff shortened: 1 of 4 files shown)


packages/application-shell/src/components/quick-access/butler/butler.js
```diff
@@ -430,6 +434,9 @@ export default class Butler extends React.Component {
             <input
               id="quick-access-search-input"
               ref={this.searchInputRef}
+              placeholder={this.props.intl.formatMessage(
```
tdeekens: I tried for some minutes to have `FormattedMessage` work with the new FaaC API but failed as it returns a react element which the placeholder can not render.

emmenko: You mean like

```js
<FormattedMessage {...messages.inputPlacehoder}>
  {label => (
    <input placeholder={label} />
  )}
</FormattedMessage>
```

packages/application-shell/src/components/quick-access/butler/butler.mod.css
```diff
@@ -77,6 +77,10 @@
   padding: var(--spacing-16) var(--spacing-16) var(--spacing-8) var(--spacing-8);
 }
 
+.searchText::placeholder {
```
montezume: Question, do we have auto prefixing happening?

tdeekens: I do assume so as the rollup config specifically has it: 

```js
postcssPresetEnv({
            browsers: browserslist.production,
            autoprefixer: { grid: true },
          }),
```

montezume: Because if not we need at least the edge prefixed version  https://caniuse.com/#feat=css-placeholder

packages/application-shell/src/components/quick-access/butler/butler.js
tdeekens: Ah maybe that would work. How important is this for you? You'd request the change here?

packages/application-shell/src/components/quick-access/butler/butler.mod.css
montezume: 🎉

packages/application-shell/src/components/quick-access/butler/butler.js
emmenko: The label should be a string actually. Let me double check

emmenko: I just tried and it works as expected. Feel free to use that or not

emmenko: screenshot 2019-02-04 at 15 06 55

tdeekens: Ah, got it wrong the other way around. Thanks! I'll keep it this way however. I actually like this more as it's a bit less noisy and easy to read at the cost of using a HoC.

PR #346, opened 2019-02-06
refactor: to not setupTestFramework for jest (so we can update)

tdeekens:
#### Summary

This pull request removes the `setupTestFrameworkScriptFile` from the jest preset.

#### Description

We thought to be able to be backwards compatible but can't really as jest complains you can't have both:

cleanshot 2019-02-06 at 13 39 07 2x

The fact that the jest-preset is an object not a function also doesn't allow us to pass in an `version` option or something.

My question: do we consider this breaking? We maybe should.


tdeekens:
I will choose patch. Strictly speaking semver it maybe should be breaking but it also fails very obvious to consumers (if we have any other than us) of the package. So they just need to update jest.

tdeekens:
It should have been `jest@25` not 24. This is kind of the definition of a breaking change to me: https://github.com/facebook/jest/commit/6a7e6d66115a2ffe855bdf9b86936fdc52829966#diff-01d323d7b4d36c55f0247cf2086585d3L8

PR #348, opened 2019-02-06
Replace `query-string` with `qss`

tdeekens:
#### Summary

This pull request proposes to also remove `query-string` for `qss` in app-kit.

#### Description

There are some change I will comment on inline.

### DIFF
--- a/packages/application-shell/src/components/login-sso-callback/login-sso-callback.js
+++ b/packages/application-shell/src/components/login-sso-callback/login-sso-callback.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { compose, withProps } from 'recompose';
-import qs from 'query-string';
+import { decode } from 'qss';
 import jwtDecode from 'jwt-decode';
 import { connect } from 'react-redux';
 import * as storage from '@commercetools-frontend/storage';
@@ -41,7 +41,7 @@ export class LoginSSOCallback extends React.PureComponent {
   };
 
   componentDidMount() {
-    const fragments = qs.parse(this.props.location.hash.substring(1));
+    const fragments = decode(this.props.location.hash.substring(1));
     const idToken = fragments.id_token;
     const decodedIdToken = jwtDecode(idToken);
     // Validate the nonce.
--- a/packages/application-shell/src/components/login-sso-callback/login-sso-callback.spec.js
+++ b/packages/application-shell/src/components/login-sso-callback/login-sso-callback.spec.js
@@ -9,7 +9,7 @@ jest.mock('@commercetools-frontend/storage');
 
 const createTestProps = props => ({
   location: {
-    hash: '#id_token=111',
+    hash: '#id_token=abcTestIdToken',
   },
   redirectTo: jest.fn(),
   requestAccessToken: jest.fn(),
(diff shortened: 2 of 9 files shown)


packages/application-shell/src/components/login-sso-callback/login-sso-callback.js
```diff
@@ -41,7 +41,7 @@ export class LoginSSOCallback extends React.PureComponent {
   };
 
   componentDidMount() {
-    const fragments = qs.parse(this.props.location.hash.substring(1));
+    const fragments = decode(this.props.location.hash.substring(1));
```
tdeekens: This is needed otherwise the object has `{ '?foo': 'bar, ... }` which is not bad but not what we want here.

packages/application-shell/src/components/login-sso-callback/login-sso-callback.spec.js
```diff
@@ -64,7 +64,7 @@ describe('lifecylcle', () => {
       });
       it('should call requestAccessToken with idToken', () => {
         expect(props.requestAccessToken).toHaveBeenCalledWith(
-          expect.objectContaining({ idToken: '111' })
+          expect.objectContaining({ idToken: 'abcTestIdToken' })
```
tdeekens: Cause 111 will be casted to an int as qss casts boolean and numbers for us I felt this may be a better test token name.

packages/browser-history/index.js
```diff
@@ -1,7 +1,9 @@
 import createBrowserHistory from 'history/createBrowserHistory';
 import withQuery from 'history-query-enhancer';
-import queryString from 'query-string';
+import { encode, decode } from 'qss';
 
-const history = withQuery(queryString)(createBrowserHistory());
+const history = withQuery({ parse: decode, stringify: encode })(
```
tdeekens: This is what the `withQuery` expects, source: https://github.com/yenshih/history-query-enhancer/blob/master/src/withQuery.ts#L12

packages/sdk/src/utils/utils.js
```diff
@@ -7,7 +7,7 @@ export const parseUri = uri => {
 
   return {
     pathname: parser.pathname,
-    search: qs.parse(parser.search),
+    search: decode(parser.search.substring(1)),
```
tdeekens: This is needed otherwise the object has `{ '?foo': 'bar, ... }` which is not bad but not what we want here. Same as with the hashbang.

packages/sdk/src/utils/utils.spec.js
```diff
@@ -18,13 +18,13 @@ describe('parseUri', () => {
...
-    it('should extract the', () => {
+    it('should extract the pathname', () => {
       expect(result.pathname).toBe('/project/products/new');
     });
     it('should extract the search', () => {
       expect(result.search).toEqual({
-        foo: 'true',
-        bar: '5',
+        foo: true,
+        bar: 5,
```
tdeekens: Cause things are casted.

tdeekens:
Thanks for merging but was there a reason too? I wanted to test this locally with the app tomorrow before merging. I think it should be fine but wanted to make sure.

emmenko:
You can use the canary release for testing if you want 😉

tdeekens:
I know I can use a canary release. I did before on multiple occasions abd they are good for such a and other things. However, his now (speaking as a creator of the PR) would be part of any future release (which is out of my control). My main point being is just (as I seemed go fail to communicate it before) lets keep merging to a creator of a PR unless he/she labels or comments on the PR with a „ship it“ comment. 

P.S.: I do not know what smiley means in your previous comment but it made me write this.

emmenko:
My bad, I assumed this was approved and ready to go.

tdeekens:
Yeah, all cool. i should have added a label „please do not merge“ or WIP but didn‘t really expect such a quick/late merge. Will do next time 🐇.

PR #350, opened 2019-02-06
WIP feature: mc-http-server helm chart

Freyert:
#### Summary

This adds a helm chart for deploying the `mc-http-server` on Kubernetes. 

#### Description
Things to do before merging:

* [ ] understand the values that go into the `--config=env.json`
* [ ] Document all chart values including `env.json`
* [ ] Add `ignores` to `Docker` and `npm` for the `chart` directory.
* [ ] For users who do not want to use a CDN, add a configuration file to deploy their minified code as a config map and mount it into the correct directory inside the container. (which directory exactly?)

### DIFF
--- a/packages/mc-http-server/chart/templates/deployment.yaml
+++ b/packages/mc-http-server/chart/templates/deployment.yaml
@@ -1,0 +1,56 @@
+apiVersion: v1
+kind: Deployment
+metadata:
+  name: {{ template "mc-http-server.fullname" . }}
+  labels:
+    app: {{ template "mc-http-server.name" . }}
+    chart: {{ template "mc-http-server.chart" . }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+  annotations:
+    checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+spec:
+  minReadySeconds: {{.Values.mc-http-server.minReadySeconds}}
+  replicas: {{.Values.mc-http-server.replicas}}
+  selector:
+    matchLabels:
+      app: {{ template "mc-http-server.name" . }}
+      chart: {{ template "mc-http-server.chart" . }}
+      heritage: {{ .Release.Service }}
+      release: {{ .Release.Name }}
+  template:
+    metadata:
+      name: {{ template "mc-http-server.fullname" . }}
+      labels:
+        app: {{ template "mc-http-server.name" . }}
+        chart: {{ template "mc-http-server.chart" . }}
+        heritage: {{ .Release.Service }}
+        release: {{ .Release.Name }}
+    spec:
+      containers:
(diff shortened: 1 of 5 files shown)


packages/mc-http-server/chart/templates/deployment.yaml
```diff
@@ -0,0 +1,56 @@
...
+        image: {{ .Values.mc-http-server.image}}:{{ .Values.mc-http-server.tag}}
+        imagePullPolicy: IfNotPresent
+        livenessProbe:
+          httpGet:
+            path: /version
+            port: {{.Values.mc-http-server.port}}
+          initialDelaySeconds: 5
+        readinessProbe:
+          httpGet:
+            path: /version
```
tdeekens: The server has a ready and health probe. Why do we use version?

emmenko:
Thanks for your contribution!

I have a couple of questions first:
- how is the chart supposed to be used? I mean, I know how to use it but in terms of release process. Do we need one and if so, what would that be?
- did you left out things like Service and Ingress on purpose?
- as Tobi mentioned, we should align it with our internal chart, although it can't be the same because e.g. we don't expose the app through an Ingress
- what is the real intention here from your side? It would have been helpful first to open an issue and discuss it before jumping into the implementation

To be honest, I'm a bit skeptical that this would be useful. I'd rather have it only documented how to run a custom application in a K8s cluster. I'm saying this because from my experience it's a bit hard to have a chart that fits different use cases, setups, naming conventions etc.

I would suggest for now to maybe add it to our k8s repo if you wish to use internally at commercetools.

Looking forward to your feeback!

Freyert:
Originally I think I wanted to use it as a repeatable pattern we can use for any partners or customers that want to deploy a merchant center plugin (like the one I'm working with now), but I agree with your point @emmenko that it's hard to make one that fits every use case.

Perhaps we could use it as a blueprint or a way to document the deployment in a very clear format that follows the conventions seen in [helm/charts](https://github.com/helm/charts/tree/master/stable/nginx-ingress#configuration). It would also make documenting items such as the liveness and readiness probes, which I've missed, very straight forward.

tdeekens:
We could expose our internal helm chart and make the Ingress configurable and being able to set it to `enabled: false` for out cases. Still, I am not sure, as mentioned, if we can make one chart to rule all cases.

Freyert:
How about I close this one and just add more documentation to the internal chart. Then when the internal chart is feeling generalize-able we can think of opening it up.

tdeekens:
Sounds cool!

emmenko:
Eventually we will document such things in the upcoming documentation website. For instance, providing examples/guides how to deploy an application in different ways (k8s, now, aws, netlify, etc).

emmenko:
And please reach out to us if you have some questions, missing documentation, requirements, etc.

Issue #293, opened 2019-01-16
[Proposal] Define CSP and Security Headers in MC Proxy, deprecate csp.json


emmenko:
This proposal is part of the tooling improvements in #214 

## Motivation

At the moment, using the `mc-http-server` to run an application in production will include **security headers**, such as CSP. The `mc-http-server` comes with a pre-defined configuration that can be extended using a `csp.json` file.

While this approach is perfectly fine, it has some drawbacks:
- security headers are defined within the `mc-http-server` and are not enforced by our production system
- people can choose to deploy the application in a different way, by not using the `mc-http-server` package, which results in a "loss" of the headers (unless the user explicitly defines them again)

To support those use cases, of developing applications using other tooling (e.g. parcel), we need to rethink a bit our approach and **make it easier for people to have secured applications without much effort**.

## Proposal

Given that in production each application is served by our MC Proxy, we can define the security headers (including CSP) in the proxy server itself.

#### Local development

At the moment, the development server (webpack-dev-server) is configured to apply the security headers. In general, I find it useful to have that for development, because it's a setup that resembles production and it's useful to have the app secured.

So how should we deal with that and the MC proxy?

I suppose we could move the security headers config into a separate package, which can be used by both the MC Proxy and for whatever server is used for local development.

#### CSP for custom applications

Custom applications still need to extend the CSP based on the domains they are hosted and other possible external services used by the application.

I guess there are a couple of options that we can use to do so:
- extend the CSP config using custom HTTP headers (e.g. `"x-mx-csp-script-src"="foo.com,bar.com"`)
- extend the CSP config as settings (where you register your applications in the MC)

## Outcome

It should be possible to develop applications using different tooling and different ways to deploy them (even as static apps with e.g. Netlify) without worrying about security headers.

tdeekens:
Sorry for loosing track of this. I think it makes sense to extract the headers into an app-kit package which just exposes them. Then anybody can use them in any way needed. We ourselves could just use them in the proxy.

PR #354, opened 2019-02-08
fix(application-shell): redirect to welcome by default

tdeekens:
⚠️ Do not merge yet ⚠️ 

#### Summary

This pull requests suggest to redirect to the welcome page.

### DIFF
--- a/packages/application-shell/src/components/application-shell/application-shell.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.js
@@ -285,13 +285,13 @@ export const RestrictedApplication = props => (
                                 render={props.render}
                               />
                               {/* Project routes */}
-                              {/* Redirect from base project route to dashboard */}
+                              {/* Redirect from base project route to welcome */}
                               <Route
                                 exact={true}
                                 path="/:projectKey"
                                 render={({ match }) => (
                                   <Redirect
-                                    to={joinPaths(match.url, 'dashboard')}
+                                    to={joinPaths(match.url, 'welcome')}
                                   />
                                 )}
                               />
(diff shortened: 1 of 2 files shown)


packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -285,13 +285,13 @@ export const RestrictedApplication = props => (
                                 render={props.render}
                               />
                               {/* Project routes */}
-                              {/* Redirect from base project route to dashboard */}
+                              {/* Redirect from base project route to welcome */}
                               <Route
```
emmenko: If we do that, I would simply remove this route. What will happen then is that the fallback app (which most likely is the application served when requests are e.g. `/` or `/:projectKey`) will handle the redirect `/:projectKey` -> `/:projectKey/welcome`.

I would avoid hardcoding any more routes here, who know what will come in the future.

tdeekens: Makes sense. I am confused by apparently two opinions towards this:

1. Keep it backwards compatible
2. Value in removing the route here entirely

I don't mind and am really open to what the best solution is. What would you prefer?

emmenko: Well, eventually we should remove it. It's just the question if we want to do it now or maybe wait for the "next major" release?

tdeekens: When would be the next major release. Did we plan one? I personally don't mind major releases now as custom apps are in beta. Just my take though.

emmenko: Probably #293 but I haven't planned yet when to work on that. So, yeah I'm fine with the breaking change now.

tdeekens:
@emmenko this would be ready now. We got the translations for the welcome screen. Are you fine with this change. Then I would release it as breaking, integrate it with a proxy router change and deploy a branch for UAT.

PR #355, opened 2019-02-08
fix(browser-history): to omit the question mark

tdeekens:
#### Summary

Caused by: https://github.com/commercetools/merchant-center-application-kit/pull/348

We need to omit the `?` otherwise parts of applications relying on it will break.


tdeekens:
Ref: https://github.com/yenshih/history-query-enhancer/blob/master/src/withQuery.ts#L12

PR #362, opened 2019-02-11
chore: update deps

tdeekens:
#### Summary

- Closes #361
- Closes #360
- Closes #358
- Closes #359
- Closes #357
- Closes #356


tdeekens:
This is mostly dev deps but I'd like to have the same react, uikit and formik version across repos.

PR #363, opened 2019-02-11
chore(mc-scripts): add webpack filter plugin to suppress css order warnings

montezume:
#### Summary

[This issue](https://github.com/webpack-contrib/mini-css-extract-plugin/issues/250) describes the issue well. 

Right now, we are pinning the dependency `mini-css-extract-plugin` to an older version just so we don't get these warnings. However, the output CSS remains the same as the new version, the only difference is that the older version didn't output the warnings. So why not just suppress the warnings?


tdeekens:
Maybe instead of having it here we could also just push onto the app plugins. I don't mind. Either way is fine by me.

montezume:
Sure 👍

emmenko:
I agree that the workaround should we done in the specific app. We can close this then.

PR #364, opened 2019-02-12
feat(app-shell): export withApplicationsMenu from app-shell

tdeekens:
#### Summary

This PR re-exports the `withApplicationsMenu` from the application-shell.

#### Description

Given we want to use the `withApplicationsMenu` in the new welcome screen we need it exported from app-kit. As it wasn't before this PR intends to do so.

Just a side note: personally I am not yet 100% sold if this being of high value or being really a good idea given the use case:

1. Always having permissions in sync is super ✅ 
2. Assuming that the menu names ever change and can get out of sync maybe a bit far fetched 
    - I am not even sure design wants the same name all the time
3. Logic to deduct the icon (maybe even with another field on the query feels a bit over done
    - The welcome page can be in 100% control of the icons nothing can get out of sync

Maybe I am missing something. Maybe it could be improved by augmenting the schema by e.g. new fields like `isVisibleInWelcomeScreen` or `welcomeScreen: { icon, title }` which allow us to more generically target things. I just wanted to raise my thoughts here.

### DIFF
--- a/packages/application-shell/src/index.js
+++ b/packages/application-shell/src/index.js
@@ -18,6 +18,9 @@ export { GtmContext } from './components/gtm-booter';
 export {
   default as handleApolloErrors,
 } from './components/handle-apollo-errors';
+export {
+  default as withApplicationsMenu,
+} from './components/with-applications-menu';
 
 /**
  * NOTE:


tdeekens:
> Hmm, I'm not sure we want to expose this.

Me neither. Hence my concerns.

> However, I honestly do not think those links are a long-term solution.

Can you elaborate? Have you talked to design about it? What will be the long-term solution?

> For instance, we can also read the data from the apollo cache only, assuming that the menu has been fetched.

I thought about that too. However, then you have a very implicit coupling to the query being fetched and the things in the cache which can break quite unexpectedly. Then for me the cost and benefit don't balance right anymore.

> Or we could also expose a separate query from the GraphQL API that is targeted for the welcome screen links.

True, but it's something really static, nothing inter-app dependent, so I don't see a massive advantage in going over the wire. The only risk right now (to my naive view) is that the permissions can get outdated. Given they change.

> I would prefer to keep the query "duplicated" in the welcome screen app for now.

Sorry, I can't follow. What query do you mean? There is different loading behaviour between dev and stage/prod? How can this work? I thought the component meant to encapsulates that and this shouldn't leak into the app.

Over all right now it feels like there are lots of uncertainties without (at least for me) as it is right now clear benefits. Maybe that's just me obviously viewing it like that.

tdeekens:
I will close this. I don't think implementation complexity and benefit yet balance right to justify this.

PR #365, opened 2019-02-14
feat(l10n): rename Macedonia -> North Macedonia 🇲🇰 🇲🇰 🇲🇰 🇲🇰 🇲🇰 🇲🇰 🇲🇰 

montezume:
As of February 12th, Macedonia is now known as Northern Macedonia. 🇲🇰 

I got the german name from [here](https://de.wikipedia.org/wiki/Streit_um_den_Namen_Mazedonien)
And the spanish name from [here](https://es.wikipedia.org/wiki/Disputa_sobre_el_nombre_de_Macedonia)
and english from [here](https://www.theguardian.com/world/2019/feb/12/nato-flag-raised-ahead-of-north-macedonias-prospective-accession)

### DIFF
--- a/packages/l10n/data/countries/de.json
+++ b/packages/l10n/data/countries/de.json
@@ -149,7 +149,7 @@
   "mf": "St. Martin",
   "mg": "Madagaskar",
   "mh": "Marshallinseln",
-  "mk": "Mazedonien",
+  "mk": "Nordmazedonien",
   "ml": "Mali",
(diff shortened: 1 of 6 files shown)


montezume:
@emmenko yeah it seems like http://cldr.unicode.org/index/downloads is outdated... and node-cldr is using that. Hopefully we catch regression in future PRs that run that script.

PR #366, opened 2019-02-15
Add Storybook for developing application components

emmenko:
The setup is similar to the one we have in ui-kit.

screenshot 2019-02-15 at 11 06 08

#### Additional changes
I cleaned up a bit the current components (PageNotFound, ServicePageResponseLayout) and used a css grid to center the elements in the page.
I also renamed `ServicePageResponseLayout` to `MaintenancePageLayout`, as the name was very confusing. However, the `ServicePageResponseLayout` export is still available but marked as deprecated (so no breaking change necessary).


.storybook/webpack.config.js
```diff
@@ -0,0 +1,43 @@
...
+  storybookBaseConfig.module.rules = [
+    // Disable require.ensure as it's not a standard language feature.
+    { parser: { requireEnsure: false } },
+    // add story source
+    {
+      test: /\.story\.js$/,
+      loaders: [require.resolve('@storybook/addon-storysource/loader')],
+      enforce: 'pre',
+    },
+    ...mcWebpackConfig.module.rules,
```
emmenko: I found it easier to use like this as we need things like postcss, svg loaders etc.

packages/application-components/src/components/maintenance-page-layout/maintenance-page-layout.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Constraints, Text, Spacings } from '@commercetools-frontend/ui-kit';
+import deprecateComponent from '../../utils/deprecate-component';
+import styles from './maintenance-page-layout.mod.css';
+
+const MaintenancePageLayout = props => (
+  <div className={styles.container}>
```
emmenko: This renders a css grid. We can't use the ui-kit Grid as we need additional styles (e.g. `height`)

packages/application-components/src/components/maintenance-page-layout/maintenance-page-layout.js
```diff
@@ -0,0 +1,47 @@
...
+    <Constraints.Horizontal constraint="l">
+      <Spacings.Stack scale="m">
```
emmenko: We now use `Constraints` and `Spacings` components 🎉

packages/application-components/src/components/maintenance-page-layout/maintenance-page-layout.js
```diff
@@ -0,0 +1,47 @@
...
+        <div>
+          <img src={props.imageSrc} />
+        </div>
+        <div>
+          <Text.Headline elementType="h2">{props.title}</Text.Headline>
+        </div>
```
emmenko: The wrapping divs are for the margins 😕

package.json
```diff
@@ -38,6 +40,12 @@
   "devDependencies": {
     "@commitlint/cli": "7.5.1",
     "@commitlint/config-conventional": "7.5.0",
+    "@storybook/addon-actions": "4.1.11",
```
tdeekens: Can't that we in the package of application-components?

emmenko: Sure, I'll move it in there.

montezume:
I'm wondering if it might actually make sense to move `application-components` to another repo? We could use a percy free plan probably.

tdeekens:
> I'm wondering if it might actually make sense to move `application-components` to another repo? We could use a percy free plan probably.

My head goes like: no please not more repos 😄

emmenko:
> We could use a percy free plan probably

What do you mean? 🤔

qmateub:
Maybe Malcolm means to have that new repo with VRT (percy) for all the common components like the PageNotFound 🤔 

I like a lot the idea of having this Nicola 🥇

montezume:
I mean if we triggered VRT tests everytime there was a chance in any of the packages, we would be wasting snaps and spending a lot of money, whereas if we had a way of running and requiring VRT when rollup config and or the components themselves changed, that would be more reasonable.

emmenko:
We can run percy only if something in the package changed right?

montezume:
@emmenko yes, but if we enabled percy for the whole repo, it would block merging of PRs that didn't run Percy.

tdeekens:
We could have a "global" run script called `test:vrts` this would run in all packages via `lerna`. Each package then has a `test:vrts` which for all, apart one, does `exit 0` as they don't have vrts while for one it runs percy. I would suggest this to be another PR though. I don't really like having another repo just to have VRTs to be honest.

package.json
```diff
@@ -25,6 +25,8 @@
     "test:watch": "jest --config jest.test.config.js --watch",
     "playground:start": "yarn --cwd playground start",
     "playground:build": "yarn --cwd playground build",
+    "storybook:build": "yarn --cwd packages/application-components storybook:build",
```
montezume: shouldn't these scripts be part of the `application-components` package?

montezume:
Yeah, sorry for throwing us off track. Let's concentrate on Storybook in this PR of course

emmenko:
I think what Malcom means is the github check for Percy?

package.json
tdeekens: We should be able to use `yarn workspace application-components storybook:build` here? [ref](https://yarnpkg.com/lang/en/docs/cli/workspace/)

montezume:
@emmenko exactly.

packages/application-components/src/utils/deprecate-component.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { getDisplayName } from 'recompose';
+import warning from 'warning';
+
+export default function({ message = '' }) {
```
adnasa: why is this here?
Is the intention then to import from app-kit when using `deprecate-component` in the future?

packages/application-shell/src/components/project-suspended/project-suspended.spec.js
```diff
@@ -1,29 +1,19 @@
...
-    });
+  it('when suspension is temporary it should print correct message', async () => {
+    const { getByText } = render(
+      <Route
+        path="/:projectKey"
+        render={() => <ProjectSuspended isTemporary={true} />}
+      />,
+      { route: '/my-project' }
+    );
+    await waitForElement(() =>
```
adnasa: no assertion that follows?

tdeekens: This would throw already so the test fails.

adnasa: 🤦‍♂️

Issue #323, opened 2019-01-28
Adding UI components for building applications


emmenko:
Now that we started moving components into the new package `@commercetools-frontend/application-components` (#316), we can continue adding more components that will help building applications.

This issue is meant to show a priority list of the components that we want to move/add first.

_Components should adhere to the same principles of components developed in ui-kit._

---

Storybook integration: #366 

#### Maintenance pages
- [x] `PageNotFound` (#316)
- [x] `ServicePageResponseLayout` (#316)
- [ ] `PageUnauthorized`

#### Dialogs
- [x] `InfoDialog` #367 
- [x] `ConfirmationDialog` #377 
- [x] `FormDialog` #399 

#### Page layouts
_These components should be implemented carefully, considering the use cases we've encountered so far. Most likely they are composed by lower-lever components that can be reused as well for special cases._
- [ ] `TabPage`
- [x] `InfoModalPage`, `FormModalPage`, `TabularModalPage` #642 
- [ ] `ListPage`

#### Controls
- [ ] `Pagination`
- [ ] `WarnOnLeave`

montezume:
Thanks for keeping track of this! 🙌

emmenko:
FYI: I started drafting a `ModalPage` component that we will try out first internally in the apps. If everything works fine, we will then move it here in the app-kit.

mmaltsev-ct:
The ticket is not required anymore. Some tasks have already separate tickets in the ui-kit and app-kit repos. In case of getting back to the migration topic, the review is advised. Hence, closing.

PR #367, opened 2019-02-15
feat: add <InfoDialog>

emmenko:
Based on #366 

Refs #323


packages/application-components/src/components/dialogs/info-dialog/info-dialog.story.js
```diff
@@ -0,0 +1,97 @@
...
+import {
+  SecondaryButton,
+  Spacings,
+  Text,
+} from '@commercetools-frontend/ui-kit';
+import Readme from './README.md';
+import InfoDialog from './info-dialog';
+
+const DialogController = props => {
+  const [isOpen, toggle] = React.useState(false);
```
tdeekens: I wouldn't do this even though I like it. That means anybody building an app with our application-components needs to be on the latest React version. Not that it's unlikely but it's a bit of a harsh restriction balancing the benefit of it (mainly being funky) 😄

packages/application-components/src/components/dialogs/info-dialog/info-dialog.story.js
```diff
@@ -0,0 +1,97 @@
...
+  return (
+    <div
+      style={{
+        display: 'grid',
+        alignItems: 'center',
+        justifyContent: 'center',
+        height: '100%',
+        width: '100%',
+      }}
```
tdeekens: Sorry, can we extract this? Either in a .mod.css or emotion. I don't feel like we have a rule for inline styles apart from: "it's not that much this time is it".

tdeekens:
@emmenko I would think it being helpful for other teams to also review this? So they know about the package, what's going on etc? Maybe just assign somebody from another team?

packages/application-components/src/components/dialogs/info-dialog/info-dialog.js
```diff
@@ -0,0 +1,68 @@
...
+    contentLabel="info-dialog"
+  >
+    <Constraints.Horizontal constraint={props.horizontalConstraint}>
+      <Card>
+        <Spacings.Stack scale="m">
+          <Stretched>
+            <Text.Subheadline
+              elementType="h4"
+              truncate={true}
+              title={typeof props.title === 'string' ? props.title : undefined}
```
tdeekens: Why do we need this check?

packages/application-components/src/components/dialogs/info-dialog/info-dialog.story.js
emmenko: This is just the story for storybook, I don't think it's a problem to be honest.

packages/application-components/src/components/dialogs/info-dialog/info-dialog.story.js
emmenko: We plan to migrate to emotion, so this is just temporary.

packages/application-components/src/components/dialogs/info-dialog/info-dialog.js
emmenko: The `props.title` can be any react node, whereas the `Text.title` needs to be a string (it's the HTML `title` attribute)

packages/application-components/src/components/dialogs/info-dialog/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+| Props                  | Type       | Required | Values   | Default | Description                                                                                                                 |
+| ---------------------- | ---------- | :------: | -------- | ------- | --------------------------------------------------------------------------------------------------------------------------- |
+| `title`                | `node`     |    ✅    | -        | -       | The title of the Info Dialog                                                                                                |
+| `isOpen`               | `boolean`  |    ✅    | -        | -       | Indicates whether the dialog should open or not                                                                             |
+| `onClose`              | `function` |    ✅    | -        | -       | Called when the dialog closes (click on overlay, click on close button, press ESC)                                          |
+| `closeTimeoutMS`       | `number`   |    -     | -        | `150`   | The timeout in ms for the dialog to close                                                                                   |
```
montezume: can we rename this to `closeAfter`?

packages/application-components/src/components/dialogs/info-dialog/README.md
```diff
@@ -0,0 +1,41 @@
...
+  </Spacings.Stack>
+</InfoDialog>
+```
+
+#### Properties
+
+| Props                  | Type       | Required | Values   | Default | Description                                                                                                                 |
+| ---------------------- | ---------- | :------: | -------- | ------- | --------------------------------------------------------------------------------------------------------------------------- |
+| `title`                | `node`     |    ✅    | -        | -       | The title of the Info Dialog                                                                                                |
+| `isOpen`               | `boolean`  |    ✅    | -        | -       | Indicates whether the dialog should open or not                                                                             |
```
montezume: I'm not really sure what this prop means. Does passing `isOpen` mean that it has to be controlled by the prop? Or does passing `{isOpen=true}` set a kind of "initial open state", where afterwards the opening and closing is managed by the component and not the parent?

packages/application-components/src/components/dialogs/info-dialog/README.md
```diff
@@ -0,0 +1,41 @@
...
+    </div>
+  </Spacings.Stack>
+</InfoDialog>
+```
+
+#### Properties
+
+| Props                  | Type       | Required | Values   | Default | Description                                                                                                                 |
+| ---------------------- | ---------- | :------: | -------- | ------- | --------------------------------------------------------------------------------------------------------------------------- |
+| `title`                | `node`     |    ✅    | -        | -       | The title of the Info Dialog                                                                                                |
```
montezume: is this a node to make it easier to pass in `react-intls` formatted message? Can we make it a `string`?

packages/application-components/src/components/dialogs/info-dialog/info-dialog.js
montezume: I think it would be cleaner to just force people to pass in strings.

packages/application-components/src/components/dialogs/info-dialog/info-dialog.js
```diff
@@ -0,0 +1,68 @@
...
+            >
+              {props.title}
+            </Text.Subheadline>
+            <SecondaryIconButton
+              label="Close dialog"
```
montezume: shouldn't this be translated?

packages/application-components/src/components/dialogs/info-dialog/info-dialog.js
```diff
@@ -0,0 +1,68 @@
...
+);
+InfoDialog.displayName = 'InfoDialog';
+InfoDialog.propTypes = {
+  title: PropTypes.node.isRequired,
+  children: PropTypes.node.isRequired,
+  horizontalConstraint: PropTypes.oneOf(['m', 'l']),
+  // React modal props
+  isOpen: PropTypes.bool.isRequired,
+  onClose: PropTypes.func.isRequired,
+  closeTimeoutMS: PropTypes.number,
```
montezume: nitpick: this could be required since it has a default?

packages/application-components/src/components/dialogs/info-dialog/info-dialog.story.js
montezume: I also don't mind using inline styles in stories.

packages/application-components/src/components/dialogs/info-dialog/README.md
emmenko: Sure (the name is the same as the prop in `react-modal` btw)

packages/application-components/src/index.js
```diff
@@ -1,6 +1,9 @@
+// Maintenance pages
 export {
   default as MaintenancePageLayout,
   DeprecatedServicePageResponseLayout as ServicePageResponseLayout,
 } from './components/maintenance-page-layout';
-
 export { default as PageNotFound } from './components/page-not-found';
+
+// Dialogs
+export { default as InfoDialog } from './components/dialogs/info-dialog';
```
montezume: nice! not forgetting the export 💯

packages/application-components/src/components/dialogs/info-dialog/README.md
emmenko: It's a controlled component, I mentioned that in the README description. How can we make it more clear?

packages/application-components/src/components/dialogs/info-dialog/README.md
montezume: Yeah I saw that. I figured that we aren't married to the idea of keeping our component in sync with `react-modal`. Maybe in the future we migrate aware from `react-modal`. It makes more sense to keep our props consistent

packages/application-components/src/components/dialogs/info-dialog/README.md
emmenko: I thought about it and decided to keep the API open for now. In theory, we should be able to pass a string, but I'm not sure if we might need to pass a react component from time to time.
I can double check though. For me it's also fine to restrict it to a string.

packages/application-components/src/components/dialogs/info-dialog/info-dialog.js
emmenko: Maybe...maybe not. It's just for a11y.

packages/application-components/src/components/dialogs/info-dialog/README.md
montezume: `Indicates the initial open state of the dialog`?

emmenko: Hmm wait, the `isOpen` state needs to be managed outside of the dialog component. There is no "initial" state.

montezume: `Indicates whether the dialog is open or closed` then?

packages/application-components/src/components/dialogs/info-dialog/info-dialog.js
```diff
@@ -0,0 +1,68 @@
...
+  <div className={styles.stretched}>{props.children}</div>
+);
+Stretched.displayName = 'Stretched';
+Stretched.propTypes = { children: PropTypes.node.isRequired };
+
+const InfoDialog = props => (
+  <ModalContainer
+    isOpen={props.isOpen}
+    closeTimeoutMS={props.closeTimeoutMS}
+    onRequestClose={props.onClose}
```
montezume: we have onClose here but also in the secondary icon button. Does that mean it will be called twice? Once immediately, and once after 150ms?

emmenko: No, the dialog can be closed in 3 ways, all of them map to the `onClose function`
- clicking on the overlay (modal triggers `onRequestClose`)
- pressing ESC (modal triggers `onRequestClose`)
- clicking on the close button (secondary button triggers `onClick`)

montezume: In this case should we remove the leave delay functionality since it doesn't apply when you click the secondary button? Or add a setTimeout so it matches the other cases?

packages/application-components/src/components/dialogs/info-dialog/info-dialog.story.js
tdeekens: True. Missed that. All good then. But it is worthwhile being on the lookout 😁

packages/application-components/src/components/dialogs/info-dialog/info-dialog.js
emmenko: Hmm good point, I haven't thought about it. I guess we can remove it then, I don't see much the value on having that.

packages/application-components/src/components/dialogs/info-dialog/info-dialog.js
tdeekens: I am fine with having it untranslated for now.

packages/application-components/src/components/dialogs/info-dialog/info-dialog.js
```diff
@@ -0,0 +1,65 @@
...
+          </Stretched>
+          <div className={styles.content}>{props.children}</div>
+        </Spacings.Stack>
+      </Card>
+    </Constraints.Horizontal>
+  </ModalContainer>
+);
+InfoDialog.displayName = 'InfoDialog';
+InfoDialog.propTypes = {
+  title: PropTypes.string.isRequired,
```
tdeekens: Should we make this a `node`? This would require anybody to use `formatMessage` and apply the HoC over just rendering a `<FormattedMessage />`?

emmenko: I had it as a `node` initially but @montezume asked to make it a `string`, because the value is also used for the HTML `title` attribute (which can only be a string).

tdeekens: Okay, sorry missed that. All good then.

packages/application-components/src/components/dialogs/info-dialog/info-dialog.js
```diff
@@ -0,0 +1,65 @@
...
+  Constraints,
+  SecondaryIconButton,
+  Spacings,
+  Text,
+} from '@commercetools-frontend/ui-kit';
+import ModalContainer from '../modal-container';
+import styles from './info-dialog.mod.css';
+
+// Like "Spacings.Inline", but with the `justify-content: space-between` style
+const Stretched = props => (
```
qmateub: Any plans on having this in UI Kit at some point? I can verify that we will remove A LOT of custom styles in the MC 🤓 cc/ @montezume

packages/application-components/src/components/dialogs/info-dialog/info-dialog.js
```diff
@@ -0,0 +1,65 @@
...
+        </Spacings.Stack>
+      </Card>
+    </Constraints.Horizontal>
+  </ModalContainer>
+);
+InfoDialog.displayName = 'InfoDialog';
+InfoDialog.propTypes = {
+  title: PropTypes.string.isRequired,
+  children: PropTypes.node.isRequired,
+  horizontalConstraint: PropTypes.oneOf(['m', 'l']),
```
qmateub: So, we only support `m` and `l` sizes here? Is it a design requirement? I'm saying because I think the original component does not have any of that right? 🤔

packages/application-components/src/components/dialogs/info-dialog/info-dialog.spec.js
```diff
@@ -0,0 +1,49 @@
...
+
+    fireEvent.click(getByLabelText(/Open Info Dialog/));
+    await wait(() => {
+      expect(queryByText(/Lorem ipsus/)).toBeInTheDocument();
+    });
+    expect(queryByText(/Hello/)).toBeInTheDocument();
+
+    fireEvent.click(getByLabelText(/Close dialog/));
+    await wait(() => {
+      expect(queryByText(/Lorem ipsus/)).not.toBeInTheDocument();
```
qmateub: Oh! I didn't know that you can await and do an `expect` 😅 . I will keep this in mind for future test writes 👍

packages/application-components/src/components/dialogs/info-dialog/info-dialog.story.js
qmateub: Our first hook component 🥺

packages/application-components/src/components/dialogs/info-dialog/info-dialog.js
montezume: @qmateub you can has an issue create 🐕

emmenko: I guess we can also extend the `Spacings.Inline` component to support this mode

packages/application-components/src/components/dialogs/info-dialog/info-dialog.js
emmenko: Yes, it's a design requirement

packages/application-components/src/components/dialogs/info-dialog/info-dialog.spec.js
emmenko: Yeah, this comes quite in handy

emmenko:
FYI: I'll probably wait for a couple of fixes in ui-kit before merging this

PR #375, opened 2019-02-18
Add Percy VRT tests

montezume:
#### Summary

Adds a `visual-testing-app` to the app-kit. Any file ending in `.visualroute` will be build into the `visual-testing-app`, just like in `ui-kit`.

Adds a visual test for the `application-components` component `PageNotFound`.

Adds a conditional running logic for Percy based off of whether there are changes in `application-components`, `rollup.config.js` or `yarn.lock`.


scripts/run-percy.js
```diff
@@ -0,0 +1,32 @@
...
+const changedFiles = getChangedFiles(range);
+
+const rulesGlob = [
+  'packages/application-components/**',
+  'rollup.config.js',
+  'yarn.lock',
+];
+
+const diff = multimatch(changedFiles, rulesGlob);
+console.log('diff', diff);
```
tdeekens: :)

montezume: debugging on percy is not fun

scripts/run-percy.js
```diff
@@ -0,0 +1,32 @@
+/* eslint-disable no-console */
+
+const multimatch = require('multimatch');
+const shelljs = require('shelljs');
+
+const getChangedFiles = range => {
```
emmenko: In case it helps, this is how we used to do it on CircleCI:

```bash
hasChangesInMatchingFiles() {
  # Extract commit range
  # (`cut -d/ -f7` means "split by `/` and pick the 7th field in the list")
  # E.g. https://github.com/commercetools/merchant-center-frontend/compare/1945a7a7a011...3ec9fece99cf
  #   => 1945a7a7a011...3ec9fece99cf
  COMMIT_RANGE=$(echo "${CIRCLE_COMPARE_URL}" | cut -d/ -f7)

  # NOTE: unfortunately in case of force pushes (e.g. rebases) the commit range is useless
  # as the checked out code does not contain the original commit.
  # The error you would see contains `unknown revision or path not in the working tree.`
  # https://github.com/travis-ci/travis-ci/issues/2668
  # In this case we check if git considers the value to be a valid range. If that's the
  # case the evaluated value would be `0`, otherwise something like `128`.
  IS_COMMIT_RANGE_VALID=$(git diff "$COMMIT_RANGE" &> /dev/null; echo $?)

  # Furthermore, we check if the range is actually a range, as sometimes the `CIRCLE_COMPARE_URL`
  # can be a single commit. In this case we also consider it to not be a valid range.
  if [[ $COMMIT_RANGE != *"..."* ]]; then
    IS_COMMIT_RANGE_VALID=9 # some custom exit code that is not `0`
  fi

  if [[ $IS_COMMIT_RANGE_VALID != 0 ]]; then
    echo "CircleCI commit range diff failed, probably due to a new PR or a force push (rebase), falling back to single commit"
    if git diff-tree --no-commit-id --name-only -r "$CIRCLE_SHA1" |grep --quiet -E "$MATCHING_PATH_TRIGGERS"; then
      return 0
    else
      return 1
    fi
  else
    if git diff --no-commit-id --name-only "$COMMIT_RANGE" |grep --quiet -E "$MATCHING_PATH_TRIGGERS"; then
      return 0
    else
      return 1
    fi
  fi
}
```

visual-testing-app/webpack.config.js
```diff
@@ -0,0 +1,75 @@
...
+            presets: [
+              require.resolve('@commercetools-frontend/babel-preset-mc-app'),
+            ],
+          },
+        },
+      },
+      {
+        test: /\.svg$/,
+        use: {
+          loader: require.resolve('svg-url-loader'),
```
montezume: we need this because `PageNotFound` is doing 

```
import PageNotFoundSVG from '@commercetools-frontend/assets/images/page-not-found.svg';
```

So the file isn't actually base64 encoded yet.

visual-testing-app/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "visual-testing-app",
```
emmenko: It probably makes sense to include this in the yarn workspace, otherwise it needs to be installed separately

packages/application-components/test-utils/visual/spec.js
```diff
@@ -0,0 +1,122 @@
...
+  omitPropsList: PropTypes.bool,
+};
+
+Spec.defaultProps = {
+  omitPropsList: false,
+};
+
+Spec.displayName = 'Spec';
+
+export default Spec;
```
tdeekens: Wondering, and maybe a premature thought, if this could be a package (or even repo) so it's shared between UIKit and app-kit?

montezume: Yeah I want to make this a repo but for now I think it makes sense just to copy paste it

scripts/run-percy.js
montezume: Didn't we have a lot of problems with that though and ended up removing it?

visual-testing-app/webpack.config.js
tdeekens: Can we use svgr here as we use that elsewhere too?

visual-testing-app/package.json
montezume: oh that's the problem I was having, thanks!

packages/application-components/test-utils/visual/spec.js
tdeekens: Yeah sure. Just came to mind.

visual-testing-app/webpack.config.js
montezume: Nah, we use SVGR just for SVGs ending in `.react.svg`. Generally icons.

scripts/run-percy.js
emmenko: We removed it for other reasons. My point was more about the edge cases to consider when trying to extract the commit logs.

montezume: Yeah I think I'll need some help with this part

.percy.yml
```diff
@@ -0,0 +1,4 @@
+version: 1
+snapshot:
+  widths: [1024]
```
montezume: this isn't working

the docs say
```Percy Agent is primarily configured by creating a .percy.yml file. This yaml file allows you to globally set configuration for each build. The .percy.yml configuration file must exist in the directory where you run the percy from.

```

Maybe something to do with yarn workspaces? 🤔

tdeekens: Ah, so it doesnt look for it upwards?

emmenko:
@montezume should we have a look at it tomorrow together?

package.json
```diff
@@ -25,7 +25,7 @@
     "test:watch": "jest --config jest.test.config.js --watch",
     "test:visual": "jest --config jest.visual.config.js --runInBand",
     "run-if:percy": "node scripts/run-percy.js",
-    "percy": "percy exec -- jest --config jest.visual.config.js --runInBand",
+    "vrt": "percy exec -- jest --config jest.visual.config.js --runInBand",
```
tdeekens: What would speak against keeping some "command consistency" across UIKIt and app-kit?

emmenko: I wanted to inspect the percy CLI command by doing `yarn run percy --help` and instead this command was executed 🙃

image

tdeekens: Ah okay. Nice catch. Maybe we can change the command at UIKit then too to avoid this? 👍

PR #379, opened 2019-02-20
chore: comment out failing test

montezume:
This part of the `QuickAccess` test has been failing locally (but passing on CI...) for some time. Both @emmenko and I looked into it and we couldn't figure out why. At this point, I'm not sure it makes sense to continue spending time on it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 sudo: false
+os: osx
 
 addons:
   apt:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,4 +1,5 @@
 sudo: false
+os: osx
```
tdeekens: What is this solving/doing? :)

montezume: Trying to figure out if this would make the test that fails locally also fail on CI 🤷‍♂️

PR #378, opened 2019-02-20
chore: update to ui-kit v9.0.0

emmenko:
Nothing to be migrated in the app-kit

PR #380, opened 2019-02-20
fix: broken layouts due to Card component structure changes

emmenko:
The last ui-kit update #378 caused the login form layout to be messed up a little, specifically the Card component having a child `div` for the padding. The login box was injecting some custom style to the Card component to reset some styles...hacky stuff.

I refactored the layout a bit to use as much ui-kit components as possible (/cc @jonnybel this might be interesting for you for the new login screens).

image


emmenko:
I pushed a couple more fixed that I saw while testing things around.

packages/application-components/src/components/dialogs/confirmation-dialog/README.md
```diff
@@ -35,7 +35,7 @@ Confirmation dialogs are controlled components used to prompt the user to confir
...
 | `children`                      | `node`                     |    ✅    | -                 | -                                 | Content rendered within the dialog. If the content is long in height (depending on the screen size) a scrollbar will appear                                                                                                     |
-| `horizontalConstraint`          | `string`                   |          | `m`, `l`, `scale` | `l`                               | Horizontal width limit of the dialog card                                                                                                                                                                                       |
+| `size`                          | `string`                   |          | `m`, `l`, `scale` | `l`                               | Horizontal width limit of the dialog card                                                                                                                                                                                       |
```
emmenko: The `Constraints.Horizontal` component actually sets a `max-width`, but for the modal sizes we need an actual `width`, hence this change

packages/application-components/src/utils/get-display-name.js
```diff
@@ -0,0 +1,14 @@
+// From https://github.com/acdlite/recompose/blob/master/src/packages/recompose/getDisplayName.js
```
emmenko: Just to avoid having a dependency on `recompose`

packages/application-shell/src/components/user-settings-menu/user-settings-menu.js
```diff
@@ -95,7 +94,7 @@ const UserSettingsMenuBody = props => {
     [];
 
   return (
-    <Card className={styles.menu}>
```
emmenko: This `Card` component was basically useless, as the `.menu` styles were overwriting all of the card styles 🙃

PR #381, opened 2019-02-21
fix(app-shell): quick-access to support full reloads

tdeekens:
#### Summary

This pull request adds the feature (needed as a fix) for QuickAccess to support redirects with full page reloads. 

When QuickAccess is rendered in the application-fallback we can not render a `RouteCatchAll` as a result never triggering a full page reload from there to any other app. As a result we see the 404 page served by the application-fallback app itself.

We have the notion of `INTERNAL__isApplicationFallback` the NavBar already uses for these cases. We use the same concept, passing it the same as `useFullRedirectsForLinks` to `open` instead of `push` to history.


tdeekens:
Fix or feature is debatable but I think it's nothing public facing so we can keep it a fix.

PR #383, opened 2019-02-21
feat(app-shell/login): Login page redesign

jonnybel:
#### Summary

Implementation of the Log In page redesign.

Added `@emotion/core`.

image

Some text is still not updated because of the translations.


package.json
```diff
@@ -40,7 +40,9 @@
   "devDependencies": {
     "@commitlint/cli": "7.5.2",
     "@commitlint/config-conventional": "7.5.0",
+    "@emotion/babel-preset-css-prop": "^10.0.7",
```
montezume: can you use the exact versions without the `^`?

packages/babel-preset-mc-app/index.js
```diff
@@ -61,6 +61,13 @@ module.exports = function getBabePresetConfigForMcApp() {
           useBuiltIns: true,
         },
       ],
+      [
+        '@emotion/babel-preset-css-prop',
+        {
+          sourceMap: isEnvDevelopment,
```
montezume: nice 💯

packages/application-shell/package.json
```diff
@@ -52,6 +52,7 @@
     "@commercetools-frontend/sentry": "8.0.0",
     "@commercetools-frontend/storage": "8.0.0",
     "@commercetools-frontend/url-utils": "8.0.0",
+    "@emotion/core": "^10.0.7",
```
tdeekens: 💯

packages/babel-preset-mc-app/index.js
montezume: I'm wondering though if it makes sense to have a separate babel config for mc "libraries" 🤔

packages/i18n/data/core.json
```diff
@@ -37,23 +39,27 @@
...
   "LoginForm.forgotPasswordTitle": "We are redirecting you to the Forgot Password form in {countdown} seconds.",
   "LoginForm.InvalidCredentials": "Sorry, but the login details entered below are not in our system. Please enter your correct email address and password below.",
   "LoginForm.invalidEmail": "Please enter a valid Email address.",
   "LoginForm.logout": "We have logged you out of the Merchant Center. If you would like to log back in, please fill out the form below.",
   "LoginForm.noProjectsLink": "set up a project in the current Merchant Center",
   "LoginForm.noProjectsTitle": "Sorry, but you do not have a project associated to your account and we are therefore unable to log you in. Please {link} before trying again.",
   "LoginForm.password": "Password",
-  "LoginForm.signin": "Sign in",
-  "LoginForm.title": "Sign in to your account",
+  "LoginForm.signin": "Log in",
```
montezume: login is the noun, "log in" is the verb. I think we want the noun here, since "Password" is also a noun.

packages/i18n/data/core.json
```diff
@@ -37,23 +39,27 @@
...
+  "LoginForm.title": "Log in",
   "LoginForm.unauthorized": "Your session has expired. Please sign in again to continue using the Merchant Center.",
   "LoginForm.validating": "Validating...",
   "LoginLocked.helpDeskLink": "Help Desk",
   "LoginLocked.paragraph1": "For security reasons, your account has been locked because of too many failed log-in attempts. You can {resetPasswordLink} or contact the commercetools {helpDeskLink}",
   "LoginLocked.resetPasswordLink": "reset your password",
   "LoginLocked.title": "Account Locked",
-  "LoginPage.privacyPolicy": "Privacy Policy",
+  "LoginPage.privacyPolicyLink": "Privacy Policy",
+  "LoginPage.termsOfServiceLink": "Terms of Service",
```
montezume: why do we use german capitalization in English 🤦‍♂️

packages/application-shell/src/components/public-page-container/public-page-container.js
```diff
@@ -1,31 +1,96 @@
...
-            </Text.Detail>
-          </a>
-          <Text.Detail>{`${year} © commercetools`}</Text.Detail>
-        </Spacings.Inline>
-      </Spacings.Stack>
-    </div>
-  </React.Fragment>
-);
+const PublicPageContainer = props => {
+  const privacyPolicyLink = (
```
tdeekens: This is a bit odd maybe. Maybe make then React components in the same file? `PrivacyPolicyLink` and `TermsOfServiceLink`?

packages/application-shell/src/components/public-page-container/public-page-container.js
```diff
@@ -1,31 +1,96 @@
...
+  return (
+    <React.Fragment>
+      <PortalsContainer />
+      <div
+        css={css`
+          height: 100vh;
+          display: flex;
+          padding-top: ${customProperties.spacing32};
+          justify-content: center;
+          background-color: #073538;
```
tdeekens: I guess this is just trying to match the bg-image?

packages/application-shell/src/components/public-page-container/public-page-container.js
```diff
@@ -1,31 +1,96 @@
...
+      <FormattedMessage {...messages.termsOfServiceLink} />
+    </a>
+  );
+
+  return (
+    <React.Fragment>
+      <PortalsContainer />
+      <div
+        css={css`
+          height: 100vh;
```
montezume: are we sure we want 100vh? won't the page scroll then? wouldn't it make more sense to subtract the height of the top bar?

packages/application-shell/src/components/public-page-container/public-page-container.js
```diff
@@ -1,31 +1,96 @@
...
+                <span style={{ color: 'white' }}>
+                  <FormattedMessage {...messages.welcome} />
+                </span>
+              </Text.Headline>
+            </div>
+          </Spacings.Stack>
+          <Spacings.Stack scale="s">
+            {/* Login Card */}
+            {props.children}
+            <div
```
tdeekens: Can you explain why this wrapper is needed and Spacings.Stack can't be used with the same props?

packages/application-shell/src/components/public-page-container/public-page-container.js
montezume: maybe extract this into a js var

packages/application-shell/src/components/public-page-container/public-page-container.js
```diff
@@ -1,31 +1,96 @@
...
+                    <span style={{ color: 'white' }}>
+                      <FormattedMessage
+                        {...messages.termsAndPrivacy}
+                        values={{ privacyPolicyLink, termsOfServiceLink }}
+                      />
+                    </span>
+                  </Text.Detail>
+                </div>
+                <Text.Detail>
+                  <span style={{ color: 'white' }}>
```
montezume: can you use css in js here instead of inline style?

packages/application-shell/src/components/public-page-container/public-page-container.js
```diff
@@ -1,31 +1,96 @@
...
+              css={css`
+                display: flex;
+                align-items: center;
+                justify-content: space-between;
+              `}
+            >
+              <Spacings.Stack scale="xs">
+                <div>
+                  <Text.Detail>
+                    <span style={{ color: 'white' }}>
```
montezume: css in js again

packages/application-shell/src/components/public-page-container/public-page-container.js
```diff
@@ -1,31 +1,96 @@
...
+            <img
+              css={css`
+                width: 275px;
+                height: 45px;
+              `}
+              src={LogoBWWhiteSVG}
+            />
+            <div>
+              <Text.Headline elementType="h2">
+                <span style={{ color: 'white' }}>
```
montezume: css in js

packages/application-shell/src/components/public-page-container/public-page-container.js
emmenko: Or simply inline them to the message values

packages/application-shell/src/components/login/login.js
```diff
@@ -170,70 +171,78 @@ export class Login extends React.PureComponent {
           return this.handleSubmit(values, addHandlers);
         }}
         render={formikProps => (
-          <LoginBox>
```
emmenko: Can we try to keep this?

I'm not sure if you know but we also have a second login screen for SSO (`/login/sso`)

packages/application-shell/src/components/login/login.js
```diff
@@ -170,70 +171,78 @@ export class Login extends React.PureComponent {
...
+          <Card
+            css={css`
+              width: 330px;
```
emmenko: Curious why we can't use the `m` constraint here from the design tokens? 🤔

(it's a question for the design team)

packages/application-shell/src/components/public-page-container/public-page-container.js
tdeekens: It's feels to me like this is a hack around the fact that our Text components can't do white colours. Should they maybe?

packages/application-shell/src/components/login/login.js
```diff
@@ -170,70 +171,78 @@ export class Login extends React.PureComponent {
...
+                      value={formikProps.values.password}
+                      title={this.props.intl.formatMessage(messages.password)}
+                      onBlur={formikProps.handleBlur}
+                      errors={formikProps.errors.password}
+                      touched={formikProps.touched.password}
+                      onChange={formikProps.handleChange}
+                      isRequired={true}
+                      isDisabled={formikProps.isSubmitting}
+                    />
+                    <Countdown
```
emmenko: This can be removed now that we will have a an actual page for reset password

packages/application-shell/src/components/public-page-container/public-page-container.js
jonnybel: Yes, I added these before adding `emotion` and they're only here as a "hack" until we have proper theming.
Maybe I should just use css in js and add a comment to remind to remove the hack later?

emmenko: There is an `inverted` tone for the white color

montezume: can't you just do at the top of your file (outside the react component)
```
const white = css`color: white`
```

And then do css={white} on the spans?

packages/application-shell/src/components/login/login.js
jonnybel: I talked with Filip and this is really supposed to be a different value from the constraints.

emmenko: I'll talk to him...why the hell do we have the design tokens for?!

packages/application-shell/src/components/public-page-container/public-page-container.js
montezume: yeah you can use `<Text.Body tone="inverted" >`

packages/application-shell/src/components/public-page-container/public-page-container.js
jonnybel: Yes, tho I'm not sure if I should keep it. It's just a color picked from the current background image to reduce the contrast while it loads.

packages/application-shell/src/components/public-page-container/public-page-container.js
jonnybel: Now that I took another look, I think this is a leftover 😅

packages/application-shell/src/components/public-page-container/public-page-container.js
jonnybel: Ah cool, I wasn't aware!

packages/babel-preset-mc-app/index.js
jonnybel: Well, to be honest I just ~~stole~~ borrowed these configs from the `ui-kit` 😁

packages/i18n/data/core.json
jonnybel: Agreed

packages/application-shell/src/components/public-page-container/public-page-container.js
montezume: Too bad we don't have gatsby-image style image loading 🤔

packages/application-shell/src/components/login/login.js
jonnybel: Do we really need it tho? I realized I didn't need it because it was just an old version of the Card with overridden styles, but the new designs are totally compatible with just using the `Card` component as-is.

packages/application-shell/src/components/public-page-container/public-page-container.js
jonnybel: Hmm, but there is no top bar at the public page, or am I missing something?

packages/application-shell/src/components/public-page-container/public-page-container.js
jonnybel: Yeah this was unnecessary

packages/application-shell/src/components/public-page-container/public-page-container.js
jonnybel: Is there any reason why only the `Text.Body` component has the "inverted" tone? (and not also the Headline, Subheadline and Detail)

packages/application-shell/src/components/login/login.js
jonnybel: But should we remove this in this PR or only when we actually have the page?

packages/application-shell/src/components/public-page-container/public-page-container.js
montezume: 🤷‍♂️ . If there's a use case for it, we can add it. I doubt there is though.

jonnybel: Just the Headline of this page. Right now it's solved with the styled span wrapper

montezume: If it's just one in one place in the whole MC, I would stick with your approach 👍

emmenko:
Closing this as we're working on that in a separate app.

PR #384, opened 2019-02-21
feat(components/dialogs): make overlay z-index configurable

emmenko:
As I was analyzing places in the mc apps that make use of modals/dialogs, I noticed that in some case the `z-index` of the overlay container needs to be increased.
This can be the case if you have **stacking modals**.

Therefore, we can pass a `zIndex` prop that will be injected to the `overlay` container

screenshot 2019-02-21 at 18 43 54


emmenko:
> Maybe we could expose, if needed, a level as enum? low, mediun, high?

I thought about it but it's hard to have predefined values that make sense.

> Should it then maybe only increased or is there a case where it should be behind something?

Maybe...but even then if you put a low level you will probably notice that it won't work because you won't see anything.

I think for now I would like to keep it "flexible" and eventually apply more restrictions as we cover more use cases.

tdeekens:
> probably notice that it won't work because you won't see anything.

Yes, that's why I thought that we could just bump the index. A modal should always be in front of everything all the time anyways?

> I think for now I would like to keep it "flexible" and eventually apply more restrictions as we cover more use cases.

Sure, but changing the API would be a breaking change. That's why I wanted to see/discuss if we actually need it.

> I thought about it but it's hard to have predefined values that make sense.

To me it sounds like the zindex-wars for which people ended up just having low, mid, high. Given this is a public component and we can't really have expectations towards other peoples zIndex maybe keeping it like that is the only way to integrate with them.

packages/application-components/src/components/dialogs/internals/dialog-styles.mod.css
```diff
@@ -1,8 +1,5 @@
 :root {
   --color-modalBackground: rgba(32, 62, 72, 0.5);
-
-  /* Modals */
-  --z-index--overlay: 1000;
```
adnasa: 🔥

PR #385, opened 2019-02-22
chore(app-shell): expose ApplicationShellProvider

emmenko:
As we started working on a new app for the login routes etc, we need part of the AppShell setup, which is now exposed by `ApplicationShellProvider`


packages/application-shell/src/components/application-shell-provider/application-shell-provider.spec.js
```diff
@@ -0,0 +1,157 @@
...
+  });
+  it('should access redux store from context', () => {
+    let hasStore = false;
+    render(
+      <ApplicationShellProvider {...createTestProps()}>
+        {() => (
+          <ReactReduxContext.Consumer>
+            {({ storeState }) => {
+              if (storeState) {
+                hasStore = true;
```
tdeekens: Nifty.

Issue #100, opened 2018-11-14
[Tooling] Add script for checking pin dependencies


emmenko:
We should have the `throw-if-pinned-dependency-version-changes.js` script that we use in the MC FE in this repo as well and update the list accordingly.

tdeekens:
That'd be very reasonable. I wonder if that is actually a nice separate Open Source package.

emmenko:
Or to include it in like renovate & co.?

tdeekens:
After thinking about this some more I'd recommend making this a Probot over a script. Given it's a script it would have to integrate with Renovate and potentially any other service. Given it's a Probot it can hook into things on a GitHub level. Just like the "WIP bot". It would assign a label or generally just reject a review.

emmenko:
A bot would be cool yes. It's a bit more work though, so maybe the script as a first step is enough.

In any case, if someone wants to spend some time on this I'm all for it 💯

tdeekens:
Given that we removed the script from the mc-fe and Renovate is configured to not update deps I would go ahead and close this.

PR #395, opened 2019-02-25
chore: update deps

montezume:
* Closes #395 
* Closes #394 
* Closes #393 
* Closes #392 
* Closes #391 
* Closes #390


application-templates/starter/package.json
```diff
@@ -32,6 +32,6 @@
     "@commercetools-frontend/mc-scripts": "9.0.0",
     "enzyme": "3.9.0",
     "jest": "24.1.0",
-    "react-testing-library": "5.8.0"
+    "react-testing-library": "6.0.0"
```
montezume: just removes testHooks, which we don't use

packages/application-components/package.json
```diff
@@ -45,27 +45,27 @@
...
+    "react": "16.8.3",
+    "react-dom": "16.8.3",
     "react-intl": "2.8.0",
-    "react-testing-library": "5.8.0",
+    "react-testing-library": "6.0.0",
     "storybook-readme": "4.0.5"
   },
   "peerDependencies": {
-    "@commercetools-frontend/assets": "7.x",
+    "@commercetools-frontend/assets": "9.x",
```
montezume: not sure why this was pegged to version 7...

packages/application-shell-connectors/package.json
```diff
@@ -39,10 +39,10 @@
     "lodash-es": "4.17.11",
     "moment-timezone": "^0.5.23",
     "prop-types": "15.7.2",
-    "recompose": "0.26.0"
+    "recompose": "^0.30.0"
```
montezume: it's mostly performance optimizations, no breaking changes 🤔

packages/mc-scripts/package.json
```diff
@@ -47,7 +47,7 @@
     "optimize-css-assets-webpack-plugin": "5.0.1",
     "pegjs-loader": "0.5.4",
     "postcss": "7.0.14",
-    "postcss-color-mod-function": "2.4.3",
+    "postcss-color-mod-function": "3.0.3",
```
montezume: I can't find any **real** breaking changes. Mostly just dep updates, relying on postcss 7...

packages/application-shell-connectors/package.json
tdeekens: This is a bit dangerous. In the mc-fe this version breaks the Modified Products view. I never understood why. But I would recommend playing it safe and keep it as is. Maybe update in the mcfe first to ensure the version works. Whatever causs things to break...

montezume: thanks. Should we add a `throw-if-pinned-deps-changed` to app kit as well then?

tdeekens: Sure. I always wondered why it is not in yarn natively or why there isnt a pkg for this. Could just be an entry in the pkg.json called noUpdate or something 🙃

emmenko: FYI: there is an issue to work on that (#100)

As for `recompose`, I think we can completely remove it as we only use it for the display name.

I'll push a commit

emmenko: Actually I'll open a new PR, as there are a few more changes.

packages/application-components/package.json
emmenko: Maybe we can be a bit more loose with the assets package. Like e.g. `>=7`?

emmenko:
> As mentioned, I'll open a PR to remove `recompose`

Here we go! #400

PR #397, opened 2019-02-25
Update uglify plugin / stop shipping es6 code to ie

montezume:
#### The problem

Currently, our production bundles have code like this in it.

```js
const n=r("../node_modules/omit-empty/node_modules/kind-of/index.js");
```

```js
const a=(e,t)=>{let r=!!t&&t.omitZero```
```

This is bad because arrow functions are **NOT** supported by IE 11.

### The solution

Passing a new parameter to createWebpackConfigForProduction and compiling those deps.

Explicitly compiling dependencies that are shipping ES6 code with babel.

Here's a screenshot of trying to use the MC with IE11. Notice how it crashes and doesn't load the login page.

screen shot 2019-02-25 at 1 17 41 pm

### DIFF
--- a/application-templates/starter/webpack.config.prod.js
+++ b/application-templates/starter/webpack.config.prod.js
@@ -3,7 +3,7 @@ const createWebpackConfigForProduction = require('@commercetools-frontend/mc-scr
 
 const distPath = path.resolve(__dirname, 'dist');
 const entryPoint = path.resolve(__dirname, 'src/index.js');
-const vendorJsFolders = [/node_modules\/omit-empty/];
+const vendorsToTranspile = [/node_modules\/omit-empty/];
 
 const sourceFolders = [path.resolve(__dirname, 'src')];
 
@@ -11,5 +11,5 @@ module.exports = createWebpackConfigForProduction({
   distPath,
   entryPoint,
   sourceFolders,
-  vendorJsFolders,
+  vendorsToTranspile,
 });
--- a/application-templates/starter/webpack.config.prod.js
+++ b/application-templates/starter/webpack.config.prod.js
@@ -3,7 +3,9 @@ const createWebpackConfigForProduction = require('@commercetools-frontend/mc-scr
 
 const distPath = path.resolve(__dirname, 'dist');
 const entryPoint = path.resolve(__dirname, 'src/index.js');
-const sourceFolders = [path.resolve(__dirname, 'src')];
+const sourceFolders = [path.resolve(__dirname, 'src')].concat([
+  'node_modules/omit-empty',
+]);
 
 module.exports = createWebpackConfigForProduction({
   distPath,
--- a/application-templates/starter/webpack.config.prod.js
+++ b/application-templates/starter/webpack.config.prod.js
@@ -3,9 +3,10 @@ const createWebpackConfigForProduction = require('@commercetools-frontend/mc-scr
 
 const distPath = path.resolve(__dirname, 'dist');
 const entryPoint = path.resolve(__dirname, 'src/index.js');
-const sourceFolders = [path.resolve(__dirname, 'src')].concat([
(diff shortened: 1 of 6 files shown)


packages/mc-scripts/config/polyfills.js
```diff
@@ -10,7 +10,7 @@
  *    This is why `unfetch` is still used as a polyfill. Overwriting or
  *    stubbing a ponyfill is not possible as JavaScript modules are read-only.
  */
-import 'unfetch/polyfill';
+import 'unfetch/polyfill/index';
```
montezume: using the default export means we use the `mjs` file, which has untranspiled code.

playground/webpack.config.prod.js
```diff
@@ -3,7 +3,9 @@ const createWebpackConfigForProduction = require('@commercetools-frontend/mc-scr
 
 const distPath = path.resolve(__dirname, 'dist');
 const entryPoint = path.resolve(__dirname, 'src/index.js');
-const sourceFolders = [path.resolve(__dirname, 'src')];
+const sourceFolders = [path.resolve(__dirname, 'src')].concat[
+  'node_modules/omit-empty'
```
montezume: this dependency ships untranspiled  code.

emmenko: Can you explain a bit more what the actual problem is? 🤔

montezume: I updated the description

montezume:
@tdeekens no, I mean we are shipping arrow functions and const to the browser, throwing errors on browsers that don't support this es6 syntex (IE ie11)

playground/webpack.config.prod.js
montezume: The problem is the MC doesn't work in IE11 because our bundle contains ES6 features like arrow functions.

montezume:
I think it's failing on CI because `playground` is relying on the older version of mc-scripts.

Can you help out @emmenko ?

montezume:
Okay It's passing now on CI. I would like to get this merged and cut a release to make MC work in IE 11 again.

application-templates/starter/webpack.config.prod.js
```diff
@@ -3,10 +3,13 @@ const createWebpackConfigForProduction = require('@commercetools-frontend/mc-scr
 
 const distPath = path.resolve(__dirname, 'dist');
 const entryPoint = path.resolve(__dirname, 'src/index.js');
+const vendorJsFolders = [/node_modules\/omit-empty/];
```
emmenko: Can we find a "better" name for this list? Like this, it looks like you need to pass all the vendor folders or something.

Also, we should somehow document this.

montezume: Yeah, do you have any suggestions?

As for documenting, we don't have any documentation for mc-scripts so I think we could make that another task right?

emmenko: > Yeah, do you have any suggestions?

Hmm, maybe `untranspiledVendors` or something like it?

> As for documenting, we don't have any documentation for mc-scripts so I think we could make that another task right?

I would be fine to have it in the webpack config for now. Just explain what the purpose of the field is and (e.g. link it to this PR).

montezume: vendorsToTranspile sounds better to me

I'm not sure what you mean by having it in the webpack config? We can't be sure of every vendor the apps are using...

montezume: Oh, you mean with just a code comment ?

emmenko: Yes the comment/doc in the webpack config

emmenko: > vendorsToTranspile sounds better to me

👌

PR #399, opened 2019-02-25
feat(components/dialogs): add FormDialog

emmenko:
This is basically a copy of the `ConfirmationDialog`. At least, visually is the same.

I was a bit unsure if we should have this or not but I think it makes sense semantically to have a distinction between "confirmation" and "form" dialog.
The only difference in the component API is the name of the button click handlers. I had a chat about it last week with @tdeekens and we agreed to have the props named a bit more consistently with the secondary/primary format.

In the ConfirmationDialog though, I still think it's ok to keep the click handlers as `onCancel/onConfirm`, because that's what the confirmation dialog is for.

I also added some more likely-to-be-used messages that people can use for the primary button.


emmenko:
It's just weird to use a `ConfirmationDialog` for form inputs. That's the main point, to have a distinction from a semantic point of view. I guess it's a bit like idk, a Text field and a Password field.

montezume:
PasswordFields have show / hide functionality though.

Why is it semantically weird to use `ConfirmationDialog` for form inputs?

montezume:
To me a `ConfirmationDialog` implies a choice between two options. Save and Cancel. Is that not what we want with a form as well? 🤔

emmenko:
Because the purpose is different. ConfirmationDialog should be used for prompting the user to confirm an action, EditDialog should be used as a dialog form.
Having this distinction might help us in the future to optimize them for different use cases.

What is your concern in having 2 different components?

emmenko:
After talking to @montezume we agreed to rename this `FormDialog`.
FYI: material-ui also has something similar for [dialogs](https://material-ui.com/demos/dialogs/)

PR #401, opened 2019-02-26
fix: add uglifyOptions

montezume:
This fixes the production build of most apps.

### DIFF
--- a/packages/mc-html-template/index.js
+++ b/packages/mc-html-template/index.js
@@ -12,7 +12,7 @@ module.exports = templateParams => {
 
   const cssChunks = cssVendorChunks.concat(cssAppChunks).map(fileName => {
     const chunkPath = fileName.replace(/^\//, '');
-    return `<link href="__CDN_URL__${chunkPath}" rel='stylesheet' type='text/css'></link>`;
+    return `<link href="__CDN_URL__${chunkPath}" rel='stylesheet' type='text/css'>`;
   });
   const scriptChunks = templateParams.htmlWebpackPlugin.files.js.map(
     fileName => {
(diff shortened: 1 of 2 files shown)


montezume:
I'm not sure if this is the issue. I'm having problems with one of the graphql requests returning HTML instead of JSON 🤔

packages/mc-scripts/config/create-webpack-config-for-production.js
```diff
@@ -22,6 +22,22 @@ const postcssColorModFunction = require('postcss-color-mod-function');
...
+    warnings: false,
+    parse: {},
+    compress: {},
+    mangle: true,
+    output: null,
+    toplevel: false,
+    nameCache: null,
+    ie8: false,
+    // https://github.com/mishoo/UglifyJS2/issues/1609
+    collapse_vars: false,
```
montezume: this fixing a problem in a few apps

packages/mc-html-template/index.js
```diff
@@ -12,7 +12,7 @@ module.exports = templateParams => {
 
   const cssChunks = cssVendorChunks.concat(cssAppChunks).map(fileName => {
     const chunkPath = fileName.replace(/^\//, '');
-    return `<link href="__CDN_URL__${chunkPath}" rel='stylesheet' type='text/css'></link>`;
+    return `<link href="__CDN_URL__${chunkPath}" rel='stylesheet' type='text/css'>`;
```
montezume: this doesn't give a warning in IE

packages/mc-scripts/config/create-webpack-config-for-production.js
emmenko: So we're adding back what we had before? 🤔

montezume: sort of, what we had before was for a different uglifer

packages/mc-html-template/index.js
montezume: https://stackoverflow.com/questions/9943981/can-self-closing-link-tags-be-problematic

montezume:
I'm going to close this for now in favour of patching the webpack config in the apps

PR #402, opened 2019-02-27
feat: remove pegjs loader from webpack config

montezume:
#### Summary

Pegjs is only used in discounts. Instead of including this loader in our webpack config, we should just do it in the discounts webpack config like so:

```js
// discounts/webpack.config.js

config.module.rules = config.module.rules.concat({
  test: /\.pegjs$/,
  use: [require.resolve('pegjs-loader')],
});

```
I also removed it from the jest config

The migration in apps that need pegjs will have to do something like this

```js
const preset = require('@commercetools-frontend/jest-preset-mc-app');

module.exports = {
  ...preset,
  transform: {
    ...preset.transform,
    '^.+\\.pegjs$': 'pegjs-jest',
  },
  // ...
};```

### DIFF
--- a/packages/jest-preset-mc-app/jest-preset.js
+++ b/packages/jest-preset-mc-app/jest-preset.js
@@ -34,7 +34,6 @@ module.exports = {
   transform: {
     '^.+\\.js$': resolveRelativePath('./transform-babel-jest.js'),
     '^.+\\.graphql$': 'jest-transform-graphql',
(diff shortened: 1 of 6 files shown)


montezume:
@gnerkus yeah, the migration would be.

* Merge this PR.
* Release version 11 of mc-scripts
* Make PR in MC that updates all apps to version 11 of mc scripts, and adds these lines to the discounts webpack configs.

PR #403, opened 2019-02-27
feat(babel-preset-mc-app): add babel support for emotion

jonnybel:
#### Summary

This PR add a babel preset to support `emotion` and [which includes `babel-plugin-emotion`](https://github.com/emotion-js/emotion/tree/master/packages/babel-preset-css-prop#usage).

### DIFF
--- a/packages/babel-preset-mc-app/index.js
+++ b/packages/babel-preset-mc-app/index.js
@@ -60,6 +60,13 @@ module.exports = function getBabePresetConfigForMcApp() {
           // behavior for any plugins that require one.
           useBuiltIns: true,
         },
+        [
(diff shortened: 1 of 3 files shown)


montezume:
Maybe instead we can look into a way of overriding the babel config from the app side?

montezume:
What if we allowed passing in a babelConfig into the webpack config?

emmenko:
Or we keep using css modules in the apps for now 😇

jonnybel:
Closing this, as we've decided for another approach.

emmenko:
As discussed with @jonnybel we can close this for now as emotion is not necessary anymore for the original purpose.

PR #404, opened 2019-02-27
chore: add sourcemap

montezume:
Sensible defaults for the minimizer.

### DIFF
--- a/packages/mc-scripts/config/create-webpack-config-for-production.js
+++ b/packages/mc-scripts/config/create-webpack-config-for-production.js
@@ -78,7 +78,15 @@ module.exports = ({
     // https://medium.com/webpack/webpack-4-mode-and-optimization-5423a6bc597a
     optimization: {
       minimizer: [
-        new UglifyJsPlugin({ sourceMap: true }),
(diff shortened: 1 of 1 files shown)


packages/mc-scripts/config/create-webpack-config-for-production.js
```diff
@@ -78,7 +78,15 @@ module.exports = ({
...
-        new UglifyJsPlugin(),
+        new UglifyJsPlugin({
+          uglifyOptions: {
+            compress: {
+              // seems to be caused by using `tether :(`
+              collapse_vars: false,
+            },
+          },
+          sourceMap: true,
+        }),
```
emmenko: Can we maybe check the "old" config and remove the options that we don't need anymore? I'm wondering if there are also other useful options that were useful but that now are not enabled anymore.

Or have you checked all the config options already?

montezume: sure

montezume: okay, pushed a commit to use the old config minus the ones that don't exist anymore.

PR #406, opened 2019-02-27
refactor(configure-apollo): add special mapping for DeliveryItem type

qmateub:
#### Summary

Adds a special mapping for `DeliveryItem` type of graphql. Basically it contains two attributes, `id` and `quantity`. But the `id` is really the line item id referenced in the delivery item, so once it gets in the apollo cache we get the wrong data. See screenshots:

|Response  | After apollo caches it |
|---|---|
|image  | image |

See how the `typeName` is wrong for the delivery item.

The documentation is also clear about this:

image

### DIFF
--- a/packages/application-shell/src/configure-apollo.js
+++ b/packages/application-shell/src/configure-apollo.js
@@ -66,6 +66,8 @@ const fragmentMatcher = new IntrospectionFragmentMatcher({
 
 const typeNamesWithoutIdAsIdentifier = ['Project', 'BaseMenu', 'NavbarMenu'];
 
+const typeNameWithReferencedEntity = ['Reference', 'DeliveryItem'];
+
 export const createApolloClient = () =>
   new ApolloClient({
     link,
@@ -82,7 +84,10 @@ export const createApolloClient = () =>
         // However, a reference has the shape { typeId, id } where the id is the
         // id of the referenced entity. If we didn't prefix ids of References
         // they would clash with the referenced resource.
-        if (result.__typename === 'Reference')
+        // The same use case happens for DeliveryItem type in which the `id`
+        // is really the lineItem id referenced in the delivery
+        // https://docs.commercetools.com/http-api-projects-orders.html#deliveryitem
+        if (typeNameWithReferencedEntity.includes(result.__typename))
           return `${result.__typename}:${result.id}`;
 
         return result.id;


packages/application-shell/src/configure-apollo.js
```diff
@@ -82,7 +84,10 @@ export const createApolloClient = () =>
         // However, a reference has the shape { typeId, id } where the id is the
         // id of the referenced entity. If we didn't prefix ids of References
         // they would clash with the referenced resource.
-        if (result.__typename === 'Reference')
+        // The same use case happens for DeliveryItem type in which the `id`
+        // is really the lineItem id referenced in the delivery
+        // https://docs.commercetools.com/http-api-projects-orders.html#deliveryitem
+        if (typeNameWithReferencedEntity.includes(result.__typename))
           return `${result.__typename}:${result.id}`;
```
qmateub: With this we assure that for the DeliveryItem type we got that specially mapped in the apollo cache 👍

packages/application-shell/src/configure-apollo.js
```diff
@@ -66,6 +66,8 @@ const fragmentMatcher = new IntrospectionFragmentMatcher({
 
 const typeNamesWithoutIdAsIdentifier = ['Project', 'BaseMenu', 'NavbarMenu'];
 
+const typeNameWithReferencedEntity = ['Reference', 'DeliveryItem'];
```
qmateub: I think this is named accordingly to what this points to 😋

qmateub:
That's totally a valid point @emmenko , thanks for reviewing it.

However this is not really "blocking" us but is basically giving us incorrect data to show in order deliveries. Was one of the points that came out in UAT from Ida. I will vote to have this in as an intermediate step and maybe we can revisit this later on, as you said, the default caching from Apollo is what we should aim for.

Just for the record this are the two PRs that are related to this part:

* #310 
* #341 (fixing some problems of the previous one)

If I remember correctly, the main problem was that we were using the same typename for different entities and also I think a problem with how our `dereferenced` components when trying to do `readFragment` at some point (I'm not entirely sure about this last one though)

Anyways, no matter how we approach this we will still need the custom mapping for some entities right? For exmaple, Projects, we need to cache them by `key` and not `id` 🤔

emmenko:
> For example, Projects, we need to cache them by `key` and not `id`

That's fine, as long as the prefix is the `typename` (if defined)

packages/application-shell/src/configure-apollo.js
tdeekens: Thanks for the extraction.

emmenko:
> Something maybe to follow up with by adding the typename by default, building app-kit and using it (copying it over) locally and maybe running the E2E tests.

Yeah, I was thinking to manually release the change to a different dist-tag and use that version in the apps, deploy a branch and see what breaks.

adnasa:
This branch was merged, but no closure about the release.
So what are the next concrete steps for this?

emmenko:
We're releasing this afternoon

adnasa:
Looking forward.

PR #412, opened 2019-03-01
Dialog layout fixes, add VRT tests 💅

emmenko:
This PR improves and fixes some layout issues for the dialog containers, especially in `scale` size.
I'm using now a css grid layout to ensure consistent spacing for different sizes (e.g. the dialog should only be allowed to stretch to a max of 32px from the container/screen borders.

Also, since we now have VRT setup (#375) we can snapshots the components for different variants (check out the percy build to see the components rendered in different sizes).


emmenko:
VRT have been approved.

PR #413, opened 2019-03-02
fix(test-utils): create default memory history with the enhanced location

emmenko:
In our default router we use the enhanced history with the parsed location query, defined in the `@commercetools-frontend/browser-history` package.

However, in our test-utils we were just creating a memory history without the enhancement.

This PR exposes the enhancer function from the `@commercetools-frontend/browser-history` package and uses that to create the default memory history object.


packages/application-shell/src/test-utils/README.md
```diff
@@ -113,22 +113,22 @@ This section describes the methods exported by `@commercetools-frontend/applicat
...
+| `options.history`     | `Object`      | Routing                                                                                                                                                                                            | By default a memory-history is generated which has the provided `options.route` set as its initial history entry. It's possible to pass a custom history as well. In that case, we recommend using the factory function `createEnhancedHistory` from the `@commercetools-frontend/browser-history` package, as it contains the enhanced `location` with the parsed `query` object. |
```
emmenko: This diff is weird. The only thing that changed is

```diff
- By default a memory-history is generated which has the provided `options.route` set as its initial history entry. It's possible to pass a custom history as well.
+ By default a memory-history is generated which has the provided `options.route` set as its initial history entry. It's possible to pass a custom history as well. In that case, we recommend using the factory function `createEnhancedHistory` from the `@commercetools-frontend/browser-history` package, as it contains the enhanced `location` with the parsed `query` object.
```

PR #426, opened 2019-03-05
fix(apps-menu): to trigger remote query based on servedByProxy value

### LINKED ISSUE #407: Cannot start application locally in production mode
**Describe the bug**
With #289 we changed how the navigation menu is loaded. For instance, in production the menu configuration is fetched from a `/api/graphql` endpoint on the same domain of the frontend app, as a we have a proxy router between the LB and the applications.

In development though, we usually start the app directly, without the proxy router. To render the menu config of the specific application, we load the menu config from a local file, skipping the graphql request (as the endpoint does not exist).

However, when testing the production build locally like so:

```bash
NODE_ENV=production mc-http-server --csp=$(pwd)/csp.json --config=$(pwd)/env.json --use-local-assets
```

the application starts in "production" mode, triggering the non-existing graphql request.

**Expected behavior**
It should be possible to start the application locally in production mode without having the graphql request to error.

**Additional context**
One solution would be to check for the flag `servedByProxy` from the `env.json`, which usually indicates if the application runs behind the proxy or not.

A bit unrelated to the problem, but we might also want to look into manually putting the menu data into the cache when we don't fetch from the graphql endpoint, so that we can still make the graphql query which will only read from the cache.

emmenko:
Fixes #407

Previously we were basing the logic to load the apps menu config from the local file or remotely based on the `NODE_ENV` variable. However, when testing the production bundle locally in production mode, the app would break because the graphql endpoint does not exist.

Now we base the logic on the `servedByProxy` flag.

Additionally, I also removed the prop `INTERNAL__isApplicationFallback` and assume that the `useFullRedirectsForLinks` comes from the `env.json`. We only use it internally in our fallback app, so I don't consider it a breaking change.

### DIFF
--- a/packages/application-shell/src/components/application-shell/application-shell.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.js
@@ -335,6 +335,16 @@ RestrictedApplication.propTypes = {
   DEV_ONLY__loadNavbarMenuConfig: PropTypes.func,
 };
 
+const shallowlyCoerceBooleanValues = obj =>
+  Object.keys(obj).reduce((updatedObj, key) => {
+    const value = obj[key];
+    const isBoolean = value === 'true' || value === 'false';
+    return {
+      ...updatedObj,
+      [key]: isBoolean ? value === 'true' : value,
+    };
+  }, {});
+
 export default class ApplicationShell extends React.Component {
   static displayName = 'ApplicationShell';
   static propTypes = {
@@ -363,9 +373,12 @@ export default class ApplicationShell extends React.Component {
     });
   }
   render() {
+    const environmentValues = shallowlyCoerceBooleanValues(
+      this.props.environment
+    );
     return (
       <ApplicationShellProvider
-        environment={this.props.environment}
+        environment={environmentValues}
         trackingEventWhitelist={this.props.trackingEventWhitelist}
         applicationMessages={this.props.applicationMessages}
       >
@@ -373,7 +386,7 @@ export default class ApplicationShell extends React.Component {
           if (isAuthenticated)
             return (
               <RestrictedApplication
-                environment={this.props.environment}
+                environment={environmentValues}
                 defaultFeatureFlags={this.props.defaultFeatureFlags}
                 render={this.props.render}
                 applicationMessages={this.props.applicationMessages}
--- a/packages/application-shell/src/components/application-shell/application-shell.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.js
@@ -153,9 +153,7 @@ export const RestrictedApplication = props => (
                                     <QuickAccess
                                       history={routeProps.history}
                                       user={user}
-                                      useFullRedirectsForLinks={
-                                        props.INTERNAL__isApplicationFallback
-                                      }
+                                      environment={props.environment}
                                     />
                                   );
                                 return (
@@ -177,9 +175,7 @@ export const RestrictedApplication = props => (
                                           }
                                           history={routeProps.history}
                                           user={user}
-                                          useFullRedirectsForLinks={
-                                            props.INTERNAL__isApplicationFallback
-                                          }
+                                          environment={props.environment}
                                         />
                                       </ApplicationContextProvider>
                                     )}
@@ -195,6 +191,7 @@ export const RestrictedApplication = props => (
                         <AppBar
                           user={user}
                           projectKeyFromUrl={projectKeyFromUrl}
+                          environment={props.environment}
                           DEV_ONLY__loadAppbarMenuConfig={
                             props.DEV_ONLY__loadAppbarMenuConfig
                           }
@@ -239,9 +236,7 @@ export const RestrictedApplication = props => (
                                       menuVisibilities={
                                         project.menuVisibilities
                                       }
-                                      useFullRedirectsForLinks={
-                                        props.INTERNAL__isApplicationFallback
-                                      }
+                                      environment={props.environment}
                                       DEV_ONLY__loadNavbarMenuConfig={
                                         props.DEV_ONLY__loadNavbarMenuConfig
                                       }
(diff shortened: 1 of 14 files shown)


packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -691,18 +697,22 @@ export default flowRight(
...
+        menuKey: 'navBar',
+      },
+    }),
+  }),
   graphql(FetchProjectExtensionsNavbar, {
     name: 'projectExtensionsQuery',
     skip: ownProps =>
-      !ownProps.areProjectExtensionsEnabled ||
-      process.env.NODE_ENV === 'development',
+      [false, 'false'].includes(ownProps.environment.servedByProxy) ||
```
adnasa: 🙈

emmenko:
> let's test this on a canary build

Even easier 😉

```bash
yarn build
cd playground
yarn build
yarn start:prod:local
```

packages/application-shell/src/components/quick-access/quick-access.js
```diff
@@ -79,7 +79,10 @@ class QuickAccess extends React.Component {
     onChangeProjectDataLocale: PropTypes.func,
     pimSearchProductIds: PropTypes.func.isRequired,
     getPimSearchStatus: PropTypes.func.isRequired,
-    useFullRedirectsForLinks: PropTypes.bool,
+    environment: PropTypes.shape({
+      useFullRedirectsForLinks: PropTypes.oneOf([true, false, 'true', 'false'])
```
tdeekens: I guess, like we all, I find this a bit gnarly. Any component relying on this prop will have a quite bloated prop-type eventually. I think it's fine for now but somehow the application context provider (I presume) could also normalise these boolean flags coming in as strings.

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -691,18 +697,22 @@ export default flowRight(
...
+  withApplicationsMenu({
     queryName: 'applicationsMenuQuery',
-    __DEV_CONFIG__: {
-      menuLoader: ownProps.DEV_ONLY__loadNavbarMenuConfig,
-      menuKey: 'navBar',
-    },
-  })),
+    skipRemoteQuery: ownProps =>
+      [false, 'false'].includes(ownProps.environment.servedByProxy),
+    options: ownProps => ({
```
tdeekens: I like this being an getter now.

packages/application-shell/src/components/with-applications-menu/with-applications-menu.js
```diff
@@ -68,20 +68,17 @@ function withApplicationsMenu(getOptions) {
...
+                  {...this.props}
+                  {...{ [mergedConfig.queryName]: { ...data, error } }}
                 />
               )}
             </Query>
           )}
         />
       );
-    };
-    WrappedComponent.displayName = `withApplicationsMenu(${getDisplayName(
```
tdeekens: Ah, this just moved up.

packages/application-shell/src/components/quick-access/quick-access.js
emmenko: Yeah, but in the app shell we usually don't read from the context. I can coerce the values in the top component though...let me try.

packages/application-shell/src/components/with-applications-menu/with-applications-menu.js
emmenko: It's now one single component within the HOC, with some if-else.

packages/application-shell/src/components/quick-access/quick-access.js
tdeekens: Yeah. That's what I meant to suggest. Thanks for trying!

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -335,6 +335,16 @@ RestrictedApplication.propTypes = {
   DEV_ONLY__loadNavbarMenuConfig: PropTypes.func,
 };
 
+const shallowlyCoerceBooleanValues = obj =>
```
tdeekens: Nitpick: memoize based on `obj` otherwise the provider will re-render quite a bit.

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -335,6 +335,16 @@ RestrictedApplication.propTypes = {
...
+  Object.keys(obj).reduce((updatedObj, key) => {
+    const value = obj[key];
+    const isBoolean = value === 'true' || value === 'false';
+    return {
+      ...updatedObj,
+      [key]: isBoolean ? value === 'true' : value,
+    };
+  }, {});
```
tdeekens: I know this may be a bit over the top but I wanted to suggest it and at least give the full function's code for an easy copy. I just prefer very clear names for things 😄 

```js
const isBooleanAsString = environmentValue =>
  environmentValue === "true" || environmentValue === "false";

const shallowlyCoerceBooleanValues = defaultMemoize(
  uncoercedEnvironmentValues =>
    Object.keys(uncoercedEnvironmentValues).reduce(
      (uncoercedEnvironmentValues, keyOfUncoercedEnvironmentValue) => {
        const uncoercedEnvironmentValue =
          uncoercedEnvironmentValues[keyOfUncoercedEnvironmentValue];

        return {
          ...coercedEnvironmentValues,
          [key]: isBooleanAsString(uncoercedEnvironmentValue)
            ? uncoercedEnvironmentValue === "true"
            : uncoercedEnvironmentValue
        };
      },
      {}
    )
);
```

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -363,17 +373,20 @@ export default class ApplicationShell extends React.Component {
     });
   }
   render() {
+    const environmentValues = shallowlyCoerceBooleanValues(
```
tdeekens: `const coercedEnvironmentValues = ...`

emmenko:
@tdeekens thanks for the feedback!

tdeekens:
Thanks for incorporating it!

PR #425, opened 2019-03-05
fix: pin react-router to 4.4.0-beta.6

emmenko:
Refs:
- https://github.com/ReactTraining/react-router/issues/6611
- #415 

Fixes #424

PR #427, opened 2019-03-05
feat: add cypress tests for playground and template apps

emmenko:
This should help us prevent possible regressions (e.g. #425).

PS: I've created a new org/project/test-user and added the env variables on travis.


emmenko:
Hmm it seems that our strategy with "skipping" the Percy tests does not really work. I guess we just skip running it and do not require the status check.

cypress/support/commands.js
```diff
@@ -0,0 +1,52 @@
...
+//
+// -- This is will overwrite an existing command --
+// Cypress.Commands.overwrite("visit", (originalFn, url, options) => { ... })
+
+Cypress.Commands.add('logout', () => {
+  cy.visit('/logout');
+});
+
+Cypress.Commands.add('login', () => {
+  cy.visit('/logout');
```
tdeekens: Given we define the cmd above this could be cy.logout() I suppose. I guess, if it works, the same applies maybe in other repos.

package.json
```diff
@@ -100,8 +111,8 @@
     "react-router-dom": "4.4.0-beta.6"
   },
   "engines": {
-    "node": ">=8.9.4",
+    "node": ">=8.9",
     "npm": ">=5",
-    "yarn": "~1.3.0||>=1.5.0"
+    "yarn": ">=1.13"
```
tdeekens: It is interesting how one could argue this being a breaking change. 😀

cypress/plugins/index.js
```diff
@@ -0,0 +1,41 @@
...
+    require('dotenv').config({ path: envPath });
+  }
+
+  return Object.assign({}, cypressConfig, {
+    env: Object.assign({}, cypressConfig.env, {
+      LOGIN_USER: process.env.CYPRESS_LOGIN_USER,
+      LOGIN_PASSWORD: process.env.CYPRESS_LOGIN_PASSWORD,
+      PROJECT_KEY: process.env.CYPRESS_PROJECT_KEY,
+    }),
+  });
```
tdeekens: Simple and good. Wondering if one day we can actually move shared commands etc from the mc-fe here also allowing custom app authors to easily write e2e tests. Even though it ia likely an „unimportant“ feature wish right now. Anyway, it would also force us to have the same/in-sync logic across repos. So that e.g these env vars are read the same on travis and circle no matter what repo. 

These things btw is what I hoped would be nice tasks for the full time focus FLD role.

package.json
emmenko: Hmm why would you say it's a breaking change? You mean for contributors?

cypress/plugins/index.js
emmenko: > allowing custom app authors to easily write e2e tests

Yeah, I was also thinking of having a cypress setup in e.g. the example apps. Or have a plugin package that injects commands (`@commercetools-frontend/cypress-mc-app`).

emmenko:
One small side-effect is that the build now takes >15min to run 😅

package.json
tdeekens: Yeah. Anybody installing, not having that version, will at least get a warning. I am not saying it should but it is an interesting thought.

cypress/plugins/index.js
tdeekens: Yeah. My head was thinking more of @commercetools-frontend/cypress-mc-app. Maybe create an issue for it so we do not forget and can give it to an FLD who has time to pick it up? As a side effect of that we „force“ to share and gain knowledge of these things.

package.json
emmenko: Yeah, we can also look it the other way around: local and CI should use the same yarn version to avoid possible weird issues 😜

cypress/plugins/index.js
emmenko: Done: #429

tdeekens:
> One small side-effect is that the build now takes >15min to run 😅 (because we build 2 apps, run tests on each one of them, etc).

I guess we should analyse where we spend time. TravisCI generally seems a bit slower than it has to at times. Maybe running both in parallel really is an easy gain to just cut time in half.

emmenko:
I looked into the travis build stages (https://docs.travis-ci.com/user/build-stages) but they don't share artifacts between stages, you have to upload the files on your own.

I think in this case circleci will be better for defining parallel workflows (in case we want to proceed with that).

tdeekens:
Yeah I think (personal opinion) there is benefit in settling on one CI solution across repos for various reasons (familiarity etc) and pay the price for speed.

PR #428, opened 2019-03-06
chore: remove dummy percy test

emmenko:
It seems that using a dummy snapshot when percy should not run, it's causing the actual snapshots to be removed (and thus need to be approved again).

https://percy.io/commercetools-GmbH/merchant-center-application-kit/builds/1519927

So we just skip running Percy if the files did not change, and the github check for VRT is not required anymore.


tdeekens:
Lets hope folks do not merge over failing Percy. But I also do not see another way.

PR #444, opened 2019-03-12
Remove login/logout components and redirect to urls

tdeekens:
#### Summary

This pull requests removes the login/logout components from the app-kit while redirecting to the urls.


packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -438,17 +429,19 @@ export default class ApplicationShell extends React.Component {
...
-                      }}
-                    />
+
+                  /**
+                   * NOTE:
+                   *   Given the application runs in production we redirect to that host's login application.
+                   *   Given the application is in development mode (locally) we redirect to the application on staging.
+                   */
+                  const frontendHost =
+                    process.env.NODE_ENV === 'production'
```
emmenko: We should use `servedByProxy` to determine if we're running locally or on a production environment

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -438,17 +429,19 @@ export default class ApplicationShell extends React.Component {
...
+                      ? ''
```
emmenko: Should we use `window.location.origin`? I find it clearer on the intent

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -438,17 +429,19 @@ export default class ApplicationShell extends React.Component {
...
+                      : this.props.environment.frontendHost;
```
emmenko: I don't think we can use this. For local development this is `localhost:3001`. We should probably hardcode the staging domain here, or we add a new value e.g. `loginUrl` or `authUrl` in `env.json`

packages/application-shell/src/index.js
```diff
@@ -18,6 +18,10 @@ export {
 export { selectUserId, selectProjectKeyFromUrl } from './utils';
 export { default as AsyncChunkLoader } from './components/async-chunk-loader';
 export { GtmContext } from './components/gtm-booter';
+export {
+  default as GtmUserLogoutTracker,
```
emmenko: Should we move this to the auth app? It's not used anymore here.

packages/application-shell/src/components/application-shell-provider/application-shell-provider.js
```diff
@@ -61,47 +60,32 @@ export default class ApplicationShellProvider extends React.Component {
...
                   trackingEventWhitelist={this.props.trackingEventWhitelist}
                 >
-                  <Switch>
-                    {/**
-                     * No matter if the user is authenticated or not, when we go
-                     * to this route we should always log the user out.
-                     * TODO: do a hard page reload once we move the login routes
-                     * to a different app.
-                     */}
-                    <Route path="/logout" component={Logout} />
```
emmenko: This needs to stay and to do a hard redirect. Unless we change the `logout` link to be a hard redirect

packages/application-shell/src/index.js
tdeekens: I thought about it but then we spread logic: set-up in app-kit and teardown in app. That's why I liked to just export it.

packages/application-shell/src/components/application-shell/application-shell.js
tdeekens: Yeah makes sense. My only goal was to have it working on all cloud deployments.

packages/application-shell/src/index.js
emmenko: Yeah I guess...

packages/application-shell/src/components/application-shell/application-shell.js
tdeekens: Can you elaborate why `frontendHost` is not working? Adding `loginUrl` sounds reasonable but I don't follow how that is not just `${frontendHost}/login`.

tdeekens:
@emmenko pushed a commit containing some fixes.

emmenko:
Ok thanks. Have you seen this comment btw? (https://github.com/commercetools/merchant-center-application-kit/pull/444#discussion_r264612202)

I don't think we can use `frontendHost` value.

packages/application-shell/src/components/application-shell-provider/application-shell-provider.js
tdeekens: I added a hard-redirect in the same vein as in the other component now.

packages/application-shell/src/components/application-shell/application-shell.js
emmenko: Well, in development `frontendHost` points to `localhost:3001`. So when you go to `/login`, we end up in an infinite loop because the page keeps reloading as there is no "login service" running.

Instead, it should point to `https://mc.commercetools.com/login`, right? (or our staging domain, which I won't write down).

tdeekens: Understood. Thanks. Then I will add a new env variable called `authUrl` to have it be "augmented" to `/login` and `/logout`.

playground/env.prod-us.json
```diff
@@ -1,5 +1,6 @@
 {
   "frontendHost": "mc.commercetools.co",
+  "mcAuthUrl": "https://mc.commercetools.co",
```
tdeekens: Thanks. I wanted to do this after I saw that we have an `ctpAuthUrl` as an env variable.

packages/application-shell/src/components/application-shell-provider/application-shell-provider.js
```diff
@@ -18,6 +18,26 @@ import ErrorApologizer from '../error-apologizer';
...
+    const authUrl = this.props.environment.servedByProxy
+      ? window.location.origin
+      : this.props.environment.authUrl;
+    this.redirectTo(`${authUrl}/logout`);
+  }
+  render() {
+    return null;
+  }
+}
+
```
tdeekens: Many thanks!

packages/application-shell/src/components/login/login.js
```diff
@@ -1,257 +0,0 @@
...
-        storage.put(STORAGE_KEYS.IS_AUTHENTICATED, true);
-        // Ensure to remove the IdP Url value from local storage, as in this
-        // case login is not done through SSO.
-        storage.remove(STORAGE_KEYS.LOGIN_STRATEGY);
-        // Redirect to a `redirectTo` url, if present, otherwise to defaul route
-        const nextTargetUrl = this.props.location.query.redirectTo;
-        // Inspect the target url to see if the host is one of commercetools
-        try {
-          const nextTargetUrlObject = new URL(nextTargetUrl);
-          if (nextTargetUrlObject.hostname === window.location.hostname) {
```
emmenko: This is currently preventing the redirect from staging -> localhost (currently a problem only for development).

packages/application-shell/src/components/authenticated/authenticated.js
```diff
@@ -1,17 +1,54 @@
...
 
 const Authenticated = props => {
-  const isAuthenticated = getIsAuthenticated();
-  return props.children({ isAuthenticated });
+  // We attempt to see if the user was already authenticated by looking
+  // at the "cached" flag in local storage.
+  const cachedAuthenticationState = hasCachedAuthenticationState();
+  if (cachedAuthenticationState) {
+    return props.render({ isAuthenticated: true });
+  }
```
tdeekens: Still any subsequent request will fail in the mc-be in turn logging the user out again. So there is no harm in this caching here, correct?

packages/application-shell/src/components/authenticated/authenticated.js
```diff
@@ -1,17 +1,54 @@
...
+  // is valid. We return null while the query is loading.
+  return (
+    <Query
+      query={AmILoggedInQuery}
+      variables={{ target: GRAPHQL_TARGETS.MERCHANT_CENTER_BACKEND }}
+      onCompleted={() => {
+        // Even though the login page might set this flag, we just make sure that
+        // we do it here as well. This will help in the future when we eventually
+        // move the auth service to itw own domain, in which case the local storage
+        // is not shared anymore.
```
tdeekens: You're saying that in that case e.g. `login.commercetools.com` might save things to local storage as well as this but as code has/may run on different domains it remains to work?

packages/application-shell/src/components/authenticated/authenticated.js
```diff
@@ -1,17 +1,54 @@
...
+        storage.put(STORAGE_KEYS.IS_AUTHENTICATED, true);
+      }}
+    >
+      {({ loading, data, error }) => {
+        if (error) {
+          console.error(error);
```
tdeekens: Dangling log?

packages/application-shell/src/components/authenticated/authenticated.js
```diff
@@ -1,17 +1,54 @@
...
+          return props.render({ isAuthenticated: false });
+        }
+        if (!loading && data && data.amILoggedIn) {
+          return props.render({ isAuthenticated: true });
+        }
+        return null;
```
tdeekens: Q: wny not return `isAuthenticated: false` in this case?

packages/application-shell/src/components/application-shell-provider/application-shell-provider.spec.js
```diff
@@ -32,6 +31,10 @@ const createTestProps = props => ({
 });
 
 describe('rendering', () => {
+  beforeEach(() => {
+    // eslint-disable-next-line global-require
+    require('../authenticated').setIsAuthenticated('true');
```
tdeekens: Q: Can't we just import this as `import { setIsAuthenticated } from '../authenticated'`? I presume, from reading the code, you ran into issues with mocking in L11. I find this `require`-cjs work around a bit confusing. Would a 

```js
jest.mock('../authenticated', () => ({
  ...require.requireActual('../authenticated'),
  // eslint-disable-next-line react/prop-types
  Authenticated: ({ children }) => <div>{children}</div>,
}));
```

Might be a bit trouble some with the default export of `Authenticated` and I am not 100% sure what the issue here is but just wanted to share the thouhgt.

packages/application-shell/src/components/login/login.js
tdeekens: We would change that in the mc-app now though, right? Paring it somehow with `servedByProxy` again I suppose...

emmenko: I guess yes

packages/application-shell/src/components/authenticated/authenticated.js
emmenko: Yes correct

packages/application-shell/src/components/authenticated/authenticated.js
emmenko: Yes that's the idea, to probably reduce breaking changes.
In case we have our own login/auth service, I would still like to consider a different auth mechanism though (as we discussed already).

packages/application-shell/src/components/authenticated/authenticated.js
emmenko: Not really, I wasn't sure what to do with that to be honest. Most likely is going to be a "unauthorized" error, which we can safely ignore.

packages/application-shell/src/components/authenticated/authenticated.js
emmenko: If we do that, we end up rendering e.g. the login page. However, we need to wait for the request to be finished, and only then render the function (that's also why I renamed it from `children` to `render`, as it's only executed in certain cases)

packages/application-shell/src/components/application-shell-provider/application-shell-provider.spec.js
emmenko: I've spent a bit of time on this actually.

The main problem is that the `ApplicationShellProvider` renders the normal apollo provider. So if we test it and wrap it with an apollo mock provider to mock the graphql request, it won't work because the normal apollo provider takes precedence over the mocked one. This is only a "problem" with this component, as it's the top level one.

So the only solution is to mock the `Authenticated` component using a manual mock, and we need to be able to control the `isAuthenticated` value from outside of the mock.

I haven't tried with `import { setIsAuthenticated } from '../authenticated'`, will do that tomorrow. Thanks!

packages/application-shell/src/components/authenticated/authenticated.js
tdeekens: Ah, it is too late. Missed that it is a render-faac not a children-faac. Makes sense. Thanks.

packages/application-shell/src/components/authenticated/authenticated.js
tdeekens: Should we maybe report to sentry instead?

packages/application-shell/src/components/authenticated/authenticated.js
tdeekens: Yeah.

packages/application-shell/src/components/login/login.js
tdeekens: We can tackle that on the PR taking a pre-canary build of this I guess.

tdeekens:
A bit strange that the e2e tests for the playground keep failing. 😩😭

emmenko:
It’s because of the login redirect I believe. Don’t worry too much about the e2e tests for now.

packages/application-shell/src/components/authenticated/authenticated.js
emmenko: We will get spammed if we log on every unauthorized. Maybe we can send errors that are not unauthorized.

tdeekens:
BTW: I published an earlier version to pre-canary. Will publish this there later too so we can try it.

CleanShot 2019-03-13 at 11 22 01@2x

emmenko:
New release:

```
Successfully published:
 - @commercetools-frontend/application-shell@11.0.3-pre-canary.11+f3b22193
```

emmenko:
Update: for local development we need to use a login page from `localhost`, as the API response needs to set the cookie. Using a login page to another domain doesn't work in that regards.
Therefore, we don't need the `mcAuthUrl` anymore.

To amend that (for the short period), we serve a "simple" html login page from the webpack dev server, that logs the user in and redirects to the actual app running on `localhost`.

@tdeekens please try this locally with the playground app to see if everything works fine, thanks!

packages/mc-scripts/config/views/login.pug
```diff
@@ -0,0 +1,40 @@
+html
```
tdeekens: Nice. Sorry to ask but do we really need pug for this? 😄

emmenko: We don't, but on the other hand...why not? 😄

We can use plain html, I don't mind...as long as it works

tdeekens:
> @tdeekens please try this locally with the playground app to see if everything works fine, thanks!

Sure. I planned to have a little intro meeting next week with all FE engineers going through JWTs and how login/sign-up will work. There I planned to also elaborate on why we need this page for now.

tdeekens:
Confirmed sir. The temporary development login page works like a charm. Could use some more Netscape style styling but is lovely.

emmenko:
New release:

```
Successfully published:
 - @commercetools-frontend/application-shell@11.0.3-pre-canary.13+907f42f7
 - @commercetools-frontend/mc-scripts@11.0.3-pre-canary.13+907f42f7
```

tdeekens:
👑 Majestic 👑

packages/mc-scripts/config/views/login.js
```diff
@@ -20,19 +19,25 @@ window.addEventListener('load', function loaded() {
...
-      // Call a function when the state changes.
-      if (this.readyState === XMLHttpRequest.DONE) {
-        if (this.status === 200) {
-          window.localStorage.setItem('isAuthenticated', true);
-          var searchParams = new URLSearchParams(window.location.search);
-          var redirectTo = searchParams.get('redirectTo') || '/';
-          window.location.replace(redirectTo);
-        } else {
+    window
+      .fetch(data.get('url'), {
```
tdeekens: Nice. Wanted to ask why we not using fetch after thinking about this.

packages/mc-scripts/config/views/logout.pug
```diff
@@ -1,4 +1,10 @@
 html
   head
+    title Logout (development only)
     script
       include logout.js
+  body
+    div
+      h3 This is the logout page for local development.
+      p Be aware that you might still have an active session as the cookie is assigned to a production domain (e.g. commercetools.com) which we can't unset from localhost. This is only a problem on local development and we will fix this in the future.
```
tdeekens: Very cool! Thanks!

packages/application-shell/src/test-utils/test-utils.js
```diff
@@ -285,6 +292,15 @@ const experimentalRender = (ui, renderOptions) => {
...
 };
 
+const defaults = {
+  mocks: {
+    user: defaultUser,
+    project: defaultProject,
+    environment: defaultEnvironment,
+    permissions: defaultPermissions,
+  },
+};
```
tdeekens: Just an idea. It makes the tests a bit simpler. We could also expose the default rtl render there. /cc @emmenko

emmenko: > Just an idea. It makes the tests a bit simpler.

What do you mean? Those values are already available as

```js
const { user, project, environment } = renderApp(<Foo />);
```

> We could also expose the default rtl render there

What do you mean? It's already exported with `export * from 'react-testing-library';`

playground/src/components/state-machines-list/state-machines-list.spec.js
```diff
@@ -25,16 +25,22 @@ const createStateMachinesSdkMock = () => ({
   },
 });
 
+const renderApp = (options = {}) => {
```
emmenko: Careful because `renderApp` is exported from `test-utils`

tdeekens: Not on this branch yet. After I rebase, yes.

packages/application-shell/src/test-utils/test-utils.js
tdeekens: I know. It would just be easier to to test the test-utils. Anyway. Will remove.

tdeekens:
NOTE: rebased to have CircleCI take over testing here.

tdeekens:
Huuuuray. The branch is green. Given we squash merge anyway I won't clean up the commit history here.

I think generally it's ready for review though.

tdeekens:
This branch can only be merged after UAT has passed on the respective internal branch deployment.

packages/application-shell/src/components/project-container/project-container.js
```diff
@@ -112,9 +121,19 @@ export class ProjectContainer extends React.Component {
...
     }
-    // TODO: do something if there is an `error`?
+
+    /**
+     * Given the user does not have project (and as a result is not part of an organization)
+     * the account application gets control over render. If any other application
+     * is requested to render a full page redirect (to have the proxy serve the request) occurs.
+     */
     if (this.props.user && this.props.user.projects.total === 0)
-      return <Redirect to="/logout?reason=no-projects" />;
```
emmenko: Q: what should we do here as long as signup is disabled?
Should we wrap this logic into feature toggle and clean up afterwards?

packages/i18n/data/core.json
```diff
@@ -36,30 +36,6 @@
...
-  "LoginForm.password": "Password",
-  "LoginForm.signin": "Sign in",
-  "LoginForm.title": "Sign in to your account",
-  "LoginForm.unauthorized": "Your session has expired. Please sign in again to continue using the Merchant Center.",
-  "LoginForm.validating": "Validating...",
-  "LoginLocked.helpDeskLink": "Help Desk",
-  "LoginLocked.paragraph1": "For security reasons, your account has been locked because of too many failed log-in attempts. You can {resetPasswordLink} or contact the commercetools {helpDeskLink}",
-  "LoginLocked.resetPasswordLink": "reset your password",
-  "LoginLocked.title": "Account Locked",
-  "LoginPage.privacyPolicy": "Privacy Policy",
```
emmenko: Let’s also remove the translations for this

playground/src/components/entry-point/entry-point.js
```diff
@@ -28,6 +28,11 @@ export const ApplicationStateMachines = () => (
       />
     )}
     <Route path="/:projectKey/state-machines" component={AsyncStateMachines} />
+    {/* Given a user does not have any projects yet, then the app-shell redirects here. The playground renders a message instead of letting the <RouteCatchAll> handle it. */}
+    <Route
+      path="/account/projects/new"
```
emmenko: Hmm how is this useful? There is no account all running locally

packages/application-shell/src/components/project-container/project-container.js
tdeekens: Hm. Is there a problem with not wrapping it? In the AC you get your project and org straight away. If you get an account otherwise it may be nice to get to account creation (another toggle). Do you see a specific problem we can run into?

playground/src/components/entry-point/entry-point.js
tdeekens: If I remember correctly otherwise you and up in a redirect loop as the route catch all kicks in and so forth. Plus it felt kind of okay to have it for debugging as there is no accounts app.

packages/application-shell/src/components/project-container/project-container.js
emmenko: Just thinking about the case when the user does not have any projects (it’s not necessarily after signup) and we redirect to create a project but the accounts app is non available (feature toggled).

playground/src/components/entry-point/entry-point.js
emmenko: Hmm I see. Does it mean that every app needs to do that?

tdeekens: Fair question. Something we should check, true. Maybe we should redirect only if it's served by the proxy of something smater.

packages/application-shell/src/components/project-container/project-container.js
```diff
@@ -28,6 +28,15 @@ const shouldShowNotificationForTrialExpired = daysLeft =>
   daysLeft <= minDaysToDisplayNotification &&
   daysLeft >= maxDaysToDisplayNotification;
 
+export class RedirectToProjectCreate extends React.Component {
```
emmenko: I think we can fix that problem here:
- if it’s not `servedByProxy` we don’t redirect and show a simple message

If we’re in the account app already, there won’t be any redirect and the route will match and let the app handle the rest.

tdeekens: Yup I agree. Thanks. Have a good weekend 🤪

playground/src/components/entry-point/entry-point.js
tdeekens: See below.

emmenko:
@tdeekens should I take care of the rebase of you do that?

tdeekens:
I already did. Thanks for asking.

emmenko:
Ok, because there are a bunch of conflicts here

packages/application-shell/src/components/project-container/project-container.js
tdeekens: e46235f

packages/application-shell/src/components/project-container/project-container.js
tdeekens: 843ae4c

packages/i18n/data/core.json
tdeekens: 2168293

emmenko:
> why did lint-staged not catch this

Yeah, I think because you don't create any commits during rebase. Right?

emmenko:
We also need to do a new pre-canary release now, if you want the tests to pass

tdeekens:
> We also need to do a new pre-canary release now, if you want the tests to pass

Yeap I was just making ☕️ and about to do that now :)

application-templates/starter/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "merchant-center-application-template-starter",
-  "version": "12.2.0",
+  "version": "12.0.1-pre-canary.32",
```
emmenko: Q: should we revert those version before merging? I guess it shouldn't matter if we then release a major, all version will be bumped anyway.

packages/application-shell/src/components/project-container/project-container.js
```diff
@@ -112,9 +124,32 @@ export class ProjectContainer extends React.Component {
...
+    if (
+      this.props.user.projects.total === 0 &&
+      !this.props.isAccountCreationEnabled
+    )
       return <Redirect to="/logout?reason=no-projects" />;
+    if (this.props.user && this.props.user.projects.total === 0)
+      return (
+        <Switch>
+          <Route path="/account" render={this.props.render} />
+          {this.props.environment.servedByProxy && (
```
emmenko: Hmm that's not what I had in mind. We should still render the `RedirectToProjectCreate` but instead of rendering `null`, we should render a message explaining that there are no projects and that you would need the account app blabla. And of course the redirect should be prevented.

We can also have another component and render one or the other. But the point is to keep the route, otherwise things will probably break.

tdeekens: K

application-templates/starter/package.json
tdeekens: What do you mean by revert? Unpublishing won't work after 24 hours. We could remove the dist-tag.

tdeekens:
@jonnybel can you also have a look when you have time please? We can't merge it yet but it would be cool to have it ready.

jonnybel:
Yes, I've looked at parts of it while it's being worked on, gotta take a final look.

emmenko:
@tdeekens is there an ETA on when we are supposed to merge this?

tdeekens:
Alright. Rebased out all pre-canary releases. So the pkg.json's are now up-to-date with master. Furthermore in 8e95ddd I added an `aria-label` for sign up cause I realised otherwise running the mc locally will not work the the mc-fe E2E tests as it requires aria-label.

PR #458, opened 2019-03-18
fix(mc-scripts): use terser as minifier

tdeekens:
#### Summary

Uglify-js is not recommended to be used anymore. Also CRA uses terser. Furthermore, uglify-js has issues with hooks. Just google for "Cannot set property 'lastEffect' of null".

### DIFF
--- a/packages/mc-scripts/config/create-webpack-config-for-production.js
+++ b/packages/mc-scripts/config/create-webpack-config-for-production.js
@@ -12,7 +12,7 @@ const CleanWebpackPlugin = require('clean-webpack-plugin');
 // as "aliasing v1.0.0 as webpack.optimize.UglifyJsPlugin is scheduled for
 // webpack v4.0.0" (https://webpack.js.org/plugins/uglifyjs-webpack-plugin/)
 // we need to explicitly use the library to be using the newest version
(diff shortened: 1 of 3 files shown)


tdeekens:
Will quickly give this a spin cherry-picking it into a pre-canary branch to publish and branch deploy with it.

packages/mc-scripts/config/create-webpack-config-for-production.js
```diff
@@ -120,7 +78,10 @@ module.exports = ({
     // https://medium.com/webpack/webpack-4-mode-and-optimization-5423a6bc597a
     optimization: {
       minimizer: [
-        new UglifyJsPlugin(uglifyConfig),
+        new TerserPlugin({
```
amine-benselim: woah! this is too simple 👍  😍

tdeekens: It has some options too but these are the most useful I think. The compression options we can opt-into once we need them I guess. The branch builds fine. Will dig a bit into size comparison before merging.

tdeekens: Sizes are very similar. Nothing to hold back the migration I think.

PR #53, opened 2018-11-05
chore: fix this undefined due to malformed react components

montezume:
#### Summary

In `project-not-found` and `project-expired`, we have a few warnings due to the babel transformation / rollup.

```js
https://rollupjs.org/guide/en#error-this-is-undefined
src/from-core/page-not-found/page-not-found.js
13:     lineNumber: 9
14:   },
15:   __self: this
              ^
16: }, React.createElement(FormattedMessage, Object.assign({}, messages.helpDesk, {
17:   __source: {
...and 1 other occurrence
src/components/project-expired/project-expired.js
13:     lineNumber: 9
14:   },
15:   __self: this
              ^
16: }, salesEmail);
```

The issue was that we were creating React components for the links, but we weren't creating them as class or functional components. So this was undefined.

The compiled code looked like

```js
var salesEmail = 'sales@commercetools.com';
var mailto = React.createElement("a", {
  href: "mailto:".concat(salesEmail),
  __source: {
    fileName: _jsxFileName,
    lineNumber: 9
  },
  __self: undefined
}, salesEmail);
```

and now it looks like

```js

var salesEmail = 'sales@commercetools.com';

var EmailLink = function EmailLink(email) {
  return React.createElement("a", {
    href: "mailto:".concat(email),
    __source: {
      fileName: _jsxFileName,
      lineNumber: 11
    },
    __self: this
  }, salesEmail);
};

EmailLink.propTypes = {
  email: PropTypes.string.isRequired
};
EmailLink.displayName = 'EmailLink';
```

PR #461, opened 2019-03-21
chore: update to babel@8.4 and core-js@2

tdeekens:
#### Summary

Wanted to get us to the latest babel version. Includes some other minor bumps.


package.json
```diff
@@ -89,15 +89,15 @@
     "puppeteer": "1.13.0",
     "rcfile": "1.0.3",
     "read-pkg-up": "4.0.0",
-    "rollup": "1.6.0",
+    "rollup": "1.7.0",
     "rollup-plugin-babel": "4.3.2",
     "rollup-plugin-commonjs": "9.2.1",
-    "rollup-plugin-json": "3.1.0",
+    "rollup-plugin-json": "4.0.0",
```
islam3zzat: https://github.com/rollup/rollup-plugin-json/releases/tag/v4.0.0
Is there an easy way to see if this doesn't break anything?

tdeekens: Thanks for asking.

> Pass all JSON data through dataToEsm to consistently support "compact" formatting, support empty keys, abandon Node 4 support, add prettier, update dependencies (#53)

I presumed it's about dropping node@4 support which shouldn't be a big problem.

packages/mc-scripts/config/create-webpack-config-for-development.js
```diff
@@ -55,8 +55,8 @@ module.exports = ({
   entry: {
     app: [
       // Ship a few polyfills by default
-      require.resolve('./polyfills'),
-      require.resolve('babel-polyfill'),
+      require.resolve('core-js/stable'),
+      require.resolve('regenerator-runtime/runtime'),
```
tdeekens: More on this [here](https://github.com/zloirock/core-js/blob/master/docs/2019-03-19-core-js-3-babel-and-a-look-into-the-future.md#babelpolyfill) and [here](https://babeljs.io/blog/2019/03/19/7.4.0).

packages/babel-preset-mc-app/index.js
```diff
@@ -43,8 +43,9 @@ module.exports = function getBabePresetConfigForMcApp() {
           // the targeted browsers. This is safer than `usage` which performs
           // static code analysis to determine what's required.
           // This is probably a fine default to help trim down bundles when
-          // end-users inevitably import '@babel/polyfill'.
+          // end-users inevitably import 'core-js/stable'.
           useBuiltIns: 'entry',
+          corejs: 3,
```
tdeekens: Otherwise you rightfully get spammed with this warning

> WARNING: We noticed you're using the `useBuiltIns` option without declaring a core-js version. Currently, we assume version 2.x when no version is passed. Since this default version will likely change in future versions of Babel, we recommend explicitly setting the core-js version you are using via the `corejs` option.

> You should also be sure that the version you pass to the `corejs` option matches the version specified in your `package.json`'s `dependencies` section. If it doesn't, you need to run one of the following commands:

>  npm install --save core-js@2    npm install --save core-js@3
>  yarn add core-js@2              yarn add core-js@3

packages/mc-scripts/config/create-webpack-config-for-development.js
islam3zzat: Thanks for the explanation.
> Don't forget install those dependencies directly!

Have we already done that?

tdeekens: Oh course. Why would I forget that. _cough_ fc5448f. Thanks.

islam3zzat: Sorry, I meant `core-js` and  `regenerator-runtime`
as suggested in https://github.com/zloirock/core-js/blob/master/docs/2019-03-19-core-js-3-babel-and-a-look-into-the-future.md#babelpolyfill
> Don't forget install those dependencies directly!
`npm i --save core-js regenerator-runtime`

tdeekens: Yes, they're both added.

packages/mc-scripts/config/create-webpack-config-for-development.js
```diff
@@ -55,8 +55,8 @@ module.exports = ({
   entry: {
     app: [
       // Ship a few polyfills by default
-      require.resolve('./polyfills'),
```
emmenko: Does this need to be removed? 🤔

packages/mc-scripts/config/create-webpack-config-for-development.js
islam3zzat: both 
- regenerator-runtime
- core-js

was in the dependencies tree, but not directly dependencies for `@commercetools-frontend/mc-scripts`, but since we now require them directly we should add them to `mc-scripts`

packages/mc-scripts/config/create-webpack-config-for-development.js
tdeekens: b28a83c

packages/mc-scripts/config/create-webpack-config-for-development.js
tdeekens: Ah. My bad. This should solve this, right? 0b17450

tdeekens:
Note: this is not ready to merge. We get a "the request of a dependency is an expression" in the playground and percy fails to test the bundle. I am unclear why right now.

tdeekens:
The last commit may do the trick. It's an issue with webpack. I am not 100% sure why it occurs now. Anway, the way I changed the message loading is how we do it in apps how. Static and not dynamic `langs` which at least makes the playground start and webpack do its thing without errors.

tdeekens:
Given canary releases should be fixed with #467 we can try this out in apps before publishing 🎉.

tdeekens:
This fails strangely with

CleanShot 2019-03-22 at 09 43 38@2x

tdeekens:
Another note: not ready to merge yet. Will have to investigate later or feel free to help out.

emmenko:
How about doing the updates into separate PRs, given that only with the new babel version there are some problems?

tdeekens:
Sure. Will spin up a PR tmr and rebase it out. Wasn‘t expecting it to have such obscure side-effects.

tdeekens:
This is now focussed only on the babel and core-js update while keeping core-js@2 fornow.

tdeekens:
This is starting to drive me to utter confusion. It's only `yarn playground:build` failing due to `omitEmpty` which is claims to not be exported. Neither named or per dafault. Running the playground works just fine. The built files on both branches look pretty much the same. There is not omitEmpty missing in one and not in another.

emmenko:
I'll take a look at this as well

emmenko:
I opened a separate PR to see if we get the same problems (#477)

tdeekens:
Given both seem to yield similar results which one do you intend to close?

emmenko:
It doesn't matter, as long as we get it to work

emmenko:
I'll experiment a bit more on mine

tdeekens:
> It doesn't matter, as long as we get it to work

True, I'd however prefer to merge this one as it provides far more context and contains more discussions for anybody wanting to replicate or look into things later.

tdeekens:
Closing in favour of #477.

PR #462, opened 2019-03-21
feat(app-shell): quick-access to have noresults message

tdeekens:
#### Summary

Adds a no results message to QuickAccess to give better user feedback.

CleanShot 2019-03-21 at 13 48 26@2x

This is meant to look similar to the offline message.

CleanShot 2019-03-21 at 13 48 49@2x

### DIFF
--- a/packages/application-shell/src/components/quick-access/butler/butler.js
+++ b/packages/application-shell/src/components/quick-access/butler/butler.js
@@ -455,7 +455,10 @@ export class Butler extends React.Component {
                 </div>
               );
 
-            if (this.state.results.length === 0 && this.state.searchText !== '')
+            if (
+              this.state.results.length === 0 &&
+              this.state.searchText.trim().length > 0
+            )
               return (
                 <div className={classnames(styles.noResultsWarning)}>
                   <FormattedMessage {...messages.noResults} />
--- a/packages/application-shell/src/components/quick-access/butler/butler.js
+++ b/packages/application-shell/src/components/quick-access/butler/butler.js
@@ -455,7 +455,12 @@ export class Butler extends React.Component {
                 </div>
               );
 
-            if (!this.state.results) return null;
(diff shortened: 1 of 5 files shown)


packages/application-shell/src/components/quick-access/quick-access.spec.js
```diff
@@ -315,9 +365,6 @@ describe('QuickAccess', () => {
       await waitForElement(() => getByText(offlineText));
       // it should show the offline warning
       expect(getByText(offlineText)).toBeVisible();
-      // it should log the error
-      // eslint-disable-next-line no-console
-      expect(console.error).toHaveBeenCalledTimes(1);
```
tdeekens: I find testing this a point pointless and not RTL'ly.

packages/application-shell/src/components/quick-access/butler/butler.js
```diff
@@ -455,7 +455,12 @@ export class Butler extends React.Component {
                 </div>
               );
 
-            if (!this.state.results) return null;
+            if (this.state.results.length === 0 && this.state.searchText !== '')
```
qmateub: Would make sense to trim the value to check if it's empy? 🤔

packages/application-shell/src/components/quick-access/quick-access.js
```diff
@@ -201,11 +201,11 @@ class QuickAccess extends React.Component {
...
-              /${this.props.project.key}
-              /products
-              /${productId}
-              /${variantId}
-            `,
+                /${this.props.project.key}
+                /products
+                /${productId}
+                /${variantId}
+              `,
```
qmateub: You love fixing identation 😅

packages/application-shell/src/components/quick-access/butler/butler.js
tdeekens: Fair point not sure. It's not done anywhere else. So I presume it would be a separate fix.

PR #467, opened 2019-03-21
chore: remove cypress video

tdeekens:
#### Summary

Now that the working directory actually shows what is dirty

CleanShot 2019-03-21 at 18 30 38@2x

We can fix the problem 😄. This is in the `.gitignore` anyway but somehow was commited I presume (and showhow is pruned) which makes the directory dirtay.


tdeekens:
Alright. Looks like this did the trick. On `master` there are not changes to publish but we don't get the warning by lerna anymore that the working directory is dirty. Let's see what the next merge with actual changes holds for us :)

emmenko:
Thanks! 🚀

PR #476, opened 2019-03-25
refactor(sentry): to use newest sdk over deprecated raven client

emmenko:
I had this on my list for a while. Basically it just uses the new JS client for sentry instead of the deprecated one.


packages/sentry/sentry.js
```diff
@@ -1,68 +1,83 @@
...
-// NOTE: only exported for tests => use reportError below for your code
-export const createSentryErrorReporter = (
-  environment,
-  debugLogger = console,
-  productionLogger = Raven
-) => (error, extraInfo) => {
-  if (error instanceof Error === false && environment === 'development') {
-    debugLogger.warn(
-      '[SENTRY]: You called "sentry.reportError" with an argument that is not an instance of "Error". ' +
+export const reportErrorToSentry = (error, extraInfo, getIsEnabled) => {
```
emmenko: We don't need the factory function anymore as we test things a bit differently now (using the sentry-testkit).

packages/sentry/sentry.js
```diff
@@ -1,68 +1,83 @@
...
+  const environment = window.app.env || process.env.NODE_ENV;
+  const isEnabled = getIsEnabled
+    ? getIsEnabled(environment)
```
emmenko: This is used in tests

packages/sentry/sentry.js
```diff
@@ -1,68 +1,83 @@
...
-    // The error stack should be available in Sentry, so there is no
-    // need to print it in the console as well.
-    // We just notify that an error occurred and provide the error ID.
-    debugLogger.error(`[SENTRY]: An error occured (ID: ${errorId}).`);
+    Sentry.withScope(scope => {
+      if (extraInfo && extraInfo.extra) {
+        if (typeof extraInfo.extra === 'object') {
+          // See https://docs.sentry.io/platforms/javascript/react/
+          Object.keys(extraInfo.extra).forEach(key => {
+            scope.setExtra(key, extraInfo.extra[key]);
```
emmenko: The extra info needs to be passed using the `setExtra` function, as key-values, which is why we need to iterate through the properties.

packages/sentry/sentry.js
```diff
@@ -1,68 +1,83 @@
...
+          });
+        } else {
+          scope.setExtra('extra', extraInfo.extra);
```
emmenko: In case the `extra` info is not an object, we pass it as a value under the key `extra`

packages/sentry/sentry.js
```diff
@@ -1,68 +1,83 @@
...
-      tags: { role: 'frontend' },
       environment: window.app.env,
       // in order to reduce the noise in sentry we only track errors that come
       // from our code and ignore errors that come from other services
       // https://blog.sentry.io/2017/03/27/tips-for-reducing-javascript-error-noise.html
       whitelistUrls: [window.app.cdnUrl, window.app.frontendHost],
-    }).install();
+    });
+    Sentry.configureScope(scope => {
+      scope.setTag('role', 'frontend');
```
tdeekens: Nice API.

PR #477, opened 2019-03-25
chore: upgrade to babel 7.4

emmenko:
This branch aims to do the same things as #461 from a fresh perspective from me and see if we get the same errors/problems as in #461.


packages/application-shell/package.json
```diff
@@ -36,7 +36,6 @@
     "build:test-utils:cjs": "cross-env NODE_ENV=development rollup -c ../../rollup.config.js  -f cjs ./src/test-utils/index.js --o ./test-utils/index.js"
   },
   "dependencies": {
-    "@babel/runtime": "7.3.4",
```
emmenko: To my understanding, this is only needed in our babel preset package

packages/babel-preset-mc-app/package.json
```diff
@@ -21,25 +21,27 @@
...
     "@babel/plugin-transform-react-constant-elements": "7.2.0",
     "@babel/plugin-transform-react-display-name": "7.2.0",
-    "@babel/plugin-transform-regenerator": "7.3.4",
-    "@babel/plugin-transform-runtime": "7.3.4",
-    "@babel/preset-env": "7.3.4",
+    "@babel/plugin-transform-regenerator": "7.4.0",
+    "@babel/plugin-transform-runtime": "7.4.0",
+    "@babel/preset-env": "7.4.2",
     "@babel/preset-react": "7.0.0",
+    "@babel/runtime-corejs2": "7.4.2",
```
emmenko: This is required when using using the `corejs: 2` option in `transform-runtime`

https://babeljs.io/docs/en/next/babel-plugin-transform-runtime#corejs

packages/mc-scripts/config/create-webpack-config-for-development.js
```diff
@@ -56,7 +56,8 @@ module.exports = ({
     app: [
       // Ship a few polyfills by default
       require.resolve('./polyfills'),
-      require.resolve('babel-polyfill'),
+      require.resolve('core-js/stable'),
+      require.resolve('regenerator-runtime/runtime'),
```
emmenko: Replacing the deprecated `@babel/polyfill` requires to manually specifying those 2 polyfills:

https://babeljs.io/blog/2019/03/19/7.4.0#migration-from-core-js-2

packages/mc-scripts/config/create-webpack-config-for-development.js
```diff
@@ -56,7 +56,8 @@ module.exports = ({
     app: [
       // Ship a few polyfills by default
       require.resolve('./polyfills'),
-      require.resolve('babel-polyfill'),
+      require.resolve('core-js/stable'),
```
emmenko: Oh, this is only available in v3

emmenko:
Hmm should we maybe trying to upgrade to v3?

emmenko:
Update:
- I updated `core-js` to use v3.
- I patched the import of `omit-empty` to use `require()` (I tried this approach https://github.com/facebook/create-react-app/pull/6279 but it didn't help).

Let's see what happens

tdeekens:
> Hmm should we maybe trying to upgrade to v3?

I tried that before. That's why there is one commit downgrading to core-js@2.

packages/mc-scripts/config/create-webpack-config-for-development.js
tdeekens: Yes.

packages/mc-scripts/config/create-webpack-config-for-development.js
tdeekens: Correct.

packages/babel-preset-mc-app/package.json
tdeekens: Correct. Same on the other PR.

application-templates/starter/src/components/entry-point/entry-point.js
```diff
@@ -1,17 +1,13 @@
...
   ApplicationShell,
   setupGlobalErrorListener,
   RouteCatchAll,
 } from '@commercetools-frontend/application-shell';
 import { Sdk } from '@commercetools-frontend/sdk';
 import * as globalActions from '@commercetools-frontend/actions-global';
-import { Redirect, Route, Switch } from 'react-router-dom';
-
-const loadApplicationMessagesForLanguage = async lang => {
-  const messages = await import(`../../i18n/data/${lang}.json` /* webpackChunkName: "application-messages-[request]" */);
```
emmenko: FYI: this causes webpack now to throw (no idea why it didn't before)

```bash
Failed to compile.
./src/components/entry-point/entry-point.js 53:19-55:13
Critical dependency: the request of a dependency is an expression
 @ ./src/components/entry-point/index.js
 @ ./src/index.js
 @ multi ../packages/mc-scripts/config/polyfills.js ../packages/mc-scripts/node_modules/core-js/stable/index.js ../packages/mc-scripts/node_modules/regenerator-runtime/runtime.js ./src/index.js
```

tdeekens: I fixed this on my branch too. Did you look at the change set there?

emmenko: I know, I took it from your branch 😅

tdeekens: Ah kay. It sounded like thought you found out something novel 😄.

emmenko: Yes sorry, I'm just "documenting" what I find along the way for reference, even though it might be obvious coming from the other PR.

emmenko:
Alright, now the e2e test fails. I'll try to reproduce it locally

emmenko:
I can't reproduce this locally...wtf?!? 🤔

I'll try regenerating the lockfile

emmenko:
I had to start the app in production mode. So this is the error:

image

emmenko:
So it's still about the way `omit-empty` is being imported. I'll try to fiddle with the config a bit

emmenko:
Alright I think I "fixed" the problem, which is related to another problem we were trying to fix (#397).

I'll try to check on IE if this still works, but somehow the problem was related to this (https://github.com/commercetools/merchant-center-application-kit/pull/477/commits/b6934c9ce1e3bea323c77cc1c7f67244ab2a5f4e)

Maybe now that we are using terser, things will work

tdeekens:
> Maybe now that we are using terser, things will work

This was my last guess too. I thought it was done with a reason. Presumeing that `omitEmpty` module uses an arrow-fn in it's code which remains untranspiled I would assume it doesn't work in IE. I then thought about inlining the fn but got demotivatied this morning seeing that is uses an external `kind-of` dep (which we could obv also make our dep.

emmenko:
So I tried to test this on IE but couldn't manage to (e.g. browserstack free plan does not allow testing in IE, in Edge works though).

I think we can go ahead with this. Should we keep this PR or migrate the changes to the other one?

tdeekens:
> I think we can go ahead with this. Should we keep this PR or migrate the changes to the other one?

We have a paid browserstack plan.

emmenko:
> We have a paid browserstack plan

🤩

emmenko:
Alright, so it's failing on IE because of `omit-empty`

image

image

Alright, I'm going to publish an es version of that library.

tdeekens:
> Alright, I'm going to publish an es version of that library.

Yeah. That's what I would recommend too.

emmenko:
Alright, I published a version `omit-empty-es` (https://github.com/emmenko/omit-empty-es)

emmenko:
Going to release a small fix of `omit-empty-es`, then I'll merge this

emmenko:
Btw: rollup is now showing a bunch of warnings like

image

Not sure what to understand from that.

PR #480, opened 2019-03-26
docs(changelog): for release v11.4.0 [skip ci]

emmenko:


emmenko:
Hmm why was it labelled as a refactoring then?

Anyway, you're right, although v5 is the same as the beta version, so it's half-breaking. Whatever, I think is fine like this, but we should keep that in mind for the next time. Thanks for the reminder!

tdeekens:
> Hmm why was it labelled as a refactoring then?

Cause I also realised only this morning. I also don't think one has to be overly strict but it's good to keep something like this in mind.

PR #482, opened 2019-03-26
refactor(test-utils): to avoid conflicting named exports with rtl render function

emmenko:
Using the latest babel version, we get this error when testing the apps that consume the `test-utils` bundle.

image

The problem seems to be the way we export our custom `render` function, which clashes with the re-exported `render` function of RTL.

Solution: we should export our custom render function under different names.

| Before | After |
| --- | --- |
| `render` | `renderApp` |
| `renderWithRedux` | `renderAppWithRedux` |
| `experimentalRender` | `experimentalRenderAppWithRedux` |

More info:
- https://github.com/kentcdodds/react-testing-library/issues/169
- https://github.com/kentcdodds/react-testing-library/pull/179

### DIFF
--- a/application-templates/starter/src/components/main-view/main-view.spec.js
+++ b/application-templates/starter/src/components/main-view/main-view.spec.js
@@ -1,6 +1,6 @@
 import React from 'react';
 import {
-  render,
+  renderApp,
   waitForElement,
   fireEvent,
 } from '@commercetools-frontend/application-shell/test-utils';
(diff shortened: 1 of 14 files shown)


tdeekens:
Are you sure this fixes the issue? Issues date back to babel@7 as the casue which we already used before quite a bit.

emmenko:
Yes, I already tested this and it works.

The error is also quite clear if you think about it, as we are trying to override a named export. As to why he get this now, I don't know, maybe something changed in the new babel dependencies that will trigger this error.

PR #485, opened 2019-03-27
feat(mc-dev-authentication): extract routes and dev auth into pkg to …

tdeekens:


packages/mc-http-server/load-options.js
```diff
@@ -10,9 +10,10 @@ if (flags.help) {
   Usage: mc-http-server [options]
 
   Options:
-  --config=<path>        (required) The path to the environment config (defined as a JSON file, e.g. "env.json").
-  --csp=<path>           (optional) The path to the custom CSP directives config (defined as a JSON file, e.g. "csp.json").
-  --use-local-assets     (optional) If this option is enabled, the "dist/assets" will be used to start the http-server package. This requires that the assets have been built before running this script.
+  --config=<path>           (required) The path to the environment config (defined as a JSON file, e.g. "env.json").
+  --csp=<path>              (optional) The path to the custom CSP directives config (defined as a JSON file, e.g. "csp.json").
+  --use-local-assets        (optional) If this option is enabled, the "dist/assets" will be used to start the http-server package. This requires that the assets have been built before running this script.
+  --use-dev-authentication  (optional) If this option is enabled, the development "login,logout" routes will be rendered http-server package. This should only be used when running e.g. dist assets.
```
emmenko: I don't think we need this. We can determine that by reading the `servedByProxy` from the `env.json`.

tdeekens: Yeah.

tdeekens:
this works on my machine (tm) now. I will merge to see how UI on the base branch likes it. Thanks for the feedback.

PR #486, opened 2019-03-27
refactor(ci): migrate to CircleCI

### LINKED ISSUE #479: Consider moving CI to CircleCI
Our current CI flow is quite slow (~17min) as we are performing several sequential steps that take some time to finish.

If we move to CircleCI we can spread the jobs into parallel steps, which will decrease the time to execute a CI run.

tdeekens:
#### Summary

Closes #479 by migrating CI setup to CircleCI.


scripts/release_canary.sh
```diff
@@ -4,11 +4,11 @@ set -e
...
 : "${NPM_TOKEN?Required env variable NPM_TOKEN}"
 
+COMMIT_MESSAGE=$(git log --format=oneline -n 1 $CIRCLE_SHA1))
+
 # Only trigger the canary release when:
-# - the branch is `master`
-# - the build has not been triggered by a pull request
 # - the commit message does not contain `[skip publish]`
-if [ "$TRAVIS_BRANCH" = "master" ] && [ "$TRAVIS_PULL_REQUEST" = "false" ] && [[ ! "$TRAVIS_COMMIT_MESSAGE" =~ \[skip\ publish\] ]]; then
+if [[ ! "$COMMIT_MESSAGE" =~ \[skip\ publish\] ]]; then
```
tdeekens: The job is only triggered on `master` (as CircleCI's config is more powerfull). If it is `master` it also can't be a pull request.

scripts/run-percy.js
```diff
@@ -5,40 +5,39 @@ if (!process.env.CI) {
...
   return isMatchingFilesForCommitRange.code === 0;
 };
 
 const shouldSkipPercy = () => {
-  const isPullRequest = process.env.TRAVIS_PULL_REQUEST !== 'false';
-  const isTargetBranchMaster = process.env.TRAVIS_BRANCH === 'master';
+  const isPullRequest = !process.env.CIRCLE_PULL_REQUEST;
 
-  if (isPullRequest || isTargetBranchMaster) {
+  if (isPullRequest) {
```
tdeekens: As above, the notion does not exist on CircleCI. Maybe it can be extracted from the PR URL somehow if really needed.

.circleci/config.yml
```diff
@@ -0,0 +1,135 @@
...
+  - &save_yarn_cache
+    name: 'Saving yarn cache'
+    key: v1-yarn-cache-{{ checksum "yarn.lock" }}
+    paths:
+      - ~/.cache/yarn
+
+executors:
+  node_11:
+    docker:
+      - image: circleci/node:11.10
```
emmenko: Should we stick to node 10 maybe? Or is it because of the yarn version?

.circleci/config.yml
```diff
@@ -0,0 +1,135 @@
...
+
+commands:
+  install_dependencies:
+    description: 'Installing dependencies'
+    steps:
+      - checkout
+      - restore_cache: *restore_yarn_cache
+      - run: yarn install --pure-lockfile
+      - save_cache: *save_yarn_cache
+  build_artefacts:
```
emmenko: Plural is `artifacts`

.circleci/config.yml
```diff
@@ -0,0 +1,143 @@
...
+      - run: yarn run jest --projects jest.eslint.config.js jest.stylelint.config.js --maxWorkers=3 --reporters jest-silent-reporter
+  unit_test:
+    description: 'Running tests'
+    steps:
+      - run: yarn run jest --projects jest.test.config.js --maxWorkers=3 --reporters jest-silent-reporter
+  vrt_test:
+    description: 'Running Visual Regression Tests'
+    steps:
+      - run: yarn run-if:percy
+  build_playgrond:
```
emmenko: `build_playground`

.circleci/config.yml
```diff
@@ -0,0 +1,143 @@
...
+      - run:
+          name: Publishing to npm registry
+          command: ./scripts/release_canary.sh
+
+jobs:
+  lint:
+    executor: node_11
+    steps:
+      - install_dependencies
+      - lint
```
emmenko: What’s the reason from splitting this from tests?

scripts/install_app_template.sh
```diff
@@ -9,7 +9,7 @@ if [ -z "$TEMPLATE" ]; then
   exit 1
```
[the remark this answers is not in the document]
emmenko: Ah yeah we can remove this

.circleci/config.yml
```diff
@@ -0,0 +1,143 @@
...
+          store_artifacts: true
+          parallel: false
+          parallelism: 1
+  unit_test:
+    executor: node_11
+    steps:
+      - install_dependencies
+      - unit_test
+      - build_artefacts
+      - vrt_test
```
emmenko: We could think of having this running as a parallel step like for the e2e tests, it might speed up things a bit more

.circleci/config.yml
tdeekens: Will go anyway due to CircleCI orbs. I am still working on it.

.circleci/config.yml
tdeekens: I'd first like to have it running before optimizing.

tdeekens: That said, I don't see splitting out VRTs gaining too much. The unit tests run fast and we can benefit from a "living" job (checked out code, built etc).

tdeekens: I split them up anyway. Whatever.

.circleci/config.yml
```diff
@@ -0,0 +1,147 @@
...
+      - vrt_test:
+          requires:
+            - lint
+
+  e2e_test:
+    jobs:
+      - cypress/install:
+          name: cypress/install/e2e_playground
+          yarn: true
+          build: yarn playground:build
```
tdeekens: Bleah. This internally uses yarn 1.10. We require > 1.13. It has to be used like this cause the orb interally persists to the workspace. Not sure what do to here.

emmenko: I think you can pass an option to yarn to ignore the check

.circleci/config.yml
emmenko: No, the apps are part of yarn workspace, to ensure that lerna updates the deps when releasing

tdeekens: Ah missed that. Then I can merge the e2e and unit test workflows.

.circleci/config.yml
tdeekens: Thanks will try. Hoping it doesn't come from inside the orb itself.

tdeekens:
Screenshot 2019-03-27 at 21 08 36

Current workflow visually. I think it makes sense to sequentially lint and unit test to then fan out across the time expensive vrt and e2e tests to then require it all to pass before publishing.

Note: I will rename the workflows so they fit the boxes (CirlceCI problem since 1995).

tdeekens:
Nice learning from this is that the new CirlecCI CLI is actually nice. It pulls down any references orbs (e.g. Cypress) to validate the possible args and configurations. Moreover, we could obviously also publish or own orbs if we ever want to share things across repos.

.circleci/config.yml
tdeekens: I don't think this is on "our level". It comes from within the orb [here](https://github.com/cypress-io/circleci-orb/blob/master/src/orb.yml#L83). I really like the orb though and would like to use it. Either we drop it (obviously) or we open a PR for them to accept `args: string` to the orb so we can pass `--ignore-engines`.

tdeekens:
Reading the orb [source](https://github.com/cypress-io/circleci-orb/blob/master/src/orb.yml#L83) is kind of fun. I must admit I am starting to like the CircleCI config and abstractions there more and more (a bit contrary to the UI).

Turns out you can pass a custom executor to run and/or install. Which could help. When passing it to `run` the executor must have cypress in the Docker image for sure. For the install step I am not seeing why I can't just pass our `node_10` executor (hopeing this works). Otherwise options are

1. PR to the Cypress orb to allow `args` being passed so we can pass ` --ignore-engines` (started issue [here](https://github.com/cypress-io/circleci-orb/issues/102)
2. Create a custom image with Cypress we can use to pass to both as an executor with a recent yarn version
3. Remove the engines requirement from our packages
4. Copy over the orb into our source code (inline or external) and reference it from our config (maybe temporarily with 1.)
5. Just not use the orb at all and have everything inline 😭

tdeekens:
Re-iterating on my thoughts: I don't think 4. or 5. would work because we will have to use the cypress image (executor) in our job anyways too just like the orb to be able to run cypress. We do the same in the mc-fe where we do not run into any problems cause we don't have this engines requirement.

.circleci/config.yml
```diff
@@ -0,0 +1,162 @@
...
+commands:
+  install_dependencies:
+    description: 'Installing dependencies'
+    steps:
+      - checkout
+      - restore_cache: *restore_yarn_cache
+      - run: yarn install --pure-lockfile
+      - save_cache: *save_yarn_cache
+  build_artifacts:
+    description: 'Building artefacts'
```
emmenko:
```suggestion
    description: 'Building artifacts'
```

.circleci/config.yml
```diff
@@ -0,0 +1,162 @@
...
+          parallelism: 1
+      - cypress/install:
+          name: install/e2e_starter
+          requires:
+            - build
+          yarn: true
+          # Building again is needed as Cypress writes/attaches to `~/`
+          # which is not configurable and we can not use as a root.
+          build: yarn build && yarn template-starter:build
+      - vrt_test:
```
emmenko: What's the reason for running the vrt tests after the e2e tests? 🤔 
There is no dependency between them, we can run them in parallel

.circleci/config.yml
```diff
@@ -0,0 +1,162 @@
...
+    description: 'Installing dependencies'
+    steps:
+      - checkout
+      - restore_cache: *restore_yarn_cache
+      - run: yarn install --pure-lockfile
+      - save_cache: *save_yarn_cache
+  build_artifacts:
+    description: 'Building artefacts'
+    steps:
+      - run: yarn build
```
emmenko: We need to persist to workspace right?

.circleci/config.yml
tdeekens: They run after the e2e tests finished installing. I ran them with the install before but it feels reasonable to have the actual tests: e2es and vrts run in parallel after the e2e are finished installing and running too. It's minor though. I would like to keep it like this it makes the workflow graph make a bit more sense.

README.md
```diff
@@ -3,7 +3,7 @@
...
+  <a href="https://circleci.com/gh/commercetools/merchant-center-application-kit"><img src="https://circleci.com/gh/tdeekens/flopflip.svg?style=shield&circle-token=2c30a67460ed10ad965404b648940c774b3b34a3" alt="CI status" /></a> <a href="https://github.com/commercetools/merchant-center-application-kit/releases"><img src="https://badgen.net/github/release/commercetools/merchant-center-application-kit" alt="Latest release" /></a> <a href="https://github.com/commercetools/merchant-center-application-kit/blob/master/LICENSE"><img src="https://badgen.net/github/license/commercetools/merchant-center-application-kit" alt="GitHub license" /></a>
```
montezume: why flopflip.svg?

.circleci/config.yml
tdeekens: It doesn't seem that easy. I tried for a long time to have this working. The problem is the following (as I tried to document with comments):

1. The Cypress orb uses the `~/` as its root
2. We use `~/merchant-center-application-kit`
    - We can not use `~/` cause CircleCI then fails to checkout code and they recommend using `~/<projectName>`

This leaves us with a bit of problems as we want to save from `~/merchant-center-application-kit` post build to `~/` for Cypress to be able to pick it up. I tried to persist `packages` to both `/~` and `~/merchant-center-application-kit` however still it wasn't picked up correctly. I can commit c change again and see if you spot the error.

.circleci/config.yml
```diff
@@ -0,0 +1,213 @@
...
+  build_artifacts:
+    description: 'Building artifacts'
+    steps:
+      - run: yarn build
+  save_workspace:
+    steps:
+      - persist_to_workspace:
+          root: *working_directory
+          paths:
+            - packages/**/dist
```
emmenko: Hmm don't we also need the `package.json` etc?

.circleci/config.yml
```diff
@@ -0,0 +1,213 @@
...
+
+workflows:
+  version: 2
+  build_and_test:
+    jobs:
+      - lint:
+          filters:
+            branches:
+              only:
+                - refactor/use-circleci
```
emmenko: Reminder to remove this

.circleci/config.yml
```diff
@@ -0,0 +1,213 @@
...
+          requires:
+            - build
+      - e2e_test:
+          name: e2e_playground
+          requires:
+            - build
+          spec: 'cypress/integration/playground/**/*.js'
+          build: yarn playground:build
+          start: yarn playground:start:prod:local
+          wait-on: 'http://localhost:3001'
```
emmenko: This is nice 👌

.circleci/config.yml
tdeekens: We could yeah. But we checkout before restoring the workspace. I find that more intuitive. You checkout the code and restore what you intend to carry on. That means that you don't persist everything (inlcuding all the clutter).

tdeekens:
Current look of the workflow.

CleanShot 2019-03-28 at 14 57 34@2x

.circleci/config.yml
emmenko: Ah right, I missed that part. Thanks

scripts/run-percy.js
```diff
@@ -5,8 +5,15 @@ if (!process.env.CI) {
   throw new Error(`This script is meant to be executed only on CI`);
 }
 
+const COMPARE_URL = shelljs.exec(
+  `cat ${process.env.CIRCLE_WORKING_DIRECTORY}/CIRCLE_COMPARE_URL.txt`
+);
 const COMMIT_RANGE = shelljs.exec(
-  `echo ${process.env.CIRCLE_COMPARE_URL} | cut -d/ -f7`
```
tdeekens: This env variable is actually not set. Known issue at CirleCI.

.circleci/config.yml
```diff
@@ -27,6 +27,9 @@ executors:
         environment:
           TERM: xterm
 
+orbs:
+  compare-url: iynere/compare-url@0.4.9
```
tdeekens: They recommend using this orb to reconstruct the url and use it as...

scripts/run-percy.js
```diff
@@ -5,8 +5,15 @@ if (!process.env.CI) {
...
+  `echo ${COMPARE_URL} | sed 's:^.*/compare/::g'`
```
tdeekens: ...such which I kind of like. I start liking orbs.

.circleci/config.yml
emmenko: I see, ok

scripts/run-percy.js
tdeekens: Screw `sed` (great tool). I can't handle it. I'll just split that url up.

.circleci/config.yml
```diff
@@ -0,0 +1,232 @@
...
+jobs:
+  lint:
+    executor: node_10
+    steps:
+      - install_dependencies
+      - lint
+  unit_test:
+    executor: node_10
+    steps:
+      - install_dependencies
```
emmenko: Q: https://github.com/commercetools/ui-kit/pull/615#discussion_r270097618

.circleci/config.yml
```diff
@@ -0,0 +1,232 @@
...
+    name: 'Restoring yarn cache'
+    keys:
+      - v1-yarn-cache-{{ checksum "yarn.lock" }}
+      - v1-yarn-cache
+
+  - &save_yarn_cache
+    name: 'Saving yarn cache'
+    key: v1-yarn-cache-{{ checksum "yarn.lock" }}
+    paths:
+      - ~/.cache/yarn
```
emmenko:
```suggestion
      - ~/.cache/yarn
      - node_modules
```

tdeekens: As mentioned on the other PR. Won't do. Can't find this as a recommendation on CircleCI's docs. It's a cache not the installed modules.

tdeekens:
Alright. It all seems to work like before. Will merge. I might spin up a PR later to share installed node_modules between steps.

PR #489, opened 2019-03-29
fix: save other dist folders too

tdeekens:


tdeekens:
I need to look into the orb but it's unrelated to this.

PR #491, opened 2019-03-29
fix: update flopflip

tdeekens:


tdeekens:
I tested this version of flopflip before manually in the app and it works. Just didn't get to update it (and didn't have a reason). Now we do :)

tdeekens:
VRTs are failing again due to the orb as it's the only commit on the branch. Wondering if I can make this conditional then somehow. It's odd.

PR #492, opened 2019-03-29
Decoupled html template from Webpack

emmenko:
Slightly relates to #293 
Requirement for #214 (see follow up PR)

I tried to split the changes into commits, for ease reviews.

At the moment, most of our tooling/packages are built around Webpack. However, developing applications with Webpack is not the only option and we wanted to allow other bundlers to be used.

One of the most important places around bundling an application is within the `mc-html-template` package. This contains the raw html template and the utilities to compile it.

Therefore, this PR decouples the package to expose generic methods that do not necessarily rely on Webpack. There is though an extracted `webpack.js` entry point that does some mapping on the injected assets (see webpack config and the README of the html-template package).

Most of the files have only been moved around, I'll try to point out the important changes in inline comments.


packages/application-shell/src/index.js
```diff
@@ -1,8 +1,6 @@
 /**
  * THIS IS THE MODULE ENTRY POINT!
  */
-import './public-path';
```
emmenko: As mentioned, this is now required in the webpack config entry points.

packages/assets/images/ct-logo.svg
```diff
@@ -1,34 +1,9 @@
...
-                        <path d="M12.5033697,13.1677266 L12.5033697,16.7871381 C12.3299478,16.7871381 12.1565733,16.8239203 11.995159,16.8989559 L0.348932131,22.3122449 C0.136592311,22.4106312 0.000711912672,22.623683 0.000711912672,22.8572852 L0,6.76986236 C0,6.76986236 0,6.30602753 0,6.36288562 C0,6.59719981 0.134456573,6.80996678 0.34608448,6.90906502 L12.0100617,12.3693403 C12.1664452,12.4429521 12.3349312,12.4804461 12.5033697,12.4825819 C12.6513052,12.4847176 12.7992406,12.459943 12.9401044,12.4075463 L12.9202658,12.4167062 C12.5579972,12.5841481 12.5033697,12.7512103 12.5033697,13.1677266" id="Fill-18" fill="#CCCCC7"></path>
-                    </g>
-                </g>
-            </g>
-        </g>
-    </g>
-</svg>
\ No newline at end of file
+<svg width="200" height="29" xmlns="http://www.w3.org/2000/svg">
```
emmenko: I run this SVG through the https://jakearchibald.github.io/svgomg/

packages/assets/index.js
```diff
@@ -0,0 +1 @@
+exports.packageLocation = __dirname;
```
emmenko: I haven't found a better way yet to get the package location, as we need to copy files based on that

packages/mc-html-template/README.md
```diff
@@ -11,3 +11,72 @@ This package contains utils and scripts related to the `index.html` for a MC app
 ```bash
 $ npm install --save @commercetools-frontend/mc-html-template
 ```
+
+## API
```
emmenko: I tried to document here the exposed methods, let me know if it's clear enough

packages/mc-html-template/lib/generate.js
```diff
@@ -0,0 +1,77 @@
...
+    <div id="app-loader">
+      <!-- Loading screen styles -->
+      <style>__LOADING_SCREEN_CSS__</style>
+
+      <div class="loading-screen loading-screen--hidden">
+        <svg class="loading-spinner" viewBox="0 0 40 40">
+          <path class="loading-spinner-circle" d="M20.201,5.169c-8.254,0-14.946,6.692-14.946,14.946c0,8.255,6.692,14.946, 14.946,14.946 s14.946-6.691,14.946-14.946C35.146,11.861,28.455,5.169,20.201,5.169z M20.201,31.749c-6.425,0-11.634-5.208-11.634-11.634c0-6.425,5.209-11.634,11.634-11.634c6.425,0,11.633,5.209,11.633,11.634C31.834,26.541,26.626,31.749,20.201,31.749z"></path>
+          <path class="loading-spinner-pointer" d="M26.013,10.047l1.654-2.866c-2.198-1.272-4.743-2.012-7.466-2.012h0v3.312h0 C22.32,8.481,24.301,9.057,26.013,10.047z"></path>
+        </svg>
+        <svg width="200" height="29" xmlns="http://www.w3.org/2000/svg">
```
emmenko: This is the optimized SVG, just to save some bytes 😉

packages/mc-html-template/lib/html-scripts/.eslintrc.yaml
```diff
@@ -0,0 +1,2 @@
+rules:
+  prefer-arrow-callback: 0
```
emmenko: This file has only been moved

packages/mc-html-template/lib/html-scripts/data-layer.js
```diff
@@ -0,0 +1 @@
+window.dataLayer = [{ 'gtm.start': new Date().getTime(), event: 'gtm.js' }];
```
emmenko: This file has only been moved

packages/mc-html-template/lib/html-scripts/index.js
```diff
@@ -0,0 +1,22 @@
+/* eslint-disable no-console */
```
emmenko: This file has only been moved

packages/mc-html-template/lib/html-scripts/loading-screen.js
```diff
@@ -0,0 +1,26 @@
+// Assigning global callback used by ConfigureIntlProvider to remove
```
emmenko: This file has only been moved

packages/mc-html-template/lib/html-styles/index.js
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable no-console */
```
emmenko: This file has only been moved

packages/mc-html-template/lib/html-styles/loading-screen.css
```diff
@@ -0,0 +1,50 @@
+.loading-screen {
```
emmenko: This file has only been moved

packages/mc-html-template/lib/load-env.js
```diff
@@ -0,0 +1,39 @@
+const fs = require('fs');
```
emmenko: This file has only been moved

packages/mc-html-template/lib/utils/replace-html-placeholders.js
```diff
@@ -14,8 +14,10 @@ const replaceHtmlPlaceholders = (indexHtmlContent, config) =>
   indexHtmlContent
     .replace(
       new RegExp('__CDN_URL__', 'g'),
-      // Ensure there is a trailing slash
-      `${config.cdnUrl.replace(/\/$/, '')}/`
+      config.cdnUrl
```
emmenko: In case the `cdnUrl` value is an empty string, we shouldn't append a trailing slash

packages/mc-html-template/webpack.js
```diff
@@ -0,0 +1,25 @@
+const generateTemplate = require('./lib/generate');
+
+module.exports = templateParams => {
```
emmenko: This is the extracted part related to the html webpack plugin, which only does some mapping

packages/mc-http-server/bin/start-mc.js
```diff
@@ -5,7 +5,12 @@ const fs = require('fs');
 const path = require('path');
 const shell = require('shelljs');
 const fetch = require('node-fetch');
-const replaceHtmlPlaceholders = require('@commercetools-frontend/mc-html-template/utils/replace-html-placeholders');
+const {
+  replaceHtmlPlaceholders,
+} = require('@commercetools-frontend/mc-html-template');
+const {
+  packageLocation: assetsPath,
```
emmenko: This is how the exported `__dirname` is useful

emmenko:
NOTE: it's a breaking change because of the moved files in the `mc-html-template` package.

We can maybe avoid that by having proxy files.

tdeekens:
May I ask regarding

> At the moment, most of our tooling/packages are built around Webpack. However, developing applications with Webpack is not the only option and we wanted to allow other bundlers to be used.

Has that been actively requested by users already? Is that something consumers of custom apps have experienced as a painpoint?

packages/assets/package.json
```diff
@@ -23,5 +23,7 @@
     "package.json",
     "LICENSE",
     "README.md"
-  ]
+  ],
+  "main": "index.js",
+  "module": "index.js"
```
tdeekens: Is it ESM or why do we specify this?

packages/mc-html-template/README.md
```diff
@@ -11,3 +11,72 @@ This package contains utils and scripts related to the `index.html` for a MC app
...
+If you use Webpack with the `HtmlWebpackPlugin`, you can pass the `webpack` entry point that will map the Webpack template params to our generic `generateTemplate` method.
+
+```js
+new HtmlWebpackPlugin({
+  template: require.resolve(
+    '@commercetools-frontend/mc-html-template/webpack'
+  ),
+  // ...
+}),
+```
```
tdeekens: thanks!

packages/mc-html-template/webpack.js
```diff
@@ -0,0 +1,25 @@
...
+    .concat(cssAppChunks)
+    .map(fileName => fileName.replace(/^\//, ''));
+
+  const scriptChunks = templateParams.htmlWebpackPlugin.files.js.map(fileName =>
+    // Trim leading slash, the CDN_URL will ensure to have a trailing slash
+    // (see `replaceHtmlPlaceholders`)
+    fileName.replace(/^\//, '')
+  );
+
+  return generateTemplate({ cssChunks, scriptChunks });
```
tdeekens: Nit: maybe `jsChunks` makes it clearer?

packages/assets/index.js
tdeekens: Package location is not the pkg.json, right?

packages/mc-http-server/bin/start-mc.js
tdeekens: Ah we need the string not the resolved thing. Got it.

packages/assets/index.js
emmenko: It's the location of the folder

packages/assets/package.json
emmenko: It's probably not necessary but we usually specify both options in our packages. Nothing more

emmenko:
> Has that been actively requested by users already? Is that something consumers of custom apps have experienced as a painpoint?

Not really, at least not to my knowledge. However, because people don't ask for it, it does not mean that it's not useful.

Besides that, I personally would like to do that for a couple of reasons:
- it allows us to also experiment internally with different approaches that can potentially bring benefits, like using different bundlers, serving the apps differently, etc.
- it's good for documentation purposes to have different examples of different approaches
- it makes our tooling more clean and flexible, without introducing more complexity

tdeekens:
I would make this a breaking release to be honest. Maintaining proxy exports and removing them eventually sounds a bit of a hassle.

emmenko:
Yeah, it's mostly about keeping it backwards compatible until we bump the next major.

tdeekens:
> Yeah, it's mostly about keeping it backwards compatible until we bump the next major.

Given it's not too much hassle we could do it util we ship sign up/login/pw reset which will be breaking mostly anyway.

emmenko:
Exactly

emmenko:
Actually we only use this package internally in the other packages, I'm pretty sure nobody is using this directly. So it's fine to keep it like this. I removed the breaking change label.

tdeekens:
> I removed the breaking change label.

Yeah, true. Makes sense not to expect anybody using it.

PR #494, opened 2019-03-29
docs(changelog): for v12.1.0

tdeekens:
Would like to release this to unblock a branch by @jonnybel on sign up.


emmenko:
Why not using the canary release?

tdeekens:
> Why not using the canary release?

We're doing that already. We use it on a branch and it works. We'd however later like to merge the branch to `master` and then do not want to use the canary release anymore. Makes sense?

tdeekens:
> to suggest it to avoid making a new release

Sure thanks. Makes sense too. I'd however in this case have this as `next` at least as it's nothing potential temporary and I wanted to have the flopflip update to be stable for a bit longer already.

PR #506, opened 2019-04-02
fix(app-shell): css height for app container

### LINKED ISSUE #505: Vertical alignment of "Ice cream" page
**Describe the bug**

image

emmenko:
Fixes #505 

**Before**

image

**After**

image

### DIFF
--- a/packages/application-shell/src/components/application-shell/application-shell.mod.css
+++ b/packages/application-shell/src/components/application-shell/application-shell.mod.css
@@ -3,6 +3,10 @@
   --menuWidth: 200px;
 }
 
+:global(#app) {
+  height: 100%;
+}
+
 .app-layout {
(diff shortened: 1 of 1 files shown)


packages/application-shell/src/components/application-shell/application-shell.mod.css
```diff
@@ -3,6 +3,10 @@
   --menuWidth: 200px;
 }
 
+:global(#app) {
+  height: 100%;
+}
```
tdeekens: Was this something we removed before?

emmenko: No I think because I changed the layout to use css-grid (previously was flexbox) and this was missing for such pages.

PR #511, opened 2019-04-03
feat(i18n): support new locales fr-FR and zh-CN

emmenko:
This PR refactors different packages to support new locales: `fr-FR` and `zh-CN`.

Most of the changes are related to the fact that now we have to deal with **full locales** (e.g. `fr-FR`) as opposed to **language tags** (e.g. `de`).

For consistency, I also renamed different props to be `locale` instead of `language`.

In theory there shouldn't be any breaking changes.

Please also carefully check the newly generated l10n data (e.g. currencies, countries, etc).


packages/l10n/data/countries/en.json
```diff
@@ -149,7 +149,7 @@
   "mf": "St. Martin",
   "mg": "Madagascar",
   "mh": "Marshall Islands",
-  "mk": "North Macedonia",
+  "mk": "Macedonia",
```
montezume: can we revert this?

packages/l10n/data/countries/de.json
```diff
@@ -149,7 +149,7 @@
   "mf": "St. Martin",
   "mg": "Madagaskar",
   "mh": "Marshallinseln",
-  "mk": "Nordmazedonien",
+  "mk": "Mazedonien",
```
montezume: can we revert this?

packages/l10n/data/countries/es.json
```diff
@@ -149,7 +149,7 @@
   "mf": "San Martín",
   "mg": "Madagascar",
   "mh": "Islas Marshall",
-  "mk": "Macedonia del Norte",
+  "mk": "Macedonia",
```
montezume: can we revert this?

packages/i18n/README.md
```diff
@@ -56,28 +58,43 @@ const MyApplication = props => (
...
+  switch (lang) {
+    case 'de':
+      loadAppI18nPromise = import('./i18n/data/de.json' /* webpackChunkName: "app-i18n-de" */);
+      break;
+    case 'es':
+      loadAppI18nPromise = import('./i18n/data/es.json' /* webpackChunkName: "app-i18n-es" */);
+      break;
+    default:
+      loadAppI18nPromise = import('./i18n/data/en.json' /* webpackChunkName: "app-i18n-en" */);
+  }
```
tdeekens: 💯 We shouldn't be adding the new locales here yet right?

packages/i18n/utils.js
```diff
@@ -1,15 +1,18 @@
...
-      return 'es';
-    default:
-      return 'en-gb';
-  }
+export const mapLocaleToMomentLocale = locale => {
+  if (locale.startsWith('de')) return 'de';
+  if (locale.startsWith('es')) return 'es';
+  if (locale.startsWith('fr')) return 'fr';
+  if (locale === 'zh-CN') return 'zh-cn';
+  return 'en-gb';
```
tdeekens: 😕 why is the default for moment different than for intl?

packages/l10n/time-zone-information.js
```diff
@@ -1,14 +1,18 @@
...
+  const intlLocale = mapLocaleToIntlLocale(locale);
+  switch (intlLocale) {
     case 'de':
       return import(/* webpackChunkName: "timezone-data-de" */ './data/time-zones/de.json');
     case 'es':
       return import(/* webpackChunkName: "timezone-data-es" */ './data/time-zones/es.json');
+    case 'fr-FR':
+      return import(/* webpackChunkName: "timezone-data-fr-FR" */ './data/time-zones/fr-FR.json');
+    case 'zh-CN':
+      return import(/* webpackChunkName: "timezone-data-zh-CN" */ './data/time-zones/zh-CN.json');
```
tdeekens: Eventually all apps would have to adjust their `message.js` too, correct?

packages/l10n/utils.js
```diff
@@ -0,0 +1,41 @@
+export const mapLocaleToIntlLocale = locale => {
+  if (locale.startsWith('de')) return 'de';
+  if (locale.startsWith('es')) return 'es';
+  if (locale.startsWith('fr')) return 'fr-FR';
+  if (locale === 'zh-CN') return 'zh-CN';
+  return 'en';
+};
+
+const supportedLocales = ['en', 'de', 'es', 'fr-FR', 'zh-CN'];
```
tdeekens: Wondering if we should share this somehow in a pkg?

packages/l10n/utils.js
```diff
@@ -0,0 +1,41 @@
...
+const defaultLocale = 'en';
+
+// Given a locale, return the locale only if it's whitelisted.
+// If not, return a default locale.
+export function getSupportedLocale(locale) {
+  const isSupported = supportedLocales.find(supportedLocale =>
```
tdeekens: Ah okay. Just to recap for me: given we have a `language` not really a `locale` in `supportedLocales` (like `es`) we whitelist all of their corresponding `locales` (as they start with `es`).

packages/mc-scripts/extract-intl.js
```diff
@@ -30,10 +35,11 @@ if (commands.length === 0 || (flags.help && commands.length === 0)) {
   Usage: mc-scripts extract-inl [options] <glob-pattern>..
 
   Options:
-  --output-path         The location where to put the extracted messages
-  --locale=<locale>     (optional) The default locale to use [default "en"]
-  --build-translations  (optional) In case you want to manually build the locale files with the translations [default "false"]
-  --overwrite-core      (optional) By default, if a core.json file exists, existing keys won't be overwritten. This is to ensure that messages in core.json can be updated from external sources. In case you want to avoid this check, you can force writing the extracted messages to core.json [default "false"]
+  --output-path                       The location where to put the extracted messages
+  --locale=<locale>                   (optional) The default locale to use [default "en"]
+  --locales=<locale1,locale2,...>     (optional) The supported locales to map to [default ${supportedLocales.toString()}]
```
tdeekens: Nice.

packages/l10n/time-zone-information.js
emmenko: Depends. Our apps yes, custom apps probably not (that's why I also kept playground and the starter app without)

packages/l10n/utils.js
emmenko: I thought about it but each package uses the info a bit differently, so I would prefer to keep it co-located to each package for now.

packages/l10n/utils.js
emmenko: Yes, I thought this was the best approach to handle both formats

packages/mc-scripts/extract-intl.js
emmenko: This can be used for apps that only require a subset of the locales. By default, we use all our supported locales.

packages/i18n/utils.js
emmenko: That's how their data files are defined 🤷‍♂️

packages/l10n/data/countries/es.json
emmenko: That's the data we get from cldr 🤔

packages/l10n/utils.js
tdeekens: I just felt it being a bit "dangerous" given things get out of sync across packages but lets see how it goes.

packages/l10n/utils.js
tdeekens: Sure. I was just thinking out loud.

packages/l10n/data/countries/es.json
montezume: Yes but it's *wrong*. [ref](https://www.theguardian.com/world/2019/feb/12/nato-flag-raised-ahead-of-north-macedonias-prospective-accession)

emmenko: So does it mean that we can't run the script anymore, or that we need to manually patch things up? Is that the only _wrong_ thing that you found or are there are cases?

emmenko:
Alright, CI passes, this is ready for review now.

I'll do some more testing locally for using one of the new locales...

emmenko:
I did some testing locally (e.g. using `fr` and `fr-FR`) and it works as expected.

From my side, there is nothing else to do for now.

emmenko:
I'll push a commit shorty with the latest cldr data.

emmenko:
Done, now we have the latest cldr data `v35`, dated `2019-03-27` (http://cldr.unicode.org/index/downloads) 🎉

packages/l10n/data/countries/es.json
emmenko: Fixed in https://github.com/commercetools/merchant-center-application-kit/pull/511/commits/1a92f79cb3ee8798cdb77b20ba077e49ce47cba6

emmenko: PS: apparently not all languages have the latest translations (e.g. for "north macedonia") 🤷‍♂️

montezume: I added the translations in a previous PR for north macedonia.

If everyone is fine with it, I guess we can revert to the "wrong" countries and wait until CLDR catches up 🤷‍♂️ I tried to find a repo on which to log an  issue but I couldn't find one, is it some mailing list? Do you know how to contact them about their data being outdated?

montezume:
@emmenko nice!

packages/l10n/data/countries/es.json
emmenko: I'd rather prefer to use the official list instead of patching some names manually. They eventually catch up, so I wouldn't worry much about it.

montezume: Can you look into contacting them about the outdated countries?

emmenko: I found this ticket related to the macedonia translations (from 7 days ago): https://unicode.org/cldr/trac/changeset/14975

I suppose that those will be available in the next release.

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -53,7 +50,7 @@ const createTestProps = props => ({
 
 const testLocaleData = {
   isLoading: false,
-  language: 'en',
+  locale: 'en',
```
montezume: I'm wondering what the deal is with renaming language to locale since AFAIK this is a language and not a locale?

emmenko: The renaming was for two reasons:
- to fix the inconsistencies
- to technically be correct, as `fr-FR` is a locale and not a language anymore

PR #513, opened 2019-04-04
chore: fix canary releases

tdeekens:


.circleci/config.yml
```diff
@@ -48,10 +48,10 @@ commands:
       - persist_to_workspace:
           root: *working_directory
           paths:
-            - packages/**/dist
-            - playground/dist
-            - visual-testing-app/dist
-            - application-templates/**/dist
+            - packages
```
tdeekens: This was premature. packages (e.g. application-shell) sometimes contain things from src they would actually like to reference and publish. For application-shell this is test-utils.

tdeekens:
> I right, I thought I mentioned this last time I reviewed that.

We dicussed if this would work without and thouhgt yes. So we wanted to attach/persist less to the workspace. However it doesn't work 😄

PR #515, opened 2019-04-04
fix(app-shell): to unset previously set cached value

tdeekens:
#### Summary

This pull request fixes and edge case with the following repo case:

1. Log in
2. Remove the `mcAccessToken` cookie
3. Reload the page

Then you will end up in an infite redirect loop as the app assumes the user is still logged in (local storage cache) while the query `amILoggedIn` in the background fails due to the missing cookie.

### DIFF
--- a/packages/application-shell/src/components/authenticated/authenticated.spec.js
+++ b/packages/application-shell/src/components/authenticated/authenticated.spec.js
@@ -14,6 +14,9 @@ const createTestProps = custom => ({
 });
 
 describe('rendering', () => {
+  beforeEach(() => {
+    storage.put.mockClear();
+  });
   let props;
   describe('when authenticated state was cached in local storage', () => {
     beforeEach(() => {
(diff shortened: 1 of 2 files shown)


packages/application-shell/src/components/authenticated/authenticated.spec.js
```diff
@@ -90,7 +93,17 @@ describe('rendering', () => {
       it('should not put `isAuthenticated` into local storage', async () => {
         await wait(
           () => {
-            expect(storage.put).not.toHaveBeenCalledWith();
+            expect(storage.put).not.toHaveBeenCalled();
```
tdeekens: This fella was a false positive (passing falsely) as the mock was not called with `undefined`.

tdeekens:
I will take the stride to merge this as I'd like to unblock `master` and am very confident about this change.

PR #516, opened 2019-04-04
docs(changelog): for v13.0.1 [skip ci]

tdeekens:


tdeekens:
Verified with the canary that this fixes the issue.

PR #519, opened 2019-04-05
refactor(i18n): to load the i18n messages async

emmenko:
Following the [suggestions from ui-kit](https://github.com/commercetools/ui-kit/pull/643), we can actually treat the ui-kit translations as we do for the other localization files.

### DIFF
--- a/packages/application-shell/src/components/application-shell-provider/application-shell-provider.js
+++ b/packages/application-shell/src/components/application-shell-provider/application-shell-provider.js
@@ -1,3 +1,4 @@
+import './global-style-imports';
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Router } from 'react-router-dom';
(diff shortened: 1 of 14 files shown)


packages/application-shell/src/components/application-shell-provider/application-shell-provider.js
```diff
@@ -1,3 +1,4 @@
+import './global-style-imports';
```
montezume: makes sense to put this at the top 👍

packages/i18n/load-i18n.js
```diff
@@ -38,7 +39,23 @@ const getMomentChunkImport = locale => {
   }
 };
 
-const getLocalizedStringsChunkImport = locale => {
+const getUiKitChunkImport = locale => {
+  const intlLocale = mapLocaleToIntlLocale(locale);
+  switch (intlLocale) {
+    case 'de':
+      return import(/* webpackChunkName: "i18n-ui-kit-locale-de" */ '@commercetools-frontend/ui-kit/i18n/de.json');
```
emmenko: FYI: we just need to remember to update the path to `.../i18n/data/...` once we do the next ui-kit release

tdeekens: Yeah it's funny that these things tend to be breaking releases from the view of UIKit.

montezume: seems like an easy thing to forget 🙈

emmenko: It depends when we are going to release ui-kit :)

We can also revert the folder structure in the ui-kit...

emmenko: I guess we need to wait anyway for the ui-kit release, as we expect the new locales to be there

emmenko:
Waiting for a new ui-kit release, in the meantime this is blocked.

montezume:
Can use ui-kit `9.2.0`

emmenko:
FYI: there are some failing VRT tests (unrelated to this PR). There is a related PR to fix one of the problems (#542)

PR #527, opened 2019-04-08
fix(react-notifications): to use errorByExtension field

tdeekens:
#### Summary

This pull request changes the detection when rendering the `localizedMessage` coming from an request having gone through an API Extension.

Note that the response of an API Extension contains:

```json
"errorByExtension": {
   "id": "<extension-id>",
   "key": "my-extension"
}
```

The `localizedMessage` does not have to be present but can be set by the user of the API Extension.

More can be found here: https://docs.commercetools.com/http-api-projects-api-extensions.html#error-cases


tdeekens:
Note, that this is not a bug just a request by API developers to use this field for detection.

tdeekens:
The VRTs are just failing as CircleCI's orb to detect the commit range fails. Still, not sure what to do about it.

PR #530, opened 2019-04-09
fix: allow react-redux@7.x as peer dep

tdeekens:
#### Summary

This pull request adds `"6.x || 7.x"` for react-redux. react-redux@7 is compatible to v6 however requires react@16.8.4. React itself is a peer dependency of `16.8`. The consumer of app-kit should pick a pairing of both matching. Additionally, this PR updates react-redux as a dev dep to v7.


tdeekens:
I like that we have E2E tests for starter and playground. Give some really decent confidence in things working over all.

PR #542, opened 2019-04-15
fix(dialogs): content overflow. Refs #534

### LINKED ISSUE #533: FormDialog  problem showing data in select input
**Describe the bug**
I have a problem when render a FormDialog modal (i think its the same for all modals) when put a Select Input inside of it and display the data the info show inside of modal and I have to scroll down to see the values.

I try to hardcode .dialog-styles-mod_dialog-content__{dinamic} class with `overflow: visible`and work as spected.

**To Reproduce**
1. Create a FormDialog 
2. Put as children an input field.
3. Display the values and you can see it inside of modal and had to scroll down.

**Expected behavior**
See the dropdown values in front of modal.

**Screenshots**
image

emmenko:
In #534 there was a change to attempt to fix #533 

Unfortunately that change broke the normal scrolling behaviour, so we need to find another solution for the dropdown problem.

### DIFF
--- a/packages/application-components/src/components/dialogs/internals/dialog-styles.mod.css
+++ b/packages/application-components/src/components/dialogs/internals/dialog-styles.mod.css
@@ -119,6 +119,6 @@
 .dialog-content {
   border-top: 1px solid var(--border-color-separator);
   padding: var(--spacing-16) 0 var(--spacing-8);
-  overflow: visible;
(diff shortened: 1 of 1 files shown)


emmenko:
The `CIRCLE_COMPARE_URL` problem is really annoying 😑

PR #545, opened 2019-04-16
feat: add project not initialized screen

tdeekens:
#### Summary

The API has the notion of initialized/non-initialized projects. Given that when projects are created they can remain in an uninitialized state this behaviour needs to visually be reflected in the MC.

#### Description

CleanShot 2019-04-16 at 09 18 56@2x


packages/i18n/data/core.json
```diff
@@ -47,6 +47,8 @@
   "ProjectNotFound.paragraph1": "The Project you are looking for either does not exists or you are not authorized to view it. Select another Project to view.",
   "ProjectNotFound.paragraph2": "Please contact your system administrator if you have any further questions.",
   "ProjectNotFound.title": "We could not find this Project",
+  "ProjectNotInitialized.paragraph1": "Initialization should not take longer than a few minutes. Please contact us at {mailto} in case the project does not get initialized.",
+  "ProjectNotInitialized.title": "Your project is not initialized yet",
```
montezume:
```suggestion
  "ProjectNotInitialized.title": "Your project has not yet been initialized",
```

packages/application-shell/src/components/project-not-initialized/index.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+const ProjectNotInitialized = React.lazy(() =>
+  import(
+    './project-not-initialized' /* webpackChunkName: "project-not-initialzied" */
```
montezume: nitpick: typo in webpack chunk name

~~ initialzied~~ initialized

packages/assets/images/project-not-initialized.svg
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
emmenko: Is this the same as the "project-expired" svg? 🤔

tdeekens: Yes, which works fine for now I think. It's the exact same hourglass we have in the async creation in the MC. So it speaks the same visuals.

tdeekens:
⚠️ This needs an update in our gateway to fetch the field obviously ⚠️

packages/assets/images/project-not-initialized.svg
montezume: should we then rename the SVGs like we did the icons? 

~project-expeired~ hourglass?

emmenko: Do we need to have 2 identical svgs then?

tdeekens: I think we're speaking of the same thing again.

1. No, we do not need two SVGs
2. Yes, using the `ProjectSuspendedSVG` in the `ProjectNotInitialized` screeen feels odd
3. Yes, we can improve the naming.
4. Yes, Duplicating the SVG seems to be the least evil also as it is code split.

However, changing the naming is a breaking change to the assets package. We can also duplicate even more to not have it breaking and export generic and specific names. Howveer, I think that's a bigger evil again even.

emmenko: Hmm ok, whatever

tdeekens: Are you willing to go through a breaking change to have different namings in the assets package for the SVGs?

montezume: Let's track is as a separate issue then, and do it in a followup as suggested by @tdeekens. We can ask design to come up with semantic names for all the assets?

tdeekens: For future reference: https://github.com/commercetools/merchant-center-application-kit/issues/546

tdeekens:
An updated mc-be is rolled out across all envs this should be able to safely hit master.

PR #550, opened 2019-04-17
refactor: to pass intl message to Text components

emmenko:


playground/src/components/state-machines-list/state-machines-list.js
```diff
@@ -90,9 +88,12 @@ export class StateMachinesList extends React.Component {
                     <Spacings.Inline alignItems="center">
                       <DotIcon size="small" theme="green" />
                       <Text.Detail isItalic={true}>
```
tdeekens: Ha, for this we'd need `intlValues` ain't we?

emmenko: Yeah, I was going to open a PR to ui-kit 😅

tdeekens:
Merging as UIKit updated and we can have it in the release I suppose.

PR #552, opened 2019-04-17
chore: use menuPortalTarget in form-dialog visual route

montezume:
#### Summary

**Don't merge until ** #551 is merged.

We don't seem to have problems with scrolling 😄 

nice


tdeekens:
Lets do this in one release then.

tdeekens:
Done. I think this is good then?

montezume:
Screen Shot 2019-04-17 at 11 01 17 AM
Not right yet. I'm going to mess around with using a different target

montezume:
Okay, looks like we are having the exact same issue we had with trying to use VRT with tooltips. 

It would probably be worthwhile to make a reproduction repo using `react-select` and submit that with a support request to Percy...

PR #563, opened 2019-04-23
refactor(sentry): to also pass location to sentry environment, enable sentry based on tracking value

emmenko:
I'm reorganizing a bit our sentry projects, as we are rolling out more environments. For example, we can use more the new Sentry environment UI to differentiate between the different environments of a project (e.g. gcp prod eu/us).

### DIFF
--- a/packages/sentry/sentry.js
+++ b/packages/sentry/sentry.js
@@ -1,11 +1,11 @@
 import * as Sentry from '@sentry/browser';
 
 export const boot = () => {
-  if (window.app.env === 'production') {
+  if (window.app.trackingSentry) {
     Sentry.init({
       dsn: window.app.trackingSentry,
       release: window.app.revision,
-      environment: window.app.env,
+      environment: `${window.app.env}-${window.app.location}`,
       // in order to reduce the noise in sentry we only track errors that come
       // from our code and ignore errors that come from other services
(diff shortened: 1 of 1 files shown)


packages/sentry/sentry.js
```diff
@@ -1,11 +1,11 @@
 import * as Sentry from '@sentry/browser';
 
 export const boot = () => {
-  if (window.app.env === 'production') {
+  if (window.app.trackingSentry) {
     Sentry.init({
       dsn: window.app.trackingSentry,
       release: window.app.revision,
-      environment: window.app.env,
+      environment: `${window.app.env}-${window.app.location}`,
```
emmenko: This is to have more specific information about the environment within the same sentry project.

packages/sentry/sentry.js
```diff
@@ -1,11 +1,11 @@
 import * as Sentry from '@sentry/browser';
 
 export const boot = () => {
-  if (window.app.env === 'production') {
+  if (window.app.trackingSentry) {
```
emmenko: To keep it simple, we only need to check for the `trackingSentry` url value. For example, we might also include staging (previously it was only enabled for production).

tdeekens:
@amine-benselim took over FLD. Assigned him.

PR #567, opened 2019-04-23
fix(app-shell): back to project link colour

tdeekens:
#### Summary

This refactors the back to project link to be a normal LinkButton. Design requested it to be a FlatButton but actually intends the colour to not be black like now

CleanShot 2019-04-23 at 17 57 28@2x

but primary liks this

CleanShot 2019-04-23 at 17 58 48@2x

### DIFF
--- a/packages/application-shell/src/components/app-bar/app-bar.js
+++ b/packages/application-shell/src/components/app-bar/app-bar.js
@@ -1,7 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Link } from 'react-router-dom';
-import { Spacings, Text, AngleLeftIcon } from '@commercetools-frontend/ui-kit';
+import { FormattedMessage } from 'react-intl';
+import {
+  Spacings,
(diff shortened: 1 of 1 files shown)


packages/application-shell/src/components/app-bar/app-bar.js
```diff
@@ -12,14 +17,18 @@ import messages from './messages';
...
-      <AngleLeftIcon theme="green" size="small" />
-      <Text.Detail intlMessage={messages.backToProjectLink} />
-    </Spacings.Inline>
-  </Link>
+  <LinkButton
+    to={`/${props.projectKey}`}
+    iconLeft={<AngleLeftIcon />}
+    label={
+      <FormattedMessage {...messages.backToProjectLink}>
+        {txt => txt}
```
emmenko: 🤔

tdeekens: I don't want to injectIntl and this component needs a string.

PR #570, opened 2019-04-25
feat(website): implement home screen, colors, and main content layout

emmenko:
- implements home screen
- adjust main website layout
- define some basic colors palette
- add necessary plugins
- setup netlify


emmenko:
The site preview is deployed, in case you want to check it out

montezume:
Looks pretttttty nice.  🎉

website/src/colors.js
```diff
@@ -0,0 +1,11 @@
+// eslint-disable-next-line import/prefer-default-export
+export const light = {
```
emmenko: We don’t do theming now but having already a bit of a structure does not hurt.

Also, I’m still playing around with the colors choices and names, so it’s not final yet. I will go through them with the design team as well.

website/src/components/code-block.js
```diff
@@ -1,15 +1,135 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Highlight, { defaultProps } from 'prism-react-renderer';
+import { customProperties } from '@commercetools-frontend/ui-kit';
+import * as colors from '../colors';
+
+// TODO: improve colors
+const theme = {
```
emmenko: This is the “theme” config for the code syntax highlighter. It’s still a WIP at the moment, in terms of color choices.

website/src/components/markdown-layout.js
```diff
@@ -2,10 +2,67 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { MDXProvider } from '@mdx-js/react';
 import { css } from '@emotion/core';
+import styled from '@emotion/styled';
 import Layout from './layout';
 import SEO from './seo';
 
-const components = {};
+// TODO: style components
```
emmenko: Will work on that in a follow up PR

website/src/pages/404.js
```diff
@@ -1,13 +1,37 @@
...
+        justify-content: center;
+      `}
+    >
+      <h1>Page not found</h1>
+      <div
+        css={css`
+          max-width: 50%;
+        `}
+      >
+        <Lottie
```
emmenko: This is to render the animated SVG

website/src/pages/index.js
```diff
@@ -1,22 +1,222 @@
...
+
+const heroSlides = [
+  { id: 0, component: DevelopItSvg },
+  { id: 1, component: DeployItSvg },
+  { id: 2, component: RegisterItSvg },
+];
+
+const IndexPage = () => {
+  const [slideIndex, nextSlide] = React.useState(0);
+  const transitions = useTransition(slideIndex, i => i, {
```
emmenko: react-spring is pretty cool

celeste-horgan:
Beautiful as usual, Nicola!

But: if you're going to use the geometric code-like font as a header, I'd strongly suggest falling back to Open Sans as body text. Particularly given that the App Kit will have lots of codeblocks and `code highlights`, it's going to get unpleasant and difficult to read very quickly if you don't. A large part of the reason we use geometric fonts for code highlighting is to create a visual difference between code and body text, so that big documents are easier to scan.

On another note: Since you've started tackling this website, do we need to start a conversation about writing some of the overview/philosophy type text for this site? We spoke about it briefly a few months ago, but you weren't ready to work on the site so we stopped talking.

emmenko:
@celeste-horgan yes thanks for you feedback. At the moment I'm mostly focused on the basic structure, layout parts and colors. Once I'm done with that we can focus on the content part. There will be for sure other details to adjust and discuss, so I'll be happy to look at those things together with you.

PR #571, opened 2019-04-26
fix(mc-scripts): do not use ip address as socket host. Fixes #569

### LINKED ISSUE #569: CSP policy complains about webpack-dev-server hot reloading
Running the MC locally, you get a lot of messages like below, due to the webpack-dev-server's hot reloading breaking the CSP policy 
Screen Shot 2019-04-24 at 11 12 12 AM

emmenko:


emmenko:
Please test it out on your laptops to see if it really fixes the problem.

montezume:
Okay I'll try it locally with the building + copy paste into MC-FE 👍

emmenko:
Nice! I'll do a release later then

PR #574, opened 2019-04-27
refactor(components): to use emotion styles

emmenko:
This is necessary to render the components within a gatsby website. Furthermore, we wanted to use emotion in app-kit as well, so it's a win win.


packages/application-components/src/components/dialogs/form-dialog/form-dialog.visualroute.js
```diff
@@ -103,6 +103,8 @@ const FormDialogOverflowExample = props => (
               onChange={formikProps.handleChange}
               onBlur={formikProps.handleBlur}
               onFocus={formikProps.handleFocus}
+              menuPortalTarget={document.body}
```
montezume: just remove this test until we can figure out what's going on with Percy and positioning items with JavaScript 😄

packages/application-components/src/components/dialogs/form-dialog/form-dialog.visualspec.js
```diff
@@ -1,5 +1,12 @@
 import { percySnapshot } from '@percy/puppeteer';
 
+const waitForMs = (ms = 1000) =>
```
tdeekens: Nit:

```js
const waitForMs = (ms = 1000) =>
  new Promise(resolve => setTimeout(resolve, ms));
```

packages/application-components/src/components/dialogs/internals/dialog-container.js
```diff
@@ -23,33 +23,182 @@ const getDefaultParentSelector = () =>
...
+            [row3-end]
+            / minmax(${customProperties.spacingXl}, 1fr)
+            ${customProperties.constraintL} minmax(
+              ${customProperties.spacingXl},
+              1fr
+            );
+        `,
+      ];
+  }
+};
```
tdeekens: Not sure how others feel: maybe worth dumping it into another file. It's quite a chunk until you get to the component which is what I would expect when opening the file.

packages/application-components/src/components/dialogs/internals/dialog-container.js
```diff
@@ -23,33 +23,182 @@ const getDefaultParentSelector = () =>
...
+            overflow: hidden;
+          `}
+        >
+          <Card
+            // 1. For the min-height: https://stackoverflow.com/questions/28636832/firefox-overflow-y-not-working-with-nested-flexbox/28639686#28639686
+            // 2. For the scale size, we want the card to stretch to 100% height
+            // 3. For the actual "> div" container with the content, we need to use normal pointer events so that clicking on it does not close the dialog.
+            css={css`
+              min-height: 0;
+              ${props.size === 'scale' ? `height: 100%;` : ``}
```
tdeekens: Are the backricks around height needed?

packages/application-components/src/components/dialogs/form-dialog/form-dialog.visualroute.js
emmenko: Yeah, I guess we can remove it for now. I'll clean things up.

montezume: or you can do it in another PR 🤷‍♂

packages/application-components/src/components/dialogs/form-dialog/form-dialog.visualspec.js
montezume: we don't need this because that test doesn't work anyways. Just remove that from the visualroute and the "click" from the test.

packages/application-components/src/components/dialogs/internals/dialog-container.js
emmenko: I moved some of the styles into a separate file. Thanks for the suggestion

PR #585, opened 2019-04-29
docs(changelog): for v13.3.7 [skip ci]

tdeekens:


emmenko:
PS: do you want to test the canary release first maybe, in case there are other edge cases to fix?

tdeekens:
> PS: do you want to test the canary release first maybe, in case there are other edge cases to fix?

Yup. I just wanted to parepare it.

tdeekens:
Tested local builds and canary builds. Will release.

PR #586, opened 2019-04-29
chore(app-shell): extract logic into util

tdeekens:
#### Summary

This pull request extracts the project key determination logic into a little module. The whole ternary business is a bit spread out across the code base while essentially it's almost always the same.

1. Did the user select one (localStorage)
2. Does the user have a default one (mc-be)
3. Then either rendering something or not, redirect or not

### DIFF
--- a/packages/application-shell/src/components/app-bar/app-bar.js
+++ b/packages/application-shell/src/components/app-bar/app-bar.js
@@ -8,9 +8,9 @@ import {
   LinkButton,
 } from '@commercetools-frontend/ui-kit';
 import LogoSVG from '@commercetools-frontend/assets/images/logo.svg';
-import { selectProjectKeyFromLocalStorage } from '../../utils';
 import UserSettingsMenu from '../user-settings-menu';
 import ProjectSwitcher from '../project-switcher';
+import { getPreviousProjectKey } from '../../utils';
 import LoadingPlaceholder from '../loading-placeholder';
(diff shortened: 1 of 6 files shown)


packages/application-shell/src/components/app-bar/app-bar.js
```diff
@@ -30,18 +30,9 @@ export const BackToProjectLink = props => (
 BackToProjectLink.displayName = 'BackToProjectLink';
 
 BackToProjectLink.propTypes = {
-  projectKey: PropTypes.string.isRequired,
+  projectKey: PropTypes.string,
```
tdeekens: We also don't always have a `projectKey` here.

packages/application-shell/src/components/app-bar/app-bar.spec.js
```diff
@@ -5,8 +5,6 @@ import ProjectSwitcher from '../project-switcher';
 import UserSettingsMenu from '../user-settings-menu';
 import AppBar, { BackToProjectLink } from './app-bar';
 
-jest.mock('../../utils');
```
tdeekens: This stub is never actually used.

tdeekens:
> I wish that for such refactorings we would have proper types support, it would just make it dead easier to avoid mistakes.

Yup. It's also really hard to follow along with this code cause it's kind of accidentially complex.

PR #587, opened 2019-04-29
chore(app-shell): add development mode message

tdeekens:
#### Summary

Given the mc runs in devleopment mode and the user does not have any projects we shoudl show a message. Given this happens when the mc runs behind the proxy we will be redirected to `account/projects/new` anyway.


packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -283,8 +284,36 @@ export const RestrictedApplication = props => (
 
                                   if (!user) return <ApplicationLoader />;
 
-                                  // NOTE: Given no previous `projectKey` is found,
-                                  // we redirect to the base url.
+                                  if (
+                                    props.environment.servedByProxy !== true &&
+                                    !previousProjectKey
```
tdeekens: This is a bit defensive but it's just so we don't show the message by accident.

PR #590, opened 2019-04-30
refactor(app-shell): to have redirect-to-project-create component

tdeekens:
#### Summary

This pull request aims to improve how we redirect to project create while fixing a bug with it.

Generally three cases where we need to cover this exist (apart from multiple cache things):

1. User has previously visited a project _and_ a project key is in the url but his new account doesn't have access to it while she/he has no other projects
   - Redirect to project create
   - Given not served by proxy: show dev message
2. User never or has visited a project but no project key is in the url
   - Redirect to project create
   - Given not served by proxy: show dev message

Given that those cases overlap I extracted the logic into a component for reuse and isolated testing.

Moreover, we should contract (did already) UX/PO about how we may want to scope local stroage. Given you sign in and sign up across various accounts do we really want to attempt to restore the project previously visited.

### DIFF
--- a/packages/application-shell/src/components/application-shell/application-shell.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.js
@@ -286,9 +286,9 @@ export const RestrictedApplication = props => (
                                    * NOTE:
                                    *   Given the user has not been loaded a loading spinner is shown.
                                    *   Given the user was not working on a project previously nor has a default
-                                   *   project she/she will be promted to create one.
-                                   *   Given the user was working in a project previously or has a default
-                                   *   project a redirect to it happens.
+                                   *   project, the user will be prompted to create one.
+                                   *   Given the user was working on a project previously or has a default
+                                   *   project, the application will redirect to that project.
                                    */
 
                                   if (!user) return <ApplicationLoader />;
--- a/packages/application-shell/src/components/application-shell/application-shell.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.js
@@ -2,7 +2,6 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Redirect, Route, Switch } from 'react-router-dom';
 import { defaultMemoize } from 'reselect';
-import { Spacings, Text, LinkButton } from '@commercetools-frontend/ui-kit';
 import {
   joinPaths,
   trimLeadingAndTrailingSlashes,
@@ -32,6 +31,7 @@ import NavBar, { LoadingNavBar } from '../navbar';
 import ApplicationLoader from '../application-loader';
 import ErrorApologizer from '../error-apologizer';
 import Redirector from '../redirector';
+import RedirectToProjectCreate from '../redirect-to-project-create';
 import { selectProjectKeyFromUrl, getPreviousProjectKey } from '../../utils';
 import styles from './application-shell.mod.css';
 import QuickAccess from '../quick-access';
@@ -282,42 +282,20 @@ export const RestrictedApplication = props => (
                                     user && user.defaultProjectKey
                                   );
 
+                                  /**
+                                   * NOTE:
+                                   *   Given the user has not been loaded a loading spinner is shown.
+                                   *   Given the user was not working on a project previously nor has a default
+                                   *   project she/she will be promted to create one.
+                                   *   Given the user was working in a project previously or has a default
+                                   *   project a redirect to it happens.
+                                   */
+
                                   if (!user) return <ApplicationLoader />;
-
-                                  if (
-                                    props.environment.servedByProxy !== true &&
-                                    !previousProjectKey
-                                  )
-                                    return (
-                                      <Spacings.Stack scale="m">
-                                        <Text.Headline elementType="h2">
-                                          Development mode without projects
-                                        </Text.Headline>
-                                        <Text.Body>
-                                          You are using the Merchant Center in
-                                          development mode - it is not served by
-                                          a proxy. Moreover, you do not have any
-                                          projects yet. As a result we did not
-                                          redirect you anywhere (e.g. another
-                                          application).
-                                        </Text.Body>
-                                        <Text.Body>
-                                          Please go to the
-                                          `application-accounts` to create a
-                                          project first.
-                                        </Text.Body>
-                                        <LinkButton
-                                          to={`account/projects/new`}
-                                          label="Create project"
-                                        />
-                                      </Spacings.Stack>
-                                    );
-
-                                  // NOTE: Given no previous `projectKey` is found we redirect to the base url.
+                                  if (!previousProjectKey)
+                                    return <RedirectToProjectCreate />;
                                   return (
-                                    <Redirect
-                                      to={`/${previousProjectKey || ''}`}
-                                    />
+                                    <Redirect to={`/${previousProjectKey}`} />
                                   );
                                 }}
                               />
--- a/packages/application-shell/src/components/project-container/project-container.js
+++ b/packages/application-shell/src/components/project-container/project-container.js
@@ -5,7 +5,6 @@ import { Route, Switch, Redirect } from 'react-router-dom';
 import { injectIntl } from 'react-intl';
 import isNil from 'lodash/isNil';
 import flowRight from 'lodash/flowRight';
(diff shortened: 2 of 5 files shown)


tdeekens:
Sorry for the back and forth but it's really hard to expect all cases in dev and non-dev mode while always finding a "dead" spot in app-shell one missed.

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -282,42 +282,20 @@ export const RestrictedApplication = props => (
...
-                                          Please go to the
-                                          `application-accounts` to create a
-                                          project first.
-                                        </Text.Body>
-                                        <LinkButton
-                                          to={`account/projects/new`}
-                                          label="Create project"
-                                        />
-                                      </Spacings.Stack>
-                                    );
```
tdeekens: This moved into the component and was combined with the other inline component.

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -282,42 +282,20 @@ export const RestrictedApplication = props => (
...
 
-                                  // NOTE: Given no previous `projectKey` is found we redirect to the base url.
+                                  if (!user) return <ApplicationLoader />;
+                                  if (!previousProjectKey)
+                                    return <RedirectToProjectCreate />;
                                   return (
-                                    <Redirect
-                                      to={`/${previousProjectKey || ''}`}
-                                    />
+                                    <Redirect to={`/${previousProjectKey}`} />
```
tdeekens: Note: I shall remove the stringification.

packages/application-shell/src/components/project-container/project-container.js
```diff
@@ -165,12 +140,7 @@ export class ProjectContainer extends React.Component {
       return (
         <Switch>
           <Route path="/account" render={this.props.render} />
-          {this.props.environment.servedByProxy && (
-            <Route component={RedirectToProjectCreate} />
-          )}
-          {!this.props.environment.servedByProxy && (
-            <Route component={NoProjects} />
-          )}
```
tdeekens: Logic is now in the shared component.

packages/application-shell/src/components/redirect-to-project-create/redirect-to-project-create.js
```diff
@@ -0,0 +1,42 @@
...
+
+export class RedirectToProjectCreate extends React.Component {
+  static displayName = 'RedirectToProjectCreate';
+  static propTypes = {
+    servedByProxy: PropTypes.bool.isRequired,
+  };
+  constructor(props) {
+    super(props);
+
+    if (this.props.servedByProxy === true)
```
tdeekens: I went with a constructor to avoid to use the deprecated `componentWillMount` will also avoiding a potential flash of something rendering when using `componentDidMount`. Open to suggestions (e.g. double check in `render`).

packages/application-shell/src/components/redirect-to-project-create/redirect-to-project-create.js
```diff
@@ -0,0 +1,42 @@
...
+        </Text.Body>
+        <LinkButton to={`account/projects/new`} label="Create project" />
+      </Spacings.Stack>
+    );
+  }
+}
+
+export default withApplicationContext(applicationContext => ({
+  servedByProxy: applicationContext.environment.servedByProxy,
+}))(RedirectToProjectCreate);
```
tdeekens: Using the HoC again (not passing it in) so that the component can easily be used in `<Route component={} />`

packages/application-shell/src/components/redirect-to-project-create/redirect-to-project-create.spec.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { renderApp } from '../../test-utils';
```
tdeekens: Using `renderApp` as we need the router to be setup for the dev mode `LinkButton`.

tdeekens:
Alright, I tested all known cases locally with a build. 

1. Behind and infront of a proxy
2. With `projectKey` in url and without
3. with cached `projectKey` and without

I can't find any dangling issues now. Moreover, moving the redirect to `render` actually avoids a flashing content (with an error).

packages/application-shell/src/components/project-container/project-container.js
```diff
@@ -165,12 +140,7 @@ export class ProjectContainer extends React.Component {
...
+          <Route component={RedirectToProjectCreate} />
```
tdeekens: This now internally determines what do to.

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -282,42 +282,20 @@ export const RestrictedApplication = props => (
...
-                                          to={`account/projects/new`}
-                                          label="Create project"
-                                        />
-                                      </Spacings.Stack>
-                                    );
+                                  /**
+                                   * NOTE:
+                                   *   Given the user has not been loaded a loading spinner is shown.
+                                   *   Given the user was not working on a project previously nor has a default
+                                   *   project she/she will be promted to create one.
```
emmenko:
```suggestion
                                   *   project, the user will be prompted to create one.
```

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -282,42 +282,20 @@ export const RestrictedApplication = props => (
...
+                                   *   Given the user was working in a project previously or has a default
```
emmenko:
```suggestion
                                   *   Given the user was working on a project previously or has a default
```

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -282,42 +282,20 @@ export const RestrictedApplication = props => (
...
+                                   *   project a redirect to it happens.
```
emmenko:
```suggestion
                                   *   project, the application will redirect to that project.
```

packages/application-shell/src/components/redirect-to-project-create/redirect-to-project-create.js
emmenko: `useApplicationContext`?

packages/application-shell/src/components/redirect-to-project-create/redirect-to-project-create.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { withApplicationContext } from '@commercetools-frontend/application-shell-connectors';
+import { Spacings, Text, LinkButton } from '@commercetools-frontend/ui-kit';
+
+export const RedirectToProjectCreate = props => {
+  if (props.servedByProxy === true)
+    window.location.replace('/account/projects/new');
```
emmenko: Is it "safe" to do this during rendering? 🤔
I mean, the page reloads so who cares (or maybe sentry cares), but in theory we should do that using a side effect, correct?

packages/application-shell/src/components/application-shell/application-shell.js
tdeekens: @emmenko the bug was given you're behing the proxy and don't have a `projectKey` in the url (which is when we end up here) then we redirected to `/` instead of `account/projects/new`. Now we redirect there within the new component too (same as with the url containing a `projectKey`).

packages/application-shell/src/components/redirect-to-project-create/redirect-to-project-create.js
tdeekens: No we can't and shouldn't. This will run with a React version of a consumer in a custom app. We can't assume hooks are available I'd say. Plus the syntactic benefit is marginal to me.

packages/application-shell/src/components/redirect-to-project-create/redirect-to-project-create.js
tdeekens: If we do it as a side-effect it will happen after the component has been mounted. Hence you will see the message then be redirected. If we do it in render if occurs faster. I also don't like side-effects in render but in this case we just want to not even work with anything rendered anymore.

PR #591, opened 2019-04-30
docs(changelog): for v13.3.8 [skip ci]

tdeekens:
Waiting to test out canary release first. Then releasing.


tdeekens:
Tested both canary releases. This seems to be everything I ever wanted.

PR #592, opened 2019-04-30
chore(app-shell): to not fetch project without key

tdeekens:
#### Summary

This avoids a uncalled-for round trip and makes sign up smoother. Given the user doesn't have a project:

1. Either not in the url (claims she/he has one)
2. Or actually no `defaultProjectKey` from the users query which is `projects[0]`

Then we can omit the request. This makes the redirect smoother and less janky.

### DIFF
--- a/packages/application-shell/src/components/fetch-project/fetch-project.js
+++ b/packages/application-shell/src/components/fetch-project/fetch-project.js
@@ -35,8 +35,12 @@ export const mapAllAppliedToObjectShape = allAppliedShape =>
 class FetchProject extends React.Component {
   static displayName = 'FetchProject';
   static propTypes = {
-    projectKey: PropTypes.string.isRequired,
+    projectKey: PropTypes.string,
+    skip: PropTypes.bool,
     children: PropTypes.func.isRequired,
+  };
+  static defaultProps = {
+    skip: false,
   };
   shouldComponentUpdate(nextProps) {
     return !deepEqual(this.props, nextProps);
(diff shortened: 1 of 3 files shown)


packages/application-shell/src/components/fetch-project/fetch-project.js
```diff
@@ -35,9 +35,13 @@ export const mapAllAppliedToObjectShape = allAppliedShape =>
 class FetchProject extends React.Component {
   static displayName = 'FetchProject';
   static propTypes = {
-    projectKey: PropTypes.string.isRequired,
+    projectKey: PropTypes.string,
```
tdeekens: This has to be optional after all now. The url doesn't have to have one. Never had to.

packages/application-shell/src/components/redirect-to-project-create/redirect-to-project-create.js
```diff
@@ -14,9 +14,12 @@ export const RedirectToProjectCreate = props => {
    *   In turn this intends wo make explicit that we never want to
    *   render and instead just navigate away.
    */
-  if (props.servedByProxy === true)
+  if (props.servedByProxy === true) {
     window.location.replace('/account/projects/new');
 
+    return null;
```
tdeekens: Just to be more explicit even.

tdeekens:
More concretely this avoids a "flash of page notification error" we currently also have in the MC. The redirect happens fast, the project query is cancelled (by browser) and an error appears for 50ms. We can avoid that by optimizing as we know a user is "lying" if having a `projectKey` in the url but actually not even having a default project.

PR #594, opened 2019-05-02
fix(app-shell): back to project peforming full page reload

tdeekens:
#### Summary

This pull request changes the BackToProject button to perform a full page reload. A code comment explains why.


packages/application-shell/src/components/back-to-project/back-to-project.spec.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
```
tdeekens: Wanted to have a spec so it can't be changed without knowing/being informed of it.

PR #595, opened 2019-05-02
fix(website): to differentiate between internal and external links

emmenko:
Inspired by https://www.gatsbyjs.org/docs/gatsby-link/#reminder-use-link-only-for-internal-links

### DIFF
--- a/website/src/templates/markdown.js
+++ b/website/src/templates/markdown.js
@@ -1,12 +1,13 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { graphql } from 'gatsby';
+import { graphql, Link as HistoryLink } from 'gatsby';
 import { MDXRenderer } from 'gatsby-mdx';
 import { MDXProvider } from '@mdx-js/react';
 import { css } from '@emotion/core';
 import styled from '@emotion/styled';
 import { customProperties } from '@commercetools-frontend/ui-kit';
 import * as colors from '../colors';
+import ExternalLinkSvg from '../images/external-link.svg';
 import { LayoutContent } from '../layouts';
 import { SEO, CodeBlock } from '../components';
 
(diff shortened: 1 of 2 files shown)


website/src/templates/markdown.js
```diff
@@ -216,18 +216,25 @@ const Link = props => {
...
+    // E.g. if `href="./foo"` and we're on page `/getting-started/bar`, the
+    // value we need is `/getting-started/foo`.
+    // To achieve that, we let the DOM API build the full URL, then we simply
+    // extract the relative path.
+    const linkElement = document.createElement('a');
+    linkElement.href = props.href;
+    const absoluteUrl = linkElement.href; // <-- this now is the full absolute URL
+    const [, relativePath] = absoluteUrl.split(window.location.origin);
+    setLink(relativePath);
+  });
```
tdeekens: Don't have have to specify `[props.href]` as a dep? Should we add the ESLint rules for hooks on the pkg?

emmenko: The prop won't change, so it shouldn't be necessary. Or am I missing something?

tdeekens: In that case it's fine. It's just underspecifying the dependencies which I read is rarely what one wants and can lead to quite silent bugs.

emmenko: Yeah I know. I can add an empty array if you want, to make it "safer"?

tdeekens: I think that's the same. Anyway. It's fine just wanted to make sure.

website/src/templates/markdown.js
```diff
@@ -192,22 +191,59 @@ const Link = styled.a`
...
+        css={css`
+          display: inline-flex;
+          align-items: center;
+        `}
+      >
+        <a
+          {...props}
+          css={linkStyles}
+          target="_blank"
+          rel="noopener noreferrer"
```
montezume: 👍

PR #597, opened 2019-05-02
feat(website): add table of content sidebar

emmenko:

### DIFF
--- a/website/src/layouts/content.js
+++ b/website/src/layouts/content.js
@@ -4,6 +4,7 @@ import { css } from '@emotion/core';
 import LayoutContainer from './internals/layout-container';
 import LayoutHeader from './internals/layout-header';
 import LayoutSidebar from './internals/layout-sidebar';
+import LayoutToc from './internals/layout-toc';
 import LayoutFooter from './internals/layout-footer';
 import LayoutMain from './internals/layout-main';
 import Globals from './internals/globals';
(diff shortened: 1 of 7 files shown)


website/src/layouts/content.js
```diff
@@ -4,6 +4,7 @@ import { css } from '@emotion/core';
 import LayoutContainer from './internals/layout-container';
 import LayoutHeader from './internals/layout-header';
 import LayoutSidebar from './internals/layout-sidebar';
+import LayoutToc from './internals/layout-toc';
```
montezume: nitpick: LayoutTOC

website/src/layouts/internals/layout-toc.js
```diff
@@ -0,0 +1,125 @@
...
+  color: ${colors.light.text};
+  text-decoration: none;
+
+  :hover {
+    color: ${colors.light.primarySoft};
+    text-decoration: underline;
+  }
+  :hover,
+  :active {
+    outline-width: 0;
```
montezume: 😢

emmenko: ?

montezume: Why are we removing the outline? makes it really hard to use keyboard to navigate

emmenko: The focus outline for a11y is still there, this is just for hover/active state.

PR #599, opened 2019-05-03
feat(app-shell): add "Manage Projects" to Quick Access

tdeekens:
#### Summary

This pull request adds "Manage Projects" to Quick Access.

CleanShot 2019-05-03 at 16 07 33


tdeekens:
Note: the error only appears cause the route doesn't exist in the playground.

PR #602, opened 2019-05-03
refactor(app-shell): remove old feature flags

tdeekens:
#### Summary

I assumed flags are passed from the app but quick-access connects to them itself. Another case for types casue the prop-types are not followed through to the `create-commands`-fn.


packages/application-shell/src/components/quick-access/quick-access.js
```diff
@@ -356,15 +353,12 @@ export default flowRight(
   injectIntl,
   injectFeatureToggles([
     'pimSearch',
-    'customerGroups',
-    'projectSettings',
-    'developerSettings',
-    'projectSettingsChannels',
```
tdeekens: I went ahead and did a little clean up. These feature won't be untoggled and we should clean up our used flags in the MC then too.

tdeekens:
> it looks to me that I have to update this file as well, since I introduced 2 new feature flags

No. Only if those features are ought to be accessible through Quick Access and their access is toggled.

adnasa:
yeah I just realised that now. carry on 😗

packages/application-shell/src/components/quick-access/create-commands.js
```diff
@@ -312,16 +310,15 @@ export default ({
...
-                to: `/${project.key}/settings/developer/api-clients`,
-              },
+          hasPermission(permissions.ManageProject, project.permissions) && {
+            id: 'go/settings/developer',
+            text: intl.formatMessage(messages.openDeveloperSettings),
+            action: {
+              type: 'go',
+              to: `/${project.key}/settings/developer/api-clients`,
             },
-          featureToggles.customApplicationsSettings &&
```
tdeekens: This was actually wrong. There is no flag like this and the flag requested is called differently. Hence, quick access for custom app settings was never working to how I see it.

PR #603, opened 2019-05-03
docs(changelog): for v13.4.1 [skip ci]

tdeekens:
Waiting for canary then releasing again.


tdeekens:
Yay now even custo apps quick access works.

PR #604, opened 2019-05-03
fix(website): scrolling momentum on ios

emmenko:


emmenko:
I just tried on the ios simulator and it fixes the problem 😉

Issue #589, opened 2019-04-30
Static types


emmenko:
**Is your feature request related to a problem? Please describe.**
Some of the packages (e.g. AppShell) have some pretty complex logic that often times causes small bugs to appear as it gets really tricky on finding out all the edge cases. This problem becomes more evident during refactorings.

**Describe the solution you'd like**
We should start adding static types to the application logic, with the goal of giving more confidence during refactorings and to also provide more helpful insights about the logic for new contributors.

**Additional context**
We can start with Flow, as it's more easy to integrate and we already use it in our backend services.

---

## Migration progress:

- [x] actions-global @emmenko #888 
- [x] application-components @emmenko  #921 
- [ ] application-shell (**this should be the last one to migrate**) @emmenko #1069 
- [x] application-shell-connectors @emmenko
- [x] assets (_not necessary_)
- [x] babel-preset-mc-app (_not necessary_)
- [x] browser-history @emmenko #885 
- [x] constants @emmenko
- [ ] create-mc-app
- [x] eslint-config-mc-app (_not necessary_)
- [x] i18n @emmenko #774 
- [x] jest-preset-mc-app (_not necessary_)
- [x] l10n @montezume #718 
- [ ] mc-dev-authentication
- [ ] mc-html-template
- [ ] mc-http-server
- [ ] mc-scripts
- [x] notifications @emmenko #884 
- [x] permissions @tdeekens #739 
- [x] react-notifications @emmenko #896 
- [x] sdk @emmenko #985
- [x] sentry @emmenko 
- [x] storage @montezume #715 
- [x] url-utils @montezume #717

emmenko:
We decided to use TypeScript. From now on we will start migrating the packages to TS, step by step.

tdeekens:
Migration status

- [x] sentry
- [x] storage
- [x] permissions
- [x] application-shell-connectors
- [x] url-utils
- [ ] actions-global
- [ ] application-shell
- [ ] sdk
- [ ] notifications

Please mark if you're working on a package with ⚠️ or add it :)

montezume:
I just want to give some feedback from my side on the migration and my thoughts.

I did a few small simple packages, but when I got to more complex things, I realized that my knowledge of typescript is just not up to snuff. I don't have enough experience with it, or enough confidence with it, to migrate more packages. 

I also do not have time to both do my normal work, and learn TypeScript at the office. So something has to give, and for me, what gave was TypeScript. 

[...]

PR #605, opened 2019-05-05
Setup TypeScript

emmenko:
> This is a spike to get a feeling of the effort and setup that we would need to run typescript (see #589).

I decided to convert the `sentry` package, as it's quite small and it can be used as a good estimation.

The changes and setup is partly taken from flopflip monorepo and partly from the internet.

I'm not 100% certain that this is the correct setup but this is what I got so far that works. I must say that I struggled a lot to get a basic setup going and I couldn't have done it without looking things up from other repos.

Feedback so far: given that it was 10% type system work and 90% setup, I'm not bought on typescript yet. I even had to type myself a couple of modules because they didn't have declaration files, which is not much different from e.g. flowtype.


.eslintrc.yaml
```diff
@@ -1,11 +1,25 @@
-extends: '@commercetools-frontend/eslint-config-mc-app'
+extends:
```
emmenko: The eslint config is one of the messy things, probably due the unfortunate combination of our own config, the XO config and typescript.

emmenko:
@tdeekens when you have some time, can you have a look at the typescript setup and see how we can improve it? Especially around the eslint stuff.

Thanks 🙏

.eslintrc.yaml
```diff
@@ -1,11 +1,25 @@
...
+  - xo
+  - '@commercetools-frontend/eslint-config-mc-app'
+  - xo-react
+  - xo-typescript
+  - prettier/@typescript-eslint
```
tdeekens: Maybe worth adding `prettier/react`? Not sure.

package.json
```diff
@@ -103,7 +120,11 @@
     "start-server-and-test": "1.9.0",
     "stylelint": "9.10.1",
     "stylelint-config-standard": "17.0.0",
-    "stylelint-order": "0.8.1"
+    "stylelint-order": "0.8.1",
+    "typescript": "3.4.5"
+  },
+  "resolutions": {
+    "@types/react": "16.8.16"
```
tdeekens: Yeah. I couldn't update to a recent version either and I know good ol Phil has the same problem.

tsconfig.json
```diff
@@ -0,0 +1,33 @@
...
+    "emitDecoratorMetadata": false,
+    "experimentalDecorators": false,
+    "esModuleInterop": true,
+    "importHelpers": true,
+    "jsx": "react",
+    "lib": ["esnext", "dom"],
+    "module": "esnext",
+    "moduleResolution": "node",
+    "noFallthroughCasesInSwitch": true,
+    "noImplicitAny": true,
```
tdeekens: I think we have to be strict in the beginning otherwise we never to it.

tdeekens:
@emmenko do you have a preferance between the TS or Flow setup at the moment?

emmenko:
> do you have a preferance between the TS or Flow setup at the moment?

I like the setup of Flow, as it's simple and easy to start. However, we might get better type support in TS. I also have a feeling that the TS compiler is a bit faster on full rebuilds.

If we can simplify the TS setup, I think I would prefer to go with that.

.eslintrc.yaml
emmenko: I think this part here is what we need to fix, as we are extending and overwriting a lot of presets. For instance, that's why our own preset is in between the others, otherwise we get more conflicts.

I think we need to either include the typescript stuff in our preset (so that we can determine the proper order of the presets) or have a separate one tailored to TS.

WDYT?

tdeekens:
> If we can simplify the TS setup, I think I would prefer to go with that.

Maybe we could dump it into a `mc-typescript` package and reuse it from there. A bit like [ts-dx](https://github.com/palmerhq/tsdx). Maybe also after we figured out the setup once. No strong opinions.

> I like the setup of Flow, as it's simple and easy to start. However, we might get better type support in TS. I also have a feeling that the TS compiler is a bit faster on full rebuilds.

Same experience. Plus TypeScript seemed better at following through on types at times. However Flows error messages are nicer. I think generally the ecosystem right now for TS is the big win. Again, I like not buying into TS as a compiler instead uses Babel.

.eslintrc.yaml
tdeekens: We should then mix/make our own preset as a package `mc-preset-typescript` or something are carefully build up the preset there from other building blocks?

emmenko: Yeah, I don't have a good idea at the moment. I was hoping you do 😇

emmenko:
> Maybe we could dump it into a mc-typescript package and reuse it from there. A bit like ts-dx. Maybe also after we figured out the setup once. No strong opinions.

Ah nice the `ts-dx`. What if we just use that for developing app-kit? 🤔

tdeekens:
> Ah nice the ts-dx. What if we just use that for developing app-kit? 🤔

I don't have anything against it as long as it does what we want it to. However, it does not support custom babel configurations so I am not sure how easily we can adopt it.

emmenko:
Now that the babel typescript parser is faster (https://twitter.com/NicoloRibaudo/status/1130897987692056577) we might look into this setup again and see if we can make it work

emmenko:
Update: I managed somehow to fix the eslint setup. Should we invest in this and start migrating the packages to typescript?

tdeekens:
Can we maybe have a chapter meeting about it next week or something like that with UIKit etc. Feels a bit like all teams can give input in what direction they would want to move in regards to typing code bases?

emmenko:
Sure thing. Should I set up a meeting?

tdeekens:
Yeah maybe drop a line in the channel mentioning the ambassadors to find a day which suits most (maybe even via a poll or just checking calendars).

packages/sentry/src/sentry.ts
```diff
@@ -17,7 +22,7 @@ export const boot = () => {
   }
 };
 
-export const updateUser = user => {
+export const updateUser = (user: User) => {
   if (user && window.app.trackingSentry) {
```
islam3zzat: Nitpick: we don't need to check for user anymore
```suggestion
   if (window.app.trackingSentry) {
```

packages/sentry/src/sentry.ts
```diff
@@ -32,13 +37,18 @@ export const updateUser = user => {
 };
 
 export const stopTrackingUser = () => {
```
montezume: shouldn't this be 

```
export const stopTrackingUser = ():void => {

```

packages/sentry/src/sentry.ts
```diff
@@ -32,13 +37,18 @@ export const updateUser = user => {
...
-  if (window.app.trackingSentry)
+  if (window.app.trackingSentry) {
     Sentry.configureScope(scope => {
       scope.clear();
     });
+  }
 };
 
-export const reportErrorToSentry = (error, extraInfo, getIsEnabled) => {
+export const reportErrorToSentry = (
```
montezume: shouldn't this have a return type as well?

emmenko: Most of the time, type are inferred, so we don't need to explicitly annotate them.

packages/sentry/src/sentry.ts
emmenko: Yep, you're right!

packages/sentry/src/sentry.ts
montezume: Okay as long as that's the best practice. We should be careful though with our "first typescript package", as we will use it as inspiration for the rest of the typescript we write going forward...

emmenko: Sure...you will see it yourself though when you write code that if something needs to be annotated, you'll get a compiler warning/error. If not, you just leave the annotation out.

emmenko:
Nice feedback

image

tdeekens:
> Nice feedback

Yup and I happily ignored that in se flopper :D

emmenko:
Here we go! 🚀

PR #613, opened 2019-05-06
[spike] Setup Flowtype

emmenko:
> This is a spike to get a feeling of the effort and setup that we would need to run flowtype (see #589).

I decided to convert the `sentry` package, as it's quite small and it can be used as a good estimation.

Compared to the typescript setup, this feels much more lighter and simpler. I think I would prefer to go with flow for now.


rollup.config.js
```diff
@@ -34,6 +34,7 @@ const createConfig = cliArgs => {
       sourcemap: true,
     },
     plugins: [
+      patchedFlowEntry({ mode: 'strict' }),
```
emmenko: The idea here is to include the flow types into the npm package.

image

With `flow-copy-source`, we include the source files as `*.js.flow` files into the `dist/types` folder. The rollup plugin, simply creates a `*.js.flow` version of the bundle file that points to the `types` folder.
In theory, consumers of the package will be able to get the type annotations as flow understands the `*.js.flow` format and prefers to load that if it finds one.

emmenko:
Closing in favor of #605

PR #617, opened 2019-05-08
refactor(application-shell): use ui kit select input for locale switcher

montezume:
Experiment: Use `ui-kit` for the locale switcher instead of using react-select v1.

Before

localeswitchbefore

After

afterlocaleswicher


montezume:
@emmenko should we get design to do a UX review tomorrow?

emmenko:
@montezume I guess you can inform them, but I don't see a need for a review since we're using a validated uikit component.

PR #619, opened 2019-05-09
fix(application-components): ui-kit peer dep

montezume:
#### Summary

Our peer dependency for `ui-kit` is wrong, because in `application-components`, we use customProperties that were added in version `9.2.0`. [ref](https://github.com/commercetools/merchant-center-application-kit/blob/877818249aa324bbe6af447aed2517fe4183d579/packages/application-components/src/components/dialogs/internals/dialog-container.js#L92)

#### Approach

Updates the peer dep in application components to `^9.2`.

#### Open questions

Do we need to also update it in `application-shell`? Or will application-shell resolve it from `application-components`?

### DIFF
--- a/packages/application-components/package.json
+++ b/packages/application-components/package.json
@@ -65,7 +65,7 @@
     "storybook-readme": "5.0.3"
   },
   "peerDependencies": {
-    "@commercetools-frontend/ui-kit": "9.x",
+    "@commercetools-frontend/ui-kit": "> 9.2",
     "react": "16.x",
     "react-dom": "16.x",
     "react-intl": "2.x",
--- a/packages/application-components/package.json
+++ b/packages/application-components/package.json
@@ -65,7 +65,7 @@
     "storybook-readme": "5.0.3"
   },
   "peerDependencies": {
-    "@commercetools-frontend/ui-kit": "> 9.2",
+    "@commercetools-frontend/ui-kit": "^9.2",
     "react": "16.x",
     "react-dom": "16.x",
     "react-intl": "2.x",


tdeekens:
> Do we need to also update it in application-shell? Or will application-shell resolve it from application-components?

app-shell doesn't use any uikit so shouldn't have it as a peer dep from my view.

montezume:
@tdeekens not yet :trollface: #618 #617

tdeekens:
Then it should declare the version it needs to the "minimum" last major. So `9.x`. Yarn/npm will then resolve it to the highest minor of that major.

packages/application-components/package.json
```diff
@@ -65,7 +65,7 @@
     "storybook-readme": "5.0.3"
   },
   "peerDependencies": {
-    "@commercetools-frontend/ui-kit": "9.x",
+    "@commercetools-frontend/ui-kit": "^9.2",
```
emmenko: I don't think this will help much to be honest. Usually you only get warnings when you install an unmet version of a peer dependency, which you are probably going to ignore because your console is full of them.

I would rather just mention in the appkit changelog release that version x should require a ui-kit version >= x.

tdeekens: ...but warnings are better than nothing? Maybe we can just do both? Given then this would even be a breaking change in strict semver.

emmenko: Sure, we can do both I guess. Good point!

montezume: So should I merge this then 🤔 ?

emmenko:
Didn't you forget to include this in the release? 🤔

montezume:
@emmenko yes 😓

PR #621, opened 2019-05-09
chore: remove react-select v1

### LINKED ISSUE #554: react-select v1 css should be applied by default
In applications that don't use react-select v1, it's weird to have to apply the css themselves. It would make more sense if it's done as part of the app kit.

I just refactored an app not to use react-select v1, and I tried to remove it, but broke the app as app kit is expecting it to be applied.

montezume:
#### Summary

Now that #617 and #618 are merged, `react-select v1` can be removed as a dependency, and it's CSS import can be removed as well.

Closes #554

### DIFF
--- a/packages/application-shell/package.json
+++ b/packages/application-shell/package.json
@@ -82,7 +82,6 @@
     "perfume.js": "1.1.4",
     "prop-types": "15.7.2",
     "qss": "2.0.3",
-    "react-modal": "3.8.1",
     "react-required-if": "1.0.3",
     "redux-logger": "3.0.6",
     "redux-thunk": "2.3.0",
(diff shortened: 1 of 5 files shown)


yarn.lock
```diff
@@ -19911,15 +19911,6 @@ react-router@5.0.0:
     tiny-invariant "^1.0.2"
     tiny-warning "^1.0.0"
 
-react-select@1.2.1:
```
montezume: this makes me happy

packages/application-shell/package.json
```diff
@@ -82,9 +82,7 @@
     "perfume.js": "1.1.4",
     "prop-types": "15.7.2",
     "qss": "2.0.3",
-    "react-modal": "3.8.1",
```
montezume: this is now in `application-components`, I think it was just a forgotten leftover

montezume:
Thanks all! Feels good to have this done. 😅

PR #624, opened 2019-05-09
feat(permissions): add use-is-authorized hook

tdeekens:
#### Summary

This pull request adds a `useIsAuthorized()` hook. It should be pretty self explanatory from the specs.

### DIFF
--- a/packages/application-shell-connectors/src/components/application-context/index.js
+++ b/packages/application-shell-connectors/src/components/application-context/index.js
@@ -1,4 +1,5 @@
 export {
+  Context,
   ApplicationContext,
   ApplicationContextProvider,
   withApplicationContext,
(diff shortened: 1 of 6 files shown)


packages/application-shell-connectors/src/components/application-context/index.js
```diff
@@ -1,4 +1,5 @@
 export {
+  Context,
```
tdeekens: The React docs state a useContext call needs the return of `createContext`. Not only the consumer.

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.js
```diff
@@ -0,0 +1,28 @@
...
+
+      const isAuthorized = shouldMatchSomePermissions
+        ? hasSomePermissions(demandedPermissions, actualPermissions)
+        : hasEveryPermissions(demandedPermissions, actualPermissions);
+
+      return isAuthorized;
+    }
+  : () => {
+      throw new Error(
+        `React hooks do not seem to be available yet in the installed React version "${
```
jonnybel: 🆒

tdeekens: We do that on other hooks too. I didn't care to test it again. In another hook there is a teset.

tdeekens:
Feel free to merge when deemed good.

PR #636, opened 2019-05-15
feat(components): add ModalPage component

jonnybel:
#### Summary

Moving the ModalPage component from the MC app to here. Lots of changes will be done to the component, but this is the first part of the job -- to have storybook and visual regression tests first and then build on top of it.

TODO:
- [x] Storybook
- [x] VRTs

### DIFF
--- a/packages/application-components/src/components/modal-page/README.md
+++ b/packages/application-components/src/components/modal-page/README.md
@@ -3,7 +3,7 @@
 ## Usage
 
 ```js
-import ModalPage from '@commercetools-local/components/modal-page';
+import { ModalPage } from '@commercetools-frontend/application-components';
 ```
 
 #### Description
--- a/packages/application-components/src/components/modal-page/README.md
+++ b/packages/application-components/src/components/modal-page/README.md
@@ -1,0 +1,54 @@
+# ModalPage
+
+## Usage
+
+```js
+import ModalPage from '@commercetools-local/components/modal-page';
+```
+
+#### Description
+
+Modal pages are controlled components used to render a page using a modal container, which makes it appear on top of the normal page. This is useful to render detailed data that normally requires some space.
+
+#### Usage
+
+```js
+<ModalPage
+  isOpen={true}
+  onClose={handleClose}
+  title="Lorem ipsus"
+  subtitle={<Text.Body>{'Lorem ipsus ...'}</Text.Body>}
+  components={{
+    actions: (
+      <Spacings.Inline alignItems="flex-start">
+        <SecondaryButton label="Cancel" onClick={handleReset} />
+        <PrimaryButton
+          label="Save"
+          isDisabled={isDisabled}
+          onClick={handleSubmit}
+        />
+      </Spacings.Inline>
+    ),
+  }}
+>
+  <Spacings.Stack scale="m">
+    <Text.Body>{'Lorem ipsus ...'}</Text.Body>
+    <Text.Body>{'Lorem ipsus ...'}</Text.Body>
+  </Spacings.Stack>
+</ModalPage>
+```
+
+#### Properties
+
+| Props                | Type       | Required | Values                | Default | Description                                                                                                                                                                                                                                                                                       |
+| -------------------- | ---------- | :------: | --------------------- | ------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `isOpen`             | `boolean`  |    ✅    | -                     | -       | Indicates whether the page is open or closed. The parent component needs to manage this state                                                                                                                                                                                                     |
+| `title`              | `string`   |    ✅    | -                     | -       | The title of the page                                                                                                                                                                                                                                                                             |
+| `subtitle`           | `node`     |    -     | -                     | -       | The subtitle of the page, usually a React component                                                                                                                                                                                                                                               |
+| `onClose`            | `function` |    -     | -                     | -       | Called when the page closes (click on overlay, click on close button, press ESC). If the function is not provided, the page cannot be closed by any of the listed options.                                                                                                                        |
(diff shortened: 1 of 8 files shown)


packages/application-components/src/components/modal-page/README.md
```diff
@@ -0,0 +1,54 @@
+# ModalPage
+
+## Usage
+
+```js
+import ModalPage from '@commercetools-local/components/modal-page';
```
montezume:
```suggestion
import { ModalPage } from '@commercetools-frontend/application-components';
```

packages/application-components/src/index.js
```diff
@@ -11,3 +11,5 @@ export {
   default as ConfirmationDialog,
 } from './components/dialogs/confirmation-dialog';
 export { default as FormDialog } from './components/dialogs/form-dialog';
+
+export { default as ModalPage } from './components/modal-page';
```
montezume: 👍

packages/application-components/src/components/modal-page/README.md
```diff
@@ -0,0 +1,54 @@
...
+#### Usage
+
+```js
+<ModalPage
+  isOpen={true}
+  onClose={handleClose}
+  title="Lorem ipsus"
+  subtitle={<Text.Body>{'Lorem ipsus ...'}</Text.Body>}
+  components={{
+    actions: (
```
tdeekens: What if we split and have to. So actions.primary and actions.secondary. While the modal retains control over position to limit the APIs flexibility.

packages/application-components/src/components/modal-page/modal-page.js
```diff
@@ -0,0 +1,163 @@
...
+      return baseZIndex + 1;
+    case 'three':
+      return baseZIndex + 2;
+    default:
+      return baseZIndex;
+  }
+};
+
+const getColumnSpacing = level => {
+  const baseSpacing = 64;
```
tdeekens: Should this be a spacing var?

packages/application-components/src/components/modal-page/modal-page.js
```diff
@@ -0,0 +1,163 @@
...
+          <div />
+        </React.Fragment>
+      );
+    case 'three':
+      return (
+        <React.Fragment>
+          <div />
+          <div />
+          <div />
+        </React.Fragment>
```
tdeekens: Can you elaborate what happens here and why we render those empty divs?

packages/application-components/src/components/modal-page/modal-page.js
```diff
@@ -0,0 +1,163 @@
...
+          <div className={styles['page-header']}>
+            <Spacings.Inline
+              scale="m"
+              alignItems="center"
+              justifyContent="space-between"
+            >
+              <Text.Headline
+                elementType="h3"
+                truncate={true}
+                title={props.title}
```
tdeekens: Q: why pass it twice. I forgot how they are applied differently.

packages/application-components/src/components/modal-page/modal-page.js
```diff
@@ -0,0 +1,163 @@
...
+      </Card>
+    </div>
+  </Modal>
+);
+ModalPage.displayName = 'ModalPage';
+ModalPage.propTypes = {
+  isOpen: PropTypes.bool.isRequired,
+  onClose: PropTypes.func,
+  zIndex: PropTypes.number,
+  baseZIndex: PropTypes.number,
```
tdeekens: Why base and normal? Would one be enough and we just add to that. What would e.g. happen if base would be bigger the normal index?

packages/application-components/src/components/modal-page/modal-page.mod.css
```diff
@@ -0,0 +1,68 @@
...
+.page-header > * + * {
+  margin: var(--spacing-s) 0 0;
+}
+
+.page-content {
+  border-top: 1px solid var(--border-color-separator);
+  padding-top: var(--spacing-m);
+  overflow: auto;
+  flex: 1;
+}
```
tdeekens: Could we use emotion here or was there something not allowing that?

packages/application-components/src/components/modal-page/modal-page.js
jonnybel: Yes, and it will be a different value.

packages/application-components/src/components/modal-page/README.md
jonnybel: Yep, that's a good idea.

packages/application-components/src/components/modal-page/modal-page.js
jonnybel: These are for adding more space to the left of the component, with a grey semi-transparent overlay. The space should programmatically increase with the level of the modal.

However, the way this is done will change, as there will be more than 3 levels, and we'll rethink the logic.

packages/application-components/src/components/modal-page/modal-page.mod.css
jonnybel: Yes we will 😄

montezume: Our idea was that first we add it in css modules, with visual tests, and then when we migrate to emotion, we can rely on the visual testing to make sure the visuals match 🎈

jonnybel:
@tdeekens Thanks for the suggestions.
This first PR is basically just for moving the same code from the MC-FE as is, adding a storybook story and some VRTs, so that I and @montezume can then build on top of this and work simultaneously on separate new features this is going to have.

montezume:
@tdeekens our initial idea was to start by moving the current code from `application-core` to `application-kit/components`, then split up the work to make the ModalPage match the designs

packages/application-components/src/components/modal-page/modal-page.mod.css
tdeekens: Top idea!

PR #638, opened 2019-05-17
chore: update deps

tdeekens:
...before releasing.


application-templates/starter/package.json
```diff
@@ -22,7 +22,7 @@
     "react": "16.8.6",
     "react-apollo": "2.5.5",
     "react-dom": "16.8.6",
-    "react-intl": "2.8.0",
+    "react-intl": "2.9.0",
```
tdeekens: Will update with it in the mc too to have one version.

package.json
```diff
@@ -83,19 +83,19 @@
...
+    "lint-staged": "8.1.7",
+    "prettier": "1.17.1",
+    "puppeteer": "1.16.0",
     "rcfile": "1.0.3",
     "read-pkg-up": "6.0.0",
-    "rollup": "1.11.3",
+    "rollup": "1.12.1",
     "rollup-plugin-babel": "4.3.2",
-    "rollup-plugin-commonjs": "9.3.4",
+    "rollup-plugin-commonjs": "10.0.0",
```
tdeekens: These now support new side-effect features.

montezume: I was looking at this, I think the "breaking" change is it requires rollup the new rollup, which you updated as well 👍

packages/mc-scripts/package.json
```diff
@@ -55,12 +55,12 @@
     "postcss-reporter": "6.0.1",
     "postcss-safe-parser": "4.0.1",
     "promise": "8.0.3",
-    "react-dev-utils": "7.0.3",
+    "react-dev-utils": "9.0.1",
```
tdeekens: The breaking change here is the method signature of `createCompiler` to accept an object. I fixed that in `mc-scripts`.

package.json
tdeekens: Yes. That's how I understood it too when reading through their PRs.

PR #642, opened 2019-05-17
feat(components): new Modal Page design

jonnybel:
#### Summary

This PR reimplements the stacking Modal Pages according to the new designs.


packages/application-components/src/components/modal-page/README.md
```diff
@@ -14,41 +14,42 @@ Modal pages are controlled components used to render a page using a modal contai
...
-        <PrimaryButton
-          label="Save"
-          isDisabled={isDisabled}
-          onClick={handleSubmit}
-        />
-      </Spacings.Inline>
-    ),
-  }}
+  headerActions={
+    <Spacings.Inline alignItems="flex-start">
```
tdeekens: Didnt we agree that splitting actions into two would be good?

jonnybel: Yes, however after further analysing the designs and use cases for this, I found cases where there are 3 different buttons, and where the type of the buttons (primary/secondary, iconButton) also differs.
So, given this need for flexibility which would require a bunch of different parameters, I decided to keep the actions has a render prop.

tdeekens: > actions has a render prop.

It's not a render-prop. That would imply it being an fn receiving arguments. Is that what we need here?

jonnybel: Sorry, you're right, not a render prop. It's a node which its rendered aligned in the right side of the Header section of the Modal.

packages/application-components/src/components/modal-page/internals/modal-page-top-bar.js
```diff
@@ -0,0 +1,48 @@
...
+  FlatButton,
+  AngleLeftIcon,
+  SecondaryIconButton,
+} from '@commercetools-frontend/ui-kit';
+import { getTopBarStyles } from './modal-page.styles';
+
+const ModalPageTopBar = props => (
+  <div css={getTopBarStyles(props)}>
+    <Spacings.Inline>
+      <FlatButton
```
montezume: from the designs it looks like there can be an array of previous paths right?

jonnybel: At first I thought so, like breadcrumbs, but then I didn't find any other than just the previous path and the current one. I'm going to clarify this.

jonnybel: Actually, on another look I've found sometimes there are more than just one previous paths, so I'm going to refactor this to use an array for such.

jonnybel: After discussing with design, this is only supposed to act like a "Back Button". There are plans to implement a "breadcrumbs" component but not for this phase.

packages/application-components/src/components/modal-page/internals/modal-page-top-bar.js
```diff
@@ -0,0 +1,48 @@
...
+        tone="primary"
+        label={props.previousPath || ''}
```
tdeekens: defaultProps instead?

packages/application-components/src/components/modal-page/modal-page.js
```diff
@@ -29,136 +28,74 @@ const parentSelector = () =>
...
+  children: PropTypes.node.isRequired,
+  showTopBar: PropTypes.bool,
+  showHeader: PropTypes.bool,
   baseZIndex: PropTypes.number,
-  title: PropTypes.string.isRequired,
-  subtitle: PropTypes.node,
-  components: PropTypes.shape({
-    actions: PropTypes.node,
+  topBarTone: PropTypes.oneOf(['white', 'gray']),
+  topBarLabels: PropTypes.shape({
```
tdeekens: Nit: we often prefer flat props to avoid people passing in new object references. so `topBarCurrentPathLabel` etc. The sum of character needing typing is the same after all.

packages/application-components/src/components/modal-page/modal-page.js
```diff
@@ -29,136 +28,74 @@ const parentSelector = () =>
...
+  title: PropTypes.string.isRequired,
+  zIndex: PropTypes.number,
   isOpen: PropTypes.bool.isRequired,
   onClose: PropTypes.func,
-  zIndex: PropTypes.number,
+  subtitle: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+  children: PropTypes.node.isRequired,
+  showTopBar: PropTypes.bool,
+  showHeader: PropTypes.bool,
   baseZIndex: PropTypes.number,
```
tdeekens: This combincation still makes me worry but I guess it's only me.

packages/application-components/src/components/modal-page/internals/modal-page-top-bar.js
jonnybel: Ah right, and I'm actually setting it as _requiredIf showTopBar = true_

packages/application-components/src/components/modal-page/modal-page.js
jonnybel: It might be useful, but it might also never be used. I guess I can remove it and we add it later if it becomes necessary.

packages/application-components/src/components/modal-page/README.md
```diff
@@ -14,41 +14,42 @@ Modal pages are controlled components used to render a page using a modal contai
...
+| `headerActions`  | `node`             |          | -                | -       | This node will be rendered in the header section, on the right side of the subtitle, as long as the `showHeader` prop isn't falsy.                                                                                                                                                                          |
+| `topBarLabels`   | `object`           |          |                  | `{}`    | Use this object to pass `currentPath` and `previousPath` fields as labels to the top bar.                                                                                                                                                                                                                   |
+| `topBarTone`     | `string`           |          | [`white`,`gray`] | `white` | Set the colour of the top bar section.                                                                                                                                                                                                                                                                      |
```
montezume: I'm not sure I would use `tone`. Why not use `color` or `backgroundColor`?

packages/application-components/src/components/modal-page/modal-page.js
jonnybel: Yeah, this was one of the things I wasn't so sure myself and was awaiting feedback.

packages/application-components/src/components/modal-page/README.md
jonnybel: Fine by me :) I also thought `shade`, since it's only b&w, but yeah `color` is just easier to understand.

tdeekens: Good point. The MC has a history of coming up with any possible name for something like this. `shade` would be something new. We have `tone`, `color` and `theme` and I think just color for these things is fine :)

packages/application-components/src/components/modal-page/modal-page.js
```diff
@@ -29,136 +28,74 @@ const parentSelector = () =>
...
+  level: PropTypes.number,
+  title: PropTypes.string.isRequired,
+  zIndex: PropTypes.number,
   isOpen: PropTypes.bool.isRequired,
   onClose: PropTypes.func,
-  zIndex: PropTypes.number,
+  subtitle: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+  children: PropTypes.node.isRequired,
+  showTopBar: PropTypes.bool,
+  showHeader: PropTypes.bool,
```
emmenko: Have you considered having different components with specific properties instead of one component with many different combinations?
For example, the dialog components have been semantically split into Info,Confirmation,Form components, even though their APIs are very similar as they are built with the same set of internal components. The reason for splitting them is to give them semantic names, which makes it clear when to use what and to keep the respective APIs restrictive in regards to the unique properties of each components.

With that in mind, I would expect something similar to the “modal page” components, for example:
- `InfoModalPage` (the one without actions)
- `FormModalPage` (the one with actions and white header background, without any sub pages)
- `TabularModalPage` (the one with actions and gray header background, with sub pages as tabs)

The names can change, but the idea is what I’m trying to explain.

jonnybel: Yes, I've considered that, even though I first opted for this more "generic" version of this Modal Page, but I've already separated some of the elements into internal components (topBar and Header) with that in mind.

I only left out the Tabular navigation: currently the dev would pass that whole part as the content (children) of the modal, and "turn off" the header.

But if we all agree on this, I can build on top of the current implementation and provide the 3 different, more specific versions of it 🙂

emmenko: Yes, for me it comes down to make the lives of the developers easier, by reducing the cognitive overhead (in this case in terms of correct combinations of props based on the type of usage).

jonnybel: Regarding the `TabularModalPage`: If I understand correctly, if we're gonna integrate the tab-page navigation in this ModalPage, we're gonna need to migrate the Tab components from the MC-FE to app-kit as well.

In fact, I think having a Tab/Tabs component is actually agnostic enough to be a good candidate for UI-kit! What do you think?

montezume: I also think Tab component would fit for UI Kit.

emmenko: To clarify: I didn’t mean to integrate tabs in this component, it’s just a name to suggest the usage of item regards to its layout.
For instance, the FormDialog also does not implement a form, it just indicates that it should be used for form dialogs

jonnybel: Ah, I see, so in this case, the developer would use the "premade" version of `TabularModalPage` which "premade" with the rest of the pieces of the Modal such as TopBar and Header already styled accordingly, and would then nest the Tab navigation and TabContent within the `children` of the modal.

emmenko: Yes, the tabs part is left to the children.

packages/application-components/src/components/modal-page/internals/modal-page.styles.js
```diff
@@ -0,0 +1,52 @@
+import { css, styled, keyframes } from '@emotion/core';
```
montezume: does this work? I thought there was no styled export in `@emotion/core` 🤔 

[ref](https://github.com/emotion-js/emotion/blob/master/packages/core/src/index.js)

the docs say to do `import styled from '@emotion/styled` [docs](https://emotion.sh/docs/styled)

packages/application-components/src/components/modal-page/internals/modal-page.styles.js
```diff
@@ -0,0 +1,52 @@
...
+    : props.baseZIndex + props.level};
+  top: 0;
+  left: 0;
+  width: 100%;
+  height: 100%;
+  background-color: rgba(32, 62, 72, 0.5);
+  animation: ${fadeIn} 0.2s ease;
+`;
+
+export const ContentWrapper = () => styled.div`
```
montezume: why does this need to be a react component instead of just exporting styled.div?

jonnybel: It doesn't 😄

packages/application-components/src/components/modal-page/internals/modal-page.styles.js
jonnybel: I haven't tested it yet (hence the WIP state). Thanks for pointing this out 😄

emmenko:
What's the status here? Do you need help finishing the implementation?

jonnybel:
@emmenko I slowed down a bit over the last two days because of lots of meetings, but I'm trying to wrap this up today, it's all good!

But meanwhile, we also wanted to start deploying the Storybook of app-kit on Netlify so that design can properly test it!

emmenko:
> But meanwhile, we also wanted to start deploying the Storybook of app-kit on Netlify so that design can properly test it!

I would like to do that differently, and use the website. We can talk about it and I can show you what you can do with that.

packages/application-components/src/components/modal-page/internals/modal-page-top-bar.js
jonnybel: Talked with design and this now defaults to "Go Back" (translated)

packages/application-components/src/components/modal-pages/internals/modal-page-header.js
```diff
@@ -14,43 +14,50 @@ import filterDataAttributes from '../../../utils/filter-data-attributes';
...
-              isDisabled={props.isPrimaryButtonDisabled}
-              {...filterDataAttributes(props.dataAttributesPrimaryButton)}
-            />
-          </Spacings.Inline>
-        ))}
-    </Spacings.Inline>
+      <Text.Subheadline elementType="h4" truncate>
+        {props.title}
+      </Text.Subheadline>
+      {typeof props.subtitle === 'string' ? (
```
tdeekens: What if I pass `14` as a `subtitle`? Maybe using `React.isValidElement` works better?

jonnybel: You are right, `React.isValidElement` is better 👍

packages/application-components/src/components/modal-pages/internals/modal-page-header.js
```diff
@@ -0,0 +1,98 @@
...
+      & > * + * {
+        margin-left: ${customProperties.spacingM};
+      }
+    `}
+  >
+    <div
+      css={css`
+        overflow: hidden;
+        & * + * {
+          margin-top: ${customProperties.spacingM} !important;
```
jonnybel: This is basically for the same purpose as using a `<Spacings>` component, but I couldn't use it because I needed to have `overflow: hidden` in order to properly truncate the text of its children.

jonnybel:
I've added more VRTs and included examples of the Modal Pages to the new `appkit` website.

emmenko:
Thanks, I'll take a look once the deploy preview is ready 🙌

packages/application-components/src/components/modal-pages/form-modal-page/form-modal-page.story.js
```diff
@@ -0,0 +1,110 @@
...
+import withReadme from 'storybook-readme/with-readme';
+import { Formik } from 'formik';
+import { PORTALS_CONTAINER_ID } from '@commercetools-frontend/constants';
+import {
+  Spacings,
+  TextInput,
+  TextField,
+  SecondaryButton,
+} from '@commercetools-frontend/ui-kit';
+import { number } from '@storybook/addon-knobs/dist/deprecated';
```
emmenko: Is this import correct? 🤔

packages/application-components/src/components/modal-pages/form-modal-page/form-modal-page.js
```diff
@@ -0,0 +1,90 @@
...
+import PropTypes from 'prop-types';
+import { injectIntl } from 'react-intl';
+import ModalPage from '../internals/modal-page';
+import ModalPageTopBar from '../internals/modal-page-top-bar';
+import ModalPageHeader from '../internals/modal-page-header';
+import { ContentWrapper } from '../internals/modal-page.styles';
+import messages from '../internals/messages';
+
+const useLabel = (label, intl) =>
+  typeof label === 'string' ? label : intl.formatMessage(label);
```
emmenko: Wait, isn't this wrong? The prop-type says it's either a string or a node, not an intl message 🤔

packages/application-components/src/components/modal-pages/form-modal-page/form-modal-page.js
```diff
@@ -0,0 +1,90 @@
...
+  intl: PropTypes.shape({
+    formatMessage: PropTypes.func.isRequired,
+  }).isRequired,
+};
+FormModalPage.defaultProps = {
+  labelPrimaryButton: messages.confirm,
+  labelSecondaryButton: messages.cancel,
+};
+
+FormModalPage.Intl = messages;
```
emmenko: Careful, you're exposing also messages that should not be used for the button labels, e.g. "back".

packages/application-components/src/components/modal-pages/internals/modal-page-header.js
```diff
@@ -0,0 +1,98 @@
...
+
+const ModalPageHeader = props => (
+  <div
+    css={css`
+      display: flex;
+      align-items: flex-end;
+      justify-content: space-between;
+      margin: 0 ${customProperties.spacingM};
+      padding: ${customProperties.spacingM} 0;
+      border-bottom: 1px solid #afafaf;
```
emmenko: Why is this color not one of the design tokens?

packages/application-components/src/components/modal-pages/internals/modal-page-top-bar.js
```diff
@@ -0,0 +1,112 @@
...
+        label={props.intl.formatMessage(messages.close)}
+        onClick={props.onClose}
+        icon={<LargeCloseIcon />}
+      />
+    )}
+  </div>
+);
+ModalPageTopBar.displayName = 'ModalPageTopBar';
+ModalPageTopBar.propTypes = {
+  color: PropTypes.oneOf(['white', 'gray']),
```
emmenko: Can we use other names here? Like `surface` and `neutral`

packages/application-components/src/components/modal-pages/internals/modal-page-top-bar.js
```diff
@@ -0,0 +1,112 @@
...
+      align-items: center;
+      justify-content: space-between;
+      padding: ${customProperties.spacingS} ${customProperties.spacingM};
+      background-color: ${props.color === 'gray'
+        ? customProperties.colorGray95
+        : customProperties.colorWhite};
+      border-bottom: 1px solid
+        ${props.color === 'gray'
+          ? customProperties.colorWhite
+          : customProperties.colorGray};
```
emmenko: Those tokens are deprecated, we should use the new ones: `colorSurface` and `colorNeutral`

packages/application-components/src/components/modal-pages/internals/modal-page-top-bar.js
```diff
@@ -0,0 +1,112 @@
...
+  <div
+    css={css`
+      position: relative;
+      display: flex;
+      align-items: center;
+      justify-content: space-between;
+      padding: ${customProperties.spacingS} ${customProperties.spacingM};
+      background-color: ${props.color === 'gray'
+        ? customProperties.colorGray95
+        : customProperties.colorWhite};
```
emmenko: Same here: `colorNeutral95` and `colorSurface`

packages/application-components/src/components/modal-pages/internals/modal-page.styles.js
```diff
@@ -0,0 +1,53 @@
+import { css, keyframes } from '@emotion/core';
+import styled from '@emotion/styled';
+import { customProperties } from '@commercetools-frontend/ui-kit';
+
+const slideFromLeft = keyframes`
```
emmenko: Didn't we say to keep the animation disabled for now?

packages/application-components/src/components/modal-pages/form-modal-page/form-modal-page.story.js
jonnybel: Whoops, VScode automatically did this import and I didn't notice 😄
Good catch!

packages/application-components/src/components/modal-pages/form-modal-page/form-modal-page.js
jonnybel: Hmm, this is used for the props `labelSecondaryButton`, `labelPrimaryButton`, which are either `string` or an `intl` shape. Aren't you confusing it with `subtitle`?

emmenko: You're right, I got confused 🤦‍♂️

We use the same in `dialog-footer.js`, so maybe we can rename it there as well.

packages/application-components/src/components/modal-pages/internals/modal-page.styles.js
jonnybel: I've talked with Sven and also showed the animation to Filip here in Berlin and they were quite happy about it and decided it should stay 🙂

packages/application-components/src/components/modal-pages/internals/modal-page-header.js
jonnybel: I coudn't find it 😕 I can have another look, I might have missed it...

packages/application-components/src/components/modal-pages/form-modal-page/form-modal-page.js
```diff
@@ -0,0 +1,96 @@
...
+  intl: PropTypes.shape({
+    formatMessage: PropTypes.func.isRequired,
+  }).isRequired,
+};
+FormModalPage.defaultProps = {
+  labelPrimaryButton: messages.confirm,
+  labelSecondaryButton: messages.cancel,
+};
+
+FormModalPage.Intl = { confirm: messages.confirm, cancel: messages.cancel };
```
emmenko: How about we split the messages into 2 files?
- `button-messages.js`
- `messages.js`

Then we can simply `FormModalPage.Intl = buttonMessages;`

jonnybel: I considered it, but preferred this solution to prevent having too many files. But, I have no strong opinion one way or another, so if you prefer to split we can do that.

emmenko: I fear that we forget to update the list here if we add more messages.

Also, we should reuse the same messages that we had in dialogs, to avoid translating them again.

emmenko: Here you go: https://github.com/commercetools/merchant-center-application-kit/pull/642/commits/19a1ed07dada5b647f1881c4084647a66975216d

jonnybel: Thanks 😄

tdeekens:
What's the status here. I am entirely lost in the come and go and duration of this. Sorry. Is it ready for review?

emmenko:
No, we're working on improving the interactive examples and the VRTs.

emmenko:
I opened a PR to extract some of the logic to improve the interactive example setup: #692

I'll take a look also at using react-live, so that people can mess around with the example, if they want to.

montezume:
Is it possible to make the website examples not so tiny?

emmenko:
> Is it possible to make the website examples not so tiny?

Everything is possible! Keep in mind also that I haven't done a "design review" of the website yet, just focussing on the technical aspects first.

packages/application-components/src/components/dialogs/confirmation-dialog/README.md
```diff
@@ -31,7 +31,7 @@ Confirmation dialogs are controlled components used to prompt the user to confir
...
-| `title`                         | `string`                   |    ✅    | -                 | -                                 | The title of the Info Dialog                                                                                                                                                                                                                                                                                |
+| `title`                         | `string`                   |    ✅    | -                 | -                                 | The title of the Confirmation Dialog                                                                                                                                                                                                                                                                        |
```
montezume: 😄

jonnybel:
It's okay to review now. Even if we add more interactive examples, the core components are the ones which need a proper code review.

emmenko:
I added some improvements in the examples. I also noticed that the Tabular component is not implemented correctly, we can talk about it tomorrow.

emmenko:
Very nice, the improved examples looks dope!! 🙌

emmenko:
The VRT tests also need to be updated

packages/application-components/src/components/modal-pages/tabular-modal-page/tabular-modal-page.example.js
```diff
@@ -0,0 +1,174 @@
...
+                  }
+                />
+              </Spacings.Inline>
+            </Spacings.Stack>
+          </TabularModalPage>
+        )}
+      </ExampleWrapper>
+      <Spacings.Inline>
+        <Spacings.Inline alignItems="center">
+          <ToggleInput
```
emmenko: Nitpick: we should use the small size

emmenko:
@jonnybel heads up: I've rebased and I'm fixing some typos in the readmes.

emmenko:
VRTs look good now! 🙌

tdeekens:
So @emmenko you're not working on this? @jonnybel you're done adding/coordinating things and it can be reviewed?

emmenko:
We're working on this yes :)

Abel is taking care of the UX review, I worked on making UX reviews easier (#733 ). I'd like to merge that other PR first and refactor the examples here, so that we can test it properly.

jonnybel:
OK, I've rebased and integrated the new "knobs", and thrown away those toggles I had added for using custom controls, and I'm now doing that with the knobs instead.

jonnybel:
Design has reviewed and approved.

emmenko:
@Shecki @filippobocik can you also review and approve the VRTs? Thanks

packages/application-components/src/components/modal-pages/internals/modal-page-header.js
```diff
@@ -0,0 +1,84 @@
...
```
emmenko: What about this? We should use a color from uikit tokens

packages/application-components/src/components/modal-pages/internals/modal-page-top-bar.js
```diff
@@ -0,0 +1,126 @@
...
+        : customProperties.colorSurface};
+      border-bottom: 1px solid
+        ${props.color === 'neutral'
+          ? customProperties.colorSurface
+          : customProperties.colorNeutral};
+      & * + * {
+        margin-left: ${customProperties.spacingS};
+      }
+
+      /* FIXME: these "dirty" styles should be removed when the new Breadcrumbs component is implemented */
```
emmenko: Is there a task for that already?

website/src/templates/markdown.js
```diff
@@ -322,15 +323,17 @@ const components = {
...
-    <MDXProvider components={components}>
-      <TypographyPage>
-        <SEO title={props.data.mdx.frontmatter.title} />
-        {/* This wrapper div is important to ensure the vertical space */}
-        <div>
-          <MDXRenderer>{props.data.mdx.code.body}</MDXRenderer>
-        </div>
-      </TypographyPage>
-    </MDXProvider>
+    <IntlProvider locale="en">
```
emmenko: I don't think we need this, it's defined now in the `example-wrapper.js`

jonnybel: Yeah, forgot to remove this after the latest rebase

packages/application-components/src/components/modal-pages/internals/modal-page-top-bar.js
jonnybel: There isn't a task as in a "jira ticket", but design said it's planned as an upcoming new feature/component.

jonnybel:
@emmenko I've addressed the problem of the z-index of the playground InfoDialog, as well as the other remarks.
Let me know if there's something more.

packages/application-components/src/components/internals/for-docs/example-wrapper/example-wrapper.js
```diff
@@ -159,6 +160,7 @@ const ExampleWrapper = props => {
                 title="Playground"
                 size="scale"
                 isOpen={isOpen}
+                zIndex={1001}
```
emmenko: This does not solve the issue.

image

We need to set this higher, I would put it to `1100`, so like with an offset of `100` (just to be sure).

packages/application-components/src/components/internals/for-docs/example-wrapper/example-wrapper.js
```diff
@@ -19,6 +19,7 @@ import { InfoDialog } from '@commercetools-frontend/application-components';
 const PreviewContainer = styled.div`
   position: relative;
   width: 100%;
+  overflow: hidden;
```
emmenko: What's this for?

packages/application-components/src/components/internals/for-docs/example-wrapper/example-wrapper.js
jonnybel: Oh, I only tested with one opened Modal. 
👍

packages/application-components/src/components/internals/for-docs/example-wrapper/example-wrapper.js
jonnybel: It's for the shadows around the portal not to "bleed" out of the PreviewContainer.
image
Design asked about this, but in practice, those shadows won't show on the edge of the pages, only the ones on the left. So I decided to clip them from the PreviewContainer to make it clearer how it will appear on a page.

emmenko: Ah ok, makes sense 👌

jonnybel:
I think the VRTs need to be approved again, perhaps because I changed that `border-color` to use one of the design tokens (the "nearest" one) and it's not exactly the same.

emmenko:
Ok, should I just approve them again?

jonnybel:
Yeah, everything seems to be correct.

emmenko:
Done, feel free to finally merge the PR! Thanks a lot for the patience and all the improvements!! 🙌

PR #643, opened 2019-05-18
chore: update flopflip

tdeekens:
#### Summary

This pull request updates flopflip with a change suggested by @gnerkus. It should silence Sentry a bit when flopflip fails to initialize.

I will merge this as it's just a dep update.


packages/application-shell/package.json
```diff
@@ -51,9 +51,9 @@
     "@commercetools-frontend/sentry": "13.6.0",
     "@commercetools-frontend/storage": "13.3.5",
     "@commercetools-frontend/url-utils": "13.3.5",
-    "@flopflip/launchdarkly-adapter": "2.5.0",
+    "@flopflip/launchdarkly-adapter": "2.6.0",
```
tdeekens: Internally initialization changed to use another API from LD. While we can opt into getting an error thrown at us (we don't want yet) if initialzation fails.

PR #651, opened 2019-05-20
fix(circleci): to use new compare-url orb

tdeekens:
#### Summary

This release fixes the issue with a PR having one commit without an ancestor. 

https://github.com/iynere/compare-url/pull/23


tdeekens:
Seems like the new version wasn't published yet. Let's use 1.0.0 anyway.

PR #656, opened 2019-05-24
fix(app-shell): quick access env not always set

tdeekens:
#### Summary

We should not assume an env always to be set when evaluating here. This breaks quick access in the accounts app cause it currently doesn't have any custom env passed.


tdeekens:
Re-tested locally. It is a beauty.

PR #669, opened 2019-05-30
chore(mc-scripts): update with upstream react-scripts for (relevant) latest changes

emmenko:


packages/mc-scripts/config/create-webpack-config-for-production.js
```diff
@@ -79,16 +79,56 @@ module.exports = ({
     optimization: {
       minimizer: [
         new TerserPlugin({
+          terserOptions: {
```
emmenko: Differences in sizes brought by those changes:

Screenshot 2019-05-30 at 21 25 27

PR #670, opened 2019-05-31
feat(mc-scripts): add static command to compile final index.html file (and security headers)

emmenko:
This PR introduces a few changes and some new features around the mc-scripts tooling.

## Background
At the moment, custom apps are deployed using the mc-http-server, which requires the cloud provider to either run it as a node app or as a docker container.
However, a custom application is nothing more than a bunch of static files, as on production it's supposed to be served behind the mc proxy.

Therefore, I would like to enhance our tooling to make it easier to deploy custom applications as static files. This will allows us to also provide example to Netlify or Now v2.

On top of that, I would also take this chance as a starting point to experiment with an idea that I had about deploying all of our mc apps as static apps (_more on that will follow up_).

## Enhancements

- the `mc-scripts` now has a new command `compile-html`, which is more or less what previously the `mc-http-server` script was doing before starting the server
  - the `static` command accepts some CLI options to read the different config files so that it's able to compile the `index.html` and the CSP directives
- the `mc-http-server` has been refactored to use the extracted logic
- the `playground` application has been updated to be able to deploy to the new Now v2
  - it uses the `compile-html` command to prepare the static files into the `public` folder as well as generating a `*.now.json` file (using a custom transformer)
  - the deployed application contains static files, which are served by Now's [CDN at the edge](https://zeit.co/docs/v2/domains-and-aliases/cdn/)
  - I also added a "fake" route handler for /login and /logout in case someone access the domain directly and prevents an infinite redirect loop to the login route. Those routes are deployed as lambda functions.

PS: with this approach, we can have a different transformer to generate a Netlify config for the headers.


packages/mc-html-template/lib/compile-html.js
```diff
@@ -0,0 +1,66 @@
...
+const requiredOptions = ['envPath', 'publicAssetsPath'];
+
+/**
+ * Options:
+ * - envPath
+ * - cspPath
+ * - publicAssetsPath
+ * - useLocalAssets
+ */
+module.exports = async function compileHtml(options) {
```
emmenko: Most of the logic here has been extracted from the `mc-http-server/bin/start-mc.js`

packages/mc-http-server/bin/start-mc.js
```diff
@@ -1,52 +1,76 @@
...
 const fs = require('fs');
 const path = require('path');
-const shell = require('shelljs');
-const fetch = require('node-fetch');
-const {
-  replaceHtmlPlaceholders,
-} = require('@commercetools-frontend/mc-html-template');
+const mri = require('mri');
+const shelljs = require('shelljs');
+const { compileHtml } = require('@commercetools-frontend/mc-html-template');
```
emmenko: Most of the logic has been moved into the `compileHtml` function

packages/mc-http-server/bin/start-mc.js
```diff
@@ -1,52 +1,76 @@
...
 const {
-  packageLocation: assetsPath,
+  packageLocation: applicationStaticAssetsPath,
 } = require('@commercetools-frontend/assets');
 const startServer = require('../server');
-const options = require('../load-options');
+
+const flags = mri(process.argv.slice(2), { alias: { help: ['h'] } });
```
emmenko: The CLI args parsing was previously defined in `load-options` and has been now moved here

packages/mc-scripts/static.js
```diff
@@ -0,0 +1,125 @@
...
+      require.resolve(flags.transformer);
+    } catch (error) {
+      throw new Error(
+        `Could not load transformer module "${flags.transformer}"\n${
+          error.stack
+        }`
+      );
+    }
+    const transformerFn = require(flags.transformer);
+    transformerFn(compiled);
```
emmenko: This is the where we pass the compiled data to the transformer module

montezume: What does `yarn static` do? I'm confused about the naming. Does it build static version of the sites? Wouldn't it make more sense to include build in the name then? 😕

emmenko: I couldn't think of a better name. Do you suggest something like `build-static`?

What it does is basically two things:
- it compiles the `index.html.template` into `index.html`
- it moves all the assets (including some extra ones) into a `public` folder

packages/mc-html-template/lib/compile-html.js
```diff
@@ -0,0 +1,66 @@
...
+      indexHtmlContent: updatedIndexHtmlContent,
+    };
+  }
+
+  // TL;DR; This disables the `Cache-Control` and ensures that we get
+  // the latest version of the object.
+  const randomQueryParam = Date.now();
+  // Fetch `index.html.template` from remote CDN
+  const remoteIndexHtmlResponse = await fetch(
+    `${env.cdnUrl}/index.html.template?${randomQueryParam}`
```
montezume: can you explain why we need to do this?

emmenko: What exactly? The "random query" part? It's to ensure that we don't get a cached version of the file, always the up-to-date version.

montezume: I mean querying the cdn to get the template?

montezume: Why can't  we have the template locally?

emmenko: That's how it works currently (this logic was part of the mc-http-server previously).

When we bundle each app, the js/css assets are generated by webpack as well as the `index.html.template`. We upload them to our CDN, then we deploy the http server, which on startup will fetch and compile the `index.html.template` for the specific app, thus serving the latest assets (referenced in the template by webpack).

Does it clarify your question?

emmenko: > Why can't we have the template locally?

You can optionally choose to have the template locally (together with the other assets) by passing the option `--use-local-assets`. That's not the recommended approach for production though.

tdeekens:
I will review this tomorrow. I feel like already having reviewed 15 PRs today. Sorry for the delay.

emmenko:
No worries, this is low priority

playground/config/transformer-now.js
```diff
@@ -0,0 +1,44 @@
+const fs = require('fs');
```
tdeekens: Do we want to offer default transformers or do we require people to always write their own?

emmenko: I thought about it and I think for now it does not make sense to have a default transformers. It's also a bit tricky to provide a good default.
For now, I would just provide examples as documentation, so people can just copy-paste it and update it to their needs.

packages/mc-http-server/routes/logout.js
```diff
@@ -11,14 +11,7 @@ const isDev =
 module.exports = env => (request, response, next) => {
   devAuthentication.routes.logout(
     response,
-    isDev
-      ? []
-      : [
-          `Domain=${
-            require('../load-options').env.frontendHost.split('mc.')[1]
```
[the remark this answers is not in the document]
emmenko: Yep

emmenko:
Should we keep `mc-scripts static` or should we name the command differently?

montezume:
I'm sorry, I tried to review it but it's too hot and I can't concentrate and understand the code. I can try tomorrow or you can just merge.

emmenko:
No worries, take your time

emmenko:
@montezume do you still want to review this or if it's ok otherwise I would merge it.

playground/config/transformer-now.js
```diff
@@ -0,0 +1,44 @@
...
+    name: `mc-app-state-machines-${environmentKey}`,
+    alias: `mc-app-state-machines-${environmentKey}`,
+    regions: regions[env.location],
+    builds: [
+      { src: 'public/**', use: '@now/static' },
+      { src: 'config/fallback-route.js', use: '@now/node' },
+    ],
+    routes: [
+      {
+        src: '/(.*).(js.map|js|css|txt|html|png)',
```
montezume: should this be configurable? what if I want to host some svg? 🤔

montezume: or PDFs 😄

emmenko: It’s up to you to configure it the way you like it, this is just an example for the playground app.
We don’t ship a transformer out of the box.

packages/mc-scripts/bin/mc-scripts.js
```diff
@@ -15,6 +15,7 @@ if (commands.length === 0 || (flags.help && commands.length === 0)) {
   Commands:
   build            Bundles the application in production mode
   start            Starts the application using webpack dev server
+  static           Generates index.html and security headers (requires "mc-scripts build" to run before)
```
montezume:
```suggestion
  build:static           Generates index.html and security headers (requires "mc-scripts build" to run before)
```

emmenko:
Yeah I can change the name. I was also thinking of maybe include it in the `build` script behind a flag (`mc-scripts build —compile-html` or something similar)

montezume:
@emmenko that's a good idea as well.

Just to explain why "static" is not a good name. Take the same logic to our build script. Would you like to rename it `production`? Same reasoning for why we shouldn't use static 😉

emmenko:
Including it in the build script is maybe not a good idea, because of the different CLI options that we need to pass (so it's not just one). Having it as a different command makes more sense from that point of view.

Let me rename it to `mc-scripts compile-html`, because that's what it basically does. Is that better?

PR #678, opened 2019-06-03
chore: update deps

emmenko:
Closes #671 
Closes #672 
Closes #673 
Closes #674 
Closes #675 
Closes #676

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,7 +19,7 @@ aliases:
 executors:
   node_10:
     docker:
-      - image: circleci/node:10.15
+      - image: circleci/node:10.16
     working_directory: *working_directory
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -29,7 +29,7 @@ executors:
           TERM: xterm
 
 orbs:
-  compare-url: iynere/compare-url@1.0.2
+  compare-url: iynere/compare-url@1.1.0
```
emmenko: FYI: there seems to be a [different problem](https://github.com/iynere/compare-url/issues/25) with this version, it does not hurt to update though

montezume: Progress is progress 🤷‍♂

PR #682, opened 2019-06-04
refactor(http-server): to unset cookie only for local development

emmenko:
This might seem a breaking change, but for production usage it does not affect any one (besides our internal auth app).


emmenko:
We still need them for local development though...right?

PR #683, opened 2019-06-04
feat(assets): add ct logos

emmenko:


application-templates/starter/README.md
```diff
@@ -1,6 +1,6 @@
 <p align="center">
   <a href="https://commercetools.com/">
-    <img alt="commercetools logo" src="http://cdn.rawgit.com/commercetools/press-kit/master/PNG/72DPI/CT%20logo%20chrom%20black%20horizontal%20RGB%2072dpi.png">
+    <img alt="commercetools logo" src="https://unpkg.com/@commercetools-frontend/assets/logos/commercetools_primary-logo_horizontal_RGB.png">
```
montezume: don't we need to do a release first for this to work?

emmenko: Yes, but once released it will "just work" 😉

montezume: It will just work but we need to update the latest tag

emmenko: Yes, I meant to do this today. Have faith my young padawan 😁

PR #684, opened 2019-06-04
refactor: add css-in-js linting

tdeekens:
#### Summary

This enables/adds css-in-js linting. More comments inline with the code.

Closes #684

### DIFF
--- a/.stylelintignore
+++ b/.stylelintignore
@@ -1,2 +1,1 @@
-packages/application-components/
 website/
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,5 +1,8 @@
 {
-  "extends": "stylelint-config-standard",
+  "extends": [
+    "stylelint-config-standard",
+    "stylelint-config-prettier"
+  ],
   "plugins": [
     "stylelint-order"
   ],
@@ -28,6 +31,12 @@
     "rule-empty-line-before": null,
     "value-list-comma-newline-after": null,
     "indentation": null,
-    "selector-descendant-combinator-no-non-space": null
+    "selector-descendant-combinator-no-non-space": null,
+    "no-descending-specificity": null,
+    "no-duplicate-selectors": null,
+    "no-empty-source": null,
+    "declaration-empty-line-before": null,
+    "no-missing-end-of-source-newline": null,
+    "function-name-case": null,
   }
 }
--- a/jest.stylelint.config.js
+++ b/jest.stylelint.config.js
@@ -1,8 +1,15 @@
 module.exports = {
   runner: 'jest-runner-stylelint',
   displayName: 'stylelint',
-  moduleFileExtensions: ['css'],
-  modulePathIgnorePatterns: ['dist'],
-  testMatch: ['<rootDir>/**/*.css'],
+  moduleFileExtensions: ['css', 'js'],
+  modulePathIgnorePatterns: [
+    'dist',
+    '.spec.js',
+    '.visualspec.js',
+    '.visualroute.js',
+    '.story.js',
+    'packages/mc-http-server/public/',
+  ],
+  testMatch: ['<rootDir>/**/*.css', '<rootDir>/packages/**/*.js'],
   watchPlugins: ['jest-watch-typeahead/filename', 'jest-watch-master'],
 };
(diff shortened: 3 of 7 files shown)


.stylelintignore
```diff
@@ -1,2 +1 @@
-packages/application-components/
-website/
\ No newline at end of file
+website/
```
tdeekens: `stylelint` has an option `allow-empty-input` which if not passed makes stylelint fail if pointed at a location which is ignored again.

.stylelintrc
```diff
@@ -28,6 +31,12 @@
...
     "value-list-comma-newline-after": null,
     "indentation": null,
-    "selector-descendant-combinator-no-non-space": null
+    "selector-descendant-combinator-no-non-space": null,
+    "no-descending-specificity": null,
+    "no-duplicate-selectors": null,
+    "no-empty-source": null,
+    "declaration-empty-line-before": null,
+    "no-missing-end-of-source-newline": null,
+    "function-name-case": null,
```
tdeekens: Same as UI-Kit.

jest.stylelint.config.js
```diff
@@ -1,8 +1,15 @@
...
+  moduleFileExtensions: ['css', 'js'],
+  modulePathIgnorePatterns: [
+    'dist',
+    '.spec.js',
+    '.visualspec.js',
+    '.visualroute.js',
+    '.story.js',
+    'packages/mc-http-server/public/',
+  ],
+  testMatch: ['<rootDir>/**/*.css', '<rootDir>/packages/**/*.js'],
```
tdeekens: We want to css-in-js lint packages for now. The website can be enabled later.

packages/application-components/src/components/dialogs/internals/dialog-container.js
```diff
@@ -29,6 +29,11 @@ const getDefaultParentSelector = () =>
     ? document.body
     : document.querySelector(`#${PORTALS_CONTAINER_ID}`);
 
+const sizesStyles = props =>
+  css`
+    ${props.size === 'scale' ? 'height: 100%;' : ''}
+  `;
```
tdeekens: Needs to be extracted otherwise can't be parsed.

packages/application-components/src/components/dialogs/internals/dialog-container.js
```diff
@@ -43,6 +48,7 @@ const DialogContainer = props => (
         parentSelector={props.getParentSelector}
         ariaHideApp={false}
         style={{
+          // stylelint-disable-next-line selector-type-no-unknown
```
tdeekens: Assuming that this is a prop looking like a style prop but actually not being one.

jest.stylelint.config.js
emmenko: What is the reason for not enabling it now?

tdeekens: That it gives some errors I would like to follow up on after this. Also cause the website ain't 100% my turf :)

tdeekens:
Tests fail on this even though linting ain't catching it. I fail to spot the error. I went through all internsl of the modals and also stasrted commenting in/out things. @montezume any idea?

CleanShot 2019-06-04 at 16 22 59@2x

packages/application-components/src/components/dialogs/internals/dialog-footer.js
```diff
@@ -8,18 +8,18 @@ import {
 } from '@commercetools-frontend/ui-kit';
 import filterDataAttributes from '../../../utils/filter-data-attributes';
 
-const useLabel = (label, intl) =>
+const getLabel = (label, intl) =>
```
tdeekens: I just fear `useLabel` will at some point be assumed a hook by ESLint.

montezume:
@tdeekens I think changing

`${sizeStyles}` to `${sizeStyles(props)}` will fix it!

packages/application-components/src/components/dialogs/internals/dialog-container.js
```diff
@@ -69,7 +79,7 @@ const DialogContainer = props => (
             // 3. For the actual "> div" container with the content, we need to use normal pointer events so that clicking on it does not close the dialog.
             css={css`
               min-height: 0;
-              ${props.size === 'scale' ? 'height: 100%;' : ''}
+              ${sizeStyles}
```
montezume:
```suggestion
              ${sizeStyles(props)}
```

montezume:
Looks good to me. I'm really not sure why this ModalPage VRT won't stick 🤔

montezume:
I have approved it like 20 times

tdeekens:
I'll do the website css-in-js tmr as a follow up.

PR #688, opened 2019-06-05
feat(app-shell): export npm version

tdeekens:
#### Summary

This pull request adds a `version` export to the app-shell by reading the `npm_package_version`.

#### Description

This version can be used to e.g. send it to a metrics service or a logging system. 

1. Another appraoch of this would be to have a json plugin in rollup and read it from the package.json. However in that case you always get "one version" down as you build before you publish and bump the version.
2. Also the `version` could be exported through the `index.js`. However then it quickly looses context and can. `ApplicationShell.version` seems more intuitive to me.

### DIFF
--- a/packages/application-shell/src/components/application-shell/application-shell.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.js
@@ -385,6 +385,8 @@ export default class ApplicationShell extends React.Component {
   static defaultProps = {
     trackingEventWhitelist: {},
   };
+  static version = process.env.npm_package_version;
+
   redirectTo = targetUrl => window.location.replace(targetUrl);
   componentDidMount() {
     this.props.onRegisterErrorListeners({
--- a/yarn.lock
+++ b/yarn.lock
@@ -1045,6 +1045,11 @@
     react-modal "3.8.1"
     warning "4.0.3"
 
+"@commercetools-frontend/assets@13.3.5":
+  version "13.3.5"
+  resolved "https://registry.yarnpkg.com/@commercetools-frontend/assets/-/assets-13.3.5.tgz#16806a319c11223e75ae228b9c5de42ccdbd5633"
+  integrity sha512-PvXjkHbsGwgp34acuq7tKoHti2Nd31g7WVe4Tw/ho/XhjLvfi85iolFVeVdwZnkJ6qxg+EOKDCwaIocuhWfMMw==
+
 "@commercetools-frontend/ui-kit@9.8.1":
   version "9.8.1"
   resolved "https://registry.yarnpkg.com/@commercetools-frontend/ui-kit/-/ui-kit-9.8.1.tgz#16b1971ebe5999805f5cd7540b3e696607d43052"


tdeekens:
> Hmm how about we properly configure the user agent to send the library version? At the moment is kind of hardcoded (and for apollo is not configured I think).

Yes that would be another step - at least in my head. I presume you're also taking from the Node.JS SDK and Apollo Client? That way we could also see in logs what custom apps made requests.

emmenko:
Yes. Should I add those changes here or should I open a separate PR?

tdeekens:
You can add them here if you want. Whatever is easiest to you.

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -385,6 +385,8 @@ export default class ApplicationShell extends React.Component {
   static defaultProps = {
     trackingEventWhitelist: {},
   };
+  static version = process.env.npm_package_version;
```
emmenko: Btw, are the npm env variables always defined?

emmenko:
> You can add them here if you want. Whatever is easiest to you.

I'll open a separate PR

packages/application-shell/src/components/application-shell/application-shell.js
montezume: Don't we need to use a rollup plugin to define this? like rollup plugin replace?

emmenko:
How about we do something like apollo-client does, by exporting a `version.js` file?

https://github.com/apollographql/apollo-client/blob/master/packages/apollo-client/src/version.ts

We can look into how to integrate it in the lerna release workflow

packages/application-shell/src/components/application-shell/application-shell.js
tdeekens: No. It's a npm variable set. I also do the same in flopflip. More e.g. here https://github.com/facebook/create-react-app/issues/2466#issuecomment-312016096 or https://docs.npmjs.com/misc/scripts#packagejson-vars.

It's a good point that it would actual need interpolation/replacing.

tdeekens:
> How about we do something like apollo-client does, by exporting a version.js file?

Yeah that's the other way I mentioned. It's similar to just exporting a `export { version }` in the end. My point against that was that when importing you quickly loose track of what the version belongs to `import { version } from 'app-shell'` might quickly loose context whereas `ApplicationShell.version` seemed more clear to me. Open to either side or even both.

emmenko:
The import would be `application-shell/version`.

Also yeah, as @montezume mention we need to define the "process.env.packageVersion" in rollup replace

packages/application-shell/src/components/application-shell/application-shell.js
emmenko: Also, those variables are only available if you run the process as a npm script, correct? I think we need to make it more safe, instead of relying on an "optional" environment variable.

WDYT?

tdeekens:
> The import would be application-shell/version.

It wasn't about the import path but about the name of the import. Whatever. I'll just do it.

emmenko:
I think this requires a bit more work, if we want to properly generate the version within a lerna release process, which is that the version needs to be bumped before releasing the package. So on build time, the version is "outdated", thus we can't have it within the normal react code. It needs to be a different file that is generated during release.

At least that's my understanding, correct me if I'm wrong.

tdeekens:
CleanShot 2019-06-05 at 10 08 35@2x

That is correct.

emmenko:
Let's do this properly, as the implementation is more complicated than it seems. I opened an issue so that we can discuss it there #690

PR #694, opened 2019-06-05
refactor: for the http clients to send a x-user-agent with the package versions

emmenko:
Requires #693 to be merged first.

- [x] whitelist `x-user-agent` in the CORS config

### DIFF
--- a/application-templates/starter/env.json
+++ b/application-templates/starter/env.json
@@ -1,4 +1,5 @@
 {
+  "applicationName": "merchant-center-application-template-starter",
   "frontendHost": "localhost:3001",
   "mcApiUrl": "https://mc-api.commercetools.com",
   "location": "eu",
--- a/packages/application-shell/src/configure-apollo.js
+++ b/packages/application-shell/src/configure-apollo.js
@@ -15,7 +15,7 @@ import version from './version';
 const userAgent = createHttpUserAgent({
   name: 'apollo-client',
   version: apolloVersion,
-  libraryName: '@commercetools-frontend/application-shell',
+  libraryName: [window.app.name, 'application-shell'].join('/'),
   libraryVersion: version,
   contactUrl:
     'https://github.com/commercetools/merchant-center-application-kit/issues/new/choose',
--- a/packages/application-shell/src/configure-apollo.js
+++ b/packages/application-shell/src/configure-apollo.js
@@ -17,8 +17,7 @@ const userAgent = createHttpUserAgent({
   version: apolloVersion,
   libraryName: [window.app.name, 'application-shell'].join('/'),
   libraryVersion: version,
-  contactUrl:
-    'https://github.com/commercetools/merchant-center-application-kit/issues/new/choose',
+  contactUrl: 'https://git.io/fjuyC', // points to the appkit repo issues
   contactEmail: 'mc@commercetools.com',
 });
 
--- a/packages/application-shell/src/configure-apollo.js
+++ b/packages/application-shell/src/configure-apollo.js
@@ -1,4 +1,5 @@
 import ApolloClient from 'apollo-client';
+import apolloVersion from 'apollo-client/version';
 import { ApolloLink } from 'apollo-link';
 import apolloLogger from 'apollo-link-logger';
 import { createHttpLink } from 'apollo-link-http';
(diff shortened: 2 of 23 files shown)


packages/application-shell/src/configure-apollo.js
```diff
@@ -1,18 +1,32 @@
...
 } from 'apollo-cache-inmemory';
+import createHttpUserAgent from '@commercetools/http-user-agent';
 import { errorLink, headerLink, tokenRetryLink } from './apollo-links';
 import { isLoggerEnabled } from './utils/logger';
+import version from './version';
+
+const userAgent = createHttpUserAgent({
+  name: 'apollo-client',
+  version: apolloVersion,
+  libraryName: '@commercetools-frontend/application-shell',
```
montezume: couldn't we import this from package.json?

tdeekens: This way we would never spot a custom app as a `libraryName`, right?

packages/sdk/src/version.js
```diff
@@ -0,0 +1,2 @@
+// NOTE: This string will be replaced in the a `prepare` run script by the `scripts/version.js` file.
```
montezume: can we fix this typo before we put it all over the place?

packages/application-shell/src/version.js
```diff
@@ -0,0 +1,2 @@
+// NOTE: This string will be replaced in the a `prepare` run script by the `scripts/version.js` file.
+export default '__@APPLICATION_KIT_PACKAGE/VERSION_OF_RELEASE__';
```
tdeekens: Erm...

packages/sdk/src/version.js
emmenko: Which typo? I copied it from the other PR...

packages/application-shell/src/configure-apollo.js
emmenko: Hmm, you mean to put here the app name?

packages/sdk/src/version.js
tdeekens: I think he just based off my PR.

montezume:
```suggestion
// NOTE: This string will be replaced in the `prepare` script by the `scripts/version.js` file.
```

packages/application-shell/src/configure-apollo.js
tdeekens: Yeah. That's what I wanted to achieve with the metrics too. Maybe a `ApplicationShell.setApplicationName` could do the trick. Then here `ApplicationShell.getApplicationName`. However then we require every custom app to set it up. Where is we could also use `readPkgUp` in case the starter (or any custom app) has a valid `name` in the pkg.json.

emmenko: We need it to define in both sdk and app-shell packages btw...

emmenko: I think we should pass the app name to the env.json, so that the bundled code can read it from `window.app.name`

emmenko: And we can enforce the `env.json` to contain the `name` field, we already validate other fields

https://github.com/commercetools/merchant-center-application-kit/blob/6fb7ec398fbcb723fdc76211066e0595c9a45593/packages/mc-html-template/lib/load-env.js#L15-L21

tdeekens: That's a nice idea. Would be another breaking release for app-kit I suppose? We could also default it to `unknown`.

emmenko: See https://github.com/commercetools/merchant-center-application-kit/pull/694/commits/9c6873be2f8f3c7fceda45633ff4f9f8510a31d2

WDYT?

emmenko: I would keep it optional for now, so we can test it a bit. Then we can enforce it and cut a breaking release.

application-templates/starter/env.json
```diff
@@ -1,4 +1,5 @@
 {
+  "applicationName": "merchant-center-application-template-starter",
```
tdeekens: Jackpot. Would have named it the same :)

playground/README.md
```diff
@@ -40,6 +40,7 @@ A project for developing a Merchant Center Application usually consists of the f
   - `routes.js` contains the sub-routes and components rendered by the application (the main route is defined in the `<EntryPoint>` and is loaded asynchronously using code splitting)
 - `dist` contains the production bundles (this is created once you run `yarn build`)
 - `env.json` contains runtime configuration available as a global state `window.app`. The object has to be passed to the `<ApplicationShell>` as `environment` prop. The object can contain any configuration specific to the application, plus the following **required** fields:
+  - `applicationName`: the name of the application (usually the same as in the `package.json`)
```
montezume:
```suggestion
  - `applicationName`: the name of the application (usually the same as in`package.json`)
```

website/content/getting-started/deployment.mdx
```diff
@@ -82,6 +82,7 @@ When you start the `mc-http-server`, you need to pass a `--config` option pointi
 
 The `env.json` file contains runtime configuration available as a global state, `window.app`. The `window.app` object is passed to the `<ApplicationShell>` component as an `environment` prop. The `window.app` object may contain any configuration specific to the application, plus the following **required** fields:
 
+- `applicationName`: the name of the application (usually the same as in the `package.json`)
```
montezume:
```suggestion
- `applicationName`: the name of the application (usually the same as in `package.json`)
```

PR #695, opened 2019-06-05
feat(app-shell): add version tracker

tdeekens:
#### Summary

This pull request adds a `VersionTracker` component (name TBD).

#### Description

This `<VersionTracker />` will be used within the `<ApplicationShell />` that means no need to configure anything for e.g. a custom app.
It uses a new field `applicationName` from the `env.json` (we can make it a breaking change or not).

### DIFF
--- a/packages/application-shell/src/components/application-shell/application-shell.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.js
@@ -20,6 +20,7 @@ import PortalsContainer from '../portals-container';
 import ApplicationShellProvider from '../application-shell-provider';
 import { getBrowserLocale } from '../application-shell-provider/utils';
 import FetchUser from '../fetch-user';
+import VersionTracker from '../version-tracker';
 import FetchProject from '../fetch-project';
 import ConfigureIntlProvider from '../configure-intl-provider';
 import AppBar from '../app-bar';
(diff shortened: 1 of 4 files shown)


packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -113,217 +114,222 @@ export const RestrictedApplication = props => (
                   appEnv={props.environment.env}
                   defaultFlags={props.defaultFeatureFlags}
                 >
-                  <React.Fragment>
-                    {/* NOTE: the requests in flight loader will render a loading
+                  <VersionTracker>
```
tdeekens: Just intentation.

packages/application-shell/src/components/version-tracker/version-tracker.js
```diff
@@ -0,0 +1,68 @@
...
+const createVersionMetric = ({ applicationName }) => ({
+  metricName: 'versions',
+  metricLabels: {
+    application: applicationName || 'unknown',
+    package_name: pkg.name,
+    package_version: pkg.version,
+  },
+});
+
+export class VersionTracker extends React.Component {
```
tdeekens: Same as the MeasureFirstPaint thing. We can use hooks (e.g. useEffect) as we need to support older versions of React.

packages/application-shell/src/components/version-tracker/version-tracker.js
```diff
@@ -0,0 +1,68 @@
...
+import { withApplicationContext } from '@commercetools-frontend/application-shell-connectors';
+import pkg from '../../../package.json';
+import * as actions from './actions';
+
+const createVersionMetric = ({ applicationName }) => ({
+  metricName: 'versions',
+  metricLabels: {
+    application: applicationName || 'unknown',
+    package_name: pkg.name,
+    package_version: pkg.version,
```
montezume: errr, don't we want to use the new version attribute instead? if not then what was the point of that PR?

tdeekens: Oh :D

packages/application-shell/src/components/version-tracker/version-tracker.js
```diff
@@ -0,0 +1,68 @@
...
+        // reloaded whilst the request was being sent or network request was interrupted
+      });
+  }
+  render() {
+    return this.props.children;
+  }
+}
+
+const mapStateToProps = () => ({
+  userAgent: window.navigator.userAgent,
```
montezume: what is this used for 🤔

tdeekens: True not needed.

packages/application-shell/src/components/version-tracker/version-tracker.js
```diff
@@ -0,0 +1,65 @@
...
+import { withApplicationContext } from '@commercetools-frontend/application-shell-connectors';
+import applicationShellVersion from '../../version';
+import pkg from '../../../package.json';
+import * as actions from './actions';
+
+const createVersionMetric = ({ applicationName }) => ({
+  metricName: 'versions',
+  metricLabels: {
+    application: applicationName || 'unknown',
+    package_name: pkg.name,
```
tdeekens: This may look a bit odd but the idea was that we could also send more versions of more packages back given they often are not updated.

packages/application-shell/src/components/application-shell/application-shell.js
emmenko: Curious as to why it renders children? Why not just `<VersionTracker />`?

packages/application-shell/src/components/version-tracker/version-tracker.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import flowRight from 'lodash/flowRight';
+import { connect } from 'react-redux';
+import { withApplicationContext } from '@commercetools-frontend/application-shell-connectors';
+import applicationShellVersion from '../../version';
+import pkg from '../../../package.json';
```
emmenko: I would not import it like this. For example, if someone were to use CRA, it would not work because of imports outside of `src`.

Since we know then name of the package, I would just hardcore it

packages/application-shell/src/components/version-tracker/version-tracker.js
```diff
@@ -0,0 +1,64 @@
...
+    pushVersionCounter: PropTypes.func.isRequired,
+  };
+  static defaultProps = {
+    /**
+     * NOTE:
+     *   In the e.g. application-account runs out of a `projectKey` context.
+     *   In order to not have `"undefined"` in our Prometheus metrics we
+     *   send an empty string whenever no `projectKey` is passed.
+     */
+    projectKey: '',
```
emmenko: Is this correct?

packages/application-shell/src/components/version-tracker/version-tracker.js
tdeekens: Sure.

packages/application-shell/src/components/version-tracker/version-tracker.js
tdeekens: No.

tdeekens:
> Curious as to why it renders children? Why not just <VersionTracker />?

Feelt like that pattern is a bit mixed around the code base. I also like it to not render children if it doesn't pass anything down.

packages/application-shell/src/components/version-tracker/version-tracker.js
tdeekens: 426ee69

packages/application-shell/src/components/version-tracker/version-tracker.js
```diff
@@ -0,0 +1,53 @@
...
+    package_name: '@commercetools-frontend/application-shell',
+    package_version: applicationShellVersion,
+  },
+});
+
+export class VersionTracker extends React.Component {
+  static displayName = 'VersionTracker';
+  static propTypes = {
+    // withApplicationContext
+    applicationName: PropTypes.string.isRequired,
```
emmenko: It’s optional at the moment

tdeekens: It's to hot. I can't think. Sorry.

tdeekens:
Huzaaah. Tested with a local build and it works. Will merge and get to the canary next.

PR #697, opened 2019-06-06
feat(app-shell): add reporting react version

tdeekens:
#### Summary

This pull request adds tracking of react's version.


montezume:
Should we also track UI-Kit versions 🤔

emmenko:
> Should we also track UI-Kit versions 

We first need to expose the version in uikit as well. But yeah, it sounds like a good idea!

PR #698, opened 2019-06-06
feat: add ui-kit to version tracker

montezume:
Same idea is #697

### DIFF
--- a/packages/application-shell/src/components/version-tracker/version-tracker.js
+++ b/packages/application-shell/src/components/version-tracker/version-tracker.js
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import flowRight from 'lodash/flowRight';
 import { connect } from 'react-redux';
+import { version as uiKitVersion } from '@commercetools-frontend/ui-kit';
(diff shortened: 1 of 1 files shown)


montezume:
@tdeekens well it's a peer dep though? so we will be sending unknown until it's updated in the apps? 

Am I missing something?

montezume:
Ah, looks like we need to change the approach.

```
/root/merchant-center-application-kit/packages/application-shell/dist/application-shell-index.es.js
Attempted import error: 'version' is not exported from '@commercetools-frontend/ui-kit' (imported as 'version$2').

```

emmenko:
Let's just wait for the next ui-kit release then.

montezume:
@emmenko but then we will crash custom apps right?

emmenko:
What do you mean "we crash custom apps"?

tdeekens:
Yes. That was my concern with this not thinking about custom apps.

emmenko:
You mean because of the peer dependency?

emmenko:
Let's maybe keep this on hold until we release the next appkit major. Then we can bump the peer dependency version.
Also, is there an ETA for the next uikit major?

montezume:
I'll close this for now 😄

PR #703, opened 2019-06-06
fix(mc-scripts): browserlist config

tdeekens:
#### Summary

This pull request attempts to solve a regression caused by postcss where it warns about lacking a `browserlistrc`-file.

image

I am not 100% sure this solves it but it is closer to what the docs suggest. Furthermore, I couldn't see the `config/browserlist` being used anywhere. Does anybody have an idea?

### DIFF
--- a/packages/mc-scripts/config/browserslist.js
+++ b/packages/mc-scripts/config/browserslist.js
@@ -1,6 +1,0 @@
-module.exports = {
-  development: ['chrome', 'firefox'].map(
-    browser => `last 2 ${browser} versions`
-  ),
-  production: ['>1%', 'not op_mini all', 'ie 11'],
-};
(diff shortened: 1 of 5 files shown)


montezume:
Hm, are you sure you don't need to make a change in the webpack build?

montezume:
I think this line has to change?

```
                  postcssPresetEnv({
                    browsers: browserslist.production,
                    autoprefixer: { grid: true },
                  }),
```

emmenko:
The file is used by the webpack configs

https://github.com/commercetools/merchant-center-application-kit/blob/1033693ddcefb9ab89f647c76633658e09fbab62/packages/mc-scripts/config/create-webpack-config-for-production.js#L23

montezume:
also by the rollup config

tdeekens:
> also by the rollup config

Thanks. I grepped for browserlist. Not browserslist.

tdeekens:
From what I can tell from browserlist readme there are multiple ways to specify the config. Still not in a cosmiconfig style.

- `.browserlistrc` has a custom format and we can't parse it
   - We don't set `NODE_ENV` to `DEVELOPMENT` browserlist defaults it to `PRODUCTION` from what I read.
- `package.json` just put a `browserslist` entry there. 

The latter feels like the most backwards compatible and minimal change. As we can just read the env we want from there and pass it down while other tools will also read it from there which they didn't with having things in `config/browserslist`.

Issue #701, opened 2019-06-06
Fix rollup warnings when building packages


montezume:
#### Summary

sentry package

```bash
cd packages/sentry
yarn build
yarn run v1.15.2
$ rimraf dist/**
$ npm run build:es && npm run build:cjs

> @commercetools-frontend/sentry@13.7.0 build:es /Users/malcolmlaing/dev/merchant-center-application-kit/packages/sentry
> cross-env NODE_ENV=production rollup -c ../../rollup.config.js  -f es ./index.js -o ./dist/sentry.es.js

node-resolve: setting options.module is deprecated, please override options.mainFields instead
node-resolve: setting options.jsnext is deprecated, please override options.mainFields instead
node-resolve: setting options.main is deprecated, please override options.mainFields instead

./index.js → ./dist/sentry.es.js...
(!) Unresolved dependencies
https://rollupjs.org/guide/en#warning-treating-module-as-external-dependency
@babel/runtime-corejs3/core-js-stable/instance/concat (imported by sentry.js, sentry-user-tracker/sentry-user-tracker.js)
created ./dist/sentry.es.js in 904ms

> @commercetools-frontend/sentry@13.7.0 build:cjs /Users/malcolmlaing/dev/merchant-center-application-kit/packages/sentry
> cross-env NODE_ENV=production rollup -c ../../rollup.config.js  -f cjs ./index.js -o ./dist/sentry.cjs.js

node-resolve: setting options.module is deprecated, please override options.mainFields instead
node-resolve: setting options.jsnext is deprecated, please override options.mainFields instead
node-resolve: setting options.main is deprecated, please override options.mainFields instead

./index.js → ./dist/sentry.cjs.js...
(!) Unresolved dependencies
https://rollupjs.org/guide/en#warning-treating-module-as-external-dependency
@babel/runtime-corejs3/core-js-stable/instance/concat (imported by sentry.js, sentry-user-tracker/sentry-user-tracker.js)
created ./dist/sentry.cjs.js in 917ms
```

application shell

```bash
Malcolms-Commercetools-MacBook-Pro:packages malcolmlaing$ cd application-shell
Malcolms-Commercetools-MacBook-Pro:application-shell malcolmlaing$ yarn build
yarn run v1.15.2
$ rimraf dist/** test-utils/**
$ npm run build:es && npm run build:cjs && npm run build:test-utils:cjs

[...]

PR #705, opened 2019-06-06
fix: define @babel/runtime-corejs3 in each (browser) package

emmenko:
Ref: https://github.com/commercetools/merchant-center-application-kit/issues/701#issuecomment-499501254

### DIFF
--- a/packages/actions-global/package.json
+++ b/packages/actions-global/package.json
@@ -35,6 +35,7 @@
     "build:es:watch": "npm run build:es -- -w"
   },
   "dependencies": {
+    "@babel/runtime-corejs3": "7.4.5",
     "@commercetools-frontend/browser-history": "13.3.5",
     "@commercetools-frontend/constants": "13.3.5",
     "@commercetools-frontend/notifications": "13.3.5",
--- a/packages/application-components/package.json
+++ b/packages/application-components/package.json
@@ -38,6 +38,7 @@
     "storybook:start": "start-storybook -p 9002 -s .storybook/assets"
   },
   "dependencies": {
+    "@babel/runtime-corejs3": "7.4.5",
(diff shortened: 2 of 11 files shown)


tdeekens:
Does this fix it when you build from a package? I saw that this package, now added everywhere, actually got hoisted and all things where there.

emmenko:
I have to release it first to see if it builds with bundlephobia

tdeekens:
Ah. Yeah, we're speaking of two different things. Yes this will fix bundlephobia. It's unlikely though it removes the warnings when bundling with rollup within a package.

emmenko:
If you check CI it seems that the warnings are gone

montezume:
Lerna build hides the warnings though. You have to try them individually

tdeekens:
> If you check CI it seems that the warnings are gone

As @montezume mentioned: e.g. `cd packages/sentry && yarn build`. I think even running lerna with a scope mutes the warnings. Maybe it's actually not a problem - not really sure.

emmenko:
Some of the packages are now built in bundlephobia (e.g. https://bundlephobia.com/result?p=@commercetools-frontend/actions-global@canary), some others still fail with a different error (e.g. https://bundlephobia.com/result?p=@commercetools-frontend/application-components@canary)

image

emmenko:
> As @montezume mentioned: e.g. cd packages/sentry && yarn build. I think even running lerna with a scope mutes the warnings. Maybe it's actually not a problem - not really sure.

Btw, I don't see any warnings even when building individual packages.

image

tdeekens:
Cool. Then it was actually caused by this too. I guess as node doesnt resolve to the workspace root for deps when building from a package directly and hence the dep is missing. If built with lerna it resolves up and doesnt show it?

PR #709, opened 2019-06-07
feat(storybook): use contexts for locale switching

montezume:
#### Summary

Same as in UI-Kit, we use the contexts addon for changing locale instead of a knob.

contexts


packages/application-components/.storybook/configs/contexts/intl-context.js
```diff
@@ -0,0 +1,77 @@
...
+  switch (locale) {
+    case 'en':
+      return 'English';
+    case 'es':
+      return 'Español';
+    case 'de':
+      return 'Deutsch';
+    case 'fr-FR':
+      return 'Français';
+    case 'zh-CN':
```
montezume: realized from the gif I needed to update zhCN to zh-CN

montezume:
Okay so if I update storybook to latest, it breaks.

emmenko:
I'm starting to regret that we updated to core-js 3, it just breaks with a lot of libraries (e.g. CRA, parcel, etc).

PR #710, opened 2019-06-07
fix(app-shell): do not re-export the version field from flopflip

emmenko:
Fixes the error

```
TypeError: Cannot set property version of [object Object] which has only a getter
```

This is caused by the proxy export of flopflip, which also contains a `version` field that conflicts with the app-shell `version` export. The reason is that the exported keys in cjs are exported as following:

```js
Object.keys(reactBroadcast).forEach(function (key) {
  Object.defineProperty(exports, key, {
    enumerable: true,
    get: function () {
      return reactBroadcast[key];
    }
  });
});
```

### DIFF
--- a/packages/application-shell/src/index.js
+++ b/packages/application-shell/src/index.js
@@ -36,4 +36,17 @@ export { default as version } from './version';
  *
  *    More information can be found here: https://github.com/commercetools/fe-chapter-notes/issues/61
  */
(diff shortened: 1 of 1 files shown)


tdeekens:
You want to release or should I?

emmenko:
I can release it, no worries

montezume:
Awesome find 😄

PR #713, opened 2019-06-07
fix(circleci): try patch for compare-url

emmenko:
Let's try something different...


.circleci/config.yml
```diff
@@ -85,8 +85,184 @@ commands:
...
+                COMMIT_FROM_JOB_NUM=$(extract_commit_from_job $VCS_TYPE $JOB_NUM)
+                # we do a similar check later on, but it needs to be here too
+                # for edge case 1.5: an existing commit pushed to a new branch
+                if [[ $COMMIT_FROM_JOB_NUM == $CIRCLE_SHA1 ]]; then
+                  JOB_NUM=$(( $JOB_NUM - 1 ))
+                  continue
+                fi
+                FOUND_BASE_COMPARE_COMMIT=true
+                break
+                # cd ~/merchant-center-application-kit
```
emmenko: I'm skipping this part below and exit from the loop right away

tdeekens:
The let's copy it into our code style :)

emmenko:
@tdeekens should we give this a try until it's fixed upstream?

emmenko:
They fixed the previous issue but created a new one: https://github.com/iynere/compare-url/issues/25

The patch I did was just to "cut some logic", it does not really fix the underlying problem, which I have few knowledge about.

PR #715, opened 2019-06-07
feat(storage): use typescript

montezume:
#### Summary

Use typescript for the `storage` package.

### DIFF
--- a/packages/sentry/tsconfig.json
+++ b/packages/sentry/tsconfig.json
@@ -1,4 +1,5 @@
 {
   "extends": "../../tsconfig.json",
-  "files": ["./src/index.ts"]
+  "files": ["./src/index.ts"],
+  "include": ["./src/globals.d.ts"]
 }
--- a/packages/storage/index.ts
+++ b/packages/storage/index.ts
@@ -1,0 +1,5 @@
+// This file exists  because we want jest to use our non-compiled code to run tests
+// if this file is missing, and you have a `module` or `main` that points to a non-existing file
+// (ie, a bundle that hasn't been built yet) then jest will fail if the bundle is not yet built.
+// all apps should export all their named exports from their root index.js
+export * from './src';
--- a/packages/storage/index.ts
+++ b/packages/storage/index.ts
@@ -1,0 +1,5 @@
+export const put = (key: string, value: string) =>
+  window.localStorage.setItem(key, value);
+export const get = (key: string) => window.localStorage.getItem(key);
+export const remove = (key: string) => window.localStorage.removeItem(key);
+export const clear = () => window.localStorage.clear();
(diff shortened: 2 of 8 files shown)


emmenko:
Hmm weird...maybe @tdeekens has an idea? Because I basically based the implementation like in flopflip...

montezume:
I copied from [flopflip](https://github.com/tdeekens/flopflip/blob/master/packages/launchdarkly-adapter/tsconfig.json) and it seems to work now

Screen Shot 2019-06-07 at 1 59 21 PM

packages/storage/index.ts
```diff
@@ -0,0 +1,5 @@
+export const put = (key: string, value: string) =>
+  window.localStorage.setItem(key, value);
```
montezume: hey ma, I used static types 🙈

montezume:
@tdeekens is on the way to the airport I think

emmenko:
I just built it from your branch and I see this:

image

montezume:
@emmenko yeah, isn't that the intended result?

packages/storage/tsconfig.json
```diff
@@ -1,5 +1,4 @@
 {
   "extends": "../../tsconfig.json",
-  "files": ["./index.ts"],
-  "include": []
+  "files": ["./src/index.ts"]
```
montezume: did you try after removing include: []? that's the fix i got from flopflip.

emmenko:
Yeah sorry, I didn't see that you updated your comment. I was referring to your question when you opened the PR.

Also, I pushed a commit to expose the version, apparently I missed that last time out.

packages/storage/tsconfig.json
emmenko: Hmm, let me try again. I removed it because the sentry package also does not have it...

emmenko: Ooh ok, then we need to add it for each `tsconfig.json` I suppose. Good catch

emmenko: Wait, now I'm getting other errors 😱

emmenko: Ok this seems to fix it: https://github.com/commercetools/merchant-center-application-kit/pull/715/commits/19b7bb5500177e88c40040973190d86d649124f1

packages/sentry/tsconfig.json
```diff
@@ -1,4 +1,5 @@
 {
   "extends": "../../tsconfig.json",
-  "files": ["./src/index.ts"]
+  "files": ["./src/index.ts"],
+  "include": ["./src/globals.d.ts"]
```
emmenko: This is important in case we have custom declaration files

image

montezume:
Thanks for the help @emmenko  😄

PR #717, opened 2019-06-07
refactor(url-utils): use typescript

montezume:
#### Summary

Refactors the `url-utils` package to use typescript.


packages/url-utils/src/url.spec.ts
```diff
@@ -24,6 +24,7 @@ describe('joinPaths', () => {
     ).toBe('http://localhost:4000/bar');
   });
   it('should throw if a path is not a string', () => {
```
tdeekens: Hm. I am not sure if we want to type specs. Is that a goal?

montezume: well TS won't help us at runtime right? this is testing that the 

```      if (typeof path !== 'string')
        throw new Error(
          `Expected path "${path}" to be a "string", but got "${typeof path}"`
        );
```

works, not that type script works?

montezume: but I am not a TS expert at all! Please let me know if it's wrong 🤔

tdeekens: Yes, but I was asking about the spec being a .ts. Anyway, all good.

montezume: oh, now I understand 🙈 

To be honest I just used ts for the specs because @emmenko did for his sentry refactor.

This is why I mentioned that we should be sure about our first refactor, because we will reuse the same patterns all over right?

emmenko: I would like to try that we also type the specs, as it helps with configuring the tests correctly. If we see that it somehow does not work as we want, we can drop them again.

emmenko: As for this particular problem, in an ideal world we don't even need to do type guards, because the compiler will check it for us to ensure that we will always pass a string.

In this case we could "trick" it by specifying the type signature as

```ts
export const joinPaths = (...paths: (string | number)[]) => {
```

So that we can still test it to do a type guard and at the same time make the compiler happy

emmenko: But I would just keep the `// @ts-ignore`, which will make the signature more correct as only expect a string.

montezume: How will the type guard protect us in the merchant centre though? (where we haven't introduced TS?)

montezume: yeah seems like a good compromise 😄

PR #718, opened 2019-06-07
feat(l10n): use typescript

montezume:
#### Summary

Refactors the `l10n` package to use typescript.

I didn't type the spec file because it seems quite daunting.


packages/l10n/src/utils.ts
```diff
@@ -1,4 +1,13 @@
-export const mapLocaleToIntlLocale = locale => {
+interface Currency {
+  label: string;
+  symbol: string;
+}
+
+interface Currencies {
+  [key: string]: Currency;
```
montezume: is this right?: 🤔

tdeekens: Yeah looks right to me.

packages/l10n/src/country-information.ts
```diff
@@ -1,8 +1,8 @@
-import PropTypes from 'prop-types';
+import * as PropTypes from 'prop-types';
 import createL10NInjector from './create-l10n-injector';
 import { getSupportedLocale, mapLocaleToIntlLocale } from './utils';
 
-const getImportChunk = locale => {
+const getImportChunk = (locale: string): Promise<any> => {
```
emmenko: We should avoid using `any`. Try

```suggestion
const getImportChunk = (locale: string): Promise<Record<string, string>> => {
```

packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
```diff
@@ -2,14 +2,23 @@ import React from 'react';
 import { reportErrorToSentry } from '@commercetools-frontend/sentry';
 import { getDisplayName } from '../utils';
 
+interface Props {
```
tdeekens: Why interface and not `type`? To my knowledge interfaces are a little more powerful. They can be extended and things. I'd kind of use `type` for props and `interface` for objects. At least we could use `type` for props first and then make it an `interface` when it needs extending.

packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
```diff
@@ -2,14 +2,23 @@ import React from 'react';
...
   displayName,
   propKey,
   propLoadingKey,
   loadLocale,
-}) {
-  return mapPropsToLocale => WrappedComponent => {
-    class L10NComponent extends React.Component {
+}: Props) {
+  return (mapPropsToLocale: Function) => (
+    WrappedComponent: React.ComponentClass<any>
```
tdeekens: I think this should be React.ComponentType.

packages/l10n/src/country-information.ts
```diff
@@ -34,7 +34,7 @@ export const countriesShape = PropTypes.objectOf(PropTypes.string);
  * If running through webpack, code splitting makes `getCountriesForLocale`
  * a function that asynchronously loads the country data.
  */
-const getCountriesForLocale = (locale, cb) => {
+const getCountriesForLocale = (locale: string, cb: Function) => {
```
emmenko: We should specify the callback function signature

```suggestion
const getCountriesForLocale = (locale: string, cb: (error?: Error, countries: Record<string, string>) => void) => {
```

packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
```diff
@@ -2,14 +2,23 @@ import React from 'react';
 import { reportErrorToSentry } from '@commercetools-frontend/sentry';
 import { getDisplayName } from '../utils';
 
+interface Props {
+  displayName: string;
+  propKey: string;
+  propLoadingKey: string;
+  loadLocale: Function;
```
emmenko: Same here and in other places

packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
montezume: I know I am repeating myself, but I used interface because we used it in the sentry package.

https://github.com/commercetools/merchant-center-application-kit/blob/91ace37d5d7ac9334fc3a36b461d1691970f9d68/packages/sentry/src/sentry-user-tracker/sentry-user-tracker.tsx#L4

Remember,  once we merge a PR, people (me) will copy paste 😄

tdeekens: Yeah. I missed it there. Sorry.

montezume: I'm happy to change this to Props.

emmenko: The compiler suggested to prefer using interface

montezume:
Thanks for the feedback and links @emmenko I'll fix it up on Monday 😄

packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
emmenko: https://palantir.github.io/tslint/rules/interface-over-type-literal/

packages/l10n/src/utils.ts
emmenko: We can use a record: `Record<string, Currency>`

packages/l10n/src/country-information.ts
emmenko: ...or whatever the record type is (I just put string,string as an example)

packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
montezume: palantir 🤢

tdeekens: Hm. I read otherwise. I guess it's about being restrictive where we can and allow extending only where needed to restrict the "life" of the type system of the library. 🤷‍♂

emmenko: I would also prefer to use types over interfaces but I don't know what is the convention in TS though, I always see using interfaces...also in e.g. generated code (prisma, etc) they all use interfaces.

I don't know what would be the better way to go for us...

emmenko: I found this article that describes well the differences: https://link.medium.com/GW8RdH3AlX

It also recommends to use types for props etc, and most importantly to be consistent with the usage of them.
Interfaces are mostly useful to allow consumers of the library to extend the exported types.

tdeekens: Yes that‘s also how I understood it. Being restrictive with types not being extensible as long as they‘re exported and have to be extended for consumers. Thanks for the reference.

packages/l10n/src/country-information.ts
montezume: I'm not sure this works? 

https://stackoverflow.com/questions/17186566/how-do-i-fix-error-ts1015-parameter-cannot-have-question-mark-and-initializer

emmenko: The second arg should also be optional I suppose

montezume:
I can't figure out how to get typings working with `@commercetools-frontend/jest-enzyme-matchers`. Any ideas @tdeekens ? 

```
packages/l10n/src/create-l10n-injector/create-l10n-injector.spec.tsx:111:25 - error TS2339: Property 'toHaveState' does not exist on type 'Matchers<any>'.

111         expect(wrapper).toHaveState(
                            ~~~~~~~~~~~
```

packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
```diff
@@ -1,15 +1,35 @@
...
+export default function createL10NInjector<T>({
   displayName,
   propKey,
   propLoadingKey,
   loadLocale,
-}) {
-  return mapPropsToLocale => WrappedComponent => {
-    class L10NComponent extends React.Component {
+}: InjectorOptions<T>) {
+  return (mapPropsToLocale: (props: any) => string) => (
```
montezume: Any ideas @emmenko ?

packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
```diff
@@ -1,15 +1,35 @@
...
+    WrappedComponent: React.ComponentType<any>
```
montezume: really not sure what to do about the HOC react components...

packages/l10n/src/currency-information.ts
```diff
@@ -1,7 +1,16 @@
...
 import { getSupportedLocale, mapLocaleToIntlLocale } from './utils';
 
-const getImportChunk = locale => {
+type Currency = {
+  label: string;
+  symbol: string;
+};
+
+type Currencies = {
+  default: Record<string, Currency>;
```
montezume: I tried using the conditional type and it didn't work

montezume:
Okay this should be ready for reviews again.

jest.d.ts
```diff
@@ -0,0 +1,7 @@
+declare namespace jest {
+  interface Matchers<R> {
+    toHaveProp(selector: string): void;
```
emmenko: Can we specify where those custom matchers come from?

```suggestion
    // The following matchers are part of https://www.npmjs.com/package/@commercetools/jest-enzyme-matchers
    toHaveProp(selector: string): R;
```

Also, the return type should be `R` and not `void` (e.g. https://github.com/facebook/jest/blob/ccea646e76efd73654468ac32b3cb3a88778f7f4/packages/expect/src/types.ts#L87-L343)

jest.d.ts
```diff
@@ -0,0 +1,7 @@
...
+    toRender(selector: string): void;
```
emmenko:
```suggestion
    toRender(selector: string): R;
```

jest.d.ts
```diff
@@ -0,0 +1,7 @@
...
+    toHaveState(selector: string, arg2: any): void;
```
emmenko:
```suggestion
    toHaveState(selector: string, expected: unknown): R;
```

https://github.com/facebook/jest/blob/ccea646e76efd73654468ac32b3cb3a88778f7f4/packages/expect/src/types.ts#L205

packages/jest-preset-mc-app/jest-preset.js
```diff
@@ -26,7 +26,7 @@ module.exports = {
     resolveRelativePath('./setup-tests.js'),
     'jest-localstorage-mock',
   ],
-  setupFilesAfterEnv: [resolveRelativePath('./setup-test-framework.js')],
+  setupFilesAfterEnv: [resolveRelativePath('./setup-test-framework.ts')],
```
emmenko: Hmm this should stay `.js` otherwise it breaks for whoever is using the package without typescript 🤔

packages/l10n/package.json
```diff
@@ -43,6 +44,7 @@
     "react": "16.8.6"
   },
   "devDependencies": {
+    "@types/enzyme": "^3.9.3",
```
emmenko: This should be moved to the main `package.json`

packages/l10n/src/country-information.ts
```diff
@@ -34,19 +38,22 @@ export const countriesShape = PropTypes.objectOf(PropTypes.string);
...
  */
-const getCountriesForLocale = (locale, cb) => {
+const getCountriesForLocale = (
+  locale: string,
+  cb: (error?: Error, countries?: Record<string, string>) => void
+) => {
   const supportedLocale = getSupportedLocale(locale);
   // Use default webpackMode (lazy) so that we generate one file per locale.
-  // The files are named like "country-data-en-json.chunk.js" after compilation
+  // The files are named like "country-sdata-en-json.chunk.js" after compilation
```
emmenko:
```suggestion
  // The files are named like "country-data-en-json.chunk.js" after compilation
```

packages/l10n/src/country-information.ts
```diff
@@ -1,8 +1,12 @@
-import PropTypes from 'prop-types';
+import * as PropTypes from 'prop-types';
 import createL10NInjector from './create-l10n-injector';
 import { getSupportedLocale, mapLocaleToIntlLocale } from './utils';
 
-const getImportChunk = locale => {
+type Countries = {
+  default: Record<string, string>;
+};
```
emmenko: To make it less repetitive maybe we should do

```ts
type Countries = Record<string, string>;
type ImportData = { default: Countries };
```

packages/l10n/src/country-information.ts
```diff
@@ -34,19 +38,22 @@ export const countriesShape = PropTypes.objectOf(PropTypes.string);
...
-  // The files are named like "country-data-en-json.chunk.js" after compilation
+  // The files are named like "country-sdata-en-json.chunk.js" after compilation
   // https://webpack.js.org/api/module-methods/#import-
   getImportChunk(supportedLocale)
     // Prefer loading `default` (for ESM bundles) and
     // fall back to normal import (for CJS bundles).
-    .then(countries => cb(null, countries.default || countries))
-    .catch(error => cb(error));
+    .then(countries => cb(undefined, countries.default || countries))
+    .catch(error => cb(error, undefined));
```
emmenko: The second argument should not be necessary

packages/l10n/src/create-l10n-injector/create-l10n-injector.spec.tsx
```diff
@@ -1,11 +1,15 @@
...
+import { shallow, ShallowWrapper } from 'enzyme';
 import { reportErrorToSentry } from '@commercetools-frontend/sentry';
 import createL10NInjector from './create-l10n-injector';
 
 jest.mock('@commercetools-frontend/sentry');
 
-const mockData = {
+type Mock = Record<number, string>;
+
+type MockData = Record<string, Mock>;
```
emmenko: If you know the shape, you can simply do

```ts
type MockData = {
  en: Record<number, string>,
  de: Record<number, string>,
};
```

packages/l10n/src/create-l10n-injector/create-l10n-injector.spec.tsx
```diff
@@ -63,15 +67,20 @@ describe('rendering', () => {
 });
 
 describe('lifecycle', () => {
-  let wrapper;
+  let wrapper: ShallowWrapper;
   beforeAll(() => {
-    const WrappedComponent = withCandies(props => props.locale)(Foo);
+    const WrappedComponent: React.ComponentClass<any> = withCandies(
```
emmenko:
```suggestion
    const WrappedComponent: React.ComponentType<any> = withCandies(
```

packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
```diff
@@ -0,0 +1,59 @@
...
+export default function createL10NInjector<T>({
+  displayName,
+  propKey,
+  propLoadingKey,
+  loadLocale,
+}: InjectorOptions<T>) {
+  return function createHOC<U>(mapPropsToLocale: (props: any) => string) {
+    return (
+      WrappedComponent: React.ComponentType<U>
+    ): React.ComponentClass<U> => {
```
emmenko: I think this return type does not need an explicit annotation

```suggestion
    ) => {
```

packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
```diff
@@ -0,0 +1,59 @@
...
+  loadLocale: (locale: string, cb: (error?: Error, data?: T) => void) => void;
+};
+
+export default function createL10NInjector<T>({
+  displayName,
+  propKey,
+  propLoadingKey,
+  loadLocale,
+}: InjectorOptions<T>) {
+  return function createHOC<U>(mapPropsToLocale: (props: any) => string) {
```
emmenko:
```suggestion
  return function createHOC<U>(mapPropsToLocale: (props: U) => string) {
```

emmenko: PS: if it helps, we can also rename `U` as `Props`

packages/l10n/src/type-utils.ts
```diff
@@ -0,0 +1,12 @@
...
+// https://dev.to/busypeoples/notes-on-typescript-inferring-react-proptypes-1g88
+export type InferPropTypes<
+  PropTypes,
+  DefaultProps = {},
+  Props = PropTypes.InferProps<PropTypes>
+> = {
+  [Key in keyof Props]: Key extends keyof DefaultProps
+    ? Props[Key] | DefaultProps[Key]
+    : Props[Key];
+};
```
emmenko: This can be removed as we don't use it here

packages/l10n/src/utils.ts
```diff
@@ -20,15 +27,18 @@ export function getSupportedLocale(locale) {
...
+) => {
   if (currencies[currencyCode] && currencies[currencyCode].symbol) {
     return currencies[currencyCode].symbol;
   }
   return '';
 };
 
 // From https://github.com/acdlite/recompose/blob/master/src/packages/recompose/getDisplayName.js
-export const getDisplayName = Component => {
+export const getDisplayName = (Component: React.ComponentType<any>) => {
```
emmenko: Is the `any` necessary or can we leave it out?

packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
montezume: Screen Shot 2019-06-12 at 8 37 38 AM
compiler says no 😢

packages/l10n/src/utils.ts
montezume:
```
packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx:24:11 - error TS2345: Argument of type 'ComponentType<Props>' is not assignable to parameter of type 'ComponentType<{}>'.
  Type 'ComponentClass<Props, any>' is not assignable to type 'ComponentType<{}>'.
    Type 'ComponentClass<Props, any>' is not assignable to type 'ComponentClass<{}, any>'.
      Types of property 'getDerivedStateFromProps' are incompatible.
        Type 'GetDerivedStateFromProps<Props, any> | undefined' is not assignable to type 'GetDerivedStateFromProps<{}, any> | undefined'.
          Type 'GetDerivedStateFromProps<Props, any>' is not assignable to type 'GetDerivedStateFromProps<{}, any>'.
            Types of parameters 'nextProps' and 'nextProps' are incompatible.
              Type 'Readonly<{}>' is not assignable to type 'Readonly<Props>'.

24           WrappedComponent
             ~~~~~~~~~~~~~~~~```

Comment (montezume) [reply packages/l10n/src/utils.ts]: I think it's necessary 🤔

File: packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx (hunk 21 above)
Comment (montezume) [reply packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx]: Screen Shot 2019-06-12 at 8 49 12 AM

File: packages/l10n/src/utils.ts (hunk 23 above)
Comment (tdeekens) [reply packages/l10n/src/utils.ts]: I don't think `any` is correct here. You actually don't want to allow to accept any from my view. I'd say `Props extends object` in these cases. At least that's what I ended up doing at the flopper.

File: packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx (hunk 21 above)
Comment (montezume) [reply packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx]: I guess it makes sense because they can set their locale prop as anything 🤔

Comment (tdeekens) [inline packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx]: I would suggest giving types not names like `T` and `U`. To me it reminds me of Uni when we named our variables `var a,b,c,d`. I'd call it something like `PassedProps` and `AugmentedProps` or `InjectedProps`. Maybe just me though :)

Comment (tdeekens) [inline packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx]: Props is not any I think?

File: packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx (hunk 20 above)
Comment (emmenko) [reply packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx]: Can you try returning the class directly?

```js
return (WrappedComponent: React.ComponentType<U>) =>
  class L10NComponent extends React.Component<U> {}
```

...instead of `return L10NComponent`

packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
montezume: Screen Shot 2019-06-12 at 8 49 12 AM

emmenko: That's why you need to specify the props shape

```ts
type TestProps = {
  locale: string
};

const WrappedComponent = l10nInjector<TestProps>(props => props.locale)(Foo);
```

packages/l10n/src/utils.ts
emmenko: Hmm "object" is also not "safe".

I think we need to parametrize the `getDisplayName` function as well, and pass the generic type to it when we call it

```ts
export function getDisplayName<T>(Component: React.ComponentType<T>) {}
```

packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
emmenko: Yeah, I suggested that as well. For me personally is fine but it would be easier to read if we name them properly.

PS: it was me that said to name those T,U,... in the first place 😇

packages/l10n/src/create-l10n-injector/create-l10n-injector.spec.tsx
```diff
@@ -153,7 +153,7 @@ describe('when there is an error loading L10n data', () => {
       propLoadingKey: 'l10nInjector',
       loadLocale: loadLocalesErrorMock,
     });
-    const WrappedComponent: React.ComponentClass<any> = l10nInjector(
+    const WrappedComponent: React.ComponentType<any> = l10nInjector(
```
tdeekens: I am still never sure when to say `any` or if that's a good idea. I read quite a few articles back then and felt having `Props extends object` can't hurt. Also if you check the typing of React and what it defaults to it's a `{}`. Props extends object

emmenko: Yeah, but if possible we should always opt to having proper types so that we can make sure the compile has our back.

packages/l10n/src/create-l10n-injector/create-l10n-injector.spec.tsx
montezume: hm, doesn't that break this line though?

```
const loadLocalesMock = jest.fn((locale, cb) => cb(null, mockData[locale]));
```

packages/l10n/src/utils.ts
tdeekens: Not sure. Now I could type `T` to a number. That's what felt odd to me back then and why I didn't allow having the consumer pass the generic.

packages/l10n/src/country-information.ts
emmenko: 👆Can you also refactor those things?

packages/l10n/src/country-information.ts
emmenko:
```suggestion
    .catch(error => cb(error));
```

packages/l10n/src/create-l10n-injector/create-l10n-injector.spec.tsx
emmenko: Hmm how would it break it? 🤔

packages/l10n/src/create-l10n-injector/create-l10n-injector.spec.tsx
```diff
@@ -30,11 +38,11 @@ const Foo = () => <div />;
 Foo.displayName = 'Foo';
 
 describe('rendering', () => {
-  let WrappedComponent;
-  let wrapper;
+  let WrappedComponent: React.ComponentType<any>;
```
emmenko: Let's avoid `any`

packages/l10n/src/create-l10n-injector/create-l10n-injector.spec.tsx
```diff
@@ -138,9 +157,14 @@ describe('when there is an error loading L10n data', () => {
       propLoadingKey: 'l10nInjector',
       loadLocale: loadLocalesErrorMock,
     });
-    const WrappedComponent = l10nInjector(props => props.locale)(Foo);
+    const WrappedComponent: React.ComponentType<any> = l10nInjector<TestProps>(
```
emmenko: No `any`

emmenko: I also don't think we need to explicitly annotate the variable.

packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import { reportErrorToSentry } from '@commercetools-frontend/sentry';
+import { getDisplayName } from '../utils';
+
+type InjectorOptions<T> = {
```
emmenko: If you want, we can rename `T` to `LoadedData` or something similar

packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
```diff
@@ -0,0 +1,58 @@
...
+  propKey,
+  propLoadingKey,
+  loadLocale,
+}: InjectorOptions<T>) {
+  return function createHOC<Props>(mapPropsToLocale: (props: Props) => string) {
+    return (
+      WrappedComponent: React.ComponentType<Props>
+    ): React.ComponentClass<Props> => {
+      return class L10NComponent extends React.Component<Props> {
+        static displayName = `${displayName}(${getDisplayName(
```
emmenko:
```suggestion
        static displayName = `${displayName}(${getDisplayName<Props>(
```

packages/l10n/src/utils.ts
```diff
@@ -1,4 +1,11 @@
-export const mapLocaleToIntlLocale = locale => {
+type Currency = {
+  label: string;
+  symbol: string;
+};
+
+type Currencies = Record<string, Currency>;
```
emmenko: We should be able to import them from the other file.

Or we move those types (currency, countries, etc) into a shared `types.ts`

packages/l10n/src/utils.ts
emmenko: Sure, although it's an internal function.

What you're proposing is instead something like this?

```ts
function getDisplayName<Props extends object>(Component: React.ComponentType<Props>) {}
```

packages/l10n/src/time-zone-information.ts
```diff
@@ -1,16 +1,13 @@
...
 import createL10NInjector from './create-l10n-injector';
 import { getSupportedLocale, mapLocaleToIntlLocale } from './utils';
+import { TimeZones } from './types';
 
-type Timezone = {
-  name: string;
-  abbr: string;
-  offset: string;
+type ImportData = {
+  default: TimeZones;
```
montezume: 👍

packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
```diff
@@ -18,7 +18,9 @@ export default function createL10NInjector<LoadedData>({
   propLoadingKey,
   loadLocale,
 }: InjectorOptions<LoadedData>) {
-  return function createHOC<Props>(mapPropsToLocale: (props: Props) => string) {
+  return function createHOC<Props extends object>(
```
tdeekens: BTW: I think `Props extends {}` works too. Not that it matters :)

montezume:
Much thanks to @emmenko

PR #719, opened 2019-06-07
feat: generate proptypes from typescript

montezume:
#### Summary

We can generate proptypes from typescript.

This is the output now

```js
SentryUserTracker.propTypes = {
  user: _pt.any
};
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,7 @@
     "babel-eslint": "10.0.1",
     "babel-plugin-import-graphql": "2.7.0",
     "babel-plugin-transform-rename-import": "2.3.0",
+    "babel-plugin-typescript-to-proptypes": "0.17.1",
     "cross-env": "5.2.0",
     "cypress": "3.3.1",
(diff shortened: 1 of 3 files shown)


montezume:
@tdeekens don't mind if I do 😄

emmenko:
Why is it typed as `any`? 🤔

montezume:
@emmenko see this issue https://github.com/milesj/babel-plugin-typescript-to-proptypes/issues/8

emmenko:
Can you try using `type` instead of `interface`?

montezume:
Same result 🤔

montezume:
I'm not sure there currently exists a solution that allows for generating proptypes from imported typescript types 😞. Perhaps we will have to make one ourselves.

emmenko:
I would suggest we try the other solution proposed in the article I posted on Slack

emmenko:
I opened a separate PR to try the other approach: #720

montezume:
Closing in favour of #720 😄

PR #720, opened 2019-06-08
refactor: to prefer type literal instead of interface, to infer type from prop-types

emmenko:
To expose both type declarations and prop-types, we tried #719 using a babel plugin but unfortunately it currently has some limitations (https://github.com/commercetools/merchant-center-application-kit/pull/719#issuecomment-499959171).

This is another approach that we can use which works the other way around: given prop-types, we infer TS types from them.

https://dev.to/busypeoples/notes-on-typescript-inferring-react-proptypes-1g88

### DIFF
--- a/packages/sentry/src/sentry-user-tracker/sentry-user-tracker.tsx
+++ b/packages/sentry/src/sentry-user-tracker/sentry-user-tracker.tsx
@@ -1,17 +1,22 @@
 import React from 'react';
+import * as PropTypes from 'prop-types';
+import { InferPropTypes } from '../type-utils';
 import * as sentry from '../sentry';
 
-interface Props {
-  user?: sentry.User;
-}
+const userTrackerPropTypes = {
+  user: PropTypes.shape(sentry.userPropTypes),
+};
+
+type UserTrackerProps = InferPropTypes<typeof userTrackerPropTypes>;
 
 /**
  * This component will let sentry know if any information about the user has
  * changed.
  */
 
-class SentryUserTracker extends React.PureComponent<Props> {
+class SentryUserTracker extends React.PureComponent<UserTrackerProps> {
   static displayName = 'SentryUserTracker';
+  static propTypes = userTrackerPropTypes;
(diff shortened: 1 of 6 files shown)


tdeekens:
Cool. Feels a bit odd to generate compile time types from run time types 😄. What specifically is the problem with generating prop-types from ts-types again? Just that we use `interfaces` instead of `props` or is it generally that everything becomes `Proptypes.any` and it's a known issue with the library?

montezume:
@tdeekens the TL:DR is the babel plugin only really works with types declared in that file, not imported types.

tdeekens:
Then lets not use it :)

packages/sentry/src/sentry-user-tracker/sentry-user-tracker.tsx
```diff
@@ -1,17 +1,22 @@
 import React from 'react';
+import * as PropTypes from 'prop-types';
+import { InferPropTypes } from '../type-utils';
 import * as sentry from '../sentry';
 
-interface Props {
-  user?: sentry.User;
-}
+const userTrackerPropTypes = {
+  user: PropTypes.shape(sentry.userPropTypes),
```
montezume: Hm... This feels weird though, what happens if the shape was different? For example, let's say user shape was
```js
id: PropTypes.string,
name: PropTypes.string,
department: PropTypes.shape({
 ...
})
```

montezume: At this point we would need to structure the proptypes / types pretty weird right?

emmenko: Hmm I don't get your point. Can you clarify?

montezume: AH nevermind, I understand now 🤦‍♂

emmenko:
Awesome! Do you want to approve or is there still something missing?

PR #721, opened 2019-06-09
[on hold] feat(website): support react live examples

emmenko:
https://github.com/FormidableLabs/react-live

This PR introduces the usage of react-live editor, which allows to change the example code within a "editor-like textarea" (https://github.com/satya164/react-simple-code-editor) and see the changes reflected in "real time".

The idea with this is to provide full control of the example code, to let the users "play around" with it.
It's also an attempt to simplify reviews from the design team, in case the standard example (provided by us) is not enough to test the component.


montezume:
Would it make sense to code split on routes that include live editors? I imagine the bundle size gets bloated from it 🤔

emmenko:
Every route is a different static page per-se, I wouldn't worry much about it for now. The idea is also that the interactive examples are rendered with react-live.

montezume:
Did you talk with @Shecki and show him the live site? Does it solve his issues with reviewing and allow him to put in ugly data?

Shecki:
Thank for mention me.
The gif looks good. but would have to see it live and test it.
I would also suggest @filippobocik to have a  look, as he does a lot of reviews as well.

emmenko:
That's the purpose of this PR, to try out this approach and see if it improve things. That's why is marked as technical exploration.

You can try it out from the deploy link: https://deploy-preview-721--mc-appkit.netlify.com/

Shecki:
Is it possible that the Dialog could stay open and i see live feedback when i change the text?
Right, it closes and i have to reopen it again.

emmenko:
Putting this on hold as we want to check first if/how we can deploy storybook as well.

emmenko:
Closing as we solved the issue in #733

PR #730, opened 2019-06-11
chore: use new npm packages @testing-library/*

emmenko:
**DO NOT MERGE YET**

Since the peer dep changed, we should probably mark this as a breaking change. Let's maybe wait for other breaking changes before merging this.


montezume:
I thought there was a reason why we weren't updating on MC-FE? Because of the breaking change with query now returning an error if there's multiple returns?

tdeekens:
Yeah that‘s why I skipped the update. I felt like the 6-7 changes would break the MC with selectors throwing so we might want to migrate there first?

emmenko:
> Because of the breaking change with query now returning an error if there's multiple returns?

Hmm I wasn't aware of this...can you explain what the problem is?

> It might make more sense to refactor those tests on MC-FE first?

What about the RTL peer dependencies of appkit? Otherwise you end up having 2 versions of RTL, which we should avoid...

tdeekens:
> Hmm I wasn't aware of this...can you explain what the problem is?

As we return rtl from app-kit we would use it in the MC from my knowledge. rtl@7 updates dom-testing-library to v4 which has these. https://github.com/testing-library/dom-testing-library/releases/tag/v4.0.0

emmenko:
Ah ok, then let's close this. Sorry.

PR #731, opened 2019-06-11
revert(storybook): contexts addon, since it's missing the core-js dependency

emmenko:
Reverts #709 

TL;DR; the contexts package does not have an explicit dependency to `core-js`, like the other addons, which causes all sorts of problems and it's not possible to fix with e.g. yarn resolutions/nohoist.


emmenko:
I'll open an issue or a PR in their repo so maybe it will be fixed soon

PR #733, opened 2019-06-11
refactor(website): to use custom knobs to control component props

emmenko:
In order to provide a similar experience of testing components in storybook, but keeping the content and design of the website, we decided to render some input elements to control some of the component props.

A previous attempt was done in #721

### DIFF
--- a/packages/application-components/src/components/internals/for-docs/example-wrapper/example-wrapper.js
+++ b/packages/application-components/src/components/internals/for-docs/example-wrapper/example-wrapper.js
@@ -1,43 +1,199 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { Formik } from 'formik';
 import { ThemeProvider } from 'emotion-theming';
+import styled from '@emotion/styled';
 import { IntlProvider } from 'react-intl';
-import { customProperties } from '@commercetools-frontend/ui-kit';
-import ModalController from '../modal-controller';
+import {
+  Spacings,
+  TextField,
+  MultilineTextField,
+  SelectField,
+  customProperties,
+  Tooltip,
+  IconButton,
+  CodeViewIcon,
+} from '@commercetools-frontend/ui-kit';
+import { InfoDialog } from '@commercetools-frontend/application-components';
 
-const ExampleWrapper = props => (
-  <ThemeProvider
-    theme={{
-      // Reset theme to default styles, so that the example uses the default uikit tokens
-      fontFamilyDefault: customProperties.fontFamilyDefault,
-      colorSolid: customProperties.colorsSolid,
-    }}
-  >
-    <IntlProvider locale="en">
-      <div
-        style={{
-          position: 'relative',
-          width: '100%',
-          height: '400px',
-          border: `1px solid ${customProperties.colorNeutral95}`,
-        }}
-      >
-        <div id={props.containerId} style={{ flex: 1 }} />
-        <ModalController
-          title={props.controllerTitle}
-          buttonLabel={props.controllerButtonLabel}
-        >
-          {props.children}
-        </ModalController>
-      </div>
-    </IntlProvider>
-  </ThemeProvider>
-);
+const PreviewContainer = styled.div`
+  position: relative;
+  width: 100%;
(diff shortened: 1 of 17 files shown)


montezume:
This is cool for these components but I can't help but wonder why we don't just deploy storybook as well. Having to reinvent most of it's functionality just feels odd.

packages/application-components/src/components/internals/for-docs/example-wrapper/example-wrapper.js
```diff
@@ -1,43 +1,199 @@
...
+                      touched={formikProps.touched[knobConfig.name]}
+                      onChange={formikProps.handleChange}
+                      onBlur={formikProps.handleBlur}
+                      onFocus={formikProps.handleFocus}
+                    />
+                  );
+                case 'select':
+                  return (
+                    <SelectField
+                      key={knobConfig.name}
```
montezume: nottether

try tethering it?

montezume: nottether

try tethering it?

emmenko:
Deploying storybook alongside gatsby is not at all easy. And since the knobs feature is basically the only "missing feature" from storybook, I'd rather just implement that (which is actually quite simple).

packages/application-components/src/components/internals/for-docs/example-wrapper/example-wrapper.js
emmenko: Should I set the `menuPortalTarget` thingy?

montezume: Yeah. Might be weird though because we have two modals?

packages/application-components/src/components/internals/for-docs/modal-controller/modal-controller.js
```diff
@@ -9,30 +9,34 @@ import {
...
+      >
+        <Spacings.Stack>
+          <Text.Body>{props.title}</Text.Body>
+          <Spacings.Inline>
+            <SecondaryButton
+              label={props.buttonLabel}
+              onClick={() => toggle(true)}
+            />
+          </Spacings.Inline>
+          {props.children({ isOpen, toggle })}
```
tdeekens: Nit: `setIsOpen`

packages/application-components/src/components/internals/for-docs/modal-controller/modal-controller.js
```diff
@@ -9,30 +9,34 @@ import {
...
-            label={props.buttonLabel}
-            onClick={() => toggle(true)}
-          />
-        </Spacings.Inline>
-        {props.children({ isOpen, toggle })}
-      </Spacings.Stack>
-    </div>
+    <>
+      <div id={props.containerId} style={{ flex: 1 }} />
+      <div
```
tdeekens: `styled.div`?

packages/application-components/src/components/internals/for-docs/example-wrapper/example-wrapper.js
emmenko: Sorry was that a yes or no? 🤔

montezume: Yes. But instead of using document.body, you probably need to use the previous modal id

emmenko: Yes sure, I'll see what I can do then...

filippobocik:
Love it! Other than setting the base font-size to 13px I have nothing to add... Thanks @emmenko 
... @Shecki have a look as well please

packages/application-components/src/components/internals/for-docs/example-wrapper/example-wrapper.js
emmenko: @montezume Is that normal like this?

emmenko: Alright we decided to keep it as before with the overflow/scroll

PR #736, opened 2019-06-12
fix(build): use `overrideBrowsersList` to comply with new autoprefixer API

adnasa:
Let's kill that `autoprefixer` warning :)

### DIFF
--- a/packages/mc-scripts/config/create-webpack-config-for-development.js
+++ b/packages/mc-scripts/config/create-webpack-config-for-development.js
@@ -227,8 +227,10 @@ module.exports = ({
               plugins: () => [
                 postcssImport({ path: sourceFolders }),
                 postcssPresetEnv({
-                  browsers: browserslist.development,
-                  autoprefixer: { grid: true },
+                  autoprefixer: {
+                    grid: true,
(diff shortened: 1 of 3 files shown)


packages/mc-scripts/config/create-webpack-config-for-development.js
```diff
@@ -227,8 +227,10 @@ module.exports = ({
               plugins: () => [
                 postcssImport({ path: sourceFolders }),
                 postcssPresetEnv({
-                  browsers: browserslist.development,
-                  autoprefixer: { grid: true },
+                  autoprefixer: {
+                    grid: true,
+                    overrideBrowserslist: browserslist.development,
```
tdeekens: Nice. What does it use by default?

emmenko: image

(see last line)

> If the above methods did not produce a valid result Browserslist will use defaults: `> 0.5%, last 2 versions, Firefox ESR, not dead`

tdeekens: Confusing though as your package.json has the mentioned entry. Refactored from a custom config.

PR #737, opened 2019-06-12
refactor: migrate application-shell-connectors to TypeScript

emmenko:
This PR migrates the application-shell-connectors pacakge to TypeScript.
It also contains several improvements and fixes for the previous migrated packages (see inline comments for a better understanding).

Additionally, I wanted to revisit the babel plugin to generate the prop-types, as it would be the preferred choice. This was really helpful for the app-context components, given all the complex TS types. I also realised that (at least for now) you would probably define the TS types in the same file as the component, which makes the plugin work.


.eslintrc.yaml
```diff
@@ -67,6 +67,7 @@ overrides:
         [2, { accessibility: 'no-public' }]
       '@typescript-eslint/no-require-imports': 0
       '@typescript-eslint/promise-function-async': 0
+      react/prop-types: 0
```
emmenko: For TS files we now don't declare prop-types (they are generated using the babel plugin). Therefore, we need to turn off this eslint rule.

packages/application-shell-connectors/src/components/application-context/application-context.spec.tsx
```diff
@@ -43,6 +45,8 @@ describe('rendering', () => {
             env: 'development',
             cdnUrl: 'http://localhost:3001',
             servedByProxy: false,
+            // extra props
+            foo: 'bar',
```
emmenko: Just to test that the `AdditionalEnvironmentProps` works

image

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -0,0 +1,200 @@
+import React from 'react';
+import moment from 'moment-timezone';
+import getDisplayName from '../../utils/get-display-name';
+
+type AdditionalProperties = { [key: string]: unknown };
+type TRawUser = {
```
emmenko: To avoid naming conflicts, I decided to prefix most of the types here with `T`.

emmenko: Also, the "raw" user is the one we get from the API. Hopefully the name makes sense

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -0,0 +1,200 @@
...
+  id: string;
+  email: string;
+  firstName: string;
+  lastName: string;
+  language: string;
+  timeZone?: string;
+} & AdditionalProperties;
```
emmenko: This is to ensure that the `props.user` can contain additional properties, but we are only interesting in the ones defined here above

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -0,0 +1,217 @@
...
+type TApplicationContextEnvironment = {
+  applicationName?: string;
+  frontendHost: string;
+  mcApiUrl: string;
+  location: string;
+  env: string;
+  cdnUrl: string;
+  servedByProxy: string | boolean;
+};
+type TApplicationContext<ExtendedEnvironment extends {}> = {
```
emmenko: You'll see a lot of generics `ExtendedEnvironment extends {}`.
The idea is that you can specify the additional environment props that are specific to your application, which are then applied to the `context.environment`.

At least I hope this is the correct way to do that.

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -0,0 +1,217 @@
...
+  environment: ExtendedEnvironment & TApplicationContextEnvironment;
```
emmenko: This is where we "merge" the types

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -0,0 +1,217 @@
...
+  project: TApplicationContextProject | null;
+  permissions: TApplicationContextPermissions | null;
+  dataLocale: string | null;
+};
+type ProviderProps<ExtendedEnvironment extends {}> = {
+  environment: ExtendedEnvironment & TApplicationContextEnvironment;
+  user?: TRawUser;
+  project?: TRawProject;
+  projectDataLocale?: string;
+  children: React.ReactNode;
```
emmenko: I found this article with some dos and donts about TS, and the guy mentioned to explicitly type the children props.

https://medium.com/@martin_hotell/10-typescript-pro-tips-patterns-with-or-without-react-5799488d6680#55bb

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -0,0 +1,217 @@
...
+};
+type ConsumerProps<ExtendedEnvironment extends {}> = {
+  render: (
+    context: TApplicationContext<ExtendedEnvironment> | {}
+  ) => React.ReactNode;
+  children?: never;
```
emmenko: Also from the article (https://medium.com/@martin_hotell/10-typescript-pro-tips-patterns-with-or-without-react-5799488d6680#55bb).

> if your component doesn't use it, prevent it's usage with `never` type.

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -0,0 +1,217 @@
...
+};
+type DefaultMappedProps<ExtendedEnvironment extends {}> = {
+  applicationContext: TApplicationContext<ExtendedEnvironment> | {};
+};
+
+const Context = React.createContext<TApplicationContext<{}> | {}>({});
```
emmenko: This might look a bit weird but basically we need to account for the "initial" value of the context, which is an empty object.

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -0,0 +1,217 @@
...
+      props.environment,
+      props.user,
+      props.project,
+      props.projectDataLocale
+    )}
+  >
+    {props.children}
+  </Context.Provider>
+);
+ApplicationContextProvider.displayName = 'ApplicationContextProvider';
```
emmenko: The generated prop-types are:

image

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+const ApplicationContext = <ExtendedEnvironment extends {}>(
+  props: ConsumerProps<ExtendedEnvironment>
+) => (
+  <Context.Consumer>
+    {applicationContext => props.render(applicationContext)}
+  </Context.Consumer>
+);
+ApplicationContext.displayName = 'ApplicationContext';
```
emmenko: ...and

image

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+function withApplicationContext<
+  Props extends {},
```
emmenko: To clarify: these are the props passed to the wrapped component

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -0,0 +1,217 @@
...
+  MappedProps extends {} = {},
```
emmenko: ...this are the props from the `mapApplicationContextToProps`, in case it's used

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -0,0 +1,217 @@
...
+  ExtendedEnvironment extends {} = {}
```
emmenko: ...and we still need to account for the extended environment props

packages/application-shell-connectors/src/components/application-context/application-context.tsx
emmenko: PS: we can also name it something like `AdditionalEnvironmentProperties` if it makes it clearer

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -0,0 +1,217 @@
...
+>(
+  mapApplicationContextToProps?: (
+    context: TApplicationContext<ExtendedEnvironment> | {}
+  ) => MappedProps
+) {
+  return (
+    Component: React.ComponentType<
+      | Props
+      | Props & DefaultMappedProps<ExtendedEnvironment>
+      | Props & MappedProps
```
emmenko: The underlying component that we wrap with the HOC, will receive the additional props that the HOC injects, which means that we should specify the different cases:
- if we pass the map function, we expect the `MappedProps` to be spread
- if we don't pass the map function, we expect the context to be passed as a `applicationContext` prop

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -0,0 +1,217 @@
...
+    WrappedComponent.displayName = `withApplicationContext(${getDisplayName<
+      Props
+    >(Component)})`;
+    return WrappedComponent;
+  };
+}
+
+// Forward-compatibility with React Hooks 🎉
+const useApplicationContext = React.useContext
+  ? <SelectedContext extends {} = {}, ExtendedEnvironment extends {} = {}>(
```
emmenko: The hook signature is similar to the HOC, but I named the map function `selector` because it does not really inject props and simply return a value. React redux also uses those names.

packages/application-shell-connectors/src/components/project-extension-image-regex/project-extension-image-regex.spec.tsx
```diff
@@ -12,7 +12,7 @@ describe('rendering', () => {
           <ProjectExtensionProviderForImageRegex>
             <div />
           </ProjectExtensionProviderForImageRegex>
-        ).renderProp('children', { loading: true, data: undefined });
+        ).renderProp('children')({ loading: true, data: undefined });
```
emmenko: You'll see a lot of these changes. It was necessary because TS was pointing to the `renderProp` method from `enzyme` (not our own package), which has a slightly different API.
As a consequence, I refactored those and removed our extension package.

packages/application-shell-connectors/src/components/project-extension-image-regex/project-extension-image-regex.tsx
```diff
@@ -0,0 +1,82 @@
...
+type ConsumerProps = {
+  render: (imageRegex: TImageRegexContext | {}) => React.ReactNode;
+  children?: never;
+};
+
+const Context = React.createContext<TImageRegexContext | {}>({});
+
+const ProjectExtensionProviderForImageRegex = (props: ProviderProps) => (
+  <Query<
+    { projectExtension?: { imageRegex?: TImageRegex } },
```
emmenko: I felt it was ok to just inline those types.

packages/application-shell-connectors/src/graphql.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module '*.graphql' {
```
emmenko: This is to tell TS how to import `*.graphql` files.

However, it seems to only work in this location, I tried to put this file in the root folder but then it doesn't work 🤷‍♂️

packages/jest-preset-mc-app/package.json
```diff
@@ -23,7 +23,6 @@
   "dependencies": {
     "@babel/core": "7.4.5",
     "@commercetools-frontend/babel-preset-mc-app": "13.9.0",
-    "@commercetools/enzyme-extensions": "3.0.1",
```
emmenko: Bye bye 👋

emmenko: Although it might be a breaking change now that I think about it...hmm we need to find a better solution I guess 🤔

packages/l10n/package.json
```diff
@@ -31,7 +31,7 @@
   "scripts": {
     "prepare": "./../../scripts/version.js replace",
     "prebuild": "rimraf dist/**",
-    "build": "yarn build:es && yarn build:cjs",
+    "build": "yarn build:es && yarn build:cjs && yarn build:typings",
```
emmenko: This was missing

packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
```diff
@@ -18,12 +18,10 @@ export default function createL10NInjector<LoadedData>({
   propLoadingKey,
   loadLocale,
 }: InjectorOptions<LoadedData>) {
-  return function createHOC<Props extends object>(
+  return function createHOC<Props extends {}>(
     mapPropsToLocale: (props: Props) => string
   ) {
-    return (
-      WrappedComponent: React.ComponentType<Props>
-    ): React.ComponentClass<Props> => {
```
emmenko: This seems to work without the explicit annotation

packages/sentry/src/index.ts
```diff
@@ -1,6 +1,10 @@
 export { default as version } from './version';
 export * from './sentry';
-export { default as SentryUserTracker } from './sentry-user-tracker';
+export {
+  default as SentryUserTracker,
+  updateUser,
+} from './sentry-user-tracker';
```
emmenko: The changes here are so that we can define the user type in the react component file, thus getting the generated prop types.

image

tdeekens:
Oh nice. I did the permissions package yesterday and wanted to start this today as it is required for it.

packages/application-shell-connectors/src/components/application-context/application-context.tsx
tdeekens: I often postfixed typed with Type. E.g PermissionsType. I don‘t really mind which way we go as long as we all pick the same 🤣

packages/application-shell-connectors/src/components/application-context/application-context.tsx
tdeekens: Yeah. React.Node is the correct type here to what I know.

packages/application-shell-connectors/src/components/application-context/application-context.tsx
tdeekens: ...so the return of the map props?

packages/application-shell-connectors/src/components/project-extension-image-regex/project-extension-image-regex.spec.tsx
tdeekens: Will this be a breaking change?

emmenko: I’m afraid so. I tried to find a backwards compatible solution but if we can’t overload the `renderProp` signature of ShallowWrapper there is no way to make it work.

One thing that I haven’t tried is to keep the old version dependency of jest package and use yarn resolutions to point to the new version only for the appkit repo

emmenko: Seems to work with yarn resolutions: https://github.com/commercetools/merchant-center-application-kit/pull/737/commits/c674021cf0bbc17e32b28809d8118776d6165290

tdeekens: Nice. Eventually we will have to break this though. It will mostly hit the MC I think. Most custom apps are actually little to  not much tested to my knowledge.

.eslintrc.yaml
montezume: Does this apply to all only .tsx files or all files?

packages/application-shell-connectors/src/components/application-context/application-context.tsx
montezume: oh that's cool

packages/application-shell-connectors/src/components/project-extension-image-regex/project-extension-image-regex.spec.tsx
montezume: Do we still need to make these changes with the yarn resolution?

.eslintrc.yaml
emmenko: Yes, this is specific to the ts files

packages/application-shell-connectors/src/components/project-extension-image-regex/project-extension-image-regex.spec.tsx
emmenko: In the appkit repo we can use the new version, but the jest package contains the "old" version so that consumers should not need to change anything (yet).

packages/application-shell-connectors/src/components/application-context/application-context.tsx
tdeekens: At gut feeling I'd fine `AdditionalEnvironmentProperties` easier to grasp but up to you. It's really picky.

packages/application-shell-connectors/src/components/application-context/application-context.tsx
tdeekens: So none of the types in `type Props` can be in another file?

emmenko: Correct, it's a current limitation of the babel plugin.

emmenko:
Yeah, you need to get familiar with type systems in general first, especially around things like generics. But practice helps learning 😉

montezume:
@emmenko I did learn about generics... in java class... 5 years ago 🙈

PR #739, opened 2019-06-13
feat(permissions): add ts-typings

tdeekens:
#### Summary

Types the permissions package. More comments inline.


packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
-import PropTypes from 'prop-types';
+import invariant from 'tiny-invariant';
+import { TPermissions, TPermissionName } from '../../types';
```
tdeekens: We actually can't not have the same types in `types.ts` and inline them for prop-types as they won't match. At least TS complained.

packages/permissions/src/components/restricted-by-permissions/restricted-by-permissions.tsx
```diff
@@ -1,42 +1,47 @@
...
-    applicationContext: PropTypes.shape({
-      permissions: PropTypes.objectOf(PropTypes.bool).isRequired,
-    }).isRequired,
+type TRenderProp = (props: { isAuthorized: boolean }) => React.ReactNode;
+type Props = {
+  shouldMatchSomePermissions?: boolean;
+  permissions: TPermissionName[];
+  unauthorizedComponent?: React.ComponentType;
+  children: React.ReactNode | TRenderProp;
+  render: TRenderProp;
```
tdeekens: I just went with the same prefixing. Feels quite good to tell types from other things.

packages/permissions/src/components/restricted-by-permissions/restricted-by-permissions.tsx
```diff
@@ -1,42 +1,47 @@
...
+> {
+  static displayName = 'RestrictedByPermissions';
 
   render() {
+    invariant(
+      !(
+        typeof this.props.children === 'function' &&
+        !isNil(this.props.unauthorizedComponent)
+      ),
+      '@commercetools-frontend/permissions/RestrictedByPermissions: You provided both `children` and `unauthorizedComponent`. Please provide only one of them.'
```
tdeekens: You can't really do custom prop-types with ts otherwise.

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.ts
```diff
@@ -0,0 +1,36 @@
...
+}: {
+  demandedPermissions: TPermissionName[];
+  shouldMatchSomePermissions: boolean;
+}): Error | TPermissionValue => {
+  if (!React.useContext)
+    throw new Error(
+      `React hooks do not seem to be available yet in the installed React version "${React.version}". Please check the React Hooks documentation for more info: https://reactjs.org/hooks.`
+    );
+
+  const actualPermissions = useApplicationContext<SelectedApplicationContext>(
```
tdeekens: Somehow this fails. Not sure why.

packages/permissions/src/components/restricted-by-permissions/restricted-by-permissions.tsx
```diff
@@ -1,42 +1,47 @@
...
-    // Injected
-    applicationContext: PropTypes.shape({
-      permissions: PropTypes.objectOf(PropTypes.bool).isRequired,
-    }).isRequired,
+type TRenderProp = (props: { isAuthorized: boolean }) => React.ReactNode;
+type Props = {
+  shouldMatchSomePermissions?: boolean;
+  permissions: TPermissionName[];
+  unauthorizedComponent?: React.ComponentType;
+  children: React.ReactNode | TRenderProp;
```
tdeekens: This type also blows up at me and I don't fully know why. It states it's not callable. I think we may need a type guard then?

packages/permissions/src/components/restricted-by-permissions/restricted-by-permissions.tsx
emmenko: Nice, should we then have this as a convention and always prefix them with `T`?

packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -8,40 +9,29 @@ import {
...
-          `Invalid prop \`${propName}\` supplied to \`${componentName}\`. The permission(s) ${namesOfNonConfiguredPermissions.toString()} is/are not configured through \`actualPermissions\`.`
-        );
-      }
+type RenderProp = (isAuthorized: boolean) => React.ReactNode;
+type Props = {
+  shouldMatchSomePermissions?: boolean;
+  demandedPermissions: TPermissionName[];
+  actualPermissions: TPermissions;
+  render: RenderProp;
+};
```
emmenko: Let’s also add `children?: never`

emmenko: Also, we need to type the default props as well, check out this article

https://medium.com/@martin_hotell/10-typescript-pro-tips-patterns-with-or-without-react-5799488d6680

packages/permissions/src/components/restricted-by-permissions/restricted-by-permissions.tsx
emmenko: Shouldn’t both children and render be optional?

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.ts
emmenko: It’s not an object but `TPermissions` directly

packages/permissions/src/components/restricted-by-permissions/restricted-by-permissions.tsx
tdeekens: I think so right now. Only in case of overlap or when you feel fizzy would be odd. I wouldn‘t like prefixing interfaces with I cause they‘re also just types. So all just T would get my vore.

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.ts
tdeekens: Won't work either. I'll add it to the PR to proove.

packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -9,19 +9,22 @@ import {
...
+type DefaultProps = typeof defaultProps;
 type RenderProp = (isAuthorized: boolean) => React.ReactNode;
 type Props = {
   shouldMatchSomePermissions?: boolean;
   demandedPermissions: TPermissionName[];
   actualPermissions: TPermissions;
   render: RenderProp;
-};
+  children?: never;
+} & DefaultProps;
```
tdeekens: This should work.

packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -46,12 +49,15 @@ class Authorized extends React.Component<Props> {
   }
 }
 
+type InjectedProps = {
+  [propName: string]: boolean;
+};
```
tdeekens: This part I never managed to type "100%" in flopflip. You pass a `propName` of type string. Whatever is passed there would be part of the type here again. As such now the type is quite vague.

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.ts
```diff
@@ -22,7 +19,7 @@ const useIsAuthorized = ({
       `React hooks do not seem to be available yet in the installed React version "${React.version}". Please check the React Hooks documentation for more info: https://reactjs.org/hooks.`
     );
 
-  const actualPermissions = useApplicationContext<SelectedApplicationContext>(
+  const actualPermissions = useApplicationContext<TPermissions>(
```
tdeekens: Won't do the trick. Not sure why.

emmenko: I think because the generic type `SelectedContext extends {} = {}` in `useApplicationContext` is wrong, it shouldn't be only an object.

I'll open a separate PR to fix it

emmenko: Hmm actually the signature `useApplicationContext` might not be the problem. For example, if I pass `string` but return a `number`, the compiler errors:

image

emmenko: So I made some fixes/changes to the app context types (#740). With those changes, we now get the following:

image

You can also see how function overloading works in this case:

image

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -168,22 +165,20 @@ const ApplicationContext = <AdditionalEnvironmentProperties extends {}>(
 ApplicationContext.displayName = 'ApplicationContext';
 
 function withApplicationContext<
-  Props extends {},
-  MappedProps extends {} = {},
-  AdditionalEnvironmentProperties extends {} = {}
+  OwnProps extends {},
+  MappedProps extends {
+    applicationContext?: TApplicationContext<AdditionalEnvironmentProperties>;
```
tdeekens: Ah okay. So this was one type which broke in app for me.

packages/permissions/src/components/branch-on-permissions/branch-on-permissions.tsx
```diff
@@ -1,41 +1,40 @@
...
-};
-
-const branchOnPermissions = <
-  Props extends {},
-  TFallbackComponent = React.ComponentType
->(
+const branchOnPermissions = <OwnProps extends {}>(
   demandedPermissions: TPermissionName[],
-  FallbackComponent: TFallbackComponent,
+  FallbackComponent: React.ComponentType<unknown>,
```
tdeekens: Interesting.

tdeekens:
Fixed up the specs for `RestrictedByPermissions`.

packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -13,43 +13,35 @@ const defaultProps = {
...
-          this.props.actualPermissions
-        );
+  const isAuthorized = props.shouldMatchSomePermissions
+    ? hasSomePermissions(props.demandedPermissions, props.actualPermissions)
+    : hasEveryPermissions(props.demandedPermissions, props.actualPermissions);
 
-    return this.props.render(isAuthorized);
-  }
-}
+  return <React.Fragment>{props.render(isAuthorized)}</React.Fragment>;
```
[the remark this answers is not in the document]
emmenko: Otherwise TS doesn’t seem to understand it’s a react component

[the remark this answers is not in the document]
tdeekens: Pfff.

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
```diff
@@ -1,10 +1,16 @@
 import React from 'react';
-import PropTypes from 'prop-types';
-import { renderApp } from '@commercetools-frontend/application-shell/test-utils';
```
emmenko: I would like to avoid importing from app-shell from here.

tdeekens:
I don't understand why the e2e tests fail: 

https://circleci.com/gh/commercetools/merchant-center-application-kit/3730?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

packages/permissions/package.json
```diff
@@ -29,10 +31,11 @@
   "scripts": {
     "prepare": "./../../scripts/version.js replace",
     "prebuild": "rimraf dist/**",
-    "build": "yarn build:es && yarn build:cjs",
-    "build:cjs": "cross-env NODE_ENV=production rollup -c ../../rollup.config.js  -f cjs ./src/index.js -o ./dist/permissions.cjs.js",
-    "build:es": "cross-env NODE_ENV=production rollup -c ../../rollup.config.js  -f es ./src/index.js -o ./dist/permissions.es.js",
-    "build:es:watch": "yarn build:es -w"
+    "build": "yarn build:es && yarn build:cjs && yarn build:typings",
+    "build:cjs": "cross-env NODE_ENV=production rollup -c ../../rollup.config.js  -f cjs ./src/index.ts --dir ./dist --chunkFileNames permissions-[name]-[hash].cjs.js --entryFileNames permissions-[name].cjs.js",
```
emmenko: Why do we need those extra options? Maybe a bad copy/paste? 🤔

emmenko: #746

tdeekens: I was confused by that myself and took what the majority had on this PR.

PR #740, opened 2019-06-13
fix(app-shell-connectors): to cast context type to not be empty, use function overloading for useApplicationContext

emmenko:


packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -191,27 +200,47 @@ function withApplicationContext<
...
+function useApplicationContext<
+  SelectedContext,
+  AdditionalEnvironmentProperties extends {} = {}
+>(
+  selector?: (
+    context: TApplicationContext<AdditionalEnvironmentProperties>
+  ) => SelectedContext
+) {
+  // Forward-compatibility with React Hooks 🎉
+  if (!React.useContext) {
```
tdeekens: 👌🏿

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -191,27 +200,47 @@ function withApplicationContext<
...
-        `React hooks do not seem to be available yet in the installed React version "${React.version}". Please check the React Hooks documentation for more info: https://reactjs.org/hooks.`
-      );
-    };
+// Use function overloading to declare two possible signatures with two
+// distict return types, based on the selector function argument.
+function useApplicationContext<
+  AdditionalEnvironmentProperties extends {} = {}
+>(): TApplicationContext<AdditionalEnvironmentProperties>;
+function useApplicationContext<
+  SelectedContext,
```
tdeekens: I wondered if this is not the same as ReturnType typeof selector? By that consumers would not have to specify what their selector returns in many cases?

emmenko:
Hmm 🤔

```
./index.ts → dist/application-shell-connectors.es.js...
[!] Error: 'useApplicationContext' is not exported by src/components/application-context/application-context.tsx
https://rollupjs.org/guide/en#error-name-is-not-exported-by-module-
src/components/application-context/index.ts (6:2)
4:   ApplicationContextProvider,
5:   withApplicationContext,
6:   useApplicationContext,
     ^
7: } from './application-context';
Error: 'useApplicationContext' is not exported by src/components/application-context/application-context.tsx
```

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -153,7 +153,16 @@ const ApplicationContext = <AdditionalEnvironmentProperties extends {}>(
   props: ConsumerProps<AdditionalEnvironmentProperties>
 ) => (
   <Context.Consumer>
-    {applicationContext => props.render(applicationContext)}
+    {context => {
+      // Because of the way the ApplicationShell configures the Context.Provider,
+      // we ensure that, when we read from the context, we always get actual
+      // context object and not the initial value.
+      // Therefore, we can safely cast the value to be out `TApplicationContext` type.
+      const applicationContext = context as TApplicationContext<
```
tdeekens: So the alternative (not really) would be consumers using type guards to ensure it is not an empty object?

emmenko: Probably yes, because you then always get `TApplicationContext | {}` which it would never be the case given our setup.

packages/application-shell-connectors/src/components/application-context/application-context.tsx
emmenko: Hmm I'm not sure I get your point. Can you clarify what you mean?

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -191,27 +200,52 @@ function withApplicationContext<
...
+  const applicationContext = context as TApplicationContext<
+    AdditionalEnvironmentProperties
+  >;
+  return selector ? selector(applicationContext) : applicationContext;
+}
+
+// This is a workaround to trick babel/rollup to correctly export the function.
+// Most likely the problem arises with the use of overloading.
+// See related issue: https://github.com/babel/babel/issues/8361
+const useApplicationContext = useApplicationContextHook;
```
emmenko: This fixes this problem: https://github.com/commercetools/merchant-center-application-kit/pull/740#issuecomment-501858918

packages/application-shell-connectors/src/components/application-context/application-context.tsx
emmenko: Are you maybe referring to the conditional types, using `infer`?

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html

tdeekens: Sure. What I meant is that with what we pass as a generic we state what the `selector`-fn returns? We could attempt TypeScript to infer that:

```js
function getInt(a: string) {
  return parseInt(a);
}

type A = ReturnType<typeof getInt>; // => number
```

emmenko: Yes I've been looking into that and in this case it doesn't really work because the return type of the selector function can be literally _anything_

```js
useApplicationContext(context => context.permissions)
useApplicationContext(context => Object.keys(context.permissions).length > 0)
useApplicationContext(context => "foo")
```

Which means the signature would be: 

```ts
(context: TApplicationContext) => unknown
```

which is why we need a generic type

```ts
<T>(context: TApplicationContext) => T
```

But maybe for other more deterministic cases we can see if we can use that.

tdeekens: react-redux seems to do the same:

``js
export function useSelector<TState, TSelected>(
    selector: (state: TState) => TSelected,
    equalityFn?: (left: TSelected, right: TSelected) => boolean
): TSelected;
```

PR #741, opened 2019-06-14
chore: unrevert contexts for app-components storybook.

montezume:
Updated to storybook 5.1.5, and unreverted the contexts upgrade.


emmenko:
Although now that we can potentially test the components within the website, I was thinking of dropping storybook.

tdeekens:
> Although now that we can potentially test the components within the website, I was thinking of dropping storybook.

I would suggest to drop it once the website goes live.

emmenko:
> I would suggest to drop it once the website goes live.

I personally never had a need to use it, because I can just use the website to test it. Design also does not need it anymore.
@jonnybel what do you think?

If we don't need it I'd rather just drop it now (less maintenance work)

montezume:
If it's not needed for the development of new components I also say we should drop it. However, going back to this PR, is there a way in the website to test the component under different locales 🤔 ?

jonnybel:
I also don't see the need for keeping storybook anymore, since it's not deployed and we can now test the components on the website.

emmenko:
Cool, then I'll open a PR to drop storybook

>  is there a way in the website to test the component under different locales

We can add it in the "playground" mode, with the knobs

emmenko:
Here you go: https://github.com/commercetools/merchant-center-application-kit/pull/744

PR #742, opened 2019-06-14
[Feat] implement state-machines details page

TimonRey:
#### Summary
This pull request implements a details page for state machines

### DIFF
--- a/playground/src/components/state-machines-details/state-machines-details.js
+++ b/playground/src/components/state-machines-details/state-machines-details.js
@@ -1,10 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import flowRight from 'lodash/flowRight';
-import {
-  withApplicationContext,
-  Context,
-} from '@commercetools-frontend/application-shell-connectors';
+import { withApplicationContext } from '@commercetools-frontend/application-shell-connectors';
 import { connect } from 'react-redux';
 import {
   LoadingSpinner,
@@ -72,15 +69,15 @@ export class StateMachinesDetails extends React.Component {
               gridAutoColumns="1fr"
               gridTemplateColumns="repeat(2, 1fr)"
             >
-              <Text.Body>Type</Text.Body>
+              <Text.Body>{'Type'}</Text.Body>
               <Text.Body>{this.state.data.type}</Text.Body>
-              <Text.Body>Built In</Text.Body>
+              <Text.Body>{'Built In'}</Text.Body>
               <Text.Body>
-                {this.state.data.builtIn ? <CheckBoldIcon /> : null}
+                {this.state.data.builtIn ? <CheckBoldIcon /> : ''}
               </Text.Body>
-              <Text.Body>Initial</Text.Body>
+              <Text.Body>{'Initial'}</Text.Body>
               <Text.Body>
-                {this.state.data.initial ? <CheckBoldIcon /> : null}
+                {this.state.data.initial ? <CheckBoldIcon /> : ''}
               </Text.Body>
             </Grid>
           </Constraints.Horizontal>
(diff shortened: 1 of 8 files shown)


playground/src/components/state-machines-details/state-machines-details.js
```diff
@@ -0,0 +1,105 @@
...
+            </Grid>
+          </Constraints.Horizontal>
+        </Spacings.Stack>
+      </Spacings.Inset>
+    );
+  }
+}
+
+export default flowRight(
+  withApplicationContext(context => {
```
tdeekens: Nit: this could just return `context => ({...})`

playground/src/routes.js
```diff
@@ -7,15 +7,32 @@ import {
...
+        render={routerProps => (
+          <RestrictedByPermissions
+            permissions={[permissions.ViewStates, permissions.ManageStates]}
+            unauthorizedComponent={PageUnauthorized}
+            shouldMatchSomePermissions={true}
+          >
+            <StateMachinesDetails
+              id={routerProps.match.params.id}
+              projectKey={routerProps.match.params.projectKey}
+              backToListPath={props.match.url}
```
tdeekens: We often started to pass down functions for this kind of navigation in the MC. So here `goToStateMachinesList`. It ultimatively eases things a bit cause a single parent is in control of routing.

playground/src/components/state-machines-list/state-machines-list.js
```diff
@@ -38,6 +38,10 @@ const getErrorMessage = error =>
 export class StateMachinesList extends React.Component {
   static displayName = 'StateMachinesList';
   static propTypes = {
+    history: PropTypes.shape({
+      push: PropTypes.func.isRequired,
+    }).isRequired,
```
tdeekens: Given the comment below: we might not have to share the history if we just pass down a fn `goToStateMachinesList` and `goToStateMachineDetail`. Hope it makes sense.

playground/src/routes.js
emmenko: Yes but in this case we pass the string to the `LinkButton` component `to` prop, which is a string.

playground/src/components/state-machines-list/state-machines-list.js
emmenko: Right, we can use `goToStateMachineDetail` in this case. I missed that, thanks!

playground/src/components/state-machines-details/state-machines-details.js
```diff
@@ -0,0 +1,105 @@
...
+  static propTypes = {
+    id: PropTypes.string.isRequired,
+    projectKey: PropTypes.string.isRequired,
+    backToListPath: PropTypes.string.isRequired,
+    // withApplicationContext
+    dataLocale: PropTypes.string.isRequired,
+    // connect
+    fetchStateMachine: PropTypes.func.isRequired,
+  };
+  state = {
```
emmenko: FYI for the reviewers: as a "follow up step" and as an exercise , we're going to lift up the state to the redux state.

playground/src/components/state-machines-details/state-machines-details.js
```diff
@@ -0,0 +1,105 @@
...
+          <Spacings.Stack scale="xs">
+            <Text.Headline elementType="h2">
+              {this.state.data.name
+                ? this.state.data.name[this.props.dataLocale]
+                : 'n/a'}
+            </Text.Headline>
+            <Text.Detail>{this.state.data.key}</Text.Detail>
+          </Spacings.Stack>
+          <Constraints.Horizontal constraint="m">
+            <Grid
```
emmenko: FYI for the reviewers: this is just a temporary solution. As a "follow up step" and as an exercise, we're going to implement a proper form management

playground/src/components/state-machines-list/state-machines-list.js
TimonRey: 👍 
82e4aaa

playground/src/components/state-machines-details/state-machines-details.js
TimonRey: Thanks! 💯 
82e4aaa

playground/src/routes.js
tdeekens: Ok.

PR #750, opened 2019-06-14
fix(permissions): do not use shared types, keep them within the component file

emmenko:
So that we can generate proper prop-types. Yes it's a bit repetitive but the duplicated types are very simple, so I think it's worth the effort in this case.

**Before**
Screenshot 2019-06-14 at 23 13 20

**After**
Screenshot 2019-06-14 at 23 12 27


tdeekens:
Sure. I just felt from our call that we went like „meh. Then prop-types are any“.

emmenko:
Yes sure but I would still like, if possible, to have them. In this case it’s a little trade off as the types are very simple.

PR #751, opened 2019-06-15
chore: init github labels config

emmenko:
This PR simply creates an initial entry in version control about the current github labels. It uses a [new package](https://github.com/commercetools/github-labels) that we built recently to keep the labels in sync with a local config.

The idea is that the labels are managed through a config file, which can be better reviewed as it's in version control. It would also then make it easier to replicate the same config across different repositories.

### DIFF
--- a/.github-labelsrc.json
+++ b/.github-labelsrc.json
@@ -1,0 +1,170 @@
+[
+  {
+    "id": 1158981072,
+    "name": "Semver: BREAKING",
+    "color": "e3f166",
+    "description": "A breaking change"
+  },
+  {
+    "id": 1158980916,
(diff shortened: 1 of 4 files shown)


.github-labelsrc.json
```diff
@@ -0,0 +1,170 @@
+[
```
tdeekens: Nice. Would it make sense to have these published somewhere so we can just install them as a dependency and have them in sync more easily as a result?

emmenko: Hmm the problem with that is that the config needs the have the IDs of the labels for each repository.
I don’t know yet how we can solve that issue

PR #752, opened 2019-06-15
chore: update github labels config

emmenko:
Follow up of #751 

This PR updates the label config, following some new ideas:
- labels that have similar meanings and that are part of the same type, use the same color
- update colors
- add new labels

~~NOTE: I need to apply/sync the changes before merging this~~


emmenko:
This is the result now:

image

tdeekens:
Ready for 2050 with 300 customers on custom apps 😄

PR #760, opened 2019-06-17
chore(app-shell): remove org list toggle

tdeekens:
#### Summary

This removes the feature toggle for the org list. This may seem a bit quick given the release happened Thursday last week. However, the AC's views are to be turned off today too and we already rolled out 2-3 fixes.

### DIFF
--- a/packages/application-shell/src/feature-toggles.js
+++ b/packages/application-shell/src/feature-toggles.js
@@ -11,4 +11,3 @@
  */
 
 // eslint-disable-next-line import/prefer-default-export
-export const ORGANIZATIONS_LIST = 'organizationsList';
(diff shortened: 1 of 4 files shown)


packages/application-shell/src/feature-toggles.js
```diff
@@ -11,4 +11,3 @@
  */
 
 // eslint-disable-next-line import/prefer-default-export
-export const ORGANIZATIONS_LIST = 'organizationsList';
```
emmenko: This file does not seem to be used, can we remove it?

tdeekens: I'd say lets keep it in case we add feature toggles. My intend to keep it was to keep git history a bit easier to follow with when toggles got removed or added.

emmenko: Sure but even in this case it wasn't being used, as the org list toggle was injected directly.
Anyway, it's fine either way.

tdeekens: Yeah cause in quick access they're all for some strange reason.

PR #762, opened 2019-06-17
fix(i18n): fix naming bug for french and chinese translations

montezume:
#### Summary

The file committed by the transifix bot has a different name that the file we were importing. Either we do this fix, or we change the transifix bot config, but I don't know how to do that. Either is okay.

#### Before

Screen Shot 2019-06-17 at 1 10 47 PM

#### After

Screen Shot 2019-06-17 at 1 10 51 PM


emmenko:
Oh weird, in tx we have the following config:

```
filters:
  - filter_type: file
    file_format: KEYVALUEJSON
    source_language: en_US
    source_file: packages/i18n/data/core.json
    translation_files_expression: packages/i18n/data/<lang>.json
settings:
  language_mapping:
    fr_FR: fr-FR
    zh_CH: zh-CH
```

montezume:
maybe just remove the language mapping there? But either way we should remove the duplicated files, because they hide the error

PR #764, opened 2019-06-18
feat(app-shell): add user deleted logout reason

tdeekens:
#### Summary

This pull request adds redirecting a user, given his/her account does not exist, to logout with a `?reason=deleted`. Our internal app can then catch this an render an error accordingly instead of just showing a general "ice cream"-error.

### DIFF
--- a/packages/application-shell/src/components/application-shell/application-shell.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.js
@@ -43,25 +43,45 @@ import QuickAccess from '../quick-access';
  * and contains the "restricted" application part.
  */
 
+const getHasUnauthorizedError = graphQLErrors =>
+  graphQLErrors.find(
+    gqlError =>
+      gqlError.extensions &&
+      gqlError.extensions.code &&
+      gqlError.extensions.code === 'UNAUTHENTICATED'
+  );
+const getHasUserBeenDeletedError = graphQLErrors =>
+  graphQLErrors.find(
+    gqlError =>
+      gqlError.message &&
+      // NOTE: The CTP API does not provide an error code in this case.
+      gqlError.message.includes('was not found.')
+  );
 export const RestrictedApplication = props => (
   <FetchUser>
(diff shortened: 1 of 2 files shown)


packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -43,25 +43,45 @@ import QuickAccess from '../quick-access';
...
+      gqlError.extensions.code &&
+      gqlError.extensions.code === 'UNAUTHENTICATED'
+  );
+const getHasUserBeenDeletedError = graphQLErrors =>
+  graphQLErrors.find(
+    gqlError =>
+      gqlError.extensions &&
+      gqlError.extensions.message &&
+      // NOTE: The CTP API does not provide an error code in this case.
+      gqlError.extensions.message.includes('was not found.')
```
tdeekens: I don't really like this but CTP does not provide a better error yet. I also don't want to wrap CTP's API in our gateway as we don't do that in other cases (e.g. when loading an org). I think we should push CTP to provide better error codes and then come back here.

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -43,25 +43,45 @@ import QuickAccess from '../quick-access';
...
               <Redirector
                 to="logout"
                 environment={props.environment}
                 queryParams={{
-                  reason: LOGOUT_REASONS.UNAUTHORIZED,
+                  reason: (() => {
+                    if (hasUnauthorizedError)
+                      return LOGOUT_REASONS.UNAUTHORIZED;
+                    else if (hasUserBeenDeletedError) return LOGOUT_REASONS.DELETED;
+                  })(),
```
tdeekens: Just an IIFE so it's easier adding more as do-expressions are not used here yet and do not exist in TS.

tdeekens:
Yes, this will only apply once (after it's the invalid reason) and only if the account was deleted by somebody else but the user. Given the account is deleted by the user him/herself we redirect to the reason automatically.

Issue #748, opened 2019-06-14
Upcoming breaking changes (v14)


emmenko:
In the next major release I would like to go with first support for hooks, which will help us to ship less and more performant code.
Therefore, we should:
- support only react `>16.8.0`
- use hooks in the appkit packages implementations, easier to type as well
- I would also like to drop the HOC functions

We can maybe do the same in the upcoming uikit v10 and release both versions together.

Also maybe upgrade RTL.

Thoughts?

---
- [x] peer dependency react `>16.8` #795 
- [x] peer dependency react-redux `>7` #853 
- [ ] ~~(_check if it's worth it, otherwise deprecate it for now_) remove HOCs~~
- [x] (_optional_) use hooks in all appkit packages
- [x] (_optional_) use hooks for notifications #855 
- [x] (_optional_) use hooks for redux #852
- [x] (_optional_) use hooks for l10n #860 
- [x] (_optional_) upgrade to react-intl [v3](https://github.com/formatjs/react-intl/blob/master/docs/Upgrade-Guide.md) #799 #817 
- [ ] ~~(_optional_) upgrade to react-apollo~~ [v3](https://github.com/apollographql/react-apollo/tree/master/packages/hooks) #801
- [x] (_check the effort_) upgrade to latest RTL -> `@testing-library/react` #811 
- [x] upgrade ct enzyme-extensions dep in jest to latest (for the `renderProp`) #796 
- [x] (_optional_) rename assets files #834
- [x] peer dependency uikit `>10` #825
- [x] migrate uikit v10 breaking changes #825
- [x] remove IE11 specific polyfills and ponyfills #830
- [x] stop transpiling for IE 11 #836
- [x] remove vendorsToTranspile webpack option #829 
- [x] update `permissions` constants to match new permissions (e.g. `ManageProject` has been removed) #815 
- [x] deprecate `storage` package #828 
- [x] remove deprecated components: `ServicePageResponseLayout` #828

tdeekens:
I like it. Hope that people will update. However, I feel there are other reasons than bumping react why some projects do not get updated. To be fair, I don‘t think hooks for hocs will show meaningful performance improvements. It will make code easier to type and read for sure.

emmenko:
Yeah, for the performance I mean more from the React side of things. For example less wrapping components, probably react can even further optimize things internally (just a hunch from my side).
Also Babel will generate less code 🙌

montezume:
[...]

PR #766, opened 2019-06-18
feat(modal-pages): add delay before calling onClose to let animations finish

jonnybel:
* Moved the ModalPageTopBar to be part of the ModalPage, since it's always present in all of the ModalPage variations, and makes it easier to implement the next point without repeating code;

* Added a delay before calling the `onClose` function to allow the closing animation of the Modal to finish even when the ModalPage component is unmounted "abruptly" without using the `isOpen` prop. This can be turned off by the `shouldDelayOnClose` prop, which is `true` by default:
>  `shouldDelayOnClose` - Sets wether the ModalPage should delay calling its onClose function to allow the closing animation to finish. This can be turned off if the developer is controlling the ModalPage only through the `isOpen` prop, and not abruptly mounting/unmounting it or one of its parent elements.


packages/application-components/src/components/modal-pages/form-modal-page/README.md
```diff
@@ -50,6 +50,7 @@ Form Modal pages are controlled components used to render a page with a form or
...
 | `customControls`                | `node`                     |    -     | -                        | Pass a React.node to be used in place of the pre-determined controls. This can be useful if you need actions other than Cancel & Confirm, or other types of buttons, while keeping the same modal header layout                                                                    |
+| `shouldDelayOnClose`            | `bool`                     |    -     | `true`                   | Sets wether the ModalPage should delay calling its onClose function to allow the closing animation to finish. This can be turned off if the developer is controlling the ModalPage only through the `isOpen` prop, and not abruptly mounting/unmounting it or one of its parent elements. |
```
montezume:
```suggestion
| `shouldDelayOnClose`            | `bool`                     |    -     | `true`                   | Sets weather the ModalPage should delay calling its onClose function to allow the closing animation time to finish. This can be turned off if the developer is controlling the ModalPage only through the `isOpen` prop, and not abruptly mounting/unmounting it or one of its parent elements. |
```

jonnybel: 🤨
image
TIL 😄

packages/application-components/src/components/modal-pages/internals/modal-page.js
```diff
@@ -30,41 +42,182 @@ const getDefaultParentSelector = () =>
...
+    formatMessage: PropTypes.func.isRequired,
+  }),
+};
+ModalPageTopBar.defaultProps = {
+  color: 'surface',
+  previousPathLabel: messages.back,
+};
+
+const ModalPage = props => {
+  const [isOpen, setIsOpen] = React.useState(props.isOpen);
```
emmenko: Heads up: in theory the app-kit does not necessarily support hooks. We plan to do that in the v14 release (#748).
Maybe we should keep this on hold until then, if we're going to use hooks here?

jonnybel: Ah I see.
Well, I guess it's okay to put this on hold and merge on the v14 release.

The alternative would be refactoring the ModalPage into a class component to use the traditional lifecycle methods... only to refactor them back on v14, so it's not worth it IMHO.

emmenko: Yeah I think it's not a big deal at the moment, and it shouldn't stop migrating the mc-fe to the new modal pages.

emmenko: Maybe we can start a `v14` branch already, then we can point this PR to that branch

emmenko:
@jonnybel FYI: I've rebased with master in order to retrigger a percy build (#767)

packages/application-components/src/components/modal-pages/internals/modal-page.js
tdeekens: I think, for simplicity, I'd just not use hooks. Then in a consistent fashion we can refactor things to hooks in the future.

packages/application-components/src/components/modal-pages/internals/modal-page.js
```diff
@@ -30,56 +33,120 @@ const getDefaultParentSelector = () =>
...
+    }),
+  };
+  static defaultProps = {
+    level: 1,
+    baseZIndex: 1000,
+    getParentSelector: getDefaultParentSelector,
+    shouldDelayOnClose: true,
+  };
+
+  state = { forceIsOpen: null };
```
emmenko: Please double check that the controlled state logic is ok.

jonnybel:
The code seems good to go.
I've tested locally and it's working correctly ✅

jonnybel:
Alright, I've rebased to re-trigger the netlify deployment which had failed for some unknown reason, and now this can be merged.

PR #774, opened 2019-06-19
refactor(i18n): convert to TS

emmenko:


packages/i18n/src/async-locale-data/async-locale-data.spec.tsx
```diff
@@ -0,0 +1,115 @@
...
+      <>
+        <div>{`Locale: ${props.locale}`}</div>
+        <div>{`Messages: ${props.messages['CustomApp.title']}`}</div>
+      </>
+    );
+  }
+  return <div>{'Nothing'}</div>;
+};
+
+const createTestProps = (props: Partial<Props> = {}) => ({
```
tdeekens: 👌🏿

package.json
```diff
@@ -110,6 +110,7 @@
     "prettier": "1.18.2",
     "puppeteer": "1.17.0",
     "rcfile": "1.0.3",
+    "react-testing-library": "7.0.1",
```
emmenko: FYI: I moved the rtl dep to the root package. Only a couple of packages (e.g. app-shell, jest preset) define it explicitly as a peer dependency.

packages/i18n/src/async-locale-data/async-locale-data.spec.tsx
```diff
@@ -0,0 +1,115 @@
+import { mocked } from 'ts-jest/utils';
```
emmenko: I found this here: https://github.com/kulshekhar/ts-jest/issues/576#issuecomment-458178545

It's basically to have the mocked modules `jest.mock` properly typed.

packages/i18n/src/async-locale-data/async-locale-data.spec.tsx
```diff
@@ -0,0 +1,115 @@
...
+import React from 'react';
+import { reportErrorToSentry } from '@commercetools-frontend/sentry';
+import { render, wait } from 'react-testing-library';
```
emmenko: I also rewrote the tests using rtl

PR #775, opened 2019-06-19
chore: move with-mouse-over-state internal

montezume:
#### Summary

Moves the HOC `withMouseOverState` from UI-Kit internal.

FYI, I tried to refactor to not use it, but it's not so easy since `Avatar` accepts that prop and changes the image opacity and background colour of it's div. I guess we can modify Avatar to accept custom styling through (maybe accepting a default component?) and deal with it that way. But as the current Avatar component states, it's hard to change the appearance on hover without relying on it's implementation details...


montezume:
Yeah I think so.

Because avatar doesn't accept a className, so we would need to "dive in"  and target the div directly in the hover.

PR #778, opened 2019-06-23
chore(circleci): save only node_modules, persist workspace across jobs

emmenko:
Previously we were storing both `node_modules` and yarn cache folder. This lead to circleci having to store > 900 Mb of data, which increases the job time for restoring the cache:

image

Let's see if this helps, yarn [has this config](https://github.com/yarnpkg/yarn/blob/01713684fa5cc6b6d6143ac0bbea2c4f889d5fd8/.circleci/config.yml#L33-L37) as well.

Additionally, we also persist the project folder to the workspace and use it across jobs. Yarn [does the same](https://github.com/yarnpkg/yarn/blob/01713684fa5cc6b6d6143ac0bbea2c4f889d5fd8/.circleci/config.yml#L38-L41).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,6 +4,7 @@ aliases:
   - &project_directory ~/project
   - &working_directory ~/project/merchant-center-application-kit
   - &project_folder merchant-center-application-kit
+  - &cypress_cache_folder /home/circleci/.cache/Cypress
   - &restore_cache
       name: Restoring yarn cache
       keys:
@@ -27,6 +28,7 @@ executors:
       - image: cypress/base:10
         environment:
           TERM: xterm
+          CYPRESS_CACHE_FOLDER: *cypress_cache_folder
 
 # orbs:
 #   compare-url: iynere/compare-url@1.1.0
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -36,26 +23,33 @@ commands:
...
       - persist_to_workspace:
-          root: *working_directory
+          root: ~/project
           paths:
-            - packages
-            - playground
-            - visual-testing-app
-            - application-templates
-  restore_workspace:
+            - merchant-center-application-kit
```
emmenko: Yarn [does the same thing](https://github.com/yarnpkg/yarn/blob/01713684fa5cc6b6d6143ac0bbea2c4f889d5fd8/.circleci/config.yml#L38-L41), let's see if it speed things up.

tdeekens: Nice. Thanks. Nit: we could slap `~/project` into a `&project_directory`.

tdeekens:
There we go :)

- Cache size: 355 MB
- Install time: 30s

Is that where we want to end up at? On master we're at

- Cache size: 935 MB
- Install time: 25s

So I guess we cut down the cache size without harming install time 👯‍♂️.

emmenko:
So I spent some more time on this, read a lot on the internet and now I'm honestly even more confused what the appropriate setup should be 🙃

tdeekens:
#ignoranceisbliss

emmenko:
Closing in favor of #786

PR #785, opened 2019-06-24
chore: update deps

montezume:
#### Chore

Update deps

* Closes #781 
* Closes #780 
* Closes #782 
* Closes #783 
* Closes #779


packages/mc-scripts/package.json
```diff
@@ -59,7 +59,7 @@
     "postcss": "7.0.17",
     "postcss-color-mod-function": "3.0.3",
     "postcss-custom-media": "7.0.8",
-    "postcss-custom-properties": "8.0.10",
+    "postcss-custom-properties": "9.0.1",
```
montezume: requires node 8 or higher

packages/mc-scripts/package.json
```diff
@@ -40,13 +40,13 @@
...
-    "babel-plugin-react-intl": "3.2.1",
-    "browserslist": "4.6.2",
+    "babel-plugin-react-intl": "3.3.3",
+    "browserslist": "4.6.3",
     "clean-webpack-plugin": "1.0.1",
     "core-js": "3.1.4",
     "css-loader": "2.1.1",
     "cssnano": "4.1.10",
-    "file-loader": "3.0.1",
+    "file-loader": "4.0.0",
```
montezume: minimum required nodejs version is 8.9.0

packages/mc-scripts/package.json
montezume: hm, this is breaking our builds, resulting in "unknown word" errors

montezume: [ref](https://github.com/postcss/postcss-custom-properties/issues/186)

tdeekens: Wät.

packages/mc-scripts/package.json
tdeekens: Should be fine under development I think.

montezume:
After downgrading the postcss plugin, the packages build now and playground runs.

tdeekens:
Some prop-types seem to blow.

emmenko:
Sometimes I wonder why are we starting the week by getting into the dependency update mess, to then spend half a day (or more) fixing that mess 🙃

montezume:
@tdeekens 🙏

Issue #767, opened 2019-06-18
Investigate problems with Percy


montezume:
Ever since we introduced the first iteration of the `ModalPage`, Percy has been considering every PR to be introducing this new page. We need to figure out why, and fix it, because now in App Kit Percy is not adding much value.

emmenko:
Yeah it's weird. Should be maybe uninstall Percy in appkit and re-install it again? 🤔

montezume:
It's worth a shot? Can you try it?

emmenko:
I archived the project and created a new one...let's see what happens

emmenko:
I've pushed a commit to master (https://github.com/commercetools/merchant-center-application-kit/commit/23d7ed031677e29cb8d950132321e8f7749b97ee) that should trigger percy, which will set the baseline for the VRTs.

emmenko:
Wait, we don't run percy on master... 🙃

emmenko:
I'll use #769 to set the baseline for the VRTs

emmenko:
I have a feeling that when we skip the percy build, the next time it asks for approvals again. Maybe we should just run percy on each commit like in uikit?

montezume:
The weird thing is we skip Percy in UI-Kit as well, when it's not a PR

emmenko:
That's fine, but here we additionally skip it on PRs...I think that's the problem

montezume:
Let's remove the skipping and see if it fixes it?

emmenko:
Percy now runs on all PRs, let's see how it goes.

montezume:
We are still having this issue. We need to contact Percy support about it I reckon. Can anyone volunteer to take this on?

emmenko:
I guess...let's wait for a couple more PRs, just to make sure this problem is still there.

emmenko:
Did it happen again? #791 

![image](https://user-images.githubusercontent.com/1110551/60248322-b3da0d00-98c2-11e9-980c-5bce7f5cba7e.png)

montezume:
Yes, I had to approve the diff there. That's why I reopened this 😇

emmenko:
Yeah ok...if you have time please reach out to their support. Otherwise I'll do that later/tomorrow

montezume:
I have noticed something weird today. Take a look at two open PRs.

https://github.com/commercetools/merchant-center-application-kit/pull/827

and https://github.com/commercetools/merchant-center-application-kit/pull/825

[...]

PR #786, opened 2019-06-24
refactor(circleci): to try improving CI workflow run time, additionally clean things up

emmenko:
Simpler approach than #778 to see if things can improve as well (_inspired by the react circleci config_).


.circleci/config.yml
```diff
@@ -4,69 +4,72 @@ aliases:
...
+    working_directory: *working_directory
     docker:
       - image: circleci/node:10.16
-    working_directory: *working_directory
+
   cypress:
     working_directory: *working_directory
     docker:
-      - image: 'cypress/base:10'
+      - image: 'cypress/base:10.16.0'
```
emmenko: See https://github.com/cypress-io/cypress-docker-images/tree/master/base

.circleci/config.yml
```diff
@@ -4,69 +4,72 @@ aliases:
...
-      - ~/.cache/yarn
-      - node_modules
+    save_cache:
+      name: 'Saving yarn cache'
+      key: v1-yarn-cache-{{ .Branch }}-{{ checksum "yarn.lock" }}
+      paths:
+        - ~/.cache/yarn
+
+  - &run_yarn
+    run: yarn --frozen-lockfile
```
emmenko: It's recommended to use `--frozen-lockfile` on CI

.circleci/config.yml
```diff
@@ -4,69 +4,72 @@ aliases:
...
 
   - &restore_yarn_cache
-    name: 'Restoring yarn cache'
-    keys:
-      - v1-yarn-cache-{{ checksum "yarn.lock" }}
-      - v1-yarn-cache
+    restore_cache:
+      name: 'Restoring yarn cache'
+      keys:
+        - v1-yarn-cache-{{ .Branch }}-{{ checksum "yarn.lock" }}
```
emmenko: React (and other CI configs that I found) use the `.Branch` part in the cache key name. I think it makes sense, so that branches don't mess with each other dependencies.

.circleci/config.yml
```diff
@@ -4,69 +4,72 @@ aliases:
...
     steps:
       - persist_to_workspace:
           root: *working_directory
           paths:
-            - packages
-            - playground
-            - visual-testing-app
-            - application-templates
-  restore_workspace:
+            - packages/**/dist/
```
emmenko: This should be enough to persist

.circleci/config.yml
```diff
@@ -266,41 +269,29 @@ commands:
...
           command: yarn run-if:percy
-  build_packages:
-    steps:
-      - build_artifacts
-      - save_workspace
-  build_playground:
-    description: 'Building playground'
-    steps:
-      - run: yarn playground:build
-  build_starter:
```
emmenko: Those commands 👆 were never used

.circleci/config.yml
```diff
@@ -266,41 +269,29 @@ commands:
...
-          command: ./scripts/release_canary.sh
+      - run: ./scripts/release_canary.sh
 
 jobs:
   lint_and_test:
     executor: node_10
     steps:
-      - install_dependencies
+      - setup_dependencies
+      - *save_yarn_cache
```
emmenko: This is separated now from `setup_dependencies` because it's only necessary in the first job

.circleci/config.yml
tdeekens: I think we did that before or at laest elsewhere.

.yarnrc
```diff
@@ -0,0 +1 @@
+yarn-path ".yarn/releases/yarn-1.16.0.js"
```
emmenko: I saw this in the Jest repo.
This should ensure that yarn always uses the given release version (which is included in the repo), so that no matter which images are installed on CI, yarn should consistently run with the version we want.

tdeekens: How to those release versions end up in the `.yarn`-path? Will yarn pull them down in case they're missing? Are they cached?

emmenko: We manually download and save the file from the yarn release page (https://github.com/yarnpkg/yarn/releases/tag/v1.16.0)

emmenko: Ok, I just checked in CI that this works.

I can document how to deal with those

emmenko: See https://yarnpkg.com/en/docs/yarnrc#yarn-path-

> Instructs yarn to defer to another Yarn binary for execution. Useful if you want to bundle Yarn into your repository and have everyone use the same version for consistency. This was introduced in Yarn 1.0, so all developers must have Yarn >= 1.0 installed.

.circleci/config.yml
```diff
@@ -389,6 +204,11 @@ workflows:
       - vrt_test:
           requires:
             - build
+          filters:
+            branches:
+              ignore:
+                - master
```
emmenko: As mentioned here (https://github.com/commercetools/merchant-center-application-kit/issues/767#issuecomment-503512937) we should always run Percy on PRs.

emmenko:
I would like to try this and see how it goes. We can revert back in case it does not work properly.

PR #789, opened 2019-06-25
fix(jest-stylelint-runner): include run file

montezume:
#### Summary

The `run` file was omitted from package.json of `jest-stylelint-runner`. This PR adds that file.

### DIFF
--- a/packages/jest-stylelint-runner/package.json
+++ b/packages/jest-stylelint-runner/package.json
@@ -20,6 +20,7 @@
   },
   "files": [
     "package.json",
+    "run.js",
     "LICENSE",
     "README.md"
   ],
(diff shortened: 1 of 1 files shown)


packages/jest-stylelint-runner/package.json
```diff
@@ -20,6 +20,7 @@
   },
   "files": [
     "package.json",
+    "run.js",
```
montezume: question: do we need to include `index.js` here as well?

tdeekens: Yeap.

emmenko: Wait, I thought we put the files into a `src` folder?

montezume: No, one of your comments was to move it out of a src folder, no? 🤔

montezume: Should I move them back into a src folder?

emmenko: I meant the tests to move them out of the src folder. Anyway, we can do that as a follow up...

PR #791, opened 2019-06-27
chore: add babel plugin transform react constant

montezume:
As discussed on slack, we add `babel-plugin-transform-react-constant-elements` to our list of babel plugins 😇 

[ref](https://babeljs.io/docs/en/next/babel-plugin-transform-react-constant-elements.html)


tdeekens:
Can you try in a local version if this works in the MC? I dont know why it shouldn't but just to make sure.

montezume:
@tdeekens I tested it with playground, built all packages and built playground in dev and prod mode. I can also do the copy paste trick if you prefer

PR #797, opened 2019-06-27
fix(babel-preset): to remove plugin-transform-react-constant-elements

tdeekens:
#### Summary

This breaks the MC tests in many ways. 

CleanShot 2019-06-27 at 18 02 05

It's hard to say what the issue is as the plugin seems to massage the jsx. I would remove it to unblock the app-kit release. If we really have a case for the plugin we need to investigate some more.


tdeekens:
...and yes, remove this again fixes CI.

PR #800, opened 2019-06-28
[v14] refactor(app-shell): remove mouse-over hoc

emmenko:


packages/application-shell/src/components/user-settings-menu/user-settings-menu.js
```diff
@@ -18,39 +18,36 @@ import {
...
-        theme={props.isMouseOver ? 'grey' : 'black'}
-      />
-    </Spacings.Inline>
-  </div>
-);
+const UserAvatar = props => {
+  const [isMouseOver, setMouseOver] = React.useState(false);
+  return (
+    <div
+      onMouseOver={() => setMouseOver(true)}
```
tdeekens: Nit: we could define to setters: setMouseOver and setMouseOut and use React.useCallback to keep the fn reference equal.

PR #809, opened 2019-07-01
chore: update webpack loaders

tdeekens:
#### Summary

This pull request mainly takes a shot at updating the webpack loaders which we failed to update in the past. In all to be able to more easily update to the next version of webpack.

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,7 +107,7 @@
     "lint-staged": "8.2.1",
     "mri": "1.1.4",
     "postcss": "7.0.17",
-    "postcss-custom-properties": "8.0.11",
+    "postcss-custom-properties": "9.0.1",
     "postcss-import": "12.0.1",
     "prettier": "1.18.2",
     "puppeteer": "1.18.1",
--- a/packages/mc-scripts/config/create-webpack-config-for-development.js
+++ b/packages/mc-scripts/config/create-webpack-config-for-development.js
@@ -193,7 +193,6 @@ module.exports = ({
         use: [
           {
             loader: require.resolve('svg-url-loader'),
-            options: { noquotes: true },
           },
         ],
       },
--- a/packages/mc-scripts/config/create-webpack-config-for-development.js
+++ b/packages/mc-scripts/config/create-webpack-config-for-development.js
@@ -215,9 +215,12 @@ module.exports = ({
           {
             loader: require.resolve('css-loader'),
             options: {
-              modules: true,
+              modules: {
+                mode: 'local',
+                localIdentName: '[name]__[local]___[hash:base64:5]',
+                hashPrefix: 'ct',
+              },
               importLoaders: 1,
-              localIdentName: '[name]__[local]___[hash:base64:5]',
             },
           },
           {
--- a/packages/mc-scripts/config/create-webpack-config-for-production.js
+++ b/packages/mc-scripts/config/create-webpack-config-for-production.js
@@ -303,9 +303,12 @@ module.exports = ({
             {
               loader: require.resolve('css-loader'),
               options: {
-                modules: true,
+                modules: {
+                  mode: 'local',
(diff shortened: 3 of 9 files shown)


package.json
```diff
@@ -107,7 +107,7 @@
     "lint-staged": "8.2.1",
     "mri": "1.1.4",
     "postcss": "7.0.17",
-    "postcss-custom-properties": "8.0.11",
+    "postcss-custom-properties": "9.0.1",
```
tdeekens: It mainly updated an internal dep and dropped node@6 support.

packages/mc-scripts/config/create-webpack-config-for-development.js
```diff
@@ -215,9 +214,12 @@ module.exports = ({
           {
             loader: require.resolve('css-loader'),
             options: {
-              modules: true,
+              modules: {
+                mode: 'local',
```
tdeekens: This is the new way of configuring it. css-modules is disabled by default now and the modules is a boolean or object.

packages/mc-scripts/config/create-webpack-config-for-development.js
```diff
@@ -215,9 +214,12 @@ module.exports = ({
...
+                localIdentName: '[name]__[local]___[hash:base64:5]',
+                hashPrefix: 'ct',
```
tdeekens: Not sure. Just thought that would be nice.

packages/mc-scripts/config/create-webpack-config-for-production.js
```diff
@@ -160,8 +160,10 @@ module.exports = ({
     },
 
     plugins: [
-      new CleanWebpackPlugin([distPath], {
-        allowExternal: true,
+      new CleanWebpackPlugin({
+        dangerouslyAllowCleanPatternsOutsideProject: true,
+        dry: true,
+        cleanOnceBeforeBuildPatterns: [distPath],
```
tdeekens: This now hooks into the webpack config and normally doesn't need the dist path. However in this case we do need it as we seem to clean outside of the project.

packages/mc-scripts/package.json
```diff
@@ -70,11 +70,11 @@
     "regenerator-runtime": "0.13.2",
     "shelljs": "0.8.3",
     "style-loader": "0.23.1",
-    "svg-url-loader": "2.3.3",
+    "svg-url-loader": "3.0.0",
     "terser-webpack-plugin": "1.3.0",
     "thread-loader": "2.1.2",
-    "url-loader": "1.1.2",
-    "webpack": "4.35.0",
+    "url-loader": "2.0.1",
```
tdeekens: Literally changes I don't understand and I dare to assume we don't rely on. At least I've never seen those things mentioned in the changelog.

package.json
montezume: Be careful, we had problems with this update

montezume: I had a lot of problems with this version 

https://github.com/postcss/postcss-custom-properties/issues/186

tdeekens: Okay so best to not update this one I presume.

packages/mc-scripts/config/create-webpack-config-for-development.js
tdeekens: This is confusing. The `css-loader` states that "options should NOT have additional properties" which should come from a validation error against their schema [here](https://github.com/webpack-contrib/css-loader/blob/master/src/options.json). However, I think we don't violate this schema. There are also issues people have run into with this on GitHub. Also when changing it to `modules: true` the same error comes around.

montezume:
Do you need help with this @tdeekens ?

tdeekens:
If you care to look into it or have a clue. I am a bit clueless why the loader rejects a valid config.

PR #813, opened 2019-07-02
refactor(app-shell): to remove team count

tdeekens:
#### Summary

I don't see us using this or reason for anybody else having to use this. This is now available through another separate query.


tdeekens:
Not sure if we consider this a breaking change.

PR #814, opened 2019-07-03
feat(app-shell): allow to disable menu items per env

tdeekens:
#### Summary

This pull request adds allowing to disable menu items per environment by convention.

This is a PoC I would value feedback if this approach makes sense.

The idea is simple. In the MC we need to hide menu items (maybe also other custom app builders) per environment.

Given the `uriPath` is `categories/mode&search`
And a env value of `disableCategoriesModeSearch` is `true`
Then the menu will be hidden

This convention based approach works by just slugifying the `uriPath` above.

Given this is deemed a way we can agree on I will continue with tests.


.prettierignore
```diff
@@ -5,3 +5,4 @@ packages/i18n/*.json
 packages/application-shell/test-utils/index.js
 .percy.yml
 .cache
+.yarn
```
tdeekens: Otherwise we format the whole of yarn :)

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -242,14 +244,21 @@ const hasMenuEveryMenuVisibilitySetToBeHidden = (
...
 export const RestrictedMenuItem = props => {
   // NOTE: Custom application are activated/deactivated while their
   // visibility is not controlled via a visibiility overwrite.
   if (
     hasMenuEveryMenuVisibilitySetToBeHidden(
       props.namesOfMenuVisibilities,
       props.menuVisibilities
-    )
+    ) ||
+    isDisabledForEnvironment(props.uriPath, props.environment)
```
tdeekens: This simplicity of this part is okay.

packages/application-shell/src/utils/slugify/slugify.js
```diff
@@ -0,0 +1,11 @@
...
+
+export default text =>
+  text
+    .toString()
+    .toLowerCase()
+    .replace(/(\w)\'/g, '$1') // Special case for apostrophes
+    .replace(/[^a-z0-9_\-]+/g, '-') // Replace all non-word chars with -
+    .replace(/\-\-+/g, '-') // Replace multiple - with single -
+    .replace(/^-+/, '') // Trim - from start of text
+    .replace(/-+$/, ''); // Trim - from end of text
```
tdeekens: I will test this later will menu configs we ourselves have.

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -666,6 +682,7 @@ export class NavBar extends React.PureComponent {
           menuVisibilities={this.props.menuVisibilities}
           applicationLocale={this.props.applicationLocale}
           projectKey={this.props.projectKey}
+          environment={this.props.environment}
```
tdeekens: Needs to be passed through to hit the function above.

packages/application-shell/src/components/navbar/navbar.js
emmenko: I would suggest to use the `key` and not the `uriPath`. The `key` is usually already the slugified `uriPath`.

Also, how about passing a list of menu item keys to be disabled?

```json
{
  "disabledMenuItems": ["settings-project", "settings-product-types"]
}
```

tdeekens:
> I would suggest to use the key and not the uriPath. The key is usually already the slugified uriPath.

Sure. Where does the key property come from? I didn’t see it configured in all our internal apps which as a result made me slugify the uriPath.

tdeekens:
> Also, how about passing a list of menu item keys to be disabled?

Not sure I follow. Where would you define the config. Just in env.json and/or k8s? I kind of like the flat, convention based, structure right now. You would rather configure a list?

emmenko:
> Where does the key property come from?

In our internal apps, it's generated. Check out the `menu.json` files.

> Not sure I follow. Where would you define the config. Just in env.json and/or k8s? I kind of like the flat, convention based, structure right now. You would rather configure a list?

In the `env.json` yes. I just think it's easier to manage one single field as a list instead of potentially many fields. Also, given the list values are supposed to be the keys, we have a 1-to-1 mapping, so no extra transformations.

```js
// instead of this
{
  disableDashboard: true,
  disableSettingsProject: true,
  disableSettingsProductTypes: true,
}

// we do this
{
  disabledMenuItems: ['dashboard', 'settings-project', 'settings-product-types']
}
```

tdeekens:
Yo. Thanks for the feedback. I adjusted things to be `disabledMenuItems` I think it simplifies things. How do you feel about this now @emmenko? :)

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -279,6 +286,8 @@ RestrictedMenuItem.propTypes = {
   featureToggle: PropTypes.string,
   namesOfMenuVisibilities: PropTypes.arrayOf(PropTypes.string),
   menuVisibilities: PropTypes.object.isRequired,
+  disabledMenuItems: PropTypes.arrayOf(PropTypes.string),
+  keyOfMenuItem: PropTypes.string.isRequired,
```
tdeekens: We need to pass it as another prop as `key` can't be accessed in React.

packages/application-shell/src/components/navbar/navbar.spec.js
```diff
@@ -634,6 +638,50 @@ describe('rendering', () => {
...
+          );
+        });
+        it('should not render <ToggleFeature>', () => {
+          expect(wrapper).not.toRender(ToggleFeature);
+        });
+        it('should not render <RestrictedByPermissions>', () => {
+          expect(wrapper).not.toRender(RestrictedByPermissions);
+        });
+      });
+      describe('when passed but menu item should not be disabled', () => {
```
tdeekens: One bored day I will rewrite this suite to RTL.

emmenko: Yeah, actually we have a "low priority" task to improve and rewrite the navbar component...I was hoping then to rewrite the tests when we do that 😇

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -331,6 +340,7 @@ export class DataMenu extends React.PureComponent {
     menuVisibilities: PropTypes.objectOf(PropTypes.bool).isRequired,
     applicationLocale: PropTypes.string.isRequired,
     projectKey: PropTypes.string.isRequired,
+    environment: PropTypes.object.isRequired,
```
emmenko:
```suggestion
    environment: PropTypes.shape({
      disabledMenuItems: PropTypes.arrayOf(PropTypes.string.isRequired),
    }).isRequired,
```

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -503,18 +515,20 @@ export class DataMenu extends React.PureComponent {
               ? menu.submenu.map(submenu => (
                   <RestrictedMenuItem
                     key={`${menu.key}-submenu-${submenu.key}`}
+                    keyOfMenuItem={`${menu.key}-submenu-${submenu.key}`}
```
emmenko: This should just be `submenu.key`. The string concatenation was just for ensuring that the react key is unique (probably not necessary).

```suggestion
                    keyOfMenuItem={submenu.key}
```

packages/application-shell/src/components/navbar/navbar.spec.js
tdeekens: It would be kind of dope to rewrite the tests before to have a hardened suite and then just rewrite the NavBar against the existing test suite.

packages/application-shell/src/components/navbar/navbar.spec.js
```diff
@@ -634,6 +638,50 @@ describe('rendering', () => {
...
+        });
+      });
+    });
+    describe('disabled menu items', () => {
+      describe('when passed menu item should be disabled', () => {
+        beforeEach(() => {
+          props = {
+            menuVisibilities: {},
+            keyOfMenuItem: 'orders',
+            disabledMenuTiems: ['orders'],
```
emmenko:
```suggestion
            disabledMenuItems: ['orders'],
```

packages/application-shell/src/components/navbar/navbar.spec.js
```diff
@@ -634,6 +638,50 @@ describe('rendering', () => {
...
+          expect(wrapper).not.toRender(RestrictedByPermissions);
+        });
+      });
+      describe('when passed but menu item should not be disabled', () => {
+        beforeEach(() => {
+          props = {
+            menuVisibilities: {},
+            permissions: [permissions.ViewProducts],
+            keyOfMenuItem: 'products',
+            disabledMenuTiems: ['orders'],
```
emmenko:
```suggestion
            disabledMenuItems: ['orders'],
```

packages/application-shell/src/components/navbar/navbar.spec.js
emmenko: Right, even better then!

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -368,23 +368,42 @@ RestrictedApplication.propTypes = {
...
+  let parsedEnvironmentValue;
+
+  try {
+    parsedEnvironmentValue = JSON.parse(environmentValueAsString);
+  } catch (e) {
+    return false;
+  }
+
+  return parsedEnvironmentValue;
+};
```
tdeekens: @emmenko this is the easiest way I could think of solving this problem of String encoded primitives.

tdeekens:
Tested this locally with both the string and array form. Both work charmingly.

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -368,23 +368,42 @@ RestrictedApplication.propTypes = {
...
-      {}
-    )
+const shallowlyCoerceValues = defaultMemoize(uncoercedEnvironmentValues =>
+  Object.keys(uncoercedEnvironmentValues).reduce(
+    (coercedEnvironmentValues, key) => {
+      const uncoercedEnvironmentValue = uncoercedEnvironmentValues[key];
+      return {
+        ...coercedEnvironmentValues,
+        [key]:
+          getIfJSONValue(uncoercedEnvironmentValue) ||
```
emmenko: Be careful, that if the actual parsed value is `false`, but the uncoerced value is `"false"`, then we will set a value of `"false"`. Same thing for e.g. number `0`.

So we don't need this `||`, we should just return the parsed value or not from the try-catch function. Do you know what I mean?

tdeekens: Yeah you're right. Pushed the change.

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -368,23 +368,40 @@ RestrictedApplication.propTypes = {
...
+ *
+ *   - `JSON.parse('null')` => `null`
+ *   - `JSON.parse('1')` => `1`
+ *   - `JSON.parse('["a", "b"]')` => `['a', 'b']`
+ */
+const getCoerceEnvironmentValue = environmentValueAsString => {
+  let parsedEnvironmentValue;
+
+  try {
+    parsedEnvironmentValue = JSON.parse(environmentValueAsString);
```
emmenko: ...or just

```suggestion
    return JSON.parse(environmentValueAsString);
```

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import isNil from 'lodash/isNil';
 import flowRight from 'lodash/flowRight';
+import camelCase from 'lodash/camelCase';
```
emmenko: Is this still used?

tdeekens:
Should be good now I think. Thanks for the feedback again. Anything else coming to mind @adnasa and @emmenko?

montezume:
Hm, this was merged into master, and there was no problem with the snapshots.

Screen Shot 2019-07-09 at 1 23 38 PM

I'm really confused as to why this is not working properly

PR #815, opened 2019-07-03
[v14] refactor(permissions): to remove permission constants and drop ManageProject support

emmenko:
Given that `ManageProject` will not be used anymore in our new permissions model, we should just remove it.
I also decided to remove the exposed `permissions` constants. In our internal apps we have the permissions for the specific app listed as constants, I think this is the better approach and the recommended way (I mentioned this in the README of the permissions package).

There is one important thing to consider though now: previously, having `canManageProject` implied a permission to do everything in a project. However, with this gone now, the logic and behavior changed a bit. We just need to double check that there are no unwanted behaviors when we release this.


emmenko:
Can you give your feedback to my last point in the PR description, about removing the manage project check?

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
```diff
@@ -84,34 +80,16 @@ describe('when only one permission matches', () => {
     });
   });
   describe('when it should not match some permission', () => {
-    describe('when `ManageProject` permission is set', () => {
```
adnasa: I always thought it should be fine to have this removed.
It is when we migrate permissions: `ManageProject` maps to `ManageProjectSettings`, `ManageDeveloperSettings` and `ManageProductTypes` as far as I'm concerned. Everything else follows user flows.

tdeekens:
> Can you give your feedback to my last point in the PR description, about removing the manage project check?

Yeah forgot sorry. We should be safe to remove it already generally. As with our `migratePermissionsForUser` forward comptible projection we already project `ManageProject` into the three split permissions. However, I would suggest to remove it after all organisations have been migrated. Just as part of a "cleanup"-initiative also in the mc-be. So let's say we remove it by the 18th the earliest or by the end up July on another PR. I think even removing it is internal to app-kit so not another breaking change given we release v14 before.

emmenko:
@tdeekens so we remove it in v14 or after that? Sorry I didn't quite get your conclusion...

emmenko:
One side effect now with removing the "catch-all" permission `ManageProject`, is that in tests we need to explicitly define the required permissions.
What are your thoughts about this? How should we handle this?

tdeekens:
> @tdeekens so we remove it in v14 or after that? Sorry I didn't quite get your conclusion...

We can do both. I would say after just for extra safety.

tdeekens:
> One side effect now with removing the "catch-all" permission ManageProject, is that in tests we need to explicitly define the required permissions.

I think that's a positive side-effect actually as tests then have to take permissions more into account.

emmenko:
So please have another look at the last commit. I refactored the `QuickAccess` to use the test-utils properly and get the data from the application context. This makes the setup easier to test. As a consequence I had to refactor a couple of other things as well.

Thanks

emmenko:
@adnasa anything else from your side?

adnasa:
on it

PR #816, opened 2019-07-04
docs(changelog): for v13.13.0 [skip ci]

tdeekens:
13.13. This release can only go bad.


tdeekens:
> Maybe test the canary first, just in case ;)

Sure will do. I tested a local build already but will test the canary again before releasing.

tdeekens:
Okay tested the canary locally with both array and string form of disabling products and the dashboard.

PR #825, opened 2019-07-08
[v14] migrate to uikit v10

montezume:
#### Summary

Updates to latest ui-kit


emmenko:
Thanks! 🙌

PS: did you do the updates using the codemods?

montezume:
I tried, it didn't really work properly 🤦‍♂ . So I just did most of it manually. I think when there are conditions or extra props it doesn't work right.

emmenko:
Haven't you tested the codemods before? What's the point of having them if they don't work? 🤔

montezume:
I tested them in UI-Kit. I'm not sure why it didn't work here. 🤷‍♂

montezume:
Anyways, like I have mentioned tons and tons of times, we don't need a lot of codemods for UI-Kit. The changes were extremely minor and took me less than 1 minute to do with find and replace.

Actually popular open source libraries like storybook-readme had a breaking change in their last minor release without even telling anyone about it until after the fact...

I'm afraid we are wasting our time creating things that nobody will ever use instead of just working effectively on our goals

emmenko:
I believe there is a lot of value of having working codemods. Yes, find-replace change could be done with the editor but what about more complex things? I think starting to do that will make it easier even for us internally to migrate changes, and we would get more familiar in writing them, thus potentially supporting more complex scenarios.

> Actually popular open source libraries like storybook-readme had a breaking change in their last minor release without even telling anyone about it until after the fact...

So if people jump over the bridge, should we follow them as well? I don't think this is a good example for supporting your argument 😅

Anyway, I just hoped we would set a good example of handling breaking changes and migrations. If you believe it's too much work and not worth it, I would not be the one to stop you.

montezume:
Sorry, I was in a bit of bad mood yesterday. I'll work on improving the changemods.

Just to clarify though,

```js
<WarningIcon theme="orange" />
```

Can be code modded

But

```js
<WarningIcon theme={getIconTheme(this.props)} />
```

Won't work with any codemod, right?

montezume:
BTW, Percy actually caught a regression! 😄

emmenko:
> Won't work with any codemod, right? We can change the prop name, but if the value is a function or a IFFE, then it's not so trivial to update...

Well the codemod needs to be properly configured, but yeah depending on how the implementation is it might be too complex to have a codemod for it. In that case users should do the migration manually.

montezume:
I need to sync up with @jonnybel before merging this.

The visual regressions seem to be "intended" through the CSS on the component...

        ```/*  specific selector for the svg of the FlatButton */
        button > span > svg {
          height: 12px !important;
          width: 12px !important;
        }```

I'm confused why the icon wasn't small before 🤔

jonnybel:
@montezume Yeah that was unnecessary, don't remember if there was any reason for it, but now it seems there isn't. I pushed a commit to fix it.

jonnybel:
Ok, it was the introduction of this `<div>` that broke the CSS selectors and caused the problem:
https://github.com/commercetools/merchant-center-application-kit/commit/9f4f091d94fefb517bc0a34a5360eeeec947b72e#diff-bd6479eb6367f5d10cab64a6a4f71e56R70

Pushing a new commit to fix this.

emmenko:
> it seems these changes on the <div> wrappers

What changes? 🤔

Anyway, thanks for looking into that!

montezume:
Thanks for the support @jonnybel 😄

montezume:
Hm, what's with the first commit?

jonnybel:
> Hm, what's with the first commit?

Oh, that commit is already in `master` and I rebased with it before doing it with `v14` by mistake...
But if we rebase `v14` with `master` it will disappear from here, it shouldn't make any difference in practice.

montezume:
okay, i'll merge this then 🎈

emmenko:
Thanks guys!

PR #827, opened 2019-07-09
feat(app-shell): track main menu click

pa3:
#### Summary

Support tracking main menu item clicks.

#### Description

In order to get some insights about user navigation behaviour it may be useful to track how users navigate using main menu. This PR adds support of this.

In order to track user's clicking on menu items, application may provide additional field `trackingEvent` inside a structure describing application menu entries. This event will be fired whenever user click on a corresponding menu item.


packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -207,6 +213,10 @@ export class MenuItemLink extends React.PureComponent {
           if (this.props.useFullRedirectsForLinks) {
             event.preventDefault();
             this.redirectTo(this.props.linkTo);
+          } else if (this.props.trackingEvent) {
+            const { event, category, label } = this.props.trackingEvent;
+            console.log('tracking!');
```
montezume: can you remove the console log?

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -534,6 +548,7 @@ export class DataMenu extends React.PureComponent {
                           useFullRedirectsForLinks={
                             this.props.useFullRedirectsForLinks
                           }
+                          trackingEvent={submenu.trackingEvent}
```
emmenko: This has several implications, so I wouldn't define this in the menu config.

Instead, how about passing a function prop from the `ApplicationShell` component?
Something like `onMenuItemClick` or something, then you are free to handle the event the way you want.
The callback function can get the `gtm` object as an argument, together with the `event`.

pa3: Took the approach you suggested. Can you, please, take a look? Did I get your idea correctly, or you meant something different?

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -492,6 +494,9 @@ export class DataMenu extends React.PureComponent {
                 : null
             }
             useFullRedirectsForLinks={this.props.useFullRedirectsForLinks}
+            onClick={() =>
+              this.props.onMenuItemClick && this.props.onMenuItemClick(menu)
```
emmenko: You already check this `else this.props.onClick && this.props.onClick();`

```suggestion
              this.props.onMenuItemClick(menu)
```

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -534,6 +539,10 @@ export class DataMenu extends React.PureComponent {
                           useFullRedirectsForLinks={
                             this.props.useFullRedirectsForLinks
                           }
+                          onClick={() =>
+                            this.props.onMenuItemClick &&
```
emmenko: Same here, not necessary

packages/application-shell/README.md
```diff
@@ -48,6 +48,7 @@ const EntryPoint = () => (
...
 | `onRegisterErrorListeners`       | `func`             |           ✅            | -       | A callback function to setup global event listeners, called when the `ApplicationShell` is mounted. The function is called with the following named arguments: `dispatch` (the dispatch function of Redux).                                                                                                                                                                      |
+| `oMenuItemClick`                 | `func`             |           ✅            | -       | A callback function to be called on Navbar menu item click. The function is called with a single parameter, object representing menu item.                                                                                                                                                                                                                                       |
```
montezume: nitpick: typo `onMenuItemClick`

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -207,7 +208,7 @@ export class MenuItemLink extends React.PureComponent {
           if (this.props.useFullRedirectsForLinks) {
             event.preventDefault();
             this.redirectTo(this.props.linkTo);
-          }
+          } else this.props.onClick && this.props.onClick();
```
emmenko: We need to pass the `track` function as an argument, otherwise we can't use it "outside of the appshell".

```js
onClick(event, track)
```

To get the `track` function, you can access it from the context:

```js
static contextType = GtmContext;
```

pa3: Why can't we `injectTracking` in application entry point instead? Passing `track` in click handler make it very tracking specific. In case we gonna do it, i'd rename `onMenuItemClick` -> `onTrackMenuItemClick`.

emmenko: It won't work, the AppShell is the top level component, so you can't wrap anything around that.

emmenko: Maybe we can avoid passing the `track` object and you can call push the tracking event directly:

```js
onMenuItemClick={event => {
  // extract info from event.currentTarget, then push it to the dataLayer
  window.dataLayer.push({
    event: 'TrackingEvent',
    trackingCategory: '',
    trackingAction: '',
    trackingLabel: '',
  });
}}
```

That's basically what the `track` function does

emmenko:
@pa3 have you seen my last comment? https://github.com/commercetools/merchant-center-application-kit/pull/827#discussion_r301922252

PR #828, opened 2019-07-09
[v14] Deprecate storage package

emmenko:


packages/storage/package.json
```diff
@@ -14,7 +14,7 @@
     "toolkit"
   ],
   "license": "MIT",
-  "private": false,
+  "private": true,
```
emmenko: After publishing v14, I'll mark this package as deprecated in npm.

tdeekens:
Sorry, is the rationale been discussed somewhere? As the PR doesnt have a description.

emmenko:
I just remembered yesterday that we can include this in the v14 release, so there was no mention in the tracking issue.
As for the motivation, we discussed this in the past that this package is not necessary anymore, as localStorage can be easily mocked with Jest now.

Hope it makes sense. Do you think we should keep it instead?

tdeekens:
Yeah it makes sense and it is what assumed. Let‘s document it properly in the release notes then. It just something I would not have spent my energy on refactoring it in the MC 😃. Still I value it. It removed a bit of confusion. Thanks.

emmenko:
> It just something I would not have spent my energy on refactoring it in the MC

Yeah, we actually don't need to do that. The "old" package version will still work, just there won't be any more updates. So it's more of a "slow", low priority migration.

tdeekens:
> Yeah, we actually don't need to do that. The "old" package version will still work, just there won't be any more updates. So it's more of a "slow", low priority migration.

Yeah but we should "lead by example" in app-kit adoption and not release/remove something we ourselves are not able/willing to follow up on. But I agree we can adapt to the removal slowly even though we've not been good at that in the past.

packages/application-shell/src/test-utils/test-utils.js
```diff
@@ -294,7 +294,26 @@ const experimentalRenderAppWithRedux = (ui, renderOptions) => {
   });
 };
 
+// Mock utilities
+const windowMocks = {
```
emmenko: I added this mocks object to the `test-utils`, which helps setting up the mock for window objects (e.g. localStorage)

tdeekens: Why have you decided against: https://www.npmjs.com/package/jest-localstorage-mock. As this "custom" `windowMocks` construct supports more mocks in the future?

montezume:
Death by deprecation

packages/application-shell/src/test-utils/test-utils.js
emmenko: I checked that out yes but felt a "bit too complicated" for this use case.

We can use that instead if you prefer to, I don't have a strong opinion.

emmenko: Oh wait, we are already using it 🤦‍♂️

Issue #429, opened 2019-03-06
[Idea] Have a package to share cypress plugins/commands


emmenko:
**Is your feature request related to a problem? Please describe.**
We can expose a cypress package `@commercetools-frontend/cypress-mc-app` to share some of our plugin setup/commands to test mc apps.

Ref: https://github.com/commercetools/merchant-center-application-kit/pull/427#discussion_r262891224

emmenko:
I guess we can close this now that we have a basic `cypress` package.

PR #836, opened 2019-07-11
[v14] refactor(mc-scripts): to remove IE from browserslit

emmenko:
@montezume is this everything that's left?

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -1,5 +1,10 @@
 import 'cypress-testing-library/add-commands';
 
+function isLocalhost() {
+  const url = new URL(Cypress.config('baseUrl'));
+  return url.hostname === 'localhost';
+}
+
 // ***********************************************
 // This example commands.js shows you how to
 // create various custom commands and overwrite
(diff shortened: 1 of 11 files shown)


montezume:
I think so. I would do all of the hooks refactorings afterwards, as they aren't breaking changes (since we require React 16.8 now)

packages/mc-scripts/package.json
```diff
@@ -25,7 +25,7 @@
     "production": [
       ">1%",
       "not op_mini all",
-      "ie 11"
+      "not dead"
```
tdeekens: That's a thing? Awesome :)

playground/menu.json
```diff
@@ -2,7 +2,7 @@
   "key": "state-machines",
   "uriPath": "state-machines",
   "icon": "RocketIcon",
-  "permissions": ["ViewStates", "ManageStates"],
+  "permissions": ["ManageDeveloperSettings", "ViewDeveloperSettings"],
```
emmenko: Thank god we have integrations tests!

Screenshot 2019-07-11 at 13 25 13


Not sure why they haven't been caught up before though...

playground/src/components/state-machines-list/state-machines-list.js
```diff
@@ -91,7 +91,7 @@ export class StateMachinesList extends React.Component {
                 <Spacings.Stack scale="m">
                   {this.props.cachedStateMachineObjectsCount !== null && (
                     <Spacings.Inline alignItems="center">
-                      <DotIcon size="small" theme="green" />
```
montezume: 👍

playground/menu.json
montezume: They don't seem to run on PRs aimed at v14? That was my experience anyways... I guess they only run on master would be my wild guess...

tdeekens: Yeah and the error only happens since our TS migration as the invariant now runs in render and not as a custom prop-type anymore.

Welcome to my fearfull world of breaking the MC for the last weeks with trying to slowly integrating permissions.

emmenko: btw: I also had to migrate the org for the e2e test account to the new permissions and enable the view,manageDeveloperSettings permission (that's why the tests were still failing)

cypress/support/commands.js
```diff
@@ -30,25 +35,33 @@ Cypress.Commands.add('logout', () => {
...
+    cy.getByLabelText(/email/i).type(Cypress.env('LOGIN_USER'));
+    cy.getByLabelText(/password/i).type(Cypress.env('LOGIN_PASSWORD'), {
+      log: false,
+    });
+    cy.get('[aria-label="Sign in"]').click();
+    cy.get('[data-test=top-navigation').should('exist');
+    cy.visit(`${Cypress.config('baseUrl')}${redirectToUri}`);
+    cy.url().should('include', redirectToUri);
+  }
+);
```
tdeekens: Maybe we can have a `packages/cypress` at some point where we share e.g. a `commands.js` to be used by the MC too.

emmenko: Yeah, there is already an issue for that: #429

PR #841, opened 2019-07-12
chore: use rollup replace

montezume:
#### Summary

By using rollup plugin replace to set `process.env.NODE_ENV`, we can take advantage of some of rollups treeshaking capabilities 🌴

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -13,6 +13,7 @@ const json = require('rollup-plugin-json');
 const commonjs = require('rollup-plugin-commonjs');
 const postcss = require('rollup-plugin-postcss');
 const peerDeps = require('rollup-plugin-peer-deps-external');
+const replace = require('rollup-plugin-replace');
 const builtins = require('rollup-plugin-node-builtins');
 const babelPluginImportGraphQL = require('babel-plugin-import-graphql');
 const postcssCustomProperties = require('postcss-custom-properties');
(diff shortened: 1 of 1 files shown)


rollup.config.js
```diff
@@ -37,6 +38,9 @@ const createConfig = cliArgs => {
       sourcemap: true,
     },
     plugins: [
+      replace({
+        'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV),
```
emmenko: Hmm we already do this in webpack, I'm not sure if it makes sense to do it here as well.

tdeekens: Some others however might not use our webpack config and will not do this or forget to. I think this is a decent safety net?

montezume: @emmenko do we apply it to node_modules?

montezume: I'll check the built playground code to see if it achieves the same thing...

emmenko: I'm mostly concerned to "turn off" things in the libraries when they are used in development as well. So it should be up to the application bundler to turn off those things.

For example, if we do this, the logger won't work anymore when we develop apps locally
https://github.com/commercetools/merchant-center-application-kit/blob/dc3795c73216b447e231c5f3748b4b9bf9b0e867/packages/sdk/src/utils/logger.js#L9

montezume: Yeah I think you are right 👍

PR #842, opened 2019-07-14
refactor(app-shell/quick-acess): to use new permissions

tdeekens:
#### Summary

This pull request refactors Quick Access to use the new permissions. Given organizations will start being migrated we otherwise might allow users to navigate to a part of the MC through Quick Access where we then show a `PageUnauthorized`.

### DIFF
--- a/packages/application-shell/src/components/quick-access/constants.js
+++ b/packages/application-shell/src/components/quick-access/constants.js
@@ -1,0 +1,26 @@
+// eslint-disable-next-line import/prefer-default-export
+
+export const permissions = {
+  ViewOrders: 'ViewOrders',
+  ManageOrders: 'ManageOrders',
+  ViewProducts: 'ViewProducts',
+  ManageProducts: 'ManageProducts',
+  ViewCategories: 'ViewCategories',
+  ManageCategories: 'ManageCategories',
+  ViewCustomers: 'ViewCustomers',
+  ManageCustomers: 'ManageCustomers',
+  ViewCustomerGroups: 'ViewCustomerGroups',
+  ManageCustomerGroups: 'ManageCustomerGroups',
+  ViewProductDiscounts: 'ViewProductDiscounts',
+  ManageProductDiscounts: 'ManageProductDiscounts',
+  ViewDiscountCodes: 'ViewDiscountCodes',
+  ManageDiscountCodes: 'ManageDiscountCodes',
+  ViewCartDiscounts: 'ViewCartDiscounts',
+  ManageCartDiscounts: 'ManageCartDiscounts',
+  ViewProjectSettings: 'ViewProjectSettings',
+  ManageProjectSettings: 'ManageProjectSettings',
+  ViewDeveloperSettings: 'ViewDeveloperSettings',
+  ManageDeveloperSettings: 'ManageDeveloperSettings',
+  ViewProductTypes: 'ViewProductTypes',
+  ManageProductTypes: 'ManageProductTypes',
+};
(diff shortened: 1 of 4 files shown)


packages/application-shell/src/components/quick-access/constants.js
```diff
@@ -0,0 +1,26 @@
...
+  ManageDiscountCodes: 'ManageDiscountCodes',
+  ViewCartDiscounts: 'ViewCartDiscounts',
+  ManageCartDiscounts: 'ManageCartDiscounts',
+  ViewProjectSettings: 'ViewProjectSettings',
+  ManageProjectSettings: 'ManageProjectSettings',
+  ViewDeveloperSettings: 'ViewDeveloperSettings',
+  ManageDeveloperSettings: 'ManageDeveloperSettings',
+  ViewProductTypes: 'ViewProductTypes',
+  ManageProductTypes: 'ManageProductTypes',
+};
```
tdeekens: So not to depend on the permissions package anymore. Given we add more permissions there this can be removed again but I'd prefer not to do that now.

packages/application-shell/src/components/quick-access/create-commands.js
```diff
@@ -68,47 +64,54 @@ export default ({
...
         action: { type: 'go', to: `/${project.key}/categories` },
         subCommands: [
-          {
+          hasPermission(permissions.ViewCategories, project.permissions) && {
             id: 'go/categories/list',
             text: intl.formatMessage(messages.openCategoriesList),
             action: { type: 'go', to: `/${project.key}/categories?mode=list` },
           },
-          {
+          hasPermission(permissions.ViewCategories, project.permissions) && {
```
tdeekens: I decided to rather be repetitive here than trying to be smart.

packages/application-shell/src/components/quick-access/quick-access.js
```diff
@@ -351,14 +348,7 @@ class QuickAccess extends React.Component {
 export default flowRight(
   withApollo,
   injectIntl,
-  injectFeatureToggles([
-    'pimSearch',
-    'customApplications',
-    'canViewOrders',
-    'canViewCategories',
-    'canViewDashboard',
-    'canViewDiscounts',
```
tdeekens: These are not needed anymore. There is an internal MC PR with more context.

packages/application-shell/src/components/quick-access/quick-access.spec.js
```diff
@@ -1166,7 +1165,7 @@ describe('QuickAccess', () => {
 
       const searchInput = getByTestId('quick-access-search-input');
       fireEvent.change(searchInput, { target: { value: searchTerm } });
-      await waitForElement(() => getByText('Open Dashboard'));
+      await waitForElement(() => getByText('Open Manage Organizations'));
```
tdeekens: The dashboard is bound to orders now, just as in the MC. As a result we can't wait for it when we intend to test if orders are hidden.

tdeekens:
We need the release before. I would prefer not to mix those things if possible. Sorry if that causes issues on the existing branch but I think we need this before and I would prefer not to rush v14 so that QuickAccess works with new permissions.

emmenko:
Sure ok, I’ll update the branch with those changes then

emmenko:
I’ll take a look at this PR tomorrow

packages/application-shell/src/components/quick-access/create-commands.js
adnasa: redudancy is king in this case

PR #851, opened 2019-07-15
refactor(quick-access): leftover for permissions cleanup

emmenko:
Leftovers from #842

### DIFF
--- a/packages/application-shell/src/components/quick-access/create-commands.js
+++ b/packages/application-shell/src/components/quick-access/create-commands.js
@@ -108,10 +108,7 @@ export default ({
         keywords: ['Go to Customers'],
         action: { type: 'go', to: `/${project.key}/customers` },
         subCommands: [
-          hasSomePermissions(
-            [permissions.ViewCustomers],
-            project.permissions
-          ) && {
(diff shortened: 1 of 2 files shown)


tdeekens:
Thanks. Will include. I personally don't really like having the fns for one and multiple with the only difference accepting an array or not. Anyway. Thanks :)

emmenko:
> I personally don't really like having the fns for one and multiple with the only difference accepting an array or not

Yeah, actually we only use `hasPermission` in quick-access, we don't use it in the MC. Should we remove the export from the `permissions` package (in v14)?

```js
export {
  hasPermission,
  hasEveryPermissions,
  hasSomePermissions,
} from './utils/has-permissions';
```

tdeekens:
> Yeah, actually we only use hasPermission in quick-access, we don't use it in the MC. Should we remove the export from the permissions package (in v14)?

Yeah I wouldn't mind removing it. But then we'd to refactor the usages again to `hasSomePermission` in QuickAccess?

emmenko:
> But then we'd to refactor the usages again to hasSomePermission in QuickAccess?

Yes, we can also do that in the v14 branch.

PR #852, opened 2019-07-15
[v14] feat(actions-global): expose hooks for dispatching notifications

emmenko:


packages/actions-global/src/hooks.js
```diff
@@ -0,0 +1,49 @@
...
+
+// Returns a function that dispatches a notification, pre-configured to
+// a speficic notification.
+// Example:
+//   const showSuccessNotification = useShowNotification({
+//     kind: 'success',
+//     domain: DOMAINS.SIDE,
+//   });
+//   showSuccessNotification({ text: "All good!" });
+export function useShowNotification(notificationFragment) {
```
tdeekens: I struggled with naming this too. Fragment sounds good to me. I think I went with `config` and `content` in the end.

packages/actions-global/src/hooks.js
```diff
@@ -0,0 +1,49 @@
...
+  showUnexpectedErrorNotification,
+  showApiErrorNotification,
+} from './global';
+import { useDispatch } from 'react-redux';
+
+// Returns a function that dispatches an API error notification.
+// Example:
+//   const showApiErrorNotification = useShowApiErrorNotification();
+//   showApiErrorNotification({ errors });
+export function useShowApiErrorNotification() {
```
tdeekens: Nitpick: I would like to have one hook per file. It might sounds not needed or overengineered but it comes at the benefit that I can enter `useShowApiErrorNotification` and get to the file.

packages/actions-global/src/hooks.js
emmenko: Yeah, I find `config` a bit misleading though, as it seems that you can pass some configuration options (like when you're setting up a server or a request or something).
In reality is just a part of the notification object, so I figured "notification fragment" to be a more suitable name.

packages/actions-global/src/hooks.js
emmenko: Sure, I can split them. I'll do that for the global actions as well.

PR #855, opened 2019-07-15
[v14] refactor(react-notifications): to use hooks

emmenko:
- use FC
- use hooks internally
- rewrite some tests to RTL


packages/react-notifications/src/components/notifications-list/notifications-list.js
```diff
@@ -51,55 +53,49 @@ function mapNotificationToComponent(notification) {
...
+    <div className={styles[`container-${props.domain}`]}>
+      {notifications.map(notification => {
+        // 1. Check if there is a custom notification component first
+        const CustomNotificationComponent = mapCustomNotificationToComponent(
+          notification
+        );
 
-  render() {
-    return (
-      <NotificationsConnector domain={this.props.domain}>
```
emmenko: This is not needed anymore

packages/react-notifications/src/components/notifications-list/notifications-list.spec.js
```diff
@@ -1,271 +1,191 @@
...
-  let wrapper;
-  let props;
+  let rendered;
   beforeEach(() => {
-    props = createTestProps();
-    wrapper = shallow(<NotificationsList {...props} />);
+    useSelector.mockClear();
+    useSelector.mockReturnValue({ page: [], side: [], global: [] });
+    useCustomNotificationComponent.mockClear();
+    useCustomNotificationComponent.mockReturnValue(() => null);
```
emmenko: We can just mock the hooks now

packages/react-notifications/src/components/notifier/notifier.js
```diff
@@ -1,53 +1,42 @@
...
-  }
-
-  componentWillUnmount() {
-    this.notificationHandle.dismiss();
-  }
+    return () => {
+      // Remove notification when component "unmounts"
+      notification.dismiss();
+    };
+  }, []); // run only once
```
emmenko: Please double check that the `useEffect` is implemented correctly. According to the test, it should be fine.

tdeekens: Yeah, I it looks alright to me.

emmenko:
> Wondering though if such a refactoring could also happen post v14 as, given something goes wrong, this increases the "area for failure" quite a bit?

Yeah, but it also allows us to test hooks internally before shipping the release...and we can still drop things that we don't need anymore from being exported.

PR #856, opened 2019-07-15
[v14] refactor: example apps to use hooks where possible

emmenko:


playground/src/components/state-machines-details/state-machines-details.js
```diff
@@ -15,84 +15,116 @@ import {
...
+  isLoading: true,
+  error: null,
+  data: null,
+};
+
+// NOTE: using `useEffect` for asynchronously update the state, thus re-rendering
+// the component, has currently a limitation when testing the component, as it throws
+// a warning about `act`. This will be fixed in react@16.9.
+// In the meantime we can use a plain old stateful class component.
+class StateMachinesDetailsFetcher extends React.Component {
```
tdeekens: Why did this compnonent change at all?

playground/src/components/state-machines-details/state-machines-details.js
```diff
@@ -15,84 +15,116 @@ import {
...
+const StateMachinesDetails = props => {
+  const dataLocale = useApplicationContext(context => context.dataLocale);
+  const dispatch = useDispatch();
+  const showApiErrorNotification = useShowApiErrorNotification();
+
+  return (
+    <StateMachinesDetailsFetcher
+      id={props.id}
+      dataLocale={dataLocale}
+      dispatch={dispatch}
```
tdeekens: Ah, can we not separate concerns here and make the child less aware and not pass `dispatch` but a bound action creator already?

playground/src/components/state-machines-list/state-machines-list.js
```diff
@@ -35,107 +34,84 @@ export const columnsDefinition = [
...
-  measurementCache = null;
-  registerMeasurementCache = cache => {
-    this.measurementCache = cache;
-  };
-  handleRowClick = (rowIndex, results) => {
-    this.props.goToStateMachineDetail(results[rowIndex].id);
-  };
+const StateMachinesList = props => {
+  const cachedStateMachine = useSelector(selectStateMachinesFromCache);
+  const cachedStateMachineObjectsCount = cachedStateMachine
```
tdeekens: Isn't this just `hasStateMachines` as a better name?

playground/src/components/state-machines-list/state-machines-list.spec.js
```diff
@@ -81,11 +98,87 @@ describe('list view', () => {
...
+                },
+              },
+            },
+            response: {
+              ...defaultMock.response,
+              id: 'sm2',
+              key: 'sm-2',
+            },
+          },
+        ],
```
tdeekens: Maybe we can keep the setup here a bit slimmer in that we have different mocks define above?

application-templates/starter/src/routes.js
```diff
@@ -1,28 +1,38 @@
...
 
-const ApplicationRoutes = ({ match }) => (
-  <RestrictedByPermissions
-    permissions={[PERMISSIONS.ManageProducts, PERMISSIONS.ViewProducts]}
-    unauthorizedComponent={PageUnauthorized}
-    shouldMatchSomePermissions={true}
-  >
+const ApplicationRoutes = ({ match }) => {
+  const canRenderRestrictedRoutes = useIsAuthorized({
+    demandedPermissions: [PERMISSIONS.ManageProducts, PERMISSIONS.ViewProducts],
```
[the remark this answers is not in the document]
emmenko: You're right, I'll fix that thanks

playground/src/components/state-machines-details/state-machines-details.js
emmenko: Yeah so actually I wanted to use only hook (with `useEffect`) but there is still this issue with `act` in testing. So I just extracted the fetching logic into a stateful component, but it's just a temporary thing. Eventually it should all be hooks.

playground/src/components/state-machines-details/state-machines-details.js
emmenko: See https://github.com/commercetools/merchant-center-application-kit/pull/856#discussion_r303628451

playground/src/components/state-machines-list/state-machines-list.js
emmenko: Hmm what do you mean? The `cachedStateMachineObjectsCount` is a number.

playground/src/components/state-machines-details/state-machines-details.js
tdeekens: Yeah I know this problem. Haf it with sign up components too.

playground/src/components/state-machines-details/state-machines-details.js
tdeekens: That part is clear. I was just wondering is we can not pass disptach down but bound action creators.

emmenko: Changed, thanks

application-templates/starter/src/routes.js
```diff
@@ -1,28 +1,38 @@
...
+);
 
-const ApplicationRoutes = ({ match }) => (
-  <RestrictedByPermissions
-    permissions={[PERMISSIONS.ManageProducts, PERMISSIONS.ViewProducts]}
-    unauthorizedComponent={PageUnauthorized}
-    shouldMatchSomePermissions={true}
-  >
+const ApplicationRoutes = ({ match }) => {
+  const canRenderRestrictedRoutes = useIsAuthorized({
```
tdeekens:
```suggestion
  const vanViewProducts = useIsAuthorized({
```

playground/src/components/state-machines-details/state-machines-details.js
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import flowRight from 'lodash/flowRight';
-import { withApplicationContext } from '@commercetools-frontend/application-shell-connectors';
-import { connect } from 'react-redux';
+import { useDispatch } from 'react-redux';
+import { useApplicationContext } from '@commercetools-frontend/application-shell-connectors';
```
tdeekens: Are these used in this component. Feels like I fail to spot that.

playground/src/components/state-machines-details/state-machines-details.js
```diff
@@ -15,84 +15,120 @@ import {
...
-    }
-  )
-)(StateMachinesDetails);
+const StateMachinesDetails = props => {
+  const dataLocale = useApplicationContext(context => context.dataLocale);
+  const dispatch = useDispatch();
+  const fetchData = React.useCallback(
+    (...args) => dispatch(fetchStateMachine(...args)),
+    [dispatch]
+  );
```
tdeekens: Thanks.

playground/src/components/state-machines-details/state-machines-details.js
emmenko: Yes, for the `dataLocale`

PR #859, opened 2019-07-16
feat(app-shell): add passing `teamId` in link and middleware

tdeekens:
_This PR is not needed before v14. It's just a spike or thought experiment_

#### Summary

This pull request adds passing along the `teamId` in the Apollo link or Redux middleware.

#### Description

At some point the MC likely will have a team switcher. This allows e.g. admins to view permissions etc as if they were a member of another team.

All of the queries in the mc-be already work with a `teamId`, it's just not passed. The idea is for now to not have the switcher as a UI element but allow us to set the `teamId` somehow.

My first gut feeling is not to have the `teamId` as `?teamId=<UUID>` in the url. It's not really a sub-resource to anything, doesn't have to be passed around in a link. As a result we could start with the `teamId` in local storage (just like the `projectKey`). This would allow us to try this feature out silently and not to have to change all urls/links in the MC.

### DIFF
--- a/packages/sdk/src/middleware/middleware.js
+++ b/packages/sdk/src/middleware/middleware.js
@@ -72,7 +72,7 @@ export default function createSdkMiddleware({
           ...(action.payload.headers || {}),
           ...(shouldRenewToken ? { 'X-Force-Token': 'true' } : {}),
           'X-Project-Key': projectKey,
-          'X-Team-Id': teamId,
+          ...(teamId && { 'X-Team-Id': teamId }),
         };
         const body =
           action.payload.method === 'POST' ? action.payload.payload : undefined;
--- a/packages/sdk/src/middleware/middleware.js
+++ b/packages/sdk/src/middleware/middleware.js
@@ -30,11 +30,13 @@ const actionToUri = (action, projectKey) => {
 export default function createSdkMiddleware({
   getCorrelationId,
   getProjectKey,
+  getTeamId,
 }) {
   const client = createClient({ getCorrelationId });
 
   return ({ dispatch }) => next => action => {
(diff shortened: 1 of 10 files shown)


tdeekens:
This does the trick surprisingly well. I've got the integration with the API running too. The question is:

1. `<projectKey>/<applicationName>?teamId=<teamId>
2. Local storage only for now

Do you have any views towards that?

emmenko:
I sent you a DM to get some clarification

packages/sdk/src/middleware/middleware.js
```diff
@@ -70,6 +72,7 @@ export default function createSdkMiddleware({
           ...(action.payload.headers || {}),
           ...(shouldRenewToken ? { 'X-Force-Token': 'true' } : {}),
           'X-Project-Key': projectKey,
+          ...(teamId && { 'X-Team-Id': teamId }),
```
adnasa: 👍 
yeah, as discussed we might have to add `store` in the future

emmenko: You mean as an extra HTTP header?

tdeekens: Yes, so use case being: a user is member of team A and B wheere A has store access D, E and B has store access F, G while the user browsers the MC as team A _and_ only wants to see store E (not D). Let's see when we get there.

tdeekens: Let's see if it has to be a header or just can be part of a query e.g. `inStore(key: "luxury-brand")`.

tdeekens:
Given this is just a silent feature and we all agree that localstorage is more feasiable than adjusting any url I will merge.

PR #860, opened 2019-07-16
[v14] refactor(l10n): to implement loading data with hooks, refactor internals of HOCs

emmenko:
...so that we also ship hooks for the l10n package.

### DIFF
--- a/packages/l10n/src/create-l10n-injector/create-l10n-injector.spec.tsx
+++ b/packages/l10n/src/create-l10n-injector/create-l10n-injector.spec.tsx
@@ -1,17 +1,16 @@
-import * as React from 'react';
-import { shallow, ShallowWrapper } from 'enzyme';
+import { renderHook } from '@testing-library/react-hooks';
 import { reportErrorToSentry } from '@commercetools-frontend/sentry';
-import createL10NInjector from './create-l10n-injector';
+import { createL10NHook } from './create-l10n-injector';
 
 jest.mock('@commercetools-frontend/sentry');
 
 type Candy = Record<number, string>;
 type Candies = Record<string, Candy>;
-
-type TestProps = {
-  locale: string;
+type Result = {
+  isLoading: boolean;
+  data?: Candies;
+  error?: Error;
 };
-
 const candies: Candies = {
   en: {
     1: 'sugar',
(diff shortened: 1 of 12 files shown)


packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
```diff
@@ -21,44 +89,22 @@ export default function createL10NInjector<LoadedData>({
...
-                this.setState({ [propKey]: data, [propLoadingKey]: false });
-            }
-          );
-        };
-        render() {
-          return <WrappedComponent {...this.props} {...this.state} />;
-        }
+    return (WrappedComponent: React.ComponentType<Props>) => {
+      const L10NComponent = (props: Props) => {
+        const state = useL10n(mapPropsToLocale(props));
```
emmenko: Much nicer now!

packages/l10n/src/create-l10n-injector/create-l10n-injector.spec.tsx
```diff
@@ -24,155 +23,74 @@ const candies: Candies = {
...
 };
-
 const loadLocalesMock = jest.fn(
   (locale: string, cb: (error?: Error, data?: Candies) => void) =>
-    cb(undefined, candies[locale])
+    new Promise(resolve => {
+      setTimeout(() => {
+        resolve();
+        cb(undefined, candies[locale]);
+      }, 10);
```
tdeekens: ?

packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
```diff
@@ -11,8 +11,76 @@ type InjectorOptions<LoadedData> = {
...
+    const [data, dispatch] = React.useReducer<
+      (
+        prevState: State<LoadedData>,
+        action: Action<LoadedData>
+      ) => State<LoadedData>
+    >(reducer, initialState);
+
+    React.useEffect(() => {
+      let cleaning = false;
+      dispatch({ type: 'loading' });
```
tdeekens: Nit: define `const setLoading = () => dispatch({...})` and `const setError =...` ontop.

packages/l10n/src/create-l10n-injector/create-l10n-injector.spec.tsx
emmenko: Should I leave it out?

packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
emmenko: Any particular reason for doing it like this?

emmenko:
> Starting to get a bit worried how much we cramp into v14 though.

Worried about things breaking or that we include too many features?

tdeekens:
> Worried about things breaking or that we include too many features?

Too many changes increasing the surface area of intended breakage.

packages/l10n/src/create-l10n-injector/create-l10n-injector.spec.tsx
tdeekens: I was wondering why it's added? To end up on another tick?

packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
tdeekens: Readability. It's a nit.

emmenko:
Well, most of the changes are internal refactorings. I'm also quite confident that our tests setup and types caught pretty much all the problems.

With that said, we obviously need to do careful testing when migrating. We also don't need to migrated all the new features yet, but just the breaking changes. For instance, we can start using/migrating to the new hooks with time.

packages/l10n/src/create-l10n-injector/create-l10n-injector.spec.tsx
emmenko: Yeah, the intent was just to make it clear that it's an async execution.

tdeekens:
> Well, most of the changes are internal refactorings. I'm also quite confident that our tests setup and types caught pretty much all the problems.

That's my point. I don't have a lack of confidence in our setup. I just wonder why we cramp these into the release instead of releasing small incremental change we integrate early and often instead of a big bang release. Maybe that's just me though.

packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
emmenko: I see. I think I prefer to keep it like this, the hook function is already a bit long

packages/l10n/src/create-l10n-injector/create-l10n-injector.spec.tsx
tdeekens: You can just to `Promise.resolve` then? That's scheules a mircrotask whereas `setTimeout` is a regular task.

emmenko:
Sure, I also agree with you. In this case I wanted to include as many "breaking" changes as possible to avoid having to possibly release a new major version every week, for little changes. I think it's just a bit nicer DX in regards to the frequency of updating the packages to newer versions.

packages/l10n/src/create-l10n-injector/create-l10n-injector.spec.tsx
emmenko: `Promise.resolve` does not work in this case, because we resolve the loading through a callback.

I can change the loading part though, to use promises only.

emmenko: Done, it's also nicer now! https://github.com/commercetools/merchant-center-application-kit/pull/860/commits/5b36f1f3d527f3266c752d5760b34bfa0258e7bf

tdeekens:
Okay, all good. I think we would end up in a needless argument now if we continue. It's a trade-off and balancing act as usual.

PR #861, opened 2019-07-18
[v14] fix(jest-preset): missing polyfill for plural rules

emmenko:
TL;DR; the polyfill for `PluralRules` was missing (for nodejs/test environment).

### DIFF
--- a/packages/jest-preset-mc-app/polyfills/intl-pluralrules.js
+++ b/packages/jest-preset-mc-app/polyfills/intl-pluralrules.js
@@ -1,0 +1,141 @@
+/**
+ * NOTE: this is a copy of https://github.com/eemeli/intl-pluralrules
+ * as the package is shipped with ES6 code.
+ */
+const pluralRules = require('make-plural');
+const pluralCategories = require('make-plural/umd/pluralCategories');
+
+// does not check for duplicate subtags
+const isStructurallyValidLanguageTag = locale =>
+  locale.split('-').every(subtag => /[a-z0-9]+/i.test(subtag));
+
+const canonicalizeLocaleList = locales => {
+  if (!locales) return [];
+  if (!Array.isArray(locales)) locales = [locales];
+  const res = {};
+  for (let i = 0; i < locales.length; ++i) {
+    let tag = locales[i];
+    if (tag && typeof tag === 'object') tag = String(tag);
+    if (typeof tag !== 'string') {
+      // Requiring tag to be a String or Object means that the Number value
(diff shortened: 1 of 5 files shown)


packages/jest-preset-mc-app/polyfills/intl-pluralrules.js
```diff
@@ -0,0 +1,141 @@
+/**
+ * NOTE: this is a copy of https://github.com/eemeli/intl-pluralrules
+ * as the package is shipped with ES6 code.
+ */
+const pluralRules = require('make-plural');
```
emmenko: Basically changing the imports to cjs

packages/jest-preset-mc-app/polyfills/intl-pluralrules.js
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  select(number) {
+    if (typeof number !== 'number') number = Number(number);
+    if (!isFinite(number)) return 'other';
+    const fmt = this._format(Math.abs(number));
+    return pluralRules[this._locale](fmt, this._type === 'ordinal');
+  }
+}
+module.exports = PluralRules;
```
emmenko: ...and the export

packages/jest-preset-mc-app/polyfills/intl.js
```diff
@@ -0,0 +1,6 @@
+// Polyfill `Intl` for NodeJS, as `react-intl` (v3) relies on the `intl-locales-supported`
+// package, which checks if the locale is supported by the following constructors.
+// In the browser everything is fine, however in NodeJS environment we need to polyfill it.
+// https://github.com/formatjs/formatjs/tree/master/packages/intl-messageformat#modern-intl-dependency
+global.Intl = require('intl');
+global.Intl.PluralRules = require('./intl-pluralrules');
```
emmenko: FYI: we didn't catch this problem before because in appkit we don't use plural rules.

emmenko:
I'll merge and release `14.0.0-alpha.1` so that I can continue with the migration.

tdeekens:
What on earth is this? What does this do?

emmenko:
A plural rules is when you define a message like this:

```
{count} {count, plural, one {This is one} other {Those are more than one}}
```

emmenko:
The `PluralRules` polyfill is not included in the `intl` package.

https://github.com/formatjs/formatjs/tree/master/packages/intl-messageformat#modern-intl-dependency

tdeekens:
Ah that stuff.

PR #863, opened 2019-07-18
[v14] refactor(l10n): to use class component in tests until react 16.9

emmenko:
This is a bit of a workaround, due to the issue with react 16.8 and `act`.

I still want to use the hook implementation under the hood though, but when running tests (and when react is still 16.8) we can fall back to use the class component.

### DIFF
--- a/packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
+++ b/packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
@@ -2,19 +2,21 @@ import React from 'react';
 import { reportErrorToSentry } from '@commercetools-frontend/sentry';
 import { getDisplayName } from '../utils';
 
-type LoadLocale<LoadedData> = (locale: string) => Promise<LoadedData>;
-type InjectorOptions<LoadedData> = {
+type LoadLocale<LoadedData extends {}> = (
+  locale: string
+) => Promise<LoadedData>;
+type InjectorOptions<LoadedData extends {}> = {
   displayName: string;
   propKey: string;
   propLoadingKey: string;
   loadLocale: LoadLocale<LoadedData>;
 };
-type State<Data> = {
+type State<Data extends {}> = {
   isLoading: boolean;
-  data?: Data;
+  data?: Data | {};
   error?: Error;
 };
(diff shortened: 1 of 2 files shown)


packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
```diff
@@ -2,19 +2,21 @@ import React from 'react';
 import { reportErrorToSentry } from '@commercetools-frontend/sentry';
 import { getDisplayName } from '../utils';
 
-type LoadLocale<LoadedData> = (locale: string) => Promise<LoadedData>;
-type InjectorOptions<LoadedData> = {
+type LoadLocale<LoadedData extends {}> = (
```
emmenko: The `extends {}` is unrelated to the issue, but is still useful to indicate that the generic type should be an object

packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
```diff
@@ -2,19 +2,21 @@ import React from 'react';
...
   displayName: string;
   propKey: string;
   propLoadingKey: string;
   loadLocale: LoadLocale<LoadedData>;
 };
-type State<Data> = {
+type State<Data extends {}> = {
   isLoading: boolean;
-  data?: Data;
+  data?: Data | {};
```
emmenko: This is also unrelated to the issue with `act` but it's also a fix to be backwards compatible.

packages/l10n/src/create-l10n-injector/create-l10n-injector.tsx
```diff
@@ -98,3 +113,51 @@ export function createL10NInjector<LoadedData>({
     };
   };
 }
+
+export function createLegacyL10NInjector<LoadedData extends {}>({
```
emmenko: This is basically the "old" implementation

tdeekens: Hm. Looks good to me. I personally think we're trying to jump onto the hook train a bit too eagerly given this but that's just my humble view.

PR #880, opened 2019-07-22
[experiment] Enable yarn offline mirror

emmenko:
This is an experiment to make our process more resilient by using yarn offline mirror.

https://yarnpkg.com/blog/2016/11/24/offline-mirror/

More reading:
- https://twitter.com/leeb/status/1017607265115750400
- https://twitter.com/renovatebot/status/1151459996275814400
- https://medium.com/slackernoon/speed-up-ci-builds-with-npm-offline-cache-aa72a2bb63c5


tdeekens:
Thanks for trying this out. I read up on the subject too when the topic came up. Generally I am wondering if one really wants all those tarballs in git. I know it's kind of the idea for the build to be even more reproducible but would it also make sense to just have them in CircleCI's cache (for now) or would that not work at all?

emmenko:
From what I read so far you should include them in git. I wouldn't rely on the CI cache.

Btw: this branch is just meant to "experiment" with it, there is decision yet whether we are going with this or not

tdeekens:
Sure. I understood and value it as such. I am also just wondering/reflecting. I am curious to how the community accepts this.

emmenko:
Nothing much going on here, closing

PR #884, opened 2019-07-23
refactor(notifications): to convert to TS

emmenko:
Assigned @adnasa for awareness, in case you want to take a look.


packages/notifications/src/middleware.ts
```diff
@@ -0,0 +1,70 @@
...
+
+const dismissCallbacksMap = new Map<
+  TNotification['id'],
+  TNotificationOnDismiss
+>();
+let id = 0;
+
+const middleware = <Payload extends TNotification>({
+  dispatch,
+}: MiddlewareAPI) => (next: Dispatch<TNotificationAction<Payload>>) => (
```
tdeekens: TMiddlewareAPI or is it a external type?

packages/notifications/src/middleware.ts
```diff
@@ -0,0 +1,70 @@
...
+      if (callback) callback(notificationId);
+      dismissCallbacksMap.delete(notificationId);
+      return next(action);
+    }
+    default:
+      return next(action);
+  }
+};
+
+export default middleware;
```
tdeekens: I am only checking types, assuming that the remainder didn't change.

packages/notifications/src/notifications.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+  });
+
+  it('should add a notification and remove it after calling onDismiss', () => {
+    const notification: FooNotification = { id: 0, foo: 'bar' };
+
+    const store = createStore<
+      TNotificationState<FooNotification>,
+      TNotificationAction<FooNotification>,
+      unknown,
+      unknown
```
tdeekens: I read what the unknown type is can you share why we use it here?

packages/notifications/src/middleware.ts
emmenko: Yeah, it's from `redux`

packages/notifications/src/middleware.ts
emmenko: Just an internal change regarding the notification id but it does not affect the outcome.

packages/notifications/src/notifications.spec.ts
emmenko: Yes, it's because the `createStore` signature expects 4 generic types:

```ts
<S, A extends Action, Ext, StateExt>
```

But I'm only interested in the first 2. The `unknown` is kind of like a "safer" `any`.

image

adnasa:
thanks. let me have a look

packages/notifications/src/middleware.ts
```diff
@@ -0,0 +1,70 @@
...
+  TNotificationOnDismiss,
+  TNotificationAction,
+} from './types';
+import { ADD_NOTIFICATION, REMOVE_NOTIFICATION } from './action-types';
+import { removeNotification } from './action-creators';
+
+// Force TS cast of generic action to TNotificationAction
+const isNotificationAction = <Payload extends TNotification>(
+  action: Action
+): action is TNotificationAction<Payload> =>
```
adnasa: oh 🤩

tdeekens: It is a type predicate which is part of the whole type guards thing. It is a bit like an ‚as‘ but safer and used in return types.

emmenko: Yeah, Philipp actually gave me the hint 😉

PR #885, opened 2019-07-23
refactor(browser-history): to convert to TS

emmenko:


packages/browser-history/src/enhanced-history.ts
```diff
@@ -2,8 +2,10 @@ import { createBrowserHistory } from 'history';
 import withQuery from 'history-query-enhancer';
 import { encode, decode } from 'qss';
 
-// Note: the "?" needs to be removed.
-const parse = search => decode(search.substring(1));
+function parse<Q extends {}>(search: string): Q {
+  // Note: the "?" needs to be removed.
+  return decode(search.substring(1)) as Q;
```
emmenko: The casting is a bit unfortunate here, because the `parse` function of `history-query-enhancer` is typed differently than `decode`.

The `decode` signature is actually more correct, as it uses a `Partial` type.

adnasa: hahahaha! This is the one that drove me crazy 🤣 
thanks for looking into it and follow it through :trollface:

PR #891, opened 2019-07-25
fix(permissions): to drill props

tdeekens:
#### Summary

This pull request fixes drilling the props through RestrictedByPermissions to Authorized. It's the only occurance where we missed this. 

The RestrictedByPermissions is still tested via shallow rendering which is why tests nor typing barked at us. I plan to refactor those after the fix is out this or early next week.


emmenko:
> The RestrictedByPermissions is still tested via shallow rendering which is why tests nor typing barked at us. I plan to refactor those after the fix is out this or early next week.

Awesome, thanks!

PR #896, opened 2019-07-26
refactor(react-notifications): to convert to TS

emmenko:
Uff, this was a real PITA!

TL;DR: This shows me that we need to rethink the implementation of the notifications system, for a couple of reasons: 1) the code/logic is spread across multiple packages 2) the notifications system is a bit "too dynamic" and it was hard to make it sounded typed (I still had to do some workarounds) 3) we can make it simpler

Alongside the refactoring of the `react-notifications` package to TS, I also had to fix/extend types and other things across the repo. To make it "easier" to understand the context of the changes, I decided to keep the changes together in one PR.
Furthermore, I also refactored the notification components to use emotion (the VRTs should help with ensuring that nothing changed).

I'll try to document the choices I had to make as PR comments, so hold on a bit with reviewing until the PR is set to ready for review.

---

Assigning @adnasa @montezume  for awareness in case you want to have a look


tdeekens:
How about other team members also review some TypeScript to learn some? :)

@types/commercetools-frontend__ui-kit/index.d.ts
```diff
@@ -0,0 +1,16 @@
+declare module '@commercetools-frontend/ui-kit' {
```
emmenko: We really need to expose types from ui-kit!

I'll try to work on [this approach](https://github.com/commercetools/ui-kit/pull/860) until we decide to convert the components to TS as well. This should buy us some time.

/cc @montezume

@types/sentry-testkit/index.d.ts
```diff
@@ -1,26 +0,0 @@
-declare module 'sentry-testkit' {
```
emmenko: This is not needed anymore, `sentry-testkit` comes with its own types now.

package.json
```diff
@@ -68,7 +68,6 @@
     "@types/prop-types": "15.7.1",
     "@types/react": "16.8.23",
     "@types/react-dom": "16.8.4",
-    "@types/react-intl": "2.3.18",
```
emmenko: Also not needed, react-intl v3 comes with TS support

packages/actions-global/src/actions/handle-action-error.ts
```diff
@@ -51,7 +51,7 @@ export default function handleActionError(error: ActionError) {
       showApiErrorNotification({
         errors: error.body.errors || [
           // Pass a fallback error so that our error components can handle it
-          { message: error.body.message, code: 'Unknown' },
```
emmenko: The `code` should be optional

packages/actions-global/src/actions/show-notification.ts
```diff
@@ -3,15 +3,18 @@ import {
   addNotification,
   TNotificationMetaOptions,
 } from '@commercetools-frontend/notifications';
-import { DOMAINS } from '@commercetools-frontend/constants';
+import {
+  NOTIFICATION_DOMAINS,
```
emmenko: This is just an alias to `DOMAINS` (to be backwards compatible). I find it a bit more clear what "domains" refer to. We can use this internally in app-kit first.

packages/actions-global/tsconfig.json
```diff
@@ -1,5 +1,4 @@
 {
   "extends": "../../tsconfig.json",
-  "files": ["./src/index.ts"],
-  "include": ["./src/globals.d.ts"]
```
emmenko: Those have been now moved to `<root>/@types`, as I figured out the correct setup of `tsconfig.json`

packages/constants/src/constants.ts
```diff
@@ -11,31 +11,65 @@ export const HIDE_LOADING = 'HIDE_LOADING';
 export const HIDE_ALL_PAGE_NOTIFICATIONS = 'HIDE_ALL_PAGE_NOTIFICATIONS';
 
 // Notifications
-export type TAppNotificationDomain = 'global' | 'page' | 'side';
-export const DOMAINS: {
-  [key: string]: TAppNotificationDomain;
```
emmenko: This was actually still wrong, as intellisense didn't provide any meaningful options.

packages/constants/src/constants.ts
```diff
@@ -11,31 +11,65 @@ export const HIDE_LOADING = 'HIDE_LOADING';
...
+  'api-error': 'api-error';
+} = {
+  error: 'error',
+  warning: 'warning',
+  info: 'info',
+  success: 'success',
+  'unexpected-error': 'unexpected-error',
+  'api-error': 'api-error',
+};
+export type TAppNotificationDomain = (typeof NOTIFICATION_DOMAINS)[keyof typeof NOTIFICATION_DOMAINS];
```
emmenko: This "weird" syntax basically extracts the object values as union types.

See https://stackoverflow.com/a/50044130

packages/constants/src/constants.ts
```diff
@@ -49,17 +83,18 @@ export type TAppNotificationOfKind<
     ? TAppNotificationKindSide
     : never;
 };
-export type TAppNotificationApiError = {
-  code: string;
+export type TAppNotificationApiError<ExtraFields extends {} = {}> = {
```
emmenko: The `ExtraFields` is used to extend the error object, which is helpful for the API error notifications

packages/constants/src/constants.ts
```diff
@@ -68,59 +103,62 @@ export type TAppNotification<
...
-  ...NOTIFICATION_KINDS_SIDE,
-  'unexpected-error': 'unexpected-error',
-};
-export const NOTIFICATION_KINDS_PAGE: {
-  [key: string]: TAppNotificationKindPage;
-} = {
-  ...NOTIFICATION_KINDS_GLOBAL,
-  'api-error': 'api-error',
-};
+export type TAppNotificationGlobal = TAppNotification<{
```
emmenko: We can also construct specific types for each notification domain, which is necessary to be able to differentiate between the notifications when we handle them as a list.

packages/jest-preset-mc-app/setup-test-framework.js
```diff
@@ -1,13 +1,13 @@
...
-require('@testing-library/react/cleanup-after-each');
-const Enzyme = require('enzyme');
-const Adapter = require('enzyme-adapter-react-16');
-const ShallowWrapper = require('enzyme/ShallowWrapper');
-const configureEnzymeExtensions = require('@commercetools/enzyme-extensions');
-const commerceToolsEnzymeMatchers = require('@commercetools/jest-enzyme-matchers');
-require('./polyfills/intl');
+import 'unfetch/polyfill';
+import 'jest-enzyme';
+import '@testing-library/jest-dom/extend-expect';
```
emmenko: Using `import` is actually an attempt to instruct vscode to look up the extended types of the custom matchers.

We'll have to look into this more as a follow up

packages/react-notifications/src/components/notification-kinds/api-error-message/api-error-message.spec.tsx
```diff
@@ -1,17 +1,18 @@
...
 jest.mock('@commercetools-frontend/sentry');
 
-const renderMessage = ui =>
+const renderMessage = (ui: React.ReactElement) =>
   render(<IntlProvider locale="en">{ui}</IntlProvider>);
 
 describe('render', () => {
   it('should show message for InvalidInput', () => {
-    const error = { code: 'InvalidInput' };
+    const error = { code: 'InvalidInput', message: 'message-content' };
```
emmenko: Besides the missing "required" fields, I didn't need to change a single test. That was very nice!

packages/react-notifications/src/components/notification-kinds/api-error-message/api-error-message.tsx
```diff
@@ -0,0 +1,171 @@
+import React from 'react';
+import { useIntl, IntlShape } from 'react-intl';
+import has from 'lodash/has';
+import { reportErrorToSentry } from '@commercetools-frontend/sentry';
+import { TAppNotificationApiError } from '@commercetools-frontend/constants';
+import messages from './messages';
+
+const regexInvalidOperationRequiredAttribute = /Required attribute '(.*)' cannot be removed/;
+
+type ExtraErrorFields = {
```
emmenko: This type contains all the possible extra fields from any of the API errors that we explicitly handle here

packages/react-notifications/src/components/notification-kinds/api-error-message/api-error-message.tsx
```diff
@@ -0,0 +1,171 @@
...
+        validFrom: string;
+        validUntil: string;
+      };
+  detailedErrorMessage?: string;
+  attribute?: {
+    name: string;
+  };
+};
+type Props = {
+  error: TAppNotificationApiError<ExtraErrorFields>;
```
emmenko: Here is how we "extend" the notification error type

packages/react-notifications/src/components/notification-kinds/api-error-message/api-error-message.tsx
```diff
@@ -0,0 +1,171 @@
...
+};
+
+const ApiErrorMessage = (props: Props) => {
+  const intl = useIntl();
+
+  // Attempt to map the error to a specific error message
+  const formattedMessage = getMessageByErrorCode(props.error, intl);
```
emmenko: I split the logic of mapping over the errors from the main component function, to make the logic more readable

packages/react-notifications/src/components/notification-kinds/api-error-message/api-error-message.tsx
```diff
@@ -0,0 +1,171 @@
...
+  if (formattedMessage) {
+    return <>{formattedMessage}</>;
```
emmenko: The fragment is necessary to tell react that this returns a JSX element, otherwise it's a string

packages/react-notifications/src/components/notification-kinds/api-error/api-error.tsx
```diff
@@ -0,0 +1,50 @@
...
+  NOTIFICATION_KINDS_PAGE,
+} from '@commercetools-frontend/constants';
+import Notification from '../../notification';
+import ApiErrorMessage from '../api-error-message';
+
+type Props = {
+  notification: TAppNotification<{
+    domain: typeof NOTIFICATION_DOMAINS.PAGE;
+    kind: typeof NOTIFICATION_KINDS_PAGE['api-error'];
+    values: TAppNotificationValuesApiError;
```
emmenko: The conditional typings of `TAppNotification` turned out to be pretty nice, as we can specify the exact shape of the notification.

packages/react-notifications/src/components/notification-kinds/unexpected-error/unexpected-error.tsx
```diff
@@ -0,0 +1,40 @@
...
+} from '@commercetools-frontend/constants';
+import { Spacings } from '@commercetools-frontend/ui-kit';
+import Notification from '../../notification';
+import apiErrorMessages from '../api-error-message/messages';
+
+type Props = {
+  notification: TAppNotification<{
+    domain: typeof NOTIFICATION_DOMAINS.PAGE;
+    kind: typeof NOTIFICATION_KINDS_PAGE['unexpected-error'];
+    values: TAppNotificationValuesUnexpectedError;
```
emmenko: FYI: you can see here how the type system helps avoiding mistakes

Screenshot 2019-07-25 at 14 13 17
Screenshot 2019-07-25 at 14 13 26

packages/react-notifications/src/components/notifications-list/notifications-list.tsx
```diff
@@ -0,0 +1,273 @@
...
+  selectSideNotifications,
+} from './selectors';
+import { TAppState } from './types';
+import { reportErrorToSentry } from '@commercetools-frontend/sentry';
+
+type Props = {
+  domain: TAppNotificationDomain;
+};
+
+const getStyles = (props: Props) => {
```
emmenko: I'll move those into a separate file

packages/react-notifications/src/components/notifications-list/notifications-list.tsx
```diff
@@ -0,0 +1,273 @@
...
+    default:
+      return css``;
+  }
+};
+
+const NotificationsList = (props: Props) => {
+  const dispatch = useDispatch();
+  const mapCustomNotificationToComponent = useCustomNotificationComponent();
+
+  switch (props.domain) {
```
emmenko: This is one of the "mah" places, where we need to handle the different cases verbosely, otherwise the type system gets confused.
The logic is a bit repetitive but at least the typings work(_-ish_)

packages/react-notifications/src/components/notifications-list/notifications-list.tsx
```diff
@@ -0,0 +1,273 @@
...
+    case NOTIFICATION_DOMAINS.GLOBAL: {
+      const notifications = useSelector<TAppState, TAppNotificationGlobal[]>(
```
emmenko: Notice the `TAppNotificationGlobal[]` as the return type from the selector. Each switch-case has to explicitly handle this, but at least it's safer

packages/react-notifications/src/components/notifications-list/notifications-list.tsx
```diff
@@ -0,0 +1,273 @@
...
+        selectGlobalNotifications
+      );
+      return (
+        <div css={getStyles(props)}>
+          {notifications.map(notification => {
+            // 1. Check if there is a custom notification component first
+            const CustomNotificationComponent = mapCustomNotificationToComponent(
+              notification
+            ) as Notification;
```
emmenko: Since we don't know the custom notification type, we need to "cast" it to `Notification` in order to pass the `notification` prop and adhere with its types.

packages/react-notifications/src/components/notifications-list/notifications-list.tsx
```diff
@@ -0,0 +1,273 @@
...
+                />
+              );
+            }
+
+            switch (notification.kind) {
+              case NOTIFICATION_KINDS_GLOBAL.error:
+              case NOTIFICATION_KINDS_GLOBAL.warning:
+              case NOTIFICATION_KINDS_GLOBAL.info:
+              case NOTIFICATION_KINDS_GLOBAL.success: {
+                const { values, ...genericNotification } = notification;
```
emmenko: The `GenericNotification` does not accept the `values` prop, so we need to "remove" it.

packages/react-notifications/src/components/notifications-list/notifications-list.tsx
```diff
@@ -0,0 +1,273 @@
...
+                    key={notification.id}
+                    notification={genericNotification}
+                    dismiss={() => {
+                      dispatch(removeNotification(notification.id));
+                    }}
+                  />
+                );
+              }
+              case NOTIFICATION_KINDS_GLOBAL['unexpected-error']: {
+                const { text, ...errorNotification } = notification;
```
emmenko: Same here

packages/react-notifications/src/components/notifications-list/notifications-list.tsx
```diff
@@ -0,0 +1,273 @@
...
+              case NOTIFICATION_KINDS_PAGE.warning:
+              case NOTIFICATION_KINDS_PAGE.info:
+              case NOTIFICATION_KINDS_PAGE.success: {
+                const { values, ...genericNotification } = notification;
+                return (
+                  <GenericNotification
+                    key={notification.id}
+                    notification={{
+                      ...genericNotification,
+                      kind: notification.kind,
```
emmenko: I don't know what but we need to explicitly pass the `kind`, otherwise TS gets confused 🤷‍♂️

packages/react-notifications/src/components/notifications-list/notifications-list.tsx
```diff
@@ -0,0 +1,273 @@
...
+                );
+              }
+              case NOTIFICATION_KINDS_PAGE['unexpected-error']: {
+                const { text, ...errorNotification } = notification;
+                return (
+                  <UnexpectedErrorNotification
+                    key={notification.id}
+                    notification={{
+                      ...errorNotification,
+                      values: notification.values as TAppNotificationValuesUnexpectedError,
```
emmenko: This is also weird, and I don't fully understand why, but TS things that the `values` that we pass is of type `TAppNotificationValuesApiError`, which is wrong.

packages/react-notifications/src/components/notifications-list/selectors.js
```diff
@@ -1,40 +0,0 @@
...
-    notifications.filter(notification => notification.domain === DOMAINS.PAGE)
-);
-
-export const selectSideNotifications = createSelector(
-  selectNotifications,
-  notifications =>
-    notifications.filter(notification => notification.domain === DOMAINS.SIDE)
-);
-
-export const selectLatestGlobalNotificationAsList = createSelector(
```
emmenko: I removed this selector, and incorporated the `slice(0, 1)` to the `selectGlobalNotifications`

packages/react-notifications/src/components/notifications-list/selectors.js
```diff
@@ -1,40 +0,0 @@
...
-  selectGlobalNotifications,
-  notifications => notifications.slice(0, 1)
-);
-
-export const selectNotificationsByDomain = createSelector(
```
emmenko: This is also not needed anymore, as we now use each single selector in the switch-case block in the notifications-list component

packages/react-notifications/src/components/notifications-list/selectors.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export const selectGlobalNotifications = createSelector<
+  TAppState,
+  TAppState['notifications'],
+  TAppNotificationGlobal[]
+>(
+  selectNotifications,
+  notifications =>
+    // https://stackoverflow.com/a/42487130
+    (notifications as { domain: TAppNotificationDomain }[])
```
emmenko: This was also another "edge case", see SO link for more information

packages/react-notifications/src/components/notifications-list/types.ts
```diff
@@ -0,0 +1,12 @@
+import {
+  TAppNotificationGlobal,
+  TAppNotificationPage,
+  TAppNotificationSide,
+} from '@commercetools-frontend/constants';
+
+export type TAppState = {
```
emmenko: Just for reusability

packages/sentry/src/sentry.spec.ts
```diff
@@ -40,7 +40,7 @@ describe('reportErrorToSentry', () => {
       expect(report).toHaveProperty('event_id', expect.any(String));
       expect(report).not.toHaveProperty('extra');
       expect(console.error).toHaveBeenCalledWith(
-        expect.stringContaining(report.event_id)
+        expect.stringContaining(report.event_id as string)
```
emmenko: The `event_id` is marked as optional 🤷‍♂️

tsconfig.json
```diff
@@ -26,9 +26,8 @@
     "stripInternal": true,
     "target": "esnext",
     "allowJs": false,
-    "baseUrl": ".",
-    "paths": { "*": ["@types/*"] }
+    "baseUrl": "."
   },
-  "include": ["**/src/**/*", "jest.d.ts", "globals.d.ts"],
+  "include": ["@types/**/*"],
```
emmenko: This was the fix

emmenko:
> How about other team members also review some TypeScript to learn some? :)

Yeah but I wouldn't recommend to review this part as it gets quite complicated and into advanced TS mambo-jambo.

emmenko:
VRTs passed, wow I'm astonished! 🎉

packages/actions-global/src/actions/show-notification.ts
adnasa: I like this new naming a lot!

emmenko:
@tdeekens anything else I should change here?

adnasa:
but looks really good to me in general. good work 👍

PR #897, opened 2019-07-26
chore: update react-intl to 3.0.0-beta.19

emmenko:
I managed to make it work (see inline comments).


packages/application-components/src/components/page-not-found/page-not-found.js
```diff
@@ -13,13 +14,13 @@ const PageNotFound = () => (
       <FormattedMessage
         {...messages.paragraph1}
         values={{
-          link: (
+          a: msg => (
```
tdeekens: This is so nice.

packages/mc-scripts/config/vendors-to-transpile.js
```diff
@@ -0,0 +1,6 @@
+// https://github.com/formatjs/react-intl/blob/master/docs/Upgrade-Guide.md#esm-build
+module.exports = [
+  require.resolve('react-intl'),
+  require.resolve('intl-messageformat'),
+  require.resolve('intl-messageformat-parser'),
+];
```
tdeekens: I assume this won't change, right?

jest.test.config.js
```diff
@@ -1,3 +1,5 @@
+const vendorsToTranspile = require('./packages/jest-preset-mc-app/vendors-to-transpile');
```
emmenko: This is to "extend" the default vendors that are shipped by default with our jest config

jest.test.config.js
```diff
@@ -12,9 +14,7 @@ module.exports = {
     // (e.g. with babel).
     // However we need to instruct jest to actually transpile our local shared
     // dependencies (e.g. `packages`) which are symlinked as dependencies
-    // in node_modules. Since the name of those packages are scoped with
-    // `@commercetools-local` we simply match the JS files within those folders.
-    'node_modules/@commercetools-frontend/ui-kit',
```
emmenko: I'm not sure why we had this but it shouldn't be needed

packages/application-components/src/components/page-not-found/messages.js
```diff
@@ -8,10 +8,6 @@ export default defineMessages({
   paragraph1: {
     id: 'PageNotFound.paragraph1',
     defaultMessage:
-      'The item you are looking for may have been deleted, does not exist, or the URL was entered incorrectly. Check the URL and try again.\n\nPlease contact your system administrator or the commercetools {link} if you have any further questions.',
-  },
-  helpDesk: {
-    id: 'PageNotFound.helpDesk',
-    defaultMessage: 'Help Desk',
+      'The item you are looking for may have been deleted, does not exist, or the URL was entered incorrectly. Check the URL and try again.\n\nPlease contact your system administrator or the commercetools <a>Help Desk</a> if you have any further questions.',
```
emmenko: I had to change this as the current implementation was broken (the message link was compiled as `[Object]`)

emmenko: I'm not sure if this might be a breaking change or not, we need to check...

packages/i18n/data/core.json
```diff
@@ -40,8 +40,7 @@
   "FailedAuthentication.title": "We are unable to authorize you to use the Merchant Center.",
   "NavBar.MCSupport.title": "Support",
   "Notification.hideNotification": "Hide notification",
-  "PageNotFound.helpDesk": "Help Desk",
-  "PageNotFound.paragraph1": "The item you are looking for may have been deleted, does not exist, or the URL was entered incorrectly. Check the URL and try again.\n\nPlease contact your system administrator or the commercetools {link} if you have any further questions.",
+  "PageNotFound.paragraph1": "The item you are looking for may have been deleted, does not exist, or the URL was entered incorrectly. Check the URL and try again.\n\nPlease contact your system administrator or the commercetools <a>Help Desk</a> if you have any further questions.",
```
emmenko: I'll manually update the messages in TX after merging this, to reflect the new changes

packages/jest-preset-mc-app/jest-preset.js
```diff
@@ -37,6 +38,6 @@ module.exports = {
     '^.+\\.tsx?$': 'ts-jest',
     '^.+\\.graphql$': 'jest-transform-graphql',
   },
-  transformIgnorePatterns: ['node_modules'],
+  transformIgnorePatterns: [`node_modules/(?!(${vendorsToTranspile})/)`],
```
emmenko: We now ships this by default.

packages/mc-scripts/config/create-webpack-config-for-development.js
```diff
@@ -336,7 +337,7 @@ module.exports = ({
             },
           },
         ],
-        include: sourceFolders,
+        include: sourceFolders.concat(vendorsToTranspile),
```
emmenko: Same for the webpack config

packages/mc-scripts/config/vendors-to-transpile.js
emmenko: It shouldn't change no. According to this comment, they plan to ship them as ESM

https://github.com/formatjs/react-intl/issues/1371#issuecomment-514255342

tdeekens:
Thanks for the explanations.

jest.test.config.js
emmenko: Oh now I remember 🤦‍♂️
It's the only dependency out of `@commercetools-frontend` that does not need to be transpiled...

tdeekens:
Rebased over `master` due to conflicts. Assume the ship it label means it can go in to test via canary.

emmenko:
Yes thank you!

PR #898, opened 2019-07-26
chore: update deps

tdeekens:
#### Summary

We had quite a few outdated deps. After managing to fix my `node-gyp` and `libtool` I was able to install deps again.

I wanted to update flopflip only at first cause it has a nice fix for @pa3 but saw many other things being outdated.


packages/jest-preset-mc-app/package.json
```diff
@@ -39,14 +39,14 @@
     "jest-transform-graphql": "2.1.0",
     "jest-watch-master": "1.0.0",
     "jest-watch-typeahead": "0.3.1",
-    "make-plural": "4.3.0",
+    "make-plural": "5.0.0",
```
tdeekens: They only extracted their CLI into a package we don't use. So we're not affected by this.

packages/mc-http-server/package.json
```diff
@@ -25,13 +25,14 @@
...
-    "@promster/express": "2.2.4",
-    "@promster/server": "2.1.3",
+    "@promster/express": "3.1.0",
+    "@promster/server": "3.1.0",
     "compression": "1.7.4",
     "express": "4.17.1",
     "lightship": "3.0.1",
     "morgan": "1.9.1",
     "mri": "1.1.4",
+    "prom-client": "11.5.3",
```
tdeekens: This is the breaking change above in promster.

packages/mc-scripts/package.json
```diff
@@ -33,47 +33,47 @@
...
     "cssnano": "4.1.10",
-    "file-loader": "4.0.0",
+    "file-loader": "4.1.0",
     "fs-extra": "8.1.0",
     "glob": "7.1.4",
     "graphql-tag": "2.10.1",
     "html-webpack-plugin": "3.2.0",
     "json-loader": "0.5.7",
-    "mini-css-extract-plugin": "0.7.0",
+    "mini-css-extract-plugin": "0.8.0",
```
tdeekens: They just added an option to ignore order.

packages/mc-scripts/package.json
```diff
@@ -33,47 +33,47 @@
...
+    "postcss-custom-properties": "9.0.2",
     "postcss-import": "12.0.1",
     "postcss-loader": "3.0.0",
-    "postcss-preset-env": "6.6.0",
+    "postcss-preset-env": "6.7.0",
     "postcss-reporter": "6.0.1",
     "postcss-safe-parser": "4.0.1",
     "react-dev-utils": "9.0.1",
-    "regenerator-runtime": "0.13.2",
+    "regenerator-runtime": "0.13.3",
```
tdeekens: I am wondering if we actually still need this? /cc @emmenko

package.json
```diff
@@ -55,24 +55,24 @@
...
     "@types/common-tags": "1.8.0",
     "@types/enzyme": "3.10.3",
-    "@types/graphql": "14.2.2",
+    "@types/graphql": "14.2.3",
     "@types/history": "4.7.2",
     "@types/jest": "24.0.15",
     "@types/lodash": "4.14.136",
     "@types/moment-timezone": "0.5.12",
-    "@types/node": "10.14.13",
+    "@types/node": "12.6.8",
```
emmenko: We use node 10, so we should keep it pinned to v10

packages/mc-scripts/package.json
emmenko: Hmm no idea tbh...

package.json
tdeekens: Yeah I noticed and downgraded already again.

packages/application-shell/src/test-utils/test-utils.js
```diff
@@ -146,7 +147,11 @@ const renderApp = (
     ...rtl.render(
       <IntlProvider locale={locale}>
         <ApolloProviderComponent mocks={mocks} addTypename={addTypename}>
-          <ConfigureFlopFlip adapter={adpater} defaultFlags={flags}>
+          <ConfigureFlopFlip
+            adapter={adpater}
+            defaultFlags={flags}
+            adapterArgs={defaultFlopflipAdapterArgs}
```
tdeekens: This didn't change in flopflip otherwise I wouldn've released a breaking change. I assume ESLint got better or something.

tdeekens:
@emmenko I downgraded `gatsby-mdx` again to the current version. The migration made me stumble from build error to build error now. Also it is used inline at times and not only as a plugin and I am not 100% sure of the implications there. I think it makes sense to split that out.

website/package.json
```diff
@@ -17,12 +17,13 @@
     "@mdx-js/mdx": "1.1.0",
     "@mdx-js/react": "1.0.27",
     "common-tags": "1.8.0",
+    "change-case": "3.1.0",
```
emmenko: What's this for?

tdeekens: Otherwise the `website` built failed cause MDX claimed it to be missing.

emmenko: Hmm ok, I'll take a look at upgrading gatsby-mdx as a follow up. Thanks for looking into it!

PR #904, opened 2019-07-28
chore: remove (deprecated) storage package

emmenko:
The package is already marked as deprecated in npm, so there is no need to keep it around any longer.


montezume:
shouldn't this be removed in the next major version then? 🤔

emmenko:
It’s fine, we already excluded it in the v14.

emmenko:
To make it a bit more clear: marking a npm package as deprecated basically means that there are no more releases.
When we released v14, the package was already excluded from being release. It’s also mentioned in the release notes.

This is basically just a cleanup

PR #905, opened 2019-07-28
fix(mc-scripts): to use paths to package directory instead of entry points

emmenko:


tdeekens:
This is somewhat really odd to frustrating to annoying. Thanks for sticking to it.

PR #914, opened 2019-07-29
chore: enable eslint rules for react hooks

emmenko:
PS: I'll merge #896 first, as there is a component that needs to be adjusted in react-notifications

### DIFF
--- a/packages/application-shell/src/components/configure-intl-provider/configure-intl-provider.js
+++ b/packages/application-shell/src/components/configure-intl-provider/configure-intl-provider.js
@@ -19,13 +19,15 @@ const hideAppLoader = () => {
 };
 
 const ConfigureIntlProvider = props => {
+  React.useEffect(() => {
+    if (props.locale) {
+      hideAppLoader();
+    }
+  }, [props.locale]);
+
   if (!props.locale) {
     return null;
   }
-
-  React.useEffect(() => {
-    hideAppLoader();
-  }, []);
 
   return (
     <IntlProvider locale={props.locale} messages={props.messages}>
--- a/packages/react-notifications/src/components/notifications-list/notifications-list.tsx
+++ b/packages/react-notifications/src/components/notifications-list/notifications-list.tsx
@@ -30,199 +30,209 @@ type Props = {
   domain: TAppNotificationDomain;
 };
 
-const NotificationsList = (props: Props) => {
+const NotificationsListGlobal = (props: Props) => {
   const dispatch = useDispatch();
   const mapCustomNotificationToComponent = useCustomNotificationComponent();
+  const notifications = useSelector<TAppState, TAppNotificationGlobal[]>(
+    selectGlobalNotifications
+  );
+  return (
+    <div css={getStyles(props)}>
+      {notifications.map(notification => {
+        // 1. Check if there is a custom notification component first
+        const CustomNotificationComponent = mapCustomNotificationToComponent(
+          notification
+        ) as Notification;
+        if (CustomNotificationComponent) {
+          return (
+            <CustomNotificationComponent
+              key={notification.id}
+              notification={notification}
+              dismiss={() => {
+                dispatch(removeNotification(notification.id));
+              }}
+            />
+          );
+        }
 
+        switch (notification.kind) {
+          case NOTIFICATION_KINDS_GLOBAL.error:
+          case NOTIFICATION_KINDS_GLOBAL.warning:
+          case NOTIFICATION_KINDS_GLOBAL.info:
+          case NOTIFICATION_KINDS_GLOBAL.success: {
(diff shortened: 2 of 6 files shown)


packages/application-shell/src/components/configure-intl-provider/configure-intl-provider.js
```diff
@@ -19,14 +19,16 @@ const hideAppLoader = () => {
 };
 
 const ConfigureIntlProvider = props => {
+  React.useEffect(() => {
+    if (props.locale) {
+      hideAppLoader();
+    }
+  }, [props.locale]);
```
tdeekens: Curious: I thought this wouldn't work without a warning and would have to be `[locale]`?

packages/react-notifications/src/components/notifier/notifier.js
```diff
@@ -22,6 +22,7 @@ const Notifier = props => {
       // Remove notification when component "unmounts"
       notification.dismiss();
     };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []); // run only once
```
emmenko: Here the eslint rule only shows a warning

image

But I do only want to run this once...

tdeekens: Hm, Then I would add a comment and disable the rule so nobody confuses it as a mistake?

website/src/templates/markdown.js
```diff
@@ -250,6 +229,35 @@ const Link = props => {
...
+          align-items: center;
+        `}
+      >
+        <ExternalLink {...props} />
+        <ExternalLinkSvg />
+      </span>
+    );
+  }
+
+  return <PatchedLink {...props} />;
```
emmenko: To avoid to "conditionally" render a hook

packages/react-notifications/src/components/notifier/notifier.js
emmenko: Yeah...but do you think it makes sense to keep the empty array?

packages/application-shell/src/components/configure-intl-provider/configure-intl-provider.js
emmenko: No idea, it works...or at least I didn't get any errors about it 🤔

packages/react-notifications/src/components/notifier/notifier.js
tdeekens: Yeah. Otherwise changing any prop would show another notification, right? I don't think that's intended. Also I think before this used hooks it was just `componentDidMount` and no `compnentDidUpdate` which triggered another notification?

emmenko: Yes exactly

packages/application-shell/src/components/configure-intl-provider/configure-intl-provider.js
emmenko: Interesting, I looked into this a bit more and it seems that the weird "destructuring" warning only appears if you call a prop function (see below):

```js
// all good
React.useEffect(() => {
  if (props.locale) {
    hideAppLoader();
  }
}, [props.locale]);

// error
React.useEffect(() => {
  if (props.locale) {
    hideAppLoader();
    props.onChange()
  }
}, [props.locale]);
```

emmenko: So by destructuring `onChange`, the warning is gone. It seems to me that this is more of a bug in the eslint plugin 🤔

tdeekens: Ah, yeah then any assumption for reasoning I had doesn't make sense anymore 😄. Then can we just do something like:

```js
const { onChange } = props;

React.useEffect(() => {
  if (props.locale) {
    hideAppLoader();
    props.onChange()
  }
}, [props.locale, onChange]);
```

for now?

emmenko: Yes that's what I'm doing. We should maybe report this in the react repo

packages/react-notifications/src/components/notifications-list/notifications-list.tsx
```diff
@@ -30,199 +30,209 @@ type Props = {
   domain: TAppNotificationDomain;
 };
 
-const NotificationsList = (props: Props) => {
+const NotificationsListGlobal = (props: Props) => {
```
emmenko: Here I just split up the switch-case into separate components, do avoid rendering conditional hooks

tdeekens: Ah kay. My mind was expecting `GlobalNotificationsList` and `PageNotificationsList` but that's kind of the same.

PR #915, opened 2019-07-29
chore(mc-http-server): update to node v12

tdeekens:
#### Summary

This pull request suggests to update the `mc-http-server` to Node.JS v12.

### DIFF
--- a/packages/mc-http-server/Dockerfile
+++ b/packages/mc-http-server/Dockerfile
@@ -3,7 +3,7 @@
 # We use Google Cloud Container Registry to build and store the Docker image, using a Build trigger connected to our GitHub repo.
 # The build is triggered on each tag, so whenever we do a new release.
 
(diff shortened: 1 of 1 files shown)


emmenko:
What's the reason for it?

tdeekens:
What's the reason against it?

v12 has stuff like: fast `async`/`await`, async stack traves, performance improvements, better http parser.

Dunno. Just felt reasonable to update if there is nothing that speaks against it I suppose.

emmenko:
Sure makes sense. I thought there were some specific reasons.

emmenko:
🤔 why did you close it?

tdeekens:
Oh, I saw your previous comment without the approval as a "lets not do it. these reasons are not specific enough.".

emmenko:
No all good, I just forgot to approve 😅

PR #916, opened 2019-07-30
refactor: default props type declarations

emmenko:
As I'm writing and testing the type declarations for ui-kit components (https://github.com/commercetools/ui-kit/pull/860) I noticed that the way we defined the `defaultProps` for TS is wrong.

After researching things a bit I found this solution, and actually since TS 3 they have much better support for `defaultProps`.

PS: in the future React will probably "deprecate" default props for functional components, preferring variable initializers.

### DIFF
--- a/packages/react-notifications/src/components/notification/notification.tsx
+++ b/packages/react-notifications/src/components/notification/notification.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { FormattedMessage } from 'react-intl';
+import { useIntl } from 'react-intl';
 import {
   CloseBoldIcon,
   ErrorIcon,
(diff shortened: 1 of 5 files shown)


packages/react-notifications/src/components/notification/notification.tsx
```diff
@@ -40,42 +40,40 @@ const NotificationIcon = (props: PropsIcon) => {
...
+  fixed: false,
+};
 
-const Notification = (props: Props) => (
-  <div css={getStylesForNotification(props)} {...filterDataAttributes(props)}>
-    <div css={getStylesForContent(props)}>{props.children}</div>
-    {props.onCloseClick ? (
-      <div>
-        <FormattedMessage {...messages.hideNotification}>
-          {label => (
```
emmenko: About this, the argument returned by `FormattedMessage` has this signature

```ts
children?(...nodes: React.ReactNodeArray): React.ReactNode;
```

Which is a bit surprising, but at the same time we can just use `intl.formatMessage` since we need to pass a `string`.

tsconfig.json
```diff
@@ -18,6 +18,7 @@
     "noImplicitThis": true,
     "noUnusedLocals": true,
     "noUnusedParameters": true,
+    "preserveSymlinks": true,
```
emmenko: This is necessary to fix this problem when running tests in watch mode

Screenshot 2019-07-30 at 22 41 01

[Source of solution](https://github.com/TypeStrong/ts-node/issues/831#issuecomment-492238938)

PR #917, opened 2019-07-31
refactor: to use emotion styles

emmenko:
Migrate most of the css modules to css-in-js.


packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -257,21 +310,36 @@ export const RestrictedApplication = props => (
                         })()}
                       </aside>
 
-                      {/**
-                       * NOTE: in IE11 main can't be a grid-child apparently.
```
emmenko: Now that we don't support IE 11 anymore, we can use the `main` html element.

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -419,77 +487,90 @@ export default class ApplicationShell extends React.Component {
...
-        trackingEventWhitelist={this.props.trackingEventWhitelist}
-        applicationMessages={this.props.applicationMessages}
-      >
-        {({ isAuthenticated }) => {
-          if (isAuthenticated)
-            return (
-              <Switch>
-                {/* When the application redirects to this route,
+      <>
+        <Global
```
emmenko: Just adding the `Global` styles here

emmenko:
> Does the package react-notifications still need the classnames dependency

I'll remove it, I probably forgot about it

> It seems we are only left with page-bottom-spacer, state-machines-list, main-view and nav-bar which are still using css modules. Can we migrate them too in this PR?

The only relevant one is the NavBar, which is a bit of a nightmare to migrate. I would do that as a follow up. The other components are from the demo apps, we can migrate them as a follow up as well.

> The most complicated seems to be the navBar, but I'm a bit confused about it since we have another navbar component which is using css-in-js

This one is from the website, nothing to do with appkit code

packages/application-shell/src/components/app-bar/app-bar.js
```diff
@@ -1,43 +1,93 @@
...
+          display: flex;
+          align-items: center;
+        `}
+      >
         <Spacings.Inline scale="m" alignItems="center">
-          <div className={styles.switchers}>
+          <div
+            css={css`
+              /* Special sizes for dropdowns */
+              .Select-input,
```
montezume: why do we need these react-select v1 custom styles?

packages/application-shell/src/components/loading-placeholder/loading-placeholder.js
```diff
@@ -1,10 +1,94 @@
...
+const animationPulse = keyframes`
+  0% {
+    background-color: ${customProperties.colorNeutral};
+  }
+  100% {
+    background-color: ${customProperties.colorNeutral95};
+  }
+`;
+
+const getWitdhBySize = props => {
```
montezume:
```suggestion
const getWidthBySize = props => {
```

packages/application-shell/src/components/loading-placeholder/loading-placeholder.js
```diff
@@ -1,10 +1,94 @@
...
+const LoadingPlaceholder = styled.div`
+  background-color: ${customProperties.colorNeutral};
+  animation-name: ${animationPulse};
+  animation-duration: 1s;
+  animation-timing-function: ease-out;
+  animation-delay: 0;
+  animation-direction: alternate;
+  animation-iteration-count: infinite;
+
+  width: ${getWitdhBySize};
```
montezume:
```suggestion
  width: ${getWidthBySize};
```

packages/application-shell/src/components/app-bar/app-bar.js
emmenko: I just used what was already there. Do you think we don't need them anymore?

montezume: Yeah. That's all react-select v1 css app-kit doesn't use react-select anymore (it just uses the ui-kit wrappers for react select v3)

emmenko: Cool, thanks for the reminder!

packages/react-notifications/package.json
```diff
@@ -45,7 +45,6 @@
     "@commercetools-frontend/sentry": "14.2.0",
     "@emotion/core": "10.0.14",
     "@emotion/styled": "10.0.14",
-    "classnames": "2.2.6",
```
tdeekens: 💣

PR #918, opened 2019-08-01
feat(application-shell/project-query): add `allAppliedDataFences`

adnasa:
#### Summary

- adds `allAppliedDataFences` to `FetchProject` query

### DIFF
--- a/packages/application-shell-connectors/src/components/application-context/application-context.tsx
+++ b/packages/application-shell-connectors/src/components/application-context/application-context.tsx
@@ -40,18 +40,26 @@ type TApplicationContextActionRights = {
 };
 
 type TApplicationContextGroupedByPermission = {
-  [key: string]: {
-    values: string[];
-  };
+  // E.g. { canManageOrders: { values: [] } }
+  [key: string]: { values: string[] };
 };
 type TApplicationContextGroupedByResourceType = {
+  // E.g. { orders: {...} }
   [key: string]: TApplicationContextGroupedByPermission;
 };
-type TApplicationContextDataFenceGroupedByDataFenceType = {
-  [key: string]: TApplicationContextGroupedByResourceType;
-};
+/**
+ * dataFence: {
+ *   store: {
+ *     orders: {
+ *       canManageOrders: { values: ['usa', 'germany'] },
+ *       canViewOrders: { values: ['canada'] },
+ *     }
+ *   }
+ * }
+ */
 type TApplicationContextDataFences = {
-  [key: string]: TApplicationContextDataFenceGroupedByDataFenceType;
+  // E.g. { store: {...} }
+  [key: string]: TApplicationContextGroupedByResourceType;
 };
 
 type TRawProject = {
--- a/packages/application-shell-connectors/src/components/application-context/application-context.tsx
+++ b/packages/application-shell-connectors/src/components/application-context/application-context.tsx
@@ -39,6 +39,21 @@ type TApplicationContextActionRights = {
   [key: string]: TActionRight;
 };
 
+type TApplicationContextGroupedByPermission = {
+  [key: string]: {
+    values: string[];
+  };
+};
+type TApplicationContextGroupedByResourceType = {
+  [key: string]: TApplicationContextGroupedByPermission;
+};
+type TApplicationContextDataFenceGroupedByDataFenceType = {
+  [key: string]: TApplicationContextGroupedByResourceType;
+};
+type TApplicationContextDataFences = {
+  [key: string]: TApplicationContextDataFenceGroupedByDataFenceType;
+};
+
 type TRawProject = {
   key: string;
   version: number;
@@ -51,6 +66,7 @@ type TRawProject = {
   };
   permissions: TApplicationContextPermissions;
   actionRights: TApplicationContextActionRights;
+  dataFences: TApplicationContextDataFences;
 } & AdditionalProperties;
 type TApplicationContextProject = {
   key: string;
@@ -76,6 +92,7 @@ type TApplicationContext<AdditionalEnvironmentProperties extends {}> = {
   project: TApplicationContextProject | null;
   permissions: TApplicationContextPermissions | null;
   actionRights: TApplicationContextActionRights | null;
+  dataFences: TApplicationContextDataFenceGroupedByDataFenceType | null;
   dataLocale: string | null;
 };
 type ProviderProps<AdditionalEnvironmentProperties extends {}> = {
@@ -150,6 +167,7 @@ const createApplicationContext: <AdditionalEnvironmentProperties extends {}>(
   project: mapProjectToApplicationContextProject(project),
   permissions: project && project.permissions ? project.permissions : null,
   actionRights: project && project.actionRights ? project.actionRights : null,
+  dataFences: project && project.dataFences ? project.dataFences : null,
   dataLocale: projectDataLocale || null,
 });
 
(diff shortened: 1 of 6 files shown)


packages/application-shell/src/components/fetch-project/fetch-project.graphql
```diff
@@ -27,6 +27,15 @@ query ProjectQuery($projectKey: String!) {
       name
       value
     }
+    allAppliedDataFences {
+      ... on StoreDataFence {
+        type
+        storeKey
+        group
+        resourceAccessPermission
+        __typename
```
adnasa: without this, I get the following when running tests

```
       Could not find __typename on Fragment  StoreDataFence {
        type: 'store',
        storeKey: 'usa',
        group: 'orders',
        resourceAccessPermission: 'canManageOrders'
      }
```

packages/application-shell/src/components/fetch-project/fetch-project.js
```diff
@@ -77,6 +77,7 @@ class FetchProject extends React.Component {
                 permissions: mapAllAppliedToObjectShape(
                   data.project.allAppliedPermissions
                 ),
+                dataFences: data.project.allAppliedDataFences,
```
adnasa: the `mapAll*` maps permissions and action rights to 

```js
{ [permission|actionRight]: true }
```

which is not my desired shape for dataFences. However it makes a whole lot of sense to group each dataFence, which is WIP now

tdeekens: 1. We will "leak" the `__typename` here into the React context if I am not mistaken. 
2. The mapping from an array to object shape for the other two has benefits
   - No `includes` for matching just object access
   - Easier to read in React DevTools

adnasa: > We will "leak" the __typename here into the React context if I am not mistaken.

yeah that this is true, but would this be a pitfall even if we strip it off before we set it on the `applicationContext`?

---

> No includes for matching just object access

we have to do a find. for the dataFence, it will be hard to read either way. because we're introducing another "dimension" to match with `storekey + permission`. so it won't be smooth to read it even if we flatten it to an `{}`..

packages/application-shell/src/components/fetch-project/fetch-project.graphql
```diff
@@ -27,6 +27,15 @@ query ProjectQuery($projectKey: String!) {
       name
       value
     }
+    allAppliedDataFences {
+      ... on StoreDataFence {
+        type
+        storeKey
+        group
+        resourceAccessPermission
```
adnasa: I realise now that this is introducing a new term in the application-kit.
We always used `name` for it. I'll update the response accordingly

packages/application-shell/src/components/fetch-project/fetch-project.graphql
```diff
@@ -27,6 +27,15 @@ query ProjectQuery($projectKey: String!) {
       name
       value
     }
+    allAppliedDataFences {
+      ... on StoreDataFence {
+        type
+        name
+        value
```
adnasa: I updated in the MC-BE to call this `value` and not `storeKey`. This is avoid domain specific checks when we map them to applicable dataFences that we can read from in `injectAuthorized`

packages/application-shell/src/components/fetch-project/fetch-project.spec.js
```diff
@@ -190,4 +200,83 @@ describe('helpers', () => {
...
+              },
+              canViewOrders: {
+                values: ['canada'],
+              },
+            },
+          },
+        });
+      });
+    });
+    describe('with categories as datafence types', () => {
```
adnasa: We do not support other DataFence types yet in the MC-BE. However this is to test that whatever types we support in the future, we introduce a very common structure in the final output that will look as the following:

```
abstract:
{
  [dataFenceType]: {
    [resourceType]: {
      [accessType]: {
        // dataFenceSpecificValues
        values: []
      }
    }
  }
}

---

real data:
{
  [store]: {
    [orders]: {
      [canManageOrders]: {
        // dataFenceSpecificValues
        values: ['usa', 'germany']
      }
      [canViewOrders]: {
        // dataFenceSpecificValues
        values: ['canada']
      }
    }
  }
}
```

tdeekens: Yay. This is dupe.

tdeekens:
> Looks good. Do you also need to expose them in the app context?

Yes I think so. To later inject them and switch on them to perform the `inStore` query or not.

adnasa:
> Looks good. Do you also need to expose them in the app context?

yes, I missed that part. thanks for bringing it up!

playground/env.json
```diff
@@ -1,7 +1,7 @@
 {
   "applicationName": "playground",
   "frontendHost": "localhost:3001",
-  "mcApiUrl": "https://mc-api.commercetools.com",
+  "mcApiUrl": "http://localhost:8080",
```
tdeekens: Tat tat tat. 😄

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -39,6 +39,21 @@ type TApplicationContextActionRights = {
   [key: string]: TActionRight;
 };
 
+type TApplicationContextGroupedByPermission = {
+  [key: string]: {
+    values: string[];
+  };
+};
+type TApplicationContextGroupedByResourceType = {
+  [key: string]: TApplicationContextGroupedByPermission;
```
emmenko: Shouldn't this be

```suggestion
  [key: string]: { values: string[] };
```

?

adnasa: no. the new is 

```js
{
  // store
  [dataFenceType]: {
    // orders
    [resourceType]: {
      // canManageOrders
      [permission]: {
        values: []
      }
    }
  }
}
```

emmenko: Exactly...have you tried my suggestion?

emmenko: Here you go: https://github.com/commercetools/merchant-center-application-kit/pull/918/commits/e7f118b8201253068c4a1cc0444fd1ad70a00339

playground/env.json
emmenko: https://github.com/commercetools/merchant-center-application-kit/pull/918/commits/d3bf9a15b511ee4cc05df16ee53aae5f85ca7789

emmenko:
The e2e tests are failing because the backend hasn't been deployed to production yet, right?

tdeekens:
Yes. It is actually behind a flag to only work on staging. Just like permissions back them. Given we agree here, I suggest to remove the toggle, deploy and retrigger CI here.

adnasa:
> Yes. It is actually behind a flag to only work on staging. Just like permissions back them. Given we agree here, I suggest to remove the toggle, deploy and retrigger CI here.

At the moment, the dataFences are:

1. hidden behind feature toggles on the MC-FE
2. hidden behind a staging flag on the backend.

We can remove the flag on the backend, users with dataFences (just me now) will have broken views. And we have to halt deployment to production. If we are okay with that, I'll remove the flag.

tdeekens:
> And we have to halt deployment to production

Why do we have to halt a deployment. Just nobody should configure any fences and things remain sane, right? This is targeted by the LD flag in the mc-fe?

adnasa:
I'm just afraid someone inside CT already has configured data fences before we toggled it off

adnasa:
Oh yeah, I totally mixed this up with `all-applied-oauth-scopes` which included the staging flag. not `all-applied-data-fences`... so only a deploy is in order.

adnasa:
yo. check that build out, my dudes. 
thanks @emmenko <3

PR #919, opened 2019-08-01
feat: add some uikit declaration types

emmenko:
See https://github.com/commercetools/ui-kit/pull/860#issuecomment-517238813

TL;DR: we will maintain the uikit type declarations here for now. **If needed**, we can also publish the types as a package, or publish them to DefinitelyTyped.

### DIFF
--- a/@types/commercetools-frontend__ui-kit/index.d.ts
+++ b/@types/commercetools-frontend__ui-kit/index.d.ts
@@ -1,16 +1,1114 @@
 declare module '@commercetools-frontend/ui-kit' {
-  type CustomProperties = { [key: string]: string };
+  import * as React from 'react';
+  import { MessageDescriptor } from 'react-intl';
 
-  export const customProperties: CustomProperties;
+  export const version: string;
 
-  export const CloseBoldIcon: React.ElementType;
-  export const ErrorIcon: React.ElementType;
-  export const WarningIcon: React.ElementType;
-  export const InfoIcon: React.ElementType;
-  export const CheckBoldIcon: React.ElementType;
-  export const IconButton: React.ElementType;
-  export const Spacings: {
-    Inline: React.ElementType;
-    Stack: React.ElementType;
+  // Design tokens - customProperties
+  export {
+    default as customProperties,
+  } from '@commercetools-frontend/ui-kit/materials/custom-properties';
+
+  // <ContentNotification>
+  export type ContentNotificationProps = {
+    type: 'error' | 'info' | 'warning' | 'success';
+    children: React.ReactNode;
   };
+  export const ContentNotification: {
+    (props: ContentNotificationProps): JSX.Element;
+    displayName: string;
+  };
+
+  // <Avatar>
+  export type AvatarProps = {
+    firstName: string;
+    lastName: string;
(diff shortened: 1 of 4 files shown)


@types/commercetools-frontend__ui-kit/index.d.ts
```diff
@@ -1,16 +1,1114 @@
...
+  import * as React from 'react';
+  import { MessageDescriptor } from 'react-intl';
 
-  export const customProperties: CustomProperties;
+  export const version: string;
+
+  // Design tokens - customProperties
+  export {
+    default as customProperties,
+  } from '@commercetools-frontend/ui-kit/materials/custom-properties';
```
emmenko: This will be available when https://github.com/commercetools/ui-kit/pull/978 is merged and released.

For now, it's duplicated at the end of this file.

@types/commercetools-frontend__ui-kit/index.d.ts
```diff
@@ -1,16 +1,1114 @@
...
+  };
+  export const WorldIcon: {
+    (props: IconProps): JSX.Element;
+    displayName: string;
+    defaultProps: Pick<IconProps, 'size'>;
+  };
+}
+
+// TODO: remove once this is available in uikit
+declare module '@commercetools-frontend/ui-kit/materials/custom-properties' {
```
emmenko: Little trick, we will remove it once uikit ships with the ts file.

packages/react-notifications/src/components/notification/notification.tsx
```diff
@@ -24,8 +25,7 @@ import {
 
 type PropsIcon = {
   type: TAppNotificationKind;
-  // TODO: get list of supported colors from uikit
-  color: string;
+  color: IconProps['color'];
```
emmenko: 💪

packages/application-shell-connectors/package.json
```diff
@@ -40,6 +40,11 @@
   "dependencies": {
     "@babel/runtime-corejs3": "7.5.5",
     "@commercetools-frontend/constants": "14.1.1",
+    "apollo-cache-inmemory": "1.6.2",
+    "apollo-client": "2.6.3",
+    "apollo-link-http": "1.5.15",
+    "graphql": "14.4.0",
+    "graphql-tag": "^2.10.1",
```
tdeekens: ...without the `^`?

emmenko: It's the same as in app-shell actually...it's a bit weird though but those are actually the required dependencies of `react-apollo`, that's why I had to define them. The error I got if I didn't define them was something like this:

```
 Cannot find module 'apollo-cache-inmemory' from 'test-utils.js'

    However, Jest was able to find:
        './test-utils.js'
        './test-utils.spec.js'

    You might want to include a file extension in your import, or update your 'moduleFileExtensions', which is currently ['ts', 'tsx', 'js', 'jsx', 'json'].

    See https://jestjs.io/docs/en/configuration#modulefileextensions-array-string

      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:230:17)
      at Object.<anonymous> (node_modules/src/test-utils.tsx:4:1)
```

PR #921, opened 2019-08-02
refactor(application-components): to convert to TS

emmenko:
Now that we have parts of uikit typed (#919) we can put them into use for the application components. In fact, I already found some bugs! 😮


packages/application-components/src/components/modal-pages/internals/modal-page.tsx
```diff
@@ -0,0 +1,136 @@
...
+      }
+      onClose && onClose(event);
+    },
+    [onClose, props.shouldDelayOnClose]
+  );
+  return (
+    <ClassNames>
+      {({ css: makeClassName }) => (
+        <Modal
+          isOpen={forceClose === true ? false : props.isOpen}
```
emmenko: Now that we `useState`, it was a bit weird before to use `null` and `false` and the state values. Now it's either `true` or `false`, the outcome should be the same though.

packages/application-components/src/components/maintenance-page-layout/maintenance-page-layout.spec.tsx
```diff
@@ -29,9 +29,9 @@ describe('rendering', () => {
     });
     it('renders the title and paragraphs', () => {
       const { getByText } = render(<MaintenancePageLayout {...props} />);
-      expect(getByText(props.title)).toBeInTheDocument();
-      expect(getByText(props.paragraph1)).toBeInTheDocument();
-      expect(getByText(props.paragraph2)).toBeInTheDocument();
+      expect(getByText('title')).toBeInTheDocument();
```
montezume: why do you need to use the strings explicitly like this?

packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
```diff
@@ -47,7 +55,7 @@ export const getBeforeCloseOverlayAnimation = () => css`
   opacity: 0 !important;
 `;
 
-export const ContentWrapper = styled.div`
+export const ContentWrapper = styled('div')`
```
montezume: is this because @emotion/styled is missing typescript defs?

packages/application-components/src/components/maintenance-page-layout/maintenance-page-layout.spec.tsx
emmenko: Because those props are typed as react nodes but the query matcher only accepts a string or regex. We could type cast `as string` but we also usually explicitly pass the query text as string in our RTL tests.
So yeah, a bit unfortunate but it’s better like this.

packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
emmenko: It does have type defs, this is how it’s documented in their docs: https://emotion.sh/docs/typescript#emotionstyled

montezume: well in the link they also use `styled.div`

```js
// TS 2.9+ only
const Image3 = styled.div<ImageProps>`
  width: ${(props: ImageProps) => props.width};
  background: url(${(props: ImageProps) => props.src})
    center center;
  background-size: contain;
`
```

emmenko: I'll give it another try

emmenko: Thanks! https://github.com/commercetools/merchant-center-application-kit/pull/921/commits/0e1ad7f7406c1f2687d99ed022dde833339beb43

emmenko:
@montezume @jonnybel it would be great if you can have a look a this today, thanks

montezume:
Will do 👍

packages/application-components/src/components/dialogs/confirmation-dialog/confirmation-dialog.spec.tsx
```diff
@@ -1,10 +1,15 @@
...
-import PropTypes from 'prop-types';
 import { SecondaryButton } from '@commercetools-frontend/ui-kit';
 import { render, wait, fireEvent } from '../../../../test-utils';
 import ConfirmationDialog from './confirmation-dialog';
 
-const DialogController = props => {
+type DialogControllerProps = {
+  children: (renderProps: {
+    isOpen: boolean;
+    setIsOpen: React.Dispatch<boolean>;
```
montezume: I can't find anything in google (not redux related) about React.Dispatch. can you enlighten?

emmenko: It's in the `@types/react` package:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/bb36642c70f366af454e1bf67471fbfdf0b9bdc8/types/react/index.d.ts#L810

montezume: ah okay. weird

would it be more readable to just put (value: A) => void ?

emmenko: This is just for a test btw, I just went with the same API used by `useState`.

Also, if you have TS editor installed, you can see the type signatures:
image

PR #922, opened 2019-08-02
chore: do not hoist react-apollo

emmenko:


emmenko:
This change was partially done in other pending PRs, I just want to see if CI works though after reverting the uikit types commit (https://github.com/commercetools/merchant-center-application-kit/commit/71f96d7dc87135b497cac45bd364c42d997cbd74), which seems to cause the CI craziness

PR #934, opened 2019-08-06
feat(circleci): snapshot testing for documentation website

emmenko:
Source: https://docs.percy.io/docs/command-line-client#section-generating-snapshots

Let's see how good it works, to snapshot a statically built website.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,10 +97,10 @@ commands:
           command: PERCY_TOKEN="$PERCY_TOKEN_UI_COMPONENTS" yarn vrt
       - run:
           name: Building Documentation Website
-          command: yarn --cwd website build
+          command: BUILD_TARGET=percy yarn --cwd website build
       - run:
           name: Running Percy snapshots
-          command: PERCY_TOKEN="$PERCY_TOKEN_WEBSITE" yarn run percy snapshot website/public --ignore-files "website/public/code-snippets/**/*.html"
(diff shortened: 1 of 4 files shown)


emmenko:
Nice!

image

image

emmenko:
What do you guys think?

emmenko:
Btw, to give a bit more context about the reasoning: I want to be able to catch regressions like this one: #933

emmenko:
Now percy runs 2 builds

image

Not sure how the status check exactly reflects that...maybe it only keeps the "last" one? @montezume do you know about this maybe?

emmenko:
We could also maybe look into "merging" the static website and the UI components pages...

emmenko:
Ah, I think we need to set up different projects: https://docs.percy.io/docs/multiple-projects-per-repo

montezume:
@emmenko this would be my assumption of what has to happen. I'm not sure Percy supports multiple Percy runs for one monorepo...

emmenko:
I created a separate Percy project to target the website snapshots. Let's see how it goes...

emmenko:
Nice, seems to work! 🎉

.circleci/config.yml
```diff
@@ -97,10 +97,10 @@ commands:
           command: PERCY_TOKEN="$PERCY_TOKEN_UI_COMPONENTS" yarn vrt
       - run:
           name: Building Documentation Website
-          command: yarn --cwd website build
+          command: BUILD_MODE=percy yarn --cwd website build
```
tdeekens: Nit: `BUILD_TARGET` maybe be more apt?

PR #936, opened 2019-08-07
feat(e2e): take percy snapshots

### LINKED ISSUE #747: Setup VRTs for starter applications
It would be nice to start having VRTs for the application pages, specifically for app-shell stuff

emmenko:
Closes #747


montezume:
Cool! So we have 3 app kit related Percy projects then right?

emmenko:
Yes 3 projects

montezume:
Thanks for driving this 🎉 

If we can do it here, we could do it in the MC-FE too 🤔 

BTW, let's disable the Firefox snaps in percy

montezume:
BTW -> it says 1 snapshot, yet you do two snapshots? Maybe we need to add an await or something? 🤔

cypress/integration/playground/state-machines.js
```diff
@@ -1,10 +1,19 @@
...
       cy.login({ redirectToUri: URL_STATE_MACHINES });
       cy.getByText('State Machines').should('exist');
+      cy.percySnapshot();
+
+      // Go to details page
+      cy.getAllByText('Initial')
+        .first()
+        .click();
+      cy.url().should('include', URL_STATE_MACHINES_ID);
+      cy.getByText('LineItemState').should('exist');
```
montezume: ah, this isn't getting run? 🤔

cypress/integration/template-starter/main.js
```diff
@@ -4,5 +4,6 @@ describe('Main view', () => {
   it('should render page', () => {
     cy.login({ redirectToUri: URL_EXAMPLES_STARTER });
     cy.getByText('Hello, world').should('exist');
+    cy.percySnapshot();
```
montezume: seems like this is the only snapshot uploaded...

emmenko:
> BTW, let's disable the Firefox snaps in percy

Oh, I thought it was disabled. Done

cypress/integration/template-starter/main.js
emmenko: Hmm weird...

emmenko: It seems like they create 2 builds. Not sure if this is the intended behaviour

image

montezume: Hm yeah, I don't think that's what we want 🤔

emmenko: It kind of makes sense though because we run cypress on 2 different apps...maybe we need to have 2 different projects

emmenko: I created 2 projects, it should work fine now

montezume:
We should be getting 3 diffs, not 2 right?

emmenko:
All good now, check both percy projects

montezume:
Thanks!

PR #937, opened 2019-08-07
chore: update to stable react-intl v3

emmenko:
https://github.com/formatjs/react-intl/releases


emmenko:
The current issue on CI is this one: https://github.com/formatjs/react-intl/issues/1395

montezume:
So we wait 😕 ?

tdeekens:
Sounds like being will start having a `react-intl-transpiled` repository at some point which syncs with the untranspiled upstream if the owner will not provide bundled assets.

emmenko:
The fact that they don't ship transpiled code is really annoying. See how many "workarounds" we had to do to make it work.

I really like the new features of react-intl but if this keeps going it will be really annoying to maintain.

Maybe we should bundle react-intl on our own...

tdeekens:
> Maybe we should bundle react-intl on our own...

That's what I meant. I did that in the past once with `pdf.js`. You just create a `react-intl-bunled` repo and have it has a dep there and roll everything up and publish it. But let's maybe give them 1-2 more weeks. Maybe they make up their mind and our current version works for us, right?

montezume:
AFAIK we aren't having any problems with the beta version right? So let's stick with that and hope for a solution in the next weeks. I bet there will be significant pressure from the community to make it work more "out of the box" with CRA...........

emmenko:
Yeah I did the same for https://github.com/emmenko/omit-empty-es

I'm trying to apply the patch, let's see if it works...

emmenko:
Oh I have to apply it also for gatsby...

packages/mc-scripts/config/create-webpack-config-for-development.js
```diff
@@ -308,6 +308,12 @@ module.exports = ({
           },
         ],
       },
+      // Fix for react-intl
+      // https://github.com/formatjs/formatjs/issues/143#issuecomment-518774786
+      {
+        test: /\.mjs$/,
+        type: 'javascript/auto',
+      },
```
tdeekens: As a result we will transpile any `mjs` file anywhere in `node_modules` again?

emmenko: Hmm I don't think so. The way I understand it is that we're just telling webpack to treat `mjs` as "normal" JS modules so it'll look for `js` modules.
This is just a guess, I couldn't find any good resource about this stuff...

emmenko: https://webpack.js.org/configuration/module/#ruletype

tdeekens: Hm, okay. Just want to make sure we don't change bundling behaviour for all deps cause one is borked.

emmenko: I can test it in the mc-fe repo

montezume: we could also explicity target react-intl...

emmenko: Seems to work fine.

PR #938, opened 2019-08-07
chore: update ui-kit

montezume:
#### Summary

Updates `UI-Kit` to next.


montezume:
I think there's a problem with `TimeInput`

We should put on hold for now until investigate.

PR #941, opened 2019-08-08
feat(permissions/authorized): add support for demandedDataFences for injectAuthorized

adnasa:
#### Summary 

- adds support for `demandedDataFence` on `injectAuthorized`
- return `isAuthorized` as closure when `dataFence` is demanded on client

##### Status quo

Until now, we supported `demandedPermissions` and `demandedActionRights` which key-value objects and easy to with their actualPermissions and actualActionRights counterparts from the MC-API

##### With dataFences

With `dataFences` however, we're adding another dimension to match with. Namely the values stored a key on a resource. This is to fulfil the following scenario:

```
GIVEN User A
WITH no projectPermission ManageOrders
AND WITH dataFencePermission ManageOrders
for orders of store A

GIVEN Order of store A 
THEN User A can manage order in the order details
```

----

#### For the Future

At the moment, the problem I'm solving is allowing the user to view and manage orders.. **for the OrderDetails**, and not the OrderList.
This is a small limitation that we should iterate on, because the need and use case is not there yet from the MC-Frontend side.

But, with the proposed change on the API and especially that `isAuthorized` is closure as a result of `demandedDataFence`, the support to delegate permission on the UI wouldn't require a lot of effort to add. It's just not there yet.

---

#### Can we see this in action?

I have a written POC [here](https://github.com/commercetools/merchant-center-application-kit/compare/adnasa-add-data-fence-injectAuthorized...adnasa-data-fence-poc?expand=1) . **PLEASE** have a look at this to see why the closure is needed

The component tree on the client side is not perfect, however I'm happy with what is there now. I look forward to add branching on MC-FE

This is a draft PR. With approval on the approach, I will

- [ ] Update tests
- [ ] Update docs
- Anything else?

### DIFF
--- a/packages/permissions/README.md
+++ b/packages/permissions/README.md
@@ -134,7 +134,7 @@ Like `branchOnPermissions`, but without the `FallbackComponent`.
 
 ```js
 injectAuthorized(
-  permissions: [Permission],
+  permissions: [Permissions.ManageOrders],
   options: ?Object
 ): HoC
 ```
--- a/packages/permissions/README.md
+++ b/packages/permissions/README.md
@@ -148,7 +148,7 @@ injectAuthorized(
     (default `false`)
   - `actionRights`: (_optional_) an array of action rights (mentioned above)
   - `dataFences`: (_optional_) an array of `DataFence` (mentioned above)
-  - `getSelectDataFenceDataByType`: (_optional_) if `dataFences` option is specified, `getSelectDataFenceDataByType` will be called with the component's `ownProps`. The result is a mapper function which we can use to pick the necessary data to compare with depending on the `type` of the dataFence.
+  - `getSelectDataFenceDataByType`: (_optional_ unless `dataFences` is specified) the function is called with the component `props` and must return a new function that will be called on each specified `dataFence` and should return the mapped value specific to the data fence. The returned function usually contains a switch-case logic based on the data fence `type`, with the return value based on the `props`. For example, with `type: 'order'`, the return value would be something like `ownProps.order.storeRef.key`.
 
 ### Example
 
--- a/packages/permissions/README.md
+++ b/packages/permissions/README.md
@@ -160,7 +160,7 @@ const InputField = props => (
   />
 );
 
-injectAuthorized(['ViewProducts', 'ViewOrders'])(InputField);
+injectAuthorized([Permissions.ViewProducts, Permissions.ViewOrders])(InputField);
 ```
 
 ## `<RestrictedByPermissions>`
--- a/packages/permissions/README.md
+++ b/packages/permissions/README.md
@@ -190,7 +190,7 @@ match, otherwise a fallback component.
 - `permissions`: an array of `Permission`, requested by the component
 - `actionRights`: (_optional_) an array of `ActionRight`, requested by the component
 - `dataFences`: (_optional_) an array of `DataFence`, requested by the component
-- `selectDataFenceDataByType`: (_optional_) a mapper function which we can use to pick the necessary data to compare with depending on the `type` of the dataFence when `dataFences` option is specified
+- `selectDataFenceDataByType`: (_optional_ unless `dataFences` is specified) the function is called on each specified `dataFence` and should return the mapped value specific to the data fence. The function usually contains a switch-case logic based on the data fence `type`, with the return value based on the `props`. For example, with `type: 'order'`, the return value would be something like `ownProps.order.storeRef.key`.
 - `unauthorizedComponent`: (_optional_) a function return an React element to be
   rendered in case the permissions don't match
 - `render`: (_optional_) a function returning an React element or a node to be
--- a/packages/permissions/README.md
+++ b/packages/permissions/README.md
@@ -6,9 +6,10 @@
 
 React components to declaratively handle MC permissions.
 
-- `branchOnPermissions()`
-- `<RestrictedByPermissions>`
+- `useIsAuthorized()`
 - `injectAuthorized()`
+- `<RestrictedByPermissions>`
+- `branchOnPermissions()`
 
 ## Install
 
@@ -95,6 +96,131 @@ export const DATA_FENCES = {
 };
 ```
 
+## `useIsAuthorized(options)`
+
+```js
+useIsAutorized(options: HookOptions): boolean
+```
+
+### Named arguments
+
+- `demandedPermissions`: an array of `Permission`, requested for the child component to
+  be allowed to render
+- `demandedAtionRights`: (_optional_) an array of action rights (mentioned above)
+- `demandedDataFences`: (_optional_) an array of `DataFence` (mentioned above)
+- `getSelectDataFenceDataByType`: (_optional_) if `dataFences` option is specified, `getSelectDataFenceDataByType` will be called with the component's `ownProps`. The result is a mapper function which we can use to pick the necessary data to compare with depending on the `type` of the dataFence.
+- `shouldMatchSomePermissions`: (_optional_) determines if _some_ or _every_ requested permission should match (default `false`)
+
+### Example
+
+```js
+const Test = () => {
+  const canAccessProducts = useIsAuthorized({
+    demandedPermissions: ['ManageProducts'],
+  });
+  if (canAccessProducts) {
+    return <span>{'Authorized'}</span>;
+  }
+  return <span>{'Not authorized'}</span>;
+};
+```
+
+## `injectAuthorized(permissions, [options])`
+
+Like `branchOnPermissions`, but without the `FallbackComponent`.
+
+> This is meant to be used in cases where only the `isAuthorized` prop is needed
+> to be injected.
+
+```js
+injectAuthorized(
+  permissions: [Permission],
+  options: ?Object
+): HoC
+```
+
+### Arguments
+
+- `permissions`: an array of `Permission`, requested for the child component to
+  be allowed to render
+- `options` (_optional_)
+  - `shouldMatchSomePermissions`: (_optional_) determines if _some_ or _every_ requested permission should match
+    (default `false`)
+  - `actionRights`: (_optional_) an array of action rights (mentioned above)
+  - `dataFences`: (_optional_) an array of `DataFence` (mentioned above)
+  - `getSelectDataFenceDataByType`: (_optional_) if `dataFences` option is specified, `getSelectDataFenceDataByType` will be called with the component's `ownProps`. The result is a mapper function which we can use to pick the necessary data to compare with depending on the `type` of the dataFence.
+
+### Example
+
+```js
+const InputField = props => (
+  <Input
+    //...
+    disabled={!props.isAuthorized}
+  />
+);
+
+injectAuthorized(['ViewProducts', 'ViewOrders'])(InputField);
+```
+
+## `<RestrictedByPermissions>`
+
+A React component that will render its children if the requested permissions
+match, otherwise a fallback component.
+
+> This is the React version of the `branchOnPermissions` HoC.
+
+```js
+<RestrictedByPermissions
+  permissions={[]}
+  unauthorizedComponent={Unauthorized}
+  shouldMatchSomePermissions={true}
+>
+  <MyAuthorizedComponent />
+</RestrictedByPermissions>
+```
+
+### Props
+
+- `permissions`: an array of `Permission`, requested by the component
+- `actionRights`: (_optional_) an array of `ActionRight`, requested by the component
+- `dataFences`: (_optional_) an array of `DataFence`, requested by the component
+- `selectDataFenceDataByType`: (_optional_) a mapper function which we can use to pick the necessary data to compare with depending on the `type` of the dataFence when `dataFences` option is specified
+- `unauthorizedComponent`: (_optional_) a function return an React element to be
+  rendered in case the permissions don't match
+- `render`: (_optional_) a function returning an React element or a node to be
+  rendered in case the permissions match
+  - Note: when a function is passed it will get an invoked with an object
+    containing `isAuthorized` e.g. `({ isAuthorized }) => <div />`
+- `children`: (_optional_) a function returning an React element or a node to be
+  rendered in case the permissions match
+  - Note: when a function is passed it will get an invoked with an object
+    containing `isAuthorized` e.g. `({ isAuthorized }) => <div />`
+- `shouldMatchSomePermissions`: (_optional_) determines if _some_ or _every_
+  requested permission should match (default `false`)
+
+### Examples
+
+```js
+const Unauthorized = () => <p>{'No permissions to see this'}</p>;
+const Dashboard = () => (
+  <div>
+    <RestrictedByPermissions
+      permissions={['ViewProducts']}
+      unauthorizedComponent={Unauthorized}
+    >
+      <TopFiveProducts />
+    </RestrictedByPermissions>
+    <RestrictedByPermissions
+      permissions={['ViewProducts', 'ViewOrders']}
+      shouldMatchSomePermissions={true}
+    >
+      {({ isAuthorized }) => <RevenueChart isDisabled={!isAuthorized} />}
+    </RestrictedByPermissions>
+  </div>
+);
+```
+
 ## `branchOnPermissions(permissions, [FallbackComponent], [options])`
 
 A HoC that will render a fallback component if the requested permissions don't
@@ -165,100 +291,4 @@ const ReadOnlyInput = () => // ...
 const Input = () => // ...
 
 branchOnPermissions(['ViewProducts'], ReadOnlyInput)(Input)
-```
-
-## `<RestrictedByPermissions>`
-
-A React component that will render its children if the requested permissions
-match, otherwise a fallback component.
-
-> This is the React version of the `branchOnPermissions` HoC.
-
-```js
-<RestrictedByPermissions
-  permissions={[]}
-  unauthorizedComponent={Unauthorized}
-  shouldMatchSomePermissions={true}
->
-  <MyAuthorizedComponent />
-</RestrictedByPermissions>
-```
-
-### Props
-
-- `permissions`: an array of `Permission`, requested by the component
-- `actionRights`: (_optional_) an array of `ActionRight`, requested by the component
-- `dataFences`: (_optional_) an array of `DataFence`, requested by the component
-- `selectDataFenceDataByType`: (_optional_) a mapper function which we can use to pick the necessary data to compare with depending on the `type` of the dataFence when `dataFences` option is specified
-- `unauthorizedComponent`: (_optional_) a function return an React element to be
-  rendered in case the permissions don't match
-- `render`: (_optional_) a function returning an React element or a node to be
-  rendered in case the permissions match
-  - Note: when a function is passed it will get an invoked with an object
-    containing `isAuthorized` e.g. `({ isAuthorized }) => <div />`
-- `children`: (_optional_) a function returning an React element or a node to be
-  rendered in case the permissions match
-  - Note: when a function is passed it will get an invoked with an object
-    containing `isAuthorized` e.g. `({ isAuthorized }) => <div />`
-- `shouldMatchSomePermissions`: (_optional_) determines if _some_ or _every_
-  requested permission should match (default `false`)
-
-### Examples
-
-```js
-const Unauthorized = () => <p>{'No permissions to see this'}</p>;
-const Dashboard = () => (
-  <div>
-    <RestrictedByPermissions
-      permissions={['ViewProducts']}
-      unauthorizedComponent={Unauthorized}
-    >
-      <TopFiveProducts />
-    </RestrictedByPermissions>
-    <RestrictedByPermissions
-      permissions={['ViewProducts', 'ViewOrders']}
-      shouldMatchSomePermissions={true}
-    >
-      {({ isAuthorized }) => <RevenueChart isDisabled={!isAuthorized} />}
-    </RestrictedByPermissions>
-  </div>
-);
-```
-
-## `injectAuthorized(permissions, [options])`
-
-Like `branchOnPermissions`, but without the `FallbackComponent`.
-
-> This is meant to be used in cases where only the `isAuthorized` prop is needed
-> to be injected.
-
-```js
-injectAuthorized(
-  permissions: [Permission],
-  options: ?Object
-): HoC
-```
-
-### Arguments
-
-- `permissions`: an array of `Permission`, requested for the child component to
-  be allowed to render
-- `options` (_optional_)
-  - `shouldMatchSomePermissions`: (_optional_) determines if _some_ or _every_ requested permission should match
-    (default `false`)
-  - `actionRights`: (_optional_) an array of action rights (mentioned above)
-  - `dataFences`: (_optional_) an array of `DataFence` (mentioned above)
-  - `getSelectDataFenceDataByType`: (_optional_) if `dataFences` option is specified, `getSelectDataFenceDataByType` will be called with the component's `ownProps`. The result is a mapper function which we can use to pick the necessary data to compare with depending on the `type` of the dataFence.
-
-### Example
-
-```js
-const InputField = props => (
-  <Input
-    //...
-    disabled={!props.isAuthorized}
-  />
-);
-
-injectAuthorized(['ViewProducts', 'ViewOrders'])(InputField);
 ```
--- a/packages/permissions/README.md
+++ b/packages/permissions/README.md
@@ -71,6 +71,30 @@ export const ACTION_RIGHTS = {
 };
 ```
 
+## Available dataFences
+
+A `DataFence` is represented as an object with the shape `{ group: string, name: string, type: string }`. currently there is only dataFences of type `store`. The `group` relates to what the permission applies to and the `name` is the permission itself. Currently the following datafences for the `group: 'orders'` exist:
+
+- `ViewOrders`
+- `ManageOrders`
+
+We recommend to put the dataFences used by your application into a `constants.js` file.
+
+```js
+export const DATA_FENCES = {
+  ViewOrders: {
+    type: 'store',
+    group: 'orders',
+    name: 'ViewOrders',
+  },
+  ManageOrders: {
+    type: 'store',
+    group: 'orders',
+    name: 'ManageOrders',
+  },
+};
+```
+
 ## `branchOnPermissions(permissions, [FallbackComponent], [options])`
 
 A HoC that will render a fallback component if the requested permissions don't
@@ -83,6 +107,8 @@ branchOnPermissions(
   options?: {
     shouldMatchSomePermissions: boolean,
     actionRights?: [ActionRight],
+    dataFences?: [DataFence],
+    getSelectDataFenceDataByType?: ownProps => func
   }
 ): HoC
 ```
@@ -95,7 +121,9 @@ branchOnPermissions(
 - `options` (_optional_)
   - `shouldMatchSomePermissions`: determines if _some_ or _every_ requested permission should match
     (default `false`)
-  - `actionRights`: an array of action rights (mentioned above)
+  - `actionRights`: an array of `ActionRight` (mentioned above)
+  - `dataFences`: an array of `DataFence` (mentioned above)
+  - `getSelectDataFenceDataByType`: if `dataFences` option is specified, `getSelectDataFenceDataByType` will be called with the component's `ownProps`. The result is a mapper function which we can use to pick the necessary data to compare with depending on the `type` of the dataFence.
 
 ### Example
 
@@ -105,6 +133,30 @@ const TopFiveProducts = () =>
   // ...
 
   branchOnPermissions(['ViewProducts', 'ViewOrders'])(TopFiveProducts);
+```
+
+```js
+// Only render a component when the user has the the requested dataFence
+const LastOrder = ({ order : { store: "Germany" }}) =>
+  // ...
+
+  branchOnPermissions([], options: {
+    dataFences: [
+      {
+        name: 'ManageOrders',
+        group: 'orders',
+        type: 'store'
+      }
+    ],
+    getSelectDataFenceDataByType: ownProps => ({ type }) => {
+      switch (type) {
+        case 'store':
+          return [ownProps.order.store];
+        default:
+          return null
+      }
+    }
+  })(LastOrder);
 ```
 
 ```js
@@ -135,7 +187,9 @@ match, otherwise a fallback component.
 ### Props
 
 - `permissions`: an array of `Permission`, requested by the component
-- `actionRights`: an array of `ActionRight`, requested by the component
+- `actionRights`: (_optional_) an array of `ActionRight`, requested by the component
+- `dataFences`: (_optional_) an array of `DataFence`, requested by the component
+- `selectDataFenceDataByType`: (_optional_) a mapper function which we can use to pick the necessary data to compare with depending on the `type` of the dataFence when `dataFences` option is specified
 - `unauthorizedComponent`: (_optional_) a function return an React element to be
   rendered in case the permissions don't match
 - `render`: (_optional_) a function returning an React element or a node to be
@@ -190,9 +244,11 @@ injectAuthorized(
 - `permissions`: an array of `Permission`, requested for the child component to
   be allowed to render
 - `options` (_optional_)
-  - `shouldMatchSomePermissions`: determines if _some_ or _every_ requested permission should match
+  - `shouldMatchSomePermissions`: (_optional_) determines if _some_ or _every_ requested permission should match
     (default `false`)
-  - `actionRights`: an array of action rights (mentioned above)
+  - `actionRights`: (_optional_) an array of action rights (mentioned above)
+  - `dataFences`: (_optional_) an array of `DataFence` (mentioned above)
+  - `getSelectDataFenceDataByType`: (_optional_) if `dataFences` option is specified, `getSelectDataFenceDataByType` will be called with the component's `ownProps`. The result is a mapper function which we can use to pick the necessary data to compare with depending on the `type` of the dataFence.
 
 ### Example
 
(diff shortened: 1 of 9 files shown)


packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -65,21 +76,34 @@ const Authorized = (props: Props) => {
...
+type TResourceToApplyDataFence = {
+  id: string;
+};
+
 const injectAuthorized = <
-  OwnProps extends { isAuthorized?: boolean },
+  OwnProps extends {
+    isAuthorized?:
+      | boolean
+      | ((resourceToApplyDataFence: TResourceToApplyDataFence) => boolean);
```
tdeekens: Maybe name this `authorizationContext` and have the type be `type TAuthorizationContext = TResourceToApplyDataFence` so we can add things to the union type later?

packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -65,21 +76,34 @@ const Authorized = (props: Props) => {
...
+    : hasDemandedPermissions && hasDemandedActionRights;
+  return <React.Fragment>{props.render(isAuthorized)}</React.Fragment>;
 };
 Authorized.displayName = 'Authorized';
 Authorized.defaultProps = defaultProps;
 
 type TInjectAuthorizedOptions = {
   shouldMatchSomePermissions?: boolean;
   actionRights?: TDemandedActionRight[];
+  dataFence?: TDemandedDataFence;
```
tdeekens: Can you elaborate why we should not support `dataFences` here?

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -141,3 +172,71 @@ export const getInvalidPermissions = (
...
+  const actualDataFenceByResourceGroup = get(
+    options.actualDataFences,
+    `${options.demandedDataFence.type}.${options.demandedDataFence.group}`
+  );
+  // { canManageOrders: { values: [] }}
+  if (actualDataFenceByResourceGroup) {
+    const hasDemandedDataFence = Object.entries(
+      actualDataFenceByResourceGroup
+    ).every(([actualDataFenceName, actualDataFence]): boolean => {
+      switch (options.demandedDataFence.type) {
```
adnasa: what happens in here could be determined by the client.
that is... the implementation of `hasDemandedDataFenceForStore` can move to MC-FE if we are not happy to have it on app-kit

packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -88,22 +112,26 @@ const injectAuthorized = <
...
-          actualActionRights={applicationContext.actionRights}
-          render={isAuthorized => (
-            <Component {...props} {...{ [propName]: isAuthorized }} />
-          )}
-        />
-      )}
-    />
-  );
+  const WrappedComponent = (props: OwnProps) => {
+    return (
```
adnasa: ignore this.. I will revert it. this is a result of a console.logging

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -141,3 +172,71 @@ export const getInvalidPermissions = (
...
+
+const hasDemandedDataFenceForStore = (
+  resourceToApplyDataFence: TResourceToApplyDataFence,
+  options: {
+    actualDataFence: TDataFenceByPermissionName;
+    actualDataFenceName: string;
+    demandedDataFence: TDemandedDataFence;
+  }
+): boolean => {
+  const hasDemandedPermission = hasPermission(options.demandedDataFence.name, {
```
tdeekens: So `options.demandedDataFence.name` and `options.actualDataFenceName` are two different things?

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -141,3 +172,71 @@ export const getInvalidPermissions = (
...
+    [options.actualDataFenceName]: true,
+  });
+  if (!hasDemandedPermission) return false;
+  switch (options.demandedDataFence.group) {
+    case 'orders': {
```
tdeekens: Maybe we can have a const `DATA_FENCE_GROUPS` to match with here?

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -141,3 +172,71 @@ export const getInvalidPermissions = (
...
+      return options.actualDataFence.values.includes(
+        resourceToApplyDataFence.storeRef &&
+          resourceToApplyDataFence.storeRef.key
+          ? resourceToApplyDataFence.storeRef.key
+          : ''
+      );
+    }
+    default:
+      return false;
```
tdeekens: Should we log, throw, use Sentry or anything in this case?

packages/permissions/src/utils/has-permissions.ts
adnasa: actualDataFenceName derives from `actualDataFences`, and `demanded*` from what was demanded

packages/permissions/src/components/authorized/authorized.tsx
adnasa: yes sure thing

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -141,3 +172,71 @@ export const getInvalidPermissions = (
     isNil(actualPermissions[toCanCase(demandedPermission)])
   );
 };
+
+const hasDemandedDataFenceForStore = (
+  resourceToApplyDataFence: TResourceToApplyDataFence,
```
tdeekens: Here we could then do `authorizationContext` which is `TResourceToApplyDataFence`. That might even nicely work then with type guards in teh future.

packages/permissions/src/utils/has-permissions.ts
tdeekens: I think it should be in app-kit as custom apps might want to use it too.

packages/permissions/src/utils/has-permissions.ts
tdeekens: Yeah but are they every different? Seems to me that `actualDataFenceName` is just the key while `options.actualDataFenceName.name` is the value which is always similar to the key?

packages/permissions/src/components/authorized/authorized.tsx
adnasa: I thought about it, but I wrote this in parallel with the POC so I decided to keep it simple. 

This wasn't a firm decision.
But if we accept that `isAuthorized` is a closure, this is simple enough to change to an `[]`

adnasa: on it

packages/permissions/src/utils/has-permissions.ts
adnasa: IMO we should throw

packages/permissions/src/components/authorized/authorized.tsx
tdeekens: Not more necessary easy to change later.

Given we pass an list of datafences
Then the `isAuthorized`-fn would also accept an array
Where the index matches to the datafence context

Or it receives an object with the `name` or `group` or whatever of the datafence to bind it too which could be more flexible. Then you could have "two datafences" bound to one context (e.g. one order).

packages/permissions/src/utils/has-permissions.ts
tdeekens: Throwing unmounts the whole app and results in a ice cream. We could throw in dev mode but log to sentry in prod mode.

packages/permissions/src/components/authorized/authorized.tsx
adnasa: > Or it receives an object with the name or group or whatever of the datafence to bind it too which could be more flexible.

You lost me here.
The `dataFence` on this commit already is an object with `name` and `group`

packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -19,19 +20,29 @@ type TDemandedActionRight = {
...
 type TActionRight = {
   [key: string]: boolean;
 };
 type TActionRights = {
   [key: string]: TActionRight;
 };
-
+type TDataFence = {
+  // E.g. { store: {...} }
+  [key: string]: {};
```
amine-benselim: how about 
```typescript
type TDataFenceKey = "store"

type TDataFence = {
 [key: TDataFenceKey]: {};
}
```

then in the future we can take advantage of typescript unions:
`type TDataFenceType = "store" | "price" `

packages/permissions/src/components/authorized/authorized.tsx
tdeekens: I'll come by.

adnasa: > Given we pass an list of datafences
> Then the isAuthorized-fn would also accept an array
> Where the index matches to the datafence context

I don't think we should do this, where `dataFence as a list = isAuthorized accepts a list`.

injectAuthorized is not something we use on the list itself, but on the actionable buttons like **Bulk update**. So given you have a list of orders where you have N with `View*` and N are `Manage*`, and injectAuthorized is used on `BulkActions` then you define one `demandedDataFence` and match whether all selected Orders on the list matches this.

tdeekens: Thanks. Got it now. We can accept a list of `dataFences` but `isAuthorized`-fn will always receive one context value to evaluate all fences against.

adnasa: talked to tobi IRL and it's worth clarifying it here.

`demandedDataFences` (list) is a list of:

- dataFences applied to a single **context** (this context is called **group** in the `allAppliedDataFences`). This means that it is bound to a single resource type.
- dataFences can be of different datafence-types (right now we have only `Store`, but there is no reason to allow for more, so that in the future, we only need to implement that the logic of which key we want to read from the resource).

This means, that the following won't work as you would expect

```js
export default injectAuthorized(
  [PERMISSIONS.ViewOrders],
  {
    dataFences: [
      {
        type: 'store',
        group: 'orders',
        name: PERMISSIONS.ManageOrders,
      },
      {
        type: 'category',
        group: 'products',
        name: PERMISSIONS.AddCategory,
      },
    ],
  },
  'isAuthorizedToManageOrders'
)(OrderDetailsGeneralInfoTab)
```

because what you are doing, is that you are demanding that:

1. a user needs to have ViewOrders in `orders` for `isAuthorizedToManageOrders` to be `true`
2. a user also needs to have `AddCategory` in products for `isAuthorizedToManageOrders` to be `true`

packages/permissions/src/components/authorized/authorized.tsx
adnasa: that's a good idea. however, the `fetch-project` already supports passing down multiple data-fence type so we could keep this.

packages/permissions/src/utils/has-permissions.ts
adnasa: yes they are different. the `demanded*` can be `ManageOrders` whereas the `actual*` can be `ViewOrders`

adnasa:
I forgot that i need to add support on `RestrictedByPermissions` and `branchOnPermissions`. will have a look

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -16,6 +22,37 @@ type TActionRight = {
 type TActionRights = {
   [key: string]: TActionRight;
 };
+type TDataFence = {
```
amine-benselim: i'm no expert in types but the repetition is bugging me. can we have them in a central place and import them in these files ?

emmenko: As long as it's not a React component yes, otherwise the babel plugin to generate PropTypes doesn't work if types are in external files.

adnasa: so where does this leave me?
because this type is used in `authorized` (react) and `has-permissions` (not react)

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -141,3 +178,98 @@ export const getInvalidPermissions = (
...
+        ).every(([name, value]): boolean => {
+          switch (demandedDataFence.type) {
+            case 'store':
+              return hasDemandedDataFenceForStore(authorizationContext, {
+                actualDataFence: { name, dataFenceValue: value },
+                demandedDataFence,
+              });
+            default: {
+              if (process.env.NODE_ENV !== 'production') {
+                throw new Error(
```
emmenko: This is not necessary since we already use `reportErrorToSentry`

packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -19,19 +20,38 @@ type TDemandedActionRight = {
...
+  storesRef?: TKeyReference[];
+};
+type TAuthorizationContext = TResourceToApplyDataFence;
 type Props = {
   shouldMatchSomePermissions?: boolean;
   demandedPermissions: TPermissionName[];
   demandedActionRights?: TDemandedActionRight[];
   actualPermissions: TPermissions | null;
   actualActionRights: TActionRights | null;
+  demandedDataFences?: TDemandedDataFence[];
```
emmenko: Nitpick: move up to have them grouped by name

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -141,3 +178,98 @@ export const getInvalidPermissions = (
...
+    actualDataFence: TActualDataFence;
+    demandedDataFence: TDemandedDataFence;
+  }
+): boolean => {
+  const hasDemandedPermission = hasPermission(options.demandedDataFence.name, {
+    [options.actualDataFence.name]: true,
+  });
+  if (!hasDemandedPermission) return false;
+  switch (options.demandedDataFence.group) {
+    case DATA_FENCE_GROUPS.ORDERS: {
```
emmenko: Why does it need to be hardcoded? 🤔

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -141,3 +178,98 @@ export const getInvalidPermissions = (
...
+      return options.actualDataFence.dataFenceValue.values.includes(
+        authorizationContext.storeRef && authorizationContext.storeRef.key
+          ? authorizationContext.storeRef.key
+          : ''
+      );
+    }
+    default: {
+      if (process.env.NODE_ENV !== 'production') {
+        throw new Error(
```
emmenko: This is not necessary since we already use `reportErrorToSentry`

packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -65,21 +85,32 @@ const Authorized = (props: Props) => {
...
   actionRights?: TDemandedActionRight[];
+  dataFences?: TDemandedDataFence[];
 };
+
 const injectAuthorized = <
-  OwnProps extends { isAuthorized?: boolean },
+  OwnProps extends {
+    isAuthorized?:
+      | boolean
+      | ((authorizationContext: TAuthorizationContext) => boolean);
```
emmenko: Why do we need this?

packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -65,21 +85,32 @@ const Authorized = (props: Props) => {
...
 
-  return (
-    <React.Fragment>
-      {props.render(hasDemandedPermissions && hasDemandedActionRights)}
-    </React.Fragment>
-  );
+  const isAuthorized = Boolean(
+    props.demandedDataFences && props.demandedDataFences.length
+  )
+    ? createHasAppliedDataFence({
```
emmenko: Nitpick: can we move this to a variable like the other ones? I think it's a bit easier to read and understand the logic.

```js
const hasDemandedPermissions = /**/;
const hasDemandedActionRights = /**/;

if (Boolean(props.demandedDataFences && props.demandedDataFences.length > 0)) {
  const hasDemandedDataFences = hasAppliedDataFences(/**/);
  return <>{props.render(hasDemandedDataFences)}</>
}
return <>{props.render(hasDemandedPermissions && hasDemandedActionRights)}</>
```

packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -19,19 +20,38 @@ type TDemandedActionRight = {
...
   [key: string]: TActionRight;
 };
-
+type TDataFence = {
+  [key: string]: {};
+};
+type TResourceToApplyDataFence = {
+  id: string;
+  // Orders
+  storeRef?: {
```
emmenko: Maybe I'm missing something but why do we need to hardcode those things?

packages/permissions/src/utils/has-permissions.ts
emmenko: You can export the type from `authorized` and use it in the utility functions.

But we can also do that as a separate step...

packages/permissions/src/components/authorized/authorized.tsx
tdeekens: That‘s the question with the cool answer. I think Adnan can explain it best.

adnasa:
```js
injectAuthorized(
  ['ViewOrders'],
  {
    dataFences: [
      {
        // there is no way for me to pass down the resource here
        // e.g an Order, which has the values (`storeRef.key`) that I need to determine if this key is part of the data fence 
        // so we pass a closure
        name: 'ManageOrders',
        group: 'orders',
        type: 'store'
      }
    ]
  },
  'canManageOrders',
)
```

if you look at the POC, it will be clearer

https://github.com/commercetools/merchant-center-application-kit/blob/10aa34928bfe40aa633091c0f1560a0c639f38dc/playground/src/components/order-details/order-details-general-info-tab.js#L11-L15

packages/permissions/src/components/authorized/authorized.tsx
adnasa: How are these hardcoded?

packages/permissions/src/utils/has-permissions.ts
adnasa: I don't get it. what's hardcoded?
we need the `group` to determine how to match the resource value to the given datafence

packages/permissions/src/components/authorized/authorized.tsx
emmenko: With the premise that I had a call with Adnan yesterday about this, where we clarified different things and discussed further suggestions, I've been also thinking a bit about the injected `isAuthorized` prop optionally being a function.

From a static typing point of view, it's always tricky to deal with types being of different shapes/types (_represented as unions_) as it might require additional logic to determine the correct type. For example, we would need to use generics, type guards, etc.
Therefore, if possible, we should make our types as unambiguous as possible.

As for the `isAuthorized` prop being either a `boolean` or a `function`, we also need to consider the different options that the package providers, namely:
- `injectAuthorized`
- `useIsAuthorized`
- `<Authorized>`
- `<RestrictedByPermissions>`,
- `branchOnPermissions`

The question is: how can we make it work for all those cases? Does it need to be a function in the first place?

For example, I can image the hook being something like this:

```js
const TestComponent = (props: TestProps) => {
  const isAuthorized = useIsAuthorized({
    demandedPermissions: ['ManageOrders'],
    demandedDataFences: [
      {
        name: 'ManageOrders',
        group: 'orders',
        type: 'store',
      },
    ],
    selectDataFenceDataByType: ({ type }) => {
      switch (type) {
        case 'store':
          return props.order.storeRef.key;
        default:
          return null
      }
    }
  });
  return (
    <ul>
      <li>Is authorized: {isAuthorized ? 'Yes' : 'No'}</li>
    </ul>
  );
};
```

That's because the hook is already used within the component, so we have access to the `props`.

For the `injectAuthorized` we can then have the mapper function accept `ownProps`, like e.g. `react-apollo` does, which we can use to pick the necessary data to map to:

```js
injectAuthorized(
  ['ManageOrders'],
  {
    dataFences: [
      {
        name: 'ManageOrders',
        group: 'orders',
        type: 'store',
      },
    ],
    selectDataFenceDataByType: (ownProps, { type }) => {
      switch (type) {
        case 'store':
          return ownProps.order.storeRef.key;
        default:
          return null
      }
    }
  }
)(Component);
```

This way `isAuthorized` always returns a `boolean` and we can encapsulate all the logic within the permissions packages.

adnasa: I appreciate the feedback on this. I tried this out last night and it definitely simplified the logic inside `injectAuthorized`. 

@emmenko and I will pair on this today.

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -205,41 +205,55 @@ const hasDemandedDataFenceByType = (options: {
...
-      if (actualDataFenceByPermissionGroup) {
-        const hasDemandedDataFence = Object.entries(
-          actualDataFenceByPermissionGroup
-        ).every(([name, value]): boolean => {
+const getDataFenceByPermissionGroup = (
+  dataFences: TDataFences,
+  storeKey: TDataFenceType,
+  groupKey: string
+) => {
+  if (storeKey in dataFences) {
```
adnasa: 👍

emmenko:
@amine-benselim there are still things left to do, to support the data fence options in all other components/functions besides `injectAuthorized`

emmenko:
Let me know if you need help, we can also pair...

amine-benselim:
> @amine-benselim there are still things left to do, to support the data fence options in all other components/functions besides `injectAuthorized`

yes I am working now on adding test cases for this and then I will move to the next components, I will definitely be pinging you :p

emmenko:
Alright, so no reviews yet :)

packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -25,13 +28,38 @@ type TActionRight = {
 type TActionRights = {
   [key: string]: TActionRight;
 };
+// Data fences
+type TDataFenceGroupedByPermission = {
+  // E.g. { canManageOrders: { values: [] } }
+  [key: string]: { values: string[] } | null;
```
amine-benselim: @emmenko do you have any idea why is values an array of strings and not of booleans ? I'm confused of what should be inside now 🤔

tdeekens: Maybe the storeKey?

emmenko: It should be the store keys as far as I understood from Adnan...

amine-benselim: of course! thank you :D

emmenko:
@amine-benselim I've rebased this branch with master.

I also pushed a commit (https://github.com/commercetools/merchant-center-application-kit/pull/941/commits/134a967a4b040ef8d3b054965ef760de49aca562) with some small changes, some fixes and supporting data fences in the `useIsAuthorized` hook

packages/permissions/README.md
```diff
@@ -6,9 +6,10 @@
 
 React components to declaratively handle MC permissions.
 
-- `branchOnPermissions()`
-- `<RestrictedByPermissions>`
+- `useIsAuthorized()`
```
emmenko: I added this and the documentation for it (it was missing). The rest is just reordering

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -141,3 +176,78 @@ export const getInvalidPermissions = (
...
+    const actualDataFenceByPermissionGroup = getDataFenceByPermissionGroup(
+      options.actualDataFences,
+      demandedDataFence.type,
+      demandedDataFence.group
+    );
+
+    if (actualDataFenceByPermissionGroup) {
+      const hasDemandedDataFence = Object.entries(
+        actualDataFenceByPermissionGroup
+      ).every(([name, value]) => {
```
tdeekens: Nit: `dataFenceName` and `dataFenceValue`.

packages/permissions/README.md
```diff
@@ -71,48 +72,95 @@ export const ACTION_RIGHTS = {
...
+- `permissions`: an array of `Permission`, requested for the child component to
+  be allowed to render
 - `options` (_optional_)
-  - `shouldMatchSomePermissions`: determines if _some_ or _every_ requested permission should match
+  - `shouldMatchSomePermissions`: (_optional_) determines if _some_ or _every_ requested permission should match
     (default `false`)
-  - `actionRights`: an array of action rights (mentioned above)
+  - `actionRights`: (_optional_) an array of action rights (mentioned above)
+  - `dataFences`: (_optional_) an array of `DataFence` (mentioned above)
+  - `getSelectDataFenceDataByType`: (_optional_) if `dataFences` option is specified, `getSelectDataFenceDataByType` will be called with the component's `ownProps`. The result is a mapper function which we can use to pick the necessary data to compare with depending on the `type` of the dataFence.
```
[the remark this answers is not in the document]
emmenko:
```suggestion
  - `getSelectDataFenceDataByType`: (_optional_ unless `dataFences` is specified) the function is called with the component `props` and must return a new function that will be called on each specified `dataFence` and should return the mapped value specific to the data fence. The returned function usually contains a switch-case logic based on the data fence `type`, with the return value based on the `props`. For example, with `type: 'order'`, the return value would be something like `ownProps.order.storeRef.key`.
```

packages/permissions/README.md
```diff
@@ -135,7 +183,9 @@ match, otherwise a fallback component.
 ### Props
 
 - `permissions`: an array of `Permission`, requested by the component
-- `actionRights`: an array of `ActionRight`, requested by the component
+- `actionRights`: (_optional_) an array of `ActionRight`, requested by the component
+- `dataFences`: (_optional_) an array of `DataFence`, requested by the component
+- `selectDataFenceDataByType`: (_optional_) a mapper function which we can use to pick the necessary data to compare with depending on the `type` of the dataFence when `dataFences` option is specified
```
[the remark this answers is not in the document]
emmenko: It's usually `select`, but the HOC has `getSelect` because we inject `props`.

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.ts
```diff
@@ -40,7 +68,28 @@ const useIsAuthorized = ({
   const actualActionRights = useApplicationContext<TActionRights | null>(
     applicationContext => applicationContext.actionRights
   );
+  const actualDataFences = useApplicationContext<TDataFences | null>(
+    applicationContext => applicationContext.dataFences
+  );
+
+  // Check first for data fences
+  if (actualDataFences && demandedDataFences && demandedDataFences.length > 0) {
```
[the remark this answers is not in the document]
emmenko: Sure

packages/permissions/README.md
```diff
@@ -71,48 +72,95 @@ export const ACTION_RIGHTS = {
...
-  unauthorizedComponent: ?UnauthorizedComponent,
-  options?: {
-    shouldMatchSomePermissions: boolean,
-    actionRights?: [ActionRight],
+export const DATA_FENCES = {
+  ViewOrders: {
+    type: 'store',
+    group: 'orders',
+    name: 'ViewOrders',
+  },
```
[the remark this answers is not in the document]
amine-benselim: you would demand data fences of the shape:
```js
[{
	type: 'store',
    group: 'orders',
    name: 'ManageOrders'
}, 
{
	type: 'prices',
    group: 'orders',
    name: 'ManageOrders'
},
]
```

and then with `getSelectDataFenceDataByType` you will provide the right selector to use for each type. (I'm unsure how that would look for the prices type for example)

packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -68,6 +68,34 @@ const defaultProps: Pick<Props, 'shouldMatchSomePermissions'> = {
 };
 
 const Authorized = (props: Props) => {
+  // Check first for data fences
```
amine-benselim: @emmenko  why do we need to first check for dataFences ? I would assume that General permissions have precedence.

emmenko: From the discussions I read yesterday from you I assumed that data fences take precedence and the other permissions are ignored.

If I understood it wrong, please feel free to adjust.

amine-benselim: in the permissions UI whenever there is a general permission for example to manage orders, any datafence is overruled by it, ideally `authorized` should mirror the same behaviour.

but in practice, since they are going to be mutually exclusive (https://github.com/commercetools/merchant-center-services/pull/686) the order does not matter since if there is no datafence then the check for `actualDataFences` would fail anyway and that block of code would not run.

packages/permissions/src/utils/has-permissions.ts
amine-benselim: there is no `options.actualDataFenceName `, there is `options.actualDataFence`, maybe you misread the code or maybe I misunderstood you 😄 please elaborate more to help me understand this.

packages/permissions/README.md
```diff
@@ -71,48 +72,95 @@ export const ACTION_RIGHTS = {
 };
 ```
 
-## `branchOnPermissions(permissions, [FallbackComponent], [options])`
+## Available dataFences
 
-A HoC that will render a fallback component if the requested permissions don't
-match with the user permissions.
+A `DataFence` is represented as an object with the shape `{ group: string, name: string, type: string }`. currently there is only dataFences of type `store`. The `group` relates to what the permission applies to and the `name` is the permission itself. Currently the following datafences for the `group: 'orders'` exist:
```
emmenko:
```suggestion
> This is a **beta** feature and should be used with caution. For more information, please [open an issue](https://github.com/commercetools/merchant-center-application-kit/issues/new/choose).

A `DataFence` is represented as an object with the shape `{ group: string, name: string, type: string }`, where:
- `type`: is one of `['store']`
- `name`: is the `Permission` value
- `group`: is one of the resource types (e.g. `orders`)
```

packages/permissions/README.md
```diff
@@ -71,48 +72,95 @@ export const ACTION_RIGHTS = {
...
+useIsAutorized(options: HookOptions): boolean
+```
+
+### Named arguments
+
+- `demandedPermissions`: an array of `Permission`, requested for the child component to
+  be allowed to render
+- `demandedAtionRights`: (_optional_) an array of action rights (mentioned above)
+- `demandedDataFences`: (_optional_) an array of `DataFence` (mentioned above)
+- `getSelectDataFenceDataByType`: (_optional_) if `dataFences` option is specified, `getSelectDataFenceDataByType` will be called with the component's `ownProps`. The result is a mapper function which we can use to pick the necessary data to compare with depending on the `type` of the dataFence.
```
emmenko:
```suggestion
- `selectDataFenceDataByType`: (_optional_ unless `dataFences` is specified) the function is called on each specified `dataFence` and should return the mapped value specific to the data fence. The function usually contains a switch-case logic based on the data fence `type`, with the return value based on the `props`. For example, with `type: 'order'`, the return value would be something like `ownProps.order.storeRef.key`.
```

packages/permissions/README.md
emmenko:
```suggestion
- `selectDataFenceDataByType`: (_optional_ unless `dataFences` is specified) the function is called on each specified `dataFence` and should return the mapped value specific to the data fence. The function usually contains a switch-case logic based on the data fence `type`, with the return value based on the `props`. For example, with `type: 'order'`, the return value would be something like `ownProps.order.storeRef.key`.
```

packages/permissions/README.md
```diff
@@ -171,38 +221,74 @@ const Dashboard = () => (
...
+- `permissions`: an array of `Permission`, requested by the component
+- `UnauthorizedComponent`: (_optional_) a reference to a React component to be
+  rendered in case the permissions don't match
 - `options` (_optional_)
   - `shouldMatchSomePermissions`: determines if _some_ or _every_ requested permission should match
     (default `false`)
-  - `actionRights`: an array of action rights (mentioned above)
+  - `actionRights`: an array of `ActionRight` (mentioned above)
+  - `dataFences`: an array of `DataFence` (mentioned above)
+  - `getSelectDataFenceDataByType`: if `dataFences` option is specified, `getSelectDataFenceDataByType` will be called with the component's `ownProps`. The result is a mapper function which we can use to pick the necessary data to compare with depending on the `type` of the dataFence.
```
emmenko:
```suggestion
  - `getSelectDataFenceDataByType`: (_optional_ unless `dataFences` is specified) the function is called with the component `props` and must return a new function that will be called on each specified `dataFence` and should return the mapped value specific to the data fence. The returned function usually contains a switch-case logic based on the data fence `type`, with the return value based on the `props`. For example, with `type: 'order'`, the return value would be something like `ownProps.order.storeRef.key`.
```

packages/permissions/src/components/authorized/authorized.tsx
tdeekens: True however wouldn't it be confusing if you pass both and only see one being evaluated? Should we maybe also log/reportToSentry in that case saying: hej, only one of those two work?

amine-benselim: There are to people in this scenario, there is the developer of the app that demands both general permissions and dataFences (which is legit as he does not know what kind of permission the end user has). and then there is the end user with a specific kind of permission (either general one or dataFence, but not both).

We can think of `demanded*` as an OR combination, then it should not be confusing to pass both and only see one being evaluated.

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.ts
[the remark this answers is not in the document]
amine-benselim: this is used in two places, one time it returns a boolean and one time it returns a react component. I am not sure if it is worth the abstraction.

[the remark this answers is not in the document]
tdeekens: True, thanks for clarifying!

packages/permissions/src/components/authorized/authorized.tsx
adnasa: Yes it should be storekeys

adnasa:
>Awesome, let's get this in! 🚀

_while Adnan is not here to enjoy it_

h8it

PR #954, opened 2019-08-09
[experiment] feat: add linter for text files

emmenko:
I found this linter to check for spelling mistakes etc. in md files.

Let's see how it goes (I still need to look at the errors in more detail).


emmenko:
The docs team is currently using this: https://errata-ai.github.io/vale/

We'll look into that, therefore we can close this one.

PR #962, opened 2019-08-12
chore: update deps

emmenko:
Closes #955 
Closes #956


tdeekens:
Sorry, I would like to include this in a new PR. I am already updating tons of other deps which got outdated.

PR #963, opened 2019-08-12
chore: update deps

tdeekens:
#### Summary

This updates most deps but not all. React we should do in one separate step I think.


packages/application-shell/package.json
```diff
@@ -51,14 +51,14 @@
     "@commercetools-frontend/sdk": "14.3.0",
     "@commercetools-frontend/sentry": "14.3.0",
     "@commercetools-frontend/url-utils": "14.3.0",
-    "@emotion/core": "10.0.14",
-    "@emotion/styled": "10.0.14",
-    "@flopflip/launchdarkly-adapter": "2.7.2",
+    "@emotion/core": "10.0.15",
+    "@emotion/styled": "10.0.15",
+    "@flopflip/launchdarkly-adapter": "2.8.0",
```
tdeekens: This has changes we don't use yet to defer flag flushing.

packages/application-shell/package.json
```diff
@@ -51,14 +51,14 @@
...
     "@flopflip/memory-adapter": "1.3.2",
-    "@flopflip/react-broadcast": "8.3.3",
-    "@sentry/browser": "5.5.0",
-    "apollo-cache-inmemory": "1.6.2",
-    "apollo-client": "2.6.3",
+    "@flopflip/react-broadcast": "9.0.0",
```
tdeekens: @emmenko contribution to prefix the lifecycle is in this one.

packages/application-shell/package.json
```diff
@@ -95,11 +95,11 @@
   },
   "devDependencies": {
     "@commercetools-frontend/ui-kit": "10.1.2",
-    "@testing-library/react": "8.0.7",
+    "@testing-library/react": "9.1.0",
     "react": "16.8.6",
     "react-apollo": "2.5.8",
     "react-dom": "16.8.6",
-    "react-intl": "3.1.1",
+    "react-intl": "3.1.6",
```
tdeekens: Praying that this won't break the build.

packages/mc-scripts/package.json
```diff
@@ -65,17 +65,17 @@
     "postcss-preset-env": "6.7.0",
     "postcss-reporter": "6.0.1",
     "postcss-safe-parser": "4.0.1",
-    "react-dev-utils": "9.0.1",
+    "react-dev-utils": "9.0.2",
     "shelljs": "0.8.3",
-    "style-loader": "0.23.1",
+    "style-loader": "1.0.0",
```
tdeekens: CleanShot 2019-08-12 at 09 30 37@2x

I can't see us using any of these.

packages/mc-scripts/package.json
```diff
@@ -65,17 +65,17 @@
...
+    "thread-loader": "2.1.3",
     "url-loader": "2.1.0",
-    "webpack": "4.38.0",
-    "webpack-bundle-analyzer": "3.3.2",
-    "webpack-dev-server": "3.7.2",
-    "webpackbar": "3.2.0"
+    "webpack": "4.39.1",
+    "webpack-bundle-analyzer": "3.4.1",
+    "webpack-dev-server": "3.8.0",
+    "webpackbar": "4.0.0"
```
tdeekens: node@6 is a requirement.

packages/react-notifications/package.json
```diff
@@ -54,10 +54,10 @@
   },
   "devDependencies": {
     "@commercetools-frontend/ui-kit": "10.1.2",
-    "@testing-library/react": "8.0.7",
+    "@testing-library/react": "9.1.0",
```
tdeekens: CleanShot 2019-08-12 at 09 37 45@2x

This is the main change. If this doens't break out tests we did a good job. Clean up is now automagic.

packages/application-shell/package.json
emmenko: Actually they removed the `mjs` part, after they got some pressure: https://github.com/formatjs/react-intl/issues/1395#issuecomment-520185742

Can you remove the workarounds we did?

application-templates/starter/package.json
```diff
@@ -33,7 +33,7 @@
   "devDependencies": {
     "@commercetools-frontend/jest-preset-mc-app": "14.3.2",
     "@commercetools-frontend/mc-scripts": "14.3.0",
-    "@testing-library/react": "8.0.7",
+    "@testing-library/react": "9.1.0",
```
emmenko: Ah, interesting:

```
It looks like you're using a version of react-dom that supports the "act" function, but not an awaitable version of "act" which you will need. Please upgrade to at least react-dom@16.9.0 to remove this warning.
```

emmenko: I guess the question is: do we want to bump React as well or should we wait a bit?

packages/application-shell/package.json
tdeekens: Sure. Did so in the last commit. One note: can we maybe be a bit more conservative with these things next time. It doesn't hurt too much adopting a new version 2-4 weeks late to the party and saves us time and energy :)

application-templates/starter/package.json
tdeekens: I would argue we wait and do it in the breaking release of v15 then.

emmenko:
Should we maybe split the updates so that we have more chances of finding out what breaks?

tdeekens:
It was react-intl and some other dep ending up in a hoist "battle" and react@16.9 in there which gave us two renderers which broke things. Should be fixed now.

packages/application-shell/package.json
emmenko: I agree...I honestly didn't expect all those problems with react-intl.

emmenko:
image

Should I have a look at this?

tdeekens:
> Should I have a look at this?

Please do. Thanks!

tdeekens:
This should succeed now. Given it does we can update the website deps if we want to in another PR.

emmenko:
FYI: https://github.com/formatjs/react-intl/issues/1412

packages/application-components/src/components/page-not-found/page-not-found.tsx
```diff
@@ -16,7 +16,7 @@ const PageNotFound = () => (
       >
         {...messages.paragraph1}
         values={{
-          a: msg => (
+          a: (msg: React.ReactNodeArray) => (
```
emmenko: Yeah this also works but it's more of a workaround. Thanks

tdeekens: Yeah I know.

tdeekens: What his this attempting to tell me now?

CleanShot 2019-08-12 at 12 32 51@2x

emmenko: I think we should keep the version pinned to '3.1.1' until the issue I reported is fixed

emmenko: I downgraded react-intl to 3.1.1

emmenko: > What his this attempting to tell me now?

It's related to the bug I reported

tdeekens:
Now we get `Attempted import error: 'TYPE' is not exported from 'intl-messageformat-parser'.` in our E2Es.

jest.test.config.js
```diff
@@ -1,5 +1,3 @@
-const vendorsToTranspile = require('./packages/jest-preset-mc-app/vendors-to-transpile');
```
tdeekens: This can go now. Only react-intl itself needs to be transpiled.

PR #966, opened 2019-08-12
chore: update to react@16.9

tdeekens:
#### Summary

This depends on the dependency update (#963) as otherwise we end up with two react versions again.

Then tests would run and we could see where we are.

This does not update the peer depenedency but I don't think we have to stricly speaking?


emmenko:
> This does not update the peer depenedency but I don't think we have to stricly speaking?

Let's see, maybe it's not necessary

tdeekens:
The trick here is to keep `release/15` up to date with `master` and this up to date with the release branch so we have one version of react. Otherwise tests fail.

package.json
```diff
@@ -144,7 +144,8 @@
     "typescript": "3.5.3"
   },
   "resolutions": {
-    "@types/react": "16.9.1"
+    "@types/react": "16.9.1",
+    "react": "16.9.0"
```
tdeekens: Just to experiment. Without continue to end up with two react versions due to `react-intl`.

tdeekens:
I think now the question may be how much of those warnings come from our libs and should we silence their logging as discussed.

tdeekens:
Done. You want to do the rebase of the apollo branch?

emmenko:
Sure I can do that later. Thanks

PR #971, opened 2019-08-13
refactor(constants): use const assertion

emmenko:
Found this here: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

image


emmenko:
Hmm, it seems that the bot does not wait for Percy to finish. Do we need to configure that? @tdeekens

PR #973, opened 2019-08-13
fix: prop-types for dataFences optional field

emmenko:
Leftover of #941 

I also changed a bit the CI workflow to build the packages first, then run the linters and tests. This way we should be able to catch the prop-types warnings.

### DIFF
--- a/packages/application-shell-connectors/src/components/application-context/application-context.tsx
+++ b/packages/application-shell-connectors/src/components/application-context/application-context.tsx
@@ -73,9 +73,9 @@ type TRawProject = {
   owner: {
     id: string;
   };
-  permissions: TApplicationContextPermissions;
-  actionRights: TApplicationContextActionRights;
-  dataFences: TApplicationContextDataFences;
+  permissions: TApplicationContextPermissions | null;
+  actionRights: TApplicationContextActionRights | null;
+  dataFences: TApplicationContextDataFences | null;
 } & AdditionalProperties;
 type TApplicationContextProject = {
   key: string;
--- a/packages/application-shell/src/components/fetch-project/fetch-project.js
+++ b/packages/application-shell/src/components/fetch-project/fetch-project.js
@@ -23,16 +23,24 @@ import ProjectQuery from './fetch-project.graphql';
  * the fetching logic. Given this mapping needs to be used elsewere feel free
  * to move this over to `permissions` and export it there.
  */
-export const mapAllAppliedToObjectShape = allAppliedShape =>
-  allAppliedShape.reduce(
+export const mapAllAppliedToObjectShape = allAppliedShape => {
+  if (allAppliedShape.length === 0) {
+    return null;
+  }
+  return allAppliedShape.reduce(
     (transformedAllApplied, allApplied) => ({
       ...transformedAllApplied,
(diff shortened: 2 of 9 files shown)


packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -73,9 +73,9 @@ type TRawProject = {
   owner: {
     id: string;
   };
-  permissions: TApplicationContextPermissions;
-  actionRights: TApplicationContextActionRights;
-  dataFences: TApplicationContextDataFences;
+  permissions: TApplicationContextPermissions | null;
+  actionRights: TApplicationContextActionRights | null;
+  dataFences: TApplicationContextDataFences | null;
```
emmenko: All those should be either defined or null.

emmenko: For instance, if we pass an empty object to `dataFences` it's expected that the object has at least a field named `store`. So basically it does not make sense to pass empty objects.

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -73,9 +73,9 @@ type TRawProject = {
   owner: {
     id: string;
   };
-  permissions: TApplicationContextPermissions;
-  actionRights: TApplicationContextActionRights;
-  dataFences: TApplicationContextDataFences;
+  permissions: TApplicationContextPermissions | null;
+  actionRights: TApplicationContextActionRights | null;
```
tdeekens: Hm. Why are they nulled and not `{}`, now? Any specific reason?

packages/application-shell/src/components/fetch-project/fetch-project.js
```diff
@@ -23,16 +23,24 @@ import ProjectQuery from './fetch-project.graphql';
  * the fetching logic. Given this mapping needs to be used elsewere feel free
  * to move this over to `permissions` and export it there.
  */
-export const mapAllAppliedToObjectShape = allAppliedShape =>
-  allAppliedShape.reduce(
+export const mapAllAppliedToObjectShape = allAppliedShape => {
+  if (allAppliedShape.length === 0) {
+    return null;
```
emmenko: So in those map functions we return `null` instead of an empty object

packages/application-shell-connectors/src/components/application-context/application-context.tsx
emmenko: https://github.com/commercetools/merchant-center-application-kit/pull/973#discussion_r313384958

packages/application-shell-connectors/src/components/application-context/application-context.tsx
tdeekens: "it does not make sense" sorry I did doubt sense. I do not dare to counter argue now 😄

packages/application-shell/src/test-utils/test-utils.js
```diff
@@ -70,8 +70,8 @@ const defaultEnvironment = {
 };
 
 const defaultPermissions = {};
```
tdeekens: Why are they not null then?

PR #975, opened 2019-08-14
refactor: remove reselect

tdeekens:
#### Summary

This pull request removes `reselect` from the app-kit where possible. All `defaultMemoize` calls are `memoize` calls now from `memoize-one`. 

There is only one occurance of `reselect` where a selector is used: in the notifations package. I think removing it would be a separate step which we can also dicsuss if we want to after all.


packages/react-notifications/package.json
```diff
@@ -50,7 +50,7 @@
     "moment": "^2.24.0",
     "moment-timezone": "^0.5.26",
     "prop-types": "15.7.2",
-    "reselect": "^4.0.0"
+    "reselect": "4.0.0"
```
tdeekens: Just to pin it.

packages/permissions/package.json
```diff
@@ -43,7 +43,6 @@
     "lodash": "4.17.15",
     "lodash-es": "4.17.15",
     "prop-types": "15.7.2",
-    "reselect": "^4.0.0",
```
tdeekens: The package actually doesn't use it.

package.json
```diff
@@ -144,7 +144,8 @@
     "typescript": "3.5.3"
   },
   "resolutions": {
-    "@types/react": "16.9.1"
+    "@types/react": "16.9.1",
+    "react": "16.8.6"
```
tdeekens: Otherwise react-intl among others due to loose ranges keep pulling in v16.9 which leaves us with two versions and breaks out tests as we throw when we log.

tdeekens:
@emmenko I presume you joint the steps in another PR but didn't update the branch restrictions for `master`. I updated them now.

PR #976, opened 2019-08-14
fix(actions-global): distinguish unknown/api errors based on presence of body field

pa3:
Customers keep complaining about obscure error messages show in MC (in `application-products` in particular), whenever CRUD operations are failing because of not passing server-side validation.

What happening in this case is that `handleActionError` tries to distinguish between `sdk errors` and `unknown errors` based on a wrong assumption that `sdk errors` are never subclasses of `Error` class. After that it strips away all valuable information about the actual failure from passed `sdk error` object. In reality though, `sdk errors` are always instances of `Error` class.

With this PR we are changing the way described check is made so that it will be based on presence of absence of `body` field of supplied error object.

### DIFF
--- a/packages/actions-global/src/actions/handle-action-error.spec.ts
+++ b/packages/actions-global/src/actions/handle-action-error.spec.ts
@@ -1,0 +1,68 @@
+import { TStatusCode } from '@commercetools-frontend/constants';
+import handleActionError from './handle-action-error';
+
+let dispatch: jest.Mock;
+beforeEach(() => {
+  dispatch = jest.fn();
+  console.error = jest.fn();
+});
+
+describe('handleActionError', () => {
+  it('notifies user about unknown error', () => {
+    const error = new Error('oupsy, something went wrong');
+    handleActionError(error)(dispatch);
+    expect(dispatch).toHaveBeenCalledWith({
+      meta: {
+        dismissAfter: 0,
+      },
+      payload: {
+        domain: 'page',
+        id: 0,
+        kind: 'unexpected-error',
+        values: {},
+      },
+      type: 'ADD_NOTIFICATION',
+    });
+  });
+
+  it('notifies user about sdk error', () => {
+    class FakeSdkError extends Error {
+      statusCode: TStatusCode;
+      body: {
+        message: string;
+      };
+
+      constructor(statusCode: TStatusCode, message: string) {
+        super();
+        this.statusCode = statusCode;
+        this.body = {
+          message,
+        };
+      }
+    }
+
+    const error = new FakeSdkError(
+      401,
+      "Required attribute 'key-required' cannot be removed."
+    );
+    handleActionError(error)(dispatch);
+    expect(dispatch).toHaveBeenCalledWith({
+      meta: {
+        dismissAfter: 0,
+      },
+      payload: {
+        domain: 'page',
+        id: 0,
+        kind: 'api-error',
+        values: {
+          errors: [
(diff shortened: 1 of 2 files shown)


packages/actions-global/src/actions/handle-action-error.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+    );
+    handleActionError(error)(dispatch);
+    expect(dispatch).toHaveBeenCalledWith({
+      meta: {
+        dismissAfter: 0,
+      },
+      payload: {
+        domain: 'page',
+        id: 0,
+        kind: 'api-error',
```
tdeekens: Ah, these tests made it clear for me.

packages/actions-global/src/actions/handle-action-error.ts
```diff
@@ -9,15 +9,19 @@ import browserHistory from '@commercetools-frontend/browser-history';
...
+  statusCode: TStatusCode;
+  body: {
+    message: string;
+    errors?: TAppNotificationApiError | TAppNotificationApiError[];
+  };
+};
+
+type ActionError = Error | SdkError;
+
+function isSdkError(error: ActionError): error is SdkError {
```
tdeekens: 👍 for type guards.

packages/actions-global/src/actions/handle-action-error.ts
```diff
@@ -9,15 +9,19 @@ import browserHistory from '@commercetools-frontend/browser-history';
...
+  return (error as SdkError).body !== undefined;
```
tdeekens: Hm, didn't know that when you're "inside" a type guard we still need to cast the thing.

emmenko: To avoid the casting we should also be able to do:

```ts
return ‘body’ in error && error.body !== undefined;
```

packages/actions-global/src/actions/handle-action-error.ts
emmenko: Nice to use a type guard!

packages/actions-global/src/actions/handle-action-error.ts
```diff
@@ -9,15 +9,19 @@ import browserHistory from '@commercetools-frontend/browser-history';
...
-type ActionError =
-  | Error
-  | {
-      statusCode: TStatusCode;
-      body: {
-        message: string;
-        errors?: TAppNotificationApiError | TAppNotificationApiError[];
-      };
-    };
+type SdkError = {
```
emmenko: Nitpick: not 100% sure to name this `SdkError`, maybe it’s more like `ApiError`

pa3: Absolutely. Overlooked that we are using term ApiError in all other places.

packages/actions-global/src/actions/handle-action-error.ts
pa3: They use casting in TS documentation, so I thought it's a more idiomatic way: https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates

But I personally don't mind to use `in`-syntax as well. WDYT?

emmenko: Yeah I don’t mind either. In other examples I found they use `in` operator (also documented below the section in the doc link).

I find it without the casting a bit more clear, but it’s fine either way. I just wanted to mention it

PR #985, opened 2019-08-19
refactor(sdk): to convert to TS

emmenko:


@types/commercetools__api-request-builder/index.d.ts
```diff
@@ -0,0 +1,12 @@
+declare module '@commercetools/api-request-builder' {
```
tdeekens: Shouldn't this be part of the Node.JS SDK more? Have you talked to the team?

emmenko: Yeah, I haven't talked to them yet. Eventually I think we should migrate the nodejs repo to TS as well, since we also provide a TS builder SDK.

This is also not a new topic, there are already issues regarding TS declarations: https://github.com/commercetools/nodejs/issues/329

For now, this is the bare minimum for us to work with. I'll check with @daern91 what are the plans for moving to TS but I assume it's a low priority for them.

tdeekens: I know that @acbeni started a TypeScript SDK but I am not sure where it feeds into this. I just would avoid having too much "cross team/library" typings flying around.

emmenko: The TS builder SDK is just for building the request URL (like the api-builder package does). You still need the rest of the SDK packages, which would need some typings.

emmenko: Of course I'd be happy to have the types built into the SDK packages and avoid maintaining them manually.

emmenko: FYI: https://github.com/commercetools/nodejs/issues/329#issuecomment-522578246

tdeekens: Thanks!

emmenko:
@adnasa @amine-benselim do you feel like taking a look at this PR?

adnasa:
on it

amine-benselim:
I did look at that PR and the discussions you had with Tobi. I don’t feel comfortable accepting a PR if im not a 100% I get everything :sweat_smile: I know you’re adding types to the code but there seems to be some code changes as well.  I will look at it again later but don't wait for my approval :p

PR #986, opened 2019-08-19
feat(sdk): allow to pass a mcApiProxyTarget, to the SDK action

emmenko:
All requests (apart from graphql) to the MC API should be prefixed with `/proxy/<service>`.

This PR adds support for a `mcApiProxyTarget,` option in the SDK action creators to make it easier for consumers to configure the action. The available proxy target are also exposed as constants `MC_API_PROXY_TARGETS`.

PS: one of the reasons for having this is to help migrate our current action creators to use the `/proxy/ctp` endpoint instead of the legacy one (which is without the prefix).
Eventually we can remove the legacy route from the MC API.

### DIFF
--- a/packages/sdk/README.md
+++ b/packages/sdk/README.md
@@ -51,27 +51,29 @@ import { actions as sdkActions } from '@commercetools-frontend/sdk';
 
 #### Methods
 
-There are three action creators which all have the same behavior and API. The only difference is the resulting HTTP method:
+The supported action creators are:
 
-- `sdkActions.get(description)`: `GET`
-- `sdkActions.post(description)`: `POST`
-- `sdkActions.del(description)`: `DELETE`
+- `sdkActions.get(config)`: sends a HTTP `GET`
+- `sdkActions.post(config)`: sends a HTTP `POST`
+- `sdkActions.del(config)`: sends a HTTP `DELETE`
+- `sdkActions.head(config)`: sends a HTTP `HEAD`
 
 #### Specifying an endpoint
 
 There are two ways to describe an endpoint:
 
-- direct: `description.uri` is used as-is
-- combined: `description.service` and `description.options` are combined to form a `uri`
+- by `uri`: pass the URI as string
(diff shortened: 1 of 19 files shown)


packages/sdk/README.md
```diff
@@ -87,7 +89,8 @@ When both, `uri` and `options` (or `service`) are present, the `uri` takes prece
 {
   service: String,
   options: Object,
-  payload: Object | String
+  uriPrefix?: String,
```
tdeekens: Given we want to remove it should we not document it?

emmenko: 🤔

What we want to remove is the legacy MC API endpoint...

tdeekens: Ah, got yah. I thought you, given we migrated all routes, also want to remove the option.

packages/sdk/src/middleware/middleware.js
```diff
@@ -40,7 +40,13 @@ export default function createSdkMiddleware({
     if (!action) return next(action);
 
     if (action.type === 'SDK') {
-      const uri = actionToUri(action, projectKey);
+      const uri = [
+        action.payload.uriPrefix &&
+          action.payload.uriPrefix.replace(/\/?$/, ''),
```
tdeekens: Can we add a comment why we do this?

emmenko: The `replace` part?

emmenko:
Good idea, I refactored it in the last commit. Please have another look.

emmenko:
Yeah but `target` is too generic. While `target` works for graphql, because it's tight to our MC API, the `sdk` package can be used potentially with any API. So I would prefer to keep it more specific, as it's clearer what the purpose is. Hope it makes sense.

tdeekens:
Yeah, it does. Maybe even `mcProxyTarget` makes it more specific.

emmenko:
Renamed to `mcApiProxyTarget` 😉

PR #987, opened 2019-08-20
refactor(vrt): move visual specs into VRT app

emmenko:
> This PR is still WIP and it's meant to explore other options for running the VRTs. This has been on my TODO list for quite some time and I just wanted to kick it off now. I opened a PR so that I can write down my thoughts about the intentions behind this.
> Don't bother reviewing this yet, unless you are interested and would like to give some early feedback.

---

**Background**

Our current VRT setup works pretty well but it still doesn't feel 100% right:
- we have a quite complex webpack config
- we use aliases which makes it more difficult to get proper import definitions and proper typechecking
- we load the visual specs using `require.context`
- it's a bit hard to get a good overview of all the visual specs, as they are spread across different packages

**New approach**

What I would like to have instead is:
- a very simple setup, the tooling shouldn't get in the way
- a good overview of what we have under test
- proper import definitions and typechecking, to help finding additional problems

With that in mind, I decided to try using `parcel`, as it basically does not require any explicit setup. Furthermore, migrating the visual specs to TS has some extra advantages:
- we have extra checks about using the components
- easier to write the specs, with intellisense etc.

For the reasons above, I decided to move the specs into the VRT app package, without relying on `require.context`. I think it's fine even though we don't have the specs co-located to the components anymore.

---

Update: let's keep Parcel our for now, at least until they release v2. We can use our own mc-scripts in the meantime to start/build the visual testing app.


montezume:
Just a quick comment, but I'm not sure I agree that `it's a bit hard to get a good overview of all the visual specs, as they are spread across different packages` is a good thing? 

With that logic, why don't we put all of our spec files in a /test folder? Putting the .visualroute and .visualspec files alongside the files was a bug, but a feature 🤔

emmenko:
So this is now ready for review from my side. Let me know what you think

emmenko:
Thanks for the feedback. Let me try to answer your points. 🤗

Using Parcel is not a must, we can still keep using Webpack. However, I feel that trying out another bundler could help us to evaluate alternatives and making sure that our packages work not only with Webpack.
One of the ideas I had was also to make a starter template using Parcel, it does not necessarily need to be in the VRT app but it's a start.

Besides that, one of the things that I found weird and annoying was to have huge relative imports and using import aliases, as opposed to importing from the real production bundle👇

image

Because of that, the Webpack config got unnecessarily complicated, which is also why I wanted to look into "zero config" tools. I feel like now (if you look at the screenshot above) the imports are easier to understand and to follow.

Again, my intent with this PR is to try making the VRT setup simpler.

> Introducing type-checking albeit usefull is side-passing the UIKit team's decicion not yet to put effort/time into using/learning them while focussing work elsewhere

I'm a bit confused what this has to do with the uikit and the visual specs 🤔 It's nothing new, we already have some type declaration because we need them in some of the appkit packages.

emmenko:
Alright I removed Parcel now in favor of our own mc-scripts: https://github.com/commercetools/merchant-center-application-kit/pull/987/commits/08bcf7c941611b1298cf562bb4a87d0b2c7214a7

/cc @tdeekens

emmenko:
@jonnybel anything to add from your side here?

PR #990, opened 2019-08-21
refactor(permissions/has-permissions): pass `actualDataFenceValues` into `selectDataFenceData`

adnasa:
#### Summary

- passes `actualDataFence.values` into `selectDataFenceData` (previously named `selectDataFenceDataByType`

#### Description

Previously we wrote `selectDataFenceDataByType` with the assumption that the data returned from `selectDataFenceDataByType` will come only from the resource we demand a dataFence for e.g `Orders`.  This is true in the case for Details and List Views.  

And it will look as the following:

```
{ 
  dataFences: [
    // N number of datafences
  ],
  selectDataFenceData: ownProps => ({ type }) => {
    switch(type) {
      ...
    }
  }
}
```

---

In the Navbar, we also need to apply datafences.
This is especially necessary because in the MC-API, [we are distinguishing](https://github.com/commercetools/merchant-center-services/pulls?q=is%3Apr+is%3Aclosed+author%3Aamine-benselim) project permissions (via `allAppliedPermissions`) and dataFence permissions (`allAppliedDataFences`).

As a result for this, when the following permissions are applied on the user

```
GIVEN no project permissions
AND user has ManageOrders on dataFence permission
```

Then `appliedPermissions` will imply false on `ManageOrders` which leads to the menu not displaying.

#### Approach to solve this

- we demanded dataFence on `NavBar`
- we pass in demandedDataFence (coming from elsewhere)
- we pass in `selectDataFenceData`
  - at this point, we will receive `actualDataFenceValues`. Because the rule is if there are dataFences, we display the application menu, so we return `actualDataFenceValues`

With the approach discussed above, we make sure the `authorized` works for both NavBar and Details/List view. The difference is the on NavBar, it will always be true.

If there are no dataFence values, we "fall back" to the project permission (general permission)

--- 

This is the outcome of yesterday's meeting. A PR with changes to `NavBar` will follow

### DIFF
--- a/packages/permissions/README.md
+++ b/packages/permissions/README.md
@@ -114,7 +114,7 @@ useIsAutorized(options: HookOptions): boolean
   be allowed to render
 - `demandedAtionRights`: (_optional_) an array of action rights (mentioned above)
 - `demandedDataFences`: (_optional_) an array of `DataFence` (mentioned above)
-- `selectDataFenceDataByType`: (_optional_ unless `dataFences` is specified) the function is called on each specified `dataFence` and should return the mapped value specific to the data fence. The function usually contains a switch-case logic based on the data fence `type`, with the return value based on the `props`. For example, with `type: 'order'`, the return value would be something like `ownProps.order.storeRef.key`.
+- `selectDataFenceData`: (_optional_ unless `dataFences` is specified) the function is called on each specified `dataFence` and should return the mapped value specific to the data fence. The function usually contains a switch-case logic based on the data fence `type`, with the return value based on the `props`. For example, with `type: 'order'`, the return value would be something like `ownProps.order.storeRef.key`.
 - `shouldMatchSomePermissions`: (_optional_) determines if _some_ or _every_ requested permission should match (default `false`)
 
 ### Example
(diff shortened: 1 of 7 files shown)


packages/permissions/README.md
```diff
@@ -193,7 +193,7 @@ match, otherwise a fallback component.
 - `permissions`: an array of `Permission`, requested by the component
 - `actionRights`: (_optional_) an array of `ActionRight`, requested by the component
 - `dataFences`: (_optional_) an array of `DataFence`, requested by the component
-- `selectDataFenceDataByType`: (_optional_ unless `dataFences` is specified) the function is called on each specified `dataFence` and should return the mapped value specific to the data fence. The function usually contains a switch-case logic based on the data fence `type`, with the return value based on the `props`. For example, with `type: 'order'`, the return value would be something like `ownProps.order.storeRef.key`.
+- `selectDataFenceData`: (_optional_ unless `dataFences` is specified) the function is called on each specified `dataFence` and should return the mapped value specific to the data fence. The function usually contains a switch-case logic based on the data fence `type`, with the return value based on the `props`. For example, with `type: 'order'`, the return value would be something like `ownProps.order.storeRef.key`.
```
amine-benselim: should we change the documentation about what we pass to this function ?

adnasa: I think this doc is clear enough. And I personally think that `actualDataFenceValues` is up for discovery, since we rarely want to encourage devs to use this.

emmenko:
PS: ignore the percy snapshots for the website

PR #991, opened 2019-08-21
feat(application-shell/navbar): add demandedDataFences and return dataFenceValues given to the user

adnasa:
#### Summary

- adds `dataFences`(demandedDataFence) into RestrictedByPermission for NavBar
- follow up on https://github.com/commercetools/merchant-center-application-kit/pull/990

#### Description

we know that we demand for DataFences to apply on `Authorized`, we are required to provide a `selectDataFenceData`.

For Details/List, we could use the resource to provide values e.g `order.storeRef.key`.
However in Navbar, we do not have any resource. So to "circumvent" this limitation, we let the `Authorized` give us the values we are looking for and return that as the default.

This is to enable the case where user has dataFence permissions but not project permissions

### DIFF
--- a/packages/application-shell/src/components/navbar/navbar.js
+++ b/packages/application-shell/src/components/navbar/navbar.js
@@ -265,10 +265,19 @@ export const RestrictedMenuItem = props => {
     return null;
 
   const permissionsWrapper =
-    props.permissions.length > 0 ? (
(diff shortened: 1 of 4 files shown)


adnasa:
thank you so much for the help! ❤️ 
This is exactly what I was hoping to finish before I went on vacation in the first place

PR #992, opened 2019-08-21
fix: tsconfig to correctly include custom type definitions and check test files

emmenko:
The `include` option in the `tsconfig` didn't do what we expected and so we didn't catch some errors in the tests.

After some research and try-error I found that using `typeRoots` solves the issue. You can see then all the places in tests that we had to fix.


packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
```diff
@@ -62,9 +62,9 @@ const testRender = ({
   demandedPermissions: TPermissionName[];
   demandedActionRights?: TDemandedActionRight[];
   shouldMatchSomePermissions: boolean;
-  actualPermissions?: TPermissions;
-  actualActionRights?: TActionRights;
-  actualDataFences?: TDataFences;
+  actualPermissions: TPermissions | null;
+  actualActionRights: TActionRights | null;
+  actualDataFences: TDataFences | null;
```
tdeekens: Maybe one time we build a Nullable type for this case 🤞

emmenko: https://github.com/Microsoft/TypeScript/issues/19944#issuecomment-344026764

🤷‍♂️

emmenko:
Aaaaah, this is driving me crazy!! 😵

emmenko:
I think I might have got this right this time 🤞

packages/constants/src/constants.ts
```diff
@@ -131,3 +131,17 @@ export const MC_API_PROXY_TARGETS = {
   MC_METRICS: 'mc-metrics',
 } as const;
 export type TMcApiProxyTargets = (typeof MC_API_PROXY_TARGETS)[keyof typeof MC_API_PROXY_TARGETS];
+
+// Global application environment on window object
+export interface ApplicationWindow extends Window {
```
tdeekens: Oh neato.

emmenko: Yeah I think it's better to ship this with the packages. I chose to put it in the constants package because it's usually used by all other packages.

emmenko:
There seem to be some issues with CircleCI

PR #1008, opened 2019-08-27
Rework Form Modal Pages

jonnybel:
#### Summary

Following the discussion from [here](https://github.com/commercetools/merchant-center-frontend/pull/7424#discussion_r314775129), this PR addresses those suggestions.

This is still WIP, but let's use this PR to discuss if we're happy with this proposed new approach. 

#### Description

Because this is WIP, please focus on the `FormModalPage` and `CustomFormModalPage`, and the `default-form-buttons.tsx` (which I'm still not sure where to place), and let me know if you would prefer an alternative to any of the following key points:
* New component `CustomFormModalPage`, which doesn't expect any default controls but exposes some pre-made control buttons: Confirm, Cancel, and Delete, to ease its use for the most common cases which still require dynamic controls.
* The `controls` prop is an array of React elements.
* `FormModalPage`, which uses `CustomFormModalPage`under the hood, is a simpler version of the latter with all the defaults set (and the possibility to still hide the controls in case of need).
* New prop `hideControls`, which simply acts as a switch that does what it says regardless of anything else (no conditional props).
* I haven't changed much of `TabularModalPage` yet, but I'm thinking of keeping it flexible, with no default controls, but we can discuss if we also want a simpler version with defaults.


packages/application-components/src/components/modal-pages/form-modal-page/form-modal-page.tsx
```diff
@@ -1,24 +1,14 @@
...
 import buttonMessages from '../../../utils/button-messages';
-import ModalPage from '../internals/modal-page';
-import ModalPageHeader from '../internals/modal-page-header';
-import ModalPageHeaderDefaultControls from '../internals/modal-page-header-default-controls';
-import { ContentWrapper } from '../internals/modal-page.styles';
+import CustomFormModalPage from '../custom-form-modal-page';
+import {
+  ConfirmButton,
+  CancelButton,
+} from '../../../utils/default-form-buttons';
```
emmenko: We should use `CustomFormModalPage.ConfirmButton` and `CustomFormModalPage.CancelButton` right?

packages/application-components/src/utils/default-form-buttons.tsx
```diff
@@ -0,0 +1,96 @@
...
+  'label' | 'isDisabled' | 'dataAttributes'
+> = {
+  label: buttonMessages.confirm,
+  ...commonDefaultProps,
+};
+
+const getFormattedLabel = (label: Label, intl: IntlShape) =>
+  typeof label === 'string' ? label : intl.formatMessage(label);
+
+const ConfirmButton = (props: Props) => {
```
emmenko: I would name this `PrimaryButton`

packages/application-components/src/components/modal-pages/internals/modal-page-header-default-controls.tsx
```diff
@@ -1,16 +1,12 @@
...
-} from '@commercetools-frontend/ui-kit';
+import { MessageDescriptor } from 'react-intl';
 import filterDataAttributes from '../../../utils/filter-data-attributes';
+import { SaveButton, CancelButton } from '../../../utils/form-buttons';
 
 type Label = string | MessageDescriptor;
 type Props = {
-  labelSecondaryButton: Label;
-  labelPrimaryButton: Label;
+  labelSecondaryButton?: Label;
```
emmenko: Shouldn't those props be required? 🤔

packages/application-components/src/utils/default-form-buttons.tsx
tdeekens: `FormPrimaryButton` maybe so it's

1. Not confused with UIKit's `PrimaryButton`
2. Mapps nicely to the `isPrimaryButtonDisbabled`-prop

Hope that makes sense.

packages/application-components/src/components/modal-pages/custom-form-modal-page/custom-form-modal-page.tsx
```diff
@@ -0,0 +1,64 @@
...
+  getParentSelector?: () => HTMLElement;
+  shouldDelayOnClose?: boolean;
+  // TopBar Props
+  topBarCurrentPathLabel?: string;
+  topBarPreviousPathLabel?: Label;
+  // Header Props
+  subtitle?: string | React.ReactElement;
+  // Controls Props
+  controls: React.ReactNode[];
+  hideControls?: boolean;
```
tdeekens: This should default to `false`, right?

packages/application-components/src/components/modal-pages/custom-form-modal-page/custom-form-modal-page.tsx
```diff
@@ -0,0 +1,64 @@
...
+    baseZIndex={props.baseZIndex}
+    currentPathLabel={props.topBarCurrentPathLabel || props.title}
+    previousPathLabel={props.topBarPreviousPathLabel}
+    getParentSelector={props.getParentSelector}
+    shouldDelayOnClose={props.shouldDelayOnClose}
+  >
+    <ModalPageHeader title={props.title} subtitle={props.subtitle}>
+      {!props.hideControls && props.controls.length !== 0 && (
+        <Spacings.Inline alignItems="flex-end">
+          {props.controls.map(controlItem => controlItem)}
```
tdeekens: This should (to my understanding) not be necessary. It can just be `{props.controls}` unless we want to wrap it or anything.

packages/application-components/src/components/modal-pages/custom-form-modal-page/custom-form-modal-page.tsx
```diff
@@ -0,0 +1,64 @@
...
+  baseZIndex?: number;
+  getParentSelector?: () => HTMLElement;
+  shouldDelayOnClose?: boolean;
+  // TopBar Props
+  topBarCurrentPathLabel?: string;
+  topBarPreviousPathLabel?: Label;
+  // Header Props
+  subtitle?: string | React.ReactElement;
+  // Controls Props
+  controls: React.ReactNode[];
```
tdeekens: Why is this an array? Can't it just be one Node or would that mean the consumer has to often wrap in a `Fragment`?

packages/application-components/src/utils/default-form-buttons.tsx
```diff
@@ -0,0 +1,96 @@
...
+  dataAttributes: {},
+};
+
+const confirmDefaultProps: Pick<
+  Props,
+  'label' | 'isDisabled' | 'dataAttributes'
+> = {
+  label: buttonMessages.confirm,
+  ...commonDefaultProps,
+};
```
tdeekens: This is clever but also maybe harder to follow than it has to given we have this shared `Props` and `commonDefaultProps` thing. We have `Props`, pick only some, use common default props and spread in some more. Would it make sense to just have duplication here instead?

packages/application-components/src/utils/default-form-buttons.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+const cancelDefaultProps: Pick<
+  Props,
+  'label' | 'isDisabled' | 'dataAttributes'
+> = {
+  label: buttonMessages.cancel,
+  ...commonDefaultProps,
+};
+
+const CancelButton = (props: Props) => {
```
tdeekens: `FormSecondaryButton` normally but given we have two variants I suppose `FormCancelButton`.

packages/application-components/src/components/modal-pages/internals/modal-page-header-default-controls.tsx
jonnybel: Oh, actually I "abandoned" this `ModalPageHeaderDefaultControls` component because it would just become an array of (or a fragment wrapping) the buttons in [default-form-buttons.tsx](https://github.com/commercetools/merchant-center-application-kit/blob/9ddaf1a1abe817575a44d30d085109422fce520f/packages/application-components/src/utils/default-form-buttons.tsx). And the buttons there already have default labels.

packages/application-components/src/utils/default-form-buttons.tsx
jonnybel: Yeah, and it doesn't even save that many lines. I'll just duplicate them.

packages/application-components/src/utils/default-form-buttons.tsx
jonnybel: I'll still call it `FormSecondaryButton`, since the other one is technically not a SecondaryButton but an IconButton with a trash icon, specifically for deletion, which I'll call `FormDeleteButton`.

jonnybel:
I've updated the PR. Sorry for the delay (I was on vacations last week 😄)

tdeekens:
Thanks. For the future: small commits of the changes resulting from the discussions would made re-review easier especially with such a time gap. Those commits could then still be rebased out or squashed away.

packages/application-components/src/components/modal-pages/custom-form-modal-page/custom-form-modal-page.mdx
```diff
@@ -0,0 +1,81 @@
...
+
+# CustomFormModalPage
+
+import CustomFormModalPageExample from './custom-form-modal-page.example';
+
+<CustomFormModalPageExample />
+
+## Description
+
+Custom Form Modal pages are controlled components used to render a page with a form or something that requires user input, similar to `InfoModalPage` but semantically with a different role. The header includes an area for controil buttons, which can be customized at will. The component exposes three default buttons which can be useful for most use cases.
```
emmenko:
```suggestion
Custom Form Modal pages are controlled components used to render a page with a form or something that requires user input, similar to `InfoModalPage` but semantically with a different role. The header includes an area for control buttons, which can be customized at will. The component exposes three default buttons which can be useful for most use cases.
```

packages/application-components/src/components/modal-pages/custom-form-modal-page/custom-form-modal-page.tsx
```diff
@@ -0,0 +1,64 @@
...
+  baseZIndex?: number;
+  getParentSelector?: () => HTMLElement;
+  shouldDelayOnClose?: boolean;
+  // TopBar Props
+  topBarCurrentPathLabel?: string;
+  topBarPreviousPathLabel?: Label;
+  // Header Props
+  subtitle?: string | React.ReactElement;
+  // Controls Props
+  controls: React.ReactNode;
```
emmenko: Nitpick: should we name this `formControls` maybe? We do that in the tabs form...

packages/application-components/src/components/modal-pages/custom-form-modal-page/custom-form-modal-page.tsx
```diff
@@ -0,0 +1,64 @@
...
+    onClose={props.onClose}
+    baseZIndex={props.baseZIndex}
+    currentPathLabel={props.topBarCurrentPathLabel || props.title}
+    previousPathLabel={props.topBarPreviousPathLabel}
+    getParentSelector={props.getParentSelector}
+    shouldDelayOnClose={props.shouldDelayOnClose}
+  >
+    <ModalPageHeader title={props.title} subtitle={props.subtitle}>
+      {!props.hideControls && props.controls && (
+        <Spacings.Inline alignItems="flex-end">
```
emmenko: I'm not sure if we should leave this to the consumer 🤔

packages/application-components/src/components/modal-pages/form-modal-page/form-modal-page.mdx
```diff
@@ -52,14 +52,13 @@ import { FormModalPage } from '@commercetools-frontend/application-components';
...
-| `customControls`                | `node`                     |    -     | -                         | Pass a React.node to be used in place of the pre-determined controls. This can be useful if you need actions other than Cancel & Confirm, or other types of buttons, while keeping the same modal header layout                                                                                                                                                                                                                                                                                                     |
```
emmenko: I guess this is considered a breaking change? 🤔

jonnybel: Yep. And it's not the only one, since TabularModalPage also won't have defaults anymore. But we can discuss if we want to avoid these.

packages/application-components/src/components/modal-pages/custom-form-modal-page/custom-form-modal-page.tsx
jonnybel: What do you mean? Should we move the `controls` and `hideControls` to be part of the ModalPageHeader?

packages/application-components/src/components/modal-pages/custom-form-modal-page/custom-form-modal-page.visualroute.js
```diff
@@ -0,0 +1,98 @@
...
+  SearchIcon,
+  BinLinearIcon,
+  TextField,
+} from '@commercetools-frontend/ui-kit';
+import { CustomFormModalPage } from '@local-build/application-components';
+import { Suite, Spec } from '../../../../../../visual-testing-app/test-utils';
+
+export const routePath = '/custom-form-modal-page';
+
+const ModalPageWithPortalParentSelector = ({ portalId, ...props }) => (
```
tdeekens:
```suggestion
const ModalPageWithPortalParentSelector = ({ portalId, ...remainingProps }) => (
```

...otherwise it's easily confused with the full props further down in code.

packages/application-components/src/utils/default-form-buttons.tsx
```diff
@@ -0,0 +1,94 @@
...
+const primaryDefaultProps: Pick<
+  Props,
+  'label' | 'isDisabled' | 'dataAttributes'
+> = {
+  label: buttonMessages.confirm,
+  isDisabled: false,
+  dataAttributes: {},
+};
+
+const getFormattedLabel = (label: Label, intl: IntlShape) =>
```
tdeekens: Just a thought I had looking at this. Cause it could be:

```js
const useFormattedLabel = (label: Label) => {
   const intl = useIntl();

   return typeof label === 'string' ? label : intl.formatMessage(label);
}
```

then in code

```js
const formattedSecondaryButtonLabel = useFormattedLabel(props.label)
```

packages/application-components/src/components/modal-pages/custom-form-modal-page/custom-form-modal-page.tsx
emmenko: I mean the Spacing wrapper. I think it can be up to the consumer to define that, I fear that we might run into some unwanted constraints by having the spacing component on our side.

jonnybel: Well, in all the migrations I did in the MC, they always respect this layout.
If we need to add more flexibility in the future, might as well just add a prop to render a customHeader or a customElement in the header.
This `formControls` prop already allows rendering anything, and the only "controls specific" thing it's doing right now is placing the rendered node inside that `Spacings`.

So.. I'd prefer to keep this here 😄

packages/application-components/src/utils/default-form-buttons.tsx
```diff
@@ -27,20 +27,21 @@ const primaryDefaultProps: Pick<
...
-      {...filterDataAttributes(props.dataAttributes)}
-    />
-  );
+
+  return typeof label === 'string' ? label : intl.formatMessage(label);
 };
+
+const FormPrimaryButton = (props: Props) => (
+  <PrimaryButton
+    label={useFormattedLabel(props.label)}
```
tdeekens: Careful. You might want to move this to L37 and store it in a var cause if we ever conditionally render the order of hooks can change easily.

jonnybel:
Considering this has breaking changes to `application-components`, should we merge this to master or have a separate branch for `v16`?

emmenko:
Is there a way that we can make it backwards compatible? 🤔

jonnybel:
Since one of the main points it to get rid of `customControls` in FormModalPage and the conditional types that it implies, I'm not seeing how 😕

But because this also doesn't really bring any groundbreaking new stuff to the table, it might be worth to keep this in a separate branch and then merge to master when we have other necessary breaking changes to go together. Just as long as we keep this in mind if we want to update stuff in the ModalPages.

tdeekens:
@emmenko and @jonnybel what is missing here? @emmenko anything from your side and a review perspective?

emmenko:
The changes are fine. I was just asking about how should deal with the breaking changes. I'm fine with whatever you guys will decide.

tdeekens:
@jonnybel I guess we can't map the `customControls` back to the new API to avoid the breaking change (if we would want to)?

jonnybel:
@tdeekens We can map them back, but that implies we also restore the conditionally required props (which we wanted to get rid of), and IMO it defeats the purpose of having a CustomFormModalPage component.

tdeekens:
Okay, thanks for explaining. Then I guess we just ship it as a breaking change. The question then is when. Either now or with a "scheduled" release which could include other breaking changes.

jonnybel:
Yes. I think we can do it _scheduled_ because this isn't something critical, as I wrote above.

emmenko:
@jonnybel should we keep this PR/branch up-to-date with `master`? Or should we close it and open a new one?

jonnybel:
It depends, if we're not releasing a new major version any time soon, it might be better to close this for now. Then I can check if it's easier to re-open and resolve the conflicts, or open a new one.

jonnybel:
Closing this PR.

Too many conflicts, it was better to re-do the changes on a fresh branch from master (opening the PR right away)

PR #1009, opened 2019-08-28
CON-187: properly pass dataFences from menu to `Authorized.js`

amine-benselim:
the prop drilling we were doing was incomplete 😅 I needed to pass dataFences one more level down. 

TODO:
- [x] change logic of appliedDataFences from AND to OR

### DIFF
--- a/packages/permissions/src/utils/has-permissions.ts
+++ b/packages/permissions/src/utils/has-permissions.ts
@@ -262,7 +262,7 @@ export const hasAppliedDataFence = (options: TOptionsForAppliedDataFence) =>
         name: dataFenceName,
         // we do the type casting because at this point we are sure that
         // specificActualDataFence.dataFenceValue is not null
-        dataFenceValue: dataFenceValue as { values: string[] },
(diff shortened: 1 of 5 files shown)


packages/permissions/src/utils/has-permissions.ts
```diff
@@ -242,20 +244,27 @@ export const hasAppliedDataFence = (options: TOptionsForAppliedDataFence) =>
...
+
+    if (!specificActualDataFence) return false;
+
+    const [dataFenceName, dataFenceValue] = specificActualDataFence;
+    return hasDemandedDataFence({
+      actualDataFence: {
+        name: dataFenceName,
+        // we do the type casting because at this point we are sure that
+        // specificActualDataFence.dataFenceValue is not null
+        dataFenceValue: dataFenceValue as { values: string[] },
```
emmenko: Yeah I think it's fine like this for now. You can simplify this by doing it like:

```ts
dataFenceValue as TActualDataFence['dataFenceValue']
```

PR #1013, opened 2019-08-29
chore: update flopflip

tdeekens:
#### Summary

Includes flopflip only flushing changes to flag values.


yarn.lock
```diff
@@ -7163,6 +7163,15 @@ create-jest-runner@^0.5.3:
     jest-worker "^24.0.0"
     throat "^4.1.0"
 
+create-react-class@^15.6.0:
```
emmenko: Hmm why do we get this dependency again? 🤔

tdeekens: I have no idea. Flopflip doesn't use it. I just ran changed the flopflip versions. I would not manually fiddle with the lockfile. Was this causing issues?

emmenko: Just wondering...maybe try regenerating the lockfile.

tdeekens: 566b306

tdeekens: CleanShot 2019-08-30 at 09 49 07@2x

tdeekens:
@emmenko can we merge this?

PR #1014, opened 2019-08-30
feat: add eslint plugin for react testing library 🔥

emmenko:
To help us following best practices when writing RTL tests, I thought it would be nice to have some eslint rules.

The first rule I created is about using `queryBy*` instead of `getBy*` when using `expect()`.


packages/eslint-plugin-testing-library-react/src/rules/prefer-expect-query-by.ts
```diff
@@ -0,0 +1,114 @@
...
+
+function mapNodesForWrongGetByQuery(node: TSESTree.CallExpression) {
+  const nodeArguments = node.arguments as ExtendedExpression[];
+  return nodeArguments
+    .map(arg => {
+      if (!arg.callee) {
+        return null;
+      }
+      // Example: `expect(rendered.getBy*)`
+      if (isMemberExpression(arg.callee)) {
```
emmenko: This case is represented like this as an AST:

image

packages/eslint-plugin-testing-library-react/src/rules/prefer-expect-query-by.ts
```diff
@@ -0,0 +1,114 @@
...
+        const node = arg.callee.property as ExtendedExpression['callee'];
+        if (isIdentifier(node) && isUsingWrongQueries(node)) {
+          return node;
+        }
+        return null;
+      }
+
+      // Example: `expect(getBy*)`
+      if (isIdentifier(arg.callee) && isUsingWrongQueries(arg.callee)) {
```
emmenko: This case is represented like this instead:

image

packages/eslint-plugin-testing-library-react/src/rules/prefer-expect-query-by.ts
```diff
@@ -0,0 +1,114 @@
...
+    messages: {
+      expectQueryBy:
+        'Using `expect(getBy*)` is not recommended, use `expect(queryBy*)` instead.',
+    },
+    schema: [],
+    type: 'suggestion',
+    fixable: 'code',
+  },
+  defaultOptions: [],
+  create: context => ({
```
amine-benselim: 👀  so that's how you "fix" mistakes 🔥

packages/eslint-plugin-testing-library-react/src/rules/prefer-expect-query-by.ts
amine-benselim: I could not spot the difference 😅

emmenko: The `arguments` list, one is a `CallExpression` object with a `callee` of type `MemberExpression`, the other one of type `Identifier`.

emmenko:
That's what I'm talking about!

image

packages/eslint-plugin-testing-library-react/src/rules/prefer-expect-query-by.ts
emmenko: Yeah, it was "surprisingly simple" 😅

packages/eslint-plugin-testing-library-react/src/rules/prefer-expect-query-by.ts
amine-benselim: ohhh. now it makes sense. can you make the distinction in the "fixing" part ? in order not to change the first case which would lead to an error.

emmenko: What do you mean?

packages/eslint-plugin-testing-library-react/README.md
```diff
@@ -0,0 +1,49 @@
+# eslint-plugin-testing-library-react
+
+<p align="center">
+  <a href="https://www.npmjs.com/package/eslint-plugin-testing-library-react"><img src="https://badgen.net/npm/v/eslint-plugin-testing-library-react" alt="Latest release (latest dist-tag)" /></a> <a href="https://www.npmjs.com/package/eslint-plugin-testing-library-react"><img src="https://badgen.net/npm/v/eslint-plugin-testing-library-react/next" alt="Latest release (next dist-tag)" /></a> <a href="https://bundlephobia.com/result?p=eslint-plugin-testing-library-react"><img src="https://badgen.net/bundlephobia/minzip/eslint-plugin-testing-library-react" alt="Minified + GZipped size" /></a> <a href="https://github.com/commercetools/merchant-center-application-kit/blob/master/LICENSE"><img src="https://badgen.net/github/license/commercetools/merchant-center-application-kit" alt="GitHub license" /></a>
+</p>
+
+ESLint plugin with React Testing Library rules for enforcing best practices.
```
tdeekens:
```suggestion
ESLint plugin for React Testing Library rules guiding best practices.
```

packages/eslint-plugin-testing-library-react/src/rules/prefer-expect-query-by.ts
amine-benselim: I mean if it is `rendered.getBy..` (so a `MemberExpression `) then fix it, but if it is `getBy` (`CallExpression`) don't fix it. IDK if this works :p

packages/eslint-plugin-testing-library-react/docs/rules/prefer-expect-query-by.md
```diff
@@ -0,0 +1,28 @@
+# Avoid using `expect(getBy*)` (prefer-expect-query-by)
+
+React (DOM) Testing Library support two types of queries: `getBy*` and `queryBy*`. Using `getBy*` throws an error in case the element is not found. This is useful when using method like `waitForElement`, which are `async` functions and therefore throwing an error will cause the test to stop.
+However, when trying to assert if an element is not in the document, we obviously can't use `getBy*`. Instead we should use `queryBy*`, which does not throw and therefore we can assert that `expect(queryByText("Foo")).not.toBeInTheDocument()`.
```
tdeekens: I would avoid words like "obviously" it can easily be perceived as "pointy" especially as this is Open Source.

"Instead it is recommended to use `getQueryBy*` to receive better error messages." or something.

packages/eslint-plugin-testing-library-react/docs/rules/prefer-expect-query-by.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Rule details
+
+This rule triggers an error if `expect(getBy*)` is used.
```
tdeekens: I think error and warning can be configured.

```suggestion
This rule gives a notification whenever `expect(getBy*)` is used.
```

packages/eslint-plugin-testing-library-react/docs/rules/prefer-expect-query-by.md
```diff
@@ -0,0 +1,28 @@
...
+
+This rule is enabled by default.
+
+### Default configuration
+
+The following patterns is considered error:
```
tdeekens:
```suggestion
The following patterns is considered erroneous:
```

packages/eslint-plugin-testing-library-react/docs/rules/prefer-expect-query-by.md
```diff
@@ -0,0 +1,28 @@
...
+
+```js
+test('some test', () => {
+  expect(getByText('Foo')).not.toBeInTheDocument();
+});
+```
+
+The following pattern is not considered error:
```
tdeekens:
```suggestion
The following pattern is considered non erroneous:
```

packages/eslint-plugin-testing-library-react/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "eslint-plugin-testing-library-react",
+  "version": "14.0.0",
+  "description": "ESLint plugin with React Testing Library rules for enforcing best practices",
```
tdeekens: ^^

packages/eslint-plugin-testing-library-react/src/configs/recommended.json
```diff
@@ -0,0 +1,6 @@
+{
+  "plugins": ["testing-library-react"],
+  "rules": {
+    "testing-library-react/prefer-expect-query-by": "error"
```
tdeekens: Wondering if we should start with it as a warning?

packages/eslint-plugin-testing-library-react/src/rules/prefer-expect-query-by.ts
```diff
@@ -0,0 +1,105 @@
...
+  node: ExtendedExpression['callee']
+): node is TSESTree.Identifier {
+  return node.type === AST_NODE_TYPES.Identifier;
+}
+
+function isMemberExpression(
+  node: ExtendedExpression['callee']
+): node is TSESTree.MemberExpression {
+  return node.type === AST_NODE_TYPES.MemberExpression;
+}
```
tdeekens: 💯

packages/eslint-plugin-testing-library-react/src/rules/prefer-expect-query-by.ts
emmenko: Why would you want to do that? Yes, if it's a `CallExpression` you need to change the destructured method as well, but that it not a concern with this rule and it shouldn't be.

It's also stated in the eslint docs:

image

packages/eslint-plugin-testing-library-react/src/rules/prefer-expect-query-by.ts
```diff
@@ -0,0 +1,105 @@
...
+
+function isUsingWrongQueries(node: TSESTree.Identifier) {
```
tdeekens: Reading the code I am wondering if this isn't more `isCandidateWithWrongQuery` cause it depends on the `callee.name` being `expect` too, right?

packages/eslint-plugin-testing-library-react/src/rules/prefer-expect-query-by.ts
```diff
@@ -0,0 +1,105 @@
...
+  return node.name.startsWith('getBy') || node.name.startsWith('getAllBy');
+}
+
+// https://github.com/microsoft/TypeScript/issues/16069#issuecomment-369374214
+function isNotNullOrUndefined<T>(input: null | undefined | T): input is T {
+  return input != null;
+}
+
+function mapNodesForWrongGetByQuery(node: TSESTree.CallExpression) {
```
tdeekens: Relating to the comment above this is just finding the candidates, right?

packages/eslint-plugin-testing-library-react/src/rules/prefer-expect-query-by.ts
amine-benselim: makes sense yeah. maybe the second rule should be to not destruct rendered 😄

packages/eslint-plugin-testing-library-react/src/rules/prefer-expect-query-by.ts
```diff
@@ -0,0 +1,105 @@
...
+
+      return null;
+    })
+    .filter(isNotNullOrUndefined);
+}
+
+function hasExpectWithWrongGetByQuery(node: TSESTree.CallExpression) {
+  if (
+    node.callee &&
+    node.callee.type === AST_NODE_TYPES.Identifier &&
```
tdeekens: Is this `isIdentifier(node.callee)`?

packages/eslint-plugin-testing-library-react/src/rules/prefer-expect-query-by.ts
emmenko: Yeah, that would be a nice rule to have as well. Would you like to give it a try? 😉

packages/eslint-plugin-testing-library-react/test/rules.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  it('should have the correct amount of rules', () => {
+    const { length } = ruleNames;
+    if (length !== numberOfRules) {
+      throw new Error(
+        `There should be exactly ${numberOfRules} rules, but there are ${length}. If you've added a new rule, please update this number.`
+      );
+    }
+  });
+});
```
tdeekens: Should we in the spec also throw some code at our function above?

packages/eslint-plugin-testing-library-react/src/configs/recommended.json
emmenko: This is the "recommended" preset, so I would prefer to have it as an error as we want to "enforce" some best practices. In the default preset, it's turned off (I think).

packages/eslint-plugin-testing-library-react/src/rules/prefer-expect-query-by.ts
amine-benselim: challenge accepted 😄

packages/eslint-plugin-testing-library-react/src/rules/prefer-expect-query-by.ts
emmenko: Hmm well this function only checks if the name of the node is either `getBy` or `getAllBy`.

packages/eslint-plugin-testing-library-react/src/rules/prefer-expect-query-by.ts
emmenko: Yes, the nodes that refer to the `getBy` and `getAllBy`, which we also use for fixing the rule.

packages/eslint-plugin-testing-library-react/src/rules/prefer-expect-query-by.ts
emmenko: Yeah I guess we can use that.

packages/eslint-plugin-testing-library-react/test/rules.spec.ts
emmenko: I guess, I have to find an example how to test the rules.

ahmehri:
What's the motivation behind this rule? Doesn't this mean that it won't be possible anymore to use `getBy*` with `expect`? I mean it's still a valid use case if we're sure that the dom element should exist in the document.

emmenko:
If you want to use `getBy`, I recommend to use `await waitForElement(() => getBy*)`, otherwise there is no reason not to use `queryBy`

packages/eslint-plugin-testing-library-react/src/rules/prefer-expect-query-by.ts
ahmehri: Do you mean here avoid such `const { getByText } = render();` destructuring?
If yes, I think then we should agree first on why it's considered a bad practice before writing the rule. I personally don't see any harm in doing such destructuring.

amine-benselim: I have not strongly held opinion about it either. I just thought it'd be fun to write 🤷‍♂ 😅

ahmehri:
Using `await waitForElement(() => getBy*)` when you're sure that asynchronicity doesn't play any role is overkilling. If using `queryBy` instead of `getBy` is the mainstream, then why both of them are part of the api? We would have `queryBy` only instead.

packages/eslint-plugin-testing-library-react/src/rules/prefer-expect-query-by.ts
emmenko: Yes that destructuring.

It's not much about good or bad practice, it's just easier to work with. From my experience by not destructuring you get the following benefits:
- use the variable (e.g. `rendeded`) with autocompletion
- no need to destructure a new function everytime you need one
- makes it easier to swap things out

emmenko:
I really don't understand your argument to be honest. Using `getBy` throws an error, which is why it should be used for async functions like `waitForElement`. Using `queryBy` makes sense for expectations.

packages/eslint-plugin-testing-library-react/src/rules/prefer-expect-query-by.ts
ahmehri: If it's not considered a bad practice, then I wouldn't create a rule prohibiting it. I'm seeing it as a style rule and not a code quality rule. I personally don't have any issues with the points mentioned above.

ahmehri:
`getBy` throws an error if no elements match, which shouldn't happen if our expectation is the opposite. And we can still use it in sync context (without `waitForElement`), we have a lot of working examples in our codebase like this one.
https://github.com/commercetools/merchant-center-frontend/blob/e6981d49cd9b2af1d64b70cd3d4dbd6b2f6718ae/packages-application/application-products/components/pim-search-list/bulk-update/bulk-update-attributes-list-item-edit/bulk-update-attributes-list-item-edit.spec.js#L40
I'm expecting the element with label text `/attribute name/` to exist, if `getBy` throws an error then the desired behaviour is to fail the test.

emmenko:
I think you're misunderstanding here a bit. The rule is just about not using `getBy` in an `expect`.

In the link you posted, you are calling `getBy` on its own, assigning the element to a variable. This is fine as long as you don't use it in an `expect`.

Does it clear things out for you now?

packages/eslint-plugin-testing-library-react/docs/rules/prefer-expect-query-by.md
```diff
@@ -0,0 +1,28 @@
+# Avoid using `expect(getBy*)` (prefer-expect-query-by)
+
+React (DOM) Testing Library support two types of queries: `getBy*` and `queryBy*`. Using `getBy*` throws an error in case the element is not found. This is useful when using method like `waitForElement`, which are `async` functions and therefore throwing an error will cause the test to stop.
```
pa3: That is not quite accurate. According to RTL doc (https://testing-library.com/docs/dom-testing-library/api-async#waitforelement) error thrown by `getBy*` will cause `waitForElement` to keep waiting for element to appear, until timeout happens.

Usage of `getBy*` even have some advantages in scope of `waitForElement`. With `getBy*` we can wait for several elements in a single `waitForElement`, like this:

```
const [elementA, elementB] = await waitForElement(() => [getByText('elementA'), getByText('elementB')])```
```
, which is not possible with `queryBy`.

ahmehri:
Yes I get that, if we translate that to the previous example it will become:
`expect(getByLabelText(/attribute name/i)).toBeInTheDocument();`

- Assertion passes: good, my code is working as expected
- Assertion fails: bad, my code is not working as expected and I need to fix the test

What I'm saying here is that the above still a valid use case that we won't be able to use it anymore because of the rule.

At the end it's not a big deal because I tend to use `getBy` without `expect`. I just wanted to share my opinion.

packages/eslint-plugin-testing-library-react/docs/rules/prefer-expect-query-by.md
emmenko: Yes right, I'll clarify that point. Thanks for pointing that out 👌

emmenko:
> Yes I get that, if we translate that to the previous example it will become:

That's not entirely true, because you're not calling `getBy*` within the `expect`. So if the function throws, the code will never reach the `expect`.

Also to be clear, using `getBy*` might work as well with `expect`. That's not the point though, but in the cases where you want to assert that an element is **not** in the document, you can't use `getBy*` anymore in an `expect`. This rule aims to bring consistency in preferring to use `expect(queryBy*)`, to avoid unexpected "surprises".

packages/eslint-plugin-testing-library-react/docs/rules/prefer-expect-query-by.md
```diff
@@ -1,7 +1,7 @@
 # Avoid using `expect(getBy*)` (prefer-expect-query-by)
 
-React (DOM) Testing Library support two types of queries: `getBy*` and `queryBy*`. Using `getBy*` throws an error in case the element is not found. This is useful when using method like `waitForElement`, which are `async` functions and therefore throwing an error will cause the test to stop.
-However, when trying to assert if an element is not in the document, we obviously can't use `getBy*`. Instead it is recommended to use `queryBy*`, which does not throw and therefore we can assert that `expect(queryByText("Foo")).not.toBeInTheDocument()`.
+React (DOM) Testing Library support two types of queries: `getBy*` and `queryBy*`. Using `getBy*` throws an error in case the element is not found. This is useful when using method like `waitForElement`, which are `async` functions that will wait for the element to be found until a certain timeout, after that the test will fail.
```
pa3: ❤️

packages/eslint-plugin-testing-library-react/src/rules/prefer-expect-query-by.ts
```diff
@@ -58,7 +58,7 @@ function mapNodesForWrongGetByQuery(node: TSESTree.CallExpression) {
 function hasExpectWithWrongGetByQuery(node: TSESTree.CallExpression) {
   if (
     node.callee &&
-    isIdentifier(node.callee) &&
+    node.callee.type === AST_NODE_TYPES.Identifier &&
```
emmenko: Using `isIdentifier` here makes the compiler confused

image

PR #1015, opened 2019-08-30
fix(mc-http-server): to add referrer policy

tdeekens:
#### Summary

Adds a referrer policy header to the mc-http-server. More details on the interal PR.

### DIFF
--- a/packages/mc-http-server/server.js
+++ b/packages/mc-http-server/server.js
@@ -4,7 +4,6 @@ const path = require('path');
 const http = require('http');
 const morgan = require('morgan');
 const hidePoweredBy = require('hide-powered-by');
-const referrerPolicy = require('referrer-policy');
 const { createLightship } = require('lightship');
 const {
   createMiddleware: createPrometheusMetricsMiddleware,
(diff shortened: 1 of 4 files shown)


packages/mc-http-server/server.js
```diff
@@ -3,6 +3,8 @@ const fs = require('fs');
 const path = require('path');
 const http = require('http');
 const morgan = require('morgan');
+const hidePoweredBy = require('hide-powered-by');
```
tdeekens: Just so we use the same package across all services. It's not really needed but I somehow value consistency here.

tdeekens:
> Also, can you include the Feature-Policy as well? It was also mentioned in the document...

On another PR to keep things focussed, okay?

> We should define the header in the headers file, alongside the CSP, etc.

Can you elaborate why? As far as I read the code, through a call chain of 3-5 functions the `headers` are ending up in `createServerIndezMiddleware` which just tdoes `response.setHeader`. Why do you prefer having it from there if it's not configurable?

emmenko:
Using our http-server is not the only way of hosting an app, so the headers should be shipped together.

tdeekens:
Thanks for the explanation. 502d43f

PR #1016, opened 2019-08-30
feat: add ability to specify feature policies

tdeekens:
#### Summary

This pull request suggests to add the ability to add feature policies via the respective HTTP header.

#### Description

Feature policies (as outlined [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Feature-Policy)) are a way to restrict what browser features an application should be able to request on a given "allow-list".

An example would be to:

1. Allow the microphone only to be used in the document
2. Allow geolocation in an iframe

Note that this also overlapps (the example of the iFrame) with CSPs.

In general I would suggest, for now, not to default feature policies as:

1. It should (likely can) be considered a breaking change was all of a sudden a custom app can't e.g. use the mic
2. The merging is a bit more complicated as zipping up a `connect-src` whitelist

If we want to support meaningful defaults we coould do that later maybe. Right now I would suggest that the MC itself can disable most things using this feature.

### DIFF
--- a/packages/mc-html-template/README.md
+++ b/packages/mc-html-template/README.md
@@ -44,13 +44,14 @@ This method will attempt to load and parse the `env.json` file, performing some 
 
 This method will attempt to load and parse the `env.json` file, performing some validation and returning the parsed JSON.
 
-#### `loadHeaders(env: Object, { cspPath: String }): Object`
+#### `loadHeaders(env: Object, { cspPath: String, featurePoliciesPath: String }): Object`
 
 This method will return the security headers to be used on the server response, serving the `index.html`.
 
 The `env` argument, is the parsed `env.json` file (see `loadEnv`).
 
 Optionally, you can pass the path to the `csp.json` that contains custom CSP directives. Those will be merged with the default ones.
+Optionally, you can pass the path to the `feature-policies.json` that contains custom feature policy directives.
 
 The final headers object contains the following headers:
 
@@ -60,7 +61,8 @@ The final headers object contains the following headers:
   "X-XSS-Protection": "1; mode=block",
   "X-Content-Type-Options": "nosniff",
   "X-Frame-Options": "DENY",
-  "Content-Security-Policy": "..."
+  "Content-Security-Policy": "...",
+  "Feature-Policies": "..."
 }
 ```
 
--- a/packages/mc-html-template/README.md
+++ b/packages/mc-html-template/README.md
@@ -59,7 +59,7 @@ Optionally, you can pass the path to a `headers.json` that contains custom CSP a
 }
 ```
 
-The `cspPath` has been deprecated in favour of the `headerpath option. To migrate away from the`cspPath`please use the`headerPath`while moving the contents of a e.g.`csp.json`into e.g. a`headers.json`behind a`csp`property while passing`headerPath`.
+The `cspPath` has been **deprecated** in favour of the `headerpath` option. You can migrate to the new option by creating a `headers.json` (previously `csp.json`) and assigning the content of the `csp.json` into the `csp` field in the `headers.json` file.
 
 The final headers object contains the following headers:
 
--- a/packages/mc-html-template/README.md
+++ b/packages/mc-html-template/README.md
@@ -40,11 +40,7 @@ At the moment we define the following placeholders:
 
 This method will attempt to load and parse the `env.json` file, performing some validation and returning the parsed JSON.
 
-#### `loadEnv(configPath: String): Object`
-
-This method will attempt to load and parse the `env.json` file, performing some validation and returning the parsed JSON.
-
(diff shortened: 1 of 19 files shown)


emmenko:
Thanks for looking into that. I'm wondering though if it would make more sense to look again at #293 to "solve those issues" in a centralized manner. I'm also not super sure how that would work but maybe it's worth thinking about it first.

tdeekens:
For CSPs I tend to agree but also havent felt too much pain with the current approach. However feature policies I am not sure if we should control/centralize to default them as it's hard to judge what good defaults should be. Especially as extending them is a bit harder.

packages/mc-http-server/bin/start-mc.js
```diff
@@ -21,6 +21,7 @@ if (flags.help) {
   Options:
   --config=<path>           (required) The path to the environment config (defined as a JSON file, e.g. "env.json").
   --csp=<path>              (optional) The path to the custom CSP directives config (defined as a JSON file, e.g. "csp.json").
+  --featurePolicies=<path>  (optional) The path to the feature policy directives config (defined as a JSON file, e.g. "feature-policies.json").
```
emmenko: I'm not sure I like adding more CLI options, especially since this one is very specific.

I'd rather have one general file for specifying custom headers, like `headers.json`. We would then **deprecate** `csp.json` as it will be included in the `headers.json` object as `csp`.

```json
{
  "csp": {},
  "featurePolicies": ""
}
```

I think this would scale better and it would be easier to maintain.

tdeekens: Sure. Makes sense to me. Let me give it a shot with a deprecation message.

emmenko: Ok cool! 💯

tdeekens: 6f576b5 I presume we only use the deprecated file if the path is passed and the headers don't contain CSPs. In other words merging the contents of the deprecated option and the new headers would be odd ot me.

packages/mc-html-template/README.md
```diff
@@ -44,13 +44,24 @@ This method will attempt to load and parse the `env.json` file, performing some
 
 This method will attempt to load and parse the `env.json` file, performing some validation and returning the parsed JSON.
 
-#### `loadHeaders(env: Object, { cspPath: String }): Object`
+#### `loadHeaders(env: Object, { headersPath: String }): Object`
```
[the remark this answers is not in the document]
tdeekens: Sure. Was this sort of what you had in mind? 2135c3c1498baf6684dbf84b1a1a7d707af8c65c

packages/mc-html-template/README.md
```diff
@@ -40,17 +40,26 @@ At the moment we define the following placeholders:
...
+  "csp": {
+    "script-src": ["storage.googleapis.com/my-bucket-path/"]
+  },
+  "featurePolicies": {
+    "microphone": "none"
+  }
+}
+```
+
+The `cspPath` has been deprecated in favour of the `headerpath option. To migrate away from the`cspPath`please use the`headerPath`while moving the contents of a e.g.`csp.json`into e.g. a`headers.json`behind a`csp`property while passing`headerPath`.
```
emmenko:
```suggestion
The `cspPath` has been **deprecated** in favour of the `headerpath` option. You can migrate to the new option by creating a `headers.json` (previously `csp.json`) and assigning the content of the `csp.json` into the `csp` field in the `headers.json` file.
```

website/content/deployment/runtime-configuration.mdx
```diff
@@ -59,21 +59,41 @@ An example configuration for local development:
 
 ## `csp.json`
 
+This file as been deprecated by the `headers.json` mentioned below.
+
+To migrate nest the contents of your `csp.json` into the `csp`-property of the `headers.json`:
```
emmenko:
```suggestion
To migrate, nest the contents of your `csp.json` into the `csp`-property of the `headers.json`:
```

website/content/deployment/runtime-configuration.mdx
```diff
@@ -59,21 +59,41 @@ An example configuration for local development:
...
+
+```json
+{
++  "csp": {
```
emmenko:
```suggestion
  "csp": {
```

website/content/deployment/runtime-configuration.mdx
```diff
@@ -59,21 +59,41 @@ An example configuration for local development:
...
+    "script-src": ["my-apps.com"],
+    "connect-src": ["my-apps.com"],
+    "style-src": ["my-apps.com"]
++  }
```
emmenko:
```suggestion
  }
```

tdeekens:
Danke.

PR #1023, opened 2019-09-02
refactor(renovate): update config to have one pr per week

tdeekens:
#### Summary

This pull request changes renovate's behaviour to have one PR a week while splitting major and minor updates.

#### Description

This is often seen as an anti-pattern and was also a feature request to Greenkeeper as far as I recall.

The downside is, that one breaking dependency update it is hard to spot.
The upside is, that we don't need to merge all updates by hand.

The alternative would be to enable automerging. However, that comes at the downside of releasing a canary per dep update while having long changelogs as a result.

Ref to GitHub: https://github.com/renovatebot/renovate/issues/48

CleanShot 2019-09-02 at 11 07 16@2x
CleanShot 2019-09-02 at 11 07 28@2x

### DIFF
--- a/.renovaterc.json
+++ b/.renovaterc.json
@@ -2,7 +2,23 @@
   "extends": [
     "config:base",
     ":pinOnlyDevDependencies",
-    "group:monorepos",
-    "schedule:weekly"
+    "schedule:weekly",
+    {
+      "groupName": "all dependencies",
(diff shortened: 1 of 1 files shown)


tdeekens:
I assigned @montezume for awareness. I think we can try it here for a week or two and then move over to e.g. UIKIt if they deem this helpful too.

PR #1028, opened 2019-09-02
fix(app-shell): null check for dataFences in navbar component

emmenko:
Caused by #991


adnasa:
Thanks for the fix, @emmenko

Issue #1032, opened 2019-09-04
Package l10n:  Unsupported countries


islam3zzat:
**Describe the bug**
Package @commercetools-frontend/l10n

Adding any of the [addedd countries](https://github.com/commercetools/merchant-center-application-kit/pull/988/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2) in  #988
 e.g `cg-alt-variant `, `un-alt-short `, ...`us-alt-short` in the zones
causes issue.
![issue](https://user-images.githubusercontent.com/7680511/64244358-2e4d8b80-cf09-11e9-99a0-199e8af7fc94.gif)
After checking the Graphql definition for the ZoneLocation
- ![image](https://user-images.githubusercontent.com/7680511/64244559-91d7b900-cf09-11e9-9ada-6a62f7f5c81d.png)
- ![image](https://user-images.githubusercontent.com/7680511/64244649-b764c280-cf09-11e9-86fe-0b541a8a551f.png)

So the supported codes are those
https://en.wikipedia.org/wiki/ISO_3166-1#Officially_assigned_code_elements
**To Reproduce**

1. Go to `settings` the `international` tab;
2. Click on edit icon in the zones pane;
3. Add `un-alt-short` in the `Country codes` column of any row.

**Expected behavior**
The zone to be updated

emmenko:
Thanks for reporting. I guess we need to filter them out in the generate script

https://github.com/commercetools/merchant-center-application-kit/blob/d71122f41dae4ccbf6af2b3775bbc55eb78116d2/packages/l10n/scripts/generate-l10n-data.js#L40-L44

@islam3zzat do you want to open a PR?

islam3zzat:
I'm on it.

PR #1033, opened 2019-09-04
Remove unsupported countries

islam3zzat:
#### Summary

In this PR we remove some countries that are no supported
We only support ISO 3166-1 country code.
See https://en.wikipedia.org/wiki/ISO_3166-1

Fixes: #1032


packages/l10n/data/countries/de.json
```diff
@@ -40,13 +40,10 @@
   "ca": "Kanada",
   "cc": "Kokosinseln",
   "cd": "Kongo-Kinshasa",
-  "cd-alt-variant": "Kongo (Demokratische Republik)",
```
montezume: I wouldn't modify this directly as it's generated

packages/l10n/data/currencies/de.json
```diff
@@ -181,7 +181,7 @@
   },
   "DZD": {
     "label": "Algerischer Dinar",
-    "symbol": "د.ج.‏"
+    "symbol": "DA"
```
emmenko: Hmm why did those change? Did you install the latest deps?

islam3zzat: I thought I did!
Updated the deps, and re-ran the `generate-data`

packages/l10n/data/time-zones/de.json
```diff
@@ -4,8 +4,8 @@
     "abbr": "-12",
     "offset": "-12:00"
   },
-  "Pacific/Niue": {
-    "name": "Pacific/Niue",
+  "Etc/GMT+11": {
```
islam3zzat: I'm not sure why this has changed.
But my deps are updated, and this was generated when I ran `generate-data`.
Will be very helpful if someone regenerated locally to make sure that this is how it is supposed to be.

emmenko: I regenerated the data. I think you might need to re-install all deps.

islam3zzat: Thanks a lot.
Will keep that in mind next time.

PR #1034, opened 2019-09-05
chore: update ts-eslint to v2

emmenko:
Breaking changes release notes: https://github.com/typescript-eslint/typescript-eslint/releases/tag/v2.0.0

NOTE: the main question I have is about enabling the `@typescript-eslint/recommended-requiring-type-checking`, as it "slows down" the linting process as it needs to perform some typechecking as well.
From their docs they mention this:

image

I'm not sure how "splitting into stages" works though...

We can also keep the extra config disabled for now.


.eslintrc.yaml
```diff
@@ -4,8 +4,11 @@ parserOptions:
     jsx: true
   ecmaVersion: 8
   sourceType: module
+  project: './tsconfig.eslint.json'
```
tdeekens: Yeah did the same in flopflip last weekend :)

packages/application-shell/src/components/quick-access/butler/butler.js
```diff
@@ -249,37 +249,34 @@ export class Butler extends React.Component {
...
-                      }
-                    : null
-                );
-              } else {
-                this.shake();
-              }
-            },
-            () => this.setNetworkError()
-          );
+          const nextCommands = this.props.getNextCommands(command);
```
tdeekens: This is not sync, is it?

packages/mc-scripts/extract-intl.js
```diff
@@ -225,32 +225,30 @@ const extractFromFile = async fileName => {
   // We usually do not build the translation-files because they are
   // handled by transifex and sync by the transifex CLI.
   if (shouldBuildTranslations) {
-    await Promise.all(
-      locales.map(async locale => {
-        const translationFileName = `${outputPath}/${locale}.json`;
+    locales.forEach(locale => {
```
tdeekens: Ah so this was wrapping it into a promise needlessly?

emmenko:
Yeah well, for example when commiting, the lint-staged runs and takes much longer now.

packages/mc-scripts/extract-intl.js
emmenko: Yes, that was one of the new rules from the additional config

tdeekens: I like it. Async behaviour in JS is "toxic"/breaking to the consumer so should be avoided.

packages/application-shell/src/components/quick-access/butler/butler.js
emmenko: Yes it's sync. The commands return a plain object or a list of objects. Never a promise.

tdeekens:
We could also "split" the usage of the new rules and the update to v2.

emmenko:
What do you mean? The new rules are part of v2 as far as I understand...

tdeekens:
Split out the `@typescript-eslint/recommended-requiring-type-checking` into another PR from the upgrade to v2.

emmenko:
What would be the benefit of doing that? 🤔

tdeekens:
I thought that slows the linting down. If it doesn't I got it wrong.

emmenko:
Well yes it's slower now, I just don't understand what splitting it into a separate PR would solve. I would say we try that and if it's too painful we can turn it off...

PR #1043, opened 2019-09-09
Revert "fix(deps): update all dependencies"

emmenko:
Reverts commercetools/merchant-center-application-kit#1039

Something is wrong here, even though CI is green. See https://github.com/commercetools/merchant-center-application-kit/pull/1040#issuecomment-529658758


emmenko:
Hmm, I'm still getting that error...maybe it's not related to the dependencies?

Let me re-install everthing...

tdeekens:
On the weekend I had the strangest issue in flopflip with TS. I had to dedupe the lockfile and it worked. There is even a package to do it.

emmenko:
The only thing that worked for me so far was to increase the nodejs memory limit

```
export NODE_OPTIONS=--max_old_space_size=4096
```

Am I the only one with this problem? Can you run e.g. `yarn lint packages/sentry/src/sentry.spec.ts` without problems?

emmenko:
Ah, it's the `plugin:@typescript-eslint/recommended-requiring-type-checking` rules that is causing the OOM.

Without that, things are fine.

PR #1044, opened 2019-09-09
fix: js OOM caused by requiring type checking rules in eslint

emmenko:
- https://github.com/commercetools/merchant-center-application-kit/pull/1043#issuecomment-529688349
- https://github.com/commercetools/merchant-center-application-kit/pull/1040#issuecomment-529658758


tdeekens:
Tried `yarn lint packages/sentry/src/sentry.spec.ts` locally and it works but I am also fine with disabling this.

PR #1048, opened 2019-09-10
feat: validate graphql files using eslint and schema introspection

emmenko:
I think it's pretty much self explanatory. I've also documented the usage here: https://github.com/commercetools/merchant-center-application-kit/blob/b124a1427ed516196afb4d2f139c75c29b75a89f/CONTRIBUTING.md#graphql-files-and-linting

PS: I think it makes sense to keep the schemas in source control so that other people contributing to to the repo don't need to setup anything.


.graphqlconfig.yaml
```diff
@@ -0,0 +1,42 @@
+projects:
+  users:
```
tdeekens: I guess you are unhappy with the naming too. It is confuses me to call it users where it exposes more. Or what was your reasoning? Internal would also be a strange name. Maybe mcInternal or mcApi.

.graphqlconfig.yaml
```diff
@@ -0,0 +1,42 @@
...
+    schemaPath: schemas/users.json
+    includes:
+      - '**/*.users.graphql'
+    extensions:
+      endpoints:
+        production:
+          url: 'https://mc-api.commercetools.com/graphql'
+          headers:
+            X-Graphql-Target: mc
+  menus:
```
tdeekens: Also wondering. Maybe appShell?

.graphqlconfig.yaml
```diff
@@ -0,0 +1,42 @@
...
+        production:
+          url: 'https://mc.commercetools.com/api/graphql'
+  settings:
+    schemaPath: schemas/settings.json
+    includes:
+      - '**/*.settings.graphql'
+    extensions:
+      endpoints:
+        production:
+          url: 'https://mc-api.commercetools.com/graphql'
```
tdeekens: How would a work flow be btw where the schema is only on stage not on prod?

.graphqlconfig.yaml
emmenko: Yeah, couldn’t find a better name. We can also use `mc`

.graphqlconfig.yaml
emmenko: Hmm or maybe `proxy`? However the Schema is about menu configs, so I think it’s fine to keep it that way for now

.graphqlconfig.yaml
emmenko: I see two options:
- specify an additional `staging` environment
- pass the URL as an env variable

Given that we probably don’t want to expose the staging domain, we could go with option 2. What do you think?

.graphqlconfig.yaml
tdeekens: I couldn’t really stop thinking about it. I would likely also suggest to go with mc. Just to align it with the GraphQL targets we have. So it is easy to map and correlate even if the name is not perfect.

.graphqlconfig.yaml
tdeekens: I think for this menus is fine as it is not a GraphQL target. Still to be honest I would not mind proxy. Also feels like a good name as we name these as to the service they hit less to what content they return.

.graphqlconfig.yaml
tdeekens: I agree. I would make it an env variable while adding the prod urls into the env template do it's the default. Just wanted to make sure we somehow enable working with this before schema updates hit production.

emmenko:
I think we started hitting a wall with Netlify. Recent builds started to fail because the entire build process takes longer than 15min. What we currently run on Netlify is: installing deps, building packages, building website. Apparently, installing deps and building packages takes around 12min, which leaves a very small room for building the website with Gatsby.

We need to think about a solution here...

tdeekens:
> What we currently run on Netlify is: installing deps, building packages, building website. 

Yeah, maybe we can have a Docker container for it or use yarn's offline thing. Funny that it's for a website which isn't even public yet (silly pun intended) 😄.

CONTRIBUTING.md
```diff
@@ -164,3 +165,20 @@ On `master` branch, we automatically publish **canary** releases from CI to the
...
+
+In order to be able to validate GraphQL queries and mutations, defined as `.graphql` files, we use the [eslint-plugin-graphql](https://github.com/apollographql/eslint-plugin-graphql), which requires **introspection schemas** from the different GraphQL APIs being used.
+
+To download the remote schemas simply run `yarn get-schemas`. The configuration of each schema is defined in the `.graphqlconfig.yml` file, in the root directory. Running this script will download the schemas in the `schemas/*.json` files.
+
+> NOTE that you need your user `mcAccessToken` to be defined as an environment variable `MC_ACCESS_TOKEN` in `.env` file. This will be used by the introspection queries to be able to download the schemas from the MC API. Additionally, you also need to specify one of your `CTP_PROJECT_KEY` where you have access to.
+
+Since the MC uses multiple GraphQL APIs, we need to differentiate which queries use which schema. To do so, prefix the file extension with one of the GraphQL targets:
+
+- **users**: instead of `.graphql` use `*.mc.graphql` (_graphql target: `mc`_)
```
adnasa: I really like this!

PR #1049, opened 2019-09-11
fix: to avoid using Mapped Types as babel plugin does not properly generate prop-types

emmenko:
Reported the issue here: https://github.com/milesj/babel-plugin-typescript-to-proptypes/issues/23

image


adnasa:
Awesome find!!! 🔥

PR #1056, opened 2019-09-16
[experiment] Try Roboto font-family

montezume:
####

!not to merge

I need to do a beta release so that Sven can test this in the MC

Found a bug in UI Kit

https://github.com/commercetools/ui-kit/pull/1061

PR #1057, opened 2019-09-16
fix(app-shell): sanitize correlation id parts

pa3:
`X-Correlation-ID` header of requests sent to backend have to be -- quoting CT documentation -- "strings that may only contain alphanumeric characters, underscores and hyphens and have a length of 8 to 256 characters". 

`X-Correlation-ID` is being built using user-provided data (namely, projectKey). In case this data does not satisfies above-mentioned constraints, request fails with an error, describing malformed `correlation id`. This error, in turn is being reported to Sentry, which spoils FLD job. 

In order to avoid reporting this errors, this PR introduces `correlation id` sanitization. Original suggestion was to URL-encode provided `projectKey`, but that wont' work because, basically, `%` character would not pass `X-Correlation-ID` validation. 

Instead, we just gonna drop malformed parts.

### Paying for review

1. https://github.com/commercetools/merchant-center-application-kit/pull/1056#pullrequestreview-288502223
2. https://github.com/commercetools/ui-kit/pull/1061#pullrequestreview-288502543

### DIFF
--- a/packages/application-shell/src/utils/get-correlation-id/get-correlation-id.js
+++ b/packages/application-shell/src/utils/get-correlation-id/get-correlation-id.js
@@ -1,8 +1,13 @@
 import uuid from 'uuid/v4';
 import selectProjectKeyFromUrl from '../select-project-key-from-url';
 
+const VALID_ID_PART_FORMAT = /^[\w-/]+$/;
+
+const skipMalformedPart = part =>
+  Boolean(part) && VALID_ID_PART_FORMAT.test(part);
+
 export default function getCorrelationId({ userId } = {}) {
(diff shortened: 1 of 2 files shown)


packages/application-shell/src/utils/get-correlation-id/get-correlation-id.js
```diff
@@ -1,8 +1,13 @@
 import uuid from 'uuid/v4';
 import selectProjectKeyFromUrl from '../select-project-key-from-url';
 
+const VALID_ID_PART_FORMAT = /^[\w-/]+$/;
```
montezume: would you mind adding a comment explaining what this does for those of us stupid enough not to be regex fluent 😅

PR #1061, opened 2019-09-17
refactor(app-shell): start converting package to TS

emmenko:
Ref: #589 

Since the app-shell package is quite big, I will split the migration into separate PRs. For that to work, we need to use `allowJs` option so that TS can still compile normal `*.js` files.


packages/application-shell/src/utils/setup-global-error-listener/setup-global-error-listener.ts
```diff
@@ -22,15 +25,14 @@ export default function setupGlobalErrorListener() {
           'handled. This is most likely a bug in the software. Please ensure ' +
           'that the promise is correctly handled.'
       );
-    internalReduxStore.dispatch(
-      showUnexpectedErrorNotification({ error: { message: event.reason } })
```
emmenko: This was actually wrong! We didn't do anything with the `error`, as the notification only accepts a `errorId`.
Now we report the error to sentry and pass the sentry `errorId`.

packages/sentry/src/sentry.ts
```diff
@@ -21,45 +21,44 @@ export const boot = () => {
...
-        '"Error" should be preferred so that a stack-trace can be made available in sentry. ' +
+      '[SENTRY]: You called "reportErrorToSentry" with a string argument. ' +
+        '"Error" objects should be preferred so that a stack-trace can be made available in Sentry. ' +
         'See: https://docs.sentry.io/clients/javascript/usage/#try-catch'
     );
   }
 
   if (isEnabled) {
-    // logs error to sentry
-    Sentry.withScope(scope => {
```
emmenko: Since we want to return a value from the `reportErrorToSentry` function, we can't use a callback here. Luckily, Sentry exposes some methods from the static export, so we don't need to use `withScope`.

packages/application-shell/src/components/gtm-user-tracker/gtm-user-tracker.js
```diff
@@ -21,7 +21,7 @@ class GtmUserTracker extends React.PureComponent {
     if (this.props.user) this.syncUser();
   }
   syncUser = () => {
-    gtm.updateUser(this.props.user);
+    gtm.updateUser(this.props.user.id);
```
tdeekens: This was broken?

emmenko: No i just changed it to pass only the id, it’s easier to type instead of typing the entire user.
I guess with the generated types I could revert it back now

tdeekens: Ah okay. I don‘t mind either way.

emmenko:
I'll create a new base branch to work on the app-shell migration, so that at the end we merge only one big PR to master but the migration can be split into multiple little ones

packages/application-shell/src/utils/select-user-id/select-user-id.ts
```diff
@@ -0,0 +1,24 @@
+import ApolloClient from 'apollo-client';
+import { NormalizedCache } from 'apollo-cache-inmemory';
+import { TFetchUserIdQuery } from '../../types/generated/mc';
```
emmenko: That's what I'm talking about! 🎉

PR #1062, opened 2019-09-17
feat: generate TS types for graphql schemas

emmenko:
We can generate types of the schemas and the graphql operations using the graphql code generator: https://graphql-code-generator.com/docs/plugins/typescript-operations

As follow ups, we can start using those generated types where we need the types of the query results.


emmenko:
FYI: I improved a couple of things regarding the generated types. For instance:
- we now generate `d.ts` files for the graphql query document imports, instead of one generic `*.graphql` declaration. For that, we're using the plugin https://graphql-code-generator.com/docs/plugins/typescript-graphql-files-modules
- most of the queries are in the app-shell package. However, the app-shell-connectors also require some types that are derived from the queries. Therefore, besides generating types for all the schemas in the app-shell package, we also generate some of them in the app-shell-connectors. I tried different solutions but this is the only one I managed to make it work, which requires some duplication in the `codegen.yml`. 

For demo purposes, I also refactored the `project-extension-image-regex` connector to use the generated types.

emmenko:
Will merge when CI passes, so that I can continue with the migration

PR #1063, opened 2019-09-18
fix(app-shell): apollo header link to omit empty headers

tdeekens:
#### Summary

This pull request omits empty header values. Instead of passing potentially `null` through the links we remove the value.

#### Description

This is an attempt to reduce the "Failed to fetch"-errors. On some issues on Apollo it was said that doing this or passing `''` removed the issue. I think it might be worth a shot.

### DIFF
--- a/packages/application-shell/src/apollo-links/header-link.js
+++ b/packages/application-shell/src/apollo-links/header-link.js
@@ -1,4 +1,5 @@
 import { ApolloLink } from 'apollo-link';
+import omitEmpty from 'omit-empty-es';
 import { GRAPHQL_TARGETS } from '@commercetools-frontend/constants';
 import {
   getCorrelationId,
@@ -7,7 +8,9 @@ import {
   selectUserId,
 } from '../utils';
 
-const isKnownTarget = target => Object.values(GRAPHQL_TARGETS).includes(target);
+const isKnownGraphQlTarget = target =>
+  Object.values(GRAPHQL_TARGETS).includes(target);
+const omitEmptyHeaders = headers => omitEmpty(headers);
 
 /* eslint-disable import/prefer-default-export */
 // Use a middleware to update the request headers with the correct params.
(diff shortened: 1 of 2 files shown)


packages/application-shell/src/apollo-links/header-link.js
```diff
@@ -7,7 +8,9 @@ import {
   selectUserId,
 } from '../utils';
 
-const isKnownTarget = target => Object.values(GRAPHQL_TARGETS).includes(target);
+const isKnownGraphQlTarget = target =>
```
tdeekens: Just renaming I couldn't keep my fingers from.

packages/application-shell/src/apollo-links/header-link.js
```diff
@@ -7,7 +8,9 @@ import {
...
+  Object.values(GRAPHQL_TARGETS).includes(target);
+const omitEmptyHeaders = headers => omitEmpty(headers);
```
tdeekens: Just to give it a "meaningful" name.

packages/application-shell/src/apollo-links/header-link.js
```diff
@@ -43,12 +46,12 @@ const headerLink = new ApolloLink((operation, forward) => {
 
   operation.setContext({
     credentials: 'include',
-    headers: {
+    headers: omitEmptyHeaders({
```
tdeekens: Found this easier to follow than the "optional spreading" stuff.

packages/application-shell/src/apollo-links/header-link.js
emmenko: Nit: is this function necessary? Why not simply using `omitEmpty` directly?

emmenko: > Just to give it a "meaningful" name.

To be honest, I find `headers: omitEmpty({ ... })` already quite clear. But yeah, was just curious

PR #1064, opened 2019-09-18
refactor(app-context): to derive provider types from generated query types

emmenko:
Now that we have the types of the graphql queries generated, we can start make use of them.

For instance, the app context requires the user/project that are coming from the graphql queries. 

One thing to note about this refactoring is that I moved the normalize functions for the `allApplied-` fields to the app-context. This has a couple of advantages:
- we already map the user/project, so it feels more natural to do all transformations there
- types are more colocated for both the app context and the normalize functions
- the `FetchProject` component does not need to care about the transformation anymore


packages/application-shell-connectors/src/components/application-context/application-context.spec.tsx
```diff
@@ -41,12 +45,21 @@ describe('rendering', () => {
             countries: ['us'],
             currencies: ['USD'],
             languages: ['en'],
-            permissions: { canManageProjectSettings: true },
-            actionRights: null,
-            dataFences: null,
+            allAppliedPermissions: [
```
emmenko: The shape of the user/project that we pass to the context provider now maps 1:1 to the query response

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -66,62 +51,7 @@ type TApplicationContextGroupedByResourceType = {
...
-  project?: TRawProject;
-  projectDataLocale?: string;
-  children: React.ReactNode;
-};
-type ConsumerProps<AdditionalEnvironmentProperties extends {}> = {
-  render: (
-    context: TApplicationContext<AdditionalEnvironmentProperties>
-  ) => React.ReactNode;
-  children?: never;
+  store?: TApplicationContextGroupedByResourceType;
```
emmenko: NOTE: that this is now optional (see normalize function)

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -180,9 +137,10 @@ const createApplicationContext: <AdditionalEnvironmentProperties extends {}>(
   environment,
   user: mapUserToApplicationContextUser(user),
   project: mapProjectToApplicationContextProject(project),
-  permissions: project && project.permissions ? project.permissions : null,
-  actionRights: project && project.actionRights ? project.actionRights : null,
-  dataFences: project && project.dataFences ? project.dataFences : null,
+  permissions: normalizeAllAppliedPermissions(project),
+  actionRights: normalizeAllAppliedActionRights(project),
+  menuVisibilities: normalizeAllAppliedMenuVisibilities(project),
+  dataFences: normalizeAllAppliedDataFences(project),
```
emmenko: This is now the only place where we do all the necessary mapping.

packages/application-shell-connectors/src/components/application-context/normalizers.ts
```diff
@@ -0,0 +1,218 @@
...
+ *     }
+ *   }
+ * }
+ */
+type NormalizedStoresGroupByResourceType = {
+  [key: string]: TStoreDataFence[];
+};
+
+const normalizeAppliedDataFencesForStoresByResourceType = (
+  dataFences: Maybe<TStoreDataFence>[]
```
tdeekens: Is this a regular `type Maybe<T> = T | null`?

packages/application-shell-connectors/src/components/application-context/normalizers.ts
```diff
@@ -0,0 +1,218 @@
...
+          store: normalizeAppliedDataFencesForStoresByResourceType(
+            groupedByType.store
+          ),
+        }
+      : {}),
+  };
+  return Object.keys(normalizedDataFences).length > 0
+    ? normalizedDataFences
+    : null;
+};
```
tdeekens: This is nice. I assume the contents/behaviour here didn't change. It's hard/impossible to trace those changes without very small commits while renaming and moving happens.

packages/permissions/src/components/restricted-by-permissions/restricted-by-permissions.spec.tsx
```diff
@@ -3,29 +3,42 @@ import { ApplicationContextProvider } from '@commercetools-frontend/application-
...
+  group: string;
 };
-type TActionRights = {
-  [key: string]: TActionRight;
+type TAllAppliedMenuVisibility = {
+  name: string;
+  value: boolean;
+};
+type TAllAppliedDataFence = {
+  __typename: 'StoreDataFence';
```
tdeekens: Hm. @adnasa shouldn't this be not here?

packages/permissions/src/components/restricted-by-permissions/restricted-by-permissions.spec.tsx
```diff
@@ -3,29 +3,42 @@ import { ApplicationContextProvider } from '@commercetools-frontend/application-
...
+  allAppliedMenuVisibilities = [],
+  allAppliedDataFences = [],
   component,
 }: {
-  actualPermissions?: TPermissions;
-  actualActionRights?: TActionRights;
+  allAppliedPermissions?: TAllAppliedPermission[];
+  allAppliedActionRights?: TAllAppliedActionRight[];
+  allAppliedMenuVisibilities?: TAllAppliedMenuVisibility[];
+  allAppliedDataFences?: TAllAppliedDataFence[];
```
tdeekens: Will this result in a breaking change for `test-utils`? No right?

packages/application-shell-connectors/src/components/application-context/normalizers.ts
```diff
@@ -0,0 +1,218 @@
...
+        ...previousGroupsOfSameType,
+        [appliedDataFence.type]: [...(previousGroup || []), appliedDataFence],
+      };
+    },
+    {}
+  );
+  const normalizedDataFences = {
+    ...('store' in groupedByType
+      ? {
+          store: normalizeAppliedDataFencesForStoresByResourceType(
```
emmenko: NOTE that this function is specific to the `store` type. Therefore, we can use this function to explicitly normalize the store values.
In the future, if we have more data fence types, we can have specific normalize functions for each one of them. In case some share the same format, we can reuse the normalize functions.
However, for now I would prefer to keep it explicit.

Also note that the return type of the `normalizeAllAppliedDataFences` can be `null` or an object optionally containing a `store` key, hence the need for this: https://github.com/commercetools/merchant-center-application-kit/pull/1064/files#r325633637

packages/application-shell/src/components/application-shell/application-shell.js
```diff
@@ -297,9 +297,6 @@ export const RestrictedApplication = props => (
                                     <NavBar
                                       applicationLocale={locale}
                                       projectKey={projectKeyFromUrl}
-                                      menuVisibilities={
-                                        project.menuVisibilities
```
emmenko: This is now read from the app context within the `NavBar` component

packages/application-shell-connectors/src/components/application-context/normalizers.ts
emmenko: It's in the generated types: by default is `export type Maybe<T> = T | null;` but we map it to `export type Maybe<T> = T | undefined;`
See: https://github.com/commercetools/merchant-center-application-kit/blob/bd3eb0d9b2817c1bab08a82e3f558c2853167223/codegen.yml#L47

packages/application-shell-connectors/src/components/application-context/normalizers.ts
emmenko: Yeah, it basically didn't change, I only added the type annotations

packages/permissions/src/components/restricted-by-permissions/restricted-by-permissions.spec.tsx
emmenko: The generated types marked this as required, I assume because of the union type expansion

https://github.com/commercetools/merchant-center-application-kit/blob/bd3eb0d9b2817c1bab08a82e3f558c2853167223/packages/application-shell-connectors/src/types/generated/mc.ts#L824

packages/permissions/src/components/restricted-by-permissions/restricted-by-permissions.spec.tsx
emmenko: Ah right, I forgot about that.

Hmm well in theory yes...maybe we can be backwards compatible and denormalize the value to be passed to the context provider.

tdeekens: Yeah I think we might have to while logging something maybe. Bit of a annoying side-effect of this.

emmenko: Yeah, I didn't think about this... 🙃

emmenko: https://github.com/commercetools/merchant-center-application-kit/pull/1064/commits/a1f3854ebb73685f2f1b32e4a5a0e0dfd6fb648c

emmenko:
Alright, I'd like to merge this PR if there are no more objections, so that we can continue with the app-shell migration

emmenko:
I tested the bundles in the mc apps, all good then!

emmenko:
Will merge this later

PR #1079, opened 2019-10-01
fix: modal page header title spacing

jonnybel:
The current style (with `& * + *`) doesn't work for cases where a user passed an element which will only render text without any wrapping HTML element tag (such as `FormattedMessage`), and would also break the style of any passed elements with more than one children.
The usage of `!important` could also be annoying in some cases.

This PR fixes this without breaking the current layout.


jonnybel:
I approved the only Percy diff which was completely unrelated to this PR.

PR #1080, opened 2019-10-02
data locale should be one of the project languages

amine-benselim:
this leads to tests failing when using 'LocalizedTextInput.createLocalizedString' because we pass the locales from the project and then in the test we check the locale the user chose.

### DIFF
--- a/packages/application-shell/src/test-utils/test-utils.js
+++ b/packages/application-shell/src/test-utils/test-utils.js
@@ -38,7 +38,7 @@ const defaultProject = {
   name: 'Test with big data',
   countries: ['de', 'en'],
   currencies: ['EUR', 'GBP'],
-  languages: ['de', 'en-GB'],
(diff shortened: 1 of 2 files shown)


packages/application-shell/src/test-utils/test-utils.js
```diff
@@ -268,7 +268,7 @@ const renderApp = (
     permissions, // <-- deprecated option, use `{ project: { allAppliedPermissions } }`
     actionRights, // <-- deprecated option, use `{ project: { allAppliedActionRights } }`
     dataFences, // <-- deprecated option, use `{ project: { allAppliedDataFences } }`
-    dataLocale = 'en',
```
emmenko: Ah good catch. Maybe we can add `en` to the project languages list instead.

amine-benselim: I think that way it's simpler. will do that.

PR #1081, opened 2019-10-08
chore: update yarn to 1.19.0 due to security vulnerability

tdeekens:
#### Summary

As you may know the older versions of yarn have a security vulnerability. Given that we now also use this version on dev machines and don't allow them to use their own there should be even more reason to keep is up-to-date.


emmenko:
VRTs are flaky, ignore them

PR #1082, opened 2019-10-08
refactor(app-shell): convert redux store setup to TS

emmenko:
Converts the `configure-store` module to TS and the related imports


packages/application-shell/src/configure-store.ts
```diff
@@ -39,23 +71,22 @@ const sdkMiddleware = createSdkMiddleware({
...
-  additionalMiddlewares = []
+  additionalMiddlewares: Middleware[] = []
 ) => {
   const store = createStore(
     createInternalReducer(),
     preloadedState,
     compose(
       applyDefaultMiddlewares(
         ...additionalMiddlewares,
-        sentryTrackingMiddleware,
```
emmenko: This middleware wasn't necessary anymore as:
- in the global error handler we are already sending the error to sentry
- in the handleActionError we also send the error to sentry now

I found that having the middleware caused some unnecessary indirection of the logic. It should be easier to follow along now.

packages/application-shell/src/configure-store.ts
```diff
@@ -18,16 +29,37 @@ import {
...
 
 const patchedGetCorrelationId = () =>
   getCorrelationId({
     userId: selectUserId({ apolloCache: apolloClient }),
   });
 
-const createInternalReducer = (injectedReducers = {}) =>
-  combineReducers({
+const createInternalReducer = (injectedReducers: ReducersMapObject = {}) =>
+  combineReducers<unknown>({
```
emmenko: This is a bit of a "workaround" because of this:

image

adnasa: I think I know I'm familiar with this problem a bit too well.
code comment that this is a workaround and what the original problem was?

packages/application-shell/src/components/requests-in-flight-loader/reducer.ts
```diff
@@ -0,0 +1,41 @@
...
+const isHideRequestInFlightAction = (
+  action: Action
+): action is THideRequestInFlightAction => action.type === HIDE_LOADING;
+
+const excludeFirstOccurrence = <Item>(list: Item[], item: Item) => {
+  const index = list.indexOf(item);
+  return [...list.slice(0, index), ...list.slice(index + 1)];
+};
+
+export default (requestsInFlight?: string[], action?: Action) => {
```
tdeekens: Nit: should we type this to `type RequestInFlight = string` and `type RequestsInFlight = RequestInFlight[]`?

packages/application-shell/src/components/requests-in-flight-loader/requests-in-flight-loader.spec.tsx
```diff
@@ -0,0 +1,37 @@
...
+describe('rendering', () => {
+  let rendered: RenderResult;
+  describe('when there are no requests in flight', () => {
+    beforeEach(() => {
+      rendered = renderAppWithRedux(<RequestsInFlightLoader />, {
+        storeState: { requestsInFlight: [] },
+      });
+    });
+    it('should not render loading spinner', async () => {
+      await wait(
```
tdeekens: Could we just return this promise?

packages/application-shell/src/components/requests-in-flight-loader/requests-in-flight-loader.spec.tsx
```diff
@@ -0,0 +1,37 @@
...
+    });
+  });
+  describe('when there are requests in flight', () => {
+    beforeEach(() => {
+      rendered = renderAppWithRedux(<RequestsInFlightLoader />, {
+        storeState: { requestsInFlight: ['one', 'two'] },
+      });
+    });
+    it('should render loading spinner', async () => {
+      await waitForElement(() => rendered.queryByText(/^Processing/));
```
tdeekens: ^^

packages/application-shell/src/configure-store.ts
```diff
@@ -18,16 +29,37 @@ import {
...
+  injectReducers: (payload: {
+    id: string;
+    reducers: ReducersMapObject;
+  }) => void;
+  removeReducers: (payload: { id: string }) => void;
+};
+
+const identityStoreEnhancer: StoreEnhancer<unknown, {}> = noop => noop;
+
+const mergeObjectValues = <T>(object: { [key: string]: T }) =>
```
tdeekens: What is this for?

packages/application-shell/src/configure-store.ts
emmenko: Yeah I can do that for now, maybe we can find a better solution in the future.

emmenko:
FYI: don't worry about the failing CI, it's expected since the migration of the package is not finished yet. This is just one small part of it. Eventually the base branch will be the reference one.

packages/application-shell/src/test-utils/test-utils.tsx
```diff
@@ -272,19 +336,25 @@ const renderApp = (
...
+  const mergedEnvironment = mergeOptional<
+    ProviderProps<AdditionalEnvironmentProperties>['environment']
+  >(defaultEnvironment, environment);
+  const mergedGtmTracking = { ...defaultGtmTracking, ...gtmTracking };
 
-  const ApplicationProviders = ({ children }) => (
+  const ApplicationProviders = (props: TApplicationProvidersProps) => (
     <IntlProvider locale={locale}>
       <ApolloProviderComponent mocks={mocks} addTypename={addTypename}>
         <ConfigureFlopFlip
```
emmenko: FYI: the types are currently failing here because of this: https://github.com/tdeekens/flopflip/pull/890

PR #1084, opened 2019-10-10
refactor(app-context): restructure tests to RTL

emmenko:


packages/application-shell-connectors/src/components/application-context/application-context.spec.tsx
```diff
@@ -1,141 +1,172 @@
 import React from 'react';
-import { shallow, ShallowWrapper } from 'enzyme';
-import { ApplicationContextProvider } from './application-context';
+import { render, waitForElement, RenderResult } from '@testing-library/react';
```
tdeekens: Why don't we use our own test-utils which exports this?

packages/application-shell-connectors/src/components/application-context/application-context.spec.tsx
```diff
@@ -1,141 +1,172 @@
...
-          },
-        })
-      );
+const renderAppWithContext = (ui: React.ReactElement) =>
+  render(
+    <ApplicationContextProvider<AdditionalEnvironmentProps>
+      user={createTestUser()}
+      project={createTestProject()}
+      environment={createTestEnvironment()}
+      projectDataLocale="en"
```
tdeekens: Is this a valid locale?

packages/application-shell-connectors/src/components/application-context/application-context.spec.tsx
```diff
@@ -1,141 +1,172 @@
...
+      />
+    );
+  });
+  it('should render project name from context', () =>
+    waitForElement(() => rendered.getByText('Project name: Foo 1')));
+});
+
+describe('useApplicationContext', () => {
+  let rendered: RenderResult;
+  const App = () => {
```
tdeekens: Nit `TestComponent`.

packages/application-shell-connectors/src/components/application-context/application-context.spec.tsx
```diff
@@ -1,141 +1,172 @@
...
+    const projectName = useApplicationContext(
+      context => context.project && context.project.name
+    );
+    return <div>{`Project name: ${projectName}`}</div>;
+  };
+  beforeEach(() => {
+    rendered = renderAppWithContext(<App />);
+  });
+  it('should render project name from context', () =>
+    waitForElement(() => rendered.getByText('Project name: Foo 1')));
```
tdeekens: Why do we need to waitFor here?

packages/application-shell-connectors/src/components/application-context/application-context.spec.tsx
emmenko: It will introduce a cycle dependency, which I would like to avoid. Unless we extract the test-utils into a separate package...

packages/application-shell-connectors/src/components/application-context/application-context.spec.tsx
emmenko: Hmm what do you mean?

packages/application-shell-connectors/src/components/application-context/application-context.spec.tsx
emmenko: Cause the context or project can be null on the first render

emmenko: To be clear: in the apps this is not a problem, cause the app gets rendered when the user/project are defined. However, since we're just testing the connector here, we need to account for this case.

packages/application-shell-connectors/src/components/application-context/application-context.spec.tsx
tdeekens: Hm. Okay. Maybe it then actually makes sense to have it on app-kits roadmap to have it as a package?

emmenko: Could be, but also it's only a "problem" with some of the app kit packages. In theory, using the test-utils is not necessary for most of them, since there is no need for an app-like setup.

packages/application-shell-connectors/src/components/project-extension-image-regex/project-extension-image-regex.spec.tsx
```diff
@@ -126,7 +126,11 @@ describe('rendering', () => {
       expect(rendered.queryByText('Not found')).toBeInTheDocument();
     });
   });
-  describe('when request fails', () => {
+  // TODO: the test is a bit flaky, where the `onError` callback does not get called
+  // consistently. This seems to be an internal issue with the apollo client.
+  // For now we can skip the test and maybe check again if it works in future versions
+  // of apollo.
+  describe.skip('when request fails', () => {
```
tdeekens: Should we then also change the implementation? I think we will actually never change this back knowing how we handled stuff like this in the past :)

tdeekens: Then lets remove it.

PR #1085, opened 2019-10-11
chore: update deps

tdeekens:
#### Summary

Manually updating some deps prior to releasing.


cypress/.eslintrc.yaml
```diff
@@ -3,4 +3,4 @@ plugins:
 env:
   cypress/globals: true
 rules:
-  prefer-object-spread/prefer-object-spread: false
+  prefer-object-spread/prefer-object-spread: 0
```
tdeekens: The config changed for the rule.

package.json
```diff
@@ -160,7 +160,7 @@
     "@types/react": "16.9.5",
     "@types/react-router": "5.1.1",
     "react": "16.10.2",
-    "**/react": "16.10.1",
+    "**/react": "16.10.2",
```
tdeekens: Otherwise we won't get the latest. Will try to remove resolutions in a follow up.

packages/mc-html-template/lib/load-env.js
```diff
@@ -35,7 +35,7 @@ module.exports = configPath => {
     servedByProxy: false,
   };
 
-  // eslint-disable-next-line prefer-object-spread/prefer-object-spread
```
tdeekens: This rule doesn't exist anymore and doesn't have to be disabled.

PR #1091, opened 2019-10-14
fix(app-shell): use first input delay

tdeekens:
#### Summary

Time to interactive is a metric which is not recommended anymore. It also is not exposed by most metric libraries as a result.

With this we use the First Input Delay which is said to reflect more what users experience. 

Before this is rolled out in our applications I will make sure our prometheus gateway is configured to consume it.


tdeekens:
CleanShot 2019-10-14 at 15 42 23@2x

PR #1102, opened 2019-10-17
feat(permissions): allow appliedPermission with Manage to bypass dataFence

adnasa:
#### Summary

If user has `*Manage` as their appliedPermissions, we simply discard dataFence since they would have permissions to manage orders, regardless what dataFence Permissions they have on their team.

This is because General Permissions overrules DataFence Permissions in the case of GeneralPermissions has `*Manage`

From an implementation POV, it doesn't look great that we "leak" appliedPermissions into `hasAppliedDataFence`, but I was not really sure how I want to ensure separate of concerns

### DIFF
--- a/packages/permissions/src/components/authorized/authorized.spec.tsx
+++ b/packages/permissions/src/components/authorized/authorized.spec.tsx
@@ -208,8 +208,8 @@ describe('rendering', () => {
   });
 
   describe('dataFence to view orders on specific store', () => {
-    describe('if cannot view or manage orders', () => {
-      describe('has demanded dataFence to MANAGE orders on the specific store', () => {
+    describe('with general permission', () => {
+      describe('when general permissions is manage', () => {
         beforeEach(() => {
           props = createTestProps({
             actualDataFences: {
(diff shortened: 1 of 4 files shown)


packages/permissions/src/utils/has-permissions.ts
```diff
@@ -234,43 +236,52 @@ const getDataFenceByPermissionGroup = (
   return null;
 };
 
-export const hasAppliedDataFence = (options: TOptionsForAppliedDataFence) =>
+export const hasAppliedDataFence = (options: TOptionsForAppliedDataFence) => {
+  // If user has `*Manage` as their `appliedPermission`, we bypass `dataFence`.
+  const hasAppliedManagePermission = options.demandedPermissions.some(
+    demandedPermission =>
+      hasManagePermission(demandedPermission, options.actualPermissions)
```
adnasa: I commited with `--no-verify` because I don't really know how to deal with this.

image

emmenko: The message explains well what is happening.

Basically `hasManagePermission` expects a `actualPermissions: TPermissions` but `options.actualPermissions` is of type `TPermissions | null`.

Therefore, I suppose you would do something like this:

```js
demandedPermission => {
  if (options.actualPermissions) {
    return hasManagePermission(demandedPermission, options.actualPermissions);
  }
  return false;
}
```

packages/permissions/src/components/authorized/authorized.spec.tsx
```diff
@@ -222,9 +222,84 @@ describe('rendering', () => {
...
+                  case 'store':
+                    return ['store-1'];
+                  default:
+                    return null;
+                }
+              },
+            });
+            shallow(<Authorized {...props} />);
+          });
+          it('should return true', () => {
```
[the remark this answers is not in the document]
adnasa: 🤦‍♂

packages/permissions/src/utils/has-permissions.ts
adnasa: aaah. thanks, bud!

adnasa: fixed ;)

packages/permissions/src/components/authorized/authorized.spec.tsx
```diff
@@ -222,9 +222,84 @@ describe('rendering', () => {
...
+                case 'store':
+                  return ['store-1'];
+                default:
+                  return null;
+              }
+            },
+          });
+          shallow(<Authorized {...props} />);
+        });
+        it('should return true', () => {
```
tdeekens:
```suggestion
        it('should indicate being authorized', () => {
```

packages/permissions/src/components/authorized/authorized.spec.tsx
```diff
@@ -222,9 +222,84 @@ describe('rendering', () => {
...
+                  case 'store':
+                    return ['store-1'];
+                  default:
+                    return null;
+                }
+              },
+            });
+            shallow(<Authorized {...props} />);
+          });
+          it('should return false', () => {
```
tdeekens:
```suggestion
          it('should indicate not being authorized', () => {
```

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -234,43 +236,59 @@ const getDataFenceByPermissionGroup = (
...
+      if (options.actualPermissions) {
+        return hasManagePermission(
+          demandedPermission,
+          options.actualPermissions
+        );
+      }
+      return false;
+    }
+  );
+  if (hasAppliedManagePermission) return true;
```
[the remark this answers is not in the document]
adnasa: yepp!

PR #1107, opened 2019-10-21
chore(app-shell): update flopflip to allow passing `featureFlags`

tdeekens:
#### Summary

This pull request builds on the latest `flopflip` release to allow passing a `featureFlags` prop. This should allow us to simplify maintaining long-running flags in LaunchDarkly.

https://github.com/tdeekens/flopflip/releases/tag/%40flopflip%2Freact-redux%409.1.0

This will be released as a chore/fix as it's not part of out documented APIs.


application-templates/starter/src/components/entry-point/entry-point.js
```diff
@@ -55,6 +56,7 @@ class EntryPoint extends React.Component {
         DEV_ONLY__loadNavbarMenuConfig={() =>
           import('../../../menu.json').then(data => data.default || data)
         }
+        featureFlags={FEATURE_FLAGS}
```
tdeekens: This trick actually helps so that custom apps don't get any flags at all. Otherwise the whole point would defeated. Regardless, custom apps being out there receiving all flags will kind of delay the helpfulness of this.

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.js
```diff
@@ -68,6 +71,7 @@ export class SetupFlopFlipProvider extends React.PureComponent {
             : ldClientSideIdStaging,
           this.props.user && this.props.user.id,
           this.props.projectKey,
+          this.props.flags,
```
tdeekens: Note, that `flags` are part of `adapterArgs` in `flopflip` as not all adapter do, will and can support this feature meainingfully.

application-templates/starter/src/components/entry-point/entry-point.js
emmenko: Nice!

adnasa:
> This will be released as a chore/fix as it's not part of out documented APIs.

btw.. this won't mean releasing permissions too, right?

tdeekens:
> btw.. this won't mean releasing permissions too, right?

Sorry, I don't follow. Can you elaborate?

PR #1109, opened 2019-10-21
fix(permissions/has-permissions): use `hasExactPermission` in `hasAppliedDataFences`

adnasa:
#### Summary

- use `startsWith` and `hasExactPermissions` for `hasAppliedDataFences`

#### Description

previously I wrote this to use `hasManagePermission`. The issue is that `hasManagePermissions` returns a truth value if `demandedPermission` starts with `View*`. 

In the case of `hasAppliedDataFence`, we want to:

1. check if the demanded Permissions is `Manage*`
2. use `hasExactPermission` to `actualPermissions`

### DIFF
--- a/packages/permissions/src/components/authorized/authorized.spec.tsx
+++ b/packages/permissions/src/components/authorized/authorized.spec.tsx
@@ -212,6 +212,7 @@ describe('rendering', () => {
       describe('when general permissions is manage', () => {
         beforeEach(() => {
           props = createTestProps({
+            demandedPermissions: ['ManageOrders'],
             actualDataFences: {
               store: {
                 orders: {
@@ -250,6 +251,7 @@ describe('rendering', () => {
         describe('when data fence permission is manage', () => {
           beforeEach(() => {
             props = createTestProps({
+              demandedPermissions: ['ViewOrders'],
               actualDataFences: {
                 store: {
                   orders: {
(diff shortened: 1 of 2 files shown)


packages/permissions/src/components/authorized/authorized.spec.tsx
```diff
@@ -212,6 +212,7 @@ describe('rendering', () => {
       describe('when general permissions is manage', () => {
         beforeEach(() => {
           props = createTestProps({
+            demandedPermissions: ['ManageOrders'],
```
adnasa: this test broke after applying the change.

This worked previously because the demandedPermission was `ViewProduct` and since there was no match, we did not do any early return and where we "fall back" to checking through `allAppliedDataFences` (which has canManageOrders`

adnasa:
I already tried this locally (hacking it) and it works as expected

```js
 var hasAppliedManagePermission = _someInstanceProperty(_context2 = options.demandedPermissions).call(_context2, function (demandedPermission) {
    if (options.actualPermissions) {
      return _startsWithInstanceProperty(demandedPermission)
      .call(demandedPermission, 'Manage') &&
      options.actualPermissions[toCanCase(demandedPermission)]
    }

    return false;
  });
  if (hasAppliedManagePermission) return true;
```

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -241,9 +241,9 @@ export const hasAppliedDataFence = (options: TOptionsForAppliedDataFence) => {
   const hasAppliedManagePermission = options.demandedPermissions.some(
     demandedPermission => {
       if (options.actualPermissions) {
-        return hasManagePermission(
-          demandedPermission,
-          options.actualPermissions
+        return (
+          demandedPermission.startsWith('Manage') &&
```
tdeekens: Can't we use `hasManagePermission` here?

emmenko: Let's leave a comment as to why it needs to check for this.

PR #1116, opened 2019-10-24
chore(mc-http-server): to expose prefixed metric

tdeekens:
#### Summary

This pull request comes after OPs intend to prefix request timing and application label metrics. Here we only have one request timing metric.

Why prefix? Cause cardinality explosions in Prometheus occur if you have a single, generically named metric, shared across services with lots of different labels. Promtheus can deal better with different metric names with similar labels. All due to its storage model. This all is aligned with best practices from the web and books.

Note, that by specifying two metrics here we expose both names and are backwards compatible for our dashboards. We can do the clean up after and expose just one.


adnasa:
nice

tdeekens:
This is small change and I started things locally and am confident with it. I'll merge to continue on the topic. Thanks for the view.

PR #1117, opened 2019-10-24
refactor(mc-http-server): startup logic

tdeekens:
#### Summary

This pull request carries out the same refactoring we did on all services in regards to: start up logic and graceful shutdown while splitting code into the same functions.

### DIFF
--- a/packages/mc-http-server/bin/start-mc.js
+++ b/packages/mc-http-server/bin/start-mc.js
@@ -10,7 +10,7 @@ const { compileHtml } = require('@commercetools-frontend/mc-html-template');
 const {
   packageLocation: applicationStaticAssetsPath,
 } = require('@commercetools-frontend/assets');
-const startServer = require('../server');
+const launchServer = require('../server');
 
 const flags = mri(process.argv.slice(2), { alias: { help: ['h'] } });
 
@@ -101,7 +101,7 @@ const start = async () => {
   );
 
   // Start the server
-  await startServer({
+  await launchServer({
     env: compiled.env,
     headers: compiled.headers,
     paths,
--- a/packages/mc-http-server/server.js
+++ b/packages/mc-http-server/server.js
@@ -17,13 +17,54 @@ const express = require('express');
 const { createLogoutHandler, createLoginHandler } = require('./routes');
 
 // Config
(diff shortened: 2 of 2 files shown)


packages/mc-http-server/bin/start-mc.js
```diff
@@ -10,7 +10,7 @@ const { compileHtml } = require('@commercetools-frontend/mc-html-template');
 const {
   packageLocation: applicationStaticAssetsPath,
 } = require('@commercetools-frontend/assets');
-const startServer = require('../server');
+const launchServer = require('../server');
```
tdeekens: Anywhere else we name the export `launchServer` whereas start is the `listen` callback.

packages/mc-http-server/server.js
```diff
@@ -17,13 +17,54 @@ const express = require('express');
...
+const applicationServerPort = process.env.HTTP_PORT || 3001;
+const prometheusMetricsServerPort = 7788;
+const applicationServerUri = `http://localhost`;
+const applicationServerUrl = `${applicationServerUri}:${applicationServerPort}`;
+
+let lightshipServer;
+let applicationServer;
+let prometheusMetricsServer;
+
+const createLightshipServer = options => {
```
tdeekens: This is all the same plit into things as on these other PRs.

packages/mc-http-server/server.js
```diff
@@ -17,13 +17,54 @@ const express = require('express');
...
+  return lightship;
+};
+
+const startServer = (server, port) =>
+  new Promise((resolve, reject) => {
+    server.listen(port, error => {
+      if (error) reject(error);
+      else resolve();
+    });
+  });
```
tdeekens: Wrapped too here so we can simplify logic below in async/await.

packages/mc-http-server/server.js
```diff
@@ -75,7 +116,7 @@ const throwIfIndexHtmlIsMissing = options => {
   }
 };
 
-const startServer = options => {
+const configureApplication = options => {
```
tdeekens: Extracted this as we do this in all our other servers too.

tdeekens:
Did an additional integration setup test. All fine.

CleanShot 2019-10-24 at 14 51 23@2x

PR #1130, opened 2019-10-28
refactor(components): use common messages from i18n package

emmenko:
This is a follow up of #1009 
The components now are using the common messages exported by the i18n package, and thus the "old ones" have been removed.

The following PRs from transifex can also be closed, as I already included those messages in this PR:

Closes #1125 
Closes #1127 
Closes #1129


packages/application-components/src/components/dialogs/confirmation-dialog/confirmation-dialog.tsx
```diff
@@ -1,10 +1,10 @@
 import React from 'react';
 import { MessageDescriptor } from 'react-intl';
+import { commonMessages } from '@commercetools-frontend/i18n';
```
tdeekens: Sorry nit: I'd call it `sharedMessages`. Here and everywhere else (readme) as that's what the package is called too.

PR #1136, opened 2019-10-30
fix(app-shell): how metric is read

tdeekens:
#### Summary

This is a strange inconsistency in `perfume.js` API.

The function does not return a promise whenever the value is not supported in a browser.


Zizzamia:
Yeah @tdeekens the all `observe` methods were an overkilling from my side. I later understood the simple way to work with all those async values is to have only one method behaving as a listener, which as probably you learned is `analyticsTracker`.

PR #1138, opened 2019-10-31
New documentation website designs

emmenko:
**Don't bother review the code, I opened a PR just for the deploy links**

TL;DR: we are implementing the new designs according to the new guidelines for writing commercetools documentations.
At the moment there is a gatsby theme folder (which will be moved to another repo) which acts as a dependency.

Furthermore, the appkit.ct.com domain will not be used anymore, but the website will be served within the docs.ct.com domain.


emmenko:
PS: I'm testing the **now** bot to compare the "performance" compared to Netlify

PR #1152, opened 2019-11-07
fix(app-shell): add flags used in menu

tdeekens:
#### Summary

Otherwise these menu items will not show up in custom apps.


tdeekens:
We will solve this differently. Hopefully soon-ish.

PR #1153, opened 2019-11-07
feat(app-shell): to use all feature toggles query

tdeekens:
#### Summary

This pull request builds on a new query in the http proxy `allFeatureToggles`.

#### Description

This query will return all toggls configured in both nav and app menu. This data we can then use to send it down to flopflip.


packages/application-shell/src/hooks/use-all-menu-feature-toggles/use-all-menu-feature-toggles.js
```diff
@@ -0,0 +1,24 @@
+import { useQuery } from 'react-apollo';
+import { useApplicationContext } from '@commercetools-frontend/application-shell-connectors';
+import { FetchAllMenuFeatureToggles } from './fetch-all-menu-feature-toggles.proxy.graphql';
```
emmenko: Don't forget to pull the schema before merging

tdeekens: Yup. Just need it deployed :). Will continue next week.

tdeekens:
@emmenko ready for review. I qualified it as a bug as I am not sure it can be seen as a feature.

tdeekens:
Once approved I will deploy the mc proxy and pull the schemas.

packages/application-shell/src/hooks/use-all-menu-feature-toggles/use-all-menu-feature-toggles.js
emmenko: Did you pull the schema?

tdeekens: CleanShot 2019-11-11 at 12 54 55@2x

packages/application-shell/src/hooks/use-all-menu-feature-toggles/use-all-menu-feature-toggles.spec.js
```diff
@@ -0,0 +1,105 @@
...
+      await waitForElement(() => rendered.getByText('Loading'));
+      await waitForElement(() => [
+        rendered.getByText(/Number of toggles: 2/i),
+        rendered.getByText(/Toggle: flagA/i),
+        rendered.getByText(/Toggle: flagB/i),
+      ]);
+    });
+  });
+
+  describe('with error', () => {
```
emmenko: I think we can remove this test, it’s not really super important. It has also been kind of flaky in the past, not sure if with the more recent versions of apollo it got more stable or not.

tdeekens: Hm, thanks. I'll keep it and see how stable it'll be. The ones in the MC for signup etc include loads of error states and are quite stable.

PR #1157, opened 2019-11-11
fix(app-shell): falling back to default

tdeekens:
#### Summary

The default is the origin. We only allow configuring.


tdeekens:
Will merge to proceed with testing. Hope that's fine. This is inline with other code.

PR #1161, opened 2019-11-13
fix(app-shell): passing of flag config

tdeekens:
We need this config. Not flag as it's now.

CleanShot 2019-11-13 at 09 39 57@2x


amine-benselim:
I am happy to approve this PR but I have no idea why we need this config in that shape specifically 😅

tdeekens:
This is what we have

CleanShot 2019-11-13 at 09 46 44@2x

This leads to LD being configured falsely.

PR #1163, opened 2019-11-13
fix the logic of isAuthorized and Authorized

amine-benselim:
this PR was meant to address a bug we found on Customer dataFences but eventually turned out to be an issue we have with dataFences and permissions in general. 

previously, if in the code we demand dataFences AND general permissions, we checked first for the dataFences, and if the user turned out not to have those permissions, we return early with `isAuthorized = false` WITHOUT checking the general permissions (and action rights).

now, we first check dataFences and THEN we also check for the general permissions if they are met. 

I also fixed the tests that used deprecated params (mainly to silence Typescript 🤷‍♂)


amine-benselim:
`yarn typecheck` does not fail for me locally so I don't know what those TS errors are 🤔

packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -76,39 +76,32 @@ const defaultProps: Pick<Props, 'shouldMatchSomePermissions'> = {
...
+  if (props.demandedDataFences && props.demandedDataFences.length > 0) {
     if (!props.selectDataFenceData) {
       reportErrorToSentry(
         new Error(
           `@commercetools-frontend/permissions/Authorized: Missing data fences selector "selectDataFenceData".`
         )
       );
-      return <React.Fragment>{props.render(false)}</React.Fragment>;
     }
-    return (
```
amine-benselim: it didnt make sense to return here because the user might meet the general permissions without meeting the dataFence permissions

emmenko:
> `yarn typecheck` does not fail for me locally so I don't know what those TS errors are

Hmm you might need to remove your `dist` folders

packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -122,18 +115,21 @@ const Authorized = (props: Props) => {
...
+  hasDemandedActionRights = hasEveryActionRight(
     props.demandedActionRights || [],
     props.actualActionRights
   );
 
   return (
     <React.Fragment>
-      {props.render(hasDemandedPermissions && hasDemandedActionRights)}
+      {props.render(
+        hasDemandedDataFences ||
```
amine-benselim: this is where we fallback to general permissions if dataFence permissions are not met.

packages/permissions/src/components/branch-on-permissions/branch-on-permissions.spec.tsx
```diff
@@ -55,15 +55,13 @@ const renderWithPermissions = (demandedPermissions: string[]) => {
         suspension: {
           isActive: true,
         },
-        allAppliedPermissions: [{ name: 'canViewProducts', value: true }],
-        allAppliedActionRights: [
-          {
-            group: 'products',
-            name: 'canEditPrices',
-            value: false,
+        permissions: { canViewProducts: true },
```
emmenko: Why did you change those? 🤔

emmenko: They need to stay as they were

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -238,18 +244,15 @@ const getDataFenceByPermissionGroup = (
 
 export const hasAppliedDataFence = (options: TOptionsForAppliedDataFence) => {
   // If user has `*Manage` as their `appliedPermission`, we bypass `dataFence`.
-  const hasAppliedManagePermission = options.demandedPermissions.some(
-    demandedPermission => {
-      if (options.actualPermissions) {
```
amine-benselim: I removed the check from here and put it inside `hasExactPermission`

packages/permissions/src/components/branch-on-permissions/branch-on-permissions.spec.tsx
amine-benselim: I thought those were deprecated. I changed the tests to reflect the new params (permissions, dataFences, actionRights). also, typescript was complaining about them inside the test...

emmenko: I'm not sure what you mean with "deprecated" but this is what we get from the graphql query

amine-benselim:
> > `yarn typecheck` does not fail for me locally so I don't know what those TS errors are
> 
> Hmm you might need to remove your `dist` folders

deleting the `dist` folders actually revealed the errors. why is that so? 🤔

packages/permissions/src/components/branch-on-permissions/branch-on-permissions.spec.tsx
amine-benselim:
```
// For backwards compatibility we need to denormalize the given `permissions` option
// (which is now deprecated) to `allAppliedPermissions`, in order to pass the value
// to the `project` prop in the application context provider.
```

I got this backwards 😓

amine-benselim: reverted 🙏

emmenko:
> deleting the `dist` folders actually revealed the errors. why is that so?

I suppose because of the type declarations in the packages that get resolved from the `dist` entry point

packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -122,18 +115,21 @@ const Authorized = (props: Props) => {
       { extra: namesOfNonConfiguredPermissions }
     );
 
-  const hasDemandedPermissions = props.shouldMatchSomePermissions
```
emmenko: This can stay as `const`, right?

packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -122,18 +115,21 @@ const Authorized = (props: Props) => {
...
+  hasDemandedPermissions = props.shouldMatchSomePermissions
     ? hasSomePermissions(props.demandedPermissions, props.actualPermissions)
     : hasEveryPermissions(props.demandedPermissions, props.actualPermissions);
 
-  const hasDemandedActionRights = hasEveryActionRight(
```
emmenko: Same here

packages/permissions/src/components/authorized/authorized.spec.tsx
```diff
@@ -283,8 +283,8 @@ describe('rendering', () => {
             });
             shallow(<Authorized {...props} />);
           });
-          it('should indicate as not authorized', () => {
-            expect(props.render).toHaveBeenCalledWith(false);
+          it('should indicate as authorized', () => {
+            expect(props.render).toHaveBeenCalledWith(true);
```
emmenko: Just double checking: is this the expected outcome now?

amine-benselim: this test case is exactly what was wrong about our logic, 
if you look at the actual and demanded permissions: 
```
demandedPermissions: ['ViewOrders'],
actualPermissions: {	        
  canViewOrders: true,	           
  canManageOrders: false,	               
},
```
the user IS authorized even if he did not have a manage orders DataFence, because in the code we also demanded the general permission `ViewOrders` and the user meets this permission.

packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -76,64 +75,47 @@ const defaultProps: Pick<Props, 'shouldMatchSomePermissions'> = {
...
+      actualPermissions: props.actualPermissions,
+      demandedPermissions: props.demandedPermissions,
+    });
   }
 
-  // If no data fences have been provided, fall back to normal permissions + action rights.
-  if (!props.actualPermissions)
-    return <React.Fragment>{props.render(false)}</React.Fragment>;
-
-  const namesOfNonConfiguredPermissions = getInvalidPermissions(
```
emmenko: Can you explain again why we don't need this anymore?

packages/permissions/src/utils/has-permissions.spec.tsx
```diff
@@ -185,23 +184,3 @@ describe('hasEveryActionRight', () => {
     ).toBe(false);
   });
 });
-
-describe('getInvalidPermissions', () => {
```
tdeekens: This feels like an unrelated change.

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -60,10 +59,10 @@ type TActualDataFence = {
 
 type TOptionsForAppliedDataFence = {
   demandedDataFences: TDemandedDataFence[];
-  actualDataFences: TDataFences;
-  selectDataFenceData: TSelectDataFenceData;
+  actualDataFences: TDataFences | null;
+  selectDataFenceData?: TSelectDataFenceData;
   actualPermissions: TPermissions | null;
-  demandedPermissions: TPermissionName[];
+  demandedPermissions: TPermissionName[] | null;
```
tdeekens: Why did the type signaure have to change?

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -175,27 +175,13 @@ export const hasSomePermissions = (
...
-  );
-};
-
 const hasDemandedDataFence = (options: {
   actualDataFence: TActualDataFence;
   demandedDataFence: TDemandedDataFence;
-  selectDataFenceData: TSelectDataFenceData;
+  selectDataFenceData?: TSelectDataFenceData;
 }): boolean => {
+  if (!options.selectDataFenceData) return false;
```
tdeekens: Would you log in this case too?

packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -76,64 +75,47 @@ const defaultProps: Pick<Props, 'shouldMatchSomePermissions'> = {
...
-  if (namesOfNonConfiguredPermissions.length > 0)
-    reportErrorToSentry(
-      new Error(
-        `@commercetools-frontend/permissions/Authorized: Invalid prop "demandedPermissions" supplied. The permission(s) ${namesOfNonConfiguredPermissions.toString()} is/are not configured through "actualPermissions".`
-      ),
-      { extra: namesOfNonConfiguredPermissions }
-    );
-
-  const hasDemandedPermissions = props.shouldMatchSomePermissions
+  hasDemandedPermissions = props.shouldMatchSomePermissions
```
tdeekens: This feels wrong. Why initialise it to a boolean argument?

packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -76,64 +75,47 @@ const defaultProps: Pick<Props, 'shouldMatchSomePermissions'> = {
...
 
 const Authorized = (props: Props) => {
-  // Check first for data fences
-  if (
-    props.actualDataFences &&
-    props.demandedDataFences &&
-    props.demandedDataFences.length > 0
-  ) {
+  // if the user has no permissions and no dataFences assigned to them, they are not authorized
+  if (!props.actualDataFences && !props.actualPermissions) {
```
adnasa: this case won't happen since `actualDataFences` and `actualPermissions` won't be `undefined`, right?

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.ts
```diff
@@ -109,7 +110,9 @@ const useIsAuthorized = ({
     actualActionRights
   );
 
-  return hasDemandedPermissions && hasDemandedActionRights;
+  return (
+    hasDemandeDataFences || (hasDemandedPermissions && hasDemandedActionRights)
```
adnasa: okay, I see that we lift up the fallback

packages/permissions/src/components/authorized/authorized.spec.tsx
amine-benselim: yes

amine-benselim:
I admit I have went too far with the changes, I will revert the change not related to the bug we're facing and then I will open another PR to explore simplifying the code.

packages/permissions/src/components/authorized/authorized.tsx
amine-benselim: actualDataFences can be, and if we look at the old code we where doing this: 
```
if (!props.actualPermissions)	
    return <React.Fragment>{props.render(false)}</React.Fragment>;
```

so also actual permissions can be `null`

packages/permissions/src/components/authorized/authorized.tsx
amine-benselim: I can explain this, but I will do so in another PR 👍

packages/permissions/src/components/authorized/authorized.tsx
amine-benselim: I don't understand, what do you mean exactly? 🤔

packages/permissions/src/utils/has-permissions.ts
amine-benselim: we log already in the Component level (Authorized and useIsAuthorized).

packages/permissions/src/components/restricted-by-permissions/restricted-by-permissions.tsx
```diff
@@ -55,7 +55,7 @@ const RestrictedByPermissions = (props: Props) => {
       render={applicationContext => (
         <Authorized
           shouldMatchSomePermissions={props.shouldMatchSomePermissions}
-          actualPermissions={applicationContext.permissions}
+          actualPermissions={applicationContext.permissions || {}}
```
tdeekens: Can you explain why this needs to change?

amine-benselim: I reverted this change 👍

packages/permissions/src/components/authorized/authorized.tsx
adnasa: 👍 thanks

tdeekens:
Sorry, I fail to follow at all. What is "state of inputs" or "on state" mean here?

adnasa:
state of input: whether a field should be disabled or not.

tdeekens:
@adnasa thanks for clarifying. I think however we need to separate two classes of problems:

1. We architecture doesn't allow us to achieve what we have to (component, hook, render props)
2. This PR doesn't fullfil all cases of UATs

What you're saying sounds like 1. but I think it's more like 2. So we just need to iterate on this and cover all cases and have a test each.

amine-benselim:
I have allowed myself to refactor the permission package to have `Authorized` use `useIsAuthorized` under the hook as the only source of logic. this makes debugging and trying to fix issues simpler. 

I have also moved the tests from `Authorized` to `useIsAuthorized` and they all passed 🎉 

regarding the logic itself, I think the solution we have so far is sound and sufficient. I will discuss this with Adnan in person and will add more test cases for the scenarios we want to cover.

amine-benselim:
@emmenko if you take a look at the Typescript error, you see that it's complaining about `{ __typename: 'StoreDataFence' }` 

the reason for that is because when taking a look at `mc.ts` you notice that this typename is the only one that is not optional (without `?`). I know that the types are generated (althought I don't know where from exactly), so I have no idea why this typename is an exception

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -237,19 +240,6 @@ const getDataFenceByPermissionGroup = (
 };
 
 export const hasAppliedDataFence = (options: TOptionsForAppliedDataFence) => {
-  // If user has `*Manage` as their `appliedPermission`, we bypass `dataFence`.
-  const hasAppliedManagePermission = options.demandedPermissions.some(
```
amine-benselim: because we fall back now to general permissions if dataFence permissions are not met, we do not need to check if the use has `manage` general permission first INSIDE `hasAppliedDataFence`, this simplifies the logic further and makes the flow for permission check straight forward: 

1- check if the user meets the dataFence permissions
2- if not, check if he meets the general permissions + action rights

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
```diff
@@ -18,7 +18,6 @@ type TAllAppliedMenuVisibility = {
   value: boolean;
 };
 type TAllAppliedDataFence = {
-  __typename: 'StoreDataFence';
```
emmenko: Can you explain why this is removed? 🤔

amine-benselim: because that 'fixes' the Typescript error I told you about in the previous comment, but does not address the original issue which is that only `__typename: 'StoreDataFence'` is not optional in `mc.ts`

emmenko: But what was the original error?

amine-benselim: there was no original error. when I was adding more test cases I had to add `__typename: 'StoreDataFence'` to all of my applied dataFences which I think is not right.

emmenko: You mean in the tests? I would define it where is needed to be honest, if that's how the types are. The types are generated from the graphql queries -> schema.

amine-benselim: done that 👍

emmenko: Thanks. Btw, if you want to be less repetitive in tests, you can simply have a `createTestStore` function or something which has some defaults values 😉

packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -161,24 +73,16 @@ const injectAuthorized = <
   Component: React.ComponentType<OwnProps>
 ): React.ComponentType<OwnProps & InjectedProps> => {
   const WrappedComponent = (props: OwnProps) => (
-    <ApplicationContext
```
adnasa: nice

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
```diff
@@ -187,4 +208,271 @@ describe('when only one permission matches', () => {
...
+            demandedActionRights: [
+              { group: 'products', name: 'PublishProducts' },
+            ],
+            shouldMatchSomePermissions: true,
+          });
+
+          expect(queryByText('Is authorized: No')).toBeInTheDocument();
+        });
+      });
+      describe('if can edit prices on products group', () => {
```
adnasa: This is a good test.
This proves that it is on a consumer side to be careful when demanding permissions.

because one can expect `isAuthorized=true` if `actualPermissions.ViewProducts=true` and `actualActionRights.canEditPrices=true`, then you get `isAuthorized=true` if you demand this.

However if this happens, we may have a flaw in how we save action rights.

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
```diff
@@ -187,4 +208,271 @@ describe('when only one permission matches', () => {
...
+          expect(queryByText('Is authorized: No')).toBeInTheDocument();
+        });
+      });
+    });
+  });
+
+  describe('dataFences', () => {
+    describe('with general permission', () => {
+      describe('when general permissions is manage', () => {
+        it('should indicate as authorized', () => {
```
adnasa: should add another describe `when demanded general permissions is...`

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
```diff
@@ -187,4 +208,271 @@ describe('when only one permission matches', () => {
...
+
+          expect(queryByText('Is authorized: Yes')).toBeInTheDocument();
+        });
+      });
+      describe('when general permissions is view', () => {
+        describe('when data fence permission is manage', () => {
+          it('should indicate as authorized', () => {
+            const { queryByText } = testRender({
+              shouldMatchSomePermissions: true,
+              allAppliedPermissions: [{ name: 'canManageOrders', value: true }],
```
adnasa: but the describe says `when general permissions is view`

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
amine-benselim: I already had that 👍 I could add more abstraction but the test cases require me to change the demanded and actual dataFences and permissions.

amine-benselim:
> it is clearer now with the test.
> We could go write it to be a bit more verbose and make the language clearer
> 
> example
> 
> ```js
> describe('with general permissions view', () => {
>   describe('with data fence permissions view', () => {
>     describe('when demanding data fence manage', () => {
>       it('should...', () => {
>         
>       })
>     })
>   })
>   describe('with data fence permissions manage', () => {
>     describe('when demanding data fence manage', () => {
>       it('should...', () => {
> 
>       })
>     })
>   })
> })
> ```

thanks @adnasa for the suggestion, I already rewrote the tests before reading your comments, can you check them now if they make sense or should I change them even further 🤔

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
amine-benselim: this is refactored now.

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
```diff
@@ -130,86 +130,90 @@ const testRender = ({
...
-            group: 'products',
-            name: 'canEditPrices',
-            value: true,
-          },
-        ],
-        demandedPermissions: ['ManageCustomers'],
-        demandedActionRights: [{ group: 'products', name: 'PublishProducts' }],
-        shouldMatchSomePermissions: false,
-      });
+    describe('when it should not match some permission', () => {
```
adnasa: I don't get this part.. does this refer to the outcome or criteria?
if it is the criteria, then `when shouldMatchSomePermissions=false`

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
```diff
@@ -436,11 +469,10 @@ describe('when only one permission matches', () => {
           expect(queryByText('Is authorized: Yes')).toBeInTheDocument();
         });
       });
-
-      describe('has demanded dataFence to VIEW or MANAGE orders on the specific store', () => {
-        it('should indicate as authorized', () => {
+      describe('when dataFence permission is not met', () => {
```
adnasa: I don't know. I prefer the language of `when general permissions...`, `when datafence permissions` vs `when demanded datafence permissions ...`

amine-benselim: I mean there is a difference between the demanded and the actual and we need to make that difference clear in our test cases.

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
```diff
@@ -103,88 +122,431 @@ const testRender = ({
...
-        demandedPermissions: ['ManageCustomers', 'ManageOrders'],
-        allAppliedPermissions: [
-          {
-            name: 'canManageCustomers',
-            value: true,
-          },
-        ],
-        shouldMatchSomePermissions: true,
+  describe('General Permissions', () => {
+    describe('when "shouldMatchSomePermissions=true"', () => {
```
tdeekens: Nit: I wouldn't even bind the name of the spec the prop name. It's just "describe(`when all permissions should be matched`"

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
amine-benselim: the problem is that it does not make sense to talk about general permissions being view or manage anymore, the dataFence permissions are agnostic to what the general permission are, so in the context of data fences we talk only about wether the user meets the general permissions or not. 

overall, I think the test cases and their structure is good enough for now and would like to get this PR through.

packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -76,66 +39,15 @@ const defaultProps: Pick<Props, 'shouldMatchSomePermissions'> = {
...
-    props.demandedActionRights || [],
-    props.actualActionRights
-  );
-
-  return (
-    <React.Fragment>
-      {props.render(hasDemandedPermissions && hasDemandedActionRights)}
-    </React.Fragment>
-  );
+  return <React.Fragment>{props.render(isAuthorized)}</React.Fragment>;
```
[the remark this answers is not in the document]
emmenko: Yes otherwise TS cannot recognize it as a react component

adnasa:
thanks for this, @amine-benselim 😉

packages/permissions/src/components/authorized/authorized.tsx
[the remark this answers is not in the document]
amine-benselim: I think we don't, I don't know why it was the case before and I just did not change it 
I will remove it now.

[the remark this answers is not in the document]
amine-benselim: maybe to allow the render prop to return more than one ReactNode? 🤔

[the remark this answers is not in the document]
amine-benselim: Typescript complained a lot when I deleted it, so I guess I have to keep it 😅

PR #1165, opened 2019-11-14
fix: cannot use rerender with app-kit custom render functions

ahmehri:
#### Summary

Currently, it's not possible to use [rerender](https://testing-library.com/docs/react-testing-library/api#rerender). This PR fixes the issue.

#### Description

When rendering using one of app-kit custom render functions, `renderApp`, `renderAppWithRedux` or `experimentalRenderAppWithRedux`, using `rerender` on the `rendered` result won't work and will result in errors.

**Why using rerender?**

Even though it's considered to be better to test the component updating the props instead of using `rerender`, we still need to be able to use it without errors when we want to.

> It'd probably be better if you test the component that's doing the prop updating to ensure that the props are being updated correctly (see the Guiding Principles section). That said, if you'd prefer to update the props of a rendered component in your test, this function can be used to update props of the rendered component.

In addition, there are some use cases where using `rerender` is very useful and not tightly related to updating props. Because of the current limitation, testing in those use cases is done by separating what would have been tested using `rerender` in another test instead. Which means that we will have tests as the following

```js
it('scenario 1', () => {
  // test scenario 1
});

it('scenario 2', () => {
  // test scenario 2
})
```

instead of a single test without the boilerplate

```js
it('test all scenarios in a single test', () => {
  // test scenario 1
  const rendered = renderApp(...);

  // test scenario 2
  rendered.rerender(...);
})
```

Being able to use `rerender` will allow us to avoid the first pattern and simplify tests.

**The fix**

The fix is simple, RTL offers, out of the box, wrapping the provided UI into a wrapper, if provided using `wrapper` render option. I adapted app-kit custom render functions to use this feature.

**Requirements**

In order to be able to use this fix, a custom render function using one of the app-kit custom render functions (`renderApp`, `renderAppWithRedux` and `experimentalRenderAppWithRedux`) should from now on use the `wrapper` render option to wrap the provided UI instead of manually wrapping it. This means that the custom render function API should change from

```js
const render = (ui, renderOptions) =>
  renderApp(
    <Provider1>
      <Provider2>
        <Provider3>{ui}</Provider3>
      </Provider2>
    </Provider1>,
    renderOptions
  );
```

to

```js
const Wrapper = ({ children }) => (
  <Provider1>
    <Provider2>
      <Provider3>{children}</Provider3>
    </Provider2>
  </Provider1>
);
Wrapper.propTypes = {
  children: PropTypes.node.isRequired
};

const render = (ui, renderOptions) =>
  renderApp(ui, {
    ...renderOptions
    wrapper: Wrapper,
  });
```

**Limitations**

The fix doesn't work if app-kit custom render function was used with render options other than `wrapper`. I didn't spend so much time trying to fix it as I think that being able to use `rerender` for most of the use cases still a big win. And maybe we can fix the issue in a future separate PR.

### DIFF
--- a/packages/application-shell/src/test-utils/test-utils.js
+++ b/packages/application-shell/src/test-utils/test-utils.js
@@ -284,7 +284,7 @@ const renderApp = (
   const mergedEnvironment = mergeOptional(defaultEnvironment, environment);
   const mergedGtmTracking = mergeOptional(defaultGtmTracking, gtmTracking);
 
-  const Wrapper = ({ children }) => (
+  const ApplicationProviders = ({ children }) => (
     <IntlProvider locale={locale}>
       <ApolloProviderComponent mocks={mocks} addTypename={addTypename}>
         <ConfigureFlopFlip
@@ -317,14 +317,17 @@ const renderApp = (
       </ApolloProviderComponent>
     </IntlProvider>
   );
-  Wrapper.propTypes = {
+  ApplicationProviders.propTypes = {
     children: PropTypes.node.isRequired,
   };
 
   const rendered = rtl.render(ui, {
     ...renderOptions,
     wrapper: ({ children }) =>
-      wrapIfNeeded(Wrapper, wrapIfNeeded(renderOptions.wrapper, children)),
+      wrapIfNeeded(
+        ApplicationProviders,
+        wrapIfNeeded(renderOptions.wrapper, children)
+      ),
   });
 
   return {
@@ -415,7 +418,7 @@ const renderAppWithRedux = (
     return createReduxStore(storeState, [testingMiddleware]);
   })();
 
-  const Wrapper = ({ children }) => (
+  const ReduxProviders = ({ children }) => (
     <NotificationProviderForCustomComponent
       mapNotificationToComponent={mapNotificationToComponent}
     >
@@ -429,14 +432,17 @@ const renderAppWithRedux = (
       </StoreProvider>
     </NotificationProviderForCustomComponent>
   );
-  Wrapper.propTypes = {
+  ReduxProviders.propTypes = {
     children: PropTypes.node.isRequired,
   };
 
   const rendered = renderApp(ui, {
     ...renderOptions,
     wrapper: ({ children }) =>
-      wrapIfNeeded(Wrapper, wrapIfNeeded(renderOptions.wrapper, children)),
+      wrapIfNeeded(
+        ReduxProviders,
+        wrapIfNeeded(renderOptions.wrapper, children)
+      ),
   });
 
   return {
--- a/packages/application-shell/src/test-utils/test-utils.js
+++ b/packages/application-shell/src/test-utils/test-utils.js
@@ -234,6 +234,9 @@ const denormalizeDataFences = dataFences => {
   );
 };
 
+const wrapIfNeeded = (wrapper, children) =>
+  wrapper ? React.createElement(wrapper, null, children) : children;
+
 // This function renders any component within the application context, as if it
 // was rendered inside <ApplicationShell />.
 // The context is not completely set up yet, some things are missing:
@@ -259,7 +262,7 @@ const renderApp = (
       createMemoryHistory({ initialEntries: [route] })
     ),
     // flopflip
-    adpater = memoryAdapter,
+    adapter = memoryAdapter,
     flags = {},
     // application-context
     environment,
@@ -280,41 +283,53 @@ const renderApp = (
   const mergedProject = mergeOptional(defaultProject, project);
   const mergedEnvironment = mergeOptional(defaultEnvironment, environment);
   const mergedGtmTracking = mergeOptional(defaultGtmTracking, gtmTracking);
-  return {
-    ...rtl.render(
-      <IntlProvider locale={locale}>
-        <ApolloProviderComponent mocks={mocks} addTypename={addTypename}>
-          <ConfigureFlopFlip
-            adapter={adpater}
-            defaultFlags={flags}
-            adapterArgs={defaultFlopflipAdapterArgs}
-          >
-            <ApplicationContextProvider
-              user={mergedUser}
-              project={
-                mergedProject && {
-                  ...mergedProject,
-                  allAppliedPermissions: denormalizePermissions(permissions),
-                  allAppliedActionRights: denormalizeActionRights(actionRights),
-                  allAppliedDataFences: denormalizeDataFences(dataFences),
-                }
+
+  const Wrapper = ({ children }) => (
+    <IntlProvider locale={locale}>
+      <ApolloProviderComponent mocks={mocks} addTypename={addTypename}>
+        <ConfigureFlopFlip
+          adapter={adapter}
+          defaultFlags={flags}
+          adapterArgs={defaultFlopflipAdapterArgs}
+        >
+          <ApplicationContextProvider
+            user={mergedUser}
+            project={
+              mergedProject && {
+                ...mergedProject,
+                allAppliedPermissions: denormalizePermissions(permissions),
+                allAppliedActionRights: denormalizeActionRights(actionRights),
+                allAppliedDataFences: denormalizeDataFences(dataFences),
               }
-              environment={mergedEnvironment}
-              projectDataLocale={dataLocale}
-            >
-              <GtmContext.Provider value={mergedGtmTracking}>
-                <Router history={history}>
-                  <React.Suspense fallback={<LoadingFallback />}>
-                    {ui}
-                  </React.Suspense>
-                </Router>
-              </GtmContext.Provider>
-            </ApplicationContextProvider>
-          </ConfigureFlopFlip>
-        </ApolloProviderComponent>
-      </IntlProvider>,
-      renderOptions
-    ),
+            }
+            environment={mergedEnvironment}
+            projectDataLocale={dataLocale}
+          >
+            <GtmContext.Provider value={mergedGtmTracking}>
+              <Router history={history}>
+                <React.Suspense fallback={<LoadingFallback />}>
+                  {children}
+                </React.Suspense>
+              </Router>
+            </GtmContext.Provider>
+          </ApplicationContextProvider>
+        </ConfigureFlopFlip>
+      </ApolloProviderComponent>
+    </IntlProvider>
+  );
+  Wrapper.propTypes = {
+    children: PropTypes.node.isRequired,
+  };
+
+  const rendered = rtl.render(ui, {
+    ...renderOptions,
+    wrapper: ({ children }) =>
+      wrapIfNeeded(Wrapper, wrapIfNeeded(renderOptions.wrapper, children)),
+  });
+
+  return {
+    ...rendered,
+
     // adding `history` to the returned utilities to allow us
     // to reference it in our tests (just try to avoid using
     // this to test implementation details).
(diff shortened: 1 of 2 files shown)


tdeekens:
Thanks first of all for the length description and explanation. You clearly invested some time and effort into this.

I've now read the react-testing-library source code again, re-visited how I made use of the `wrapper` and `wrapIfNeeded` on flopflip and took some time to reflect on this.

First I'd like to reflect on `rerender` itself. Rtl itself somehwat discourages the use of `rerender` as it only needs to be used if props have changed where the same should be achievable via rendering again even though this produces more boilerplate when writing tests. `rerender` itself has the signature of `rerender( rerenderUi: UI)` which means you pass in what you want to re-render while it recyles only the `baseElement` and `container`. So essentially the actual DOM underneath. It however recycles the previously passed in `wrapper`.
So there is `wrapper` which does as the name suggests. The common use case is e.g. when setting up Redux or other providers around the component under test. I actually do this with a `components`-prop in flopflip [here](https://github.com/tdeekens/flopflip/blob/master/packages/test-utils/index.js#L72) where one can pass in a `Wrapper` and `ConfigureFlopFlip` which makes tests quite lean as we can see [here](https://github.com/tdeekens/flopflip/blob/master/packages/react-redux/modules/components/inject-feature-toggle/inject-feature-toggle.spec.js#L13).

This brings me to these first points:

1. To clarify _for others and me_ `wrapper` is being closed over in `render` by rtls internal `wrapUiIfNeeded`. This means that when
   - Calling `render` `wrapUiIfNeeded` closes over it and "remembers" it for `rerender`
   - When calling `rerender` without `wrapper` (from `render` one looses context and can not render our providers again (as they're internal to app-kit). This renders (pun intended) `rerender` useless through our `test-utils`
2. Is it only a convienence your're seeking? So calling `rerender` instead of e.g. abstracting into a `renderWithFoo` in a specific spec which then would entail the only difference in a spec being:

```js
let rendered;
it('...', () => {
  rendered = renderWithFoo();
})
it('...', () => {
  rendered.rerender();
})
```

and

```js
it('...', () => {
  const rendered = renderWithFoo();
})
it('...', () => {
  const rendered = renderWithFoo();
})
```

This brings me to my twisted plot in mind brain 😸. I think fixing up `rerender` by passing our providers as `wrapper` is absolutely correct and wonderful. However, aiming to make stark use of `rerender` make me feel uncertain. We should surely export a functioning wrapping of rtl. However, I fail to convince myself of a good use case of `rerender` in general and need some convincing 😄.

packages/application-shell/src/test-utils/test-utils.js
```diff
@@ -280,41 +283,53 @@ const renderApp = (
...
-              user={mergedUser}
-              project={
-                mergedProject && {
-                  ...mergedProject,
-                  allAppliedPermissions: denormalizePermissions(permissions),
-                  allAppliedActionRights: denormalizeActionRights(actionRights),
-                  allAppliedDataFences: denormalizeDataFences(dataFences),
-                }
+
+  const Wrapper = ({ children }) => (
```
tdeekens: Nit: maybe call it `ApplicationProviders` otherwise this becomes fairly nifty in the bottom as it's "wrapper the wrapper if needed with the wrapper's wrapper" and you go 😑.

packages/application-shell/src/test-utils/test-utils.spec.js
```diff
@@ -264,3 +274,117 @@ describe('router', () => {
     expect(history.location.pathname).toBe('/foo');
   });
 });
+
+describe('custom render function', () => {
```
tdeekens:
```suggestion
describe('custom render functions', () => {
```

packages/application-shell/src/test-utils/test-utils.spec.js
```diff
@@ -264,3 +274,117 @@ describe('router', () => {
...
+  describe('wrapper', () => {
+    describe('when provided', () => {
```
tdeekens: Nit: I would collapse these describes into `with wrapper` and `without wrapper`.

packages/application-shell/src/test-utils/test-utils.spec.js
```diff
@@ -264,3 +274,117 @@ describe('router', () => {
...
+      const Context = React.createContext();
+      const ProvidedWrapper = ({ children }) => (
+        <Context.Provider value="provided wrapper">{children}</Context.Provider>
+      );
+      ProvidedWrapper.propTypes = {
+        children: PropTypes.node.isRequired,
+      };
+
+      it('should renderApp merges the provided wrapper with its own wrapper', () => {
```
tdeekens: This sentence seems broked 😅. 

```suggestion
      it('should merge the passed wrapper with renderApp\'s internal wrapper', () => {
```

packages/application-shell/src/test-utils/test-utils.spec.js
```diff
@@ -264,3 +274,117 @@ describe('router', () => {
...
+          return value;
+        };
+
+        const rendered = renderApp(<TestComponent number={1} />, {
+          wrapper: ProvidedWrapper,
+        });
+        rendered.getByText(/provided wrapper/i);
+      });
+
+      it('should renderAppWithRedux merges the provided wrapper with its own wrapper', () => {
```
tdeekens: ˆˆ

packages/application-shell/src/test-utils/test-utils.spec.js
```diff
@@ -264,3 +274,117 @@ describe('router', () => {
...
+        rendered.getByText(/provided wrapper/i);
+      });
+    });
+
+    describe('when not provided', () => {
+      it.each([renderApp, renderAppWithRedux])(
+        'should %p still manage',
+        renderFn => {
+          const TestComponent = ({ number }) => {
+            return number;
```
tdeekens: Nit: return is uncalled-for can just be `=> number`.

packages/application-shell/src/test-utils/test-utils.spec.js
```diff
@@ -264,3 +274,117 @@ describe('router', () => {
...
+        const rendered = renderAppWithRedux(<TestComponent number={1} />, {
+          wrapper: ProvidedWrapper,
+        });
+        rendered.getByText(/provided wrapper/i);
+      });
+    });
+
+    describe('when not provided', () => {
+      it.each([renderApp, renderAppWithRedux])(
+        'should %p still manage',
```
tdeekens: I fail to grasp the description here. Maybe it's due to the structure or description. Wondering if a structure as in

```txt
describe: when rendering with redux
   describe: with wrapper
       it: should render and merge the passed and internal wrapper
       it: should allow rerendering the passed component
   describe: without wrapper
       it: should render and internal wrapper
       it: should allow rerendering the passed component
   describe: with experimental rendering
       it: should render and internal wrapper
       it: should allow rerendering the passed component
describe with rendering without redux
   describe: with wrapper
       it: should render and merge the passed and internal wrapper
       it: should allow rerendering the passed component
   describe: without wrapper
       it: should render and internal wrapper
       it: should allow rerendering the passed component
```

packages/application-shell/src/test-utils/test-utils.spec.js
```diff
@@ -264,3 +274,117 @@ describe('router', () => {
...
+
+      const rendered = renderApp(<TestComponent number={1} />);
+      rendered.getByText(/1/);
+
+      rendered.rerender(<TestComponent number={2} />);
+      rendered.getByText(/2/);
+      expect(rendered.queryByText(/1/)).not.toBeInTheDocument();
+    });
+
+    it('should works with renderAppWithRedux', () => {
```
tdeekens: Grammar seems off here.

ahmehri:
> rerender itself has the signature of rerender( rerenderUi: UI) which means you pass in what you want to re-render while it recyles only the baseElement and container. So essentially the actual DOM underneath. It however recycles the previously passed in wrapper.

> When calling rerender without wrapper (from render one looses context and can not render our providers again (as they're internal to app-kit). This renders (pun intended) rerender useless through our test-utils

I don't think this is true, even though the `rerender` accepts only the `rerenderUi` as argument and not the wrapper, it will still remember the wrapper (passed in the first call of `render`) because `wrapUiIfNeeded` closes over it and remembers it for all consecutive `rerender` [calls](https://github.com/testing-library/react-testing-library/blob/11460a3b31d3826d59a8920a25ba2e3fb7e17495/src/pure.js#L47-L71).

```js
  const wrapUiIfNeeded = innerElement =>
    WrapperComponent
      ? React.createElement(WrapperComponent, null, innerElement)
      : innerElement

  act(() => {
    if (hydrate) {
      ReactDOM.hydrate(wrapUiIfNeeded(ui), container)
    } else {
      ReactDOM.render(wrapUiIfNeeded(ui), container)
    }
  })

  return {
    container,
    baseElement,
    debug: (el = baseElement) =>
      Array.isArray(el)
        ? // eslint-disable-next-line no-console
          el.forEach(e => console.log(prettyDOM(e)))
        : // eslint-disable-next-line no-console,
          console.log(prettyDOM(el)),
    unmount: () => ReactDOM.unmountComponentAtNode(container),
    rerender: rerenderUi => {
      render(wrapUiIfNeeded(rerenderUi), {container, baseElement})
```

The fact that it's still be possible to use the app-kit providers when calling `rerender` is what this whole PR is based on. Otherwise, the application custom render function using the app-kit custom render function won't be able to work when using `rerender`. I tested the result with my own custom render function (consuming app-kit render functions) and it went very well.

> However, I fail to convince myself of a good use case of rerender in general and need some convincing 😄.

I had very useful use cases of `rerender` but I don't remember them exactly to share them. One thing for sure is that I will let you know when I will encounter the next use case.

tdeekens:
> I don't think this is true, even though the rerender accepts only the rerenderUi as argument and not the wrapper, it will still remember the wrapper (passed in the first call of render) because wrapUiIfNeeded closes over it and remembers it for all consecutive rerender calls.

Sorry yes, I agree and follow. There is a mix-upn in my message cause I drafted it last nite and went over it this morning. As `rerender` closes over the wrapper it doesn't have to be passed again. That's the idea.

packages/application-shell/src/test-utils/test-utils.spec.js
ahmehri: The whole test will have as title 

*rerender should work with renderAppWithRedux*

I will remove the `s` in `works` which was a typo but other than that I think the grammar is correct. I'm open to any suggestions.

packages/application-shell/src/test-utils/test-utils.spec.js
ahmehri: After implementing [this suggestion](https://github.com/commercetools/merchant-center-application-kit/pull/1165#discussion_r346733517), the structure becomes

```
  custom render functions
    with wrapper
      ✓ should merge the passed wrapper with renderApp internal wrapper (6ms)
      ✓ should merge the passed wrapper with renderAppWithRedux internal wrapper (11ms)
    without wrapper
      ✓ should [Function renderApp] still manage (7ms)
      ✓ should [Function renderAppWithRedux] still manage (6ms)
    rerender
      ✓ should work with renderApp (5ms)
      ✓ should work with renderAppWithRedux (10ms)
      ✓ should work with experimentalRenderAppWithRedux (9ms)
```

*still manage* is not clear enough and I will replace it by *still work*. This test is meant to ensure that app-kit custom render functions will still work in case a `wrapper` wasn't provided.

tdeekens: Looks way better. 

1. [Function renderApp] looks odd obvously 

Super nitty now and sorry about it

```txt
describe: rendering
  it: should be supported by renderApp
  ...
```

ahmehri: No worries. The thing with `[Function renderApp]` is that it was the best I can do when using `it.each`.

https://github.com/commercetools/merchant-center-application-kit/blob/02eccdbe6808552508d0213bb18499a72d6cb42d/packages/application-shell/src/test-utils/test-utils.spec.js#L322-L330

I needed to pass the actual functions in the array and not as strings to be able to use them here.

https://github.com/commercetools/merchant-center-application-kit/blob/02eccdbe6808552508d0213bb18499a72d6cb42d/packages/application-shell/src/test-utils/test-utils.spec.js#L327

`%p` is the best formatting that I found. In the end, I preferred having `[Function renderApp]`with a single test and avoid duplication rather than write two explicit tests. It's true still that it looks odd but I find it readable at the same time.

> should be supported by renderApp

doesn't reflect clearly that we're guarding against issues when rendering without providing a wrapper. The reason why I added this test is because of the `wrapIfNeeded` function.

https://github.com/commercetools/merchant-center-application-kit/blob/02eccdbe6808552508d0213bb18499a72d6cb42d/packages/application-shell/src/test-utils/test-utils.js#L237-L238

In the beginning, I implemented it without guarding against the `wrapper` being undefined. This broke the rendering logic and none of the tests available back then caught it.

tdeekens: Ah got it. Thanks for the explanation!

tdeekens:
@ahmehri I assigned the bug label. This is used during releasing to generate a changelog :)

PR #1167, opened 2019-11-15
docs(changelog): for release v15.3.7 [skip ci]

emmenko:


amine-benselim:
please do not update MC frontend with this version yet. I think the permission HOCs are not working as expected and I am investigating the issue. 

sorry I was in meetings and did not see this PR.

PR #1173, opened 2019-11-18
refactor: only some subset of demanded DF values should belong to actual DF values

amine-benselim:
As we are integrating dataFences more in the merchant center, we have encountered so far three use case: 
1- on the route level: check if the user has `manage` or `view` permissions in order to render the route. in this case, the demanded dataFence values are all always the `actualDataFenceValues`.
```
      getSelectDataFenceData: () => ({ actualDataFenceValues }) =>
        actualDataFenceValues || [],
```

2- for orders: we check if the `order.store.key` is inside the `acualDataFenceValues`:
```
const createSelectOrderDataFenceData = componentProps => demandedDataFence => {
  switch (demandedDataFence.type) {
    case 'store':
      return componentProps.order.store?.key
        ? [componentProps.order.store.key]
        : [];
    default:
      return null;
  }
};
```

3- for customers. we also check if the `customer.stores` are inside `actualDataFenceValues` 
```
const createSelectCustomerDataFenceData = componentProps => demandedDataFence => {
  switch (demandedDataFence.type) {
    case 'store':
      return (componentProps.customer?.stores ?? []).map(
        customerStore => customerStore.key
      );
    default:
      return null;
  }
};
```

For cases 1 and 2, the old code works because we are either checking if an array belongs to itself (which would always be true), or if one value belongs to an array (because an order can only belong to one store).

For case 3, however, the issue is that a customer might belong to multiple stores, but the user has the permission to manage only a subset of those stores, in which case we still want him to manage that customer, but our code does not allow him to (`isAuthorized` is `false` in this scenario`)

## the fix
The fix, for now, is to use `some` instead of `every` for checking the demanded dataFence values against the actual ones. 

If in the future we encounter a use case where we want to have a strict matching requirement, we might introduce a flag like `shouldMatchSomePermissions` but specifically for DataFence values.

We have not faced such a use case yet so I refrain from adding complexity to the code now.

### DIFF
--- a/packages/permissions/src/utils/has-permissions.spec.tsx
+++ b/packages/permissions/src/utils/has-permissions.spec.tsx
@@ -5,6 +5,7 @@ import {
   hasEveryPermissions,
   hasSomePermissions,
   getInvalidPermissions,
+  hasAppliedDataFence,
 } from './has-permissions';
 
 type TPermissionName = string;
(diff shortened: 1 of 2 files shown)


packages/permissions/src/utils/has-permissions.ts
```diff
@@ -218,7 +218,8 @@ const hasDemandedDataFence = (options: {
     return false;
   }
 
-  return selectedDataFenceData.every(value =>
+  // it is enough to only have a subset of demanded dataFence data belonging to actual dataFence values
+  return selectedDataFenceData.some(value =>
```
adnasa: thank you

packages/permissions/src/utils/has-permissions.spec.tsx
```diff
@@ -205,3 +206,178 @@ describe('getInvalidPermissions', () => {
     });
   });
 });
+
+describe('hasAppliedDataFence', () => {
+  describe('user has not datafence permissions', () => {
```
tdeekens: Nit "has no datafence permissions"

packages/permissions/src/utils/has-permissions.spec.tsx
```diff
@@ -205,3 +206,178 @@ describe('getInvalidPermissions', () => {
...
+    it('should return false', () => {
```
tdeekens: Nit: I wouldn't indicate the return value but what it means: so here "should indicate not having applied datafences".

PR #1174, opened 2019-11-18
fix(app-shell/test-utils): to not require children

tdeekens:
#### Summary

This pull request removes requiring children on the providers.

#### Description

I am unsure why we set this but I don't see why we have to require children here. This fixes the failing test after the upgrade in the application https://github.com/commercetools/merchant-center-frontend/pull/7968

### DIFF
--- a/packages/application-shell/src/test-utils/test-utils.js
+++ b/packages/application-shell/src/test-utils/test-utils.js
@@ -318,7 +318,7 @@ const renderApp = (
     </IntlProvider>
   );
   ApplicationProviders.propTypes = {
-    children: PropTypes.node.isRequired,
+    children: PropTypes.node,
   };
 
   const rendered = rtl.render(ui, {
@@ -433,7 +433,7 @@ const renderAppWithRedux = (
     </NotificationProviderForCustomComponent>
   );
   ReduxProviders.propTypes = {
-    children: PropTypes.node.isRequired,
+    children: PropTypes.node,
   };
 
   const rendered = renderApp(ui, {


packages/application-shell/src/test-utils/test-utils.js
```diff
@@ -433,7 +433,7 @@ const renderAppWithRedux = (
     </NotificationProviderForCustomComponent>
   );
   ReduxProviders.propTypes = {
-    children: PropTypes.node.isRequired,
+    children: PropTypes.node,
```
tdeekens: not rendering children as in `null` should be allowed i think.

tdeekens:
Thanks for the feedback! I am also still scatching my head a bit. Specifically cause `rerender` was working contrary to claimed in the PR about the addition of working with wrappers. 

Let me have a look at `wrapIfNeeded` and see if it makes sense to handle it there.

emmenko:
What I mean is something like this in the wrapper function:

```js
children || <></>
```

emmenko:
Actually, isn't `rerender` supposed to get the same rendered component with different props? Why are we calling `rerender()` without any component? 🤔

tdeekens:
> Actually, isn't rerender supposed to get the same rendered component with different props? Why are we calling rerender() without any component? 🤔

Yes, I think the issue is in the app. I just don't follow why this spec was written as is it is.

Lastly, I changed the `wrapIfNeeded` to be `wrapper ? React.createElement(wrapper, null, children) : children || <></>;` but that just makes little sense to me.

tdeekens:
The odd and weird thing is that the spec in our application is fully flawed. We just render nothing to then expect something to not be thre which it obviously isn't casue we rendered nothing.

tdeekens:
Thanks for the critical eyes. Much appriciated. Closing.

ahmehri:
Sorry for the delayed reply, yes the problem is the wrong usage of `rerender` in that test.

PR #1183, opened 2019-11-26
fix(app-shell): portal container height to fix tabbing in modals

jonnybel:
#### Summary

This PR fixes [the bug reported by Support](https://jira.commercetools.com/browse/SUPPORT-7309) about not being able to use the Tab key to navigate between elements when inside a Modal.

Apparently, the latest `react-modal` release now ignores focusing a Modal if its parent container doesn't have any `height` or `width`. See https://github.com/reactjs/react-modal/pull/774.

This PR simply adds a `height` to the `PortalsContainer` but keeping it hidden from sight, fixing the problem.

### DIFF
--- a/packages/application-shell/src/components/portals-container/portals-container.js
+++ b/packages/application-shell/src/components/portals-container/portals-container.js
@@ -9,6 +9,8 @@ const PortalsContainer = () => (
     id={PORTALS_CONTAINER_ID}
     css={css`
       display: flex;
+      height: 1px;
+      margin-top: -1px;
     `}
   />
 );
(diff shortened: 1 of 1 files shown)


emmenko:
Thanks for looking into it. Is it possible to fix this in the modal pages directly? Or is it something related only with the portal container?

jonnybel:
No, this is directly related to the `PortalsContainer`, and affects any Modal that uses it as a parent.

Since the latest `react-modal` version, it needs to have a height/width > 0.

packages/application-shell/src/components/portals-container/portals-container.js
```diff
@@ -9,6 +9,8 @@ const PortalsContainer = () => (
     id={PORTALS_CONTAINER_ID}
     css={css`
       display: flex;
+      height: 1px;
```
tdeekens: #comment :)

emmenko:
Ok I see, then let's leave a comment with the explanation on why we need this. I think a link to this PR will suffice.

PR #1192, opened 2019-12-04
test: refactor some tests to RTL

emmenko:


packages/application-shell/src/components/gtm-booter/gtm-booter.js
```diff
@@ -16,28 +16,26 @@ export const GtmContext = React.createContext({
...
       ...defaultTrackingEventWhitelist,
-      ...this.props.trackingEventWhitelist,
+      ...props.trackingEventWhitelist,
     });
-  }
-  render() {
-    return this.props.children;
-  }
-}
+  }, [props.trackingEventWhitelist]);
```
tdeekens: Note that this is actually a change in behaviour where boot will be called again if the white list changes. Whereas before it would not.

packages/application-shell/src/components/service-page-project-switcher/service-page-project-switcher.spec.js
```diff
@@ -1,81 +1,59 @@
...
+  });
+  describe('when user has access to some projects', () => {
+    it('should render dropdown', async () => {
+      const mockedRequest = [
+        createGraphqlResponseForProjectsQuery({
+          numberOfProjects: 1,
+        }),
+      ];
+      const rendered = renderApp(
+        <Route path={`/:projectKey`} component={ServicePageProjectSwitcher} />,
```
tdeekens: This pattern will be deprecated in react-router. I think it would make sense to just render it as children.

```js
  <Route path={`/:projectKey`}><ServicePageProjectSwitcher /></Route>
```

packages/application-shell/src/components/route-catch-all/route-catch-all.js
```diff
@@ -1,51 +1,38 @@
...
-    // In case we are developing the app locally, we simply render a 404
-    // page because we most likely don't have other "apps" running at the same
-    // time.
-    return <Route component={PageNotFound} />;
-  }
-}
+  // In case we are developing the app locally, we simply render a 404
+  // page because we most likely don't have other "apps" running at the same
+  // time.
+  return <Route component={PageNotFound} />;
```
tdeekens: While we're here we could remove the component prop and just render children.

packages/application-shell/src/components/route-catch-all/route-catch-all.js
```diff
@@ -1,51 +1,38 @@
...
+  // For example, if the current "loaded" app is products and I click
+  // on a link to discounts, the products app does not know about the
+  // discount routes, thus falling back to this "catch all route" component.
+  // At this point we force a page reload, effectively handing the
+  // route control logic to the reverse proxy in our cluster. There,
+  // the router mapping will match the discounts route and it will forward
+  // the request to the discounts app.
+  // If no route matches, the application fallback will handle the request
+  // instead, showing e.g. a 404 page.
+  if (servedByProxy) return <Route component={ForcePageReload} />;
```
tdeekens: I somewhat feel strange about this as it would be true if servedByProxy was 'false' (string) which it is when it's set at times I think. Maybe in the future we can expose small hooks encapsulating this. Like `useIsServedByProxy` so we encapsulate that.

emmenko: No we coerce the values in the app provider, so we can safely expect a boolean here

https://github.com/commercetools/merchant-center-application-kit/blob/4819f78a3eb3fb7055d372f8337020980fa8993c/packages/application-shell/src/components/application-shell/application-shell.js#L490-L492

packages/application-shell/src/components/service-page-project-switcher/service-page-project-switcher.spec.js
emmenko: True, I'll change that.

packages/application-shell/src/components/gtm-booter/gtm-booter.js
emmenko: Ah good point, let me remove it then

packages/application-shell/src/components/route-catch-all/route-catch-all.js
tdeekens: Ah forgot about that. Maybe we can simplify some code in the mc-fe then.

PR #1193, opened 2019-12-05
refactor: use uikit single packages

emmenko:
With uikit 10.13.0 we now have at our disposal individual packages that we can use instead of one big package.

This should help reducing the bundle size.

PS: I split the type declarations of uikit into single type modules


emmenko:
I don't think so, cause the uikit preset just re-exports the individual packages, so there should be no duplications.

packages/application-components/package.json
```diff
@@ -52,13 +52,26 @@
...
+    "@commercetools-uikit/secondary-icon-button": "10.13.0",
+    "@commercetools-uikit/spacings": "10.13.0",
+    "@commercetools-uikit/text": "10.13.0",
     "react": "16.12.0",
     "react-dom": "16.12.0",
     "react-intl": "3.4.0"
   },
   "peerDependencies": {
-    "@commercetools-frontend/ui-kit": "10.x",
+    "@commercetools-uikit/design-system": "10.x",
```
emmenko: The only thing that "changed" is that the peer dependencies are now different.

If the apps use `@commercetools-frontend/ui-kit`, it should be fine as all those packages are already included as dependencies.
If not, apps should make sure to use the correct peerDependencies.

emmenko: @tdeekens @jonnybel do you have other ideas what we should do about it?

tdeekens: Not much I think we can do. I am not sure if peer deps work as in if you require to have C and A user B which requires C that the peer dep is satisfied or if one has to directly require C (not only transitively through B). If we can not solve this it would be a breaking change for app-kit right?

emmenko: I guess the alternative would be to not make them peer deps anymore...

tdeekens: We had that before didn't we and then people complained arguing that it's super hard to spot what peer deps app-kit has?

emmenko: Correct...it's a bit of a trade off. I think we can try with the new peer deps first and see how it goes.

jonnybel: I would keep them as peer deps. Don't know of a better alternative for this.

PR #1195, opened 2019-12-07
docs(changelog): for release v15.5.0 [skip ci]

emmenko:


CHANGELOG.md
```diff
@@ -1,3 +1,57 @@
...
+
+- `application-components`, `application-shell`, `i18n`, `jest-stylelint-runner`, `mc-scripts`, `react-notifications`
+
+  - [#1193](https://github.com/commercetools/merchant-center-application-kit/pull/1193) refactor: use uikit single packages ([@emmenko](https://github.com/emmenko))
+
+  Packages that previously required `@commercetools-frontend/ui-kit` as a peer dependency now require peer dependencies of specific uikit packages.
+
+  > If you keep using `@commercetools-frontend/ui-kit` you don't need to change anything. If you start using specific uikit packages though, you need to make sure you have the proper peer dependencies in place.
+
+  - `application-components`
```
tdeekens: Would it be worht linking to `install-peerdeps` to install those peer deps in that case automatically?

emmenko: Yeah I think we should add this to each package README also.

PR #1198, opened 2019-12-13
chore: check npm authentication on prerelease

jonnybel:
Same as https://github.com/commercetools/ui-kit/pull/1219.

> When doing a `yarn release`, if you forgot to log in to the `npm` CLI, lerna will still perform some of the steps but then fail on the `npm publish` process, resulting in dirty state (see https://github.com/lerna/lerna/issues/455).
> This PR prevents this problem by checking if the user is logged in to npm on the `prerelease` step.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "license": "MIT",
   "private": true,
   "scripts": {
+    "auth": "npm_config_registry=https://registry.npmjs.org npm whoami",
     "changelog": "node scripts/changelog.js",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -20,7 +20,7 @@
     "format:yaml": "prettier --write --parser yaml '**/*.yaml'",
     "format:graphql": "prettier --write --parser graphql '**/*.graphql'",
     "start": "yarn playground:start",
-    "prerelease": "yarn build && yarn typecheck",
+    "prerelease": "npm_config_registry=https://registry.npmjs.org npm whoami && yarn build && yarn typecheck",
```
tdeekens: `preprerelease` would work I think. A joke but still.

Could also be `"auth": npm_config_registry=https://registry.npmjs.org npm whoami `

amine-benselim: `auth` or `auth:npm` is a good idea I guess

PR #1202, opened 2019-12-16
fix: babel macros with react-intl

tdeekens:
I found this comment and wanted to give it a shot so at least we tried and have a PR documented with the attempt.

https://github.com/formatjs/react-intl/issues/1511#issuecomment-563191853


tdeekens:
Noice. It works.

PR #1204, opened 2019-12-16
chore: downgrade webpack

tdeekens:
#### Summary

In the react-intl upgrade PR the build works. Since the dep update of webpack it seems to fail. To stabalise master I propose to downgrade webpack in case this build passes.

### DIFF
--- a/application-templates/starter/package.json
+++ b/application-templates/starter/package.json
@@ -27,7 +27,7 @@
     "react": "16.12.0",
     "react-apollo": "3.1.3",
     "react-dom": "16.12.0",
-    "react-intl": "3.9.2",
+    "react-intl": "3.4.0",
     "react-redux": "7.1.3",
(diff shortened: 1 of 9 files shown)


.vscode/settings.json
```diff
@@ -18,5 +18,8 @@
     }
   },
   "typescript.tsdk": "node_modules/typescript/lib",
-  "vscode-vale.path": "${workspaceFolder}/node_modules/.bin/commercetools-vale"
+  "vscode-vale.path": "${workspaceFolder}/node_modules/.bin/commercetools-vale",
+  "workbench.colorCustomizations": {
+    "tab.unfocusedActiveBorder": "#fff0"
```
emmenko: 😅

tdeekens:
So webpack wasn't it it seems. Reverting the react-intl commit which flipping built on the other PR. So confusing.

PR #1210, opened 2019-12-28
chore: update flopflip to latest

emmenko:
To try to isolate the issue with the update to flopflip, as recently we had to downgrade it everytime renovate wanted to bump the version.


tdeekens:
Thanks. To my understanding the issue comes from adding memoization to the adapter args. Not having it was reported as a bug and I was not able to understand the issue. 

Relevant changes are from this issue and relevant prs: https://github.com/tdeekens/flopflip/issues/868.

emmenko:
You mean on our side or internally? How do you suggest to proceed here?

tdeekens:
To be honest. Not sure. I haven’t looked into it yet.

tdeekens:
Had a first look. First I rewrote the integration test (which is super nice to have) to use hooks and async expectation. This gives a a bit nicer output

CleanShot 2019-12-29 at 14 13 48@2x

This outlines that the issue is in this changeset: https://github.com/tdeekens/flopflip/pull/869/files#diff-5f7cbb5101b8386b106a525025c03783R25-R35 which is a bit confusing so far.

If I remove the memoization in the cjs-bundle used in the test, then the tests pass. So far so good. Sadly this issue is a bit a back and forth thing. Once changed to not memoize [here](https://github.com/tdeekens/flopflip/commit/4b6297dd6ea327dce44d669d382074524d5dcd1c) it is reported by the community to break in certain browsers. Adding memoization again it breaks for us. So far I don't really understand why.

tdeekens:
I think under test we actually create multiple ConfigureFlopflips and when memoizing all event handlers are bound to "one instance" which is problematic. I think somehow this needs to be solved with an sequential id on mount or a instance variable/ref uuid. Not super sure yet. Will mess around a bit.

tdeekens:
Okay the last commit should pass. The issue/challenge truely is that under test we render multiple adapter configurations. All with maybe different default flags. If flopflip starts to too eargerly create and memoize the configuration the handlers will all be bound to the first created configuration component. Which means that the later configuration will update the previous state but not their own. To strike a balance with firefox crashing (bug by community) without any memoization I added a sequential counter to the component. So that each configuration can re-render without creating new context values while different configurations are not falsely referencing each other.

tdeekens:
TIL: Wow, `lodash.memoize` really works a whole lot different than I expected. Kind of not at all 😄 

CleanShot 2019-12-29 at 16 16 34@2x

tdeekens:
Can’t stop thinking about this. The current fix is definitely fine to merge. That said I think the issue is in using lodash‘s memoize. It is truely different than expected. As it has a global cache, it caches multiple configurations far too eager as it only relies on the first parameter. This however is always the same - it is just the adapterArgs. If it would take the 2nd arg (event handlers) into account it would work as those are on the component instance (assumption). In hindsight my suspicions is we don’t even need a sequence id just a „proper“ last-call/all-args memoize fn. Taking the first arg is just an entirely different use case. Maybe Ill spin up a code sandbox later to debug the two memoization libs in isolation.

emmenko:
Ah wow, thanks a lot for looking into this and making some good progress! 🙌

Do you think you should write your own memoization function in case the existing libs do not fulfill the purpose?

tdeekens:
Interesting idea. In many cases the „let me copy that into my code“ thing might be good. However the memoize-one thing is pretty chill too. On the flip side again, having the very common things shared from app-kit (compose, memoize etc) might help avoding bundle bloat creep in by pulling in different versions all over the place.

PR #1214, opened 2019-12-30
fix(app-shell): retry unauthenticated requests for graphql errors

emmenko:
Apparently the "apollo-link-retry" does not handle graphql errors, only network errors: https://www.apollographql.com/docs/link/links/retry/

Screenshot 2019-12-30 at 10 16 20

We do want however to retry unauthenticated requests for graphql, to automatically refetch an access token as we do in the retry-link. 
I found this document that explains how to implement this use case: https://www.apollographql.com/docs/link/links/error/#retrying-failed-requests

### DIFF
--- a/packages/application-shell/src/apollo-links/error-link.js
+++ b/packages/application-shell/src/apollo-links/error-link.js
@@ -7,10 +7,31 @@ import history from '@commercetools-frontend/browser-history';
 
 // Checks response from GraphQL in order to scan 401 errors and redirect the
 // user to the login page resetting the store and showing the proper message
-const errorLink = onError(({ networkError }) => {
-  if (networkError && networkError.statusCode === STATUS_CODES.UNAUTHORIZED) {
-    history.push(`/logout?reason=${LOGOUT_REASONS.UNAUTHORIZED}`);
+const errorLink = onError(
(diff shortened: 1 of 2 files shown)


packages/application-shell/src/apollo-links/error-link.js
```diff
@@ -7,10 +7,31 @@ import history from '@commercetools-frontend/browser-history';
...
+  ({ graphQLErrors, networkError, operation, forward }) => {
+    if (networkError && networkError.statusCode === STATUS_CODES.UNAUTHORIZED) {
+      history.push(`/logout?reason=${LOGOUT_REASONS.UNAUTHORIZED}`);
+    }
+    // In case of graphql errors, we want to retry unauthenticated requests by
+    // forcing our API to fetch a new token, using the `X-Force-Token` header.
+    // https://www.apollographql.com/docs/link/links/error/#retrying-failed-requests
+    // We need to do this as the `token-retry-link` only works for network errors.
+    // https://www.apollographql.com/docs/link/links/retry/
+    if (graphQLErrors) {
```
tdeekens: Nice! Thanks for the comment above. Do you think we can slap a test at this? I remember Jan Florian back then did some work to test links and their observables.

PR #1215, opened 2019-12-31
fix: building type declarations

emmenko:
I noticed that the type declarations shipped with the packages were including all types from all packages, which is wrong:

image

Instead, only the type declarations of the related package should be built and shipped:

image

I think this has been "broken" since #1047

### DIFF
--- a/packages/actions-global/tsconfig.declarations.json
+++ b/packages/actions-global/tsconfig.declarations.json
@@ -1,5 +1,6 @@
 {
   "extends": "./tsconfig.json",
   "compilerOptions": { "declaration": true, "isolatedModules": false },
-  "exclude": ["**/*.spec.ts", "**/*.spec.tsx"]
+  "exclude": ["**/*.spec.ts", "**/*.spec.tsx"],
+  "include": []
 }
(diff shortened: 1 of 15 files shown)


packages/i18n/package.json
```diff
@@ -20,8 +20,8 @@
   },
   "main": "./dist/i18n-index.cjs.js",
   "module": "./dist/i18n-index.es.js",
-  "typings": "./dist/typings/index.d.ts",
-  "types": "./dist/typings/index.d.ts",
+  "typings": "./dist/typings/src/index.d.ts",
+  "types": "./dist/typings/src/index.d.ts",
```
emmenko: Somehow for the i18n and the l10n package the type declarations are put into a `src` folder

image

This is probably because of the `data` folder (I haven't been able to exclude it)

packages/actions-global/tsconfig.declarations.json
```diff
@@ -1,5 +1,6 @@
 {
   "extends": "./tsconfig.json",
   "compilerOptions": { "declaration": true, "isolatedModules": false },
-  "exclude": ["**/*.spec.ts", "**/*.spec.tsx"]
+  "exclude": ["**/*.spec.ts", "**/*.spec.tsx"],
+  "include": []
```
emmenko: The "fix" that I found is basically to override the `include` list of the base `tsconfig.json`, and only include stuff that is necessary where needed.

PR #1228, opened 2020-01-09
fix(app-shell): denormalization of action rights

tdeekens:
#### Summary

By not initializing the reducer's accumulator we end up messing up the first value of the accumulated result.


tdeekens:
I decided against adding tests for this for now as we plan to migrate apps away from it anyway. Shady argument I know but I am not aware of custom app users yet to test with permissions at all.

PR #1229, opened 2020-01-09
chore: regenerate graphql types

emmenko:


packages/application-shell-connectors/src/components/application-context/application-context.spec.tsx
```diff
@@ -58,6 +59,7 @@ const createTestProject = (custom = {}) => ({
   },
   suspension: {
     isActive: false,
+    reason: TProjectSuspensionReason.Other,
```
emmenko: The `reason` field was previously missing in the graphql query

packages/application-shell/src/components/fetch-user/fetch-user.mc.graphql
```diff
@@ -1,5 +1,3 @@
-#import "../project-switcher/project-switcher-fragments.mc.graphql"
```
emmenko: I had to remove this as the codegen was complaining of "duplicate types" related to `FetchLoggedInUser`

packages/application-shell/src/components/fetch-user/fetch-user.mc.graphql
```diff
@@ -18,7 +16,14 @@ query FetchLoggedInUser {
     projects {
       total
       results {
-        ...projectFragment
```
tdeekens: Hm. The idea behind this was to always have these two in sync so we avoid an additional fetch. A bit frustrating the tooling forces us to be "simple" here at a risk.

packages/application-shell/src/components/fetch-project/fetch-project.mc.graphql
```diff
@@ -13,6 +13,7 @@ query FetchProject($projectKey: String!) {
     }
     suspension {
       isActive
+      reason
```
emmenko: This was missing although we were using it here: https://github.com/commercetools/merchant-center-application-kit/blob/df68ce4bc5da71f9cbc763e356d777a6cbb4f6c6/packages/application-shell/src/components/project-container/project-container.js#L158

packages/application-shell/src/components/fetch-user/fetch-user.mc.graphql
emmenko: Yeah, I still don't know exactly what changed but that's the only solution I could find so far that would work.

PR #1069, opened 2019-09-20
refactor(app-shell): migrate to TS

emmenko:
Ref #589

PR #1231, opened 2020-01-09
test(app-shell): rewrite top level tests to integration and e2e tests

emmenko:
This PR rewrites the top level AppShell tests.

At the moment, the AppShell has enzyme (shallow) tests which are very much tighten to the implementation. As part of our testing strategy migration to RTL, we want to move to refactor the AppShell tests as well.
Doing that comes with some benefits, like less painful and more confident refactorings. This will also help with the migration to TS (#1069).

Testing the AppShell is actually not that trivial. We can't use the test-utils, because they effectively mock out the AppShell, which is what we want to test.
What we can do though is to mock the network requests (_basically the same approach used with the experimental render_).

Long story short: I ended up creating graphql mocks for both RTL and Cypress. The AppShell tests have been rewritten to test the full component and mock the underlying XHR requests.

image

/cc @pa3 for awareness


tdeekens:
Oh wow. This is really nice 💯. Need to set aside some time to have a closer look. Wondering if it would maybe make sense to announce this in a fe-chapter meeting and pick 1-2 dedicate reviewers how will get a walk high level walkthrough to then review the change?

emmenko:
Sure, I was hoping to get this merged soon-ish so that I can move on with other things.
I can present this in the next fe tasting maybe, although it's really nothing new.

I was also wondering if we would publish a package for those utils, so that we can use them in the apps. We might also include the pulled schemas in the package (still not 100% sure about it though).

tdeekens:
> Sure, I was hoping to get this merged soon-ish so that I can move on with other things.

I see but maybe that's not always possible. Still I think it's not scalable that I am assigned the only reviewer again. Albeit I might have some context. It's not a way we can or should with or towards the team I feel.

emmenko:
Yeah sure, I can ask in slack.

cypress/integration/playground/application-shell.js
```diff
@@ -0,0 +1,28 @@
...
+    cy.findByText('Logout').click();
+    const queryParams = encode({
+      reason: LOGOUT_REASONS.USER,
+    });
+    cy.url().should('include', `/logout?${queryParams}`);
+    cy.findByText('This is the logout page for local development.').should(
+      'exist'
+    );
+  });
+  describe('when going to an unknown route', () => {
```
tdeekens:
```suggestion
  describe('when navigating to an unknown route', () => {
```

cypress/integration/playground/application-shell.js
```diff
@@ -0,0 +1,28 @@
...
+    it('should render page not found', () => {
```
tdeekens:
```suggestion
    it('should render a not found page', () => {
```

graphql-test-utils/create-graphql-mock-server.js
```diff
@@ -0,0 +1,69 @@
+import { graphql, buildClientSchema } from 'graphql';
+import { addMockFunctionsToSchema } from 'graphql-tools';
+import { formatApolloErrors } from 'apollo-server-errors';
+import loadSchema from './load-schema';
+
+const getRootValue = (operations, operationName, variables) => {
+  if (!operationName || !operations[operationName]) {
+    return {};
+  }
+  const op = operations[operationName];
```
tdeekens: The rootValue is an operation, right?

graphql-test-utils/create-graphql-mock-server.js
```diff
@@ -0,0 +1,69 @@
...
+  if (typeof op === 'function') {
+    try {
+      return op(variables);
+    } catch (e) {
+      return e; // properly handle dynamic throw new GraphQLError("message")
+    }
+  }
+  return op;
+};
```
tdeekens: Nit and subjective :) this would have helped me:

```js
const executeGraphQlOperation = (operation, variables) => {
   if (typeof operation === "function") {
      try {
         return operation(variables);
      } catch (e) {
         // TODO: Properly handle dynamic throw new GraphQLError("message")
         return e; 
      }
   }

   return operation
};
const getRootValue = (operations, operationName, variables) => {
   if (!operationName || !operations[operationName]) {
      return {};
   }

   const operation = operations[operationName];

   return executeGraphQlOperation(operation, variables);
};

```

graphql-test-utils/create-graphql-mock-server.js
```diff
@@ -0,0 +1,69 @@
...
+
+const createGraphqlMockServer = (
+  xhrMock,
+  { mocksByTarget = {}, operationsByTarget = {}, onRequest } = {}
+) => {
+  const processGraphQLRequest = async (targetName, req, res) => {
```
tdeekens: Nit: but we always mix casing.

```suggestion
  const processGraphQlRequest = async (targetName, req, res) => {
```

graphql-test-utils/create-graphql-mock-server.js
```diff
@@ -0,0 +1,69 @@
...
+    return res.status(200).body(JSON.stringify(response));
+  };
+  const handleProxyGraphQLRequest = async (req, res) => {
+    const { variables } = JSON.parse(req.body());
+    const targetName = variables.target;
+    return await processGraphQLRequest(targetName, req, res);
+  };
+
+  xhrMock.post('http://localhost:8080/graphql', async (req, res) => {
+    if (onRequest) {
```
tdeekens: When does this happen? Why do we need to wrap the next?

graphql-test-utils/create-graphql-mock-server.js
```diff
@@ -0,0 +1,69 @@
...
+      const next = async () => {
+        return await handleProxyGraphQLRequest(req, res);
+      };
+      return await onRequest(req, res, next);
+    }
+    return await handleProxyGraphQLRequest(req, res);
+  });
+  xhrMock.post(/\/api\/graphql$/, async (req, res) => {
```
tdeekens: This is the mc-proxy right? Should we add a comment what requests we're handling maybe?

graphql-test-utils/cypress/add-commands.js
```diff
@@ -0,0 +1,13 @@
...
+
+Cypress.Commands.add('useMockedGraphqlApi', (customOperations = {}) => {
+  cy.task('getGraphQLSchema', 'mc').then(schema => {
+    cy.mockGraphql({
+      schema,
+    });
+  });
+  const operations = mocksForMc.createMockOperations(customOperations);
+  cy.mockGraphqlOps({ operations });
+  return cy.wrap(operations);
```
tdeekens: Do you know what this does?

graphql-test-utils/graphql-models/user.js
```diff
@@ -0,0 +1,26 @@
...
+  )
+  .attr('launchdarklyTrackingTeam', () => [faker.random.word()])
+  .attr('launchdarklyTrackingTenant', 'gcp-eu')
+  .attr('defaultProjectKey', () => null)
+  .attr('projects', () => ({
+    total: 0,
+    results: [],
+  }));
+
+export default User;
```
tdeekens: Nice that we use rosie and faker here too :)

graphql-test-utils/load-schema.js
```diff
@@ -0,0 +1,19 @@
+const fs = require('fs');
```
tdeekens: Is this really loading? Is it more `resolveAndCacheSchema`?

packages/application-shell/src/components/application-shell-provider/utils.spec.js
```diff
@@ -0,0 +1,29 @@
...
+  describe('when the locale is supported by the MC', () => {
+    beforeEach(() => {
+      testWindow = {
+        navigator: {
+          language: 'de-DE',
+        },
+      };
+    });
+    it('should return the locale', () => {
+      expect(getBrowserLocale(testWindow)).toBe('de-DE');
```
tdeekens: Would it be easier to avoid beforeEach and just do

```js
expect(getBrowserLocale({
  navigator: {
     language: 'de-DE',
  }
})).toBe('de-DE');
```

maybe?

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -245,7 +245,8 @@ const hasMenuEveryMenuVisibilitySetToBeHidden = (
   Array.isArray(namesOfMenuVisibilities) &&
   namesOfMenuVisibilities.length > 0 &&
   namesOfMenuVisibilities.every(
-    nameOfMenuVisibility => menuVisibilities[nameOfMenuVisibility] === true
+    nameOfMenuVisibility =>
```
tdeekens: What's the reasoning behind the change? I think that's also why the prop becomes non-required, right?

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -1,508 +1,446 @@
...
+    de: { 'CustomApp.title': 'Titel' },
+  },
   ...props,
 });
 
-const testLocaleData = {
-  isLoading: false,
-  locale: 'en',
-  messages: { 'AppKit.title': 'Title en', 'CustomApp.title': 'Title en' },
+const mockSuperfluousRequests = () => {
```
tdeekens: Nit: but should we extract this somehow? Feels like we might have more in the future, right?

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -1,508 +1,446 @@
...
-          })
-        );
-      });
-      it('should not render <ApplicationContextProvider>', () => {
-        expect(wrapper).not.toRender(ApplicationContextProvider);
-      });
+  });
+  it('should not render NavBar', async () => {
+    const rendered = renderApp();
+    await waitForElement(() => rendered.getByText('OK'));
```
tdeekens: Nit I think we can use `await findByText` here everywhere to make it a bit leaner maybe?

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -1,508 +1,446 @@
...
+    createGraphqlMockServer(xhrMock, {
+      operationsByTarget: {
+        mc: mocksForMc.createMockOperations({
+          FetchLoggedInUser: {
+            me: UserMock.build({
+              defaultProjectKey: null,
+              projects: { total: 0, results: [] },
+            }),
+          },
+        }),
```
tdeekens: This is nice!

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -1,508 +1,446 @@
...
-    it('should return the default locale, `en`', () => {
-      expect(getBrowserLocale(testWindow)).toBe('en');
+    await wait(() => {
+      expect(window.location.replace).toHaveBeenCalledWith(
+        `${window.location.origin}/login?${queryParams}`
+      );
+      expect(rendered.queryByText('OK')).not.toBeInTheDocument();
     });
   });
 });
```
tdeekens: The test cases are so nice and close to the requirements!

graphql-test-utils/create-graphql-mock-server.js
emmenko: Yeah I took this from here: https://github.com/tgriesser/cypress-graphql-mock/blob/6000875ebc9a16f0a3a20fbef72a2c41f0379eca/src/index.ts#L194-L212

I can apply your suggestion though, thanks

graphql-test-utils/create-graphql-mock-server.js
```diff
@@ -0,0 +1,69 @@
...
+    } catch (error) {
+      // https://github.com/apollographql/apollo-server/blob/61e69412c0aa88b358a2eef4ba635e8618db9946/packages/apollo-server-core/src/runHttpQuery.ts#L295-L297
+      response = { error: formatApolloErrors([error]) };
+    }
+    return res.status(200).body(JSON.stringify(response));
+  };
+  const handleProxyGraphQLRequest = async (req, res) => {
+    const { variables } = JSON.parse(req.body());
+    const targetName = variables.target;
+    return await processGraphQLRequest(targetName, req, res);
```
emmenko:
```suggestion
    return await processGraphQlRequest(targetName, req, res);
```

graphql-test-utils/create-graphql-mock-server.js
emmenko: We use it here: https://github.com/commercetools/merchant-center-application-kit/pull/1231/files#diff-935055bd8b0f453f6f0596df2db12c6aR422-R430

This is useful when we want to control the network request, for example failing with 401, which is different from a graphql error.

graphql-test-utils/create-graphql-mock-server.js
emmenko: Yes the proxy.

graphql-test-utils/cypress/add-commands.js
emmenko: Yes, it allows to get the value like this:

```js
cy.useMockedGraphqlApi().then(operations => {
})
```

emmenko: Otherwise it won't work, because Cypress queues everything that happens in the commands and executes them asynchronously.

graphql-test-utils/load-schema.js
emmenko: Couldn't find a better name 😅

packages/application-shell/src/components/application-shell/application-shell.spec.js
emmenko: I guess..should I put it in the graphql-test-utils?

packages/application-shell/src/components/application-shell/application-shell.spec.js
emmenko: Hmm what do you mean?

graphql-test-utils/create-graphql-mock-server.js
tdeekens: Ah kay. Doesn't matter too much. Just what my mind translated the code to :)

packages/application-shell/src/components/navbar/navbar.js
emmenko: Yes, it's an optional prop, the prop types were incorrect

packages/application-shell/src/components/application-shell/application-shell.spec.js
tdeekens: Yeah could help to just have a `applyExternalNetworkRequests` there or something.

packages/application-shell/src/components/application-shell/application-shell.spec.js
tdeekens: To my understandting RTL released and documented the `findBy`-APIs which are like what we have here so `await waitForElement(() => rendered.getByText('OK'));` can be shortenede to `await rendered.findByText('OK')`.

emmenko: Ah, I thought it was only used in cypress or something.

Sure, I we can do that, thanks for the heads up

graphql-test-utils/create-graphql-mock-server.js
tdeekens: Ah so it's like intercepting it? I thought it was on any request.

graphql-test-utils/cypress/add-commands.js
tdeekens: Ah this auto queueing promisification. Got it.

graphql-test-utils/create-graphql-mock-server.js
emmenko: No only if you pass the `onRequest` option

tdeekens: Yes, it you pass an request interceptor which is called onRequest then it will be intercepted.

packages/application-shell/src/test-utils/fixtures/mc.js
```diff
@@ -0,0 +1,60 @@
...
+  .attr('owner', {
+    id: faker.random.uuid(),
+  });
+const User = new Factory()
+  .sequence('sequenceId')
+  .attr('email', faker.internet.email())
+  .attr('gravatarHash', faker.random.alphaNumeric(16))
+  .attr('firstName', faker.name.firstName())
+  .attr('lastName', faker.name.lastName())
+  .attr('language', faker.locale)
```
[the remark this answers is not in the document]
emmenko: I want to keep the language to a "known" locale. I can actually just hardcode it to 'en' to avoid confusion.

cypress/integration/playground/application-shell.js
```diff
@@ -0,0 +1,69 @@
...
+  //   cy.task('getGraphQLSchema', 'mc').then(schema => {
+  //     cy.mockGraphql({
+  //       schema,
+  //     });
+  //   });
+  // });
+  describe('when user is not authenticated', () => {
+    beforeEach(() => {
+      cy.server();
+      cy.on('window:before:load', win => {
```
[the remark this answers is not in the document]
emmenko: I tried that first but I wasn't able to make it work.

cypress/integration/playground/application-shell.js
```diff
@@ -70,45 +70,28 @@ describe('when user is authenticated', () => {
...
-        .type(nextProject.name);
-      cy.findByText(nextProject.name).click();
-      cy.url().should('include', `/${nextProject.key}/state-machines`);
+      cy.useMockedGraphqlApi().then(resolvers => {
+        const nextProject = resolvers.FetchLoggedInUser.me.projects.results[1];
+        cy.login();
+        cy.get('[data-test=top-navigation').should('exist');
+        cy.get('input[aria-labelledby="project-switcher"]')
+          .focus()
+          .click({ force: true });
```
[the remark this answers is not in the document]
emmenko: Somehow it doesn't do anything without `force` 🤷‍♂️

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -1,508 +1,354 @@
...
+  window.localStorage.setItem.mockClear();
+  window.localStorage.getItem.mockClear();
+  window.localStorage.removeItem.mockClear();
+  window.location.replace = jest.fn();
+  reportErrorToSentry.mockClear();
+  window.localStorage.getItem.mockReturnValue('true');
+  ApplicationShellProvider.history = defaultHistory;
+  appShellUtils.getPreviousProjectKey.mockClear();
+});
+afterEach(() => {
```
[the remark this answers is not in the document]
emmenko: I would prefer to keep `afterEach`, it gives a better understanding on what needs to be cleaned up.

PR #1234, opened 2020-01-11
refactor(app-shell): migrate more components to TS

emmenko:
Base branch #1069


packages/application-shell/src/components/handle-apollo-errors/handle-apollo-errors.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { QueryResult } from 'react-apollo';
+import { handleActionError } from '@commercetools-frontend/actions-global';
+import getDisplayName from '../../utils/get-display-name';
+
+type QueryResultNames = string[];
+type Props = {
+  [key: string]: QueryResult;
+};
+
+function handleApolloErrors(queryResultNames: QueryResultNames) {
```
tdeekens: When should we use this again? :)

emmenko: I think we should deprecate it and eventually remove it. I don’t think anyone is using this 😅

PR #1238, opened 2020-01-13
feat(actions-global): add use-hide-all-page-notifications

tdeekens:
#### Summary

This pull request adds a hook to hide all page notifications.


emmenko:
Can you also add it for the `handleActionError`?

tdeekens:
> Can you also add it for the handleActionError?

Yup, send PR. At least that was my plan.

PR #1239, opened 2020-01-13
refactor(app-shell): migrate UserMenu to TS

emmenko:
Base branch #1069 

FYI: to make the types more "sound" and easy to use, I refactored the `useApplicationMenu` hook a bit.


packages/application-shell/src/components/user-settings-menu/user-settings-menu.tsx
```diff
@@ -151,29 +169,30 @@ const UserSettingsMenuBody = props => {
...
+              </Link>
+            </OptionalFeatureToggle>
+          ))}
         <MenuItem hasDivider={true} />
         <a
           href={`https://commercetools.com/privacy#suppliers`}
           target="_blank"
           rel="noopener noreferrer"
-          onClick={props.downshiftProps.toggleMenu}
+          onClick={() => props.downshiftProps.toggleMenu()}
```
amine-benselim: what's the difference in here? 🤔

emmenko: Simply that the HTML `event` is not "compatible" with the expected argument of `toggleMenu` 🤷‍♂️

PR #1240, opened 2020-01-13
feat(actions-global): to add use-on-action-error

tdeekens:
#### Summary

Similar to the prior PR this adds useOnActionError.

### DIFF
--- a/packages/actions-global/src/hooks/use-on-action-error.ts
+++ b/packages/actions-global/src/hooks/use-on-action-error.ts
@@ -8,6 +8,7 @@ export default function useOnActionError(error) {
     Dispatch<ReturnType<typeof handleActionError>>
   >();
   return React.useCallback(() => {
-    dispatch(handleActionError(error));
+    const action = handleActionError(error);
+    if (action) dispatch(action);
   }, [dispatch, error]);
 }
--- a/packages/actions-global/src/hooks/use-on-action-error.ts
+++ b/packages/actions-global/src/hooks/use-on-action-error.ts
@@ -1,0 +1,13 @@
+import React from 'react';
+import { Dispatch } from 'redux';
+import { useDispatch } from 'react-redux';
+import { handleActionError } from '../actions';
+
+export default function useOnActionError(error) {
+  const dispatch = useDispatch<
+    Dispatch<ReturnType<typeof handleActionError>>
+  >();
+  return React.useCallback(() => {
+    dispatch(handleActionError(error));
+  }, [dispatch, error]);
+}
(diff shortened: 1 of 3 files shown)


tdeekens:
The `handleActionError` return seems to be https://github.com/commercetools/merchant-center-application-kit/blob/master/packages/actions-global/src/actions/handle-action-error.ts#L31 but  this indicates it can return null (https://github.com/commercetools/merchant-center-application-kit/blob/master/packages/actions-global/src/actions/handle-action-error.ts#L51). Maybe that's why the useActionError doesn't pass the type check.

packages/actions-global/src/hooks/use-on-action-error.ts
```diff
@@ -0,0 +1,13 @@
...
+import { Dispatch } from 'redux';
+import { useDispatch } from 'react-redux';
+import { handleActionError } from '../actions';
+
+export default function useOnActionError(error) {
+  const dispatch = useDispatch<
+    Dispatch<ReturnType<typeof handleActionError>>
+  >();
+  return React.useCallback(() => {
+    dispatch(handleActionError(error));
```
emmenko: You need to separate this call.

```suggestion
    const action = handleActionError(error);
    if (action) dispatch(action);
```

tdeekens: Thanks TypeScript. Thanks @emmenko :)

packages/actions-global/src/hooks/use-on-action-error.ts
```diff
@@ -0,0 +1,14 @@
...
+import { useDispatch } from 'react-redux';
+import { handleActionError } from '../actions';
+
+export default function useOnActionError(error) {
+  const dispatch = useDispatch<
+    Dispatch<ReturnType<typeof handleActionError>>
+  >();
+  return React.useCallback(() => {
+    const action = handleActionError(error);
+    if (action) dispatch(action);
```
emmenko:
```suggestion
    if (action) return dispatch(action);
```

We should probably return this

emmenko:
Ah wait, let me have a look

packages/actions-global/src/hooks/use-on-action-error.ts
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { useDispatch } from 'react-redux';
+import handleActionError, {
+  ActionError,
+  DispatchActionError,
+} from '../actions/handle-action-error';
+
+export default function useOnActionError(error: ActionError) {
+  const dispatch = useDispatch<DispatchActionError>();
+  return React.useCallback(() => handleActionError(error)(dispatch), [
```
emmenko: @tdeekens note that it's not `dispatch(handleActionError(error))` but `handleActionError` is a factory function

PR #1241, opened 2020-01-15
feat(app-shell): add launchdarkly tracking subgroup

tdeekens:
#### Summary

This pull request _prematurely_ adds the `launchdarklyTrackingsubgroup`.

#### Description

This is still under development in the API. I wanted to see what parts have to change and get some early feedback.


schemas/mc.json
```diff
@@ -6081,4 +6093,4 @@
       "timestamp": "Tue Dec 31 2019 16:44:28 GMT+0100 (Central European Standard Time)"
```
tdeekens: I will pull this once deployed.

tdeekens:
This is non customer facing but a feature for us as the MC. I am not sure if we want to mark it as chore? WDYT? :)

tdeekens:
@adnasa and @emmenko this is now ready for review :)

tdeekens:
> Do you know what changed here?

Not really. I just pulled the schema down. I think what changes is the "custom part" of the schema for the test project. So e.g. what custom types are created. Just a guess after looking into the file locally.

emmenko:
Ok thanks for checking

PR #1242, opened 2020-01-15
fix(mc-scripts): allowing to opt-out of parallel builds

tdeekens:
#### Summary

This pull request would allow us to opt out of parallel builds. E.g. when `proces.env.CIRCLE_CI` is true.


tdeekens:
Don't feel this would really be feature worthy. Still obviously willing to release it as such.

PR #1246, opened 2020-01-18
refactor(app-shell): migrate more components to TS

emmenko:
Base branch #1069 

Continue to migrate the remaining "simple" components of appshell.


packages/application-shell/src/components/version-tracker/version-tracker.tsx
```diff
@@ -0,0 +1,56 @@
...
+    Dispatch<ReturnType<typeof pushDependencyVersionCounter>>
+  >();
+  React.useEffect(() => {
+    const result = (dispatch(
+      pushDependencyVersionCounter({
+        payload: createVersionMetric({
+          applicationName,
+        }),
+      })
+    ) as unknown) as Promise<unknown>;
```
emmenko: FYI: I bumped into this issue and therefore noticed that the SDK types still need to be adjusted.
The main issue is that the SDK creates “normal” plain actions but the middleware returns a Promise with the HTTP response. When using `dispatch`, it doesn’t know anything about the return type, unless otherwise specified, which means that we need to “patch it up”. For example we can expose a `useAsyncDispatch` which just dispatches the action normally but provides the "correct” types.
I’ll do that in a separate PR.

Issue #1247, opened 2020-01-19
Perfomance metrics are not being sent anymore


emmenko:
**Describe the bug**
As I was migrating the `MeasureFirstPaint` component to TS, I noticed that the API of Perfume changed, as the types were complaining that a lot of things we are currently using do not work/exist anymore.

I checked our metrics and confirmed that indeed we are not getting any metrics since at least a couple of weeks.

<img width="898" alt="Screenshot 2020-01-19 at 00 05 27" src="https://user-images.githubusercontent.com/1110551/72678269-ad9cbf80-3aa4-11ea-8d21-af70d4dad290.png">

Not sure what happened. Checking the changelog of perfume shows that there were some breaking changes in between "minor" versions, which is just ridiculous! https://github.com/Zizzamia/perfume.js/blob/master/CHANGELOG.md

If we inspect the `perfume` object, we notice that the things we were using do not exist anymore:

- `perfume.observeFirstContentfulPaint` is `undefined`
- `perfume.firstPaintDuration` is `undefined`

<img width="557" alt="image" src="https://user-images.githubusercontent.com/1110551/72678337-30be1580-3aa5-11ea-81e5-baf0576eb1f8.png">

We also didn't notice this as `Promise.all([undefined])` does not throw any error.

```
> Promise.all([undefined]).then(r => {console.log(r)})
Promise { <pending> }
> [ undefined ]
```

**Expected behavior**
Metrics should be correctly gathered and sent

---

I'm thinking about implementing the metrics on our own, as I don't feel much comfortable anymore relying on this library.

emmenko:
We can try to make it work again first, and then think about how we can maintain this in the future.

tdeekens:
Thanks for writing it down. I also had my frustrations with the library. It is less of what it does but more the release process. Maybe we can downgrade for now and then discuss updating or migrating away. If I am not mistaken 3.0.4 should work as I remember fixing something in October https://github.com/commercetools/merchant-center-application-kit/pull/1091/files

emmenko:
I'm actually looking into using the `analyticsTracker` callback of the library, as you get all the values there. I'm also considering sending all those values to GA instead of our prometheus server, as we can track more information. We should probably also have this tracking as a default in the app-shell, since we already use GA, no need to manually configure it.

I'll try to spin up a PR and we can discuss there if we want to go with that approach or not.

emmenko:
[...]

Issue #1248, opened 2020-01-19
Upcoming breaking changes (v16)


emmenko:
As we're nearing towards the official "beta" release of custom apps, we should probably take the chance to release a new major version with some of the things we would like to have included in the beta phase from the beginning.

Here are some of the things that come to my mind:
- [x] finally merge ~~#1008~~ #1269 
- [x] ~~remove static `Intl` exports from the dialog/modal pages components~~ #1283 
- [x] make `applicationName` a required field in `env.json` (this would be important for tracking usages during the beta phase) #1279 
- [x] finish and ship TS declarations (ref #589)
- [x] remove `MeasureFirstPaint` #1280 
- [x] remove ui component readmes and only keep mdx files. The readme then links to the docs page https://github.com/commercetools/merchant-center-application-kit/pull/1281/commits/e0d58ea93185ddd2153636ba545050011e70d6a5
- [x] change links to new website docs
- ???

emmenko:
@jonnybel would you have time to pick up #1008 next week?

jonnybel:
Yes, I can update it  👍

tdeekens:
I have to think of some changes we might want to do to our permissions package. I am not sure it's even used by any custom app. Maybe also addition warnings and deprecation things first would be better than just going all mad on it.

emmenko:
Yeah, whatever you might want to do add it to this list. I plan to release this within the next 2 weeks.

emmenko:
@tdeekens can you think of anything we would need to include here regarding the custom apps tracking thing?

tdeekens:
I commented somewhere else if you mean the differentiation between internal and external apps in logs?

PR #1249, opened 2020-01-19
feat: send performance timing values to GTM

emmenko:
See #1247 

image

TODOs:
- [ ] decide what to do with custom dimensions for tracking `applicationName` and `projectKey`
- [ ] configure tracking event `TimingEvent` in GTM

### DIFF
--- a/packages/application-shell/src/track-performance.js
+++ b/packages/application-shell/src/track-performance.js
@@ -1,0 +1,83 @@
+import Perfume from 'perfume.js';
+import { trackTiming } from './utils/gtm';
+
+const trackPerformance = () => {
+  new Perfume({
+    logging: false,
+    dataConsumption: true,
+    largestContentfulPaint: true,
+    analyticsTracker: ({ metricName, data, duration }) => {
+      switch (metricName) {
+        case 'firstPaint':
+        case 'firstContentfulPaint':
+        case 'firstInputDelay':
+        case 'largestContentfulPaint': {
+          trackTiming({
+            category: 'Loading Performance',
+            variable: metricName,
+            value: duration,
+          });
+          break;
+        }
+        // Navigation Timing collects performance metrics for the life and
+        // timings of a network request.
+        case 'navigationTiming': {
+          trackTiming({
+            category: 'Navigation Timing',
+            variable: metricName,
+            value: data.fetchTime,
+            label: 'Fetch time',
+          });
+          trackTiming({
+            category: 'Navigation Timing',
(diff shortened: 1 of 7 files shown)


packages/application-shell/src/track-performance.js
```diff
@@ -0,0 +1,83 @@
...
+            category: 'Navigation Timing',
+            variable: metricName,
+            value: data.timeToFirstByte,
+            label: 'Time To First Byte',
+          });
+          trackTiming({
+            category: 'Navigation Timing',
+            variable: metricName,
+            value: data.dnsLookupTime,
+            label: 'DSN lookup time',
```
adnasa: nit: this is `DNS`, right?

emmenko: Yes thanks

emmenko:
Will pick this up in the next days as I need to make some changes in the GTM settings first.

emmenko:
image

emmenko:
I also added the custom dimensions (only on staging for now):

image

tdeekens:
> I also added the custom dimensions (only on staging for now):

What would happen if those dimensions don't exist on one environment? I think GCP Prod EU is mostly/only used to get tracking insight so far. So it might be fine to also for know just have it there?

emmenko:
The dimensions are optional, in a sense that if they don't get mapped/configured nothing happens, even if we "push them to gtm".

tdeekens:
Then we might only need to configure them on EU (maybe US) and from my view that should be enough. What do you think?

emmenko:
Sure, although I would like to configure gtm to be same in all envs. It's not much to do, I can show you if you want

tdeekens:
> Sure, although I would like to configure gtm to be same in all envs. It's not much to do, I can show you if you want

Yeah true. Just wanted to reduce the amount of work if nobody uses it in the end. Maybe we should add it to one of the confluence sites how to setup a new env so we also do it for new envs?

emmenko:
Already documented 😉

adnasa:
> It's not much to do, I can show you if you want

I'm in on this

tdeekens:
> Already documented 😉

Yup, but we might want to add these from this PR there too, right?

emmenko:
FYI: I updated the document.

emmenko:
Alright, I updated the GA/GTM settings to accept the new events

emmenko:
As a follow up, I'll deprecate the `MeasureFirstPaint` component, which we can remove in the #1248

PR #1250, opened 2020-01-19
feat(sdk): add useAsyncDispatch for resolving async sdk actions (static types)

emmenko:
Addresses this: https://github.com/commercetools/merchant-center-application-kit/pull/1246#discussion_r368243447

### DIFF
--- a/packages/sdk/src/hooks/use-async-dispatch.ts
+++ b/packages/sdk/src/hooks/use-async-dispatch.ts
@@ -1,0 +1,12 @@
+import { Dispatch } from 'redux';
+import { useDispatch } from 'react-redux';
+import { TSdkAction } from '../types';
+
+// Wraps `dispatch` and cast the return type to a Promise, as the middleware
+// returns a Promise.
+function useAsyncDispatch<Action extends TSdkAction, Data>() {
+  const dispatch = useDispatch<Dispatch<Action>>();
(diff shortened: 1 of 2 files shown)


emmenko:
@tdeekens what changes do you request?

tdeekens:
Woooops. My bad. Admittedly a bad button press from the morning coffee iPad session :)

packages/sdk/src/hooks/use-async-dispatch.ts
```diff
@@ -0,0 +1,12 @@
+import { Dispatch } from 'redux';
+import { useDispatch } from 'react-redux';
+import { TSdkAction } from '../types';
+
+// Wraps `dispatch` and cast the return type to a Promise, as the middleware
+// returns a Promise.
+function useAsyncDispatch<Action extends TSdkAction, Data>() {
+  const dispatch = useDispatch<Dispatch<Action>>();
+  return (action: Action) => (dispatch(action) as unknown) as Promise<Data>;
```
tdeekens: `dispatch(action) as unknown` <= can you share why/what the `as unknown` means here? Thanks :)

emmenko: It's to force a type cast. You cast it to `unknown/any` (I personally always try to avoid `any`) and then cast it again to whatever you need to.

PR #1254, opened 2020-01-20
fix(babel-preset-mc-app): remove uncalled-for plugins

tdeekens:
#### Summary

With the latest release of babel some of the plugins we're still using are removable.

CleanShot 2020-01-20 at 09 03 16@2x

### DIFF
--- a/packages/babel-preset-mc-app/index.js
+++ b/packages/babel-preset-mc-app/index.js
@@ -127,10 +127,7 @@ module.exports = function getBabePresetConfigForMcApp() {
         },
       ],
       require('@babel/plugin-proposal-do-expressions').default,
-      require('@babel/plugin-proposal-optional-chaining').default,
-      require('@babel/plugin-proposal-nullish-coalescing-operator').default,
       // Adds syntax support for import()
-      require('@babel/plugin-syntax-dynamic-import').default,
       isEnvTest &&
         // Transform dynamic import to require
         require('babel-plugin-transform-dynamic-import').default,
--- a/packages/babel-preset-mc-app/index.js
+++ b/packages/babel-preset-mc-app/index.js
@@ -127,10 +127,6 @@ module.exports = function getBabePresetConfigForMcApp() {
         },
       ],
       require('@babel/plugin-proposal-do-expressions').default,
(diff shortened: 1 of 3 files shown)


packages/babel-preset-mc-app/index.js
```diff
@@ -127,10 +127,7 @@ module.exports = function getBabePresetConfigForMcApp() {
         },
       ],
       require('@babel/plugin-proposal-do-expressions').default,
-      require('@babel/plugin-proposal-optional-chaining').default,
-      require('@babel/plugin-proposal-nullish-coalescing-operator').default,
       // Adds syntax support for import()
-      require('@babel/plugin-syntax-dynamic-import').default,
```
emmenko: Ah you removed this but remove the dependency of `babel-plugin-transform-dynamic-import`.

Which one should we remove?

tdeekens: Good point. The changelog is not the most expressive on this. 

CleanShot 2020-01-20 at 09 38 58@2x

Somehow it sounds to me as the support it natively now. I removed both. Will build and give it a spin locally and see what happens.

tdeekens: So the tests locally with this version of the preset pass. Also a production build can be made. Anything else besides the later canary build we should test before moving on?

emmenko: If everything passes, it should be fine

.vscode/settings.json
```diff
@@ -18,5 +18,8 @@
     }
   },
   "typescript.tsdk": "node_modules/typescript/lib",
-  "vscode-vale.path": "${workspaceFolder}/node_modules/.bin/commercetools-vale"
+  "vscode-vale.path": "${workspaceFolder}/node_modules/.bin/commercetools-vale",
+  "workbench.colorCustomizations": {
+    "tab.unfocusedActiveBorder": "#fff0"
```
emmenko: 😔

I guess we need to git ignore `settings.json` file again

tdeekens: FF*. Every lovely morning day. I commit the bloody settings.json here.

PR #1257, opened 2020-01-21
docs(changelog): for v15.8.0 [skip ci]

tdeekens:


tdeekens:
I would like to have the design team to be able to use subgroup targeting.

PR #1258, opened 2020-01-21
test(app-shell): migrate navbar specs to RTL

emmenko:


packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -130,28 +134,48 @@ MenuExpander.propTypes = {
...
-    )}
-  >
-    {props.children}
-  </ul>
-);
+export const MenuGroup = props => {
+  const isSublistActiveWhileIsMenuExpanded =
+    props.level === 2 && props.isActive && props.isExpanded;
+  const isSublistActiveWhileIsMenuCollapsed =
+    props.level === 2 && props.isActive && !props.isExpanded;
```
tdeekens: Nice, the rest just moved I suppose.

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -650,7 +678,7 @@ export class DataMenu extends React.PureComponent {
 
   render() {
     return (
-      <MenuGroup level={1}>
+      <MenuGroup id="main" level={1}>
```
tdeekens: What is the id used for? Tests? 👀

packages/application-shell/src/components/navbar/navbar.spec.js
```diff
@@ -1,1293 +1,4 @@
...
-        it('should update isForcedMenuOpen to true', () => {
-          expect(window.localStorage.setItem).toHaveBeenCalledWith(
-            STORAGE_KEYS.IS_FORCED_MENU_OPEN,
-            true
-          );
-        });
-      });
-    });
-  });
-});
```
tdeekens: Nostalgia: I spent so much time bolting e.g. permissions and menu visibilities in here 🐼

playground/menu.json
```diff
@@ -4,7 +4,27 @@
   "icon": "RocketIcon",
   "permissions": ["ManageDeveloperSettings", "ViewDeveloperSettings"],
   "featureToggle": null,
-  "submenu": [],
+  "submenu": [{
```
tdeekens: This is needed for the VRTs?

packages/application-shell/src/components/navbar/navbar.js
adnasa: would it make sense to move these numeric values into a const e.g `SUB_MENU_LEVEL`

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -633,3 +639,596 @@ describe('when dispatching a loading notification', () => {
...
+      );
+    });
+    // Check that the support link is rendered
+    // Get the nav container, to narrow down the search area
+    const container = await rendered.findByTestId('left-navigation');
+    const navbarRendered = within(container);
+    expect(navbarRendered.queryByText('Support')).toBeInTheDocument();
+  });
+});
+describe('when rendering navbar menu links from local config', () => {
```
tdeekens: Good to have this tested.

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -633,3 +639,596 @@ describe('when dispatching a loading notification', () => {
...
+      'true'
+    );
+    const menuGroup = within(rendered.getByTestId(groupId));
+    expect(menuGroup.queryByText(mainSubmenuLabel.value)).toBeInTheDocument();
+    expect(menuGroup.queryByText(mainSubmenuLabel.value)).not.toHaveAttribute(
+      'aria-current'
+    );
+    // Go to the link to check if the link becomes active
+    fireEvent.click(menuGroup.queryByText(mainSubmenuLabel.value));
+    await wait(() => {
```
adnasa: what is the reason for the `wait` here and not having that for L717?

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -633,3 +639,596 @@ describe('when dispatching a loading notification', () => {
...
+    });
+    // Get the nav container, to narrow down the search area
+    const container = await rendered.findByTestId('left-navigation');
+    const navbarRendered = within(container);
+
+    const applicationLocale = mcOperations.FetchLoggedInUser.me.language;
+    const mainMenuLabel = settingsOperations.FetchProjectExtensionsNavbar.projectExtension.applications[0].navbarMenu.labelAllLocales.find(
+      localized => localized.locale === applicationLocale
+    );
+    const mainSubmenuLabel = settingsOperations.FetchProjectExtensionsNavbar.projectExtension.applications[0].navbarMenu.submenu[0].labelAllLocales.find(
```
adnasa: this looks fragile. I'm not really sure how to tackle this otherwise however :(

packages/application-shell/src/components/navbar/navbar.js
```diff
@@ -130,28 +134,48 @@ MenuExpander.propTypes = {
...
+          [styles['sublist-no-children']]: props.level === 2 && !props.children,
+        },
+        {
+          [styles[
+            'sublist-expanded__active'
+          ]]: isSublistActiveWhileIsMenuExpanded,
+        },
+        {
+          [styles[
+            'sublist-collapsed__active'
```
adnasa: nit; what if we just remove the dash so this can be `styles.sublistCollapsed__active`?

sidenote: I'm not sure if I like BEM approach when using CSSmodules anymore

packages/application-shell/src/components/application-shell/application-shell.spec.js
tdeekens: ...would this be shorter maybe

```js
const renderedMainSubmenu = menuGroup.queryByText(mainSubmenuLabel.value);

fireEvent.click(renderedMainSubmenu);

await wait(() => {
      expect(renderedMainSubmenu).toHaveAttribute(
        'aria-current',
        'page'
      );
})
```

be shorter?

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -633,3 +639,596 @@ describe('when dispatching a loading notification', () => {
...
+    const rendered = renderApp(null, {
+      featureFlags: {
+        [PROJECT_EXTENSIONS]: false,
+      },
+      environment: {
+        servedByProxy: 'true',
+      },
+    });
+    // Get the nav container, to narrow down the search area
+    const container = await rendered.findByTestId('left-navigation');
```
tdeekens: Nit: maybe naming container a bit better would help understanding where we are in the DOM. Like `leftNavigationContainer` or something?

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -633,3 +639,596 @@ describe('when dispatching a loading notification', () => {
...
+    );
+    const menuGroup = within(rendered.getByTestId(groupId));
+    expect(menuGroup.queryByText(mainSubmenuLabel.value)).toBeInTheDocument();
+    expect(menuGroup.queryByText(mainSubmenuLabel.value)).not.toHaveAttribute(
+      'aria-current'
+    );
+    // Go to the link to check if the link becomes active
+    fireEvent.click(menuGroup.queryByText(mainSubmenuLabel.value));
+    await wait(() => {
+      expect(menuGroup.queryByText(mainSubmenuLabel.value)).toHaveAttribute(
```
tdeekens: Is there a reason why we always repeat the queryBy? Does calling it one not return a live view of the DOM node maybe?

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -633,3 +639,596 @@ describe('when dispatching a loading notification', () => {
...
+  it('should not render item', async () => {
+    const navbarMock = ApplicationNavbarMenuMock.build({
+      permissions: ['ViewOrders'],
+    });
+    const rendered = renderApp(null, {
+      featureFlags: {
+        [PROJECT_EXTENSIONS]: false,
+      },
+      DEV_ONLY__loadNavbarMenuConfig: () => Promise.resolve([navbarMock]),
+    });
```
tdeekens: Somehow, when reading this, I am starting to wonder if we not wastefully loose 10LoC on every it until it gets to something. Maybe this can be put into a `render` helper which just receives 1-2 args which is the variation?

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -633,3 +639,596 @@ describe('when dispatching a loading notification', () => {
...
+    // Get the nav container, to narrow down the search area
+    const container = await rendered.findByTestId('left-navigation');
+    const navbarRendered = within(container);
+
+    const applicationLocale = operations.FetchLoggedInUser.me.language;
+    const mainMenuLabel = navbarMock.labelAllLocales.find(
```
tdeekens: I "trick" I did the other day to keep tests shorter was to return e.g. `return { mainMenuLabel, renderedNavBar, ...rendered}` etc from `render`. This encapsulates things a bit to keep tests lean. That would maybe help avoiding repetition here too?

packages/application-shell/src/components/application-shell/application-shell.spec.js
emmenko: Hmm no reason, I can maybe await above as well. Thanks for pointing it out

packages/application-shell/src/components/application-shell/application-shell.spec.js
emmenko: Hmm I haven't tried that. Good point.

packages/application-shell/src/components/application-shell/application-shell.spec.js
emmenko: Hmm yeah, although if this fails it means that the data is wrong

packages/application-shell/src/components/application-shell/application-shell.spec.js
emmenko: I though about this but decided to go with being more explicit and have things duplicated. I can have another look to see if some stuff can be simplified

packages/application-shell/src/components/navbar/navbar.js
emmenko: Yes, I extracted those into meaningful variables, as I'm using them for the `aria-expanded` attribute

emmenko: Yeah, but this navbar component needs to be rewritten to be honest. I don't want to start refactoring too much unless we actively work on it.

packages/application-shell/src/components/navbar/navbar.js
emmenko: Yeah, like I said above the navbar component needs to be rewritte, so we can get rid of all this stuff once we do that.

packages/application-shell/src/components/navbar/navbar.js
emmenko: No the id is used to reference the element using `aria-owns`

https://developers.google.com/web/fundamentals/accessibility/semantics-aria/aria-labels-and-relationships#aria-owns

playground/menu.json
emmenko: Ah sorry I was testing something...I forgot to remove it

packages/application-shell/src/components/application-shell/application-shell.spec.js
adnasa: hmm.. if there is no reason to do so here and above, then don't :)
my intention was to understand the workflow of the menu that requires that we use `wait`

packages/application-shell/src/components/application-shell/application-shell.spec.js
tdeekens: Yeah makes total sense. Always a hard balance how much to extract into test setup at the risk of having too much complexity there.

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -633,3 +637,554 @@ describe('when dispatching a loading notification', () => {
...
+    });
+    // Check that the support link is rendered
+    // Get the nav container, to narrow down the search area
+    const container = await rendered.findByTestId('left-navigation');
+    const navbarRendered = within(container);
+    expect(navbarRendered.queryByText('Support')).toBeInTheDocument();
+  });
+});
+describe('navbar menu links interactions', () => {
+  async function checkLinksInteractions(
```
emmenko: This function now encapsulates the interaction logic of the 3 tests below, so it's less duplication now.

PR #1259, opened 2020-01-21
fix(permissions): data fences to be overwritten by manage resource permission

tdeekens:
## Summary

This pull request fixes an issue when a data fence is configured while the corresponding resource access permission should overwrite it.

## Description

This case is best described in ACs

### Status Quo

```
In the permission configuration UI
Given I configure a resource access (e.g. orders) 
And I grant access to manage orders
Then any condition will be overruled by the general permission
```

```
Given I configure a resource access (e.g. orders)
And I grant access to view orders
Then I can add a condition for a specific store to allow managing orders in this store
```
 
In the evaluation of permissions

```
Given I configure a resource access (e.g. orders) 
And I grant access to manage orders
Then the condition will always gain precedence over the general permission
```

### Desired state

```
Given I configure a resource access to manage (e.g. orders)
And I have an existing condition configured too
Then the resource access to manage should take precedence
```

```
Given I configure a resource access to view (e.g. orders)
And I have an existing condition configured to manage a store
Then the condition to manage should take precedence
```

These pairings become particularily obvious with action rights and data fences.

```
Given I configure manage orders
And I configure a action right to not edit prices
And I configure a store to manage orders
```

In this case the data fence has to be disregarded (but is saved) and we fall back to evaluating the resource access and action right.

During this PR I fixed a lot of naming and test structure. Commit by commit review should be possible.

### DIFF
--- a/packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
+++ b/packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
@@ -282,7 +282,7 @@ describe('action rights', () => {
         expect(rendered.queryByText('Is authorized: No')).toBeInTheDocument();
       });
     });
-    describe('with applied action right not matching demanded', () => {
+    describe('with applied action right matching demanded', () => {
       it('should indicate being authorized', () => {
         const rendered = render({
           allAppliedPermissions: [
@@ -328,7 +328,7 @@ describe('action rights', () => {
     });
   });
   describe('with applied and none matching demanded view permission', () => {
-    describe('with applied action right matching demanded', () => {
+    describe('with applied action right not matching demanded', () => {
       it('should indicate not being authorized', () => {
         const rendered = render({
           allAppliedPermissions: [{ name: 'canViewProducts', value: false }],
--- a/packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
+++ b/packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
@@ -375,10 +375,56 @@ describe('action rights', () => {
 describe('data fences', () => {
   describe('when general permission is matched', () => {
     describe('when data fence is matched', () => {
-      it('should indicate as authorized', () => {
+      describe('when not overwritten', () => {
+        it('should indicate as authorized', () => {
+          const rendered = render({
+            shouldMatchSomePermissions: true,
+            allAppliedPermissions: [{ name: 'canManageOrders', value: true }],
+            allAppliedDataFences: [
+              {
+                __typename: 'StoreDataFence',
+                type: 'store',
+                group: 'orders',
+                name: 'canViewOrders',
+                value: 'store-1',
+              },
+            ],
+            demandedPermissions: ['ManageOrders'],
+            demandedDataFences: [
+              {
+                type: 'store',
+                group: 'orders',
+                name: 'ViewOrders',
+              },
+            ],
+            selectDataFenceData: ({ type }) => {
+              switch (type) {
+                case 'store':
+                  return ['store-1'];
+                default:
+                  return null;
+              }
+            },
+          });
+
+          expect(
+            rendered.queryByText('Is authorized: Yes')
+          ).toBeInTheDocument();
+        });
+      });
+    });
+    describe('when overwritten', () => {
+      it('should indicate as not authorized', () => {
         const rendered = render({
           shouldMatchSomePermissions: true,
           allAppliedPermissions: [{ name: 'canManageOrders', value: true }],
+          allAppliedActionRights: [
+            {
+              group: 'products',
+              name: 'canEditPrices',
+              value: false,
+            },
+          ],
           allAppliedDataFences: [
             {
               __typename: 'StoreDataFence',
@@ -389,6 +435,9 @@ describe('data fences', () => {
             },
           ],
           demandedPermissions: ['ManageOrders'],
+          demandedActionRights: [
+            { group: 'products', name: 'PublishProducts' },
+          ],
           demandedDataFences: [
             {
               type: 'store',
@@ -406,7 +455,7 @@ describe('data fences', () => {
           },
         });
 
-        expect(rendered.queryByText('Is authorized: Yes')).toBeInTheDocument();
+        expect(rendered.queryByText('Is authorized: No')).toBeInTheDocument();
       });
     });
     describe('when data fence is not matched', () => {
--- a/packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
+++ b/packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { ApplicationContextProvider } from '@commercetools-frontend/application-shell-connectors';
-import { render } from '@testing-library/react';
+import { render as rtlRender } from '@testing-library/react';
 import useIsAuthorized from './use-is-authorized';
 
 type TPermissionName = string;
@@ -62,7 +62,7 @@ const TestComponent = (props: TestProps) => {
   );
 };
 
-const testRender = ({
+const render = ({
   demandedPermissions,
   demandedActionRights,
   demandedDataFences,
@@ -83,7 +83,7 @@ const testRender = ({
   allAppliedDataFences?: TAllAppliedDataFence[];
   shouldMatchSomePermissions?: boolean;
 }) =>
-  render(
+  rtlRender(
     <ApplicationContextProvider
       project={{
         key: 'test-with-big-data',
@@ -130,11 +130,11 @@ const testRender = ({
     </ApplicationContextProvider>
   );
 
-describe('when only one permission matches', () => {
-  describe('General Permissions', () => {
-    describe('when "shouldMatchSomePermissions=true"', () => {
+describe('general permissions', () => {
+  describe('when only some permission should be matched', () => {
+    describe('with one applied permission and one matching demanded permission', () => {
       it('should indicate being authorized', () => {
-        const { queryByText } = testRender({
+        const rendered = render({
           demandedPermissions: ['ManageCustomers', 'ManageOrders'],
           allAppliedPermissions: [
             {
@@ -145,12 +145,30 @@ describe('when only one permission matches', () => {
           shouldMatchSomePermissions: true,
         });
 
-        expect(queryByText('Is authorized: Yes')).toBeInTheDocument();
+        expect(rendered.queryByText('Is authorized: Yes')).toBeInTheDocument();
       });
     });
-    describe('when "shouldMatchSomePermissions=false"', () => {
+    describe('with one applied permission and no matching demanded permission', () => {
+      it('should indicate not being authorized', () => {
+        const rendered = render({
+          demandedPermissions: ['ManageCustomers', 'ManageOrders'],
+          allAppliedPermissions: [
+            {
+              name: 'canManageCustomersGroups',
+              value: true,
+            },
+          ],
+          shouldMatchSomePermissions: true,
+        });
+
+        expect(rendered.queryByText('Is authorized: No')).toBeInTheDocument();
+      });
+    });
+  });
+  describe('when all permissions should be matched', () => {
+    describe('with one applied permission and one not matching demanded permission', () => {
       it('should indicate being not authorized', () => {
-        const { queryByText } = testRender({
+        const rendered = render({
           allAppliedPermissions: [
             {
               name: 'canManageCustomers',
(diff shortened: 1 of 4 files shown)


packages/permissions/src/hooks/use-is-authorized/use-is-authorized.ts
```diff
@@ -4,7 +4,7 @@ import {
   hasSomePermissions,
   hasEveryPermissions,
   hasEveryActionRight,
-  hasAppliedDataFence,
+  hasSomeDataFence,
```
tdeekens: This is to follow the naming patterns we have already and communicate what this function does: finding if any a match of one data fence.

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
```diff
@@ -130,11 +130,11 @@ const testRender = ({
     </ApplicationContextProvider>
   );
 
-describe('when only one permission matches', () => {
-  describe('General Permissions', () => {
-    describe('when "shouldMatchSomePermissions=true"', () => {
+describe('general permissions', () => {
+  describe('when only some permission should be matched', () => {
```
tdeekens: The nesting is mostly shuffled a bit in the first commits to not mix things.

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -79,18 +81,32 @@ const hasExactPermission = (
...
+  demandedPermission.startsWith('Manage');
+
 // Check that the user permissions match the required MANAGE permission.
 // The shapes of the arguments are:
 // - demandedPermission:
 //     'ViewProducts'
 // - actualPermissions:
 //     { canViewProducts: true, canManageOrders: false }
-const hasManagePermission = (
+const doesManagePermissionInferViewPermission = (
```
tdeekens: This function name and denseness confused me for too long :)

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -190,11 +209,14 @@ export const getInvalidPermissions = (
   );
 };
 
-const hasDemandedDataFence = (options: {
+type TGetHasDemandedDataFenceOptions = {
```
tdeekens: Just moved out of the args list.

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -224,23 +246,49 @@ const hasDemandedDataFence = (options: {
...
 };
 
-const getDataFenceByPermissionGroup = (
-  dataFences: TDataFences | null,
-  storeKey: TDataFenceType,
-  groupKey: string
+const getDataFenceByTypeAndGroup = (
+  actualDataFences: TDataFences | null,
+  demandedDataFenceType: TDataFenceType,
+  demandedDataFenceGroup: string
```
tdeekens: I found the arg naming here to be out of sync between caller and callee.

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -270,16 +318,25 @@ export const hasAppliedDataFence = (options: TOptionsForAppliedDataFence) => {
...
+        demandedDataFence
+      );
+
+      const hasDemandedDataFence = getHasDemandedDataFence({
+        actualDataFence,
         demandedDataFence,
         selectDataFenceData: options.selectDataFenceData,
       });
+
+      return !isPermissionOverwritingDataFence && hasDemandedDataFence;
```
tdeekens: Note, we return false here now in some cases. This makes the `useIsAuthorized` fall through to the resource access and action right evaluation.

packages/permissions/src/utils/has-permissions.spec.tsx
```diff
@@ -207,11 +207,12 @@ describe('getInvalidPermissions', () => {
...
 });
 
-describe('hasAppliedDataFence', () => {
+describe('hasSomeDataFence', () => {
   describe('user has not datafence permissions', () => {
     it('should return false', () => {
       expect(
-        hasAppliedDataFence({
+        hasSomeDataFence({
+          actualPermissions: null,
```
tdeekens: Just due to the type change.

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.ts
```diff
@@ -92,7 +91,8 @@ const useIsAuthorized = ({
         )
       );
     }
-    hasDemandeDataFences = hasAppliedDataFence({
+    hasDemandeDataFences = hasSomeDataFence({
+      actualPermissions,
```
tdeekens: We need this now to match up against the acutal permission using the demanded data fences.

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
```diff
@@ -323,230 +407,277 @@ describe('when only one permission matches', () => {
...
-            ],
-            selectDataFenceData: ({ type }) => {
-              switch (type) {
-                case 'store':
-                  return ['store-1'];
-                default:
-                  return null;
-              }
+    });
+    describe('when overwritten', () => {
```
tdeekens: This is the new case.

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -270,16 +316,25 @@ export const hasAppliedDataFence = (options: TOptionsForAppliedDataFence) => {
...
 
       const [dataFenceName, dataFenceValue] = specificActualDataFence;
-      return hasDemandedDataFence({
-        actualDataFence: {
-          name: dataFenceName,
-          // we do the type casting because at this point we are sure that
-          // specificActualDataFence.dataFenceValue is not null
-          dataFenceValue: dataFenceValue as TActualDataFence['dataFenceValue'],
-        },
+      const actualDataFence = {
```
tdeekens: I just moved this up and gave it the name.

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -224,23 +244,49 @@ const hasDemandedDataFence = (options: {
...
+   * NOTE:
+   *    A data fence relates, exists in relation to a resource access.
+   *    This relation is constructed by the group of the data fence.
+   *
+   *    Given the user already has manage access on the data fence's group
+   *     Then the data fence is overruled by it.
+   *    Given the user _does not_ have manage access on the data fence's group
+   *     Then the data fence itself takes precedence
+   */
+  const demandedPermissionByDataFence = `Manage${upperFirst(
```
tdeekens: This is the crutch to make the relation from the datafence to it's "host".

tdeekens:
Basically I think our API here is a bit crooked I think. We don't have to or should allow specifying multiple permissions. For instance

demandedResourceAccess: ManageCustomers, ViewOrders
demandedActionRights: [EditPrices, orders]
dataFence: [store, ManageCustomers, store-1]
shouldMatchSomePermission=true

There is likely no way the above example has any meaninigful output. I think eventually we should educe resource access to singlular form and remove shouldMatchSome (used 4-5 times in MC).

However, we the solution now works and manages to correcly match each data fence to a respective resource access and allow it to overwite the data fence itself. In the future we might need action rights in data fences and then this API starts to be challenged even more.

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
```diff
@@ -145,155 +145,239 @@ describe('when only one permission matches', () => {
...
+          demandedPermissions: ['ViewCustomersGroups'],
+        });
+
+        expect(rendered.queryByText('Is authorized: No')).toBeInTheDocument();
+      });
+    });
+  });
+});
+
+describe('action rights', () => {
```
emmenko: Wondering if we should write the describes using the ACs. Would make the specs easier to read maybe? 🤔

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
```diff
@@ -145,155 +145,239 @@ describe('when only one permission matches', () => {
...
+          demandedActionRights: [
+            { group: 'products', name: 'PublishProducts' },
+          ],
+          shouldMatchSomePermissions: true,
+        });
+
+        expect(rendered.queryByText('Is authorized: No')).toBeInTheDocument();
+      });
+    });
+    describe('with applied action right not matching demanded', () => {
```
emmenko:
```suggestion
    describe('with applied action right matching demanded', () => {
```

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
```diff
@@ -145,155 +145,239 @@ describe('when only one permission matches', () => {
...
+          shouldMatchSomePermissions: true,
+        });
 
-          expect(queryByText('Is authorized: No')).toBeInTheDocument();
+        expect(rendered.queryByText('Is authorized: No')).toBeInTheDocument();
+      });
+    });
+  });
+  describe('with applied and none matching demanded view permission', () => {
+    describe('with applied action right matching demanded', () => {
```
emmenko:
```suggestion
    describe('with applied action right not matching demanded', () => {
```

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -79,18 +81,30 @@ const hasExactPermission = (
...
   actualPermissions: TPermissions
-) =>
-  demandedPermission.startsWith('View') &&
-  actualPermissions[toCanCase(demandedPermission.replace('View', 'Manage'))];
+) => {
+  const isDemandedPermissionViewPermission = getIsViewPermission(
+    demandedPermission
+  );
+  const isViewPermissionInferredByManagePermission =
+    actualPermissions[toCanCase(demandedPermission.replace('View', 'Manage'))];
```
emmenko:
```suggestion
    Boolean(actualPermissions[toCanCase(demandedPermission.replace('View', 'Manage'))]);
```

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
tdeekens: Yeah I felt the urge too at some point. However, didn’t want to mix changing code and tests too much in a single PR. If we decide to to that I would take that on on another PR so things are easier to trace and this isn’t the „when they changed too much PR and things broke“ 🤪🧚🏻

adnasa: ye if that makes things easier for you, you can split this into a separate step

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
```diff
@@ -375,10 +375,56 @@ describe('action rights', () => {
...
+                case 'store':
+                  return ['store-1'];
+                default:
+                  return null;
+              }
+            },
+          });
+
+          expect(
+            rendered.queryByText('Is authorized: Yes')
```
adnasa: now this is where things become confusing.
IMO data-fence is overwritten here, since the general permission is `manage` which matches the demanded.

additionally in this specific test case, when indicating `is authorized`, there is no indication if what we authorized is the a match of `demanded=manage, applied=manage` or `demanded=view, applied=view`

--- 

our goal is to update the convention. "override" is now amended to take `action-right` in account

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -241,25 +247,44 @@ const getHasDemandedDataFence = (options: {
...
+  demandedDataFence: TDemandedDataFence
 ) => {
+  if (!actualPermissions) return false;
+
+  /**
+   * NOTE:
+   *    A data fence relates, exists in relation to a resource access.
+   *    This relation is constructed by the group of the data fence.
+   *
+   *    Given the user already has manage access on the data fence's group
```
adnasa: altho I understand you, we are changing the language.

I would stick to using `general permission` and `datafence permission` since all our docs are using these terms.

```
Given user has general permission `manage` on the data fence's group
```

tdeekens:
@emmenko and @adnasa regarding more of a UAT style pattern for the next PR to align early. Is that roughly what you had in mind (cases missing):

```txt
with a single resource access permission
    without an action right
        when resource access is matched
            it should indicate being authorized
        when resource access is not matched
            it should indicate being unauthorized
    with a single action right
        when resource access is matched
            when action right is matched
                it should indicate being authorized
            when action right is not matched
                it should indicate being unauthorized
        when resource access is not matched
            when action right is matched
                it should indicate being unauthorized
    with multiple action rights
        when resource access is matched
            when all action rights are matched
                it should indicate being authorized
        when resource access is not matched
            when one action right is not matched
                it should indicate being unauthorized
        when resource access is not matched
            when all action rights are matched
                it should indicate being unauthorized
with a multiple resource access permission
    without an action right
        when all resource accesses are matched
            it should indicate being authorized
        when a single resource access is not matched
            when all resource accesses should be matched
                it should indicate being unauthorized
            when at least one resource accesses should be matched
                it should indicate being authorized
```

emmenko:
Hmm I was thinking more of something like this (only 1 level nesting):

```js
describe(`
  Given I configure a resource access to manage (e.g. orders)
  And I have an existing condition configured too
  Then the resource access to manage should take precedence
`, () => {})
```

tdeekens:
I see. Yeah, also makes sense. Just works a bit against having all permuations somehow. The nesting makes that a bit more "obvious" where you missed a combination of things. At least for me when writing them.

emmenko:
With the nesting I found myself having to read it 2-3 times to trying to understand what the combination should be...I thought that maybe having it written as an AC might make it easier to understand the intention and what to expect as a result.

tdeekens:
Ok, I'd like to finish this PR and then would give the combined/longer version a shot.

emmenko:
Sure thing 💯

adnasa:
I'm for the nesting, since this works in combinations.
if combination fails then one can work/debug with it in isolation.

Doing the `all in one test()` requires a dive not only to the single failing test, but all others as well since they also use the combinations.

---

sidenote: I still don't think we should call general permission as `resource access` because this is field used in a data fence permission as well

emmenko:
> Doing the all in one test() requires a dive not only to the single failing test, but all others as well since they also use the combinations.

I'm not sure what you mean but what I suggested was to still have one describe for each test, not one describe for multiple tests. The only difference is that the nested describes are put together using a more readable description.

tdeekens:
> I'm not sure what you mean but what I suggested was to still have one describe for each test, not one describe for multiple tests. The only difference is that the nested describes are put together using a more readable description.


I think what @emmenko suggests is sort of a `git squash` or nested describes. At least that's what I have in mind. We have all the permutations but flat in one "commit"/describe.

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -255,6 +256,13 @@ const getDataFenceByTypeAndGroup = (
   return null;
 };
 
+const getIsPermissionOverwritingDataFence = (
+  actualPermissions: TPermissions | null,
+  actualDataFence: TActualDataFence
+) => {
+  return false;
```
amine-benselim: hmmmm. this is always returning false 🤔

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
amine-benselim: I really don't understand this new test, so if:
1- you have datafence permissions
2- you also have a general permission, but not the action right
then `isAuthorized` would return `false`?

packages/permissions/src/utils/has-permissions.ts
amine-benselim: nevermind this I was reviewing commit by commit 😓

tdeekens: This is my story: bolt something in, make it not change anything. Adjust behaviour :)

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.spec.tsx
tdeekens: Let me try:

So the story is you have a data fence configured. However, this data fence is practially not valid. This is cause you have manage access on the resource permission level. The mental modal is: it's additive from general permission to data fences. Not substractive. Saying you can't have: manage orders (resource access/general permission) and take away manage right to only view orders in a specific store. What you have to do instead is to set the resource access to view and then add manage to the stores the team can manage in the data fence.

This means that
Having set manage orders as general permission
Having set manage orders as data fence in e.g. store-1
-> These two mean the same and doesn't impact anything
Having set add discount to false as action right 
-> This now makes it contradicting here the action right and general permission should take precedence. That is cause the data fence itself is not valid in combination with the resource access. You already said: all orders can be managed - no matter the store. There is no need to say store-1 can also manage orders. It's implied by the general permission already.

Does that help? @adnasa does that make sense?

tdeekens: We can also have a call if that helps! Just ping me in Slack :)

adnasa:
>  sort of a git squash or nested describes

sounds good to me

tdeekens:
Okay. So I also had a call with @amine-benselim and yesterday @adnasa. I think we all agree that this is wonky but somehow needed. Also we all seem to agree that changing the tests isn't for this PR. So I'd ask for reviews :)

packages/permissions/src/utils/has-permissions.ts
adnasa: do you want to address this first before merging?

tdeekens:
> do you want to address this first before merging?

BTW: was addressed. :)

PR #1260, opened 2020-01-22
refactor(app-shell): deprecate MeasureFirstPaint

emmenko:
Ref #1248 

Follow up of #1249


packages/permissions/package.json
```diff
@@ -41,8 +41,7 @@
     "@commercetools-frontend/application-shell-connectors": "15.8.0",
     "lodash": "4.17.15",
     "lodash-es": "4.17.15",
-    "prop-types": "15.7.2",
-    "warning": "4.0.3"
```
tdeekens: I will add tiny-invariant. At least I am working on a PR to do so :).

PR #1261, opened 2020-01-22
chore(permissions): add warnings to propose improvements

tdeekens:
#### Summary

This pull request iterates on observations and discusisons from the previous pull request by adding warnings to guide the user to a more appripriate usage of the permissions package.

#### Description

There are some cases we seem to all agree on that they're dangerous:

1. Multiple demanded permissions: you can and should just use one and compose yourselr
2. `shouldMatchSomePermissions`: leads to unexpected behaviour when used with e.g. action rights and should be avoided
3. Multiple action rights: should also be avoided and multiple usages of the hook (or if used as HoC) should be preferred
4. Implied permissions: This should not happen when using a single demanded permission. Howeever, as we're only logging suggestions here I propsoe to inform the user about uncalled-for usage. So `['ManageOrders', 'ViewOrders']` should just be `['ViewOrders']` (or in future `'ManageOrders`) while `ViewOrders` is implied by `ManageOrders`.

While on this I recommend us using a babel plugin to remove these warnings in production bundles. Also to avoid: if ever using `tiny-invariant` again (remember?) the app to unmount.

### DIFF
--- a/packages/jest-preset-mc-app/setup-tests.js
+++ b/packages/jest-preset-mc-app/setup-tests.js
@@ -10,18 +10,27 @@ global.window.app = {
 
 const shouldSilenceWarnings = (...messages) =>
   [
-    /Warning: componentWillReceiveProps has been renamed/,
-    /CellMeasurerCache should only measure a cell's width or height/,
+    /.*Warning: componentWillReceiveProps has been renamed.*/,
+    /.*CellMeasurerCache should only measure a cell's width or height.*/,
   ].some(msgRegex => messages.some(msg => msgRegex.test(msg)));
+
+const shouldNotThrowWarnings = (...messages) =>
+  [/.*@commercetools-frontend\/permissions.*/].some(msgRegex =>
+    messages.some(msg => msgRegex.test(msg))
+  );
 
 // setup file
 const logOrThrow = (log, method, messages) => {
   const warning = `console.${method} calls not allowed in tests`;
   if (process.env.CI) {
-    if (shouldSilenceWarnings(messages)) {
-      return;
-    }
+    if (shouldSilenceWarnings(messages)) return;
+
     log(warning, '\n', ...messages);
+
+    // NOTE: That some warnings should be logged allowing us to refactor graceully
+    // without having to introduce a breaking change.
+    if (shouldNotThrowWarnings(messages)) return;
+
     throw new Error(...messages);
   } else {
     log(colors.bgYellow.black(' WARN '), warning, '\n', ...messages);
--- a/packages/permissions/src/hooks/use-is-authorized/use-is-authorized.ts
+++ b/packages/permissions/src/hooks/use-is-authorized/use-is-authorized.ts
@@ -83,7 +83,7 @@ const useIsAuthorized = ({
   );
   warning(
     shouldMatchSomePermissions === false,
-    `@commercetools-frontend/permissions: It is recommended not to use 'shouldMatchSomePermissions' but instead use the hook, HoC or comoponent multiple times.`
+    `@commercetools-frontend/permissions: It is recommended not to use 'shouldMatchSomePermissions' but instead use the hook, HoC or component multiple times.`
   );
   warning(
     !impliedPermissions || impliedPermissions.length === 0,
--- a/packages/permissions/src/hooks/use-is-authorized/use-is-authorized.ts
+++ b/packages/permissions/src/hooks/use-is-authorized/use-is-authorized.ts
@@ -1,3 +1,4 @@
+import warning from 'tiny-warning';
 import { useApplicationContext } from '@commercetools-frontend/application-shell-connectors';
 import { reportErrorToSentry } from '@commercetools-frontend/sentry';
 import {
@@ -5,6 +6,7 @@ import {
   hasEveryPermissions,
   hasEveryActionRight,
   hasSomeDataFence,
+  getImpliedPermissions,
 } from '../../utils/has-permissions';
 
 // Permissions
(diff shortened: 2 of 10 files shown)


packages/permissions/src/hooks/use-is-authorized/use-is-authorized.ts
```diff
@@ -69,6 +71,27 @@ const useIsAuthorized = ({
...
+    !demandedActionRights || demandedActionRights.length === 1,
+    `@commercetools-frontend/permissions: It is recommended to pass a single demanded action right while using the hook, HoC or component multiple times.`
+  );
+  warning(
+    !demandedPermissions || demandedPermissions.length === 1,
+    `@commercetools-frontend/permissions: It is recommended to pass a single demanded permission while using the hook, HoC or component multiple times.`
+  );
+  warning(
+    shouldMatchSomePermissions === false,
+    `@commercetools-frontend/permissions: It is recommended not to use 'shouldMatchSomePermissions' but instead use the hook, HoC or comoponent multiple times.`
```
emmenko:
```suggestion
    `@commercetools-frontend/permissions: It is recommended not to use 'shouldMatchSomePermissions' but instead use the hook, HoC or component multiple times.`
```

packages/permissions/README.md
```diff
@@ -211,22 +207,18 @@ match, otherwise a fallback component.
...
-);
+const Dashboard = () => {
+  const isAuthorized = useIsAuthorized({
+    demandedPermissions: ['ViewProducts'],
+  });
+
+  return (
+    <div>
+      {isAuthoried ? <TopFiveProducts /> : <Unauthorized />}
+      <RevenueChart isDisabled={!isAuthorized} />}
```
adnasa: `ViewOrders` is missing

tdeekens: Good point. Thanks.

packages/permissions/README.md
```diff
@@ -211,22 +207,21 @@ match, otherwise a fallback component.
...
+    demandedPermissions: ['ViewProducts'],
+  });
+  const canViewOrders = useIsAuthorized({
+    demandedPermissions: ['ViewOrders'],
+  });
+
+  return (
+    <div>
+      {canViewProducts ? <TopFiveProducts /> : <Unauthorized />}
+      <RevenueChart isDisabled={!canViewOrders} />}
```
adnasa: `!canViewOrders || !canViewProducts`

tdeekens: 🤦‍♂

.circleci/config.yml
```diff
@@ -7,14 +7,14 @@ aliases:
     restore_cache:
       name: "Restoring yarn cache"
       keys:
-        - v2-yarn-cache-{{ .Branch }}-{{ checksum "yarn.lock" }}
-        - v2-yarn-cache-{{ .Branch }}
-        - v2-yarn-cache
+        - v3-yarn-cache-{{ .Branch }}-{{ checksum "yarn.lock" }}
```
emmenko: What happened? 🤔

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.ts
emmenko: PS: still a typo 👆

.circleci/config.yml
tdeekens: Assumed CI‘s cache got choked. It is an odd react-intl issue. Will check later.

emmenko:
We should include this in v16 #1248 since it will throw an error on CI because of the `console.warn` logs, so it's kind of a "breaking change".

tdeekens:
Agreed. Thanks for looking into it! I will extract the improvements made here into another PR while keeping this on hold.

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.ts
```diff
@@ -69,6 +71,27 @@ const useIsAuthorized = ({
   selectDataFenceData?: TSelectDataFenceData;
   shouldMatchSomePermissions?: boolean;
 }) => {
+  const impliedPermissions = getImpliedPermissions(demandedPermissions);
+
+  warning(
```
emmenko: It would be nice if we can provide more "context" in the warning message, that makes it easier to identify the component where things need to be changed.

Maybe trying to get the stack trace and include it in the message

tdeekens:
This now only contains the breaking changes.

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.ts
tdeekens: Yes, that's why I initially wanted to use React.warn but that's not part of React anymore. I don't really see an easy way to get a hand on the component stack trace here. I would didn't check what the log looks like. Formik uses the same way to do these things if I recall correctly.

packages/jest-preset-mc-app/setup-tests.js
```diff
@@ -10,18 +10,27 @@ global.window.app = {
...
-    if (shouldSilenceWarnings(messages)) {
-      return;
-    }
+    if (shouldSilenceWarnings(messages)) return;
+
     log(warning, '\n', ...messages);
+
+    // NOTE: That some warnings should be logged allowing us to refactor graceully
+    // without having to introduce a breaking change.
+    if (shouldNotThrowWarnings(messages)) return;
```
tdeekens: What do you think of this @emmenko and @adnasa.

Reasoning being: we want to allow graceful migration (yes, at the cost of slowing things down). However, we promote those warnigns after being logged to errors. So in a non-breaking release we can always whitelist logs. Then say in the next major we will promote those if we can.

Actually in this case the `shouldMatchSomePermissions` will be very hard cause it's wired into the menu and all.

packages/jest-preset-mc-app/setup-tests.js
```diff
@@ -10,18 +10,27 @@ global.window.app = {
 
 const shouldSilenceWarnings = (...messages) =>
   [
-    /Warning: componentWillReceiveProps has been renamed/,
-    /CellMeasurerCache should only measure a cell's width or height/,
+    /.*Warning: componentWillReceiveProps has been renamed.*/,
+    /.*CellMeasurerCache should only measure a cell's width or height.*/,
```
tdeekens: @ahmehri I think this is why it was never silenced in the first place. Cause it starts with warning and contains more stuff after.

packages/jest-preset-mc-app/setup-tests.js
emmenko: Good idea!

packages/jest-preset-mc-app/fail-tests-because-there-was-an-unhandled-rejection.lock
```diff
@@ -0,0 +1 @@
+undefined
```
emmenko: This file should be ignored

tdeekens: Good point. Amended into the last commit: ab81f42

tdeekens:
Yes, it will only log things. During test. In the browser but not on prod (that's what the babel plugin sorts out) as it prepares the code to be removed by the minifier.

packages/jest-preset-mc-app/setup-tests.js
ahmehri: Actually it was silenced and it worked in many cases, not sure why it didn't work for you though.

packages/jest-preset-mc-app/setup-tests.js
adnasa: very nice!

PR #1263, opened 2020-01-22
chore: update jest

tdeekens:
#### Summary

I planned to release today and wanted to update things accordingly.

#### Description

Jest comes with some nice improvements in performance which I'd hope our MC FE CI chains to be speedier.

https://jestjs.io/blog/2020/01/21/jest-25

However, it also comes with us (here at least) logging some things:

CleanShot 2020-01-22 at 18 23 01@2x

which is a cool idea, should be fine and resolve itself.

Then I also took a shot at updating to jsdom v16 cause we could use it as we require node 12. Jest keeps node 8 support to be nice even though it's EOL.

CleanShot 2020-01-22 at 18 21 37@2x

However, it changes a bit how we can fiddle with the window

CleanShot 2020-01-22 at 18 20 45@2x

so I would do that in another step maybe.

Note that none of jest's breaking changes affect us.

CleanShot 2020-01-22 at 17 41 51@2x


jest.d.ts
```diff
@@ -7,7 +7,6 @@ declare namespace jest {
     toRender(component: React.ComponentType): R;
     toHaveState(selector: string, expected: unknown): R;
     // The following matchers are part of https://www.npmjs.com/package/jest-dom
-    toBeInTheDocument: () => R;
```
tdeekens: Now comes for free as far as I can tell.

CleanShot 2020-01-22 at 18 17 39@2x

package.json
```diff
@@ -128,14 +128,13 @@
     "eslint-plugin-prettier": "3.1.2",
     "eslint-plugin-react": "7.18.0",
     "eslint-plugin-react-hooks": "2.3.0",
-    "eslint-plugin-testing-library": "1.4.1",
+    "eslint-plugin-testing-library": "1.5.0",
```
emmenko: I forgot to do it last time, we need to explicitly enable `prefer-expect-query-by` now, as it does not come enabled by default anymore: https://github.com/testing-library/eslint-plugin-testing-library/releases/tag/v1.4.1

More on the reasons here: https://github.com/testing-library/eslint-plugin-testing-library/issues/59

application-templates/starter/package.json
```diff
@@ -38,6 +38,6 @@
     "@commercetools-frontend/mc-scripts": "15.8.0",
     "@testing-library/react": "9.4.0",
     "enzyme": "3.11.0",
-    "jest": "24.9.0"
+    "jest": "25.1.0"
```
emmenko: Should we consider this a breaking change? Cause the jest peer dep also needs to be bumped.

jest.d.ts
emmenko: Ah, good to know!

package.json
emmenko: FYI: we're replacing that rule with a new one, targeted only for cases when you want to assert that an element is not in the document.

https://github.com/testing-library/eslint-plugin-testing-library/pull/65

tdeekens: A good to now. Thanks for sharing.

tdeekens: Ah, interesting. Read up on the background now. Makes sense what's in store there :)

application-templates/starter/package.json
tdeekens: Thanks for raising. I think to me it's not a breaking change. Only node 6 support is dropped and we require node 10/12 in other cases already. All other breaking changes I see are dependent to how people use jest within our peer dependency range. So if somebody imported from `@jest/core` or `jest-cli` and that now changed is not something we can make guarantees about. If their remain to have 24 installed nothing would break for them in anyway. Do you agree or have other concerns about this?

emmenko: Yeah if we support both version it should be fine like this. Let's maybe mention it in the changelog that people can start using jest 25.

PR #1267, opened 2020-01-23
refactor: fix and improve docs and usage of permissions

tdeekens:
#### Summary

This pull request extracts items from #1261 to stop proposing/using to be deprecated APIs in the docs.

### DIFF
--- a/application-templates/starter/menu.json
+++ b/application-templates/starter/menu.json
@@ -2,7 +2,7 @@
   "key": "examples-starter",
   "uriPath": "examples-starter",
   "icon": "RocketIcon",
-  "permissions": ["ViewProducts", "ManageProducts"],
+  "permissions": ["ViewProducts"],
   "featureToggle": null,
   "submenu": [],
   "labelAllLocales": [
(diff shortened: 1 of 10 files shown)


tdeekens:
Note that this code changes have already been approoved.

packages/application-shell/src/components/quick-access/create-commands.js
```diff
@@ -18,7 +18,7 @@ export default function createCommands({
       applicationContext.permissions &&
       featureToggles.canViewDashboard &&
       hasSomePermissions(
```
tdeekens: hasSomePermissions calls hasPermission which determines itself if the passed permissions (e.g. a view) is implied by a manage.

application-templates/starter/menu.json
```diff
@@ -2,7 +2,7 @@
   "key": "examples-starter",
   "uriPath": "examples-starter",
   "icon": "RocketIcon",
-  "permissions": ["ViewProducts", "ManageProducts"],
+  "permissions": ["ViewProducts"],
```
tdeekens: Same here. This is implied and doesn't need to be added.

PR #1269, opened 2020-01-23
feat(components): rework ModalPages to have more controls customization

jonnybel:
Updated version of #1008. See the discussion there.
It was better to create a new PR than to try to resolve all conflicts in the old one because too many things have changed since then.

Ref #1248

#### Summary

Following the discussion from [here](https://github.com/commercetools/merchant-center-frontend/pull/7424#discussion_r314775129), this PR addresses those suggestions.

#### Description

Key changes:
* New component `CustomFormModalPage`, which doesn't expect any default controls but exposes some pre-made control buttons: Confirm, Cancel, and Delete, to ease its use for the most common cases which still require dynamic controls.
* `FormModalPage`, which uses `CustomFormModalPage`under the hood, is a simpler version of the latter with all the defaults set (and the possibility to still hide the controls in case of need).
* New prop `hideControls`, which simply acts as a switch that does what it says regardless of anything else (no conditional props).

Breaking changes: `customControls` prop has been removed from `FormModalPage`. To build a customized form, use `CustomFormModalPage`, and the new prop `formControls` which replaces the old one. The same prop was replaced in `TabularModalPage`.


emmenko:
@jonnybel FYI: I created a `v16` branch that we use as the base branch for the breaking changes for the v16 release. This PR now points to that branch.

packages/application-components/src/components/modal-pages/custom-form-modal-page/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+Pre-made form control buttons which can be useful for the most common use cases: Confirm, Cancel, and Delete
+
+```js
+CustomFormModalPage.FormPrimaryButton = FormPrimaryButton;
+CustomFormModalPage.FormSecondaryButton = FormSecondaryButton;
+CustomFormModalPage.FormDeleteButton = FormDeleteButton;
+```
+
+### `CustomFormModalPage.Intl`
```
emmenko: PS: I would recommend to point to the shared messages that are exposed from the `i18n` package

packages/application-components/src/components/modal-pages/form-modal-page/README.md
```diff
@@ -8,7 +8,7 @@ import { FormModalPage } from '@commercetools-frontend/application-components';
 
 ## Description
 
-Form Modal pages are controlled components used to render a page with a form or something that requires user input, similar to `InfoModalPage` but semantically with a different role. The header includes buttons to control the submission or cancellation of the form (or which can be overriden with other custom controls).
+Form Modal pages are controlled components used to render a page with a form or something that requires user input, using `CustomFormModalPage` with default controls.
```
emmenko: Maybe we should specify a bit more in detail when you would need to opt-out and use the custom component.

packages/application-components/src/index.ts
```diff
@@ -9,6 +9,8 @@ export { default as InfoDialog } from './components/dialogs/info-dialog';
 export { default as ConfirmationDialog } from './components/dialogs/confirmation-dialog';
 export { default as FormDialog } from './components/dialogs/form-dialog';
 
+// Modals
```
emmenko: Nit: Modal pages

packages/application-components/src/components/modal-pages/custom-form-modal-page/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Usage
+
+```js
+import CustomFormModalPageExample from './custom-form-modal-page.example';
+```
+
+## Description
+
+Custom Form Modal pages are controlled components used to render a page with a form or something that requires user input, similar to `InfoModalPage` but semantically with a different role. The header includes an area for control buttons, which can be customized at will. The component exposes three default buttons which can be useful for most use cases.
```
tdeekens: Maybe we can be more clear here when to use which component?

Given that you intend to build yada
Then we recommend using an InfoModalPage
Given you intend to buildfoo
Then we recommend using an CustomFormModalPage

Not sure. Just a thought to have clearer usage guidelines.

packages/application-components/src/components/modal-pages/custom-form-modal-page/README.md
tdeekens: I think it's the shared messages underneath. I think it's nice consumer ergonomics to expose them on the component so he doesn't have o install another package and import it.

packages/application-components/src/components/modal-pages/internals/default-form-buttons.tsx
```diff
@@ -0,0 +1,101 @@
...
+};
+
+const useFormattedLabel = (label: Label) => {
+  const intl = useIntl();
+
+  return typeof label === 'string' ? label : intl.formatMessage(label);
+};
+
+const FormPrimaryButton = (props: Props) => {
+  const label = useFormattedLabel(props.label);
```
tdeekens: 💯

packages/application-components/src/components/modal-pages/internals/default-form-buttons.tsx
```diff
@@ -0,0 +1,101 @@
...
+  label: Label;
+  onClick: (event: React.SyntheticEvent) => void;
+  isDisabled: boolean;
+  dataAttributes: { [key: string]: string };
+  children?: never;
+};
+
+const primaryDefaultProps: Pick<
+  Props,
+  'label' | 'isDisabled' | 'dataAttributes'
```
tdeekens: Nit: I would not inline these types have have it ontop `type PrimaryDefaultProps = ...` 🤷‍♂  sorry :)

packages/application-components/src/components/modal-pages/tabular-modal-page/README.md
```diff
@@ -19,6 +19,11 @@ Tabular Modal pages are controlled components used to render a page with custom
   onClose={handleClose}
   topBarCurrentPathLabel="Lorem ipsum"
   topBarPreviousPathLabel="Back"
+  formControls={[
```
tdeekens: I need my memory refreshed but also don't want to start a long discussion again. In anycase: `components: { ConfirmButton, CancelButton, DeleteButton }` we decided again why?

packages/application-components/src/components/modal-pages/form-modal-page/form-modal-page.tsx
```diff
@@ -33,70 +20,55 @@ type CommonProps = {
...
     baseZIndex={props.baseZIndex}
-    currentPathLabel={props.topBarCurrentPathLabel || props.title}
-    previousPathLabel={props.topBarPreviousPathLabel}
+    topBarCurrentPathLabel={props.topBarCurrentPathLabel}
+    topBarPreviousPathLabel={props.topBarPreviousPathLabel}
     getParentSelector={props.getParentSelector}
     shouldDelayOnClose={props.shouldDelayOnClose}
+    hideControls={props.hideControls}
+    formControls={
+      <React.Fragment>
```
tdeekens:
```suggestion
      <>
```

packages/application-components/src/components/modal-pages/form-modal-page/form-modal-page.tsx
```diff
@@ -33,70 +20,55 @@ type CommonProps = {
...
+          isDisabled={props.isSecondaryButtonDisabled}
+          dataAttributes={props.dataAttributesSecondaryButton}
+        />
+        <CustomFormModalPage.FormPrimaryButton
+          label={props.labelPrimaryButton}
+          onClick={props.onPrimaryButtonClick}
+          isDisabled={props.isPrimaryButtonDisabled}
+          dataAttributes={props.dataAttributesPrimaryButton}
+        />
+      </React.Fragment>
```
tdeekens:
```suggestion
      </>
```

visual-testing-app/src/components/custom-form-modal-page/custom-form-modal-page.visualroute.tsx
```diff
@@ -0,0 +1,126 @@
...
+  email: string;
+};
+
+const ModalPageWithPortalParentSelector = ({
+  portalId,
+  ...props
+}: ContainerProps) => (
+  <>
+    <div id={portalId} style={{ position: 'relative', height: '750px' }} />
+    <Formik<FormValues>
```
tdeekens: ℹ️ We could actually do `useFormik()` here I think to avoid nesting.

website-components-playground/src/pages/custom-form-modal-page.js
```diff
@@ -0,0 +1,112 @@
...
+      ]}
+    >
+      {({ values }) => (
+        <ModalController
+          title="Open the Custom Form Modal Page by clicking on the button"
+          buttonLabel="Open Custom Form Modal Page"
+          containerId={containerId}
+        >
+          {({ isOpen, setIsOpen }) => (
+            <Formik
```
tdeekens: If you want you can also useFormik here to simplify things. Doesn't matter though or we do it in one go for all examples in the app.

packages/application-components/src/components/modal-pages/custom-form-modal-page/README.md
jonnybel: Yes, it's the shared messages underneath. Should we mention that here in the Readme?

packages/application-components/src/components/modal-pages/tabular-modal-page/README.md
jonnybel: You mean their namings, or the fact that these are the defaults?

packages/application-components/src/components/modal-pages/custom-form-modal-page/README.md
tdeekens: I wouldn't. I don't think it should matter to the consumer how we compose below ourselves :)

packages/application-components/src/components/modal-pages/tabular-modal-page/README.md
tdeekens: Their namings. Ah and now I get why we keep it as is. Thanks :)

packages/application-components/src/components/modal-pages/custom-form-modal-page/README.md
emmenko: Actually I would prefer for people to use the shared messages directly. For example, it's confusing if you need to provide a message that is not exposed by the component. I would also avoid unnecessary static exports.

I would actually remove those exports from the other components as well in the v16 release.

tdeekens: Okay. Let's keep it like this here cause it's consistent with other PRs and not to dive into discussion land again :). Then we can spin up a PR to remove them. Even though I feel I disagree with doing that but at least it's a seperate discussion then.

emmenko:
Damn, I screw things up...let me fix it

emmenko:
Alright, this should be good now. I added the missing MDX file for the component, and I rephrased the description to make it clear what the use case is.

NOTE: disregard the README files of the components, it will be "cleaned up" in another PR to link to the docs website.

emmenko:
Hmm I can't seem to be able to merge this 🤔

image

PR #1272, opened 2020-01-24
chore: regenerate schema types, use ctp project used for e2e tests

emmenko:
This should prevent having different ctp schemas when using different projects.


.env.template
```diff
@@ -3,4 +3,4 @@ AUTH_TOKEN= # for github labels
 MC_API_URL=https://mc-api.commercetools.com/graphql
 MC_PROXY_URL=https://mc.commercetools.com/api/graphql
 MC_ACCESS_TOKEN= # for schema introspection
-CTP_PROJECT_KEY= # one of your CTP projects where you have access to
\ No newline at end of file
+CTP_PROJECT_KEY=mc-e2e-app-kit-01 # one of your CTP projects where you have access to
```
emmenko: FYI: @tdeekens let's use this project for pulling the schema.

PR #1273, opened 2020-01-25
refactor(app-shell): migrate NavBar to TS

emmenko:
Base branch #1069 

This PR refactors the treacherous <NavBar> component to TS, while also migrating the components to use hooks.
At the end it wasn't that bad, I only had to accommodate a couple of things to make it work (which kind of makes sense, see inline comments).


packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -94,7 +94,7 @@ export const mapProjectToApplicationContextProject = (
 };
 
 type TApplicationContextEnvironment = ApplicationWindow['app'];
-type TApplicationContext<AdditionalEnvironmentProperties extends {}> = {
+export type TApplicationContext<AdditionalEnvironmentProperties extends {}> = {
```
emmenko: Exporting so that it can be used from other packages

packages/application-shell-connectors/src/components/application-context/index.ts
```diff
@@ -1,8 +1,14 @@
...
+  ProviderProps,
+  TApplicationContext as ApplicationContext,
+} from './application-context';
 
 export type TProviderProps<
   AdditionalEnvironmentProperties extends {}
 > = ProviderProps<AdditionalEnvironmentProperties>;
+export type TApplicationContext<
+  AdditionalEnvironmentProperties extends {}
+> = ApplicationContext<AdditionalEnvironmentProperties>;
```
emmenko: As a reminder, because we're using babel with typescript, we need to re-export the type as an alias.

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -654,14 +654,14 @@ describe('when clicking on navbar menu toggle', () => {
     await wait(() => {
       expect(window.localStorage.setItem).toHaveBeenCalledWith(
         STORAGE_KEYS.IS_FORCED_MENU_OPEN,
-        false
+        'false'
```
emmenko: The `localStorage.setItem` type signature expects a string

image

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -0,0 +1,667 @@
...
+  )
+    return null;
+
+  const permissionsWrapper =
+    (Array.isArray(props.permissions) && props.permissions.length > 0) ||
+    (Array.isArray(props.dataFences) && props.dataFences.length > 0) ? (
+      <RestrictedByPermissions
+        permissions={props.permissions}
+        actionRights={props.actionRights}
+        dataFences={props.dataFences}
```
emmenko: The `dataFence` value provided in the menu config contains the `type` field which is typed as a `string`

image

This is fine, however in the permissions package we used to expect a specific value like `store`. Now in the permissions package we don't have this hard constraint anymore, the `type` field is just a string.

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -0,0 +1,667 @@
...
+  const {
+    navBarNode,
+    isMenuOpen,
+    isExpanderVisible,
+    activeItemIndex,
+    handleToggleItem,
+    handleToggleMenu,
+    shouldCloseMenuFly,
+    allApplicationNavbarMenu,
+  } = useNavbarStateManager({
```
emmenko: All the hooks have been extracted into this single hook.

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -245,15 +244,22 @@ const getHasDemandedDataFence = (
...
   demandedDataFenceGroup: string
 ) => {
   if (!actualDataFences) return null;
 
   if (demandedDataFenceType in actualDataFences) {
-    const actualDataFence = actualDataFences[demandedDataFenceType];
-    if (actualDataFence && demandedDataFenceGroup in actualDataFence) {
-      return actualDataFence[demandedDataFenceGroup];
+    switch (demandedDataFenceType) {
+      case 'store': {
```
emmenko: Just a consequence of making the `type` field a generic `string`.

@types-extensions/css.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.mod.css' {
```
tdeekens: Good start intro reviewing. Haha

packages/application-shell-connectors/src/components/application-context/application-context.tsx
tdeekens: While reading: we actually do verify that when using withApplicationContext, that what is passed as a generic for the return type is actually something that is available? Just curious/a  thought.

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
```diff
@@ -0,0 +1,197 @@
...
+type HookProps = {
+  environment: TApplicationContext<{}>['environment'];
+  DEV_ONLY__loadNavbarMenuConfig?: () => Promise<TApplicationsMenu['navBar']>;
+};
+
+const useNavbarStateManager = (props: HookProps) => {
+  const navBarNode = React.useRef<HTMLElement>(null);
+  const areProjectExtensionsEnabled = useFeatureToggle(PROJECT_EXTENSIONS);
+  const applicationsNavBarMenu = useApplicationsMenu<'navBar'>('navBar', {
+    queryOptions: {
```
tdeekens: These are the apollo options, right?

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
```diff
@@ -0,0 +1,197 @@
...
+    variables: {
+      target: GRAPHQL_TARGETS.SETTINGS_SERVICE,
+    },
+    fetchPolicy: 'cache-and-network',
+    onError: reportErrorToSentry,
+  });
+  const customAppsMenu =
+    projectExtensionsQuery &&
+    projectExtensionsQuery.projectExtension &&
+    projectExtensionsQuery.projectExtension.applications
```
tdeekens: I know it's an obvious question but can't TypeScript recently optionally chain into it?

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
```diff
@@ -0,0 +1,197 @@
...
+      // - if screen is small we close it
+      setIsMenuOpen(shouldOpen);
+      setIsExpanderVisible(true);
+    } else if (!isNil(isForcedMenuOpen) && isMenuOpen !== isForcedMenuOpen) {
+      // User has setting, we should use that and ignore the screen size.
+      // Note: if viewport size is small, we should ignore the user settings.
+      setIsMenuOpen(isForcedMenuOpen);
+      setIsExpanderVisible(true);
+    }
+  }, [isExpanderVisible, isMenuOpen, isForcedMenuOpen]);
```
tdeekens: This is still mad logic to me 😄. But it's good to move it over for now and then maybe partially redo the thing.

packages/permissions/src/components/authorized/authorized.tsx
```diff
@@ -13,11 +13,10 @@ type TDemandedActionRight = {
 };
 
 // Data fences
-type TDataFenceType = 'store';
 type TDemandedDataFence = {
   group: string;
   name: string;
-  type: TDataFenceType;
+  type: string;
```
tdeekens: Why did we have to change the type?

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
```diff
@@ -0,0 +1,197 @@
...
+  const isForcedMenuOpen =
+    typeof cachedIsForcedMenuOpen === 'string'
+      ? cachedIsForcedMenuOpen === 'true'
+      : null;
+
+  const [activeItemIndex, setActiveItemIndex] = React.useState<string | null>(
+    null
+  );
+  const [isExpanderVisible, setIsExpanderVisible] = React.useState(true);
+  const [isMenuOpen, setIsMenuOpen] = React.useState(false);
```
tdeekens: Not sure if I recall correctly but this is "just" a toggle right? Do we have more of these in app-kit so that we could add a `const {on, toggle, setOn, setOff} = useToggle()` maybe? Just ain idea...

packages/permissions/src/utils/has-permissions.ts
tdeekens: Can you say way you had to do that?

packages/application-shell/src/components/navbar/navbar.tsx
tdeekens: Got it. Disregard previous comment.

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -0,0 +1,667 @@
...
+        // Always check that some of the given permissions match.
+        shouldMatchSomePermissions={true}
+      >
+        {props.children}
+      </RestrictedByPermissions>
+    ) : (
+      <>{props.children}</>
+    );
+
+  if (props.featureToggle) {
```
tdeekens: Nit: could acutally be a useFeatureToggle and ternary, right?

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -0,0 +1,667 @@
...
+    handleToggleItem,
+    handleToggleMenu,
+    shouldCloseMenuFly,
+    allApplicationNavbarMenu,
+  } = useNavbarStateManager({
+    environment: props.environment,
+    DEV_ONLY__loadNavbarMenuConfig: props.DEV_ONLY__loadNavbarMenuConfig,
+  });
+  const disabledMenuItems = props.environment.disabledMenuItems;
+  const menuVisibilities = useApplicationContext(
```
tdeekens: It's not for this PR but those context accessors should often be `useMenuVisibilities` and build ontop of the application context but consumers don't care. We have sooo much code about this in the MC about languages and stuff. That simple wrappers could ease things and hide that detail.

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -0,0 +1,667 @@
...
+
+  const isMainMenuRouteActive = (link: string) => {
+    const match = matchPath(props.location.pathname, {
+      path: `/${props.projectKey}/${link}`,
+      exact: false,
+      strict: false,
+    });
+    return Boolean(match);
+  };
+  const renderLabel = (labelAllLocales: TLocalizedField[]) => {
```
tdeekens: This was like that right? Feels more like a component than a render prop.

packages/application-shell-connectors/src/components/application-context/application-context.tsx
emmenko: Hmm what do you mean? The `AdditionalEnvironmentProperties` part?

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
emmenko: Yep

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
emmenko: True, will change thanks :)

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
emmenko: Hmm yes and now...I refactored the 3 states into a reducer now, as I was noticing some flickering in some edge cases, probably because of having multiple "`setState`" calls.
I think a reducer in this case makes more sense, cause those 3 states are often updated together.

packages/application-shell/src/components/navbar/navbar.tsx
emmenko: That's tricky, cause the prop is optional, however the `useFeatureToggle` expects a flag value but because there is an early return in this component, declaring a conditional hook won't work. 🤷‍♂️

packages/application-shell/src/components/navbar/navbar.tsx
emmenko: Yeah, you're right: https://github.com/commercetools/merchant-center-application-kit/pull/1273/commits/ccf8cc5bea3ba9edff296057a6e49d7fe19e0be5

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
emmenko: https://github.com/commercetools/merchant-center-application-kit/pull/1273/commits/ccf8cc5bea3ba9edff296057a6e49d7fe19e0be5

emmenko:
@tdeekens thanks for the quick/initial review ❤️

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
tdeekens: Makes sense yes. I often start to think that there are more cases for useReducer than one initially thinks. After all, useState just builds on it.

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
emmenko: Mäh, we don't use the babel plugin yet.

image

Will keep it like this for now and look into enabling optional chaining in another PR.

packages/application-shell-connectors/src/components/application-context/application-context.tsx
tdeekens: Sort of. What I thought of isn’t possible though I think. I hoped that if we habe conext props and additional props. That whenever using some of them through a hook/hoc they are at least a subset of them.

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -0,0 +1,678 @@
...
+    </ul>
+  );
+};
+MenuGroup.displayName = 'MenuGroup';
+
+type MenuItemProps = {
+  hasSubmenu: boolean;
+  isActive: boolean;
+  isMenuOpen: boolean;
+  isBottomItem?: boolean;
```
dogayuksel: couldn't find the use for this prop.

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -0,0 +1,678 @@
...
+  </div>
+);
+MenuItemDivider.displayName = 'MenuItemDivider';
+
+// This component wraps `<RestrictedByPermissions>`
+// and the `<ToggleFeature>` components. However, it's necessary to have it as
+// the `<ToggleFeature>` wrapper should be rendered only if the `featureToggle`
+// prop is defined. This is because `<ToggleFeature>` will not render any
+// children if the flag is missing/not found.
+const hasMenuEveryMenuVisibilitySetToBeHidden = (
```
dogayuksel: maybe should be like `isEveryMenuVisibilitySetToHidden', i find it hard to understand.

tdeekens: Good point. I think that initial name is my bad :)

packages/application-shell/src/components/navbar/navbar.tsx
emmenko: Good catch, I'll remove it!

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
```diff
@@ -0,0 +1,248 @@
...
+              shouldRenderDivider: false,
+            };
+          })
+          .filter(nonNullable)
+      : [];
+  const cachedIsForcedMenuOpen = window.localStorage.getItem(
+    STORAGE_KEYS.IS_FORCED_MENU_OPEN
+  );
+  const isForcedMenuOpen =
+    typeof cachedIsForcedMenuOpen === 'string'
```
dogayuksel: `(method) Storage.getItem(key: string): string | null`

Seems like getItem either returns a string or null, string type check may be redundant.

Possibly can use: `cachedIsForcedMenuOpen === 'true' || null`

emmenko: True, will change it. thanks

emmenko:
@tdeekens @dogayuksel anything else from your side?

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
```diff
@@ -0,0 +1,245 @@
...
+  isExpanderVisible: boolean;
+  isMenuOpen: boolean;
+};
+type Action =
+  | { type: 'setActiveItemIndex'; payload: string }
+  | { type: 'unsetActiveItemIndex' }
+  | { type: 'setIsExpanderVisible' }
+  | { type: 'toggleIsMenuOpen' }
+  | { type: 'setIsMenuOpenAndMakeExpanderVisible'; payload: boolean }
+  | { type: 'reset' };
```
tdeekens: Nice way of avoiding the good old constants for actions.

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
```diff
@@ -0,0 +1,245 @@
...
+    STORAGE_KEYS.IS_FORCED_MENU_OPEN
+  );
+  const isForcedMenuOpen = cachedIsForcedMenuOpen === 'true' || null;
+
+  const [state, dispatch] = React.useReducer<
+    (prevState: State, action: Action) => State
+  >(reducer, initialState);
+
+  const checkSize = React.useCallback(
+    throttle(() => {
```
tdeekens: This is kind of like having a useThrottle, right? I wonder if having that would make more sense?

packages/permissions/src/utils/has-permissions.ts
tdeekens: 👋

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -654,14 +654,14 @@ describe('when clicking on navbar menu toggle', () => {
...
     });
     fireEvent.click(button);
     await wait(() => {
       expect(window.localStorage.setItem).toHaveBeenCalledWith(
         STORAGE_KEYS.IS_FORCED_MENU_OPEN,
-        true
+        'true'
       );
     });
     // Check that the support link is rendered
```
tdeekens: Nice to do tests first and then change the impl.

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
emmenko: Actually the logic was correct as it was, I reverted the change here: https://github.com/commercetools/merchant-center-application-kit/pull/1273/commits/88b45d8d0cbac2cd9fb55a7b636097034dd120bb

The thing is, only in case the `getItem` returns a string, we should check for the boolean value, otherwise keep it `null`.

By doing `cachedIsForcedMenuOpen === 'true' || null`, we lose the `false` state.

dogayuksel: Sorry about this. It makes sense now, but very implicit.

tdeekens:
@emmenko https://github.com/commercetools/merchant-center-application-kit/pull/1273#discussion_r371396257 any insight to share here?

emmenko:
I thought I mentioned this already: https://github.com/commercetools/merchant-center-application-kit/pull/1273#discussion_r370955824

Or are you specifically referring to the switch-case?

tdeekens:
Ah okay. Didn't see it relating to that. Thanks!

PR #1277, opened 2020-01-27
refactor(app-shell): migrate QuickAccess to TS

emmenko:
Base branch: #1069 

This was a bit of a tough one, mostly to convert the components to use hooks and to handle some edge cases with async tasks.

All the commands now have a "clean" interface and it's nice to see all the related parts coming up nicely together.

I'll add some comments here and there to point out some of the important things.


packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -94,7 +94,7 @@ export const mapProjectToApplicationContextProject = (
 };
 
 type TApplicationContextEnvironment = ApplicationWindow['app'];
-type TApplicationContext<AdditionalEnvironmentProperties extends {}> = {
+export type TApplicationContext<AdditionalEnvironmentProperties extends {}> = {
```
emmenko: FYI: I have the same change in the other PR #1273

tdeekens:
Maybe something nicer thant he navigation for @dogayuksel to have a look at too?

packages/application-shell/src/components/quick-access/butler/butler.tsx
```diff
@@ -0,0 +1,742 @@
...
+  hasNetworkError: boolean;
+  isLoading: boolean;
+  searchText: SearchText;
+  selectedResult: SelectedResult;
+  // Used for UX when browsing through history
+  enableHistory: boolean;
+  results: Command[];
+  stack: Stack[];
+};
+type Action =
```
emmenko: This is a perfect example for using a reducer: there are lot of state properties which are updated together most of the time, and most of the logic resides in the reducer.
This means that we can just dispatch an action and the reducer takes care of the rest.

packages/application-shell/src/components/quick-access/butler/butler.tsx
```diff
@@ -0,0 +1,742 @@
...
+
+          // only allow diving in when cursor is at end of input or when
+          // the complete text is selected (when browsing through history)
+          if (!isCursorAtEnd && !isEverythingSelected) return;
+
+          unsetHasNetworkError();
+
+          // NOTE: since we need to fetch the "next command", which is an async operation,
+          // we use a IIFE to process that and eventually update the state.
+          (async () => {
```
emmenko: This was one of the edge cases. Getting the "next" commands returns a promise, because some of the subcommands require to make a network request to fetch some data. As a consequence we need to treat this asynchronously.

Before I think it was kind of broken, or at least the promises were not properly "awaited".

packages/application-shell/src/components/quick-access/create-commands.ts
```diff
@@ -1,18 +1,33 @@
...
+};
+
+const createCommands = ({
   intl,
+  applicationContext,
   featureToggles,
   changeProjectDataLocale,
-}) {
+}: CreateCommandsOptions): Command[] => {
+  console.log('featureToggles', featureToggles);
```
emmenko: Forgot to remove this. However, there seem to be something wrong with the feature toggles in the tests, I need to investigate this a bit.

packages/application-shell/src/components/quick-access/create-commands.ts
```diff
@@ -479,26 +494,32 @@ export default function createCommands({
           'set project data language',
           'set project data locale',
         ],
+        action: () => void 0,
```
emmenko: The action property is required for a `Command`

packages/application-shell/src/components/quick-access/create-commands.ts
```diff
@@ -479,26 +494,32 @@ export default function createCommands({
...
         // We would know these statically, but we define them here as we don't
         // want to include them in the top-level search results
         subCommands: () =>
-          applicationContext.project.languages.map(language => ({
-            id: `action/set-resource-language/${language}`,
-            text: oneLineTrim`
+          Promise.resolve(
```
emmenko: A subcommand that is a function, needs to return a promise.

packages/application-shell/src/components/quick-access/quick-access.tsx
```diff
@@ -0,0 +1,508 @@
...
+    isCustomApplicationsEnabled,
+    isCanViewDashboardEnabled,
+  ] = useFeatureToggles({
+    pimSearch: true,
+    customApplications: true,
+    canViewDashboard: true,
+  });
+  const applicationContext = useApplicationContext();
+
+  const dispatchFetchProductIds = useAsyncDispatch<
```
emmenko: I kind of like our custom async dispatch function, makes it easy to use and get the correct types.

packages/application-shell/src/components/quick-access/quick-access.tsx
```diff
@@ -0,0 +1,508 @@
...
+    applicationContext.project,
+    dispatchFetchPimIndexerStatus,
+  ]);
+
+  const getProjectIndexStatus = React.useCallback(async () => {
+    // skip when there is no project
+    if (!applicationContext.project) return pimIndexerStates.NOT_INDEXED;
+
+    const canViewProducts = hasSomePermissions(
+      ['ViewProducts'],
```
emmenko: I kept only one permission now, as `ManageProducts` implies `ViewProducts`.

packages/application-shell/src/components/quick-access/quick-access.tsx
```diff
@@ -0,0 +1,508 @@
...
+    [apolloClient]
+  );
+
+  const getNextCommands = React.useCallback<
+    (command: Command) => Promise<Command[]>
+  >(
+    async command => {
+      if (!command.subCommands) return [];
+      if (Array.isArray(command.subCommands)) return command.subCommands;
+      return await command.subCommands(execQuery);
```
emmenko: As mentioned in the other comments, this is where the subcommand is called as a function, which needs to return a Promise

packages/application-shell/src/components/quick-access/sub-commands.ts
```diff
@@ -0,0 +1,230 @@
...
+      action: {
+        type: 'go',
+        to: oneLineTrim`
+          /${applicationContext.project.key}
+          /products
+          /${productId}
+          /variants
+        `,
+      },
+      subCommands: execQuery =>
```
emmenko: This is the async subcommand

packages/application-shell/src/components/quick-access/types.ts
```diff
@@ -0,0 +1,34 @@
+import { DocumentNode } from 'graphql';
```
emmenko: This time I decided to extract some of the common types into a separate file, to avoid too much repetition.

The main reason we usually don't do this is because of the babel plugin for generating prop types, as it can only do that when types are within the same file.
However, `<QuickAccess>` is an internal component and is not used "outside" of the library, therefore prop types are not very important.

emmenko:
> Maybe something nicer thant he navigation for @dogayuksel to have a look at too?

I think this is actually even more complicated than the navigation 😅

emmenko:
I'll fix the tests first, then ping you both for review.

emmenko:
Alright, tests should be fixed now. There were some tricky edge cases that were previously not covered. Now, because of the strict structure given by the type system, certain things had to be "properly implemented", which led to uncover some failing tests.

At the end it was actually kind of cool, the test coverage was well done and I basically didn't change or write new tests tests but only fixed some of the setup parts.

packages/application-shell/src/components/quick-access/butler-command/butler-command.tsx
```diff
@@ -1,13 +1,22 @@
...
+  onMouseEnter: MouseEventHandler<HTMLDivElement>;
+};
+
+const ButlerCommand = (props: Props) => (
   <div
-    key={command.id}
-    data-testid={`quick-access-result(${command.id})`}
+    key={props.command.id}
+    data-testid={`quick-access-result(${props.command.id})`}
+    aria-current={props.isSelected === true ? 'true' : 'false'}
```
emmenko: This helps in the tests to make some assertions (previously css colors were used)

tdeekens:
Sorry, maybe somebody else should review this to spread the review load and getting in touch with TypeScript a bit?

emmenko:
To be honest there's a lot to take in here and it can get quite overwhelming. I'm not sure it's a good idea to use this as a "starting point".
I also don't expect a "full review", I'm fine to rely on the test coverage and the static types to ensure things are working.

tdeekens:
I agree but we've used this argument nomerious times by now and it also doesn't solve much. Quick Access is within the domain of PCM and we're changing their code without, from what I can tell, any prior notice? I would be apt to at least inform them and ask if they want to get involved by having a look at it? I don't feel "things are working" is the only measure of a success for a code review it's also about them being able to work with their code and being able to understand it. I don't want to make this a meta-level discussion but somehow we need to at least make an attempt to take that perspective and contac them?

emmenko:
I see your point and I value your feedback of course. To be fair, I did ask and mentioned in Slack (about 2 weeks ago) that I was going to finish the refactoring and that anyone interested should let me know that I'll gladly put them as reviewers. So far only Doga expressed his interest (besides you).

> Quick Access is within the domain of PCM and we're changing their code without, from what I can tell, any prior notice? I would be apt to at least inform them and ask if they want to get involved by having a look at it?

Sure, I'll let them know but realistically nobody from the PCM team ever did any maintenance work on this part of the code. I'm just saying, the timing is a bit odd to ask for a review from someone who might not even know how things works.

In any case, you're right, we should at least let them know.

tdeekens:
Thanks. I know and understand the timing is off. Also if they never did do any maintenance. They at least shouldn’t have to open the book of TypeScript surprised when wanting of having to. Not saying they have to review in full but maybe they even want to glimpse at it. After all functionality is unlikely to break but maybe a high level review gives a bit of familiarity. After all, gaining knowledge of how this works can only be started to be gathered by getting feet wet. Otherwise we never move out of our status quo.

dogayuksel:
I read through, looks good to me.

emmenko:
Thanks again for the quick feedback and reviews. Only one PR left to finish the migration 🙌

PR #1279, opened 2020-01-28
refactor: make applicationName a required field

emmenko:
> Part of the v16 breaking changes #1248

### DIFF
--- a/packages/sdk/src/middleware/client.ts
+++ b/packages/sdk/src/middleware/client.ts
@@ -14,7 +14,7 @@ declare let window: ApplicationWindow;
 
 const userAgent = createHttpUserAgent({
   name: '@commercetools/sdk-client',
-  libraryName: [window.app.applicationName || 'unknown', 'sdk'].join('/'),
+  libraryName: [window.app.applicationName, 'sdk'].join('/'),
   libraryVersion: version,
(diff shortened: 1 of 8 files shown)


packages/sdk/src/middleware/client.ts
```diff
@@ -14,7 +14,7 @@ declare let window: ApplicationWindow;
 
 const userAgent = createHttpUserAgent({
   name: '@commercetools/sdk-client',
-  libraryName: [window.app.applicationName || 'unknown', 'sdk'].join('/'),
+  libraryName: [window.app.applicationName, 'sdk'].join('/'),
```
tdeekens: Somehow here something like `internal/external` would help?

emmenko: Hmm yeah, the question is how can we reliably tell the difference?

tdeekens: Not really. I think ultiatively it should be a field in the logs of the MC API. No body wants to split a user agent string's library by a regex. 

I am just trying to think simple. Maybe our internal apps could but a `internal` property onto the window we could send as a header to the MC API to then put into Kibana. I think there are maybe more important things but it would help us to see how many requests we have etc.

PR #1282, opened 2020-01-28
refactor(app-shell): migrate remaining components to TS

emmenko:
Base branch #1069 

The last piece of the puzzle! 🎉


packages/application-shell/src/components/application-shell/application-shell.tsx
```diff
@@ -200,19 +254,18 @@ export const RestrictedApplication = props => (
...
                                 // when used outside of a project context,
                                 // or when the project is expired or supsended
-                                const useProjectContext =
-                                  project &&
-                                  !(
-                                    (project.suspension &&
-                                      project.suspension.isActive) ||
-                                    (project.expiry && project.expiry.isActive)
-                                  );
+                                const useProjectContext = !(
```
tdeekens: While we're here to not give a hooky impression we might want to rename this `shouldUseProjectContext`.

packages/application-shell/src/components/application-shell/application-shell.tsx
```diff
@@ -454,130 +503,113 @@ const shallowlyCoerceValues = memoize(uncoercedEnvironmentValues =>
...
+                              // This will be used after being logged in,
+                              // to redirect to this location.
+                              redirectTo: window.location.origin,
+                            }),
+                      }}
+                    />
+                  )}
+                />
+                <Route
+                  render={() => (
```
tdeekens: Nit: I think we can just use children as a component here to avoid a the render prop.

packages/application-shell/src/track-performance.ts
```diff
@@ -12,38 +15,40 @@ const trackPerformance = () => {
...
             variable: metricName,
-            value: duration,
+            value: paintValue,
           });
           break;
         }
         // Navigation Timing collects performance metrics for the life and
         // timings of a network request.
         case 'navigationTiming': {
+          const navigationTiming = data as Required<IPerfumeNavigationTiming>;
```
tdeekens: Nice utility type.

packages/application-shell/src/utils/gtm.ts
```diff
@@ -114,7 +114,7 @@ export const trackTiming = ({
 }: {
   category: string;
   variable: string;
-  value: string;
+  value: unknown;
```
tdeekens: Can you share why we need this to be unknown. Fail to follow.

emmenko: In theory it's a string or a number. At this point the exact type does not matter much, but yeah let's be more specific.

.eslintrc.yaml
```diff
@@ -49,12 +49,12 @@ rules:
...
   react:
     version: 'detect'
-  import/ignore:
-    - /test-utils/
-    - test-utils$
-    - test-utils\.js$
+  # import/ignore:
+  #   - /test-utils/
+  #   - test-utils$
+  #   - test-utils\.js$
```
tdeekens: Hm?

emmenko: Yeah, not sure why but eslint was complaining about `import/named`

image

CI was also failing

image

I can't remember exactly why we had those rules...

tdeekens: Strange. Would we get those in the MC too? Does intellisense work there?

emmenko: I suppose because eslint doesn't recognize the ts files from the test-utils import, which are being "imported" in a js file. Those tests are in the playground app, I suppose because the package does not have any tsconfig.

emmenko: > Would we get those in the MC too? Does intellisense work there?

Once bundled, there should be no problem.

emmenko: image

tdeekens: Should we remove the comment then before merging?

emmenko: I'll keep it there for now, just in case we bump into that issue again

PR #1283, opened 2020-01-29
docs(components): improve documentation of static Intl exports

emmenko:
> Part of the v16 breaking changes #1248

I guess this is still up for discussion but I would prefer if we remove the static `Intl` export.

From my experience so far, most of the time where I needed to use a custom message, I had to fall back to use one of the shared messages from the i18n package.
I don't see the point in keeping the static export anymore, as the recommended way of using predefined messages is to import them from the i18n package.

Thoughts?


tdeekens:
I don’t have a strong opinion from a theoretical perspective. Only that one import is less than two for consumers and we have the room to change translation _per component_ without changing the API as it is hidden. Tree scalability is another argument but mostly a strawman case to me. Lastly, app code in MC as to change. That is on us to pull through and change. In sum I personally wouldn’t make the change but am open to it.

emmenko:
> Only that one import is less than two for consumers and we have the room to change translation per component without changing the API as it is hidden.

Not 100% sure what you mean but the current export is just a proxy to the shared messages.

```js
ConfirmationDialog.Intl = sharedMessages;
```

As you said, there is one less import, but other than that I don't see any other benefit.

We can keep it like this, I don't mind. I just want us to evaluate this now because of the possible breaking change.

tdeekens:
What I mean is that each component exports its messages which behind the scenes can be less. Say two modals re-export the same message. Now we decide that they should have different messages: that is in our control. We define the messages and their mapping to our shared messages. Saying to import shared messages is a different thing. We can still control the content but not the usage/mapping as strictly. Again, I don’t mind too much but that‘s why I like the static props.

emmenko:
Hmm I feel like there is a bit of confusion. The components internally already use the shared messages. The static export was just a convenience to allow consumers to avoid having to define their own messages. Then we moved those shared message to the i18n package, which the component now use as well. The static export is not used by the components, it's just a proxy export basically.

> We define the messages and their mapping to our shared messages.

There is no mapping at the moment. Are you suggesting that we need to add some sort of mapping?

Sorry for being persistent but I feel like we are talking about different things. Maybe we should just have a 5 min call to clarify this 😅

tdeekens:
We are talking about the same thing while working another long night. I know that they are proxy exports and only convince. Nice convince however from my view. I don’t understand why we go through a change here for reasons outlined. I like the simple API of proxy exports. It encapsulates and hides information detail.

emmenko:
Alright, we keep them. I documented the export a bit more now.

PR #1290, opened 2020-01-30
chore: define needed type declaration deps in each package.json

emmenko:
According to this (https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#including-declarations-in-your-npm-package) packages should declare the type declarations used by them.

For declaration of packages that are defined as peer dependencies I defined them as peer dependencies as well, for example react.


tdeekens:
✋

PR #1291, opened 2020-01-30
chore: add homepage link and directory to each package.json

emmenko:
The `directory` field is for monorepo use cases (see https://docs.npmjs.com/files/package.json#repository)


tdeekens:
🖐

PR #1293, opened 2020-01-31
fix(components): explicitly define intl message type, to generate correct prop-types

emmenko:
The `MessageDescriptor` type imported from `react-intl` does not get parsed by the prop-types babel plugin because it's not "visible" within the file. For now, we need to explicitly define it.

Before:

image

After:

image

### DIFF
--- a/packages/application-components/src/components/dialogs/confirmation-dialog/confirmation-dialog.tsx
+++ b/packages/application-components/src/components/dialogs/confirmation-dialog/confirmation-dialog.tsx
@@ -1,11 +1,18 @@
 import React from 'react';
-import { MessageDescriptor } from 'react-intl';
 import { sharedMessages } from '@commercetools-frontend/i18n';
 import DialogContainer from '../internals/dialog-container';
 import DialogHeader from '../internals/dialog-header';
 import DialogContent from '../internals/dialog-content';
 import DialogFooter from '../internals/dialog-footer';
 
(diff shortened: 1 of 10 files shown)


packages/application-components/src/components/dialogs/confirmation-dialog/confirmation-dialog.tsx
```diff
@@ -1,11 +1,18 @@
...
 import { sharedMessages } from '@commercetools-frontend/i18n';
 import DialogContainer from '../internals/dialog-container';
 import DialogHeader from '../internals/dialog-header';
 import DialogContent from '../internals/dialog-content';
 import DialogFooter from '../internals/dialog-footer';
 
+// NOTE: the `MessageDescriptor` type is exposed by `react-intl`.
+// However, we need to explicitly define this otherwise the prop-types babel plugin
+// does not recognize the object shape.
+type MessageDescriptor = {
```
tdeekens: Could this be a type exported by `/i18n`?

emmenko: What do you mean? The problem is that the type needs to be in the file to generate the correct prop types

tdeekens: Ah sorry. Got it. :)

PR #1294, opened 2020-01-31
fix(permissions): to match inferred data fence (view when manage)

tdeekens:
#### Summary

This pull request fixes data fencing matching the inferred permission: view should be inferred by manage as with general permissions.

#### Description

I went commit by commit and review should also be possible as such.

### DIFF
--- a/packages/application-shell/src/components/application-shell/application-shell.spec.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.spec.js
@@ -1142,7 +1142,7 @@ describe('when navbar menu items do not match given data fences', () => {
               value: 'usa',
               type: 'store',
               group: 'orders',
-              name: 'canManageOrders',
+              name: 'canViewOrders',
             },
           ],
         }),
@@ -1162,7 +1162,7 @@ describe('when navbar menu items do not match given data fences', () => {
         {
           type: 'store',
           group: 'orders',
-          name: 'ViewOrders',
+          name: 'ManageOrders',
         },
       ],
     });
--- a/packages/permissions/src/utils/has-permissions.spec.tsx
+++ b/packages/permissions/src/utils/has-permissions.spec.tsx
@@ -33,7 +33,7 @@ describe('hasPermission', () => {
       demandedPermission = 'ManageProducts';
       actualPermissions = { canManageProducts: true };
     });
-    it('should return true', () => {
+    it('should indicate the permission the permission matches', () => {
       expect(hasPermission(demandedPermission, actualPermissions)).toBe(true);
     });
   });
@@ -48,7 +48,7 @@ describe('hasPermission', () => {
           canManageProducts: true,
         };
       });
-      it('should return true', () => {
+      it('should indicate the permission the permission matches', () => {
         expect(hasPermission(demandedPermission, actualPermissions)).toBe(true);
       });
     });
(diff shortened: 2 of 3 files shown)


packages/permissions/src/utils/has-permissions.ts
```diff
@@ -313,10 +319,17 @@ export const hasSomeDataFence = (options: TOptionsForAppliedDataFence) => {
...
-      ).find(
-        ([dataFenceName, dataFenceValue]) =>
-          dataFenceValue && dataFenceName === toCanCase(demandedDataFence.name)
-      );
+      ).find(([dataFenceName, dataFenceValue]) => {
+        // Either the manage permission matches or the
+        // a demanded view permission is implied by a manage permission.
+        const doesDataFenceMatchByName =
+          dataFenceName === toCanCase(demandedDataFence.name) ||
+          dataFenceName === toCanCase(toManageCase(demandedDataFence.name));
```
tdeekens: The fix. The above is just moving logic around to be reusable.

packages/permissions/src/utils/has-permissions.spec.tsx
```diff
@@ -290,8 +318,60 @@ describe('hasSomeDataFence', () => {
...
+          demandedDataFences: [
+            {
+              type: 'store',
+              group: 'orders',
+              name: 'ViewOrders',
+            },
+          ],
+          selectDataFenceData: () => ['store-1'],
+        })
+      ).toBe(true);
```
adnasa: yes! 💯

packages/permissions/src/utils/has-permissions.ts
adnasa: to clarify:

```
demandedDF: VIEW
dataFenceName VIEW
output: true

demandedDF: MANAGE
dataFenceName VIEW
output: false

demandedDF: VIEW
dataFenceName MANAGE
output: true

demandedDF: MANAGE
dataFenceName MANAGE
output: true
```

tdeekens: Yes.

tdeekens:
I don't reproduce this test failure locally. Even with `CI=true` not. Hm.

packages/permissions/src/utils/has-permissions.spec.tsx
```diff
@@ -33,7 +33,7 @@ describe('hasPermission', () => {
       demandedPermission = 'ManageProducts';
       actualPermissions = { canManageProducts: true };
     });
-    it('should return true', () => {
+    it('should indicate the permission the permission matches', () => {
```
emmenko: Duplicated words?

packages/permissions/src/utils/has-permissions.ts
```diff
@@ -66,7 +66,11 @@ type TOptionsForAppliedDataFence = {
 
 // Build the permission key from the definition to match it to the format coming
 // from the API.
```
emmenko: This comment refers to the `toCanCase` function

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -1142,7 +1142,7 @@ describe('when navbar menu items do not match given data fences', () => {
               value: 'usa',
               type: 'store',
               group: 'orders',
-              name: 'canManageOrders',
+              name: 'canViewOrders',
```
emmenko: Thanks! Was this related to the changes in this PR or was it wrong before?

tdeekens: It related to the changes and was kind of right but also wrong. data fence permissions should infer view from manage. Just like with any other permission.

PR #1300, opened 2020-02-03
fix(app-shell): to read project key from url again when location changes

emmenko:
I noticed that on the very first load of the page, where your localStorage is empty, the NavBar does not get rendered.
This is because the appshell does not render it when there is no "projectKey" in the URL.
However, when you open the MC at `/`, we then redirect you to a project `/:projectKey`.
The redirect works, but the part of the appshell where we read `selectProjectKeyFromUrl` does not re-render.

This PR adds a `useLocation` so that changes in the routes always cause a re-render.

### DIFF
--- a/packages/application-shell/src/components/application-shell/application-shell.spec.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.spec.js
@@ -136,7 +136,7 @@ describe('when route does not contain a project key (e.g. /account)', () => {
     });
   });
 });
-describe('when user fist visits "/" with no projectKey defined in localStorage', () => {
+describe('when user first visits "/" with no projectKey defined in localStorage', () => {
   let operations;
   beforeEach(() => {
     operations = mocksForMc.createMockOperations();
--- a/packages/application-shell/src/components/application-shell/application-shell.spec.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.spec.js
@@ -26,7 +26,6 @@ import {
   ApplicationNavbarMenuMock,
   ApplicationNavbarSubmenuMock,
 } from '../../../../../graphql-test-utils';
-import selectProjectKeyFromUrl from '../../utils/select-project-key-from-url';
 import { createApolloClient } from '../../configure-apollo';
 import { STORAGE_KEYS } from '../../constants';
 import { PROJECT_EXTENSIONS } from '../../feature-toggles';
@@ -34,7 +33,6 @@ import ApplicationShellProvider from '../application-shell-provider';
 import ApplicationShell from './application-shell';
 
 jest.mock('@commercetools-frontend/sentry');
-jest.mock('../../utils/select-project-key-from-url');
 
 const createTestProps = props => ({
   environment: {
@@ -122,7 +120,7 @@ describe('when rendering', () => {
     await rendered.findByText('Application name: my-app');
   });
 });
-describe('when route does not contain a project key', () => {
+describe('when route does not contain a project key (e.g. /account)', () => {
   beforeEach(() => {
     createGraphqlMockServer(xhrMock, {
       operationsByTarget: { mc: mocksForMc.createMockOperations() },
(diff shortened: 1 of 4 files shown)


packages/application-shell/src/components/application-shell/application-shell.tsx
```diff
@@ -141,336 +141,345 @@ export const RestrictedApplication = <
...
-              | typeof LOGOUT_REASONS[keyof typeof LOGOUT_REASONS]
-              | undefined;
-            if (hasUnauthorizedError)
-              logoutReason = LOGOUT_REASONS.UNAUTHORIZED;
-            else if (hasUserBeenDeletedError)
-              logoutReason = LOGOUT_REASONS.DELETED;
-            return (
-              <Redirector to="logout" queryParams={{ reason: logoutReason }} />
+) => {
+  const location = useLocation();
```
emmenko: Here we read the location

packages/application-shell/src/components/application-shell/application-shell.tsx
```diff
@@ -141,336 +141,345 @@ export const RestrictedApplication = <
...
+                );
+              }}
+            </AsyncLocaleData>
+          );
         }
-        // Since we do not know the locale of the user, we pick it from the
-        // user's browser to attempt to match the language for the correct translations.
-        const userLocale = getBrowserLocale(window);
+
+        const projectKeyFromUrl = selectProjectKeyFromUrl(location.pathname);
```
emmenko: And here we read the project key from the URL. Note that we're passing the pathname from the router location.

emmenko:
I'll try to write a test for this as well

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -138,6 +136,25 @@ describe('when route does not contain a project key', () => {
...
 });
+describe('when user fist visits "/" with no projectKey defined in localStorage', () => {
+  let operations;
+  beforeEach(() => {
+    operations = mocksForMc.createMockOperations();
+    createGraphqlMockServer(xhrMock, {
+      operationsByTarget: { mc: operations },
+    });
+  });
+  it('should not render the NavBar first, then redirect to "/:projectKey" and render the NavBar', async () => {
```
emmenko: FYI: I wanted to test that the navbar is not shown **before** the redirect, but I can't think of a way to do that. I would still like to keep the test description like this, to make it clear what the intention is.

packages/application-shell/src/components/application-shell/application-shell.tsx
emmenko: FYI: So this is a bit of a tech debt. I commented above what the trade-off is. There is no "easy" way not to achieve what we want to without having to subscribe to route changes.
Maybe if we rewrite the component to use hooks and split it up a bit, we might be able to "reduce" the area of influence regarding re-rendering when a route changes.
We should also look out a bit for performance, in case we notice that the app got "slower" because of the many re-rendering.

packages/application-shell/src/components/quick-access/index.tsx
```diff
@@ -98,12 +97,11 @@ const QuickAccessTrigger = (props: Props) => {
       document.removeEventListener('keydown', keyHandler);
     };
   }, [keyHandler]);
-  const location = useLocation();
 
   if (!isVisible) return null;
 
   return (
-    <ErrorBoundary pathname={location.pathname}>
+    <ErrorBoundary>
```
emmenko: We actually don't need the pathname for this component.

emmenko:
@adnasa @tdeekens please have another look. I left some comments for the tech debt. I would like then to release this today so that the upgrade PR is ready for review/merge. Thanks

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -122,7 +120,7 @@ describe('when rendering', () => {
     await rendered.findByText('Application name: my-app');
   });
 });
-describe('when route does not contain a project key', () => {
+describe('when route does not contain a project key (e.g. /account)', () => {
```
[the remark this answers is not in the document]
emmenko: Correct

emmenko:
> Did you check if this however means that we re-fetch the project on each route change?

Yeah I double checked. Apollo reads the data from cache, so there is no network request.

emmenko:
@tdeekens anything else from your side?

PR #1305, opened 2020-02-04
fix(create-mc-app): escape white spaces in project path

### LINKED ISSUE #1303: Installing a custom app breaks if the folder path contains a white space
**Describe the bug**
When running the script to install the starter template, if one of the folder in the path to the current working directory contains a white space, the command breaks.

**To Reproduce**

1. Assuming the current working directory is:

  ```
  /Users/emmenko/my projects/test
  ```

2. When running the install command

  ```
  npx @commercetools-frontend/create-mc-app \
    my-new-custom-application-project \
    --template starter
  ```

3. The command fail because the template could not be copied

**Expected behavior**
The command should not fail if the path contains whitespaces in the folder names

emmenko:
Fixes #1303

### DIFF
--- a/packages/create-mc-app/src/steps/download-template.js
+++ b/packages/create-mc-app/src/steps/download-template.js
@@ -1,3 +1,4 @@
+const fs = require('fs');
 const os = require('os');
 const path = require('path');
 const execSync = require('child_process').execSync;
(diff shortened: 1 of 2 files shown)


packages/create-mc-app/src/steps/download-template.js
```diff
@@ -50,13 +51,26 @@ module.exports = function downloadTemplate({
...
-      `Could not copy template "${templateName}" into "${projectDirectoryPath}"`
+      `Could not copy template "${templateName}" into "${sanitizedProjectDirectoryPath}"`
+    );
+  }
+
+  const templatePackageJsonPath = path.join(
+    projectDirectoryPath,
+    'package.json'
+  );
+  if (!fs.existsSync(templatePackageJsonPath)) {
```
emmenko: This is just an extra check

packages/create-mc-app/src/steps/download-template.js
```diff
@@ -50,13 +51,26 @@ module.exports = function downloadTemplate({
     tagOrBranchVersion
   );
 
+  const sanitizedProjectDirectoryPath = projectDirectoryPath
+    // Escape white spaces
+    .replace(/ /g, '\\ ');
```
emmenko: This is the fix

PR #1306, opened 2020-02-04
fix: use main ui-kit package as peer dependency

### LINKED ISSUE #1304: Starter template is missing some peer dependencies
**Describe the bug**
Some of the peer dependencies from AppShell are not defined in the starter template.

**To Reproduce**
Install and run a starter template, webpack fails because of missing dependencies

**Expected behavior**
The starter app should not fail to start

emmenko:
Fixes #1304 

This is a bit tricky. The appkit uses the individual packages from uikit and at the moment it requires them as peer dependencies (from appshell, app components, react-notifications).

In theory, the applications should require all the peer dependency packages from all those appkit packages. I think this is a bit "too much" to keep track and maintain those peer deps.

I propose for the appkit packages to required the main `@commercetools-frontend/ui-kit` as a peer dependency so that apps should only need to required 1 dependency.
However, apps and app-kit packages should be able to import the individual packages, as all of them are dependencies of the main `@commercetools-frontend/ui-kit` package.

WDYT?


emmenko:
Wat?? 🤔

image

tdeekens:
So the path forward to adopt split out packages in the apps would be to add the "all"-preset while pulling in the split out packages too for importing them if needed? This would result in the app to still only bundle the used packages, right? Just thinking out loud.

emmenko:
> So the path forward to adopt split out packages in the apps would be to add the "all"-preset while pulling in the split out packages too for importing them if needed? This would result in the app to still only bundle the used packages, right? Just thinking out loud.

Yeah kind of. If "you know what your're doing" you can still pull single packages, but you need to keep track of the ones used by the app-kit packages.
I think this way it will be more straight-forward.

Alternative is to not have those uikit single packages as peer deps, but that would complicate updating to a newer version of uikit without waiting for appkit to upgrade as well. Unless you use yarn resolutions.

So for the end user, I feel like it's easier that way.

application-templates/starter/package.json
```diff
@@ -19,17 +19,13 @@
...
-    "@commercetools-uikit/spacings": "10.13.0",
-    "@commercetools-uikit/text": "10.13.0",
+    "@commercetools-frontend/ui-kit": "10.13.0",
     "prop-types": "15.7.2",
     "react": "16.12.0",
     "react-apollo": "3.1.3",
     "react-dom": "16.12.0",
     "react-intl": "3.4.0",
     "react-redux": "7.1.3",
-    "react-router": "5.1.2",
```
emmenko: It's included as a dep of `react-router-dom`

image

tdeekens:
> So for the end user, I feel like it's easier that way.

...or app-kit doesn't pin those it lists as peer dependencies and uses actually a range in accordance with the peer dep range?

emmenko:
Yeah, we can do that as well. Actually I'm going to try that because the test are failing for something I don't understand

PR #1309, opened 2020-02-04
chore: run CI jobs with github actions

emmenko:
TL;DR: remove CircleCI and use GitHub Actions to run the CI workflow.

The main reason is that we want to offload some of the jobs from circleci to reduce the amount of the resources that we use there, especially once we switch to the new plan.
Since we can now use github actions with our github org account, we wanted to use those for some of our OSS repositories.

The docs-kit repo has already been migrated to github actions.


emmenko:
Alright, this should be good to go now. I'll clean up CircleCI after merging this.

.github/workflows/main.yml
```diff
@@ -0,0 +1,110 @@
...
+
+# The event triggers are configured as following:
+# - on branch master, trigger the workflow on every push
+# - on any pull request, trigger the workflow
+# This is to avoid running the workflow twice on pull requests.
+on:
+  push:
+    branches:
+    - master
+  pull_request:
```
tdeekens: Nice. This is the same at docs-kit, right?

package.json
```diff
@@ -23,6 +23,8 @@
     "format:graphql": "prettier --write --parser graphql '**/*.graphql'",
     "start": "yarn playground:start",
     "prerelease": "yarn auth && yarn build && yarn typecheck",
+    "prerelease:alpha": "yarn prerelease",
```
tdeekens: Will this be published to another dist-tag? Should it maybe while also having another prefix in the name of the bundle?

playground/package.json
```diff
@@ -8,7 +8,7 @@
     "start": "mc-scripts start",
     "compile-html:eu": "mc-scripts compile-html --headers=$(pwd)/production-eu.headers.json --config=$(pwd)/production-eu.env.json --use-local-assets --transformer $(pwd)/config/transformer-now.js",
     "compile-html:us": "mc-scripts compile-html --headers=$(pwd)/production-us.headers.json --config=$(pwd)/production-us.env.json --use-local-assets --transformer $(pwd)/config/transformer-now.js",
-    "start:prod:local": "NODE_ENV=production mc-http-server --csp=$(pwd)/csp.json --config=$(pwd)/env.json --use-local-assets",
+    "start:prod:local": "NODE_ENV=production mc-http-server --headers=$(pwd)/headers.json --config=$(pwd)/env.json --use-local-assets",
```
tdeekens: 💯

.github/workflows/main.yml
emmenko: Yep

package.json
emmenko: Hmm I'm not sure I follow 🧐

tdeekens: Ah sorry, gotta get some coffee. Didn't see the `pre` infront.

.github/workflows/main.yml
```diff
@@ -0,0 +1,117 @@
...
+          PERCY_TOKEN: ${{ secrets.PERCY_TOKEN_E2E_PLAYGROUND }}
+
+      - name: Running End-to-End tests for Starter template application
+        uses: cypress-io/github-action@v1
+        with:
+          headless: true
+          build: 'yarn template-starter:build'
+          start: 'yarn template-starter:start:prod:local'
+          wait-on: 'http://localhost:3001'
+          spec: 'cypress/integration/playground/**/*.js'
```
ahmehri:
```suggestion
          spec: 'cypress/integration/template-starter/**/*.js'
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,117 @@
...
+          env: 'CI=true'
+          command-prefix: 'percy exec --'
+        env:
+          NODE_ENV: test
+          PERCY_TOKEN: ${{ secrets.PERCY_TOKEN_E2E_STARTER }}
+
+      - name: Uploading Cypress artifacts
+        # Test run video was always captured, so this action uses "always()" condition
```
ahmehri:
```suggestion
        # Test run video was always captured because the action was using "always()" condition
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,117 @@
...
+        uses: actions/upload-artifact@v1
+        if: failure()
+        with:
+          name: cypress-videos
+          path: cypress/videos
+
+  publish_to_canary:
+    needs: build_lint_and_test
+
```
ahmehri:
```suggestion
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,117 @@
...
+    if: github.ref == 'refs/heads/master'
+
```
ahmehri:
```suggestion
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,117 @@
...
+    runs-on: ubuntu-latest
+
```
ahmehri:
```suggestion
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,117 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+
```
ahmehri:
```suggestion
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,117 @@
...
+      - name: Install dependencies
+        uses: bahmutov/npm-install@v1
+
```
ahmehri:
```suggestion
```

jest-puppeteer.config.js
```diff
@@ -1,4 +1,8 @@
 module.exports = {
+  launch: {
+    args: ['--no-sandbox', '--disable-setuid-sandbox'],
```
ahmehri: Is it maybe `--disable-setup-sandbox`?

.github/workflows/main.yml
emmenko: I think you are reviewing an outdated version

jest-puppeteer.config.js
emmenko: No it's correct: https://github.com/puppeteer/puppeteer/blob/master/docs/troubleshooting.md#setting-up-chrome-linux-sandbox

.github/workflows/main.yml
ahmehri: Yes, I already resolved the outdated conversations, no need to check them 👍 .

PR #1312, opened 2020-02-05
refactor(create-mc-app): to run tasks using listr output

emmenko:
This PR improves the output when using `create-mc-app`.
This is the same library (https://github.com/SamVerschueren/listr) that e.g. lint-staged uses. 

image

image


packages/create-mc-app/bin/create-mc-app.js
```diff
@@ -1,16 +1,66 @@
...
+  console.log(
+    `🎉 🎉 🎉 The Merchant Center application ${options.projectDirectoryName} was successfully bootstrapped based on the ${options.templateName} template.`
+  );
+  console.log('');
+  console.log(`To get started:`);
+  console.log(`$ cd ${options.projectDirectoryName}`);
+  console.log(`$ ${useYarn ? 'yarn' : 'npm'} start`);
+  console.log('');
+  console.log(
+    `Visit https://docs.commercetools.com/custom-applications for more info about developing Custom Applications. Enjoy 🚀`
```
emmenko: FYI: I took the chance to add the link to the docs

packages/create-mc-app/bin/create-mc-app.js
```diff
@@ -1,16 +1,66 @@
...
+    --template-version <version>     (optional) The version of the template to install [default "master"]
+    `);
+    process.exit(0);
+  }
+  console.log(`Version: v${currentVersion}`);
+  hintOutdatedVersion(currentVersion);
+  console.log('');
+  const options = parseArguments(flags);
+
+  const taskList = new Listr([
```
emmenko: Most of the changes are related to "moving things around a bit". The previous steps now are called tasks and return a task-like object for the Listr

packages/create-mc-app/src/hint-outdated-version.js
```diff
@@ -1,17 +1,27 @@
 const semver = require('semver');
-const execSync = require('child_process').execSync;
-const pkg = require('../../package.json');
-const logger = require('../logger');
+const execa = require('execa');
```
emmenko: We now use this library, which wraps the child_process and provides a nicer interface to work with, also with promises

packages/create-mc-app/src/utils.js
```diff
@@ -1,14 +1,10 @@
...
 const isSemVer = version => /^(v?)([0-9].[0-9].[0-9])+/.test(version);
 
 const shouldUseYarn = () => {
-  try {
-    execSync('yarnpkg --version', { stdio: 'ignore' });
-    return true;
-  } catch (e) {
-    return false;
-  }
+  const result = execa.sync('yarnpkg', ['--version'], { stdio: 'ignore' });
```
ahmehri: Why not using `yarn` here?

packages/create-mc-app/src/hint-outdated-version.js
tdeekens: I presume we could use it most cases instead of shelljs. In case we still use it?

emmenko:
> Cool. Looks good and fun. Do you think generally these things should be tested? I remember the Node.js SDK at times tests their CLI.

Yes, I plan to test the installation of a template as a follow up step

tdeekens:
> Yes, I plan to test the installation of a template as a follow up step

Cool! Yeah, maybe it's best to just test the outcome and not much of the internals. Just that something downladed, some files are there and the name is corect or something like that. You probably thought about it a bunch more already :)

packages/create-mc-app/src/utils.js
emmenko: Hmm no idea 😅

packages/create-mc-app/src/hint-outdated-version.js
emmenko: Hmm maybe yeah, although I like shelljs as well 🙂

PR #1313, opened 2020-02-05
feat(actions): add job to test installation of template apps

emmenko:
The idea with this PR is to test the installation of the application templates, as if it would be done by a normal user (so outside of the monorepository).

Normally, when running `create-mc-app`, the following happens:
- appkit repo is cloned into a temporary system folder
  - the `master` branch is used by default
  - a specific branch can be used instead
- the selected template is copied to the target location
- the template dependencies are installed

So how can we simulate the same scenario on CI?
We need to make some assumptions first:
- the template needs to be installed "outside" of the monorepository, to make sure it's installed in isolation
- the template should be cloned from the branch that is being tested
- the template dependencies should be the ones from the branch itself, as we want to test the "latest" changes and not the published versions

To achieve this we need to do 2 things:
- bundle the packages to be ready for being installed locally. We use `npm pack` to create the tarballs
- simulate the template installation in a folder outside of the monorepo

## Ideal workflow
The ideal workflow is to test the installation process on different environments (macOs and Windows) and different Node versions (10 and 12).
This would be defined as a new **job** that runs after the main workflow runs.
The second job would use the tarballs created from the previous job, to avoid installing and building the packages all over again. To persist data across jobs we can use **artifacts**.

## Current situation
However, the actions to upload/download the artifacts are currently lacking of some configuration options. For instance, we only need to persist the artifacts between jobs, but after a workflow is completed we don't need the artifacts anymore.

There are some open issues about this and hopefully it will be supported soon.
In the meantime, we can keep the workflow simple and only test the installation on ubuntu. For future references, the workflow includes the ideal setup but it's commented out.


scripts/install_template.sh
```diff
@@ -0,0 +1,39 @@
...
+echo "Current branch: $BRANCH_NAME"
+
+mkdir "$SANDBOX_FOLDER"
+
+# Copy the `create-mc-app` package into this folder, so that we can install
+# the dependencies only for this package.
+echo "Installing copy of create-mc-app package in folder $SANDBOX_FOLDER"
+cp -R packages/create-mc-app "$SANDBOX_FOLDER/create-mc-app"
+cd "$SANDBOX_FOLDER/create-mc-app"
+npm install
```
tdeekens: Should we use yarn or npm? 😕

emmenko:
Alright, this should be good now. I bumped into some "issues" (see PR description) and this is what we can achieve so far.

Here you can see the output of the installation test, that successfully builds the template.

image

.github/workflows/main.yml
```diff
@@ -108,3 +107,90 @@ jobs:
...
+      # after the workflow ends, and the artifacts will stay for 90 days.
+      # Some related issues:
+      # - https://github.com/actions/upload-artifact/issues/5
+      # - https://github.com/actions/upload-artifact/issues/45
+      - name: Installing and building template starter
+        run: node ./scripts/install-template.js
+        env:
+          TEMPLATE_NAME: starter
+
+  #     - name: Uploading tarball artifacts
```
emmenko: From here below is the "ideal setup". I kept it commented for future references.

.github/workflows/main.yml
```diff
@@ -108,3 +107,90 @@ jobs:
...
+
+  # install_template_starter:
+  #   needs: build_lint_and_test
+
+  #   runs-on: ${{ matrix.os }}
+
+  #   strategy:
+  #     matrix:
+  #       os: [macos-latest, windows-latest]
+  #       node: [10, 12]
```
emmenko: This is what we would like to test ideally: macOs/Windows for Node 10/12.

.github/workflows/main.yml
```diff
@@ -108,3 +107,90 @@ jobs:
...
+
+  #     - name: Installing shelljs globally
+  #       run: yarn add global shelljs
+
+  #     - name: Installing and building template starter
+  #       run: node ./scripts/install-template.js
+  #       env:
+  #         TEMPLATE_NAME: starter
+
+  #     - name: Fetching workflow artifacts
```
emmenko: This was just an attempt to "clean up" the artifacts, but it's a bit more complicated that this.
It's cool though to be able to make API requests using actions and to parse the results.

https://github.com/octokit/request-action
https://github.com/gr2m/get-json-paths-action

packages/create-mc-app/bin/create-mc-app.js
```diff
@@ -22,9 +25,10 @@ async function execute() {
 
   Options:
 
-    --template <name>                (optional) The name of the template to install [default "starter"]
-                                     Available options: ["starter"]
-    --template-version <version>     (optional) The version of the template to install [default "master"]
+    --template <name>               (optional) The name of the template to install [default "starter"]
+                                    Available options: ["starter"]
+    --template-version <version>    (optional) The version of the template to install [default "master"]
+    --skip-install                  (optional) Skip installing the dependencies after cloning the template [default "false"]
```
emmenko: This is a new option that we need to make the CI tests work

packages/i18n/package.json
```diff
@@ -41,12 +41,12 @@
   "dependencies": {
     "@babel/runtime-corejs3": "7.8.3",
     "@commercetools-frontend/sentry": "16.0.0",
+    "@commercetools-uikit/i18n": "^10.13.0",
```
emmenko: This was actually a bug, caught by the new tests. This dependency needs to be defined, as we did in #1306 and #1308

scripts/install-template.js
```diff
@@ -0,0 +1,98 @@
...
+const branchName =
+  process.env.GITHUB_EVENT_NAME === 'pull_request'
+    ? process.env.GITHUB_HEAD_REF
+    : 'master';
+
+const matchingPackages = /^@commercetools-frontend\/(?!ui-kit)(.*)/;
+
+const applicationName = 'my-starter-app';
+const rootPath = path.join(__dirname, '..');
+const sandboxPath = path.join(rootPath, '..', 'sandbox-installation-template');
```
emmenko: This is the folder "outside" of the monorepo where we install the template

scripts/install-template.js
```diff
@@ -0,0 +1,102 @@
+const fs = require('fs');
+const path = require('path');
+const shelljs = require('shelljs');
+
+const matchingPackages = /^@commercetools-frontend\/(?!ui-kit)(.*)/;
+const applicationName = 'my-starter-app';
+const rootPath = path.join(__dirname, '..');
+const tarballsDistPath = path.join(rootPath, 'dist-tarballs');
+const sandboxPath = path.join(rootPath, '..', 'sandbox-installation-template');
```
emmenko: This is the folder "outside" of the monorepo where we install the template

scripts/install-template.js
```diff
@@ -0,0 +1,102 @@
...
+shelljs.mkdir('-p', sandboxPath);
+
+const createMcAppTarballName = shelljs
+  .ls(tarballsDistPath)
+  .find(tarballName => tarballName.includes('create-mc-app'));
+if (!createMcAppTarballName) {
+  throw new Error('Failed to find tarball for create-mc-app package');
+}
+
+console.log(`Installing create-mc-app from tarball`);
```
emmenko: We install this first, so that we can install the template using the binary script

scripts/install-template.js
```diff
@@ -0,0 +1,102 @@
...
+);
+shelljs.exec(
+  [
+    binaryPath,
+    applicationName,
+    '--template=starter',
+    `--template-version=${branchName}`,
+    // Don't install deps automatically, we need to patch the package.json to use the
+    // local tarballs for resolving the dependencies.
+    '--skip-install',
```
emmenko: Here we skip the installation step

scripts/install-template.js
```diff
@@ -0,0 +1,102 @@
...
+  ].join(' '),
+  { cwd: sandboxPath }
+);
+
+console.log(
+  'Patching the application package.json to use local imports for appkit dependencies'
```
emmenko: Here we rewrite the appkit dependencies defined in the `package.json` by using the `file:/path/to/tarball` instead of the npm version. This way we make sure that the "latest" changes are tested.

scripts/install-template.js
```diff
@@ -0,0 +1,102 @@
...
+});
+Object.keys(appPackageJson.devDependencies).forEach(depName => {
+  patchDependencyImportPath(depName, appPackageJson.devDependencies);
+});
+fs.writeFileSync(appPackageJsonPath, JSON.stringify(appPackageJson, null, 2), {
+  encoding: 'utf8',
+});
+
+console.log('Installing the application dependencies');
+shelljs.exec('yarn install --no-lockfile', { cwd: applicationPath });
```
emmenko: Now we can install the application dependencies

.github/workflows/main.yml
```diff
@@ -108,3 +107,90 @@ jobs:
...
+  #       with:
+  #         name: dist-tarballs
+  #         path: dist-tarballs
+
+  #     - name: Setup node
+  #       uses: actions/setup-node@v1
+  #       with:
+  #         node-version: ${{ matrix.node }}
+
+  #     - name: Installing shelljs globally
```
ahmehri: Should it maybe be `execa` instead?

.github/workflows/main.yml
tdeekens: Wow. This is really dope!

packages/create-mc-app/bin/create-mc-app.js
```diff
@@ -49,6 +55,9 @@ async function execute() {
   console.log('');
   console.log(`To get started:`);
   console.log(`$ cd ${options.projectDirectoryName}`);
+  if (flags['skip-install']) {
```
tdeekens: Just wondering. Could also be `auto-install` and default to `true` while CI passes `false` maybe?

scripts/build-tarballs.js
```diff
@@ -0,0 +1,31 @@
...
+
+const rootPath = path.join(__dirname, '..');
+const tarballsDistPath = path.join(rootPath, 'dist-tarballs');
+
+function extractTarball(packageInfo) {
+  const appPackageJsonPath = path.join(packageInfo.location, 'package.json');
+  const appPackageJson = require(appPackageJsonPath);
+  if (appPackageJson.private) return;
+
+  shelljs.exec('npm pack', { cwd: packageInfo.location, silent: true });
```
tdeekens: Confused when we use execa and when shelljs now. Can you elaborate?

scripts/install-template.js
tdeekens: Ah nice!

.github/workflows/main.yml
emmenko: No this is needed for the `install-template.js` script

packages/create-mc-app/bin/create-mc-app.js
emmenko: Hmm I think I find "skip" to be more appropriate, because we are skipping one of the steps/tasks. Thanks for the suggestion though!

scripts/build-tarballs.js
emmenko: I still prefer `shelljs`, because it has good cross OS support and it can run synchronously. For example commands like `shelljs.cp`, `shelljs.mv`, etc.

The `execa` can be used with Promises, which was good for the Listr implementation.

In general for scripts I would personally still prefer to use shelljs.

.github/workflows/main.yml
ahmehri: Yeah I meant replacing the usage of `shelljs` by `execa` as @tdeekens mentioned in another PR. But I guess each tool has its own usages.

PR #1316, opened 2020-02-06
chore(actions): skip workflow if commit message contains [skip ci]

emmenko:


emmenko:
Hmm, it does not seem to be supported as we don't get the commit message in the `pull_request` event 🙃

PR #1320, opened 2020-02-08
refactor: to suggest using new url schemes and change docs and cors

tdeekens:
#### Summary

This pull request changes the occurance of the urls suggested to be used in app-kit to match the new url schema.

#### Description

Of course the old urls will continue to work. However, I think it makes sense to

1. Suggest from the start to use the new urls
2. Amend them to our CORS rules (while eventually removing the old) to allow us to migrate local development environments

As a side note, all MC services already use the new hostnames internally.

Lastly, I suggest not to change the hostname of the MC frontend just yet. We should do that only _after_ the SSO users have migrated all their callback urls and added the new urls in to avoid non matching expectations.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -1,6 +1,6 @@
 GITHUB_AUTH= # for lerna changelog
 AUTH_TOKEN= # for github labels
-MC_API_URL=https://mc-api.commercetools.com/graphql
+MC_API_URL=https://mc-api.europe-west1.gcp.commercetools.com/graphql
 MC_PROXY_URL=https://mc.commercetools.com/api/graphql
 MC_ACCESS_TOKEN= # for schema introspection
 CTP_PROJECT_KEY=mc-e2e-app-kit-01 # one of your CTP projects where you have access to
--- a/packages/mc-html-template/lib/load-headers.js
+++ b/packages/mc-html-template/lib/load-headers.js
@@ -83,6 +83,8 @@ module.exports = (env, options) => {
       ),
       'connect-src': [
         "'self'",
+        'mc-api.europe-west1.gcp.commercetools.com',
+        'mc-api.us-central1.gcp.commercetools.com',
         'mc-api.commercetools.com',
         'mc-api.commercetools.co',
         'app.launchdarkly.com',
--- a/application-templates/starter/env.json
+++ b/application-templates/starter/env.json
@@ -1,7 +1,7 @@
 {
   "applicationName": "merchant-center-application-template-starter",
   "frontendHost": "localhost:3001",
-  "mcApiUrl": "https://mc-api.commercetools.com",
+  "mcApiUrl": "https://mc-api.europe-west1.gcp.commercetools.com",
   "location": "eu",
   "env": "development",
   "cdnUrl": "http://localhost:3001",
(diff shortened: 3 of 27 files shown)


.env.template
```diff
@@ -1,6 +1,6 @@
 GITHUB_AUTH= # for lerna changelog
 AUTH_TOKEN= # for github labels
-MC_API_URL=https://mc-api.commercetools.com/graphql
-MC_PROXY_URL=https://mc.commercetools.com/api/graphql
+MC_API_URL=https://mc-api.europe-west1.gcp.commercetools.com/graphql
+MC_PROXY_URL=https://mc-api.europe-west1.gcp.commercetools.com/api/graphql
```
tdeekens: Just so the templates contain them.

packages/mc-html-template/lib/load-headers.js
```diff
@@ -83,6 +83,8 @@ module.exports = (env, options) => {
       ),
       'connect-src': [
         "'self'",
+        'mc-api.europe-west1.gcp.commercetools.com',
+        'mc-api.us-central1.gcp.commercetools.com',
```
tdeekens: This allows a local development environment to work with them by default without changing cors rules. We can then in a loooong time (😉) remove the old urls at some point.

emmenko:
Thanks but I think we should put this on hold a bit.
We should check this first with our internal apps, if using the new mc-api hostname works while using the legacy hostname.

tdeekens:
Not sure I follow 100% but if it is a call for careful testing - I am all for it. So the status quo is:

1. Staging uses the production mc-api hostname in new scheme
2. Production environments use the SVC
3. Local development uses the legacy hostname for the mc-api

What I would suggest, and can test is:

1. Changing headers.json and CORS rules to allow new mc-api hostname while changing env.json
2. Integrate a prebuild version of this locally and then using old and new mc-api hostname locally

Would that be roughly what you had in mind or did I misread your comment and concern? 🙌🏽

emmenko:
Yes we should make sure that production apps still work when using the new api hostname. Maybe we can test this using a custom app.

tdeekens:
Sure. So what I did now to test things a bit is to use a build of the `mc-html-template` (which has the `loadHeaders`) and used it locally in the `webpack-dev-server` and a production build. Cases covered are:

1. Using legacy mc-api hostname
2. Using new mc-api hostname

Here we can see that the CSP rules (without any changes to any `headers.json`-file) contain the new hostnames automatically now:

CleanShot 2020-02-09 at 12 34 31@2x

These two are the cases with both mc-api hostname "types"

CleanShot 2020-02-09 at 12 37 00@2x
CleanShot 2020-02-09 at 12 38 35@2x

In production, as we know just as a reminder for completeness,  we already use the new hostnames

CleanShot 2020-02-09 at 12 42 15@2x

Any cases I missed to cover? I am still not 100% sure I deduced the case(s) you meant to test from your answer ;)

emmenko:
In production we don’t use the new hostnames for the API, that’s why I would like to test it on one of the production apps, or using a custom app.

emmenko:
What we need to make sure is that the legacy frontend hostname works with the new API hostname

tdeekens:
Let‘s have a quick call tomorrow. I feel we mean the same thing or at least it takes to much writing until we do.

tdeekens:
Will close this for now. We want this but also may want to take 1-2 steps before.

PR #1321, opened 2020-02-08
fix(actions): check out with depth 50 to allow proper canary versions on master branch

emmenko:


emmenko:
FYI: I tested this in docs-kit and it seems to finally work.

PR #1322, opened 2020-02-10
refactor(app-shell): sdk and apollo to fallback to build `mcApiUrl` from origin

tdeekens:
#### Summary

This pull request suggests to change building the `mcApiUrl` from the window's origin if no `mcApiUrl` is defined and the app runs behind the proxy.

#### Description

Eventually the MC needs to be able to run on multiple hostnames (at least for some time). In GCP EU that means `https://mc-api.europe-west1.gcp.commercetools.com` and `https://mc-api.commercetools.com`.

This brings some challengens. Mainly around the cookie (yes cookie based auth ain't the best). 

```txt
Given you login on `mc.commercetools.com`
And your our app talks to `https://mc-api.europe-west1.gcp.commercetools.com` will log you out again
```

```txt
Given you login on `mc.europe-west1.gcp.commercetools.com`
And your our app talks to `https://mc-api.commercetools.com` will log you out again
```

This is rooted in the fact that the cookie is not valid for the different mc api domain and hence will not be sent. The MC API itself then logs out the user.

This is however unintended. As we don't want to configure two MC API urls on the env and switch between them I propose to

- Build the `mcApiUrl` from the `window.origin` only _and only if_ the app is `servedByProxy` and no `mcApiUrl` is defined

This allows us to

1. Keep Custom Applications be able to specify an `mcApiUrl` which is entirely different from what the "window runs on"
2. Keep us interally be able to overwrite the `mcApiUrl` and be backwards compatible as no change happens if the `mcApiUrl` is not removed

### DIFF
--- a/packages/application-shell/src/index.ts
+++ b/packages/application-shell/src/index.ts
@@ -8,7 +8,7 @@ export { applyDefaultMiddlewares } from './configure-store';
 export { default as InjectReducers } from './components/inject-reducers';
 export { default as RouteCatchAll } from './components/route-catch-all';
 export { default as setupGlobalErrorListener } from './utils/setup-global-error-listener';
-export { selectUserId, selectProjectKeyFromUrl, getMcApiUrl } from './utils';
+export { selectUserId, selectProjectKeyFromUrl } from './utils';
 export { GtmContext } from './components/gtm-booter';
 export { default as GtmUserLogoutTracker } from './components/gtm-user-logout-tracker';
 export { default as SetupFlopFlipProvider } from './components/setup-flop-flip-provider';
--- a/packages/application-shell/src/index.ts
+++ b/packages/application-shell/src/index.ts
@@ -8,7 +8,7 @@ export { applyDefaultMiddlewares } from './configure-store';
 export { default as InjectReducers } from './components/inject-reducers';
 export { default as RouteCatchAll } from './components/route-catch-all';
 export { default as setupGlobalErrorListener } from './utils/setup-global-error-listener';
-export { selectUserId, selectProjectKeyFromUrl } from './utils';
+export { selectUserId, selectProjectKeyFromUrl, getMcApiUrl } from './utils';
 export { GtmContext } from './components/gtm-booter';
 export { default as GtmUserLogoutTracker } from './components/gtm-user-logout-tracker';
 export { default as SetupFlopFlipProvider } from './components/setup-flop-flip-provider';
--- a/packages/application-shell/src/utils/get-mc-api-url/get-mc-api-url.ts
+++ b/packages/application-shell/src/utils/get-mc-api-url/get-mc-api-url.ts
@@ -10,7 +10,7 @@ const getMcApiUrlFromOrigin = () => {
 };
 
 export default function getMcApiUrl(appEnv = window.app) {
-  const isServedByProxy = appEnv.servedByProxy;
+  const isServedByProxy = JSON.parse(appEnv.servedByProxy);
   const mcApiUrl = appEnv.mcApiUrl;
 
   if (isServedByProxy && !mcApiUrl) return getMcApiUrlFromOrigin();
--- a/packages/application-shell/src/utils/get-mc-api-url/get-mc-api-url.ts
+++ b/packages/application-shell/src/utils/get-mc-api-url/get-mc-api-url.ts
@@ -3,5 +3,10 @@ import { ApplicationWindow } from '@commercetools-frontend/constants';
 declare let window: ApplicationWindow;
 
 export default function getMcApiUrl(appEnv = window.app) {
+  const isServedByProxy = appEnv.servedByProxy;
+  const mcApiUrl = appEnv.mcApiUrl;
+
+  if (isServedByProxy && !mcApiUrl) return window.origin;
+
   return appEnv.mcApiUrl;
 }
--- a/packages/application-shell/src/utils/get-mc-api-url/get-mc-api-url.ts
+++ b/packages/application-shell/src/utils/get-mc-api-url/get-mc-api-url.ts
@@ -2,6 +2,6 @@ import { ApplicationWindow } from '@commercetools-frontend/constants';
 
 declare let window: ApplicationWindow;
 
-export default function getMcApiUrl() {
-  return window.app.mcApiUrl;
+export default function getMcApiUrl(appEnv = window.app) {
+  return appEnv.mcApiUrl;
 }
--- a/packages/application-shell/src/utils/get-mc-api-url/get-mc-api-url.ts
+++ b/packages/application-shell/src/utils/get-mc-api-url/get-mc-api-url.ts
@@ -1,6 +1,6 @@
 export interface ApplicationWindow extends Window {
   app: {
-    servedByProxy?: string;
+    servedByProxy: string | boolean;
     mcApiUrl: string;
   };
 }
@@ -24,7 +24,7 @@ declare let window: ApplicationWindow;
  *    2. MC: mc.europe-west1.gcp.commercetools.com with API: mc-api..commercetools.com
  *       -> Will not work as urls differ
  *
- *    Using the origin ensures that urls always match the the cookie is sent. Otherwise,
+ *    Using the origin ensures that urls always match with the cookie sent. Otherwise,
  *    the application shell will rightfully redirect to the logout page.
  */
 const getMcApiUrlFromOrigin = (actualWindow: ApplicationWindow) => {
--- a/packages/application-shell/src/utils/get-mc-api-url/get-mc-api-url.ts
+++ b/packages/application-shell/src/utils/get-mc-api-url/get-mc-api-url.ts
@@ -1,0 +1,7 @@
+import { ApplicationWindow } from '@commercetools-frontend/constants';
+
+declare let window: ApplicationWindow;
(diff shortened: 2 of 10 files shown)


packages/application-shell/src/index.ts
```diff
@@ -8,7 +8,7 @@ export { applyDefaultMiddlewares } from './configure-store';
 export { default as InjectReducers } from './components/inject-reducers';
 export { default as RouteCatchAll } from './components/route-catch-all';
 export { default as setupGlobalErrorListener } from './utils/setup-global-error-listener';
-export { selectUserId, selectProjectKeyFromUrl } from './utils';
+export { selectUserId, selectProjectKeyFromUrl, getMcApiUrl } from './utils';
```
tdeekens: A bit unfortunate but due to the location of the other helpers consistent.

packages/application-shell/src/utils/get-mc-api-url/get-mc-api-url.ts
```diff
@@ -0,0 +1,19 @@
...
+
+const getMcApiUrlFromOrigin = () => {
+  const url = new URL(window.origin);
+  const mcApiUrlHost = url.host.replace('mc.', 'mc-api.');
+
+  return `${url.protocol}//${mcApiUrlHost}`;
+};
+
+export default function getMcApiUrl(appEnv = window.app) {
+  const isServedByProxy = JSON.parse(appEnv.servedByProxy);
```
tdeekens: I somehow think this is falely typed. The `servedByProxy` is a string on `window.app` on any production app or a boolean when ran locally. It however is typed to always be boolean. /cc @emmenko

emmenko: This is on purpose, as otherwise it gets confusing and we end up with more complexity on the types level.

When we pass the `window.app` to the AppShell, the values get coerced so we end up with a `boolean`. And since we do a type cast anyway when passing the prop to the AppShell, it does not matter much what the original value was.

packages/application-shell/src/utils/get-mc-api-url/get-mc-api-url.ts
```diff
@@ -0,0 +1,19 @@
...
+  const mcApiUrl = appEnv.mcApiUrl;
```
emmenko: Hmm I don't think we need to make this optional. We should just check that the app is running behind the proxy. For custom apps that's not a problem, they also run on the same domain in production.

packages/application-shell/src/utils/get-mc-api-url/get-mc-api-url.ts
```diff
@@ -0,0 +1,19 @@
+import { ApplicationWindow } from '@commercetools-frontend/constants';
+
+declare let window: ApplicationWindow;
+
+const getMcApiUrlFromOrigin = () => {
```
emmenko: We should document a bit why we do this.

packages/sdk/src/middleware/client.ts
```diff
@@ -8,6 +8,7 @@ import { createHttpMiddleware as createSdkHttpMiddleware } from '@commercetools/
 import { createCorrelationIdMiddleware as createSdkCorrelationIdMiddleware } from '@commercetools/sdk-middleware-correlation-id';
 import createHttpUserAgent from '@commercetools/http-user-agent';
 import { ApplicationWindow } from '@commercetools-frontend/constants';
+import { getMcApiUrl } from '@commercetools-frontend/application-shell';
```
emmenko: The sdk shouldn't import from the app-shell.

packages/application-shell/src/utils/get-mc-api-url/get-mc-api-url.ts
tdeekens: Yes, I know that we coerce the values shallowly. However, what remains on the window itself is untouched. What flows down the application context is coerced. The configuration of the SDK and Apollo however do not consume anything from the application context. Hence not the coerced values. As a result the value here remains string or boolean. 

As a result you're saying here we should not use the "type-safe" variables from the window?

Or Am I missing something? I feel we mean the same again just from different angles.

tdeekens:
> I think the mcApiUrl should still be required.

Sorry, such a statement in code review doesn't feel like getting us anywhere apart from ping pong. Can you elaborate why you would suggest it to be required?

My main goal by not requiring is was to allow a graceful migration while the mcApiUrl can still be explicitely defined. I felt that also helps avoiding surprises when specifying it.

packages/application-shell/src/utils/get-mc-api-url/get-mc-api-url.ts
tdeekens: Sure, will do! 👍

packages/application-shell/src/utils/get-mc-api-url/get-mc-api-url.ts
tdeekens: > We should just check that the app is running behind the proxy

...to then what. The intend here is a bit unclear to me. Can you elaborate? You man everything running behind the proxy always gets an MC API url from the origin?

tdeekens:
Alrighto 2c2b554 aims to contain the discussed changes @emmenko 😄. The logic changed to always use the `window.origin` in case we run behind the proxy (this how can never change again 🎊). Also the code had to be duplicated. The "passing-down"-approach will not work as the it's too late to set the host from within the middleware configuration.

packages/application-shell/src/utils/get-mc-api-url/get-mc-api-url.ts
```diff
@@ -0,0 +1,45 @@
+export interface ApplicationWindow extends Window {
+  app: {
+    servedByProxy?: string;
```
emmenko:
```suggestion
    servedByProxy: string | boolean;
```

packages/application-shell/src/utils/get-mc-api-url/get-mc-api-url.ts
```diff
@@ -0,0 +1,45 @@
...
+ *    This is particularily useful with the new and old hostnames for the Merchant Center.
+ *    When using the origin, it will be made sure that the authorization cookie will
+ *    always be sent as the appropriate API url is used.
+ *
+ *    1. MC: mc.commercetools.com with API: mc-api.europe-west1.gcp.commercetools.com
+ *       -> Will not work as urls differ
+ *    2. MC: mc.europe-west1.gcp.commercetools.com with API: mc-api..commercetools.com
+ *       -> Will not work as urls differ
+ *
+ *    Using the origin ensures that urls always match the the cookie is sent. Otherwise,
```
emmenko:
```suggestion
 *    Using the origin ensures that urls always match with the cookie sent. Otherwise,
```

packages/application-shell/src/utils/get-mc-api-url/get-mc-api-url.ts
```diff
@@ -0,0 +1,45 @@
...
+const getMcApiUrlFromOrigin = (actualWindow: ApplicationWindow) => {
+  const url = new URL(actualWindow.origin);
+  const mcApiUrlHost = url.host.replace('mc.', 'mc-api.');
+
+  return `${url.protocol}//${mcApiUrlHost}`;
+};
+
+export default function getMcApiUrl(actualWindow: ApplicationWindow = window) {
+  const isServedByProxy =
+    actualWindow.app.servedByProxy &&
```
emmenko: The `servedByProxy` is required, so we don't need this check

packages/application-shell/src/index.ts
emmenko: We don't need this export anymore

packages/sdk/src/middleware/client.ts
```diff
@@ -8,6 +8,7 @@ import { createHttpMiddleware as createSdkHttpMiddleware } from '@commercetools/
 import { createCorrelationIdMiddleware as createSdkCorrelationIdMiddleware } from '@commercetools/sdk-middleware-correlation-id';
 import createHttpUserAgent from '@commercetools/http-user-agent';
 import { ApplicationWindow } from '@commercetools-frontend/constants';
+import { getMcApiUrl } from '../utils';
```
emmenko: 👌

packages/application-shell/src/utils/get-mc-api-url/get-mc-api-url.ts
tdeekens: Then we can't JSON.parse anymore cause... TS says it wants a string. Whatever, I'll change the check.

emmenko: Well you can do `JSON.parse(String(window.app.servedByProxy))`

tdeekens: Come on! Then I'd rather be explicit. That's just type checker pleasing.

emmenko: The `JSON.parse` is not typed correctly in my opinion.

Anyway, how you changed it now it's better and more explicit. Let's keep that.

tdeekens: I agree. Thanks your honor 😋.

PR #1324, opened 2020-02-10
chore: update deps 💣

tdeekens:
#### Summary

Initially I set out to update flopflip and my weekends work. Then I noticed how much is outdated and spend a good amount of minutes to update things one by one.


.eslintrc.yaml
```diff
@@ -45,7 +45,7 @@ rules:
     - error
     - env: literal
   'jest/expect-expect': 'off'
-  'testing-library/prefer-expect-query-by': 'error'
+  'testing-library/no-get-by-for-checking-element-not-present': 'error'
```
tdeekens: The old rule is gone. This one is the "smarter" replacement I guess.

package.json
```diff
@@ -176,7 +176,7 @@
     "react-test-renderer": "16.12.0",
     "**/react-test-renderer": "16.12.0",
     "**/sharp": "0.23.3",
-    "cypress": "3.8.3"
+    "cypress": "4.0.1"
```
tdeekens: This should allow us to update in the MC without a resolution /cc @ahmehri

packages/application-shell-connectors/src/components/project-extension-image-regex/project-extension-image-regex.spec.tsx
```diff
@@ -121,7 +121,7 @@ describe('rendering', () => {
       ]);
     });
     it('should not render regex info', async () => {
-      await waitForElementToBeRemoved(() => rendered.getByText('Loading...'));
+      await waitForElementToBeRemoved(() => rendered.queryByText('Loading...'));
```
tdeekens: Detected by new eslint rule.

packages/application-shell/src/components/application-shell/application-shell.tsx
```diff
@@ -2,7 +2,7 @@ import React, { SyntheticEvent } from 'react';
 import { Dispatch } from 'redux';
 import { Redirect, Route, Switch, useLocation } from 'react-router-dom';
 import { ApolloError } from 'apollo-client';
-import { Flags } from '@flopflip/types';
+import { TFlags } from '@flopflip/types';
```
tdeekens: Maybe I shouldn't made a breaking change. Didn't know we use it or somebody. Well, life is...

package.json
```diff
@@ -103,7 +103,7 @@
     "babel-plugin-transform-rename-import": "2.3.0",
     "babel-plugin-typescript-to-proptypes": "1.2.1",
     "cross-env": "7.0.0",
-    "cypress": "4.0.1",
+    "cypress": "3.8.3",
```
tdeekens: Needs to be done separately I think. The GitHub action and the asset upload don't seem in sync as paths may have changed. Just a guess so far. Still, better isolate this I think.

package.json
emmenko: FYI: this dependency does not have anything to do with the one we use in the MC. Or am I missing something?

tsconfig.json
```diff
@@ -6,7 +6,7 @@
     "esModuleInterop": true,
     "importHelpers": true,
     // See https://devblogs.microsoft.com/typescript/typescript-and-babel-7/
-    "isolatedModules": true,
+    "isolatedModules": false,
```
tdeekens: Fails if we set this to true. I've read a bit about it but not 100% sure. Did it fail in earlier versions of babel v7 when being true?

emmenko: Hmm as far as I know, if we use Babel we need to activate this flag. Did you find a reference or docs that say otherwise?

Also, I'm not sure how this change relates to any of the updated dependencies. What was the error that you got? 🤔

tdeekens: CleanShot 2020-02-11 at 09 20 21@2x

I've also not found much otherwise. However, also confusing that it works. Typechecking jest however fails without.

emmenko: See https://github.com/chalk/chalk/issues/295 and https://github.com/chalk/chalk/pull/296#issuecomment-418789492

emmenko: https://github.com/chalk/chalk/issues/382

image

emmenko: I'm wondering if we still need the `@types/jest`. 

In any case, we need to keep `isolatedModules: true` and try to downgrade one of the packages so that we don't get the problem with the chalk dependency

tdeekens: Yes, read that chalk issue too. It's a bit of a cat and mouse game. I now:

1. Downgraded ts-jest
2. Removed @types/jest (comes with jest itself)

This should work. However, it's a general issue: if any library uses language TS features babel can't compile we're doomed. This is the case somehow with the updated dependency of calk in ts-jest.

package.json
ahmehri: @tdeekens meant to point out a tip for me to be able to update cypress to v4 in mc. Thanks.

tdeekens: If you install two versions of cypress without a resolution your download of cypress seems to die.

tdeekens:
@emmenko this should be good now as discussed. Had to rebase. The last 1-2 commits are relevant.

tdeekens:
I presume the 👍 up was an approval. I will proceede to isolate some other dep updates where possible.

PR #1326, opened 2020-02-10
chore: update flopflip

tdeekens:
#### Summary

Contains a fix in the types :). This should compile ⛱.


tdeekens:
I really am unsure what's wrong with the typing here. flopflip says ToggleFeature returns a React.ReactNode. This is also what react-router returns from all their render/children etc props. Wrapping into a `<>` also doens't help. Must be something with flopflip but don't know what. Maybe @emmenko has an idea.

Maybe something is wrong with

```ts

import { TFlagName, TFlagVariation } from '@flopflip/types';
--
2 | import React from 'react';
3 | declare type Props = {
4 | flag: TFlagName;
5 | variation: TFlagVariation;
6 | };
7 | declare const ToggleFeature: {
8 | <OwnProps extends Props>(props: OwnProps): React.ReactNode;
9 | displayName: string;
10 | };
11 | export default ToggleFeature;


```

emmenko:
I can take a look later

tdeekens:
That'd be super nice. I can't see the forrest anymore. This can also be relevant: https://github.com/tdeekens/flopflip/blob/master/packages/react-broadcast/modules/components/toggle-feature/toggle-feature.tsx#L17

emmenko:
It seems that the type declarations changed between versions. See screenshot below (on top is the "old" declaration, at the bottom the "new" one).

image

emmenko:
I think `children` is missing in the props, that's why it's complaining.

Also the type declarations for the appkit components return a `JSX.Element`, and `children` returns a `React.ReactNode`. Maybe this is also part of the problem. For example:

image

I'll take a look at flopflip and in case make a PR.

tdeekens:
Thanks for investigating. Yeah, type previous types were wrong and somebody created an issue reguarding that.

emmenko:
I opened a PR in flopflip: https://github.com/tdeekens/flopflip/pull/977

PR #1327, opened 2020-02-11
chore: add skipping of mc api url inferring process

tdeekens:
#### Summary

Internally sometimes we run the MC on another origin than the underlying mc-api url requires (e.g. branch deployments). In this case we set this property `skipInferringOfApiUrlOnProduction`.


tdeekens:
It's now `disableInferringOfApiUrlOnProduction` over `skipInferringOfApiUrlOnProduction`. Feels more clear to me.

PR #1332, opened 2020-02-13
chore: update flopflip and promster

tdeekens:
#### Summary

This versions comes with the fix where under test we can run multiple versions of flopflip without getting state updates errors due to components being unmounted.


tdeekens:
Why does it fail on CI with this now. This part didn't change at all.

CleanShot 2020-02-13 at 11 00 22@2x

tdeekens:
@emmenko could this be caused by the PR you made with the proxy exports? It's the only one I can think of which could have affected this 😕 https://github.com/tdeekens/flopflip/pull/977/files

emmenko:
Hmm can you check what is installed in node_modules? All packages only list the `dist` folder to be published, no other file should be included in the bundle.

emmenko:
Ah I see, it's because you are importing the status variables from the types package

emmenko:
image

emmenko:
https://github.com/tdeekens/flopflip/blob/782acd2636ba0c6a2290b700338f72ec723bc2a9/packages/types/index.ts#L10-L13

tdeekens:
> Hmm can you check what is installed in node_modules? All packages only list the dist folder to be published, no other file should be included in the bundle.

Yes, I debugged it. Doesn't seem to be the proxy export change as it's not published at all. Must be an earlier change of mine. Maybe importing something which isn't transpiled. Not 100% sure yet.

tdeekens:
Ah, just read your last message. Yup. Either redeclare it to do something to the types package.

emmenko:
Not sure what the best way to fix this is. Maybe try running the `types` through rollup?

emmenko:
Wondering also how this didn't get caught in the flopflip tests 🤔

tdeekens:
Not too sure. I guess cause only node_modules are not transformed and for flopflip its own stuff isn't node_modules.

https://github.com/tdeekens/flopflip/pull/981

tdeekens:
Huh. Why now 😄 

CleanShot 2020-02-13 at 12 51 37@2x

tdeekens:
My guess is that SIGTERM is sent to the server and lightship would terminate the metrics server but it takes some time. I see two options:

1. Not start it on CI
2. Add a delay here to make sure the server has shut down

Why this happens now: not sure. I'll start with 2. and see.

packages/mc-http-server/server.js
```diff
@@ -178,6 +178,7 @@ const launchServer = async options => {
     });
     prometheusMetricsServer = await createPrometheusMetricsServer({
       port: prometheusMetricsServerPort,
+      detectKubernetes: true,
```
tdeekens: Let's see. Basically the new version of promster is better as it does work a bit like lightship. In that it won't start locally if this flag is passed.

TIL: passing `undefined` to `listen` seems to silently not have a server start 🤷‍♂. At least that's what lightship does.

tdeekens:
CleanShot 2020-02-13 at 14 23 27@2x

Nice, now promster is smart enough but the other thing not. FML. FMW. 😄

emmenko:
Hmm it seems that the cypress action does not "clean up" the started server process 🤔

By looking at the action implementation I'm now wondering if that's even built in...It might be that it's missing to stop the server after the tests run.

See also https://github.com/cypress-io/github-action/blob/057da53cea49223c67bb67d581666752fb578ae9/index.js#L510-L513

emmenko:
So it might be that it's a "missing feature" of the actions toolkit.

I'm wondering how it's been working so far...just luck maybe?

I guess for now we should try to "kill" the processes listening on port 3001 between multiple cypress runs.

tdeekens:
Yeah. That‘s also why I payed with the sleep and only reverted it. Either we kill what listens on 3001 (as other servers are not spawned now or we can have the server run in the background for subsequent steps.

emmenko:
https://github.com/commercetools/merchant-center-application-kit/pull/1334

tdeekens:
Yup just did the same with a `kill -9 $(lsof -t -i:3001)`.

tdeekens:
Incredible. It works. Still can't believe that I only ran into this on this branch.

PR #1333, opened 2020-02-13
docs: testing.mdx

rudavko:
Add missing left bracket to the link markdown

#### Summary

Add a missing left bracket for the link markdown


tdeekens:
The tests are only failing cause we have had an issue with long-running processes on CI which is solved now. I will merge this later.

PR #1334, opened 2020-02-13
fix(actions): kill server process running the app on port 3001 between cypress runs

emmenko:
An attempt to fix this: https://github.com/commercetools/merchant-center-application-kit/pull/1332#issuecomment-585718980

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -108,6 +108,12 @@ jobs:
           CYPRESS_LOGIN_PASSWORD: ${{ secrets.CYPRESS_LOGIN_PASSWORD }}
           CYPRESS_PROJECT_KEY: ${{ secrets. CYPRESS_PROJECT_KEY }}
 
+      # It seems that sometimes the application server started by Cypress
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -108,6 +108,12 @@ jobs:
           CYPRESS_LOGIN_PASSWORD: ${{ secrets.CYPRESS_LOGIN_PASSWORD }}
           CYPRESS_PROJECT_KEY: ${{ secrets. CYPRESS_PROJECT_KEY }}
 
+      # It seems that sometimes the application server started by Cypress
+      # does not get cleaned up properly.
+      # See also https://github.com/actions/toolkit/issues/216
+      - name: Kill possible process running the application on port 3001
```
tdeekens: Thanks. I just commited the same 😄. Arent you on sick leave 🙉.

tdeekens:
We could also kill the second server on the second run...

tdeekens:
I'll merge this to rebase my branch over it. We can change things there if we want to :). Thanks!

PR #1335, opened 2020-02-13
fix(actions-global): method signature of useOnActionError

pa3:
`useOnActionError` implementation looks wrong to me. [This doc](https://github.com/commercetools/merchant-center-application-kit/blob/master/website/src/content/development/testing.mdx) supports my feelings that its current API is not what was actually desired. So, here is the fix.

Old (wrong) API:
```js
const error = new Error();
const onActionError = useOnActionError(error);

...

// error happened here
onActionError();
```

New (fixed) API:
```js

const onActionError = useOnActionError();

...

// error happened here
const error = new Error();
onActionError(error);
```

### DIFF
--- a/packages/actions-global/src/actions/handle-action-error.ts
+++ b/packages/actions-global/src/actions/handle-action-error.ts
@@ -1,4 +1,4 @@
-import { Dispatch } from 'redux';
+import { ThunkDispatch } from 'redux-thunk';
 import {
   STATUS_CODES,
   LOGOUT_REASONS,
@@ -22,7 +22,9 @@ type ApiError = {
 };
 
 export type ActionError = Error | ApiError;
-export type DispatchActionError = Dispatch<
+export type DispatchActionError = ThunkDispatch<
+  null,
+  null,
   | ReturnType<typeof showApiErrorNotification>
   | ReturnType<typeof showUnexpectedErrorNotification>
 >;
--- a/packages/actions-global/src/hooks/use-on-action-error.spec.tsx
+++ b/packages/actions-global/src/hooks/use-on-action-error.spec.tsx
@@ -1,0 +1,40 @@
+import React from 'react';
+import { Action } from 'redux';
+import { Provider } from 'react-redux';
+import { renderHook } from '@testing-library/react-hooks';
+import useOnActionError from './use-on-action-error';
+
+jest.mock('../actions/handle-action-error', () => (payload: Error) => ({
+  type: 'fake-handle-action-error',
+  payload,
+}));
+
+const renderUseOnActionError = () => {
+  const dispatchedActions: Array<Action> = [];
+  const fakeStore = {
+    dispatch: (action: Action) => dispatchedActions.push(action),
+    getState: () => null,
+    subscribe: () => null,
+  };
+  const { result } = renderHook(() => useOnActionError(), {
+    // eslint-disable-next-line react/display-name
+    wrapper: ({ children }) => (
+      // @ts-ignore: partially implemented Redux Store is OK as a mock
+      <Provider store={fakeStore}>{children}</Provider>
+    ),
+  });
+  return { onActionError: result.current, dispatchedActions };
+};
+
+describe('useOnActionError', () => {
+  it('dispatches `handleActionError` with a given error as an payload', () => {
+    const { onActionError, dispatchedActions } = renderUseOnActionError();
+
+    const error = new Error('oupsy!');
+    onActionError(error);
+
+    expect(dispatchedActions).toMatchObject([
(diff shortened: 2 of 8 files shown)


packages/actions-global/src/hooks/use-on-action-error.ts
```diff
@@ -1,14 +1,11 @@
 import React from 'react';
 import { useDispatch } from 'react-redux';
-import handleActionError, {
-  ActionError,
-  DispatchActionError,
-} from '../actions/handle-action-error';
+import handleActionError, { ActionError } from '../actions/handle-action-error';
 
-export default function useOnActionError(error: ActionError) {
-  const dispatch = useDispatch<DispatchActionError>();
```
emmenko: The type should stay. Why did you need to remove it?

pa3: Can return it for sure. Just wonder, why explicitly putting types in the internal code (which is not a part of a public API) if TS can use its type-inference super-powers to do the job with less typing from my side?

packages/actions-global/src/hooks/use-on-action-error.spec.tsx
```diff
@@ -0,0 +1,40 @@
...
+  };
+  const { result } = renderHook(() => useOnActionError(), {
+    // eslint-disable-next-line react/display-name
+    wrapper: ({ children }) => (
+      // @ts-ignore: partially implemented Redux Store is OK as a mock
+      <Provider store={fakeStore}>{children}</Provider>
+    ),
+  });
+  return { onActionError: result.current, dispatchedActions };
+};
```
tdeekens: Interesting and concise way of testing it. Would have probably tested the whole app to see if something appears on screen. Or use the good old `redux-mock-store`.

Anyway, all cool! :)

packages/actions-global/src/hooks/use-on-action-error.ts
tdeekens: I think with this type annotation TS wouldbe be able to infer what the useCallback and dispatch actually returns. Have you checked?

pa3: @tdeekens just tried doing the following with the changes from this PR (i.e. with dropped typing inside the function body):

```js
const onActionError = useOnActionError();
onActionError('oups');
```

Got this:

Screenshot 2020-02-13 at 22 16 28

pa3:
@emmenko @tdeekens OK, I put `DispatchActionError` where it was before. That required fixing obviously wrong declaration of that type itself. As we are using `redux-thunk` the type of `dispatch` function is not just `Dispatch`, but `ThunkDispatch`.

packages/actions-global/src/actions/handle-action-error.ts
```diff
@@ -1,4 +1,4 @@
-import { Dispatch } from 'redux';
+import { ThunkDispatch } from 'redux-thunk';
```
emmenko: I'm not sure why you would need this, because we're dispatching "normal actions" and we just need the normal `dispatch` function.

The thunk that you maybe refer to is `handleActionError`, but the type here refers to the `dispatch` function used by the thunk.

pa3: @emmenko Can you please take over this PR? Because, looks like you have a certain idea in mind of how that should be implemented. So, probably will be just faster for you to do that. Thanks! ❤️

emmenko: Just revert this import, the rest is fine 😄

pa3: That will cause type check errors. I fixed that type for a reason.

emmenko: Ah because in the hook you now call the action like `dispatch(handleActionError(error))` instead of `handleActionError(error)(dispatch)`.

Ok sorry, I missed that.

We need to include the `redux-thunk` as a dependency though.

packages/actions-global/src/hooks/use-on-action-error.ts
emmenko: All good, the types are correctly inferred. Thanks for double checking!
I dropped the type params from the other hooks as well. https://github.com/commercetools/merchant-center-application-kit/pull/1335/commits/99d69fef251afbd8e8673c67548af1529364e2fb

packages/actions-global/src/actions/handle-action-error.ts
emmenko: https://github.com/commercetools/merchant-center-application-kit/pull/1335/commits/99d69fef251afbd8e8673c67548af1529364e2fb

PR #1340, opened 2020-02-14
test: fix missing async tests and remove unnecessary usage of beforeEach

emmenko:
On github actions, there is no `CI` environment variable. Without this, Jest won't fail the tests in case there are console logs.

I noticed that there are still problems with flopflip state updates.

image

cc @tdeekens


tdeekens:
Thanks. You can also downgrade flopflip to the version before here. There was one where it was a class component. I am out of ideas and energy right now to pick this up again soon in my spare time.

emmenko:
I fixed all the test setup, both `beforeEach` and missing `async` tests.

You can read more about the investigation here: https://github.com/tdeekens/flopflip/issues/982#issuecomment-586633958

emmenko:
cc @commercetools/fe-chapter

packages/application-shell/src/components/project-switcher/project-switcher-test-utils.ts
```diff
@@ -1,13 +1,19 @@
 import { GRAPHQL_TARGETS } from '@commercetools-frontend/constants';
 import ProjectsQuery from './project-switcher.mc.graphql';
 
+type CreateGraphqlResponseForProjectsQueryOptions = {
```
islam3zzat: Nitpick: I'm not deeply familiar with the AppKit, but if the preceding `T` for types like done a previous change (from `Props` to `TProps`) is a convention that u have here, should we keep it here?

emmenko: Usually we prefix the types with `T` when the types are exported and used somewhere else. For types that are only used within the module itself, it's not a must.
We also don't have a strict convention but that's more or less what we've been doing so far.

packages/application-shell/src/components/back-to-project/back-to-project.spec.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+beforeEach(() => {
+  window.location.replace = jest.fn();
+});
+
+describe('with `projectKey`', () => {
+  it('should redirect to the project with the key', async () => {
+    const rendered = renderApp(<BackToProject projectKey="test-project-key" />);
+    fireEvent.click(rendered.getByText('Back to project'));
+    await wait(() => {
```
tdeekens: It's a bit odd that this works to me. Likely just cause it puts things on the next tick of the event queue and then the warning isn't thrown.

packages/application-shell/src/components/project-switcher/project-switcher-test-utils.ts
tdeekens: I actually like having all types start with T. Internal to file or not. Maybe just me 😄

packages/application-shell/src/test-utils/test-utils.spec.tsx
```diff
@@ -374,14 +370,14 @@ describe('custom render functions', () => {
       const rendered = renderAppWithRedux(
         <TestComponent>{'one'}</TestComponent>
       );
-      rendered.getByText('one');
+      await rendered.findByText('one');
 
       rendered.rerender(<TestComponent>{'two'}</TestComponent>);
-      rendered.getByText('two');
+      await rendered.findByText('two');
```
tdeekens: Thanks.

PR #1346, opened 2020-02-17
docs(website): improve docs around csp and cloud regions

### LINKED ISSUE #1343: Improve documentation of CSP config
For instance, the MC API domain needs to be included in the `connect-src`.

At the moment, the legacy GCP-EU/US hostnames are included by default:

https://github.com/commercetools/merchant-center-application-kit/blob/40230d0606e1358eaa535adbc6bf1c19c1652857/packages/mc-html-template/lib/load-headers.js#L86-L87

This causes more confusion and we should probably remove those defaults and make it clear that people need to explicitly add those domains.

emmenko:
Closes #1343 

This PR aims to clarify how to configure the CSP directives, especially around the new hostnames.

### DIFF
--- a/website/src/content/deployment/runtime-configuration.mdx
+++ b/website/src/content/deployment/runtime-configuration.mdx
@@ -74,7 +74,7 @@ Therefore, it's important that the application is [registered](/register-applica
 
 </Warning>
 
-It is recommended to have two separate configuration files, one for development and one for production. The `env.json` **must** be used for development, therefore for production you can have a `env.prod.json` file.
+It is recommended to have two separate configuration files, one for development and one for production. The `env.json` **must** be used for development, therefore for production you can have an `env.prod.json` file.
 
 # `headers.json`
 
@@ -88,7 +88,7 @@ For development the `headers.json` file is **optional**. If used, the file must 
 
 ## Content Security Policy
 
-The Content Security Policy (CSP) header allows to instruct the browser how to deal with security measures, according to the configured instructions. Most of the times it's enough to whitelist the hostnames that the Custom Application uses, for example if [static assets](/deployment/serving-static-assets) are served by a CDN.
+The Content Security Policy (CSP) header allows to instruct the browser how to deal with security measures, according to the configured instructions. Most of the time it's enough to whitelist the hostnames that the Custom Application uses, for example if [static assets](/deployment/serving-static-assets) are served by a CDN.
 
 <Info>
 
@@ -115,7 +115,7 @@ For example, given that your application is hosted at `my-app.now.sh` and runs o
 
 <Info>
 
-This example includes two hostnames as one of them is a deprecated hostname: `mc.commercetools.com`. For more information read about [Deprecated hostnames](/main-concepts/api-gateway#deprecated-hostnames).
+This example includes two hostnames as one of them is a deprecated hostname: `mc-api.commercetools.com`. For more information read about [Deprecated hostnames](/main-concepts/api-gateway#deprecated-hostnames).
 
 </Info>
 
--- a/website/src/content/deployment/runtime-configuration.mdx
+++ b/website/src/content/deployment/runtime-configuration.mdx
@@ -6,38 +6,48 @@ Merchant Center applications require a runtime configuration to work. This is us
 
 # `env.json`
 
-The configuration is defined in a file named `env.json`. The following fields are **required** but you can provide additional fields specific to your application:
+The configuration is defined in a file named `env.json`.
 
-- `applicationName`: the name of the application (usually the same as in `package.json`)
-- `frontendHost`: the host where the Merchant Center Custom Application is running (for example `mc.commercetools.com`)
-- `mcApiUrl`: the API URL for the [Merchant Center API Gateway](/main-concepts/api-gateway)
-- `location`: the location where the Merchant Center Custom Application is running. This can be something like:
-  - `eu` for projects in **Europe**
-  - `us` for projects in the **United States**
-- `env`: the environment where the Merchant Center Custom Application is running, usually `production` or `development`
-- `cdnUrl`: the URL where the static assets are stored (see [Serving Static Assets](/deployment/serving-static-assets))
-- `servedByProxy`: a flag to indicate if the application is running behind the Merchant Center proxy or not. This is either:
-  - `true` for **production**
-  - `false` for **local development**
+<Info>
 
-The `env.json` object will eventually be injected into `window.app`, when the `index.html.template` gets compiled into `index.html`.
-The `window.app` object is then passed to the `<ApplicationShell>` component as an `environment` prop and will be available in the application context:
+For development the `env.json` file must be in the root path of the project as it's automatically loaded by the Webpack dev server.
 
-```jsx
-import { ApplicationContext } from '@commercetools-frontend/application-shell-connectors';
+</Info>
 
-const MyComponent = () => (
-  <ApplicationContext
-    render={({ environment }) => <div>{environment.applicationName}</div>}
+The following fields are **required** but you can provide additional fields specific to your application:
+
+- `applicationName`: the name of the application (usually the same as in `package.json`).
+- `frontendHost`: the host where the Merchant Center Custom Application is running. See [Merchant Center hostnames](/main-concepts/api-gateway#hostnames).
+- `mcApiUrl`: the API URL for the [Merchant Center API Gateway](/main-concepts/api-gateway).
+- `location`: the location where the Merchant Center Custom Application is running. This value is used for logging and error reporting. Usually values refer to the cloud region where the Custom Application is running, for example `gcp-eu`, `gcp-us`, `aws-ohio`, etc.
+- `env`: the environment value where the Merchant Center Custom Application is running, usually `production` or `development`.
+- `cdnUrl`: the URL where the static assets are stored (see [Serving Static Assets](/deployment/serving-static-assets)).
+- `servedByProxy`: a flag to indicate if the application is running behind the Merchant Center proxy router or not. This is either `true` for **production** and `false` for **local development**.
+
+When the `index.html.template` gets compiled into `index.html`, the `env.json` is injected into a global variable `window.app`.
+This value should be used in the `<EntryPoint>` component when rendering the `<ApplicationShell>`.
+
+```jsx highlightLines="3"
+const EntryPoint = () => (
+  <ApplicationShell
+    environment={window.app}
+    // other props
   />
 );
 ```
 
-<Info>
+The environment object is parsed and passed into a React Context, making it available to the entire application. To access it, use the `@commercetools-frontend/application-shell-connectors` package.
 
-For `development` the `env.json` file must be in the root path of the project as it's automatically loaded by the webpack dev server.
+```jsx
+import { useApplicationContext } from '@commercetools-frontend/application-shell-connectors';
 
-</Info>
+const MyComponent = () => {
+  const applicationName = useApplicationContext(
+    context => context.environment.applicationName
+  );
+  return (<div>{`Welcome to the application ${applicationName}!`}</div>);
+};
+```
 
 An example configuration for local development:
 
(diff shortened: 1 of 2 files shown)


website/src/content/deployment/runtime-configuration.mdx
```diff
@@ -53,38 +63,79 @@ An example configuration for local development:
...
+Therefore, it's important that the application is [registered](/register-applications/configuring-a-custom-application), runs behind the proxy router, and that the `servedByProxy` option in the `env.json` is set to `true`.
 
 </Warning>
 
-To migrate to the new format, wrap the `csp.json` content into a `csp` property in the `headers.json`:
+It is recommended to have two separate configuration files, one for development and one for production. The `env.json` **must** be used for development, therefore for production you can have a `env.prod.json` file.
+
+# `headers.json`
+
+This configuration file allows to configure HTTP headers. The most important one is the [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) header.
```
adinakleine:
```suggestion
This configuration file allows you to configure HTTP headers. The most important one is the [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) header.
```

website/src/content/deployment/runtime-configuration.mdx
```diff
@@ -53,38 +63,79 @@ An example configuration for local development:
...
 
-This file has been deprecated by the `headers.json` mentioned below.
+When the application is deployed on a production environment, the application will fail to make requests to the [API Gateway](/main-concepts/api-gateway). This is due to the fact that the domain, where the application is hosted, is not whitelisted by the CORS rules of the API Gateway.
+
+Therefore, it's important that the application is [registered](/register-applications/configuring-a-custom-application), runs behind the proxy router, and that the `servedByProxy` option in the `env.json` is set to `true`.
 
 </Warning>
 
-To migrate to the new format, wrap the `csp.json` content into a `csp` property in the `headers.json`:
+It is recommended to have two separate configuration files, one for development and one for production. The `env.json` **must** be used for development, therefore for production you can have a `env.prod.json` file.
```
adinakleine:
```suggestion
It is recommended to have two separate configuration files, one for development and one for production. The `env.json` **must** be used for development, therefore for production you can have an `env.prod.json` file.
```

website/src/content/deployment/runtime-configuration.mdx
```diff
@@ -53,38 +63,79 @@ An example configuration for local development:
...
+
+<Info>
+
+For development the `headers.json` file is **optional**. If used, the file must be in the root path of the project as it's automatically loaded by the Webpack dev server.
+
+</Info>
+
+## Content Security Policy
+
+The Content Security Policy (CSP) header allows to instruct the browser how to deal with security measures, according to the configured instructions. Most of the times it's enough to whitelist the hostnames that the Custom Application uses, for example if [static assets](/deployment/serving-static-assets) are served by a CDN.
```
adinakleine:
```suggestion
The Content Security Policy (CSP) header allows you to instruct the browser how to deal with security measures, according to the configured instructions. Most of the time it's enough to whitelist the hostnames that the Custom Application uses, for example if [static assets](/deployment/serving-static-assets) are served by a CDN.
```

tdeekens:
> I request changes because I assume that there is an important difference between mc.commercetools.com and mc-api.commercetools.com. Right?

Yup. The mc.commercetools.com is the MC as in where the app (Single Page Application) is hosted and accessible. The mc-api.commercetools.com is the API the SPA talks to.

website/src/content/main-concepts/api-gateway.mdx
```diff
@@ -10,9 +10,61 @@ To solve these issues, the Merchant Center provides an **API Gateway**, which pr
...
+
+<Info>
+
+The Merchant Center runs on the `mc` subdomain, whereas the API Gateway runs on the `mc-api` subdomain.
+
+</Info>
+
+| Cloud region | Hostname |
+| --- | --- |
+| North America (Google Cloud, Iowa) | `mc-api.us-central1.gcp.commercetools.com` |
```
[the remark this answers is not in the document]
emmenko: To be honest this what’s currently in the API docs. Can you check with Nikolaus?

website/src/content/main-concepts/api-gateway.mdx
```diff
@@ -10,9 +10,61 @@ To solve these issues, the Merchant Center provides an **API Gateway**, which pr
 
 ![HTTP API](/images/mc-http-api.png)
 
-All Merchant Center applications should send requests through the API Gateway, which is available at the same domain of the Merchant Center but with the `mc` subdomain being `mc-api`.
+<Info>
 
-Official domains:
+It is recommended that Custom Applications use the API Gateway to access the commercetools APIs, for security reasons. See [Proxy Endpoints](/main-concepts/proxy-endpoints).
```
adnasa:
```suggestion
For security reasons, it is recommended that Custom Applications use the API Gateway to access the commercetools APIs. See [Proxy Endpoints](/main-concepts/proxy-endpoints).
```

website/src/content/main-concepts/api-gateway.mdx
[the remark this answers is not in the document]
adinakleine: Sure, it's just a suggestions... I assume Celeste set this rule. 
I will discuss it with Nikolaus and come back to you.

@nkuehn: What is your opinion on this. Just *Google Cloud* or *Google Cloud Platform*. 
https://wiki.commercetools.com/display/DOCS/Product+naming+conventions

nkuehn:
@adinakleine  I'm sorry that you fell for an inconsistency in our writing style documents concerning the Google Cloud based Region naming.  Please also try to not link to internal wiki locations on open source repositories like this here,  there's a page "Cloud Region Naming" that is authoritative. 

@emmenko  did it correctly and consistently with the recently released API documentation: https://docs.commercetools.com/http-api#hosts    We chose to use the established everyday language "Google Cloud" without the "Platform" and also chose to use "AWS" in that case and not the fully spelled out because in that case the acronym has become the de facto everyday language use.

website/src/content/deployment/runtime-configuration.mdx
nkuehn: @adinakleine  the you is not necessary and I am relatively sure that it's not our typical style until now, too.

website/src/content/main-concepts/api-gateway.mdx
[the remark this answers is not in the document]
nkuehn: see my comment below covering the whole review.

website/src/content/main-concepts/api-gateway.mdx
```diff
@@ -10,9 +10,61 @@ To solve these issues, the Merchant Center provides an **API Gateway**, which pr
...
+| North America (Google Cloud, Iowa) | `mc-api.us-central1.gcp.commercetools.com` |
+| North America (AWS, Ohio) | `mc-api.us-east-2.aws.commercetools.com` |
+| Europe (Google Cloud, Belgium) | `mc-api.europe-west1.gcp.commercetools.com` |
+| Europe (AWS, Frankfurt) | `mc-api.eu-central-1.aws.commercetools.com` |
+
+## Deprecated hostnames
+
+Hostnames for each cloud region follow a specific naming format and therefore any hostname that does not comply to that is considered as deprecated.
+
+Current deprecated hostnames are:
```
nkuehn: The HTTP API docs use "legacy" instead of "deprecated",  we can choose either but should be consistent ( https://docs.commercetools.com/http-api#hosts )   

Overall I think the section listing all the previous host names is visually too dominant and big, it would be good to find a format that does not make it look like the key part of the docs.   It's not even strictly necessary to fully document the previous state at all.

emmenko: Thanks, I'll update that

emmenko:
@nkuehn can you take another look at the hostname section? I restructured that a bit.

website/src/content/main-concepts/api-gateway.mdx
```diff
@@ -10,9 +10,52 @@ To solve these issues, the Merchant Center provides an **API Gateway**, which pr
...
+| Cloud region | Merchant Center API Gateway hostname |
+| --- | --- |
+| North America (Google Cloud, Iowa) | `mc-api.us-central1.gcp.commercetools.com` |
+| North America (AWS, Ohio) | `mc-api.us-east-2.aws.commercetools.com` |
+| Europe (Google Cloud, Belgium) | `mc-api.europe-west1.gcp.commercetools.com` |
+| Europe (AWS, Frankfurt) | `mc-api.eu-central-1.aws.commercetools.com` |
+
+### Legacy hostnames
+
+Any hostname that does not follow the new naming format is considered a **legacy hostname**.
```
nkuehn: Thanks for the compact proposal!  When speaking to  @adinakleine  I realized that we have to keep the information what legacy hostname pointed to what region because it's not actually transparent whether to map ex-EU to GCP or AWS.  So here's a proposal (I'll propose the same change to the HTTP API docs)

```suggestion
Any hostname that does not follow the new naming format is considered a **legacy hostname** but remains available as an alias. `mc{-api}.commercetools.com` is addressing the Google Cloud Belgium Region,  `mc{-api}.commercetools.co` the Google Cloud Iowa Region. 
```

website/src/content/deployment/runtime-configuration.mdx
```diff
@@ -53,38 +63,79 @@ An example configuration for local development:
...
 
 ```json
 {
   "csp": {
-    "script-src": ["my-apps.com"],
-    "connect-src": ["my-apps.com"],
-    "style-src": ["my-apps.com"]
+    "script-src": ["my-app.now.sh"],
+    "connect-src": [
+      "my-app.now.sh",
```
nkuehn: If we propose this as a copy-paste example including the specific "my-app.now.sh"  I suggest to actually reserve this now.sh  domain in our account (it's free) to avoid potential future attempts to hijack custom applications that were created from our boilerplate, installed but never finished.  

I could be a static 200 reponse on everyting or something like that (e.g. just a now.json file)

(not necessary to merge this PR, just an idea)

website/src/content/main-concepts/api-gateway.mdx
nkuehn: @emmenko  this grammar change is correct and should be applied.   

Plus, I would not even use "recommended" but prescribe it in an "it's required to" so you have to chance to actually filter out direct API hosts in the API gateway CSP header generators.

website/src/content/main-concepts/api-gateway.mdx
```diff
@@ -10,9 +10,52 @@ To solve these issues, the Merchant Center provides an **API Gateway**, which pr
```
nkuehn: To let the index navigation better represent the actual document structure I suggest to give the big image a heading.  You can pick something else than what I propse here, but a heading would be good. 

```suggestion

# Architecture Overview

```

website/src/content/deployment/runtime-configuration.mdx
emmenko: Hmm I see your point but I wouldn't worry much about it. The name `my-app` is very generic and I don't expect people to use that.
This example is also not really meant to be a copy-paste. The deployments examples are more likely to do that.

website/src/content/main-concepts/api-gateway.mdx
emmenko: This is a bit tricky. We would probably not alias it because of the cookie issue, at least until we figure it out. So I would leave it out for now.

emmenko:
Thanks for the feedback so far. I'll merge this, any further improvements can be done separately.

website/src/content/main-concepts/api-gateway.mdx
nkuehn: @emmenko  you feedback throws both aspects of this change into a bucket.   You can decide not alias like we do in the API hosts, but you still need to tell people what new "Region" name "mc.commercetools.com" originally was and the same for .co.

emmenko: Ok, I thought you didn't want to mention that. Will open a new PR

PR #1349, opened 2020-02-18
docs: add deployment example for ZEIT Now v2, restructure other examples.

emmenko:
This PR documents the deployment example to Now v2.
The other examples have also been restructured a bit to follow the same format and because they share some of the same setup.


packages/jest-preset-mc-app/jest-preset.js
```diff
@@ -29,7 +29,7 @@ module.exports = {
   setupFilesAfterEnv: [resolveRelativePath('./setup-test-framework.js')],
   snapshotSerializers: ['enzyme-to-json/serializer'],
   testEnvironment: 'jsdom',
```
tdeekens: In theory this is a breaking change creeping in here. It will change tests (and does on the MC internally) and a user of the package has to change something. Making it crossing borders of being a breaking change 😺💀

emmenko: You mean the `testURL`?

emmenko: Yeah good catch, I’ll revert that. It does not matter much to use the new hostname for tests.

tdeekens: We can also keep it now that the change is made but add it to the changelog as to how to migrate.

PR #1358, opened 2020-02-22
fix(ci): to respect skip ci

tdeekens:
#### Summary

Played with this a bit on flopflip. Seems the only viable way for now to support skip ci which I think is nice. Just a suggestion.


emmenko:
The difference and usage of push and pull_request events is still a mystery to me to be honest...

tdeekens:
Ok cool. Worthy a shot.

PR #1368, opened 2020-02-28
bug(app-shell-connectors): path mcApiUrl on context env

tdeekens:
> I closed the previous PR and opened another one cause this affects another package.

#### Summary

Please refer to https://github.com/commercetools/merchant-center-application-kit/pull/1367 for a more detailed reasoning.

#### Description

Instead of exporting the function we patch the environment.

Note, that we can't export the function from the application-shell so we duplicate it again.

A third option is to overwrite/patch the window itself upon read/boot but I don't think that's really nice.

### DIFF
--- a/packages/application-shell-connectors/src/components/application-context/application-context.spec.tsx
+++ b/packages/application-shell-connectors/src/components/application-context/application-context.spec.tsx
@@ -7,6 +7,7 @@ import {
   withApplicationContext,
   mapUserToApplicationContextUser,
   mapProjectToApplicationContextProject,
+  mapEnvironmentToApplicationContextEnvironment,
 } from './application-context';
 import { ApplicationWindow } from '@commercetools-frontend/constants';
 
@@ -167,6 +168,24 @@ describe('mapProjectToApplicationContextProject', () => {
       currencies: expect.any(Array),
       languages: expect.any(Array),
       ownerId: expect.any(String),
+    });
+  });
+});
+
+describe('mapEnvironmentToApplicationContextEnvironment', () => {
+  it('should map environment to environment context', () => {
+    expect(
+      mapEnvironmentToApplicationContextEnvironment(createTestEnvironment())
+    ).toEqual({
+      revision: expect.any(String),
+      applicationName: expect.any(String),
+      frontendHost: expect.any(String),
+      mcApiUrl: 'https://mc-api.commercetools.com',
+      location: expect.any(String),
+      env: expect.any(String),
+      cdnUrl: expect.any(String),
+      servedByProxy: expect.any(Boolean),
+      foo: expect.any(String),
     });
   });
 });
--- a/packages/application-shell-connectors/src/components/application-context/application-context.spec.tsx
+++ b/packages/application-shell-connectors/src/components/application-context/application-context.spec.tsx
@@ -173,19 +173,44 @@ describe('mapProjectToApplicationContextProject', () => {
 });
 
 describe('mapEnvironmentToApplicationContextEnvironment', () => {
-  it('should map environment to environment context', () => {
-    expect(
-      mapEnvironmentToApplicationContextEnvironment(createTestEnvironment())
-    ).toEqual({
-      revision: expect.any(String),
-      applicationName: expect.any(String),
-      frontendHost: expect.any(String),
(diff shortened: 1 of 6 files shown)


packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -76,6 +77,16 @@ export const mapUserToApplicationContextUser = (user?: TFetchedUser) => {
   };
 };
 
+// Adjust certain fields which depend e.g. on the origin
+export const mapEnvironmentToApplicationContextEnvironment = <
+  TApplicationEnvironment extends {}
+>(
+  environment: TApplicationEnvironment
+) => ({
+  mcApiUrl: getMcApiUrl(),
```
emmenko: Shouldn't this be after `...environment`?

Also, it might be more clear to pass the `environment` object to the function `getMcApiUrl(environment)` instead of using the `window.app` value.

tdeekens: Yeah. Can adjust the fn in this case. Too late now. Will do tmr.

packages/application-shell-connectors/src/components/application-context/application-context.spec.tsx
```diff
@@ -173,19 +173,44 @@ describe('mapProjectToApplicationContextProject', () => {
...
-      revision: expect.any(String),
-      applicationName: expect.any(String),
-      frontendHost: expect.any(String),
-      mcApiUrl: 'https://mc-api.commercetools.com',
-      location: expect.any(String),
-      env: expect.any(String),
-      cdnUrl: expect.any(String),
-      servedByProxy: expect.any(Boolean),
-      foo: expect.any(String),
+  describe('when application is configured to run behind proxy', () => {
```
tdeekens: I felt this kind of narrow "integration" test is needed. The other permutations are unit tested in the the module.

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -79,12 +80,14 @@ export const mapUserToApplicationContextUser = (user?: TFetchedUser) => {
 
 // Adjust certain fields which depend e.g. on the origin
 export const mapEnvironmentToApplicationContextEnvironment = <
-  TApplicationEnvironment extends {}
+  AdditionalEnvironmentProperties extends {}
 >(
-  environment: TApplicationEnvironment
+  environment: AdditionalEnvironmentProperties & TApplicationContextEnvironment,
+  partialWindow: Pick<Window, 'origin'> = window
```
tdeekens: Otherwise the type is lying/requiring too much I think.

packages/constants/src/constants.ts
```diff
@@ -151,5 +151,6 @@ export interface ApplicationWindow extends Window {
     enableSignUp?: boolean;
     disabledMenuItems?: string[];
     useFullRedirectsForLinks?: boolean;
+    disableInferringOfMcApiUrlOnProduction?: string;
```
tdeekens: This is needed now I think as we access the property above.

emmenko: This should be a boolean

tdeekens: My bad. There we go packages/constants/src/constants.ts

tdeekens: My bad. There we go 31113e9

packages/application-shell-connectors/src/utils/get-mc-api-url/get-mc-api-url.ts
```diff
@@ -39,7 +39,11 @@ export default function getMcApiUrl(
   const isInferringOfMcApiUrlOnProductionDisabled =
     environment.disableInferringOfMcApiUrlOnProduction;
 
-  if (isServedByProxy && !isInferringOfMcApiUrlOnProductionDisabled)
+  if (
+    isServedByProxy &&
+    !isInferringOfMcApiUrlOnProductionDisabled &&
+    partialWindow.origin
```
tdeekens: Under test we don't have an window.origin but we still want to test things without caring about the mcApiUrl. Hence, the app-shell tests broke.

PR #1369, opened 2020-02-29
chore: update react-intl among others

tdeekens:
#### Summary

Wanted a cut a release for the logout issue. Then noticed interally that the MC now works with the latest react-intl version (PR up). So I wanted to propose/test to update it here too.

While doing so I updated some deps I felt would be nice to include.


tdeekens:
Hm not sure why Zeit fails. Maybe it's still from the first commit and has not been retriggered.

emmenko:
I triggered a redeployment

tdeekens:
Thanks. Too bad no luck. Hm, does it pull in the latest published version maybe?

emmenko:
Have you tried running the build locally?

```
yarn clean
yarn build:website
```

tdeekens:
> Have you tried running the build locally?

Yup, just now. Fails too. I presume the website would need a babel-config too to not stumble over formatjs as the other packages. I has some webpack overwriting but not a babelrc. I propose to then keep the website at react-intl 3.4.0 and solve it separately by adding a babel config for instance. WDYT?

emmenko:
Hmm, for gatsby we should use gatsby-node.js to customize the webpack config.

https://github.com/commercetools/commercetools-docs-kit/blob/14581f1d3d715de2cdf0d3e231b7f49777c8aa13/packages/gatsby-theme-docs/gatsby-node.js#L176

emmenko:
I can take a look on this on Monday. Thanks for give it a try!

tdeekens:
> Hmm, for gatsby we should use gatsby-node.js to customize the webpack config.

Yeah I saw and read about that. However, I am not entirely sure how much it is a webpack issue itself. More a babel caused problem by babel-macro surfacing in webpack as an error.

I tried to narrow down the error some more now:

1. Checkout `master` and yarn isntall then build the website there: ✅
2. Checkout this branch (incl. the custom babel config) and build website: ✅
3. Yarn install on this branch (to update react-intl) and build website (excl. application components, previously built): 💣

```txt
 ERROR #98123  WEBPACK

Generating JavaScript bundles failed

Can't resolve '../../../../../website/public/static/d/2048913786.json' in '/Users/tdeekens/Development/merchant-center-application-kit/node_modules/@commercetools-docs/gatsby-theme-docs/src/components'
```

So I think somehow a dependency actually causes this not even the babel config itself.

application-templates/starter/package.json
```diff
@@ -27,7 +27,7 @@
     "react": "16.13.0",
     "react-apollo": "3.1.3",
     "react-dom": "16.13.0",
-    "react-intl": "3.12.0",
+    "react-intl": "3.4.0",
```
tdeekens: Just for easier experimentation as commit.

website/babel.config.js
```diff
@@ -25,7 +25,9 @@ const isMacrosName = packageName =>
   defaultMacrosRegex.test(packageName) &&
```
tdeekens: It's a discussion in gatsby if the allow babel.config (which are more project config) or rc files. They said they want to support babel.config.js so I think it's okay and better to use it.

tdeekens:
Will close this as it's so far not "open" perse. Feel free to reopen once we came up with a simple or magical solution :)

PR #1374, opened 2020-03-02
fix(jest-preset): to not throw on createFactory

tdeekens:
#### Summary

Library code uses createFactory which in turn will trigger throwing in our tests. Either we don't adopt the latest React version of we log the error instead of throwing it. When it is actually removed we truely need to make sure no library code uses it.

### DIFF
--- a/packages/jest-preset-mc-app/setup-tests.js
+++ b/packages/jest-preset-mc-app/setup-tests.js
@@ -15,9 +15,10 @@ const shouldSilenceWarnings = (...messages) =>
   ].some(msgRegex => messages.some(msg => msgRegex.test(msg)));
 
 const shouldNotThrowWarnings = (...messages) =>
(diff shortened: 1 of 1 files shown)


emmenko:
Do you know from which libraries we are getting the warnings?

tdeekens:
> Do you know from which libraries we are getting the warnings?

Nope. Planned to check after some meetings :)

tdeekens:
Grepping the node_modules I even can't find any occurance of createFactory.

emmenko:
Could be uglified...I thought the logs/warnings would show the package, or where the log is coming from.

tdeekens:
It comes from recompose to what I can see. More specifically withProps among other things.

tdeekens:
I will merge and release this to unblock the update. We should them remove recompose from the MC I think.

emmenko:
> We should them remove recompose from the MC I think.

Yes finally! 😄

PR #1379, opened 2020-03-05
fix(app-shell): regression to render menu divider

emmenko:
The menu divider seems to have disappeared in the recent appkit versions, probably after the TS migration #1273 

You can see below that in older versions of appkit (like from the custom app) the divider is still there, but not on production.

Kapture 2020-03-04 at 16 48 38

### DIFF
--- a/packages/application-shell/src/components/navbar/navbar.tsx
+++ b/packages/application-shell/src/components/navbar/navbar.tsx
@@ -383,13 +383,12 @@ const MenuLabel = (props: MenuLabelProps) => {
 
 type ApplicationMenuProps = {
   location: RouteComponentProps['location'];
-  menuType: string;
   menu: TNavbarMenu;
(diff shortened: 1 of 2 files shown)


packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -414,9 +413,7 @@ const ApplicationMenu = (props: ApplicationMenuProps) => {
 
   return (
     <>
-      {props.menuType === 'application' && props.menu.shouldRenderDivider && (
```
emmenko: The `props.menuType === 'application'` is the actual problem, as `application` does not exist anymore.

emmenko: The rest of the changes are to simplify a bit the implementation and remove the need to pass `menuType` as prop.

Issue #1377, opened 2020-03-04
Allow to reference environment variables from env.json


emmenko:
**Is your feature request related to a problem? Please describe.**
Sometimes it's useful to have different `env.json` configurations, for example when working with different environments.

**Describe the solution you'd like**
To make this easier for people to use one `env.json` file but with dynamic values, we can allow to reference environment variables in the JSON file. For example:

```json
{
  "mcApiUrl": "${env:MC_API_URL}"
}
```

> The idea is taken from `prisma.yaml`: https://www.prisma.io/docs/prisma-cli-and-configuration/prisma-yml-5cy7/#using-variables

PR #1380, opened 2020-03-06
feat(mc-html-template): add support env variable placeholders

tdeekens:
Ref: #1377

#### Summary

This pull request adds support for env variable placeholders to our `loadEnv` script.

#### Description

After doing some research I could not find an npm package doing what the issue references. There are quite bulky solutions to interpolate env variables but I felt sucking them as a dependency is a bit too much for our use case.

As a result we parse and substitute manually. Additionally I wrote some tests.

I would suggest document this somewhere if we want to but open a PR for that separately so we can try it out internally first.

Update: This now supports multiple placeholders: `${env:foo}.${env:bar}-${env:foo}`.

### DIFF
--- a/packages/mc-html-template/lib/load-env.js
+++ b/packages/mc-html-template/lib/load-env.js
@@ -1,6 +1,4 @@
-/* eslint-disable @typescript-eslint/explicit-function-return-type */
 const fs = require('fs');
-const replaceall = require('replaceall');
 
 // Keep a reference to the loaded config so that requiring the module
 // again will result in returning the cached value.
@@ -33,13 +31,11 @@ const substituteEnvVariablePlaceholder = (
     );
   }
 
-  const substitutedEnvConfig = replaceall(
-    matchedString,
-    valueOfEnv,
-    valueOfEnvConfig
+  const escapedMatchedString = matchedString.replace(/[${}:]/g, '\\$&');
+  return valueOfEnvConfig.replace(
+    new RegExp(`(${escapedMatchedString})+`, 'g'),
+    valueOfEnv
   );
-
-  return substitutedEnvConfig;
 };
 const getValueOfPlaceholder = valueWithPlaceholder =>
   valueWithPlaceholder
--- a/packages/mc-html-template/lib/load-env.js
+++ b/packages/mc-html-template/lib/load-env.js
@@ -5,12 +5,12 @@ const fs = require('fs');
 // again will result in returning the cached value.
 let loadedEnv;
 
-const START_OF_ENV_PLACEHODER = 'env:';
+const START_OF_ENV_PLACEHOLDER = 'env:';
 const isEnvVariablesPlaceholder = valueOfEnvConfig =>
-  valueOfEnvConfig && valueOfEnvConfig.startsWith(START_OF_ENV_PLACEHODER);
+  valueOfEnvConfig && valueOfEnvConfig.startsWith(START_OF_ENV_PLACEHOLDER);
 const substitutePlaceholder = valueOfEnvConfig => {
   const nameOfEnvVariable = valueOfEnvConfig.replace(
-    START_OF_ENV_PLACEHODER,
+    START_OF_ENV_PLACEHOLDER,
     ''
   );
 
--- a/packages/mc-html-template/lib/load-env.js
+++ b/packages/mc-html-template/lib/load-env.js
@@ -13,8 +13,7 @@ const variableSyntax = RegExp('\\${([ ~:a-zA-Z0-9._\'",\\-\\/\\(\\)]+?)}', 'g');
 const envRefSyntax = RegExp(/^env:/g);
 
 const hasVariablePlaceholder = valueOfEnvConfig =>
-  typeof valueOfEnvConfig === 'string' &&
-  variableSyntax.test(valueOfEnvConfig);
+  typeof valueOfEnvConfig === 'string' && variableSyntax.test(valueOfEnvConfig);
 const isEnvVariablePlaceholder = valueOfPlaceholder =>
   envRefSyntax.test(valueOfPlaceholder);
 const substituteEnvVariablePlaceholder = (
@@ -22,7 +21,7 @@ const substituteEnvVariablePlaceholder = (
   matchedString,
   valueOfEnvConfig
 ) => {
-  const [,requestedEnvVar] = valueOfPlaceholder.split(':');
+  const [, requestedEnvVar] = valueOfPlaceholder.split(':');
   const valueOfEnv = process.env[requestedEnvVar];
 
   if (!valueOfEnv) {
--- a/packages/mc-html-template/lib/load-env.js
+++ b/packages/mc-html-template/lib/load-env.js
@@ -9,13 +9,19 @@ let loadedEnv;
  *  Allows env variable placeholders e.g. `${env:MC_API_URL}`.
  *  Other placeholder types might be supported in the future.
  */
-const variableSyntax = RegExp('\\${([ ~:a-zA-Z0-9._\'",\\-\\/\\(\\)]+?)}', 'g');
-const envRefSyntax = RegExp(/^env:/g);
+const variableSyntax = /\${([ ~:a-zA-Z0-9._\'",\-\/\(\)]+?)}/g;
+const envRefSyntax = /^env:/g;
 
 const hasVariablePlaceholder = valueOfEnvConfig =>
-  typeof valueOfEnvConfig === 'string' && variableSyntax.test(valueOfEnvConfig);
+  typeof valueOfEnvConfig === 'string' &&
+  // Using `{regex}.test()` might cause false positives if called multiple
+  // times on a global regular expression:
+  // See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test
+  //    As with exec() (or in combination with it), test() called multiple times
+  //    on the same global regular expression instance will advance past the previous match.
+  Boolean(valueOfEnvConfig.match(variableSyntax));
 const isEnvVariablePlaceholder = valueOfPlaceholder =>
-  envRefSyntax.test(valueOfPlaceholder);
+  Boolean(valueOfPlaceholder.match(envRefSyntax));
 const substituteEnvVariablePlaceholder = (
   valueOfPlaceholder,
   matchedString,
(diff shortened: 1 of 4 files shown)


packages/mc-html-template/lib/load-env.js
```diff
@@ -1,16 +1,53 @@
...
 // Keep a reference to the loaded config so that requiring the module
 // again will result in returning the cached value.
 let loadedEnv;
 
-module.exports = configPath => {
-  if (loadedEnv) return loadedEnv;
+const START_OF_ENV_PLACEHODER = 'env:';
+const isEnvVariablesPlaceholder = valueOfEnvConfig =>
+  valueOfEnvConfig.startsWith(START_OF_ENV_PLACEHODER);
+const substitutePlaceholder = valueOfEnvConfig => {
```
tdeekens: This throws if a value you want to replac doesn't exist.

packages/mc-html-template/lib/load-env.js
```diff
@@ -1,16 +1,53 @@
...
+
+  if (!valueOfEnvVariable) {
+    throw new Error(
+      `Missing '${nameOfEnvVariable}' specified in config as 'env:${nameOfEnvVariable}'.`
+    );
+  }
+
+  return valueOfEnvVariable;
+};
+const substituteEnvVariablePlaceholders = config => {
```
tdeekens: We go entries and fromEntries. Essentially a mapValues but without the dependency.

tdeekens:
@YahiaElTai the idea is we can have an `env.json` like

```json
{
   "mcApiUrl": "env:MC_API_URL"
}
```

then a start script like `MC_API_URL='foo.com' yarn start`.

packages/mc-html-template/lib/load-env.js
```diff
@@ -1,16 +1,53 @@
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
 const fs = require('fs');
 
 // Keep a reference to the loaded config so that requiring the module
 // again will result in returning the cached value.
 let loadedEnv;
 
-module.exports = configPath => {
-  if (loadedEnv) return loadedEnv;
+const START_OF_ENV_PLACEHODER = 'env:';
```
emmenko:
```suggestion
const START_OF_ENV_PLACEHOLDER = 'env:';
```

emmenko: Any reason not to use the syntax `${env:MY_ENV}`?

packages/mc-html-template/lib/load-env.js
```diff
@@ -1,16 +1,53 @@
...
+const isEnvVariablesPlaceholder = valueOfEnvConfig =>
+  valueOfEnvConfig && valueOfEnvConfig.startsWith(START_OF_ENV_PLACEHODER);
```
emmenko: This will break if the value does not only contain a placeholder.

For example:

```json
{
  "mcApiUrl": "https://mc-api.${env:CTP_REGION}.${env:CTP_CLOUD_REGION}.commercetools.com"
}
```

packages/mc-html-template/lib/load-env.js
```diff
@@ -1,16 +1,53 @@
...
+};
+
+module.exports = (configPath, { disableCache } = { disableCache: false }) => {
+  if (loadedEnv && !disableCache) return loadedEnv;
 
   // Attempt to read the JSON file provided by each application
   const rawConfig = fs.readFileSync(configPath, { encoding: 'utf8' });
   const config = JSON.parse(rawConfig);
 
+  const substitutedConfig = substituteEnvVariablePlaceholders(config);
```
emmenko: Nit: maybe "compile" instead of "substitute"? We also use that name for "compiling" the `index.html`

packages/mc-html-template/lib/load-env.js
tdeekens: Not really. I timeboxed 30 minutes for this that's why kind of :)

packages/mc-html-template/lib/load-env.js
tdeekens: Compiling to me is something different than string replacing. When going around the internet on this most people call it substitution.

packages/mc-html-template/lib/load-env.js
tdeekens: Yeah, could be useful. I didn't see the case yet apart from theoretically being nice.

1. We could achieve this by saying that the string is a template literal. Then we could provide the process.env as values. Then we however loose the `env:` prefix. 
2. We could use a RegExp to extract matches etc pp

Do we actually see supporting other things than `env:` in the future? Like inputs?

emmenko: > Do we actually see supporting other things than env: in the future? Like inputs?

Not sure, but I would stick to the syntax `${env:}` since it's also used by other tools (like Prisma).

A regex should do the trick yes in replacing the values.

packages/mc-html-template/lib/load-env.js
```diff
@@ -1,16 +1,53 @@
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
```
emmenko: Do we need this? 🤔

packages/mc-html-template/lib/load-env.js
emmenko: Usually when you have template strings or something you compile the template. See e.g. ejs, handlebars, etc.

Anyway, not a big deal. I'm also fine with this.

packages/mc-html-template/lib/load-env.js
tdeekens: Yeah. Otherwise my editor was bonkers and warns things I don't have it warn me about :)

packages/mc-html-template/lib/load-env.js
tdeekens: I am working on it. Please call me on my mobile when you don't hear from me for 2-3 hours. I might have jumped off the balcony running away from RegExps.

packages/mc-html-template/lib/load-env.js
```diff
@@ -1,38 +1,78 @@
...
-  const nameOfEnvVariable = valueOfEnvConfig.replace(
-    START_OF_ENV_PLACEHOLDER,
-    ''
-  );
+/**
+ * NOTE:
+ *  Allows env variable placeholders e.g. `${env:MC_API_URL}`.
+ *  Other placeholder types might be supported in the future.
+ */
+const variableSyntax = RegExp('\\${([ ~:a-zA-Z0-9._\'",\\-\\/\\(\\)]+?)}', 'g');
```
tdeekens: These RegExps come from Prisma. It's actually not a standard or anyting. They've also just hardcoded stuff and this into their cli-engine.

I took the RegExps (allowing us to maybe to more in the future) while adding the matching code as much as possible.

packages/mc-html-template/lib/load-env.spec.js
```diff
@@ -0,0 +1,107 @@
...
+      fs.readFileSync.mockReturnValue(fsMocks.withPlaceholders);
+
+      expect(loadEnv(undefined, { disableCache: true })).toEqual(
+        expect.objectContaining(minimalEnvConfig)
+      );
+    });
+    it('should return the parsed config and all values with placeholders', () => {
+      process.env.API_URL = 'test-api-url';
+
+      fs.readFileSync.mockReturnValue(fsMocks.withPlaceholders);
```
emmenko: Isn't this the same as the test above? 🤔

tdeekens: Not really. I wanted to split testing that both 1. the variables without placeholders and 2. the onces with placeholders are defined. Helped me when building it to spot errors.

emmenko: Sure but those 2 tests are basically the same. I would have expected to just enhance the assertion, like:

```js
expect(loadEnv(undefined, { disableCache: true })).toEqual(
  expect.objectContaining({
    ...minimalEnvConfig,
    apiUrl: process.env.API_URL,
  })
);
```

Maybe it's just me that finds this a bit confusing 🙃

packages/mc-html-template/lib/load-env.js
```diff
@@ -1,16 +1,93 @@
...
+  const requestedEnvVar = valueOfPlaceholder.split(':')[1];
+  const valueOfEnv = process.env[requestedEnvVar];
+
+  if (!valueOfEnv) {
+    throw new Error(
+      `Missing '${requestedEnvVar}' specified in config as 'env:${requestedEnvVar}'.`
+    );
+  }
+
+  const substitutedEnvConfig = replaceall(
```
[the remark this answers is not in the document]
tdeekens: Yeah, it could work without. However the library is 10 lines and helps with the auto-escaping too in the string. That's why I felt it's useful. I you want to remove it feel free to add a commit.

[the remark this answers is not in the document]
emmenko: Yeah but I would avoid depending on a library for something simple like this.

https://github.com/commercetools/merchant-center-application-kit/pull/1380/commits/ff70e05165159cf9a7ed3ef30e726cc100bcf5d5

packages/mc-html-template/lib/load-env.js
```diff
@@ -33,13 +31,11 @@ const substituteEnvVariablePlaceholder = (
     );
   }
 
-  const substitutedEnvConfig = replaceall(
-    matchedString,
-    valueOfEnv,
-    valueOfEnvConfig
+  const escapedMatchedString = matchedString.replace(/[${}:]/g, '\\$&');
```
tdeekens: Haha. This is what I liked having the lib for. All cool :)

packages/mc-html-template/lib/load-env.spec.js
emmenko: This is what I meant: https://github.com/commercetools/merchant-center-application-kit/pull/1380/commits/ccf98f7d76003e13a4a9c8040075c7ea5a3057b5

If you had something else in mind please forgive me and feel free to revert.

packages/mc-html-template/lib/load-env.js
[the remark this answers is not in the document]
tdeekens: 😎 all good.

packages/mc-html-template/lib/load-env.js
emmenko: Just because you hate regular expressions 😜

packages/mc-html-template/lib/load-env.spec.js
tdeekens: Yeah I understood before I had that at first too. I felt it's two cases to test: have the ones with placeholders and those without. It was nice to see one spec failing when working on placeholders. Anway. I don't mind :)

packages/mc-html-template/lib/load-env.js
tdeekens: I propose we rewrite the MC in Regular Expressions.

PR #1389, opened 2020-03-11
refactor(test-utils): drop unused gtmTracking option

### LINKED ISSUE #1388: Passing gtmTracking render option doesn't work
**Describe the bug**

Passing `gtmTracking` render option to `renderApp` doesn’t work. 

**To Reproduce**

In the following test code, the mocked track will never be passed to `MyComponent`.

```js
  const track = jest.fn();
  renderApp(<MyComponent />, {
    gtmTracking: {
      track,
    },
  });
  // some test code
  expect(track).toHaveBeenCalled();
```

The test will fail with `Number of calls: 0`.

**Expected behavior**
The test should pass.

**Additional context**
https://commercetools.slack.com/archives/GCSJ9T12P/p1574956056011000

emmenko:
Closes #1388 

Just to confirm that the test-utils work properly.


ahmehri:
Well, when I first saw that you're expecting that `track` was called inside `wait` I said this is it, I have to wait for the expectation to pass, but then tried it and it didn't work.

So I copied exactly the same test code that you introduced and use it in our app and the test failed. You can find it [here](https://github.com/commercetools/merchant-center-frontend/blob/am-app-kit-issue-1388/packages-application/application-products/test.spec.js).

My assumption is that for some reason it only works inside app-kit but not outside of it, maybe because of some conflict. Maybe it's worth testing it inside another custom app to be sure. In the end it's not useful if it only works inside app-kit.

emmenko:
@ahmehri I added another test in the playground app and it works as well. Can you take another look?

emmenko:
Interesting, on CI it does not work 🤔

image

I'll dig into this issue a bit more.

emmenko:
FYI: I investigated this a bit and I think there is a fundamental issue with the way we export the context.

The app-shell package exports a "pre-configured" React context (`GtmContext`), which is used by the applications when loading the context.
On the other hand, the test-utils import the `GtmContext` from the app-shell package **src** folder, which results in the `GtmContext` code to be copied/included in the the test-utils bundle.

image

Here's the issue: the context exported by the app-shell **is not the same** as the context included in the test-utils.
Basically the context used in the test-utils is **never used** by the application code.

This is actually a tricky issue and I can't think of any good options other than removing the gtm context mock from the test-utils. The apps under tests must include the `GtmContext` explicitly and pass the mocks to it.

```js
const renderApp = (options = {}) => {
  const gtmMock = { track: jest.fn(), getHierarchy: jest.fn() };
  const rendered = renderAppWithRedux(
    <GtmContext.Provider value={gtmMock}>
      <MyApp />
    </GtmContext.Provider>,
    options
  );
  return { ...rendered, gtmMock };
};
```

emmenko:
@ahmehri @tdeekens any preferences on how to proceed here?

tdeekens:
I am fine with suggesting the `<GtmContext.Provider value={gtmMock}>` trick. Then we should however remove the `gtmTracking`-property I suppose. I don't see an other solution so far. I guess it would be down to rollup and some configuration/resolving changes there?

ahmehri:
Yes, I'm fine with the `<GtmContext.Provider value={gtmMock}>` trick as well, but as @tdeekens mentioned, we should remove the `gtmTracking` render option to avoid confusion, or better document how to simulate tracking.

emmenko:
Alright, I dropped the option.

I also looked where we could document this but realised that the `trackingGtm` option in the `env.json` wasn't even documented, meaning that 99.9999% is just us internally using it.
Therefore, it does not make sense to document it in the docs website.

I'm also wondering if we should consider this a breaking change or not.

tdeekens:
> I'm also wondering if we should consider this a breaking change or not.

This can be subjective but I would say it's not. We support what we document or what's an publically exposed API. Given it was not documented and the API was broken I think it's okay to "sneak" it in as a chore.

Issue #1390, opened 2020-03-12
Apollo's caching overrides values because of the way we extract id from cache


islam3zzat:
**Describe the bug**
According to the way we currently extract id from the object wi prefix with the type only when the `__typename` is 'Reference'
https://github.com/commercetools/merchant-center-application-kit/blob/e8296717b0828e29179c7c139bb772730df5356b/packages/application-shell/src/configure-apollo.ts#L115

But I faced an issue where the cached values has been overridden when the response of a resource that queried `cartDiscount`  including the `supplyChannelRef`

```graphql
query FetchCartDiscountQuery($id: String!) {
  cartDiscount(id: $id) {
    __typename
    value {
      ... on GiftLineItemValue {
        supplyChannelRef {
          id
          typeId
          __typename
        }
       
      }
    }
  }
}
```

the `supplyChannelRef` is of type `__type` = `ChannelReferenceIdentifier` which has a different key and doesn't have name/nameAllLocales

and the response had overridden the result of another query within the view  of type `__type` = `Channel` 

```graphql
 channels {
    results {
      __typename
      id
      key
      name(locale: "en")
    }
  }
```

![2020-03-11 4 23 34 PM](https://user-images.githubusercontent.com/7680511/76433527-baf36d80-63b4-11ea-8d82-6b074e2fb2fe.gif)

I tried changing the `dataIdFromObject` to enable prefixing the id with the `__typename` for `ChannelReferenceIdentifier` also and the issue is gone.
 

**To Reproduce**

1. Go to a view that has both `cartDiscountDetails` of type gift line item (which has a supply channel reference of type `ChannelReferenceIdentifier ` ) query and `channels` query
2. do any mutation on the `cartDiscountDetails`
3. since we don't prepend the `__typename` to the `id` of the cache when it's `ChannelReferenceIdentifier` (we currently only prepend the __typename for `Reference`), the id of the `supplyChannelRef` will override the `channel` of the same id.

**Expected behavior**
Both resources should be available on the cache since they are different resources with different types.

**Screenshots**

**Additional context**

emmenko:
Thanks for the detailed explanation.

I wish there is a better way to configure the data caching map, as we might encounter this issue in the future as well.

Maybe we should expose a way for custom apps to hook into the mapping function to provide any custom logic they need to.

islam3zzat:
> I wish there is a better way to configure the data caching map, as we might encounter this issue in the future as well.

[...]

PR #1394, opened 2020-03-16
fix(apollo-caching): simplify id from object logic

islam3zzat:
#### Summary

Fixes: https://github.com/commercetools/merchant-center-application-kit/issues/1390

### DIFF
--- a/packages/application-shell/src/configure-apollo.ts
+++ b/packages/application-shell/src/configure-apollo.ts
@@ -86,6 +86,15 @@ const fragmentMatcher = new IntrospectionFragmentMatcher({
   introspectionQueryResultData,
 });
 
+const typeNamesWithoutIdAsIdentifier = [
+  'Project',
+  'BaseMenu',
+  'NavbarMenu',
+  'Store',
+];
+
+const referenceTypes = ['Reference', 'ChannelReferenceIdentifier'];
+
 export const createApolloClient = () =>
   new ApolloClient({
     link,
@@ -93,9 +102,25 @@ export const createApolloClient = () =>
       fragmentMatcher,
       dataIdFromObject: result => {
         const customResult = result as CustomIdGetterObj;
-        // We default to key, if id is not available
-        return `${customResult.__typename}:${customResult.id ||
-          customResult.key}`;
+        if (
+          !customResult.id &&
+          customResult.key &&
+          customResult.__typename &&
+          typeNamesWithoutIdAsIdentifier.includes(customResult.__typename)
+        )
+          return `${customResult.__typename}:${customResult.key}`;
+        // Generally all id's are unique across the platform, so we don't need to
+        // include the type name in the cache key.
+        // However, a reference has the shape { typeId, id } where the id is the
+        // id of the referenced entity. If we didn't prefix ids of References
+        // they would clash with the referenced resource.
+        if (
+          customResult.__typename &&
+          referenceTypes.includes(customResult.__typename)
+        )
+          return `${customResult.__typename}:${customResult.id}`;
+
+        return customResult.id;
       },
     }),
   });
--- a/packages/application-shell/src/configure-apollo.ts
+++ b/packages/application-shell/src/configure-apollo.ts
@@ -86,13 +86,6 @@ const fragmentMatcher = new IntrospectionFragmentMatcher({
   introspectionQueryResultData,
 });
 
-const typeNamesWithoutIdAsIdentifier = [
-  'Project',
-  'BaseMenu',
-  'NavbarMenu',
-  'Store',
-];
-
 export const createApolloClient = () =>
   new ApolloClient({
     link,
--- a/packages/application-shell/src/configure-apollo.ts
+++ b/packages/application-shell/src/configure-apollo.ts
@@ -100,22 +100,9 @@ export const createApolloClient = () =>
       fragmentMatcher,
       dataIdFromObject: result => {
         const customResult = result as CustomIdGetterObj;
-        if (
-          !customResult.id &&
-          customResult.key &&
-          customResult.__typename &&
-          typeNamesWithoutIdAsIdentifier.includes(customResult.__typename)
-        )
-          return `${customResult.__typename}:${customResult.key}`;
-        // Generally all id's are unique accross the platform, so we don't need to
-        // include the type name in the cache key.
-        // However, a reference has the shape { typeId, id } where the id is the
-        // id of the referenced entity. If we didn't prefix ids of References
-        // they would clash with the referenced resource.
-        if (customResult.__typename === 'Reference')
-          return `${customResult.__typename}:${customResult.id}`;
-
-        return customResult.id;
+        // We default to key, if id is not available
+        return `${customResult.__typename}:${customResult.id ||
+          customResult.key}`;
       },
     }),
   });


packages/application-shell/src/configure-apollo.ts
```diff
@@ -100,22 +100,9 @@ export const createApolloClient = () =>
...
-          customResult.__typename &&
-          typeNamesWithoutIdAsIdentifier.includes(customResult.__typename)
-        )
-          return `${customResult.__typename}:${customResult.key}`;
-        // Generally all id's are unique accross the platform, so we don't need to
-        // include the type name in the cache key.
-        // However, a reference has the shape { typeId, id } where the id is the
-        // id of the referenced entity. If we didn't prefix ids of References
-        // they would clash with the referenced resource.
-        if (customResult.__typename === 'Reference')
```
tdeekens: So "all" we do is to always prefix with typename right? Do we anyhow now why we didn't do this in the first place? /cc @emmenko maybe?

packages/application-shell/src/configure-apollo.ts
```diff
@@ -100,22 +100,9 @@ export const createApolloClient = () =>
       fragmentMatcher,
       dataIdFromObject: result => {
         const customResult = result as CustomIdGetterObj;
-        if (
-          !customResult.id &&
-          customResult.key &&
-          customResult.__typename &&
-          typeNamesWithoutIdAsIdentifier.includes(customResult.__typename)
-        )
-          return `${customResult.__typename}:${customResult.key}`;
```
islam3zzat: the new condition will return the same value, unless :
`!customResult.id ` ✔︎
`customResult.key` ✔︎
`customResult.__typename ` ✘

In this case the old condition will not fire and we'll move to the else and  return
either `${customResult.__typename}:undefined`; for Reference type or
just `undefined` if it wasn't  a Reference

tdeekens: Ah, you're saying with the simplifcation we actually cover edge cases better? Sorry Monday morning slowness :)

packages/application-shell/src/configure-apollo.ts
emmenko: Can't remember exactly. Some old PRs I found:
- https://github.com/commercetools/merchant-center-application-kit/pull/310
- https://github.com/commercetools/merchant-center-application-kit/commit/072852f434649c92f50c0801b0fa265ebf537ef7
- https://github.com/commercetools/merchant-center-application-kit/commit/20731f6a10e293fdfde23c26d610f57d97422df6

packages/application-shell/src/configure-apollo.ts
islam3zzat: > you're saying with the simplifcation we actually cover edge cases better? 


I think so. I was skeptical at first, but the only possible downside that I could find was if the `__typename` is not available.

I'm not sure if this is possible, but in this case, the proposed implementation will not do better or worse.
Maybe if this is possible, we can also default to e.g `---` so the cache id, in this case, would be

`---: resource-id`, instead of `undefined: resource-id`

tdeekens: I looked at the previous commits and PRs and I think this change should be fine. My question would be just the precedence. So always id if possible otherwise key or key if present.

islam3zzat: The most important thing for the cache id is that it must be unique. Given that, I think the `id` should have higher precedence.

emmenko:
@islam3zzat What's the status here?

islam3zzat:
@emmenko was waiting for review.
Also `build_lint_and_test` failed in the build step.

emmenko:
> Also `build_lint_and_test` failed in the build step.

That's why I asked 😄

emmenko:
Other than that, if we're confident that this change does not brake the apps, I'm fine with it.

tdeekens:
@islam3zzat my suggestion would be we fix the build/tests here. I release the quick access (and other fixes) before. Then we merge this to get a canary build. This we can then use on a branch deployment for quick testing. Sounds good?

emmenko:
Have you tried first maybe to "patch" the code in node_modules and start the apps locally?

tdeekens:
> Have you tried first maybe to "patch" the code in node_modules and start the apps locally?

Me or @islam3zzat? I think @islam3zzat could try better to test the spots he needs fixing. Before the build succeeds we also need to fix the unused variable :)

islam3zzat:
> Have you tried first maybe to "patch" the code in node_modules and start the apps locally?

Will test it again and remove the unused vars.

tdeekens:
There might already be some tests here which will help testing this a bit.

islam3zzat:
The proposed solution didn't work.
After trying it in the MC, I found that some resources don't have either `id` or `key` or `__typename`.

The old implementation will default at the end to **undefined**, but the original proposed solution will make the cache key only **"undefined:undefined"**

tried also 
```js
      dataIdFromObject: result => {
        const customResult = result as CustomIdGetterObj;
        const prefix = customResult.__typename;
        const id = customResult.id || customResult.key;
        if (prefix && id) return `${prefix}:${id}`;
        return null;
      },
```
But also didn't work.


I moved back to the original implementation, but only prefixed references.

emmenko:
> After trying it in the MC, I found that some resources don't have either id or key or __typename.

Do you have an example which resources are those?

emmenko:
Btw, the default implementation from apollo also returns null

https://github.com/apollographql/apollo-client/blob/9df809bd36c892653871a30df4d559222ee16458/src/cache/inmemory/policies.ts#L177-L184

islam3zzat:
some examples

no id & key
-
 ``` js 
 {
  "results": [
    {
      "id": "7f90e904-4fe0-4a81-b3d3-2cc7e97e8b55",
      "name": null,
      "key": "1",
      "__typename": "Channel"
    },
  ],
  "__typename": "ChannelQueryResult"
}
```

```js
{
  "isActive": false,
  "__typename": "ProjectExpiry"
}
```
```js
{
  "isActive": false,
  "__typename": "ProjectSuspension"
}
```

Sorry, I was wrong `__typename` is always there.

emmenko:
Well, `ProjectExpiry` and `ProjectSuspension` do not need to have ids or keys, because they are "fields" of a resource. Should we try it like this maybe?

```
dataIdFromObject: result => {
  if (typeof result.__typename === "string") {
    if (typeof result.id !== "undefined") return `${result.__typename}:${result.id}`;
    if (typeof result.key !== "undefined") return `${result.__typename}:${result.key}`;
  }
  return null;
}
```

islam3zzat:
I tried it also and didn't work. 
I don't understand why! but it's similar to this one ( https://github.com/commercetools/merchant-center-application-kit/pull/1394#issuecomment-600778348 )

tdeekens:
> I tried it also and didn't work.

You mean "your" current state of @emmenko proposal?

islam3zzat:
The current state is the original implementation + extra condition.

By "I tried it also and didn't work.", I meant, I wrote what Nicola proposed and build the app shell and checked the MC and it didn't work (the keys of the channels response are overwritten by the reference channel in the cart discount)

emmenko:
> the keys of the channels response are overwritten by the reference channel in the cart discount

Can you clarify how the change in this PR fixes that issue, compared to the "simpler" implementation? Just for me to understand, as I don't know what's the difference between channel and a channel reference.
Thanks

islam3zzat:
> Can you clarify how the change in this PR fixes that issue, compared to the "simpler" implementation?

The issue happens because we in one view query both `Channels` and `ChannelReferenceIdentifiers`

`Channels` have key, localized name and id, but `ChannelReferenceIdentifiers`  as a reference have only name and id.

In the current implementation, where we only prefix the referenced types (Reference, and ChannelReferenceIdentifiers) those two resources get a different entry in the cache.

In the "simpler" implementation should fix the same issue, but causes a problem somewhere and unfortunately, I don't get any errors just the error page without any error/warning in the console.

PR #1395, opened 2020-03-16
PCM-1600: routes to product varitns from quick access

vidurar:
In which routing to Product Variant Details from the quick access is redirecting to the incorrect page in both instances. Fixed by correcting the string for the route

### DIFF
--- a/packages/application-shell/src/components/quick-access/quick-access.tsx
+++ b/packages/application-shell/src/components/quick-access/quick-access.tsx
@@ -286,7 +286,13 @@ const QuickAccess = (props: Props) => {
           keywords: variantKey ? [variantKey] : undefined,
           action: {
             type: actionTypes.go,
-            to: `/${applicationContext.project.key}/products/${productId}/${variantId}`,
+            to: oneLineTrim`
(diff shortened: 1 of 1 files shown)


packages/application-shell/src/components/quick-access/quick-access.tsx
```diff
@@ -286,7 +286,13 @@ const QuickAccess = (props: Props) => {
           keywords: variantKey ? [variantKey] : undefined,
           action: {
             type: actionTypes.go,
-            to: `/${applicationContext.project.key}/products/${productId}/${variantId}`,
+            to: oneLineTrim`
```
tdeekens: CleanShot 2020-03-16 at 10 59 23@2x

This should go to a specific variant? Isn't the path correct?

vidurar: there should be a `/variants/` between the product id an the variant id, was the only change i made

tdeekens: Got it. Sorry.

PR #1397, opened 2020-03-16
refactor: to use waitFor and waitForElementToBeRemoved

tdeekens:
#### Summary

The path to react-testing-library/{react,dom} on the latest versions is quite a bit of effort.

In this pull request I tackle:

1. wait to be waitFor
2. waitForElementToBeRemoved where possible
3. Clean up some tests which are just... confusing

In the process I renamed imports from wait to waitFor. This is trick to get the dependency update PR in the end to be lean by just renaming the import again.

### DIFF
--- a/packages/application-shell/src/components/quick-access/quick-access.spec.js
+++ b/packages/application-shell/src/components/quick-access/quick-access.spec.js
@@ -3,7 +3,12 @@ import {
   GRAPHQL_TARGETS,
   MC_API_PROXY_TARGETS,
 } from '@commercetools-frontend/constants';
-import { renderAppWithRedux, fireEvent, wait } from '../../test-utils';
+import {
+  renderAppWithRedux,
+  fireEvent,
+  wait as waitFor,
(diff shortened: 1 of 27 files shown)


packages/application-shell/src/components/quick-access/quick-access.spec.js
```diff
@@ -3,7 +3,12 @@ import {
   GRAPHQL_TARGETS,
   MC_API_PROXY_TARGETS,
 } from '@commercetools-frontend/constants';
-import { renderAppWithRedux, fireEvent, wait } from '../../test-utils';
+import {
+  renderAppWithRedux,
+  fireEvent,
+  wait as waitFor,
+  waitForElementToBeRemoved,
```
emmenko: Btw, this is not used apparently.

tdeekens: Yeah I noticed ESLint being more loose on app-kit. Maybe something to change.

PR #1398, opened 2020-03-16
refactor: to use location module

tdeekens:
#### Summary

This pull request removes direct usage of `window.location`. 

#### Description

`window.location` is nor writable in all browsers. As a result jsdom v16 will also not allow it to be writable. This means all tests wanting to assign a `jest.fn` to e.g. `window.location.replace` will rightfully not pass anymore. This is correct behaviour from a library perspective miminicing the and APIs of browsers.

Options out of this issue:

1. To not use `window.location` directly but through a tiny wrapper which we can mock under test
2. To use `Object.defineProperty` to side-pass the restrictions

```js
Object.defineProperty(window, 'location', {
   writable: true,
   value: { replace: jest.fn() }
});
```

Both options is said to work. However the Jest team seems to recommend to actually do 1. I agree with it. Working around browser like behaviour in an environment which you want to work like a browser feels defeating the purpose of it all.

As a result I:

1. Added a `location` module under `utils` of the application shell
2. Use this module everywhere we for now use `replace`
3. `jest.mock` the module under test

I find this solution quite okay and it gets us one step closer towards being able to adopt jsdom v16.

### DIFF
--- a/packages/application-shell/src/components/application-shell/application-shell.spec.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.spec.js
@@ -184,8 +184,8 @@ describe('when user has no default project', () => {
     });
     await waitFor(() => {
       expect(location.replace).toHaveBeenCalledWith('/account/projects/new');
+      expect(rendered.queryByText('OK')).not.toBeInTheDocument();
     });
-    expect(rendered.queryByText('OK')).not.toBeInTheDocument();
   });
 });
 describe('when loading user fails with an unknown graphql error', () => {
@@ -226,8 +226,8 @@ describe('when loading user fails with an unauthorized graphql error', () => {
       expect(location.replace).toHaveBeenCalledWith(
         expect.stringContaining(`/logout?${queryParams}`)
       );
+      expect(rendered.queryByText('OK')).not.toBeInTheDocument();
     });
-    expect(rendered.queryByText('OK')).not.toBeInTheDocument();
   });
 });
 describe('when loading user fails with a "was not found." graphql error message', () => {
@@ -251,8 +251,8 @@ describe('when loading user fails with a "was not found." graphql error message'
       expect(location.replace).toHaveBeenCalledWith(
         expect.stringContaining(`/logout?${queryParams}`)
       );
+      expect(rendered.queryByText('OK')).not.toBeInTheDocument();
     });
-    expect(rendered.queryByText('OK')).not.toBeInTheDocument();
   });
 });
 describe('when project is not found', () => {
@@ -521,8 +521,8 @@ describe('when user is not authenticated', () => {
       expect(location.replace).toHaveBeenCalledWith(
         `${window.location.origin}/login?${queryParams}`
       );
+      expect(rendered.queryByText('OK')).not.toBeInTheDocument();
     });
-    expect(rendered.queryByText('OK')).not.toBeInTheDocument();
   });
 });
 describe('when selecting project locale "de"', () => {
--- a/packages/application-shell/src/components/application-shell/application-shell.spec.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.spec.js
@@ -34,10 +34,12 @@ import {
 import { createApolloClient } from '../../configure-apollo';
 import { STORAGE_KEYS } from '../../constants';
 import { PROJECT_EXTENSIONS } from '../../feature-toggles';
+import { location } from '../../utils/location';
 import ApplicationShellProvider from '../application-shell-provider';
 import ApplicationShell from './application-shell';
 
 jest.mock('@commercetools-frontend/sentry');
+jest.mock('../../utils/location');
 
 const createTestProps = props => ({
   environment: {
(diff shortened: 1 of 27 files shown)


packages/application-shell/src/utils/location/location.ts
```diff
@@ -0,0 +1,7 @@
+const replace = (url: string): void => window.location.replace(url);
```
tdeekens: The idea is we wrap more APIs we use from the window's location when we need it.

packages/application-shell/src/components/application-shell/application-shell.tsx
```diff
@@ -602,14 +602,14 @@ const ApplicationShell = <AdditionalEnvironmentProperties extends {}>(
 
           return (
             <Route
-              render={({ location }) => (
+              render={({ location: routeLocation }) => (
```
tdeekens: Otherwise there is a naming clash.

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -515,14 +514,15 @@ describe('when user is not authenticated', () => {
     const rendered = renderApp(null, { route: '/foo' });
     const queryParams = encode({
       reason: LOGOUT_REASONS.UNAUTHORIZED,
-      redirectTo: `${window.location.origin}/foo`,
```
tdeekens: @emmenko this is the only test I am not too sure about. I am not sure why it should be `/foo`. Cause the window.location.origin isn't on `/foo`. It's the window.history and react-router which is.

It's also another type without an actual origin.

CleanShot 2020-03-16 at 19 27 31@2x

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -182,11 +183,9 @@ describe('when user has no default project', () => {
       },
     });
     await waitFor(() => {
-      expect(window.location.replace).toHaveBeenCalledWith(
-        '/account/projects/new'
-      );
-      expect(rendered.queryByText('OK')).not.toBeInTheDocument();
+      expect(location.replace).toHaveBeenCalledWith('/account/projects/new');
     });
+    expect(rendered.queryByText('OK')).not.toBeInTheDocument();
```
emmenko: Any reason for moving this outside of the "wait"?

packages/application-shell/src/components/application-shell/application-shell.spec.js
emmenko: I'm not sure I get what you mean. The `/foo` route is what it needs to redirect to, so it needs to be in the `redirectTo` param.

packages/application-shell/src/components/application-shell/application-shell.spec.js
tdeekens: Just to isolate things a bit. We waitFor something and then expect something outside. I am fine either way just felt it wasnt needed having both in it. For one also as I wasnt sure what would happen if B passes and A doesnt. Then A passes and B too. Feels a bit odd.

packages/application-shell/src/components/application-shell/application-shell.spec.js
tdeekens: Ah true my bad. Will check again tomorrow.

packages/application-shell/src/utils/location/location.ts
```diff
@@ -0,0 +1,31 @@
...
+ * */
+const replace = (url: string): void => window.location.replace(url);
+const reload = window.location.reload;
+
+const location = {
+  replace,
+  reload,
+};
+
+export default location;
```
emmenko: Nit: wondering if the naming might be a bit confusing, as we have:
- `window.location`
- `history.location`
- (this) `location`

Not sure though...just a thought.

packages/application-shell/src/components/application-shell/application-shell.spec.js
emmenko: The only concern I have is that the two expectations are not related to each other, meaning that given the first passes, the second one might not be "ready" yet.
That's why they are both in the "wait" block. Maybe we can have 2 waiting blocks?

packages/application-shell/src/utils/location/location.ts
tdeekens: I also changed more tests below now to be as they were. Sorry for mixing things.

I also struggled with the naming. Starting to work with MC we get `import windowLocation from` which looks odd. But I don't know a better name right now. Like `import {replace} from '...'` is even strager cause then it's `replace('http/foo.com')` which is even more confusing. I frankly didn't manage to come up with anything else. Wrapping the whole window is a pain cause when you mock it you mock getters and setters which have side-effects.

packages/application-shell/src/components/application-shell/application-shell.spec.js
tdeekens: Oki doks. I agree. I changed it again :)

PR #1401, opened 2020-03-17
feat(jest-preset): add preset for jsdom sixteen

tdeekens:
#### Summary

This pull request suggests adding a jest preset for jsdom v16. 

#### Description

There are multiple ways of doing this:

1. Change the existing presets to jsdom v16 (typescript and default)
   - Definately breaking. Decided against it as jest by default also still uses the old version. Should be a separate step I think.
2. Adding another preset
   - Not breaking as people can just consume it as opt-in

Some things to note:

1. I would like to change the TypeScript template to v16 but that would also be breaking
2. I feel like we should have a better way of merging presets e.g.

```js
const { defaultPreset, typeScriptPreset } = require('@commercetools-frontend/jest-reset-mc-app');

module.exports = defaultPreset.extend({
   testEnvironment: 'jest-environment-jsdom-sixteen'
});

// or

module.exports = defaultPreset.use();

// or

module.exports = defaultPreset.use({
   testEnvironment: 'jest-environment-jsdom-sixteen'
});
```

### DIFF
--- a/packages/jest-preset-mc-app/jest-preset-for-jsdom-sixteen.js
+++ b/packages/jest-preset-mc-app/jest-preset-for-jsdom-sixteen.js
@@ -1,0 +1,6 @@
+const defaultPreset = require('./jest-preset');
+
+module.exports = {
+  ...defaultPreset,
+  testEnvironment: 'jest-environment-jsdom-sixteen',
+};
--- a/packages/jest-preset-mc-app/package.json
+++ b/packages/jest-preset-mc-app/package.json
@@ -41,6 +41,7 @@
     "jest-transform-graphql": "2.1.0",
     "jest-watch-master": "1.0.0",
     "jest-watch-typeahead": "0.4.2",
+    "jest-environment-jsdom-sixteen": "1.0.3",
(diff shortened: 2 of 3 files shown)


emmenko:
Do you want to suggest the changes about exporting the presets in a separate PR?

tdeekens:
> Do you want to suggest the changes about exporting the presets in a separate PR?

Not sure I follow. I thought about just adding them for now. Any other way, as the existing, of exporting them would be breaking. So I didn't plan to do that yet. Only if we do it though another export like `@ct-frontend/jest-preset-mc-app/presets`? Is that what you mean?

emmenko:
I don't know, I haven't give it too much thoughts, I was just referring to the things you wrote in the PR description if you wanted to follow that up or something

tdeekens:
> I don't know, I haven't give it too much thoughts, I was just referring to the things you wrote in the PR description if you wanted to follow that up or something

I am fine with this now. We just add a preset and merge them internally. Next I would maybe use the JSDOM v16 in the TS preset.

PR #1402, opened 2020-03-17
chore: to use jsdom sixteen under test

tdeekens:
#### Summary

This switches to using jsdom v16 under test.

### DIFF
--- a/packages/application-shell/src/components/application-shell/application-shell.spec.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.spec.js
@@ -83,7 +83,15 @@ const renderApp = (ui, options = {}) => {
   const rendered = render(
     <ApplicationShell {...props} render={() => ui || <p>{'OK'}</p>} />
   );
-  return { ...rendered, history: testHistory };
+  const findByLeftNavigation = () => rendered.findByTestId('left-navigation');
+  const queryByLeftNavigation = () => rendered.queryByTestId('left-navigation');
+
+  return {
+    ...rendered,
+    findByLeftNavigation,
+    queryByLeftNavigation,
+    history: testHistory,
+  };
 };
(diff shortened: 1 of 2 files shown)


tdeekens:
Ah the menu is still loading under test now I think.

```html
<div
   class="loading-dot-container"
/>
```

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -692,12 +692,13 @@ describe('navbar menu links interactions', () => {
...
-    // Get the nav container, to narrow down the search area
-    const leftNavContainer = await rendered.findByTestId('left-navigation');
-    const navbarRendered = within(leftNavContainer);
+    // Wait for the nav container
+    await rendered.findByTestId('left-navigation');
 
     // Check the relationships between the menu items of a group
-    const menuTitle = await navbarRendered.findByText(mainMenuLabel.value);
+    const menuTitle = await within(
+      await rendered.findByTestId('left-navigation')
```
tdeekens: This is odd. The tests fail cause the return of the `within`-call is bound to a subdom (only under JSDOM v16) which contains a menu which is still loading. According to the classname and content. While the debug value or rendered clearly outlines the opposite.

I then tried to use waitForElementToBeRemoved to await the loading to be removed. However, rtl rightfully tells me that it was never present in the first place. Hence, it being an invalid await.

I then inlined awaiting the left navigation which into a within in place. This then fixes the tests. It's the only way I got things to pass.

tdeekens: P.S.: I am fine with this. However, I would also like the render to be a bit more helpful like `rendered.findByLeftNavigation`.

PR #1403, opened 2020-03-18
fix(quick-access): link to pim search list

vidurar:
Link to the old products list should now point to the PIM Search 

The copy for `Open Product List` should be changed to `Open Products List`


packages/i18n/data/core.json
```diff
@@ -105,8 +105,8 @@
   "QuickAccess.openOrdersList": "Open Orders List",
   "QuickAccess.openPimSearch": "Open PIM Search",
   "QuickAccess.openProductDiscountsList": "Open Product Discounts List",
-  "QuickAccess.openProductList": "Open Product List",
   "QuickAccess.openProducts": "Open Products",
+  "QuickAccess.openProductsList": "Open Products List",
```
emmenko: Is someone taking care of updating this in TX as well?

emmenko:
@vidurar please avoid references to internal JIRA issues here. Thanks.

tdeekens:
@ahmehri how is this chapter issue currently being prioritised in your team?

packages/i18n/data/core.json
vidurar: These keys will be passed on to UX for updating in transifex after this ticket is merged

ahmehri:
I'm aware of the chapter issue and I already communicated to the team about it. It's not currently being prioritized.

ahmehri:
Maybe it makes sense first to discuss it in the chapter and then coordinate the prioritization in the team.

packages/i18n/data/core.json
ahmehri: Wouldn't it be better to edit the existing translation instead of creating a new one?

tdeekens:
> Maybe it makes sense first to discuss it in the chapter and then coordinate the prioritization in the team.

We did already and also in the round with the PO present after.

ahmehri:
I had a discussion with the team and yes the PO is aware of it, we're just heading towards fixing the current bugs and will follow up with the chapter issue soon. We're planning a meeting to discuss it tomorrow.

packages/i18n/data/core.json
vidurar: usually would edit the existing translation, but I think a new key would be more appropriate in this case because the key is the same verbiage as the text

PR #1406, opened 2020-03-20
fix(jest-preset-mc-app): add mutation-observer-shim

tdeekens:
#### Summary

This pull request adds the mutation-observer-shim to our jest preset.

#### Description

The preset was removed within dom-testing-library v7. The library itself recommends adding it or using jsdom v16 (which contains it). We offer a preset for jsdom v16 but should patch out regular preset to include the shim so it works with dom-testing-library v7

Note, that dom-testing-library used this shim itself. Adding it on older dom-testing-library versions (overwriting it) on the window should not be an issue.

CleanShot 2020-03-20 at 09 30 39@2x


tdeekens:
FYI: I tested this internally in our apps and it works.

PR #1411, opened 2020-03-23
fix: proper invocation of window.location.replace

pa3:
`utils/location/replace` method invokes `window.location.replace` with a wrong `this` value, which lead to an error when navigation between apps:

Screenshot 2020-03-23 at 14 34 07

This PR fixes that.

### DIFF
--- a/packages/application-shell/src/utils/location/location.ts
+++ b/packages/application-shell/src/utils/location/location.ts
@@ -21,7 +21,8 @@
  *    mocking system of Jest and no other trickery.
  * */
 const replace = (url: string): void => window.location.replace(url);
-const reload = (...args): void => window.location.reload(...args);
+const reload = (forceReload = false): void =>
+  window.location.reload(forceReload);
 
 const location = {
   replace,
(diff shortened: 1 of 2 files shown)


tdeekens:
Typechecker is after you :) 

CleanShot 2020-03-23 at 14 52 13@2x

packages/application-shell/src/utils/location/location.ts
```diff
@@ -21,7 +21,7 @@
  *    mocking system of Jest and no other trickery.
  * */
 const replace = (url: string): void => window.location.replace(url);
-const reload = window.location.reload;
+const reload = (): void => window.location.reload();
```
emmenko:
```suggestion
const reload = (...args): void => window.location.reload(...args);
```

tdeekens: Accordin to MDN this method doesn't accept args. Do you have a reference stating that it does. Maybe it's just my search skillz.

pa3:
@tdeekens removed changes to `yarn.lock`.

packages/application-shell/src/utils/location/location.ts
emmenko: image


https://developer.mozilla.org/en-US/docs/Web/API/Window/location

tdeekens: Ok. Then maybe let's add it to not introduce another change with the fix. Would at least be my way of dealing with it :)

PR #1416, opened 2020-03-26
Properly report unhandled rejections and ErrorEvents to Sentry

pa3:
### Context

At the moment we observe following uninformative reports in our Sentry instance:

Screenshot 2020-03-26 at 10 57 10

That's happening because `PromiseRejectionEvent` and `ErrorEvent` captured by AppShell's global error handlers are passed to `Sentry.captureException` unmodified, whereas the latter expects instance of an `Error`.

 ### Fix

This PR implements wrapping error events into `Error`s inside `reportErrorToSentry` before sending them to `Sentry`. Also default `Sentry`s reporting of unhandled rejections and global errors is disabled, to prevent double-reporting.

### Other fix options

The better option, probably, is to rely on `Sentry`'s default reporting of unhandled rejections and remove our custom global handlers, but that will prevent us from showing error_id to users in MC in case of unhandled rejections. So that's probably a subject for a follow-up discussion.

### DIFF
--- a/packages/sentry/src/sentry.spec.ts
+++ b/packages/sentry/src/sentry.spec.ts
@@ -13,8 +13,8 @@ const DUMMY_DSN = 'https://acacaeaccacacacabcaacdacdacadaca@sentry.io/000001';
 // we can drop this polyfill: https://github.com/jsdom/jsdom/issues/2401
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
-type Reason = any;
-type Payload = any;
+type Reason = unknown;
+type Payload = unknown;
 /* eslint-enable @typescript-eslint/no-explicit-any */
 
 class PromiseRejectionEvent extends Event {
--- a/packages/sentry/src/sentry.spec.ts
+++ b/packages/sentry/src/sentry.spec.ts
@@ -7,6 +7,30 @@ const { testkit, sentryTransport } = sentryTestkit();
 
 const DUMMY_DSN = 'https://acacaeaccacacacabcaacdacdacadaca@sentry.io/000001';
 
+// At the moment, jsdom doesn't support PromiseRejectionEvent type. That leads to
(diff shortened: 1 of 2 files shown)


packages/sentry/src/sentry.spec.ts
```diff
@@ -7,6 +7,30 @@ const { testkit, sentryTransport } = sentryTestkit();
...
 const DUMMY_DSN = 'https://acacaeaccacacacabcaacdacdacadaca@sentry.io/000001';
 
+// At the moment, jsdom doesn't support PromiseRejectionEvent type. That leads to
+// ReferenceError in tests. So, we add this ad-hoc polyfill just for tests for now.
+// Once the following issue is fixed (which is unlikely to happen any soon) in jsdom
+// we can drop this polyfill: https://github.com/jsdom/jsdom/issues/2401
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+type Reason = any;
+type Payload = any;
```
emmenko: Better to use `unknown` than `any`. Then eslint won't complain.

```suggestion
type Reason = unknown;
type Payload = unknown;
```

packages/sentry/src/sentry.ts
```diff
@@ -13,15 +30,22 @@ export const boot = () => {
...
+      // we don't need default Sentry's global handlers, because we  add default ones
+      integrations: [
+        new Sentry.Integrations.GlobalHandlers({
+          onunhandledrejection: false,
+          onerror: false,
+        }),
+      ],
     });
-    Sentry.configureScope((scope) => {
+    Sentry.configureScope(scope => {
```
emmenko: Forgot to apply prettier?

pa3: That's weird, but looks like some outdated prettier/config was applied. Removing `node_modules` and reinstalling all modules helped. 🤷‍♂

PR #1418, opened 2020-03-27
docs(website): mention limitations of forward-to proxy endpoint

emmenko:
image

### DIFF
--- a/website/src/content/main-concepts/proxy-to-external-api.mdx
+++ b/website/src/content/main-concepts/proxy-to-external-api.mdx
@@ -4,7 +4,7 @@ title: Proxy to External API
 
 <Warning>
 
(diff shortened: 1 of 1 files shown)


website/src/content/main-concepts/proxy-to-external-api.mdx
```diff
@@ -2,6 +2,12 @@
 title: Proxy to External API
 ---
 
+<Warning>
+
+This feature only allows for an external server to check that requests come from authenticated users. It does not provide a way for the server to perform user access validation and more fine grained permission checks. If you are interested in this functionalities, let us know and open a [support issue](https://github.com/commercetools/merchant-center-application-kit/issues/new/choose).
```
FreddieK: I would change this to info rather than warning. People should be aware of it, but we don't want to scare them away from using Custom Applications. Similarly, it might make more sense at the bottom of the 'Validating authenticated requests' section than at the top of the page?

As for the text, a small change; "If you are interested in this functionalities" should be either "If you are interested in **these** functionalities" or "If you are interested in this **functionality**".

PR #1422, opened 2020-04-01
docs(website): mention that dev login with SSO is currently not supported

emmenko:


packages/mc-dev-authentication/views/login.pug
```diff
@@ -39,5 +39,13 @@ html
           aria-label="Sign in"
         ) ➡ Sign in 🚀
 
+      div(class="info")
```
emmenko, line 42: image

PR #1423, opened 2020-04-01
feat(jest-preset): allow to configure jest preset

### LINKED ISSUE #1421: Configure Jest setup-tests

Currently, I cannot configure Jest `setup-tests` in my app.

I would like to be able to configure app-kit Jest `setup-tests` from my app.

**Additional context**

I want to silence some console errors and I don't want to duplicate the logic in my app but instead reuse `setup-tests` `silenceConsoleErrors`. Something like the following.

```js
import { configureTestSetups, testSetup } from '...'

configureTestSetup({
   silenceConsoleWarnings: [
	//,
    //
   ]
})

export testSetup
```

emmenko:
Fixes #1421

### DIFF
--- a/packages/jest-preset-mc-app/load-config.js
+++ b/packages/jest-preset-mc-app/load-config.js
@@ -43,6 +43,6 @@ module.exports = loadConfig;
 
 function createListMergerWithDefaults(key) {
   return (config) => ({
-    [key]: defaultConfig[key].concat(config[config] || []),
+    [key]: defaultConfig[key].concat(config[key] || []),
   });
 }
--- a/packages/jest-preset-mc-app/load-config.js
+++ b/packages/jest-preset-mc-app/load-config.js
@@ -19,18 +19,24 @@ const mergeSilenceConsoleWarnings = createListMergerWithDefaults(
 );
 const mergeNotThrowWarnings = createListMergerWithDefaults('notThrowWarnings');
 
+let cachedConfig;
 const loadConfig = () => {
-  const configFile = explorer.search();
-  if (!configFile || configFile.isEmpty) {
-    return defaultConfig;
+  if (cachedConfig) {
+    return cachedConfig;
   }
-  const customConfig = configFile.config;
-  return {
-    ...defaultConfig,
-    ...customConfig,
-    ...mergeSilenceConsoleWarnings(customConfig),
-    ...mergeNotThrowWarnings(customConfig),
-  };
+  const configFile = explorer.search();
+  if (configFile && configFile.config) {
+    const customConfig = configFile.config;
+    cachedConfig = {
+      ...defaultConfig,
+      ...customConfig,
+      ...mergeSilenceConsoleWarnings(customConfig),
+      ...mergeNotThrowWarnings(customConfig),
+    };
+  } else {
+    cachedConfig = defaultConfig;
+  }
+  return cachedConfig;
 };
 
 module.exports = loadConfig;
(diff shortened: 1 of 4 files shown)


packages/jest-preset-mc-app/load-config.js
```diff
@@ -0,0 +1,42 @@
...
+  const configFile = explorer.load(pathToConfig);
+  if (jestConfig.isEmpty) {
+    return defaultConfig;
+  }
+  const customConfig = configFile.config;
+  return {
+    ...defaultConfig,
+    ...customConfig,
+    ...mergeSilenceConsoleWarnings(customConfig),
+    ...mergeNotThrowWarnings(customConfig),
```
emmenko, line 32: These functions basically merge the default values with custom ones.
I'm not 100% sure about this though, if we should just let the user "override" the field.

packages/jest-preset-mc-app/load-config.js
```diff
@@ -0,0 +1,42 @@
...
+  };
+};
+
+module.exports = loadConfig;
+
+function createListMergerWithDefaults(key) {
```
tdeekens, line 38: Wow, we use hoisting. Nice! Still stumbled when reading assuming that this must be a library provided thing. Maybe just put in L5 (nit).

packages/jest-preset-mc-app/load-config.js
```diff
@@ -0,0 +1,48 @@
...
+let cachedConfig;
+const loadConfig = () => {
+  if (cachedConfig) {
+    return cachedConfig;
+  }
+  const configFile = explorer.search();
+  if (configFile && configFile.config) {
+    const customConfig = configFile.config;
+    cachedConfig = {
+      ...defaultConfig,
```
tdeekens, line 31: Should we merge these deep? Is that somewhat what I'd expect as a consumer defining my config overwrites?

packages/jest-preset-mc-app/load-config.js
```diff
@@ -0,0 +1,48 @@
+const { cosmiconfigSync } = require('cosmiconfig');
+
+const moduleName = 'jest-mc-app';
```
tdeekens, line 3: Should also be `pkg.name` almost of don't you want it to be `jest-preset-mc-app.yaml` eg?

packages/jest-preset-mc-app/load-config.js
emmenko: That was my [question](https://github.com/commercetools/merchant-center-application-kit/pull/1423#discussion_r401711259) as well 🙂

packages/jest-preset-mc-app/load-config.js
emmenko: Yeah, I felt that the "preset" part was superfluous.

tdeekens: Yeah maybe. However its also nice to have pkg name and cosmiconfig file names the same. Even if superfluous. Could bring clarity. Just a thought 😎

packages/jest-preset-mc-app/load-config.js
tdeekens: I think it is a flat structure after all. Maybe it is just fine to keep it simple for now. Deep merging would be a non-breaking change. Removing it would be. So maybe it is fine as is.

packages/jest-preset-mc-app/load-config.js
```diff
@@ -0,0 +1,48 @@
...
+    cachedConfig = defaultConfig;
+  }
+  return cachedConfig;
+};
+
+module.exports = loadConfig;
+
+function createListMergerWithDefaults(key) {
+  return (config) => ({
+    [key]: defaultConfig[key].concat(config[config] || []),
```
ahmehri, line 46: Should it be this instead?

```suggestion
    [key]: defaultConfig[key].concat(config[key] || []),
```

emmenko: Yes thanks 😅

PR #1424, opened 2020-04-01
fix(l18n): react to only be peer dependency

tdeekens:
#### Summary

I am debugging a very odd bug with redux and react-redux and invalid hook calls. This PR suggests to mainly remove react as a direct dependency from a package. As the error can somewhat only come from duplicated react versions being bundled from my view.


application-templates/starter/package.json
```diff
@@ -24,9 +24,9 @@
     "@commercetools-uikit/spacings": "10.15.1",
     "@commercetools-uikit/text": "10.15.1",
     "prop-types": "15.7.2",
-    "react": "16.13.0",
+    "react": "16.13.1",
```
tdeekens, line 27: Just while here...

tdeekens, line 27: Just while here...

packages/l10n/package.json
```diff
@@ -44,8 +44,7 @@
     "@commercetools-frontend/sentry": "16.5.3",
     "@types/prop-types": "^15.7.3",
     "moment": "^2.24.0",
-    "prop-types": "15.7.2",
-    "react": "16.13.0"
```
tdeekens, line 48: Moving this down into devDeps.

emmenko: Ah, it should probably also be a peer dep. Good catch!

tdeekens: React is a devDep mostly. So I presume we add a peerDep requirement here too.

tdeekens: 2e2d854

PR #1429, opened 2020-04-06
chore(ci): continuously deploy playground app to ZEIT

emmenko:
This PR aims to automate the deployments of the playground app to ZEIT, so that when we test the custom app in the MC it actually works and the deployment is not outdated.


.github/workflows/main.yml
```diff
@@ -226,3 +226,9 @@ jobs:
   #         artifact_id: ${{ steps.get_artifact_id.outputs.artifact_id }}
   #       env:
   #         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Deploying Playground application to ZEIT Now
+        #if: github.ref == 'refs/heads/master'
```
emmenko, line 231: Just so that I can test this on this PR, then it will only run on master builds.

scripts/deploy_playground.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -e
+
+: "${ZEIT_NOW_PLAYGROUND?Required env variable ZEIT_NOW_PLAYGROUND}"
+
+echo "Deploying Playground for production GCP-EU"
+yarn --cwd playground deploy:eu --scope=commercetools-playground --token="$ZEIT_NOW_PLAYGROUND"
```
emmenko, line 8: See https://zeit.co/docs/now-cli#options/global-options/token

playground/now-deployments/state-machines-gcp-eu/package.json
```diff
@@ -0,0 +1,6 @@
+{
```
emmenko, line 1: I moved the now deployment related stuff into those subfolders. This is kind of necessary because of Now project linking: https://zeit.co/docs/now-cli#commands/now/project-linking

I think it's actually better like this, as it decouples the Now setup from the app a bit more.

PR #1433, opened 2020-04-09
fix(app-shell): apply cached menu state synchronously as a layout effect, to avoid flickering

adnasa:
#### Summary

- update the DOM immediately when `STORAGE_KEYS.IS_FORCED_MENU_OPEN=true`

#### Description

We have `use-navbar-state-manager` to determine the state of the menu as the user loads the application. An issue within the state manager, is that we wait for the reducer to execute actions to determine the state.

This defers the experience of having the menu open _immediately_ when the user loads the application

For example
flickering menu

**Proposed solution**

Update the dom immediately, before the reducer executes action.
no-flicket

trade off, [this code](https://github.com/commercetools/merchant-center-application-kit/blob/adnasa-use-navbar-state-manage-flick-thousand-cuts/packages/application-shell/src/components/navbar/use-navbar-state-manager.ts#L178-L188) will no longer execute. I however decided to keep it, since I'm unsure about what other cases could arise that would need this execution.

---

I created this as a Draft PR, because I'm open to other approaches of updating the DOM immediate and not defer it to the reducer execution.
I added @FreddieK to UAT this.

### DIFF
--- a/packages/application-shell/src/components/navbar/navbar.tsx
+++ b/packages/application-shell/src/components/navbar/navbar.tsx
@@ -50,6 +50,7 @@ import { GtmContext } from '../gtm-booter';
 import LoadingPlaceholder from '../loading-placeholder';
 import styles from './navbar.mod.css';
 import messages from './messages';
+import useInitialForcedMenuState from './use-initial-forced-menu-state';
 import useNavbarStateManager from './use-navbar-state-manager';
 import nonNullable from './non-nullable';
 
@@ -644,6 +645,7 @@ export default NavBar;
 
 export const LoadingNavBar = () => {
   const ref = React.useRef<HTMLDivElement>(null);
+  useInitialForcedMenuState();
   return (
     <NavBarLayout ref={ref}>
       <MenuGroup id="main" level={1}>
--- a/packages/application-shell/src/components/navbar/navbar.tsx
+++ b/packages/application-shell/src/components/navbar/navbar.tsx
@@ -50,7 +50,7 @@ import { GtmContext } from '../gtm-booter';
 import LoadingPlaceholder from '../loading-placeholder';
 import styles from './navbar.mod.css';
 import messages from './messages';
-import useInitialForcedMenuState from './use-initial-forced-menu-state';
+import useLoadingMenuLayoutEffect from './use-loading-menu-layout-effect';
 import useNavbarStateManager from './use-navbar-state-manager';
 import nonNullable from './non-nullable';
 
@@ -645,7 +645,7 @@ export default NavBar;
 
 export const LoadingNavBar = () => {
   const ref = React.useRef<HTMLDivElement>(null);
-  useInitialForcedMenuState();
+  useLoadingMenuLayoutEffect();
   return (
     <NavBarLayout ref={ref}>
       <MenuGroup id="main" level={1}>
--- a/packages/application-shell/src/components/navbar/use-initial-forced-menu-state.ts
+++ b/packages/application-shell/src/components/navbar/use-initial-forced-menu-state.ts
@@ -1,0 +1,18 @@
+import React from 'react';
+import isNil from 'lodash/isNil';
+import { STORAGE_KEYS } from '../../constants';
+
+const useInitialForcedMenuState = () => {
+  const cachedIsForcedMenuOpen = window.localStorage.getItem(
+    STORAGE_KEYS.IS_FORCED_MENU_OPEN
+  );
+  const isForcedMenuOpen = isNil(cachedIsForcedMenuOpen)
+    ? null
+    : (JSON.parse(cachedIsForcedMenuOpen) as boolean);
+  React.useLayoutEffect(() => {
+    if (isForcedMenuOpen) document.body.classList.add('body__menu-open');
+    if (!isForcedMenuOpen) document.body.classList.remove('body__menu-open');
+  }, [isForcedMenuOpen]);
+};
+
+export default useInitialForcedMenuState;
--- a/packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
+++ b/packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
@@ -34,43 +34,12 @@ type Action =
   | { type: 'setIsMenuOpenAndMakeExpanderVisible'; payload: boolean }
   | { type: 'reset' };
 
-type CachedIsMenuOpenState = string | null;
-
-const initialState = {
+const getInitialState = (isForcedMenuOpen: boolean | null): State => ({
   isExpanderVisible: true,
-  isMenuOpen: false,
-};
-
-const cachedIsForcedMenuOpen = window.localStorage.getItem(
-  STORAGE_KEYS.IS_FORCED_MENU_OPEN
-);
-
-const getShouldInitiallyOpen = (
-  cachedIsMenuOpen: CachedIsMenuOpenState
-): boolean => {
-  if (!isNil(cachedIsMenuOpen)) {
-    return JSON.parse(cachedIsMenuOpen);
-  }
-  return false;
-};
-
-const shouldBeInitiallyOpen: boolean = getShouldInitiallyOpen(
-  cachedIsForcedMenuOpen
-);
-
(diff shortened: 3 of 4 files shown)


adnasa:
I did a `--no-verify` on the commit, since I get this error

```
❯ y typecheck
yarn run v1.22.4
$ tsc --noEmit
packages/sentry/src/sentry.spec.ts:39:7 - error TS2322: Type 'new (options: TransportOptions) => Transport' is not assignable to type 'TransportClass<Transport>'.
  Construct signature return types 'Transport' and 'Transport' are incompatible.
    The types of 'sendEvent' are incompatible between these types.
      Type '(event: import("/Users/adnasa/workspace/repos/merchant-center-application-kit/node_modules/@sentry/types/dist/event").Event) => PromiseLike<import("/Users/adnasa/workspace/repos/merchant-center-application-kit/node_modules/@sentry/types/dist/response").Response>' is not assignable to type '(event: import("/Users/adnasa/workspace/repos/merchant-center-application-kit/node_modules/@sentry/browser/node_modules/@sentry/types/dist/event").Event) => PromiseLike<import("/Users/adnasa/workspace/repos/merchant-center-application-kit/node_modules/@sentry/browser/node_modules/@sentry/types/dist/response").Respon...'.
        Types of parameters 'event' and 'event' are incompatible.
          Type 'import("/Users/adnasa/workspace/repos/merchant-center-application-kit/node_modules/@sentry/browser/node_modules/@sentry/types/dist/event").Event' is not assignable to type 'import("/Users/adnasa/workspace/repos/merchant-center-application-kit/node_modules/@sentry/types/dist/event").Event'.
            Types of property 'spans' are incompatible.
              Type 'import("/Users/adnasa/workspace/repos/merchant-center-application-kit/node_modules/@sentry/browser/node_modules/@sentry/types/dist/span").Span[] | undefined' is not assignable to type 'import("/Users/adnasa/workspace/repos/merchant-center-application-kit/node_modules/@sentry/types/dist/span").Span[] | undefined'.
                Type 'import("/Users/adnasa/workspace/repos/merchant-center-application-kit/node_modules/@sentry/browser/node_modules/@sentry/types/dist/span").Span[]' is not assignable to type 'import("/Users/adnasa/workspace/repos/merchant-center-application-kit/node_modules/@sentry/types/dist/span").Span[]'.
                  Type 'Span' is missing the following properties from type 'Span': child, isRootSpan

39       transport: sentryTransport,
         ~~~~~~~~~


Found 1 error.

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I'll have a look at that, but please review this approach in the mean time

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
```diff
@@ -34,10 +34,33 @@ type Action =
...
   | { type: 'reset' };
 
+const cachedIsForcedMenuOpen = window.localStorage.getItem(
+  STORAGE_KEYS.IS_FORCED_MENU_OPEN
+);
+
+const getShouldInitiallyOpen = (): boolean => {
+  if (!isNil(cachedIsForcedMenuOpen)) {
+    return (
+      typeof cachedIsForcedMenuOpen === 'string' &&
```
tdeekens, line 44: We could use the `JSON.parse` trick here to shorten this if we'd want.

CleanShot 2020-04-09 at 13 14 20@2x

adnasa: https://github.com/commercetools/merchant-center-application-kit/pull/1433/commits/ef314ecc6fb4388e349479cb3c425faac500d649

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
```diff
@@ -34,10 +34,30 @@ type Action =
...
+// GIVEN user has `STORAGE_KEYS.IS_FORCED_MENU_OPEN=true`
+// THEN update the DOM immediately and not defer that to the reducer
+// this is done to avoid the flickering experience of the menu once the user loads the application
+if (shouldBeInitiallyOpen) {
+  document.body.classList.add('body__menu-open');
+}
 const initialState = {
   isExpanderVisible: true,
-  isMenuOpen: false,
+  isMenuOpen: shouldBeInitiallyOpen,
```
emmenko, line 58: I wouldn't define it "statically" like this.

Can we maybe change this into a function `getInitialState` and pass the `isMenuOpen` to the function?

Then you do

```js
useReducer(reducer, getInitialState(cachedIsForcedMenuOpen))
```

```

Comment (adnasa) [reply packages/application-shell/src/components/navbar/use-navbar-state-manager.ts:58]: > Thanks for looking into this. I have an alternative proposal though...

I am fine with the proposal with the initial state not defined statically. With this, we can keep `cachedIsForcedMenuOpen` scoped inside the hook.

---

> ...to avoid using window and document when the module loads.

I've already tried this, since this was my initial approach

However, avoiding to use window and document when the module loads brings us back to the initial problem where updating the list to be expanded immediate is deferred to the reducer.

Comment (adnasa) [reply packages/application-shell/src/components/navbar/use-navbar-state-manager.ts:58]: basically, we need to keep this

```js
if (shouldBeInitiallyOpen) {
  document.body.classList.add('body__menu-open');
}
```

emmenko: I get that. What I'm suggesting is to pass the cached value to the `initialState` of the reducer. I haven't tried this but I assume it should work.

adnasa: okay, sure. We do that already.

https://github.com/commercetools/merchant-center-application-kit/blob/ef314ecc6fb4388e349479cb3c425faac500d649/packages/application-shell/src/components/navbar/use-navbar-state-manager.ts#L147:L147

but if we want to do this via a function, sure. we can do that too. I'll come back with an update later

adnasa: maybe something in this direction?
f7be21b

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -644,6 +645,7 @@ export default NavBar;
 
 export const LoadingNavBar = () => {
   const ref = React.useRef<HTMLDivElement>(null);
+  useInitialForcedMenuState();
```
emmenko, line 648: I decided to have a separate hook to use here, as otherwise the "big" state manager hook sets event listeners and other things that are not necessary for the "loading" navbar container.

packages/application-shell/src/components/navbar/use-initial-forced-menu-state.ts
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import isNil from 'lodash/isNil';
+import { STORAGE_KEYS } from '../../constants';
+
+const useInitialForcedMenuState = () => {
```
tdeekens, line 5: Nice! Nit: Maybe `useMenuLayoutEffect` to indicate what the hook does. It's not really returning state or anything which the name otherwise suggsts.

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
```diff
@@ -199,7 +199,7 @@ const useNavbarStateManager = (props: HookProps) => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []); // <-- run this only once!!
 
-  React.useEffect(() => {
+  React.useLayoutEffect(() => {
```
emmenko, line 202: The `useLayoutEffect` is just an additional improvement which ensures that the effect is applied to the DOM before the component is rendered.

https://kentcdodds.com/blog/useeffect-vs-uselayouteffect

adnasa: 🔥

adnasa:
I'll do a release at the end of the day

tdeekens:
> I'll do a release at the end of the day

Cool. You can also play with the canary first on a internal branch deployment.

PR #1452, opened 2020-04-21
refactor(express): support request handler for serverless functions

emmenko:
To better support the usage in serverless functions, I extracted the core logic into a `createSessionAuthVerifier`, which only requires `req, res` to be passed.

The middleware then is just a wrapper for express.

### DIFF
--- a/packages-backend/express/README.md
+++ b/packages-backend/express/README.md
@@ -16,7 +16,7 @@ $ npm install --save @commercetools-backend/express
 
 ## Session middleware
 
-This middleware should be used to handle the authentication exchange between the server and the `/proxy/forward-to` endpoint of the Merchant Center API Gateway.
+This Express.js middleware should be used to handle the authentication exchange between the server and the `/proxy/forward-to` endpoint of the Merchant Center API Gateway.
 
 > You can read more about the "Proxy to External API" concepts [here](https://docs.commercetools.com/custom-applications/main-concepts/proxy-to-external-api).
 
@@ -26,7 +26,12 @@ const {
   CLOUD_IDENTIFIERS,
 } = require('@commercetools-backend/express');
 
-app.use(createSessionMiddleware({ mcApiUrl: CLOUD_IDENTIFIERS.GCP_EU }));
+app.use(
+  createSessionMiddleware({
+    audience: 'https://my-api-server.com',
+    issuer: CLOUD_IDENTIFIERS.GCP_EU,
+  })
+);
 app.use((request, response, next) => {
   // `request.session` contains the useful information
 });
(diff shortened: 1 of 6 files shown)


packages-backend/express/README.md
```diff
@@ -26,20 +26,58 @@ const {
...
+);
 app.use((request, response, next) => {
   // `request.session` contains the useful information
 });
 ```
 
 ### Middleware options
 
-- `issuer` (_string_): either a cloud identifier or a valid URL to the Merchant Center API Gateway. The cloud identifier maps to the Merchant Center API URL of the related [cloud region](https://docs.commercetools.com/custom-applications/main-concepts/api-gateway#cloud-regions).
+- `audience` (_string_): The public-facing URL of your API server. The value should only contain the origin URL (protocol, hostname, port), the request path is inferred from the incoming request.
```
emmenko, line 42: It's better to explicitly provide the `audience`, as the `req.hostname` only exists in express.js and might not even work in a serverless environment.

packages-backend/express/src/auth.ts
```diff
@@ -0,0 +1,118 @@
...
+      !Array.isArray(cloudIdentifierHeader)
+        ? mapCloudIdentifierToUrl(cloudIdentifierHeader) ??
+          configuredDefaultIssuer
+        : configuredDefaultIssuer;
+
+    // @ts-ignore: the node HTTP request does not know about `originalUrl`
+    const requestUrlPath = request.originalUrl ?? request.url;
+    const audience = `${options.audience.replace(/\/?$/, '')}${requestUrlPath}`;
+
+    return new Promise((resolve, reject) => {
```
emmenko, line 84: This is the "changed" part, where now it returns a promise.

packages-backend/express/src/middlewares/session-middleware.ts
```diff
@@ -1,106 +1,20 @@
...
-
-  // Return the middleware
-  return (request: Request, response: Response, next: NextFunction) => {
-    const cloudIdentifierHeader = request.header('x-mc-api-cloud-identifier');
-    const issuer =
-      options.inferIssuer && cloudIdentifierHeader
-        ? mapCloudIdentifierToUrl(cloudIdentifierHeader) ??
-          configuredDefaultIssuer
-        : configuredDefaultIssuer;
+  const sessionAuthVerifier = createSessionAuthVerifier<Request, Response>(
```
tdeekens, line 7: Ah nice. Lower level building blocks :)

packages-backend/express/README.md
tdeekens: True, did we release the prior version on latest or as a beta?

emmenko: Canary 😄

PR #1453, opened 2020-04-21
refactor(express): support backwards compatibility for v1

emmenko:
In order to allow a non-breaking migration to v2, the middleware should handle the legacy hostnames when processing the request, based on a custom header sent by the MC API with the forwarded request.

### DIFF
--- a/packages-backend/express/src/auth.ts
+++ b/packages-backend/express/src/auth.ts
@@ -1,41 +1,14 @@
 import type { ServerResponse, IncomingMessage } from 'http';
 import type { TSessionMiddlewareOptions, TSession } from './types';
 
-import { CLOUD_IDENTIFIERS, MC_API_URLS } from './constants';
-import expressJwtMiddleware from 'express-jwt';
 import jwksRsa from 'jwks-rsa';
-
-const decodedTokenKey = 'decoded_token';
-
-const mapCloudIdentifierToUrl = (
-  issuer: TSessionMiddlewareOptions['issuer']
-): string | undefined => {
-  switch (issuer) {
-    case CLOUD_IDENTIFIERS.GCP_AU:
-      return MC_API_URLS.GCP_AU;
-    case CLOUD_IDENTIFIERS.GCP_EU: {
-      return MC_API_URLS.GCP_EU;
-    }
-    case CLOUD_IDENTIFIERS.GCP_US: {
-      return MC_API_URLS.GCP_US;
-    }
-    case CLOUD_IDENTIFIERS.AWS_FRA:
-      return MC_API_URLS.AWS_FRA;
-    case CLOUD_IDENTIFIERS.AWS_OHIO:
-      return MC_API_URLS.AWS_OHIO;
-    default:
-      return undefined;
-  }
-};
-const throwIfIssuerIsNotAValidUrl = (issuer: string) => {
-  try {
-    new URL(issuer);
-  } catch (error) {
-    throw new Error(
-      `Invalid issuer URL "${issuer}". Expected a valid URL to the Merchant Center API Gateway, or a cloud identifier to one of the available cloud regions. See https://docs.commercetools.com/custom-applications/main-concepts/api-gateway#hostnames.`
-    );
-  }
-};
+import expressJwtMiddleware from 'express-jwt';
+import {
+  CLOUD_IDENTIFIERS,
+  MC_API_URLS,
+  MC_API_PROXY_HEADERS,
+} from './constants';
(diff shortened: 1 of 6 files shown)


packages-backend/express/src/auth.ts
```diff
@@ -1,24 +1,52 @@
...
+
+type TDecodedJWT = {
+  sub: string;
+  iss: string;
+  [property: string]: string;
+};
 
 const decodedTokenKey = 'decoded_token';
+// Assign a session object to the request object.
+const writeSessionContext = <Request extends IncomingMessage>(
```
emmenko, line 21: Just moved up

packages-backend/express/src/auth.ts
```diff
@@ -36,47 +78,59 @@ const throwIfIssuerIsNotAValidUrl = (issuer: string) => {
...
+    // Get the `Accept-version` header, forwarded by the `/proxy/forward-to` endpoint.
+    // The version should be sent by the client making the request, to use the features of v2.
+    const proxyForwardVersion = getFirstOrThrow(
+      request.headers[MC_API_PROXY_HEADERS.FORWARD_TO_VERSION],
+      `Missing "X-MC-API-Forward-To-Version" header.`
+    );
+    if (proxyForwardVersion === 'v1') {
+      // Fall back to legacy issuer domains
+      issuer = mapToLegacyIssuer(cloudIdentifierHeader) ?? issuer;
+    }
```
emmenko, line 132: This is the part where we make the issuer backwards compatible.

packages-backend/express/src/utils.ts
```diff
@@ -0,0 +1,11 @@
+const getFirstOrThrow = (
+  value: string | string[] | undefined,
```
emmenko, line 2: Because the header can also be an array of strings.

packages-backend/express/src/auth.ts
```diff
@@ -36,47 +78,59 @@ const throwIfIssuerIsNotAValidUrl = (issuer: string) => {
     );
   }
 };
-
-type TDecodedJWT = {
-  sub: string;
-  iss: string;
-  [property: string]: string;
+// Validates required option values.
+const validateRequiredValues = (options: TSessionMiddlewareOptions) => {
```
tdeekens, line 82: Looked into Node.js `assert` but somehow it wouldn't help with this. Was just curious.

packages-backend/express/src/auth.ts
```diff
@@ -36,47 +78,59 @@ const throwIfIssuerIsNotAValidUrl = (issuer: string) => {
...
-
-  request.session = {
-    userId: decodedToken.sub,
-    projectKey: decodedToken[publicClaimForProjectKey],
-  };
-
-  delete request.decoded_token;
+// Attempt to parse the given issuer. If the value is a cloud identifier, it will
+// be mapped to one of the supported values. If not, we assume the value is a valid URL.
+const parseDefaultIssuer = (options: TSessionMiddlewareOptions) => {
```
tdeekens, line 92: Is this parsing or just mapping. Just trying to follow along with the code. `inferIssuerDefaultOrThrow`? It becomes odd :)

packages-backend/express/src/auth.ts
```diff
@@ -36,47 +78,59 @@ const throwIfIssuerIsNotAValidUrl = (issuer: string) => {
...
+  // Returns an async HTTP handler.
   return async (request: Request, response: Response) => {
-    const cloudIdentifierHeader = request.headers['x-mc-api-cloud-identifier'];
-    const issuer =
-      options.inferIssuer &&
-      cloudIdentifierHeader &&
-      !Array.isArray(cloudIdentifierHeader)
-        ? mapCloudIdentifierToUrl(cloudIdentifierHeader) ??
+    // Get the cloud identifier header, forwarded by the `/proxy/forward-to` endpoint.
+    const cloudIdentifierHeader = getFirstOrThrow(
```
tdeekens, line 112: Why do we need to assume it can be an array? I fail to follow 😎

packages-backend/express/src/auth.ts
emmenko: This function should return a value, either from the cloud identifier or the manually passed url.
I couldn’t come up with a better name

packages-backend/express/src/auth.ts
emmenko: Technically headers can be arrays, that’s how the vale is typed (in nodejs).

tdeekens: Ah. Got it.

PR #1457, opened 2020-04-22
feat: support forwardTo request config for apollo and sdk

emmenko:
This PR continues the improvements to work with the endpoint `/proxy/forward-to`.

The idea here is to allow users to keep using apollo graphql and our sdk action creators to make HTTP requests to their external API, without having to provide all the necessary configuration changes.

## Usage for Apollo

We can leverage the `context` option for Apollo queries to change a bit how the request is configured. The app-shell package now exposes an utility function to configure the Apollo context for the `/proxy/forward-to` usage.

```js
import { useQuery } from 'react-apollo';
import { createApolloContextForProxyForwardTo } from '@commercetools-frontend/application-shell';

const Fetcher = () => {
  const { loading, data, error } = useQuery(MyQuery, {
    context: createApolloContextForProxyForwardTo({
      uri: 'https://my-custom-app.com/graphql',
    }),
  });
}; 
```

## Usage for SDK

By default, all requests with the SDK are configured to be sent to the MC API.
To make it easier to make requests to the proxy endpoint using the SDK, there is a new action creator wrapper that comes with built-in configuration options.

The exported action creators have a new export `forwardTo`, which is an object containing wrappers around the normal action creators.

```js
actions.forwardTo.get({ uri: 'https://my-custom-app.com/graphql' });
actions.forwardTo.del(options);
actions.forwardTo.head(options);
actions.forwardTo.post(options);
```

The options for the action creators are the same as the normal **Usage with `uri`** action creators, except that the `uri` value needs to be the URL to the external API.

### DIFF
--- a/packages/application-shell/src/apollo-links/header-link.ts
+++ b/packages/application-shell/src/apollo-links/header-link.ts
@@ -1,3 +1,5 @@
+import type { TApolloContext } from '../utils/apollo-context';
+
 import ApolloClient from 'apollo-client';
 import { NormalizedCacheObject } from 'apollo-cache-inmemory';
 import { ApolloLink } from 'apollo-link';
@@ -12,8 +14,7 @@ import {
 
 type GraphQlTarget = typeof GRAPHQL_TARGETS[keyof typeof GRAPHQL_TARGETS];
 
-type ApolloContext = {
-  uri?: string;
+type ApolloContextWithInMemoryCache = TApolloContext & {
   cache: ApolloClient<NormalizedCacheObject>;
 };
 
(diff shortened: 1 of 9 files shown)


packages/application-shell/src/apollo-links/header-link.ts
```diff
@@ -29,20 +30,15 @@ const isKnownGraphQlTarget = (target: GraphQlTarget) =>
...
 const headerLink = new ApolloLink((operation, forward) => {
-  const { uri, cache } = operation.getContext() as ApolloContext;
-
-  // In case the `context` contains a `uri`, it means that we are not sending
-  // the request to the MC API, but to another server.
-  // For now, if that's the case, we skip the `target` validation and we do
-  // not send the custom headers.
-  if (uri) {
-    return forward(operation);
-  }
```
emmenko, line 40: This logic isn't really necessary. There is now a `skipGraphQlTargetCheck` option for that.

packages/sdk/README.md
```diff
@@ -100,15 +100,40 @@ Before using the `sdk-client` the `sdk` middleware combines `service` and `optio
...
+
+To make it easier to make requests to the proxy endpoint using the SDK, there is a new action creator wrapper that comes with built-in configuration options.
+
+The exported action creators have a new export `forwardTo`, which is an object containing wrappers around the normal action creators.
+
+```js
+actions.forwardTo.get(options);
+actions.forwardTo.del(options);
+actions.forwardTo.head(options);
+actions.forwardTo.post(options);
```
tdeekens, line 116: 💯

emmenko:
Does the approach/API make sense to you?

tdeekens:
Yes. SDK API is lovely. The creation of the Apollo context a hit "bulky" but we should recommend anyway that people have a `apollo-contexts.js` with a `forwardToContext` so the create call only happens once.

PR #1458, opened 2020-04-23
feat(playground): implement echo server view/api to demo /proxy/forward-to feature

emmenko:
Adds an echo server to the playground app (deployed on ZEIT). The serverless function uses the new "middleware" to authenticate the request.
The playground app then has a new page to demo the connection.

image

### DIFF
--- a/playground/src/components/echo-server/echo-server.js
+++ b/playground/src/components/echo-server/echo-server.js
@@ -8,8 +8,23 @@ import Text from '@commercetools-uikit/text';
 import PrimaryButton from '@commercetools-uikit/primary-button';
 import { CodeBlock } from '@commercetools-docs/ui-kit';
 
+const initialState = {
+  isLoading: false,
+  result: null,
+};
+const reducer = (state, action) => {
+  switch (action.type) {
+    case 'loading':
+      return { isLoading: true, result: null };
+    case 'complete':
+      return { isLoading: false, result: action.payload };
+    default:
+      return state;
+  }
+};
+
 const EchoServer = () => {
-  const [result, setResult] = React.useState();
+  const [state, dispatchState] = React.useReducer(reducer, initialState);
   const dispatch = useAsyncDispatch();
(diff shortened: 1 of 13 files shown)


emmenko:
I think both test and demo.

playground/src/components/echo-server/echo-server.js
```diff
@@ -8,8 +8,23 @@ import Text from '@commercetools-uikit/text';
...
+  result: null,
+};
+const reducer = (state, action) => {
+  switch (action.type) {
+    case 'loading':
+      return { isLoading: true, result: null };
+    case 'complete':
+      return { isLoading: false, result: action.payload };
+    default:
+      return state;
```
tdeekens, line 22: ...or `throw new Error(Action of type `${action.type}` not implemented.)`?

playground/src/components/echo-server/echo-server.js
```diff
@@ -8,8 +8,23 @@ import Text from '@commercetools-uikit/text';
...
+  }
+};
+
 const EchoServer = () => {
-  const [result, setResult] = React.useState();
+  const [state, dispatchState] = React.useReducer(reducer, initialState);
```
tdeekens, line 27: It's a bit too much here maybe but usually encapsulating helps so:

```js
const { dispatchComplete, dispatchLoading } = useLoadingStateReducer()
```

emmenko: Yeah I just wanted to keep it simple. I can change that though, if you prefer so.

playground/src/components/echo-server/echo-server.js
emmenko: Hmm why? The `default` case returns the inital state.

playground/src/components/echo-server/echo-server.js
tdeekens: All good. Thanks for considering.

PR #1459, opened 2020-04-24
fix(express): do not append trailing slash to audience if request path is "/"

emmenko:
Consider your server is hosted at `https://foo.com`, when you send the forwarded to request to that, the audience is set to `https://foo.com`.
However, the middleware attempts to append the request path (in this case `/`) to the configured audience value, resulting in the audience to be verified to be `https://foo.com/`.

Doing so, it will fail the token verification because of the mismatching audiences.

The fix is to not append the request path if it's just `/`.

cc @qmateub

### DIFF
--- a/packages-backend/express/src/auth.ts
+++ b/packages-backend/express/src/auth.ts
@@ -89,7 +89,7 @@ const validateRequiredValues = (options: TSessionMiddlewareOptions) => {
 };
 // Attempt to parse the given issuer. If the value is a cloud identifier, it will
 // be mapped to one of the supported values. If not, we assume the value is a valid URL.
-const parseDefaultIssuer = (options: TSessionMiddlewareOptions) => {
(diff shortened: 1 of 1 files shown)


packages-backend/express/src/auth.ts
```diff
@@ -89,7 +89,7 @@ const validateRequiredValues = (options: TSessionMiddlewareOptions) => {
 };
 // Attempt to parse the given issuer. If the value is a cloud identifier, it will
 // be mapped to one of the supported values. If not, we assume the value is a valid URL.
-const parseDefaultIssuer = (options: TSessionMiddlewareOptions) => {
+const getConfiguredDefaultIssuer = (options: TSessionMiddlewareOptions) => {
```
tdeekens, line 92: Yay 💯 :)

PR #1461, opened 2020-04-24
docs(website): rewrite docs for proxing to external API

emmenko:


website/src/content/main-concepts/proxy-to-external-api.mdx
```diff
@@ -16,74 +18,213 @@ Merchant Center applications can make [authenticated requests](/main-concepts/au
...
+
+Requests to that endpoint should additionally pass the following HTTP headers:
 
-To be able to validate that the request comes from an authenticated user, we implemented a mechanism to allow your server to verify the validity of the `mcAccessToken`.
+- `Accept-version: v2`: See [Versioning](#versioning).
+- `X-Forward-To: <url>`: The URL to your external API.
+- `X-Project-Key: <project-key>`: The key of the project currently being used by the Custom Application. The Merchant Center API Gateway will perform a validation check to ensure that the user has access to the project, then forward the request to your server only if the check was successful.
 
-Requests to your backend server should be configured as following:
+To facilitate the usage of the [built-in Apollo client and the SDK actions](/main-concepts/data-fetching), we provide some utility helpers to abstract away the request configuration for the HTTP headers listed above.
```
tdeekens, line 35: utility helpers feels a bit double :)

```suggestion
To facilitate the usage of the [built-in Apollo client and the SDK actions](/main-concepts/data-fetching), we provide some helpers to easily integrate with the request configuration for the HTTP headers listed above.
```

website/src/content/main-concepts/proxy-to-external-api.mdx
```diff
@@ -16,74 +18,213 @@ Merchant Center applications can make [authenticated requests](/main-concepts/au
...
 
-- send the request to `/proxy/forward-to` endpoint of the Merchant Center API Gateway.
-- define the original request URL to your backend server as an HTTP header `X-Forward-To`.
+## Usage for Apollo
+
+We can leverage the `context` option for Apollo queries to change a bit how the request is configured. The `@commercetools-frontend/application-shell` package now exposes an utility function to configure the Apollo context for the `/proxy/forward-to` usage.
```
tdeekens, line 39:
```suggestion
We can leverage the `context` option for Apollo queries to adjust how request is configured and executed. The `@commercetools-frontend/application-shell` package now exposes an utility function to configure the Apollo context for the `/proxy/forward-to` usage.
```

website/src/content/main-concepts/proxy-to-external-api.mdx
```diff
@@ -16,74 +18,213 @@ Merchant Center applications can make [authenticated requests](/main-concepts/au
...
-  headers: {
-    Accept: 'application/json',
-    'X-Forward-To': 'https://my-custom-app.com/foo/bar',
-  },
-  credentials: 'include',
+import { useQuery } from 'react-apollo';
+import { createApolloContextForProxyForwardTo } from '@commercetools-frontend/application-shell';
+const Fetcher = () => {
+  const { loading, data, error } = useQuery(MyQuery, {
+    context: createApolloContextForProxyForwardTo({
```
tdeekens, line 46: This is technically correct but I would suggest to set it up like this.

I would suggest that people use a `forward-to-apollo-contexts.js`-module

```js
const customServiceApolloContext = createApolloContextForProxyForwardTo({
  uri: 'https://my-custom-app.com/graphql',
})

export { customServiceApolloContext } 
```

to then more convienently do:

```js
import { customServiceApolloContext } from './forward-to-apollo-contexts'

const { loading, data, error } = useQuery(MyQuery, {
   context: customServiceApolloContext
})
```

cause I would assume they have many queries and mutations.

website/src/content/main-concepts/proxy-to-external-api.mdx
```diff
@@ -16,74 +18,213 @@ Merchant Center applications can make [authenticated requests](/main-concepts/au
...
-# Validating authenticated requests
+# Proxying requests to the external API
+
+To be able to validate that the request comes from an authenticated user, the Merchant Center API Gateway provides an endpoint that must be used to connect to your external API in a secured manner.
+
+```
+/proxy/forward-to
+```
+
+Requests to that endpoint should additionally pass the following HTTP headers:
```
FreddieK, line 29: Should as that it's optional and not required?

website/src/content/main-concepts/proxy-to-external-api.mdx
```diff
@@ -16,74 +18,213 @@ Merchant Center applications can make [authenticated requests](/main-concepts/au
...
+When a valid request is sent via the `/proxy/forward-to` endpoint, the Merchant Center API Gateway forwards the request to the external API with an `Authorization: Bearer <token>` HTTP header.
+
+The bearer token is a short-living JSON Web Token (JWT) that is used exclusively for the authorization exchange between the Merchant Center API Gateway and the external API. The token is valid for **60 seconds**.
+
+The external API must verify that the token is valid by [using the JSON Web Key Set endpoint](#using-json-web-key-set-endpoint).
 
 <Warning>
 
-For security reasons, the exchanged JWT has an expiration date of 60 seconds, as the token should only be used to verify that the request is authorized.
+It's imperative that the external API is securely protected and that only users with access to the project are able to connect to the API.
```
tdeekens, line 85:
```suggestion
It's imperative that the external API is securely protected and that only users with access to the project should able to connect to the API.
```

website/src/content/main-concepts/proxy-to-external-api.mdx
```diff
@@ -16,74 +18,213 @@ Merchant Center applications can make [authenticated requests](/main-concepts/au
...
 </Warning>
 
-## Using JSON Web Key Set endpoint
 
-At this point, your backend server will receive the request with the `Authorization` header, which you can use to validate that the JWT is valid.
-The validation is performed by using the `/.well-known/jwks.json` endpoint exposed by the Merchant Center API Gateway. You can read more about JWKS [here](https://auth0.com/docs/jwks) and [here](https://tools.ietf.org/html/rfc7517).
+## Validating the JSON Web Token
 
-The endpoint exposes a public key which can be used to verify the JWT signature, according to the OIDC spec. We also expose a `/.well-known/openid-configuration` discovery endpoint.
+To facilitate the implementation of validating the JWT from the `Authorization: Bearer <token>` HTTP header, we provide a package with built-in functions and helpers to perform the heavy work.
```
tdeekens, line 92:
```suggestion
To facilitate the implementation of validating the JWT from the `Authorization: Bearer <token>` HTTP header, we provide a package with built-in functions and helpers to facilitate the heavy lifting.
```

website/src/content/main-concepts/proxy-to-external-api.mdx
```diff
@@ -16,74 +18,213 @@ Merchant Center applications can make [authenticated requests](/main-concepts/au
...
+
+The middleware requires some options:
+
+- `audience` (_string_): The public-facing URL of your API server. The value should only contain the origin URL (protocol, hostname, port), the request path is inferred from the incoming request.
+
+  For example, given the external API is hosted at `https://my-api-server.com`, the `audience` value must be set to `https://my-api-server.com`.
+
+- `issuer` (_string_): Either a cloud identifier or a valid URL to the Merchant Center API Gateway. The cloud identifier maps to the Merchant Center API URL of the related [cloud region](/main-concepts/api-gateway#cloud-regions).
+
+  | Cloud Identifier | Merchant Center API Gateway URL |
```
tdeekens, line 119: Nit: format the table to match rows and content maybe.

website/src/content/main-concepts/proxy-to-external-api.mdx
```diff
@@ -16,74 +18,213 @@ Merchant Center applications can make [authenticated requests](/main-concepts/au
...
+  | --- | --- |
+  | `gcp-au` | `https://mc-api.australia-southeast1.gcp.commercetools.com` |
+  | `gcp-eu` | `https://mc-api.europe-west1.gcp.commercetools.com` |
+  | `gcp-us` | `https://mc-api.us-central1.gcp.commercetools.com` |
+  | `aws-fra` | `https://mc-api.eu-central-1.aws.commercetools.com` |
+  | `aws-ohio` | `https://mc-api.us-east-2.aws.commercetools.com` |
+
+- `inferIssuer` (_boolean_): Determines whether the issuer should be inferred from the custom request HTTP header `x-mc-api-cloud-identifier` which is sent by the Merchant Center API Gateway when forwarding the request. This might be useful in case the server is used in multiple regions. **Default: `false`**
```
tdeekens, line 127: The golden option.

website/src/content/main-concepts/proxy-to-external-api.mdx
```diff
@@ -16,74 +18,213 @@ Merchant Center applications can make [authenticated requests](/main-concepts/au
...
+
+- `jwks` (_object_): See [options of `jwks-rsa`](https://github.com/auth0/node-jwks-rsa/).
+
+Finally, you can use the middleware as following:
```
tdeekens, line 131: Finally is a bit odd cause we just explain above :)

```suggestion
You can use the middleware as following:
```

website/src/content/main-concepts/proxy-to-external-api.mdx
```diff
@@ -16,74 +18,213 @@ Merchant Center applications can make [authenticated requests](/main-concepts/au
...
+    response.status(401).send(JSON.stringify({ message: 'Unauthorized' }));
+    return;
+  }
+  // `request.session` contains the useful information
+};
 ```
+
+## Using JSON Web Key Set endpoint
+
+The Merchant Center API Gateway exposes a `/.well-known/jwks.json` endpoint that can be used to validate JWTs, as per Open ID Connect (OIDC) spec. The endpoint provides a rotating public key used to verify the JWT signature.
```
tdeekens, line 184:
```suggestion
The Merchant Center API Gateway exposes a `/.well-known/jwks.json` endpoint that can be used to validate JWTs, as defined Open ID Connect (OIDC) specification. The endpoint provides a rotating public key used to verify the JWT signature.
```

website/src/content/main-concepts/proxy-to-external-api.mdx
```diff
@@ -16,74 +18,213 @@ Merchant Center applications can make [authenticated requests](/main-concepts/au
...
+Additionally, there is also a `/.well-known/openid-configuration` discovery endpoint.
+
+You can read more about JSON Web Key Set [here](https://auth0.com/docs/jwks) and [here](https://tools.ietf.org/html/rfc7517).
+
+<Info>
+
+If you use the `@commercetools-backend/express` package, you don't need to do anything in this regard.
```
tdeekens, line 191:
```suggestion
The setup described above is not needed when using the `@commercetools-backend/express` package which contains all needed setup already.
```

website/src/content/main-concepts/proxy-to-external-api.mdx
```diff
@@ -16,74 +18,213 @@ Merchant Center applications can make [authenticated requests](/main-concepts/au
...
+
+| Feature maturity     | Deprecation period |
+| -------------------- | ------------------ |
+| Beta                 | 4 months           |
+| General Availability | 12 months          |
+
+The following table lists all the supported versions and their deprecation period:
+
+| Version  | Is default |  Deprecated from  |
+| -------- | ---------- | ---------- |
```
tdeekens, line 221: Nit: format.

website/src/content/main-concepts/proxy-to-external-api.mdx
```diff
@@ -16,74 +18,213 @@ Merchant Center applications can make [authenticated requests](/main-concepts/au
...
+The current version refers to the **default** version used by the API, if no `Accept-version` header is provided.
+
+When we introduce a new version, the default version enters a **deprecation period**, after which the version is marked as deprecated and the new version becomes the **default** version.
+
+| Feature maturity     | Deprecation period |
+| -------------------- | ------------------ |
+| Beta                 | 4 months           |
+| General Availability | 12 months          |
+
+The following table lists all the supported versions and their deprecation period:
```
tdeekens, line 218:
```suggestion
The following table lists all the supported versions and their potential respective deprecation start period:
```

website/src/content/main-concepts/proxy-to-external-api.mdx
emmenko: At the end it does not matter much. If you use `v2`, the `X-Project-Key` header is validated. If you use the complementary tools that we built (which are documented here), the headers are automatically set.

website/src/content/main-concepts/proxy-to-external-api.mdx
emmenko: It's not that simple. Most likely the URL is defined in the `env.json`, so it's passed through the app context.

I could maybe update the example with that, if that makes it clearer.

Defining the config statically only works if you hardcode the URL.

website/src/content/main-concepts/proxy-to-external-api.mdx
emmenko: We're repeating `facilitate`. We should use another word

website/src/content/main-concepts/proxy-to-external-api.mdx
emmenko: I would prefer it like this, otherwise it gets messy when looking at diffs.

website/src/content/main-concepts/proxy-to-external-api.mdx
tdeekens: Yeah makes sense. I wonder though how many have variable endpoints. I'd make the docs work for the majority center best practices around that and no have minior edge cases influence the majority.

website/src/content/main-concepts/proxy-to-external-api.mdx
tdeekens: True, happy to leave it as is then.

website/src/content/main-concepts/proxy-to-external-api.mdx
tdeekens: If we plan support variable endpoints then we should maybe have a `useProxyForwardToContext`-hook which people can use and which encapsulates a respective or configurable env.json value

```js
const forwardToCustomServiceContext = useProxyForwardToContext({ envKey: 'foo' })
const { loading, data, error } = useQuery(MyQuery, {
   context: forwardToCustomServiceContext
})
```

emmenko: In general you would always want to use variables. I also saw a couple of cases where `env.json` has been used. Which is actually how it was intended to be used.

I understand your point of encapsulating things but it just gets hard to find the balance to accommodate different use cases.

I think I would just update the doc example to read the value from the app context. People can then decide on their own how to access the URL.

tdeekens: Makes sense. Thanks for taking things into consideration!

PR #1463, opened 2020-04-25
refactor(permissions): use Record for dataFences

adnasa:
## Summary

- refactored `DataFences` to use `Record`

#### Descripton

I read up about `Record` and was curious about its use case. ref https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkt

This seemed a really awesome utility type and I could see that there is application for it.

For dataFences, we know that for each `key` in `DataFences` should represent a type (e.g `store`) and with that type have a specific shape, a group of resource type.

namely the following:
```
{
  [<dataFenceType]: {
    [<resourceType>]: {
      [<permissionName>]: {
        values: string[]
      }
    }
  }
}
```

I doubt that the dataFence model will change (at least for a while). we can use `Record` to solidify this domain constraint within the code.

#### With just `store`

image

#### Adding values, not supported by the Record

image

#### Extending the record, by adding more `dataFence` types to support.
image

### DIFF
--- a/packages/application-shell-connectors/src/components/application-context/application-context.tsx
+++ b/packages/application-shell-connectors/src/components/application-context/application-context.tsx
@@ -44,17 +44,11 @@ type TApplicationContextGroupedByResourceType = {
  *   }
  * }
  */
-// NOTE: we currently can't use Mapped Types as the babel transfomer does not
-// understand them yet: https://github.com/milesj/babel-plugin-typescript-to-proptypes/issues/23
-// type TApplicationContextDataFenceType = 'store';
-// type TApplicationContextDataFences = {
-//   // E.g. { store: {...} }
-//   [key in TApplicationContextDataFenceType]: TApplicationContextGroupedByResourceType;
-// };
-type TApplicationContextDataFences = {
-  // E.g. { store: {...} }
-  store?: TApplicationContextGroupedByResourceType;
-};
+type TApplicationContextDataFenceType = 'store';
+type TApplicationContextDataFences = Record<
+  TApplicationContextDataFenceType,
+  TApplicationContextGroupedByResourceType
+>;
 type TApplicationContextEnvironment = ApplicationWindow['app'];
 
 const Context = React.createContext({});
--- a/packages/application-shell-connectors/src/components/application-context/application-context.tsx
+++ b/packages/application-shell-connectors/src/components/application-context/application-context.tsx
@@ -45,9 +45,11 @@ type TApplicationContextGroupedByResourceType = {
  * }
  */
 type TApplicationContextDataFenceType = 'store';
-type TApplicationContextDataFences = Record<
-  TApplicationContextDataFenceType,
-  TApplicationContextGroupedByResourceType
+type TApplicationContextDataFences = Partial<
+  Record<
+    TApplicationContextDataFenceType,
+    TApplicationContextGroupedByResourceType
+  >
 >;
 type TApplicationContextEnvironment = ApplicationWindow['app'];
 
--- a/packages/application-shell-connectors/src/components/application-context/normalizers.ts
+++ b/packages/application-shell-connectors/src/components/application-context/normalizers.ts
@@ -105,7 +105,7 @@ type NormalizedStoresGroupByResourceType = {
 
 const normalizeAppliedDataFencesForStoresByResourceType = (
   dataFences: Maybe<TStoreDataFence>[]
-): NormalizedGroupedByResourceType => {
+) => {
   const groupedByResourceType = dataFences.reduce<
     NormalizedStoresGroupByResourceType
   >((previousGroupsOfSameType, appliedDataFence) => {
@@ -183,7 +183,7 @@ type NormalizedDataFenceRecord = Record<
 
 export const normalizeAllAppliedDataFences = (
   project: TFetchProjectQuery['project']
-): NormalizedDataFenceRecord | null => {
+) => {
   if (!project) return null;
   const allAppliedDataFences = project.allAppliedDataFences;
   if (!allAppliedDataFences || allAppliedDataFences.length === 0) {
--- a/packages/application-shell-connectors/src/components/application-context/normalizers.ts
+++ b/packages/application-shell-connectors/src/components/application-context/normalizers.ts
@@ -98,23 +98,14 @@ type NormalizedGroupedByResourceType = {
   // E.g. { orders: {...} }
   [key: string]: NormalizedGroupedByPermission | null;
 };
-/**
- * dataFence: {
- *   store: {
- *     orders: {
- *       canManageOrders: { values: ['usa', 'germany'] },
- *       canViewOrders: { values: ['canada'] },
- *     }
- *   }
- * }
- */
 type NormalizedStoresGroupByResourceType = {
+  // E.g [ { group: 'orders', value: 'us', type: 'store', name: 'canManageOrders' } ]
   [key: string]: TStoreDataFence[];
 };
 
 const normalizeAppliedDataFencesForStoresByResourceType = (
   dataFences: Maybe<TStoreDataFence>[]
-) => {
+): NormalizedGroupedByResourceType => {
   const groupedByResourceType = dataFences.reduce<
     NormalizedStoresGroupByResourceType
   >((previousGroupsOfSameType, appliedDataFence) => {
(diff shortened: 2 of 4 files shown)


adnasa:
yes. I'm already aware that I'm a$$hole for dropping a Saturday night PR ❤️ 

XOXO

packages/application-shell-connectors/src/components/application-context/normalizers.ts
```diff
@@ -98,23 +98,14 @@ type NormalizedGroupedByResourceType = {
...
- */
 type NormalizedStoresGroupByResourceType = {
+  // E.g [ { group: 'orders', value: 'us', type: 'store', name: 'canManageOrders' } ]
   [key: string]: TStoreDataFence[];
 };
 
 const normalizeAppliedDataFencesForStoresByResourceType = (
   dataFences: Maybe<TStoreDataFence>[]
-) => {
+): NormalizedGroupedByResourceType => {
```
tdeekens, line 108: The return type can not be inferred anymore? I usually felt it's a good pattern to type inputs and have the return type be inferred.

packages/application-shell-connectors/src/components/application-context/normalizers.ts
```diff
@@ -98,23 +98,14 @@ type NormalizedGroupedByResourceType = {
...
- * dataFence: {
- *   store: {
- *     orders: {
- *       canManageOrders: { values: ['usa', 'germany'] },
- *       canViewOrders: { values: ['canada'] },
- *     }
- *   }
- * }
- */
 type NormalizedStoresGroupByResourceType = {
```
tdeekens, line 101: Not subject matter in this PR but this often confuses me. what normalized implies here.

packages/application-shell-connectors/src/components/application-context/normalizers.ts
```diff
@@ -184,9 +175,15 @@ type NormalizedGroupByType = {
...
+type SupportedDataFenceType = 'store';
+type NormalizedDataFenceRecord = Record<
+  SupportedDataFenceType,
+  NormalizedGroupedByResourceType
+>;
+
 export const normalizeAllAppliedDataFences = (
   project: TFetchProjectQuery['project']
-) => {
+): NormalizedDataFenceRecord | null => {
```
tdeekens, line 186: Same here? Do we need the return type?

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.ts
```diff
@@ -37,16 +37,8 @@ type TDataFenceGroupedByResourceType = {
...
-// type TDataFences = {
-//   // E.g. { store: {...} }
-//   [key in TDataFenceType]: TDataFenceGroupedByResourceType;
-// };
-type TDataFences = {
-  // E.g. { store: {...} }
-  store?: TDataFenceGroupedByResourceType;
-};
+type TDataFenceType = 'store';
+type TDataFences = Record<TDataFenceType, TDataFenceGroupedByResourceType>;
```
tdeekens, line 41: This is nice I think.

packages/application-shell-connectors/src/components/application-context/normalizers.ts
adnasa: aha.. I like explicit 😅 
but sure, let's get rid of this then

packages/application-shell-connectors/src/components/application-context/normalizers.ts
adnasa: hmm, I think I get an issue below if I don't do this.
let me check

adnasa: yepp, we can remove this, but I need L214, otherwise we get this error

```
error TS2322: Type '{ store?: NormalizedGroupedByResourceType; } | null' is not assignable to type 'Record<"store", TApplicationContextGroupedByResourceType> | null'.
  Type '{ store?: NormalizedGroupedByResourceType; }' is not assignable to type 'Record<"store", TApplicationContextGroupedByResourceType>'.
    Property 'store' is optional in type '{ store?: NormalizedGroupedByResourceType; }' but required in type 'Record<"store", TApplicationContextGroupedByResourceType>'.

145   dataFences: normalizeAllAppliedDataFences(project),
      ~~~~~~~~~~

  packages/application-shell-connectors/src/components/application-context/application-context.tsx:111:3
    111   dataFences: TApplicationContextDataFences | null;
          ~~~~~~~~~~
    The expected type comes from property 'dataFences' which is declared here on type 'TApplicationContext<AdditionalEnvironmentProperties>'

packages/application-shell-connectors/src/components/application-context/normalizers.ts:178:6 - error TS6196: 'SupportedDataFenceType' is declared but never used.

178 type SupportedDataFenceType = 'store';
```

adnasa: yepp, we can get rid of this.
I needed L214, otherwise we get this error

```bash
error TS2322: Type '{ store?: NormalizedGroupedByResourceType; } | null' is not assignable to type 'Record<"store", TApplicationContextGroupedByResourceType> | null'.
  Type '{ store?: NormalizedGroupedByResourceType; }' is not assignable to type 'Record<"store", TApplicationContextGroupedByResourceType>'.
    Property 'store' is optional in type '{ store?: NormalizedGroupedByResourceType; }' but required in type 'Record<"store", TApplicationContextGroupedByResourceType>'.

145   dataFences: normalizeAllAppliedDataFences(project),
      ~~~~~~~~~~

  packages/application-shell-connectors/src/components/application-context/application-context.tsx:111:3
    111   dataFences: TApplicationContextDataFences | null;
          ~~~~~~~~~~
    The expected type comes from property 'dataFences' which is declared here on type 'TApplicationContext<AdditionalEnvironmentProperties>'

packages/application-shell-connectors/src/components/application-context/normalizers.ts:178:6 - error TS6196: 'SupportedDataFenceType' is declared but never used.

178 type SupportedDataFenceType = 'store';
```

packages/application-shell-connectors/src/components/application-context/normalizers.ts
tdeekens: I actually did it often at first too cause I felt explicit is good. Then when thinking about flow of types and data through the program one realises that it is redundant. If a wants A and returns AA while b wants AA while returning BB. Then it does matter if you type the return of A to AA or if you have TS infer it. If you change what A returns (implicitly or the explicit) TS will warn you that B‘s input of AA is not fulfilled anymore.

packages/application-shell-connectors/src/components/application-context/normalizers.ts
adnasa: I would not call this normalized either. it's the raw datafence we get from the API

adnasa: we can follow this up some time.
adding it to my backlog

emmenko: > it's the raw datafence we get from the API

Not really, from the API we get a list of data fences

```
type StoreDataFence implements DataFence {
  value: String!
  name: String!
  group: String!
  type: String!
}
union AppliedDataFence = StoreDataFence

allAppliedDataFences: [AppliedDataFence!]!
```

And here we need to normalize the list into a nested object.

I don't know how else you want to call this 🤔

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -44,17 +44,11 @@ type TApplicationContextGroupedByResourceType = {
...
-// NOTE: we currently can't use Mapped Types as the babel transfomer does not
-// understand them yet: https://github.com/milesj/babel-plugin-typescript-to-proptypes/issues/23
-// type TApplicationContextDataFenceType = 'store';
-// type TApplicationContextDataFences = {
-//   // E.g. { store: {...} }
-//   [key in TApplicationContextDataFenceType]: TApplicationContextGroupedByResourceType;
-// };
-type TApplicationContextDataFences = {
-  // E.g. { store: {...} }
-  store?: TApplicationContextGroupedByResourceType;
```
emmenko, line 56: To double check: the `store` was marked as an optional property before. How does it work with a `Record`?

adnasa: 🔥 that is a nice catch!
I assumed that `store` will be optional, but that is not the case. 😒
This can be rewritten to use `Partial` of `Record`, if we want.

```ts
const TDataFences = Partial<Record<DataFenceTypes, TApplicationContextGroupedByResourceType>>
```

packages/application-shell-connectors/src/components/application-context/normalizers.ts
tdeekens: Thanks. Makes sense. Forgot about this union to allow potentially future types of data fences.

adnasa:
### using `Partial<Record<K, V>>`

image

as you can see, `categories` is defined among `TDataFenceType` however missing on the `dataFences` value object and we get no warning.

Adding a key missing from `TDataFenceType` however:
image

---

let's go for this 😉

tdeekens:
Lovely.

adnasa:
there we go, my dudes..
can't wait for the build to pass.

adnasa:
oh come on 🥺

adnasa:
we can leave this as canary. no rush to release this

Issue #1466, opened 2020-04-27
[Bug] Deployment problems with firebase while using latest version of `mc-scripts`


qmateub:
**Describe the issue**

Upgrading to the latest version of `mc-scripts` which is `16.7.0` causes an error while doing a deploying of a custom app that is hosted in [firebase](https://docs.commercetools.com/custom-applications/deployment/example-deployment-firebase).

The error basically consists in a node version incompatibility because firebase (as well as Google Cloud functions) only supports until Node 10 (which is still in beta).

The line that was causing the error was `Object.fromEntries ` used in `@commercetools-frontend/mc-html-template/lib/utils/substitute-env-variable-placeholders.js:81:17`

**To Reproduce**

1. Create a basic custom app with latest releases of application-kit
2. Deploy
3. Try to run locally not using node12 or deploy to a firebase project

**Expected behavior**

Ability to deploy with node versions >= 10.

Our workaround consisted in:

- Downgrade packages `eslint-config-mc-app`, `jest-preset-mc-app`, `mc-scripts` to release `16.0.0`.
- Use resolutions field in `package.json` to fix some babel preset errors (facebook/create-react-app#8680)

**Additional context**

https://firebase.google.com/docs/functions/manage-functions#set_runtime_options

PR #1468, opened 2020-04-27
refactor(mc-script): compile it through babel, to support node 10

emmenko:
This PR addresses the issue described in #1466 for apps using an old node version.

TL;DR: compile the code with Babel.

Other PRs will follow to migrate the other "nodejs" packages.


packages/mc-scripts/README.md
```diff
@@ -4,10 +4,22 @@
...
+## Usage
+
+The package exports some functions to configure Webpack
+
+```js
+const {
+  createWebpackConfigForDevelopment,
+  createWebpackConfigForProduction,
+  vendorsToCompile,
+} = require('@commercetools-frontend/mc-scripts');
```
emmenko, line 24: This is new, but there is proxy `config` folder for backwards compatibility.

packages/mc-scripts/package.json
```diff
@@ -79,8 +93,26 @@
...
+        {
+          "targets": {
+            "node": "10"
+          },
+          "modules": "commonjs",
+          "useBuiltIns": "usage",
+          "corejs": 3
+        }
+      ]
+    ]
```
emmenko, line 116: I decided to add the babel config here, so it's co-located with the package.

application-templates/starter/webpack.config.prod.js
```diff
@@ -1,5 +1,7 @@
 const path = require('path');
-const createWebpackConfigForProduction = require('@commercetools-frontend/mc-scripts/config/create-webpack-config-for-production');
+const {
+  createWebpackConfigForProduction,
+} = require('@commercetools-frontend/mc-scripts');
```
tdeekens, line 4: Is the change in export here a breaking change or do we keep the default export for now?

packages/mc-scripts/package.json
```diff
@@ -79,8 +93,26 @@
     "webpack-dev-server": "3.10.3",
     "webpackbar": "4.0.0"
   },
+  "devDependencies": {
+    "@babel/preset-env": "7.9.5"
+  },
   "engines": {
-    "node": ">=8",
+    "node": ">=10",
```
emmenko, line 100: Not 100% sure about this, but we should determine a base line most likely.

packages/mc-scripts/README.md
tdeekens: Ah, thanks. Disregard my comment regarding that in the review then.

PR #1470, opened 2020-04-27
refactor(mc-html-template): compile it through babel, to support node 10

emmenko:
Follow up of #1468 to address the issues in #1466


packages/jest-preset-mc-app/package.json
```diff
@@ -18,7 +18,6 @@
   "main": "./jest-preset.js",
   "module": "./jest-preset.js",
   "dependencies": {
-    "@babel/core": "7.9.0",
```
emmenko, line 21: It's a dev dependency apparently

image

packages/mc-html-template/package.json
```diff
@@ -15,16 +15,50 @@
...
     "uglify-es": "3.3.9",
     "uglifycss": "0.0.29"
   },
+  "devDependencies": {
+    "@babel/plugin-transform-runtime": "7.9.0",
+    "@babel/preset-env": "7.9.5"
+  },
   "engines": {
-    "node": ">=8",
+    "node": ">=10",
```
tdeekens, line 37: Quite apparently breaking in a way but... Hm.

packages/mc-html-template/package.json
```diff
@@ -15,16 +15,50 @@
...
+        }
+      ]
+    ],
+    "plugins": [
+      [
+        "@babel/plugin-transform-runtime",
+        {
+          "corejs": 3,
+          "helpers": false
+        }
```
emmenko, line 60: We also need this otherwise things like `Object.entries` won't get transformed

packages/mc-html-template/package.json
```diff
@@ -15,16 +15,50 @@
...
+    "node": ">=10",
     "npm": ">=5"
+  },
+  "babel": {
+    "presets": [
+      [
+        "@babel/preset-env",
+        {
+          "targets": {
+            "node": "10"
```
tdeekens, line 46: Ah. As the nvm/ci etc are using 13 we can't use `true` here?

packages/mc-html-template/package.json
emmenko: Yeah that's what I asked in the other PR: https://github.com/commercetools/merchant-center-application-kit/pull/1468#discussion_r415729515

Since we are now targeting node 10 with babel, it doesn't make sense to keep node 8.

packages/mc-html-template/package.json
emmenko: Yes correct

packages/mc-html-template/package.json
tdeekens: Yeah I think it's fine. It wasn't working in the previous version anyway I presume.

PR #1471, opened 2020-04-28
refactor: to use official rollup plugins

tdeekens:
#### Summary

Two more rollup plugins have been moved into their monorepo. 

#### Description

This is hard to forget otherwise. The old plugins are deprecated and one will hardly see/wonder why they don't get any further updates. Just before we/I forget.


rollup.config.js
```diff
@@ -38,7 +38,7 @@ const createPlugins = (format) => {
     }),
     babel({
       extensions,
-      runtimeHelpers: true,
+      babelHelpers: 'runtime',
```
tdeekens, line 41: CleanShot 2020-04-28 at 13 57 09@2x

Should be the same as I read the docs. /cc @emmenko

emmenko: Should we specify `@babel/runtime` as a dependency of each package, as described?

tdeekens: Sounds like it, yes.

tdeekens: We actually already have it as a dep of the mc-scripts and mc-html-template. Adding it to the root will not change much.

Still getting this error I have to investigate. Seems like it does some preflight compilation checking if the configuration can be achieved (e.g. runtime) and then determines that it can't and opts out.

```txt
./src/index.ts → dist/browser-history.cjs.js...
[!] (plugin babel) Error: You have declared using "runtime" babelHelpers, but transforming /Users/tdeekens/Development/merchant-center-application-kit/packages/browser-history/src/index.ts resulted in "inline". Please check your configuration.
src/index.ts
Error: You have declared using "runtime" babelHelpers, but transforming /Users/tdeekens/Development/merchant-center-application-kit/packages/browser-history/src/index.ts resulted in "inline". Please check your configuration.
    at error (/Users/tdeekens/Development/merchant-center-application-kit/node_modules/rollup/dist/shared/rollup.js:10399:30)
```

tdeekens: I've tried a few things by now. I couldn't get it to work with `runtime`. Even when declaring `@babel/runtime` as external etc. I also don't want to skip the runtime check (config option) as that feels silly.

I am actually presuming that we always inlined the helpers. I checked the plugin's migration PR (into their mono-repo) and they added these checks. So I presume we specified external before (the renamed config option) but actually didn't get it.

I am trying to bundled option now which states

```txt
'bundled' - you should use it if you want your resulting bundle to contain those helpers (at most one copy of each). Useful especially if you bundle an application code.
```

which sounds half false but okay-ish.

I now build with old and new config and wanted to compare the bundle sizes of the app-shell (a bot would be dope for this 😄).

###### Before

before

###### After

after

There we can see that the bundle sizes didn't even change. I am left a bit in awe 😄

emmenko: Hmm so `bundled` is the default option, but from a library perspective we would like to use `runtime`. Did I get it right?

packages/babel-preset-mc-app/index.js
```diff
@@ -106,7 +106,8 @@ module.exports = function getBabePresetConfigForMcApp() {
         require('@babel/plugin-transform-runtime').default,
         {
           corejs: 3,
-          helpers: false,
+          // To be able to use `runtime` in Rollup babel plugin
+          helpers: true,
```
emmenko, line 110: @tdeekens This does the trick ⬆️

tdeekens: Then it was misconfigured before I think. Really odd.

tdeekens:
⚠️ Rebased out the conflicts ⚠️

packages/babel-preset-mc-app/index.js
emmenko: No idea why we had it configured that way 🤷‍♂️

Issue #1444, opened 2020-04-16
[Idea] Provide a package to help setting up an "external API" server


emmenko:
It would be helpful to have a package that takes care of most of the setup that an "external API" server would need to work with Custom Applications.

For example, including the JWKs middleware, extracting useful fields from the token, etc.

This would also help with migrations for the API versioning.

tdeekens:
Maybe just an express middleware, then maybe hapi plugin etc exposing certain things on the request would help plugging things together and getting started. A lot of that cruft we could actually steal from promster which works with many servers by now. Alternatively it could also be a complete server but then that could also build on the previous building blocks. Depends a bit on how much existing infrastructure Custom Apps on the server integrate with or if they start from scratch.

mmaltsev-ct:
Keeping the idea, with the low priority.

PR #1472, opened 2020-04-28
feat: add loggers package

emmenko:
Partially part of #1444 

This PR open sources our internal loggers package, used by our backend services.


packages-backend/loggers/src/custom-formats/json.ts
```diff
@@ -0,0 +1,55 @@
...
+import unsetIn from 'lodash/unset';
+
+/*
+ * function replacer (key, value)
+ * Handles proper stringification of Buffer output.
+ */
+function replacer(key: string, value: Buffer | string) {
+  if (key === 'queryJsonString') {
+    // We need to stringify the value as otherwise Kibana cardinality explodes.
+    // Ref: https://commercetools.slack.com/archives/C011FLZ1JLW/p1588066334005900
```
tdeekens, line 19: I think we should remove references from internal Slack convos.

packages-backend/loggers/src/create-error-report-logger.ts
```diff
@@ -0,0 +1,89 @@
...
+    DSN: string;
+    role: string;
+    environment: string;
+  };
+  errorMessageBlacklist?: Array<string | RegExp>;
+};
+
+function createErrorReportLogger(options: LoggerOptions = {}) {
+  // Sentry
+  if (options.sentry) {
```
tdeekens, line 20: Hm. Personally I find it rather specific that we use Sentry. Do we know if others using this package would too or should we allow for more reporting targets?

emmenko: In my defense, the package says "_Opinionated JSON loggers_" 😜

emmenko:
I understand your points, let's have a chat then 🙂

packages-backend/loggers/src/custom-formats/json.ts
```diff
@@ -0,0 +1,54 @@
...
+ * object into pure JSON. This was previously exposed as { json: true }
+ * to transports in `winston < 3.0.0`.
+ */
+/* eslint-disable no-param-reassign */
+const jsonFormatter = format((info, opts = {}) => {
+  info.logLevel = info.level;
+  delete info.level;
+
+  // Replace / rename fields
+  replaceField(info, 'meta.req.query', 'meta.req.queryJsonString');
```
tdeekens, line 48: Should be an option `omitMetaPaths: String[]` I think.

packages-backend/loggers/src/create-access-logger.ts
```diff
@@ -0,0 +1,51 @@
...
+
+// Inspired by https://github.com/bithavoc/express-winston/issues/62#issuecomment-396906056
+const requestFilter: RequestFilter = (req, propName) => {
+  if (propName === 'headers') {
+    return redactInsecureRequestHeaders(req);
+  }
+  return req[propName];
+};
+
+const createAccessLogger = (options: LoggerOptions = {}) => {
```
tdeekens, line 26: This is a middleware I think. So to have it inline with others maybe it's `createAccessLoggerMiddleware(options)`.

packages-backend/loggers/src/env.ts
```diff
@@ -0,0 +1,6 @@
+const env = process.env.NODE_ENV;
+const isDev = !env || env === 'development';
+const isProd = env === 'production';
+const isTest = env === 'test';
+
+export { isDev, isProd, isTest };
```
tdeekens, line 6: I would also have this as an arg to the middlewars. `envGetters` maybe. Not 100% also just making notes for myself.

packages-backend/loggers/src/custom-formats/json.ts
emmenko: Yeah, probably we need to provide some kind of fields mapping.

```js
{
  rewriteMetaInfoFields: {
    'meta.req.query': 'meta.req.queryJsonString'
  }
}
```

An alternative might be to allow to pass a custom `logform` formatter. We can also expose a bunch of default formatter (e.g. json, kibana, etc). The kibana formatter can also be additionally configured with the fields rewriter.

packages-backend/loggers/src/create-error-report-logger.ts
```diff
@@ -0,0 +1,89 @@
+import type { Request, Response, NextFunction } from 'express';
+
+import * as Sentry from '@sentry/node';
+import * as env from './env';
+import redactInsecureRequestHeaders from './utils/redact-insecure-request-headers';
```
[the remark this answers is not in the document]
emmenko: PS: one of the intents with this loggers package was to provide preconfigured loggers, meaning that for common cases there should be no need to provide extra configuration.

We could maybe have a default list of "insecure" headers, while still allowing to extend that list.

packages-backend/loggers/src/custom-formats/json.ts
tdeekens: Yeah. I think sane defaults which can be overwritten would be good. Maybe also by just saying: I want a Kibana logger as a type argument you get those defaults.

emmenko:
I cleaned things up and remove all unnecessary things. I also removed Sentry, as I think we can keep this internal for now.

@tdeekens can you take another look?

packages-backend/loggers/src/index.ts
```diff
@@ -0,0 +1,12 @@
...
+
+// Loggers
+export { default as createApplicationLogger } from './create-application-logger';
+
+// Formatters
+export { default as rewriteFieldsFormatter } from './formatters/rewrite-fields';
+
+// Re-export winston for convenience
+import * as winston from 'winston';
+export { winston };
```
tdeekens, line 12: Nit: doesn't TS allow `export * as winston` from 'winston'`?

PR #1474, opened 2020-04-29
fix(babel): keep runtime helpers

emmenko:
Given that in #1471 we fixed the runtime helpers config, we should do the same here.

### DIFF
--- a/packages/mc-html-template/package.json
+++ b/packages/mc-html-template/package.json
@@ -69,8 +69,7 @@
       [
         "@babel/plugin-transform-runtime",
         {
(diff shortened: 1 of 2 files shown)


packages/mc-html-template/package.json
```diff
@@ -69,8 +69,7 @@
       [
         "@babel/plugin-transform-runtime",
         {
-          "corejs": 3,
-          "helpers": false
```
emmenko, line 73: It's `true` by default: https://babeljs.io/docs/en/babel-plugin-transform-runtime#helpers

PR #1477, opened 2020-05-05
fix(loggers): silent option

emmenko:
Leftover of #1472

### DIFF
--- a/packages-backend/loggers/src/create-application-logger.ts
+++ b/packages-backend/loggers/src/create-application-logger.ts
@@ -3,7 +3,6 @@ import type { TLoggerOptions } from './types';
 import * as winston from 'winston';
 
 const createApplicationLogger = (options: TLoggerOptions = {}) => {
-  const shouldLog = Boolean(options.silent);
   const formatters = winston.format.combine(
     ...(options.formatters ?? []),
     options.json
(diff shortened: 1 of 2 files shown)


packages-backend/loggers/src/create-application-logger.ts
```diff
@@ -3,7 +3,6 @@ import type { TLoggerOptions } from './types';
 import * as winston from 'winston';
 
 const createApplicationLogger = (options: TLoggerOptions = {}) => {
-  const shouldLog = Boolean(options.silent);
```
emmenko, line 6: 🤦‍♂️

packages-backend/loggers/src/middlewares/create-access-logger.ts
```diff
@@ -19,7 +18,8 @@ const createAccessLoggerMiddleware = (options: TAccessLoggerOptions = {}) => {
     meta: true,
     expressFormat: true, // Use default morgan access log formatting
     colorize: !options.json,
-    skip: (req) => !shouldLog || ignoreUrls.includes(req.originalUrl),
+    skip: (req) =>
+      Boolean(options.silent) || ignoreUrls.includes(req.originalUrl),
```
tdeekens, line 22: Almost feel we could achieve the same with instead of having ignoreUrls having the skip function as an option. So it would be like

```js
skip (...args) => Boolean(options.silent) || options.skip(...args)
```

I guess that would mean some refactoring in our apps though.

emmenko: Yeah, I think passing a list of URLs is less work and more straight-forward though. After all the intent with the package is to have some sort of pre-configured loggers. 🤗

PR #1478, opened 2020-05-06
refactor(loggers): refine types

emmenko:


packages-backend/loggers/src/formatters/rewrite-fields.ts
```diff
@@ -8,29 +8,25 @@ import unsetIn from 'lodash/unset';
 type TRewriteField = {
   from: string;
   to: string;
-  replaceValue?: (value: string) => unknown;
+  replaceValue?: (value: unknown) => unknown;
```
emmenko, line 11: This is not always a string.

packages-backend/loggers/src/formatters/rewrite-fields.ts
```diff
@@ -8,29 +8,25 @@ import unsetIn from 'lodash/unset';
...
   if (val) {
-    unsetIn(info, jsonPath);
-    setIn(info, newJsonPath, replaceValue ? replaceValue(val) : val);
+    unsetIn(info, field.from);
+    setIn(info, field.to, field.replaceValue ? field.replaceValue(val) : val);
   }
 }
 
-const rewriteFieldsFormatter = format(
+const rewriteFieldsFormatter: TCreateFormatter = format(
```
emmenko, line 26: The format wrapper has the options argument typed as `any`, so we explicitly type the function with our options type.

image

packages-backend/loggers/src/formatters/rewrite-fields.ts
emmenko: I might have find a way to use a generic...

packages-backend/loggers/src/formatters/rewrite-fields.ts
tdeekens: Nice catch!

emmenko:
Putting on draft as I'm still working on it.

packages-backend/loggers/src/formatters/rewrite-fields.ts
```diff
@@ -5,35 +5,35 @@ import getIn from 'lodash/get';
 import setIn from 'lodash/set';
 import unsetIn from 'lodash/unset';
 
-type TRewriteField = {
+type TRewriteField<FieldValue> = {
   from: string;
   to: string;
-  replaceValue?: (value: string) => unknown;
+  replaceValue?: (value: FieldValue) => unknown;
```
emmenko, line 11: The field can now be typed using a generic.

adnasa: nice!

PR #1481, opened 2020-05-07
refactor: upgrade to jest v26

tdeekens:
#### Summary

Jest v26 is out (https://jestjs.io/blog/2020/05/05/jest-26). The upgrade here should be relatively smooth due to the jsdom v16 adoption work earlier.

PR #1482, opened 2020-05-08
chore: lint package.json

emmenko:
I was spooking around other repositories and found this tool to check `package.json` for monorepos.

https://www.npmjs.com/package/@manypkg/cli


application-templates/starter/package.json
```diff
@@ -33,10 +33,10 @@
     "redux": "4.0.5"
   },
   "devDependencies": {
-    "@commercetools-frontend/jest-preset-mc-app": "16.8.0",
-    "@commercetools-frontend/mc-scripts": "16.8.0",
+    "@commercetools-frontend/jest-preset-mc-app": "*",
```
emmenko, line 36: All the "version" changes in this PR are a consequence of running `manypkg fix`

packages/jest-preset-mc-app/package.json
```diff
@@ -50,11 +50,13 @@
...
+    "@testing-library/react": "10.0.4",
+    "enzyme": "3.11.0",
+    "jest": "26.0.1"
   },
   "peerDependencies": {
-    "@testing-library/react": "9.x || 10.x",
+    "@testing-library/react": ">=9 || >=10",
     "enzyme": "3.x",
-    "jest": "24.x || 25.x"
+    "jest": ">=24 || >=25 || >=26"
```
emmenko, line 60: With the recent PR to upgrade Jest #1481, should we still keep v24?

tdeekens:
Interesting. To me the changes look sane but I would like to check in more details. Is there any issues this fixes directly or rules we like or „just“ something we see valuable?

packages/jest-preset-mc-app/package.json
tdeekens: We kept 25 as it would be breaking. We kept 25 as it would be breaking 🎁🤢. To me it‘s fine. There is nothing preventing others using old versions. Maybe we can batch it for a breaking change.

emmenko:
I just noticed that many version ranges have been changed to exact version. I'll fix that.

The value I see with this is to have consistent declared dependencies...it's also just an experiment to see "how useful in reality" this is. We can always remove it again if we have problems.

emmenko:
Mäh, the versions just get screwed up, which is not what we want at the end.

Can't see us benefiting from this after all. Going to close this.

PR #1493, opened 2020-05-11
refactor(mc-http-server): to publish to cloud registry when a git tag is created

emmenko:
The goal with this PR is to build and publish a docker image whenever we release a new version of e.g. `mc-http-server`.

Up until now we have been doing this using Google Cloud Build. However, with the introduction of Changesets, each package gets its own git tag, which is something like `@commercetools-frontend/mc-http-server@16.8.1` as opposed to `v16.8.1` (as we had before).

NOTE: the `GCR_KEY` secret is still missing. We need to add this in the secrets first before merging this PR.


emmenko:
The secret has been added now. This is ready for review.

PR #1503, opened 2020-05-14
feat: support experimental header x-application-id

emmenko:


emmenko:
Working on that already

emmenko:
@tdeekens anything else from your side?

PR #1507, opened 2020-05-16
refactor(use-navbar-state): use enum for action-types

adnasa:
#### Summary

I just want to give `changeset` a spin 😜

### DIFF
--- a/.changeset/eleven-emus-lick.md
+++ b/.changeset/eleven-emus-lick.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': patch
+---
+
+move action-types to enum
--- a/packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
+++ b/packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
@@ -21,7 +21,7 @@ import useApplicationsMenu from '../../hooks/use-applications-menu';
 import FetchProjectExtensionsNavbar from './fetch-project-extensions-navbar.settings.graphql';
 import nonNullable from './non-nullable';
 
-enum ActionTypes {
+const enum ActionTypes {
   setActiveItemIndex = 'setActiveItemIndex',
   unsetActiveItemIndex = 'unsetActiveItemIndex',
   setIsExpanderVisible = 'setIsExpanderVisible',
--- a/packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
+++ b/packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
@@ -21,6 +21,15 @@ import useApplicationsMenu from '../../hooks/use-applications-menu';
 import FetchProjectExtensionsNavbar from './fetch-project-extensions-navbar.settings.graphql';
 import nonNullable from './non-nullable';
 
+enum ActionTypes {
+  setActiveItemIndex = 'setActiveItemIndex',
+  unsetActiveItemIndex = 'unsetActiveItemIndex',
+  setIsExpanderVisible = 'setIsExpanderVisible',
+  toggleIsMenuOpen = 'toggleIsMenuOpen',
+  setIsMenuOpenAndMakeExpanderVisible = 'setIsMenuOpenAndMakeExpanderVisible',
+  reset = 'reset',
+}
+
 type HookProps = {
   environment: TApplicationContext<{}>['environment'];
   DEV_ONLY__loadNavbarMenuConfig?: () => Promise<TApplicationsMenu['navBar']>;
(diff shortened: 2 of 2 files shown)


packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
```diff
@@ -129,9 +138,10 @@ const useNavbarStateManager = (props: HookProps) => {
     ? null
     : (JSON.parse(cachedIsForcedMenuOpen) as boolean);
 
-  const [state, dispatch] = React.useReducer<
-    (prevState: State, action: Action) => State
```
adnasa, line 133: this is a bit too verbose IMO, so I decided to remove it.
It is already inferred through the reducer signature 😉

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
```diff
@@ -31,12 +40,12 @@ type State = {
   isMenuOpen: boolean;
 };
 type Action =
-  | { type: 'setActiveItemIndex'; payload: string }
-  | { type: 'unsetActiveItemIndex' }
-  | { type: 'setIsExpanderVisible' }
-  | { type: 'toggleIsMenuOpen' }
-  | { type: 'setIsMenuOpenAndMakeExpanderVisible'; payload: boolean }
-  | { type: 'reset' };
+  | { type: ActionTypes.setActiveItemIndex; payload: string }
```
adnasa, line 43: If I understood enums right, 
it's enough that this is `{ type: ActionTypes }`, but this is more explicit so I keep it this way

.changeset/eleven-emus-lick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': patch
+---
+
+move action-types to enum
```
adnasa, line 5: 🤩

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
```diff
@@ -21,6 +21,15 @@ import useApplicationsMenu from '../../hooks/use-applications-menu';
...
 import nonNullable from './non-nullable';
 
+enum ActionTypes {
+  setActiveItemIndex = 'setActiveItemIndex',
+  unsetActiveItemIndex = 'unsetActiveItemIndex',
+  setIsExpanderVisible = 'setIsExpanderVisible',
+  toggleIsMenuOpen = 'toggleIsMenuOpen',
+  setIsMenuOpenAndMakeExpanderVisible = 'setIsMenuOpenAndMakeExpanderVisible',
+  reset = 'reset',
+}
```
tdeekens, line 31: Can we use a const enum here? I presume this is not exported. In essence it saves bundle size. 

More here: https://ncjamieson.com/dont-export-const-enums/

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
```diff
@@ -21,6 +21,15 @@ import useApplicationsMenu from '../../hooks/use-applications-menu';
 import FetchProjectExtensionsNavbar from './fetch-project-extensions-navbar.settings.graphql';
 import nonNullable from './non-nullable';
 
+enum ActionTypes {
```
tdeekens, line 24: Nit: Prefix with T. We somewhat settled on it in app-kit so that types are easily distinguishable from anything else.

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
tdeekens: Nice if it is inferred than we can remove it surely.

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
adnasa: that is a nice one! ❤️ 
image

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
adnasa: we don't do that for everything else in this file. May I keep it this way?

tdeekens:
True. Forgot about that. There was a babel plugin to enable them if we want. But I guess as const works just fine and we can wait for babel to natively support them.

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
tdeekens: Apologies we can‘t use it as we use babel unless we add a plugin for the transform. You‘d have to use as const. Maybe it was still a nice learning regarding TS itself ✌🏼

tdeekens:
@adnasa was there an issue with the pull request why it was closed?

PR #1518, opened 2020-05-18
refactor: update to typescript 3.9

tdeekens:
#### Summary

This pull request attempts to update to TypeScript 3.9. I had to exclude it earlier today.

CleanShot 2020-05-18 at 20 53 45

I haven't found the root of the issue but it seems in external types. Also the release notes do not mention something directly. Only around relating "exports" areas.

### DIFF
--- a/.changeset/tasty-beers-compete.md
+++ b/.changeset/tasty-beers-compete.md
@@ -1,0 +1,18 @@
+---
+"@commercetools-backend/loggers": patch
+---
+
+refactor: update to typescript 3.9
+
+The re-export of winston from the `@commercetools-backend/loggers` package has been removed. 
+
+Given the `winston` types use an `export = winston` our package cannot re-export using `export * from`. If you relied on this re-export we suggest to install winston yourself and import it regularily.
+
+After running `yarn add winston` you should change the following
+
+```diff
+-import { winston } from '@commercetools-backend/loggers'
++import winston from 'winston'
+```
+
+We might add `winston` as a peer dependency in a future release to ensure that both, ours and your, `winston` versions remain in sync.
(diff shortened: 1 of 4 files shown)


tdeekens:
I am pretty sure it comes from [here](https://github.com/winstonjs/winston/blob/9354721af36af51b127f489a366d5d8602a37049/index.d.ts#L193)

tdeekens:
Added a PR on their end. Let's see. https://github.com/winstonjs/winston/pull/1799

tdeekens:
It is all a bit odd. The issue is more here. We can * reexport a export =. Why do we do import * and export * and not „just“ import winston? Lots of examples outline it. I presume it doesn’t work?

tdeekens:
I don't see how we can fix this unless we stop re-exporting winston. The `export =` in winston doesn't work with `export * as` within our module.

emmenko:
Sorry I haven't had the time to look into this. 

> unless we stop re-exporting winston

You mean this? https://github.com/commercetools/merchant-center-application-kit/blob/662565b84a832cce9c946a9b62e1f1e1f246972f/packages-backend/loggers/src/index.ts#L10-L11

I suppose we can remove it, it was just added for convenience in case you need to add more formatters. We could maybe make `winston` a peer dependency if we remove the export.

tdeekens:
> I suppose we can remove it, it was just added for convenience in case you need to add more formatters. We could maybe make winston a peer dependency if we remove the export.

I think making it a peer dependency would make sense. However that would be breaking for the package I presume.

I tried many things to work around the issue.

1. PR to winston
2. Changing the imports/exports

Nothing really works cause semantics are different. Maybe we can remove the re-export. Are we using it ourselves?

emmenko:
> Are we using it ourselves?

I don't think so.

I'm fine if we remove it yes.

PR #1533, opened 2020-05-28
chore: ugrade to react-intl v4

emmenko:
I tried things locally and I didn't get any error this time. So maybe "it just works" now?

Eventually, we can also remove `babel-plugin-macros`. I don't know of any use case we have that require the macro plugin.

### DIFF
--- a/.changeset/slow-comics-grin.md
+++ b/.changeset/slow-comics-grin.md
@@ -1,0 +1,13 @@
+---
+'merchant-center-application-template-starter': major
+'@commercetools-frontend/application-components': major
+'@commercetools-frontend/application-shell': major
+'@commercetools-frontend/mc-scripts': major
+'@commercetools-frontend/react-notifications': major
+'playground': major
+'@commercetools-local/visual-testing-app': major
+'@commercetools-website/custom-applications': major
+'@commercetools-website/components-playground': major
+---
+
+Upgrade to `react-intl` v4. See also https://formatjs.io/docs/react-intl/upgrade-guide-4x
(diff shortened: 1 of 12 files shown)


tdeekens:
@emmenko can we actually say what's breaking on our end here towards consumers? It's late, maybe I don't see it :)

emmenko:
The version bump of the peer dependency is what makes it breaking.

tdeekens:
Just didn’t see the peer dep bump. Thought we could „claim“ we just work with v3 or v4.

emmenko:
In theory yes, but we also don't know what features of v3 users are using. So only supporting v4 seems a safer bet.

I don't have a super strong opinion though.

tdeekens:
Do we have a list of potential features breaking? We could also add it as a range of `3.x || 4.x` for now if possible.

tdeekens:
@emmenko any input regarding my previous comment ⏫.

emmenko:
Here are the documented changes: https://formatjs.io/docs/react-intl/upgrade-guide-4x

Do you prefer to keep the range then?

tdeekens:
> Do you prefer to keep the range then?

What I thought was to allow users to use both 3.x and 4.x. What they use then depends on how their messages should look like and what's allowed. Instead of forcing them to migrate to the newer version which we don't have to require.

Does that make sense? Maybe I am reading/read the release notes wrong.

adnasa:
Adding @ahmehri to the PR as he is refactoring routing.

ahmehri:
@adnasa I couldn't find the relationship between this PR and the refactoring of routing, can you elaborate on that?

adnasa:
I never checked the file changes so I just assumed they are related. 
but I think the main reason I change assignee is because my attention is needed else where and you attempted to upgrade this yourself. so it's worth looking into? https://github.com/commercetools/merchant-center-frontend/pull/8138

but if you don't want to review, all good. I'll do it

PR #1534, opened 2020-05-29
refactor(i18n): expose hook for AsyncLocaleData

emmenko:
- expose a `useAsyncLocaleData` hook. The `<AsyncLocaleData>` component now uses the hook internally.
- expose a `useAsyncIntlMessages` hook, to only load messages with dynamic imports (code splitting).

The second hook is useful for building UI packages (not custom apps) that ship with translations. The translation message should be encapsulated with the package itself and ideally they should be code splitted as well.
The UI package then can use a `IntlProvider` (which will be nested when used within a Custom Application) to load the messages of the package itself.
The `IntlProvider` then merges the messages with the "parent(s)" providers.

### DIFF
--- a/packages/application-shell/src/components/project-suspended/project-suspended.spec.js
+++ b/packages/application-shell/src/components/project-suspended/project-suspended.spec.js
@@ -1,14 +1,14 @@
 import React from 'react';
 import { Route } from 'react-router-dom';
 import { GRAPHQL_TARGETS } from '@commercetools-frontend/constants';
-import { renderApp, waitFor } from '../../test-utils';
+import { renderApp, screen } from '../../test-utils';
 import UserQuery from '../fetch-user/fetch-user.mc.graphql';
 import ProjectsQuery from '../project-switcher/project-switcher.mc.graphql';
 import ProjectSuspended from './project-suspended';
 
 describe('rendering', () => {
   it('when suspension is temporary it should print correct message', async () => {
-    const rendered = renderApp(
+    renderApp(
       <Route
         path="/:projectKey"
         render={() => <ProjectSuspended isTemporary={true} />}
@@ -98,12 +98,9 @@ describe('rendering', () => {
         ],
       }
     );
-    await waitFor(() => {
-      expect(
-        rendered.getByText(
-          /Your Project is temporarily suspended due to maintenance/
-        )
-      ).toBeInTheDocument();
-    });
+    await screen.findByText(
+      /Your Project is temporarily suspended due to maintenance/
+    );
+    await screen.findByText('Search for a project');
   });
 });
--- a/packages/i18n/README.md
+++ b/packages/i18n/README.md
@@ -20,13 +20,19 @@ $ npm install --save @commercetools-frontend/i18n
 - `fr-FR`
 - `zh-CN`
 
-### Usage
+## Usage
 
-> This package should not be used directly to load the application messages, the `application-shell` does that internally.
+> The `<AsyncLocaleData>` component, or the `useAsyncLocaleData` hook, are internally used by the `<ApplicationShell>` and there is no need to use them directly.
+> If you do need to load translation files asynchronously, we expose an additional hook `useAsyncIntlMessages` that you can use for that.
+> The difference between those is that the "async locale data" components additionally load the application-kit and ui-kit messages, as well as the moment locales.
+
+### Using a React component with render props
+
+**With static messages**
 
 ```js
+import { IntlProvider } from 'react-intl';
 import { AsyncLocaleData } from '@commercetools-frontend/i18n';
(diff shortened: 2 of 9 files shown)


packages/i18n/README.md
```diff
@@ -20,21 +20,27 @@ $ npm install --save @commercetools-frontend/i18n
...
+> The `<AsyncLocaleData>` component, or the `useAsyncLocaleData` hook, are internally used by the `<ApplicationShell>` and there is no need to use them directly.
+> If you do need to load translation files asynchronously, we expose an additional hook `useAsyncIntlMessages` that you can use for that.
+> The difference between those is that the "async locale data" components additionally load the application-kit and ui-kit messages, as well as the moment locales.
+
+### Using a React component with render props
+
+**With static messages**
 
 ```js
+import { IntlProvider } from 'react-intl';
```
tdeekens, line 34: This change is unrelated?

packages/i18n/src/async-locale-data/async-locale-data.tsx
```diff
@@ -32,63 +34,56 @@ const getMessagesForLocale = (data?: TMessages, locale?: string) => {
...
+  const loadApplicationMessages = React.useCallback(
+    async (locale: string) => {
+      if (typeof applicationMessages === 'function') {
+        return await applicationMessages(locale);
+      }
+      return getMessagesForLocale(applicationMessages, locale);
+    },
+    [applicationMessages]
+  );
+  const messagesFromKitResult = useAsyncIntlMessages({
```
tdeekens, line 50: Kit is AppKit?

packages/i18n/src/async-locale-data/use-async-intl-messages.ts
```diff
@@ -0,0 +1,52 @@
...
+  error: undefined,
+};
+
+// Low level hook to load messages for a specific locale. The loading is async
+// because it's assumed that the translation files are dynamically imported (code splitted).
+const useAsyncIntlMessages = ({ locale, loader }: THookOptions): TState => {
+  const [state, setState] = React.useState(initialState);
+
+  React.useEffect(() => {
+    let _isUnmounting = false;
```
tdeekens, line 29: I think I do the same in flopflip and use a ref inside a hook. Like `const isUnmountedRef = useIsUnmounted()`. Can't recall I just remember it took longer than I thought to get right. I'll check later.

emmenko: I took this approach from reading this: https://stackoverflow.com/a/56443045

packages/i18n/src/async-locale-data/async-locale-data.tsx
emmenko: Both, app and ui kit

packages/i18n/README.md
emmenko: Yes and now. Since I'm updating the README though, to avoid confusion I think it would be better to keep the example generic and use `IntlProvider`, not our internal `ConfigureIntlProvider`.

emmenko:
FYI: not a high prio to review. Whenever you have time...

adnasa:
I can't review this today, sorry. I'll have a look at this next week.

packages/application-shell/src/components/project-suspended/project-suspended.spec.js
```diff
@@ -98,12 +98,9 @@ describe('rendering', () => {
...
-      expect(
-        rendered.getByText(
-          /Your Project is temporarily suspended due to maintenance/
-        )
-      ).toBeInTheDocument();
-    });
+    await screen.findByText(
+      /Your Project is temporarily suspended due to maintenance/
+    );
+    await screen.findByText('Search for a project');
```
adnasa, line 104: I'm liking `screen` alot

packages/i18n/README.md
```diff
@@ -100,28 +106,51 @@ const Application = (props) => (
...
+Similarly to the `<AsyncLocaleData>`, we also expose a React hook.
+
+```js
+import { IntlProvider } from 'react-intl';
+import { useAsyncLocaleData } from '@commercetools-frontend/i18n';
+
+const Application = (props) => {
+  const { isLoading, locale, messages } = useAsyncLocaleData({
+    locale: props.locale,
+    applicationMessages: loadMessages,
```
adnasa, line 129: where is `loadMessages` coming from?

tdeekens:
@emmenko should we proceed here?

packages/i18n/README.md
emmenko: It's defined in the example above. Didn't want to copy-paste the function again.

emmenko:
Yep, we can merge this.

PR #1535, opened 2020-05-29
refactor(mc-http-server): to use Node.js v14

tdeekens:
#### Summary

As we just realised v13 has a very short current window. Given the mc-http-server contains very little deps/code I think it's valuable to bump it to v14 now before we forget.

At the same time we can spin up our dev version in the nvmrc to v14 too :)


tdeekens:
Once published I would integrate it into a branch internally.

tdeekens:
> PS: once this gets released, the CI job should trigger the docker build

Yup I know. Reviewed that thing :)

PR #1541, opened 2020-06-03
fix(ci): to publish to docker in same workflow

tdeekens:
#### Summary

This pull request changes the actions workflows to publish to the docker registry in the same workflow.

#### Description

The previous step up did not work as

> There is a limitation of workflow: An action in a workflow run can’t trigger a new workflow run.

This means workflow A can't trigger workflow B by pushing a tag. It's kind of amazing how GitHub knows that a tag was pushed from another workflow.

The suggested change here is to split the release workflow in two jobs: `npm` and `docker`. It might be easier to have one but it feels cleaner to have two if possible.

We use the `publishedPackages` variable from the changeset action.

CleanShot 2020-06-03 at 09 08 59

We then use the GitHub Action expression syntax for match on the array and find if one item is present.

CleanShot 2020-06-03 at 09 08 07


.github/workflows/release.yml
```diff
@@ -6,7 +6,7 @@ on:
       - master
 
 jobs:
-  release:
+  npm:
```
tdeekens, line 9: Two jobs now in release workflow: `npm` and `docker`.

.github/workflows/release.yml
```diff
@@ -73,3 +73,37 @@ jobs:
         if: steps.changesets.outputs.published != 'true'
         # TODO: this command will be replaced by the changesets snapshots feature
         run: yarn lerna publish 0.0.0-canary-${{ github.sha }} --exact --dist-tag canary --yes --no-push --force-publish "*"
+
+  docker:
+    runs-on: ubuntu-latest
+
+    if: contains(jobs.npm.steps.changesets.outputs.publishedPackages.*.name, '@commercetools-frontend/mc-http-server')
```
tdeekens, line 80: I hope we can reference across jobs. Haven't seen many do that but my understanding is that it should work :)

adnasa:
> The problem here is that the github.ref is not a tag, so we can't detect the tag version from that.

so even with the split of the workflow, this won't work?
I guess it's hard to verify without actually running the workflow.

tdeekens:
> so even with the split of the workflow, this won't work?

It won't work. I checked the docs last nite and forgot this thought. We basically need to extract the version and name from the `publishedPackages`. This is not possible navitely. Given we proceed with this approach (which I'll mark as draft) we would have to:

1. Use `jq` and pass the `publishedPackages` as an input and pull out the version
2. Write a custom action receiving the `publishdPackages` as input and in JS extract the version
   - Here then we would have to use the GCloud JS SDK likely

We will try to use a custom GitHub token so we can have workflows split and trigger one by a tag push of the other.

tdeekens:
The token trick seems to work it's magic which is simpler than this.

PR #1542, opened 2020-06-03
fix(workflows): use a personal access token in checkout step, to allow the workflow to trigger a new workflow

emmenko:
This is an attempt to fix the issue with the cloud registry workflow not being triggered when a tag is created through another workflow.

From the information that I could find, it seems that this restriction can be lifted by using a personal access token.

Let's see if this makes a difference or not.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,6 +20,11 @@ jobs:
           # We could set `fetch-depth: 0`, but that would cause the entire history
           # to be cloned. Using 50 seems like a good balance to start with.
           fetch-depth: 50
+          # Pass a personal access token (using our `ct-release-bot` account) to be able to trigger
+          # other workflows
+          # https://help.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token
+          # https://github.community/t/action-does-not-trigger-another-on-push-tag-action/17148/8
+          token: ${{ secrets.RELEASE_GITHUB_TOKEN }}
 
       - name: Read .nvmrc
         run: echo ::set-output name=NVMRC::$(cat .nvmrc)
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -20,6 +20,11 @@ jobs:
           # We could set `fetch-depth: 0`, but that would cause the entire history
           # to be cloned. Using 50 seems like a good balance to start with.
           fetch-depth: 50
+          # Pass a personal access token (using our `ct-release-bot` account) to be able to trigger
+          # other workflows
+          # https://help.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token
+          # https://github.community/t/action-does-not-trigger-another-on-push-tag-action/17148/8
+          token: ${{ secrets.RELEASE_GITHUB_TOKEN }}
```
adnasa, line 27: I'm not sure where this is stored. 😬

tdeekens: Ah. No probs. Eer, it's in the settings in GitHub of the repo under secrets.

tdeekens:
I will try this out.

.github/workflows/release.yml
adnasa: found it!

tdeekens:
It seems like it was triggered but it was also skipped: https://github.com/commercetools/merchant-center-application-kit/actions/runs/123667981

emmenko:
Hmm ok, at least we have solved one issue :)

Maybe the conditions are not quite correct.

This:
https://github.com/commercetools/merchant-center-application-kit/blob/0c2f2bdd4c1411ad92eb3918a9f83eec35721c4b/.github/workflows/publish-cloud-registry.yml#L5-L6

And this:
https://github.com/commercetools/merchant-center-application-kit/blob/0c2f2bdd4c1411ad92eb3918a9f83eec35721c4b/.github/workflows/publish-cloud-registry.yml#L12

emmenko:
Maybe we should use `contains` instead of `startsWith`. I guess the `ref` starts with `tags/`

tdeekens:
I think so. Worked the GitHub Actions API for 15 minutes but failed to find something. Let's see.

PR #1545, opened 2020-06-03
fix(workflows): use contains() to check for matching tag ref

emmenko:
Fixes https://github.com/commercetools/merchant-center-application-kit/pull/1542#issuecomment-638174858


tdeekens:
I worked the API's of GitHub Actions our run and check_suite a bit. Couldn't find much but also couldn't retrieve the GitHub context of the run.

PR #1547, opened 2020-06-05
feat(app-shell): fetch organization name with owner

tdeekens:
#### Summary

This pull request adds fetching the owner's name which is the organization through the project.

#### Description

This will help more easily redirect SSO users to `login/sso/<orgName>` without the application having to fetch data. Overall the addition of owner.name feels small given we already load the id.

### DIFF
--- a/packages/application-shell/src/components/quick-access/quick-access.spec.js
+++ b/packages/application-shell/src/components/quick-access/quick-access.spec.js
@@ -1173,7 +1173,8 @@ describe('QuickAccess', () => {
               currencies: ['EUR', 'GBP'],
               languages: ['en'],
               owner: {
-                id: 'project-id-1',
+                id: 'organization-id-1',
+                name: 'Organization Name',
               },
               suspension: { isActive: false },
               expiry: { isActive: false },
@@ -1190,7 +1191,8 @@ describe('QuickAccess', () => {
                     currencies: ['EUR', 'GBP'],
                     languages: ['en'],
                     owner: {
-                      id: 'project-id-1',
+                      id: 'organization-id-1',
+                      name: 'Organization Name',
                     },
                     suspension: { isActive: false },
                     expiry: { isActive: false },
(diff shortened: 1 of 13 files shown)


packages/application-shell/src/components/quick-access/quick-access.spec.js
```diff
@@ -1173,7 +1173,8 @@ describe('QuickAccess', () => {
               currencies: ['EUR', 'GBP'],
               languages: ['en'],
               owner: {
-                id: 'project-id-1',
```
tdeekens, line 1176: This is wrong. It's not the project to my knowledge.

tdeekens:
I just noticed that we map `owner.id` into `ownerId` in the application context. @emmenko can you share why? I can't recall.

emmenko:
I think for simplicity?

emmenko:
Is this in review or not yet?

tdeekens:
> Is this in review or not yet?

Wasn’t sure either. I felt it would be nice to redirect sso users to the sso login screen with their previous resoective organization name prefilled. 

When I started looking into it I felt loading the owner.name would the „easiest“ way avoding a request upon logout to determine the organization name. 

Then I noticed that we map the owner.id into ownerId and stopped for a moment. Wondering if we:

1. Want to continue this idea if prefilling
2. Load the owner.name here

emmenko:
Should we then do `ownerName`? Or start having `owner` as an object, while still keeping `ownerId` for backwards compatibility?

The question is also what kind of info we might want to expose about the project organization. Probably id and name would be enough.

tdeekens:
Yes. If we want to persue the idea it is either 

1. ownerId and ownerName
2. owner.id and owner.name while ownerId for backwards compatibility 

I think we did ownerId to not give the impression that you have a whole org under owner.*. That argument would still hold I guess. Yielding going for option 1. However, I struggle having a string opinion 😔😋.

tdeekens:
Let's go for `ownerName`. Ready for review 😄.

PR #1550, opened 2020-06-07
refactor(release): publish canary using changeset snapshot feature

emmenko:
The new shapshot feature got released recently. I tried that in a personal project and it works.


emmenko:
For publishing yes. We still use Lerna CLI to run some useful “multi-package” commands, which are not possible with `yarn workspaces`.

tdeekens:
Hm. There is yarn workspaces run now. Would that help? https://yarnpkg.com/lang/en/docs/cli/workspaces/

emmenko:
Options like `--no-private` is what I'm missing at the moment.

tdeekens:
Too bad 😋. Thanks for checking.

PR #1561, opened 2020-06-09
refactor(notifications): do not report to Sentry unmapped errors

emmenko:
We get from time to time a lot of unmapped error reports on Sentry. To avoid spamming Sentry, we just log a warning instead. We also don't seem to "care much" about mapping those errors, so there is not much of a point to keep reporting them.

If I assumed wrong, please feel free to keep things as they are.

### DIFF
--- a/.changeset/fifty-onions-suffer.md
+++ b/.changeset/fifty-onions-suffer.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/react-notifications': patch
+---
+
+Do not report to Sentry unmapped API errors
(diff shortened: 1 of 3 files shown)


tdeekens:
To my memory we actually want these cause we otherwise have no way of keeping track of them or fixing them. We could log them to Kibana or through the `metrics` service too but it felt more complicated. That's what my memory tells me.

adnasa:
> We also don't seem to "care much" about mapping those errors, so there is not much of a point to keep reporting them.

I'm sad to hear this.
Personally I want these to keep track of which features that are use and missing new ErrorCode from the API to keep pushing for new ErrorCode.

I don't think the problem here is that _we don't care_, but rather a few of us know how to produce this while developing new feature and where to escalate the problem when we discover an InvalidOperation.

https://github.com/commercetools/fe-chapter-notes/issues/33

adnasa:
> If I assumed wrong, please feel free to keep things as they are.

I propose to keep

emmenko:
Ok, sounds good to me. Let's maybe raise more the awareness to POs on how to handle this. Maybe we also need to improve the way we report those errors, so that proper action items can be planned.

Closing this then, thanks for the feedback.

PR #1570, opened 2020-06-12
build(eslint): enable react/no-unused-prop-types rule

ahmehri:
#### Summary

As a developer, I want to be notified when I'm defining a proptype that is not used in my component (similar to when I get notified when there are unused variables). This is very useful during refactoring, where I don't want to manually check whether a proptype is used.

#### Old architecture

Checking unused proptypes is done manually by the developer.

#### New architecture

Checking unused proptypes is done automatically by eslint [`react/no-unused-prop-types`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unused-prop-types.md) rule.

### DIFF
--- a/.changeset/dry-cars-kiss.md
+++ b/.changeset/dry-cars-kiss.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/eslint-config-mc-app': minor
+---
+
+enable `react/no-unused-prop-types` eslint rule
--- a/packages/eslint-config-mc-app/index.js
+++ b/packages/eslint-config-mc-app/index.js
@@ -73,6 +73,7 @@ module.exports = {
     'no-underscore-dangle': 0,
     'jest/no-identical-title': 'warn',
     'jest/no-focused-tests': 2,
+    /* eslint-plugin-react */
     'react/jsx-uses-vars': 2,
     'react/wrap-multilines': 0,
     'react/no-deprecated': 'error',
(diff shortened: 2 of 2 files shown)


packages/eslint-config-mc-app/index.js
```diff
@@ -73,6 +73,7 @@ module.exports = {
     'no-underscore-dangle': 0,
     'jest/no-identical-title': 'warn',
     'jest/no-focused-tests': 2,
+    /* eslint-plugin-react */
```
tdeekens, line 76: Is this comment intended?

ahmehri: Yes, my intent is to group the rules of the react plugin using that comment, so it's easier to spot them.

tdeekens:
As note @ahmehri one could see this as a breaking change. We add another linting rule which we previously didn't have. This might make consumers have to change their code. Which is the definition of breaking changes.

ahmehri:
@tdeekens ah yes that's true, then I need to update the changeset again by setting that I need a major version update, right? I set it to minor previously.

.changeset/dry-cars-kiss.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-frontend/eslint-config-mc-app': minor
```
ahmehri, line 2: I need to update this to major, right?

PR #1571, opened 2020-06-14
fix(app-shell): user settings menu link width for testing

tdeekens:
#### Summary

This pull request fixes the with of the `<a />` within the user settings menu. 

#### Description

The DOM structure of elements within the user settings is sort of like `div > div > div > a > TextNode` while the parent `div` has an adequate with the `a` within it has a computed height and width of 0.

CleanShot 2020-06-14 at 15 07 14

This is not acessible and also not nicely testable through Cypress. Cypress will not allow the click the element by `findByRole('link', { name: /logout/i })` as the `a` (link) itself has a height and width of 0 and is not clickable as a result. This forces us to use `findByText` which is suboptimal. Any other text on the page with `logout` would break the test if the selector wasn't scoped. Moreover, Cypress is right that the element has a wrong width and height.

The solution is to make the `a` an `inline-block` with `100%` width.

CleanShot 2020-06-14 at 15 07 01

### DIFF
--- a/cypress/integration/playground/application-shell.js
+++ b/cypress/integration/playground/application-shell.js
@@ -8,7 +8,7 @@ describe('when user is authenticated', () => {
     cy.login({ redirectToUri: URL_STATE_MACHINES });
 
     cy.findByRole('button', { name: /open user settings menu/i }).click();
-    cy.findByText('Logout').click();
+    cy.findByRole('link', { name: /logout/i }).click();
 
     const queryParams = encode({
       reason: LOGOUT_REASONS.USER,
--- a/packages/application-shell/src/components/user-settings-menu/user-settings-menu.tsx
+++ b/packages/application-shell/src/components/user-settings-menu/user-settings-menu.tsx
@@ -126,8 +126,7 @@ const MenuItem = styled.div<MenuItemProps>`
 `;
 
 const getUserSettingsMenuItemLinkStyles = () => css`
-  display: inline-block;
-  width: 100%;
+  display: block;
 `;
 
 const UserSettingsMenuBody = (props: MenuBodyProps) => {
--- a/packages/application-shell/src/components/user-settings-menu/user-settings-menu.tsx
+++ b/packages/application-shell/src/components/user-settings-menu/user-settings-menu.tsx
@@ -125,6 +125,11 @@ const MenuItem = styled.div<MenuItemProps>`
       : ''};
 `;
 
+const getUserSettingsMenuItemLinkStyles = () => css`
+  display: inline-block;
+  width: 100%;
+`;
+
 const UserSettingsMenuBody = (props: MenuBodyProps) => {
   const servedByProxy = useApplicationContext(
     (context) => context.environment.servedByProxy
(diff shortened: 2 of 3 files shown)


cypress/integration/playground/application-shell.js
```diff
@@ -6,8 +6,10 @@ import { URL_BASE, URL_STATE_MACHINES } from '../../support/urls';
 describe('when user is authenticated', () => {
   it('should log out with reason "user"', () => {
     cy.login({ redirectToUri: URL_STATE_MACHINES });
+
     cy.findByLabelText('open menu').click();
-    cy.findByText('Logout').click();
+    cy.findByRole('link', { name: /logout/i }).click();
```
tdeekens, line 11: Being able to use this selector is the goal of the PR.

packages/application-shell/src/components/user-settings-menu/user-settings-menu.tsx
```diff
@@ -125,6 +125,11 @@ const MenuItem = styled.div<MenuItemProps>`
       : ''};
 `;
 
+const getUserSettingsMenuItemLinkStyles = () => css`
+  display: inline-block;
+  width: 100%;
+`;
```
tdeekens, line 131: Extracted and shared for a readable name and as it's used 3 times below.

packages/application-shell/src/components/user-settings-menu/user-settings-menu.tsx
```diff
@@ -168,26 +173,27 @@ const UserSettingsMenuBody = (props: MenuBodyProps) => {
             </div>
           </Spacings.Inline>
         </Spacings.Inset>
-        {applicationsAppBarMenu &&
-          applicationsAppBarMenu.map((menu) => (
-            <OptionalFeatureToggle
-              key={menu.key}
-              featureToggle={menu.featureToggle}
+        {applicationsAppBarMenu?.map((menu) => (
```
tdeekens, line 176: Only changed due to the use of optional chaining here.

tdeekens:
Note, that this PR also allows us to internally in apps log out with proper selectors.

packages/application-shell/src/components/user-settings-menu/user-settings-menu.tsx
```diff
@@ -125,6 +125,11 @@ const MenuItem = styled.div<MenuItemProps>`
       : ''};
 `;
 
+const getUserSettingsMenuItemLinkStyles = () => css`
+  display: inline-block;
+  width: 100%;
```
jonnybel, line 130: I think this is the same as:
```suggestion
  display: block;
```

packages/application-shell/src/components/user-settings-menu/user-settings-menu.tsx
```diff
@@ -216,6 +223,7 @@ const UserSettingsMenuBody = (props: MenuBodyProps) => {
         </a>
         <MenuItem hasDivider={true} />
         <a
```
jonnybel, line 225: Btw since this is an `<a>` element, it's already selectable with tabs, so we could remove the `tabIndex={0}` on this line, right?  https://github.com/commercetools/merchant-center-application-kit/pull/1571/files#diff-8605567988365343f9b67c8b17dc2c7dL224

Just noticed this because when we try to navigate the menu items, the Logout item has two elements that get selected and the second one doesn't do anything when pressed and this seems to be the cause.

packages/application-shell/src/components/user-settings-menu/user-settings-menu.tsx
tdeekens: That is true. I think the tests should proove it: b46b4d4

packages/application-shell/src/components/user-settings-menu/user-settings-menu.tsx
tdeekens: Hm. Might be true would like to not mix this here though on the PR.

PR #1572, opened 2020-06-14
fix(app-shell): user settings menu labels for opening and closing menu

tdeekens:
#### Summary

This pull request fixes the user settings menu labels to be translated and custom named.

#### Description

Downshift's default naming of the `aria-label` is `open menu` and `close menu`. The library recommends to pass an own tranlation of this label

CleanShot 2020-06-14 at 16 17 15

This then allows us to use better selectors and also one which would not break with a second Downshift dropdown on the page.

CleanShot 2020-06-14 at 16 12 50

Note, that updating the tests and using the new labels prooves that this works.

### DIFF
--- a/cypress/integration/playground/application-shell.js
+++ b/cypress/integration/playground/application-shell.js
@@ -6,8 +6,10 @@ import { URL_BASE, URL_STATE_MACHINES } from '../../support/urls';
 describe('when user is authenticated', () => {
   it('should log out with reason "user"', () => {
     cy.login({ redirectToUri: URL_STATE_MACHINES });
-    cy.findByLabelText('open menu').click();
+
+    cy.findByRole('button', { name: /close user settings menu/i }).click();
     cy.findByText('Logout').click();
+
     const queryParams = encode({
       reason: LOGOUT_REASONS.USER,
     });
--- a/cypress/integration/playground/application-shell.js
+++ b/cypress/integration/playground/application-shell.js
@@ -7,7 +7,7 @@ describe('when user is authenticated', () => {
   it('should log out with reason "user"', () => {
     cy.login({ redirectToUri: URL_STATE_MACHINES });
 
-    cy.findByRole('button', { name: /close user settings menu/i }).click();
+    cy.findByRole('button', { name: /open user settings menu/i }).click();
     cy.findByText('Logout').click();
 
     const queryParams = encode({
--- a/packages/application-shell/src/components/user-settings-menu/messages.ts
+++ b/packages/application-shell/src/components/user-settings-menu/messages.ts
@@ -16,4 +16,14 @@ export default defineMessages({
     description: 'The label for privacy policy option',
     defaultMessage: 'Privacy Policy',
   },
+  openMenuLabel: {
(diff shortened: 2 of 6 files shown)


cypress/integration/playground/application-shell.js
```diff
@@ -6,8 +6,10 @@ import { URL_BASE, URL_STATE_MACHINES } from '../../support/urls';
 describe('when user is authenticated', () => {
   it('should log out with reason "user"', () => {
     cy.login({ redirectToUri: URL_STATE_MACHINES });
-    cy.findByLabelText('open menu').click();
+
+    cy.findByRole('button', { name: /close user settings menu/i }).click();
```
tdeekens, line 10: We call this user settings menu to my knowledge. Now w ecan target a button and the label instead of label text (which is strange in of itself).

packages/application-shell/src/components/user-settings-menu/messages.ts
```diff
@@ -16,4 +16,14 @@ export default defineMessages({
...
   },
+  openMenuLabel: {
+    id: 'UserSettingsMenu.openMenuLabel',
+    description: 'The label when menu is closed and would be opened by a click',
+    defaultMessage: 'Open User Settings Menu',
+  },
+  closeMenuLabel: {
+    id: 'UserSettingsMenu.closeMenuLabel',
+    description: 'The label when menu is open and would be closed by a click',
+    defaultMessage: 'Close User Settings Menu',
```
tdeekens, line 27: The elements are visible on screen. As a result it makes sense to also translate the label.

packages/application-shell/src/components/user-settings-menu/user-settings-menu.tsx
```diff
@@ -233,36 +233,49 @@ const UserSettingsMenuBody = (props: MenuBodyProps) => {
...
-          {downshiftProps.isOpen && (
-            <UserSettingsMenuBody {...props} downshiftProps={downshiftProps} />
-          )}
-        </div>
-      )}
-    </Downshift>
-  </div>
-);
+const UserSettingsMenu = (props: Props) => {
+  const intl = useIntl();
```
tdeekens, line 237: The indentation changes due to the usage of `useIntl`.

packages/application-shell/src/components/user-settings-menu/user-settings-menu.tsx
```diff
@@ -233,36 +233,49 @@ const UserSettingsMenuBody = (props: MenuBodyProps) => {
...
+                border: none;
+                padding: 0;
+                display: flex;
+                background: transparent;
+              `}
+              {...downshiftProps.getToggleButtonProps({
+                'aria-label': intl.formatMessage(
+                  downshiftProps.isOpen
+                    ? messages.closeMenuLabel
+                    : messages.openMenuLabel
```
tdeekens, line 257: This is the relevant change. If the menu is open we show a label to close it and if closed one to open it.

tdeekens:
Note, also here this allows us to target the menu better internally.

tdeekens:
@fuchodeveloper just for awareness. I don't think we need two reviews on app-kit for this.

PR #1579, opened 2020-06-17
feat(eslint-config-mc-app): to add jest-dom plugin

tdeekens:
#### Summary

This pull request adds the jest-dom plugin to our eslint config.

#### Description

Similar to our prior PR [here](https://github.com/commercetools/merchant-center-application-kit/pull/1570/files) we add another ESLint plugin.

The [eslint-plugin-jest-dom](https://github.com/testing-library/eslint-plugin-jest-dom) helps to use the right expectations towards checked, enabled state of DOM nodes etc.

### DIFF
--- a/packages/eslint-config-mc-app/index.js
+++ b/packages/eslint-config-mc-app/index.js
@@ -24,6 +24,7 @@ module.exports = {
   plugins: [
     'import',
     'jest',
+    'jest-dom',
     'jsx-a11y',
     'prettier',
     'prefer-object-spread',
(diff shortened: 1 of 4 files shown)


packages/eslint-config-mc-app/index.js
```diff
@@ -24,6 +24,7 @@ module.exports = {
   plugins: [
     'import',
     'jest',
+    'jest-dom',
```
ahmehri, line 27: I think adding only the plugin won't work because all the rules are disabled by default, as far as I recall. We can either enable them manually or use the [recommended configuration](https://github.com/testing-library/eslint-plugin-jest-dom#recommended-configuration).

tdeekens: Thanks. I missed this.

tdeekens: 9535232

PR #1583, opened 2020-06-18
feat(app-shell): to add label to maintenance page layouts

tdeekens:
#### Summary

This pull request adds a label which is used on the images alt to all maintenance pages.

#### Description

Our maintenance pages mostly have: title, two paragraphs and an image. The image itself however doesn't contain any alt text. It would be good to mostly use the `title` as the alt text on the image. 

This also eases querying for the image with `byRole('img', { name: //i })` under test.

I had multiple solution in mind when working on this:

1. Add a optional string of `label` which goes to the `img` alt. The `img` `alt` needs to be a string and can't be React.Node
2. Just use the `title` and pass it as the `alt`. Slighly less flexible but less noisy for consumers. However, the `title` is an React.Node as it's a FormattedMessage while the `alt` needs a string.
3. Refactor everything to just accept `*IntlMessage` for instance `titleIntlMessage`. This however is a breaking change to the public component.
4. Offer two sets of props: a bit too much work for a little thing like this

### DIFF
--- a/packages/application-shell/src/components/project-suspended/project-suspended.tsx
+++ b/packages/application-shell/src/components/project-suspended/project-suspended.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { FormattedMessage } from 'react-intl';
+import { useIntl, FormattedMessage } from 'react-intl';
 import ProjectSuspendedSVG from '@commercetools-frontend/assets/images/doors-closed.svg';
 import { MaintenancePageLayout } from '@commercetools-frontend/application-components';
 import ServicePageProjectSwitcher from '../service-page-project-switcher';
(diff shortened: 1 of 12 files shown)


packages/application-shell/src/components/project-suspended/project-suspended.tsx
```diff
@@ -9,22 +9,22 @@ type Props = {
...
-      if (props.isTemporary === true)
-        return (
-          <FormattedMessage
-            {...messages.temporaryMaintenanceSuspensionMessage}
-          />
-        );
-      return <FormattedMessage {...messages.defaultSuspensionMessage} />;
-    })()}
-    label={intl.formatMessage(messages.title)}
+    title={<FormattedMessage {...titleMessage} />}
```
ahmehri, line 26: Why not?

```suggestion
    title={intl.formatMessage(titleMessage)}
```

tdeekens: Could work. Didn't want to change too much. Let me try.

tdeekens: Won't work as the message goes to `Text` which needs a React.Node as children.

PR #1600, opened 2020-06-24
feat(app-shell): to allow passing `featureFlag` as variable

tdeekens:
> This is up for discussion and a proposal.

#### Summary

The MC API will at times accept a `X-Feature-Toggle` header. This header can be set through the context of Apollo.

#### Description

The header could also be set using the variables similar as the target variable which then encapsulates the header setting.

The question is if we want to expose this through our SDK or not and how much we will use it in the end.


packages/application-shell/src/apollo-links/header-link.spec.js
```diff
@@ -51,13 +51,27 @@ describe('with valid target', () => {
...
+          target: GRAPHQL_TARGETS.MERCHANT_CENTER_BACKEND,
+          featureFlag: 'test-feature-a',
+        },
       })
     );
   });
 
   it('should set headers matching snapshot', () => {
-    expect(context).toMatchSnapshot();
+    expect(context).toMatchInlineSnapshot(`
```
tdeekens, line 63: I find inline snapshots convienent for this when not too long to show what's expected.

nbryant-commercetools:
@tdeekens ,

Thanks for this proposal.  I think this works for what my team is wanting to achieve.  I could see this being used in the future as other features are in development or being refactored as well.

cc: @torihedden / @wrsenn / @adilasif

tdeekens:
So either we do this: `variable.featureToggle` to `headers['X-Feature-Toggle']` or we add `{ ...operation.getContext().headers, ...ourHeaders }`. I kind of like explicitness so I like what we're doing here. However it's also odd to pass anything through variables and dump it onto the header.

nbryant-commercetools:
> So either we do this: `variable.featureToggle` to `headers['X-Feature-Toggle']` or we add `{ ...operation.getContext().headers, ...ourHeaders }`. I kind of like explicitness so I like what we're doing here. However it's also odd to pass anything through variables and dump it onto the header.

@tdeekens ,

I agree about the explicitness of this implementation.  This is clearer about what exactly we are hoping to pass along, and it follows the same pattern as the setup for the x-graphql-target header.

PR #1614, opened 2020-06-30
refactor: rename oppressive terms ✊🏾

emmenko:


website/src/content/components/application-shell.mdx
```diff
@@ -42,7 +42,7 @@ const EntryPoint = () => (
...
-| `trackingEventWhitelist`         | `object`           |                       | -       | An optional object containing a map of tracking events.                                                                                                                                                                                                                                        |
+| `trackingEventList`         | `object`           |                       | -       | An optional object containing a map of tracking events.                                                                                                                                                                                                                                        |
```
adnasa, line 45: instead of removing it entirely, I would:

1. keep it and mark it deprecated. perhaps announce a date of when we will remove it completely from the code base e.g this https://github.com/commercetools/merchant-center-application-kit/pull/1614/files#diff-296eae346ba59229997210b8ff8ab651R557
2. document to use the new name instead.

Then we can eventually remove the old name

emmenko: I thought about that but we only use this feature internally, so I wouldn't bother document it properly.

adnasa: aahaa, I was under the impression otherwise.
so help me understand the strategy here:
are we going to refactor this off from MC-FE then remove the old naming entirely in the next step? #guessWork

tdeekens: Yes, I assumed so too.

emmenko: Yep, more or less

adnasa: all right. thanks

Issue #1623, opened 2020-07-07
connect to external graphql API


jamesgongqantas:
I have read the documentation from the following:

https://docs.commercetools.com/custom-applications/main-concepts/data-fetching#connecting-to-an-external-graphql-api

however when trying the code snippet from the above link, I have the following error message returned in response

`Missing required header "X-Forward-To".`

the request sends to `https://mc-api.australia-southeast1.gcp.commercetools.com/proxy/forward-to` which is expected I think, but the example in documentation doesn't specify how to add the missing header, or if I did something wrong?

Also what if I need to add some extra header? Do I have to create my own Apollo client for doing this?

tdeekens:
1. X-Foward-To header missing

The `createApolloContextForProxyForwardTo` maps the `uri` onto the `x-forward-to` header [here](https://github.com/commercetools/merchant-center-application-kit/blob/fd818e3a131ce71f02b738b5c9ff93978ae526fe/packages/application-shell/src/utils/apollo-context.ts#L19).

Can you check if the request which is sent to `/forward-to` contains the `x-foward-to` header?

2. Extra headers

Extra headers are not supported when proxying through the MC API itself. We have a list of allowed headers. What kind of headers would you want to forward and why?

jamesgongqantas:
> 1. X-Foward-To header missing
> 
> The `createApolloContextForProxyForwardTo` maps the `uri` onto the `x-forward-to` header [here](https://github.com/commercetools/merchant-center-application-kit/blob/fd818e3a131ce71f02b738b5c9ff93978ae526fe/packages/application-shell/src/utils/apollo-context.ts#L19).
> 
> Can you check if the request which is sent to `/forward-to` contains the `x-foward-to` header?
> 
> 1. Extra headers
> 
> Extra headers are not supported when proxying through the MC API itself. We have a list of allowed headers. What kind of headers would you want to forward and why?

1. the request doesn't have the `x-forward-to` header
2. as the external graphql api expect that header, which is `accesstoken` in my case

emmenko:
Maybe let's take a step back.

You don't need to use the forward-to endpoint to connect to your API. You can talk to it directly if you want, and take care of the auth stuff. If you do that, you also have to adjust your CSP config to allow the API domain.

The forward-to endpoint can be used to secure your API with low effort, as most of the auth validation is done by the MC API for you.

---

As for why the request doesn't have the headers, I can look into that and try to reproduce it.

[...]

PR #1624, opened 2020-07-07
 fix(app-shell): set headers for forward-to feature when using the Apollo client

emmenko:
Ref #1623

### DIFF
--- a/packages/application-shell/src/apollo-links/header-link.ts
+++ b/packages/application-shell/src/apollo-links/header-link.ts
@@ -37,6 +37,7 @@ const headerLink = new ApolloLink((operation, forward) => {
   const {
     skipGraphQlTargetCheck = false,
     cache,
+    headers: additionalHeaders = {},
   } = operation.getContext() as ApolloContextWithInMemoryCache;
 
   const variables = operation.variables as QueryVariables;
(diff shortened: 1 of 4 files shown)


packages/application-shell/src/apollo-links/header-link.ts
```diff
@@ -70,6 +71,8 @@ const headerLink = new ApolloLink((operation, forward) => {
       'X-Team-Id': teamId,
       'X-Application-Id': window.app.applicationId,
       'X-Feature-Flag': featureFlag,
+      // Merge additional headers that may have been set in the `context.headers` option.
+      ...additionalHeaders,
```
tdeekens, line 75: Can we explicitely add the `x-forward-to`? I would avoid spreading as it gives the impression that the MC API supports all headers which it doesn't. 

I understand that we want to support any header to other services (`uri`s) but that's separate to me. We could either 

1. Optimistically spread them
2. Add a `allowAnyHEader` property
3. Evaluate the `uri` to not be one of ours

Just a thought but I think spreading them (as in 1.) is a bit dangerous.

emmenko: Hmm it‘s not just one header.

Let me think about it, thanks for raising this up

Issue #1440, opened 2020-04-15
[RFC] Unified and less error-prone custom apps configuration


emmenko:
> This topic is meant to drive a decision around improving the configuration of Custom Applications

## Summary

Custom Applications are configured by an `env.json` and `headers.json` files.
The `env.json` expects to provide some values like `mcApiUrl`, while the `headers.json` usually should include the Custom Application hostname and the MC API URL.

It appears that often times this is a source of confusion and error-prone, for example to mistype the URL or to mix the cloud region of the MC.

## Discussion

To improve this and make it less error-prone, we could maybe have a unified configuration file and try to require only the minimal configuration necessary.

For example, we could also leverage fields like `cloudRegion` and `cloudProvider` instead of requiring the full URL, so that it's more explicit in which environments the app should run on.

We could also require a `version`, that we use to determine if the new format should be used or not. A JSON schema could also come in handy, also to make it available in editors like VSCode that can already provide feedback (see example below)

<img width="696" alt="image" src="https://user-images.githubusercontent.com/1110551/79316833-35a82f00-7f05-11ea-96fa-ed8ab0204fe4.png">

## Decision

To come up with a new config format that is easier to understand and less error-prone.

## Measure Completion

TBD

emmenko:
I drafted a bit a possible configuration file.

```json
{
  "name": "Starter app",
  "description": "",
  "url": "https://my-app.com",
  // "uriPath": "starter",
  // "uriEntryPoint": "starter",
  "entryPointUriPath": "starter",
  "navbarMenu": {
    // ...
  },
  "environment": {
    "servedByProxy": false,
    "cloudIdentifier": "gcp-eu",
    "cdnUrl": "",
  },
  "additionalEnvironment": {
    // ...
  },
  "headers": {
    "csp": {
      // ...
    },
  }
}
```

The idea is to provide one single configuration file that we can use for multiple things. For instance, we can drop many of the current configuration fields and try to infer all the necessary information as much as possible.

Some examples are:
- based on the `url`, where the app is hosted, we can infer the hostname, used in the csp directives. We can also use it for the `cdnUrl`, if not explicitly defined, meaning that the assets are hosted on the same domain.
- based on the `cloudIdentifier` we can determine the MC API URL, as well as the `connect-src` whitelist for the CSP directives.
- the `entryPointUriPath` (or however we want to name that) should explicitly hint at the main route path used to access the application.

[...]

PR #1626, opened 2020-07-09
feat: add new package for managing application config

emmenko:
This is the first step towards #1440 

The goal is to have a unified config that makes the setup of a Custom Application simpler and less error prone.

This PR implements the JSON schema, as I felt it would make it easier first to understand how the new JSON format works.

The idea with the package is then to provide some APIs to load the configuration file, validate it and parse the values. This will be then used by other packages, like the `mc-html-template`.

image


emmenko:
> This contains some breaking changes, right?

Not really. That's also why I opted of creating a new package, to better keep things separate. I also think that we should be able to be backwards compatible with the current config `env.json` and `headers.json`, where we transform the data into the new format.

tdeekens:
> where we transform the data into the new format.

Yeah, I had the same thought. That we can have transformations offered by the package. So we keep the old format working. The JSON Schema might be slightly different but we can transform into it easily.

I think we talked about it. How do you feel about adding a `version` to the `env.json` and `headers.json`. Most platforms have that (e.g. CircleCI). This would allow us to more easily know what state a file is in and how to transform it into another format/version.

emmenko:
I don't think the version is necessary.

I plan to name the new config file `applicationconfig.json`, which makes it also clear that it's a "new thing"

tdeekens:
Yes, I understand that it will be named differently. Still, the `applicationconfig.json` might change in shape once renamed. People could also opt-into new features by specifying a verison. Anyway, I understand if it's not needed yet.

emmenko:
Ah sure, I thought you meant for the "migration phase". We can specify a version in the schema if you want, or introduce it once we need it, so we would start by e.g. `v2`

packages/application-config/README.md
```diff
@@ -0,0 +1,35 @@
...
+In the VSCode settings (either user settings or workspace settings), reference the schema JSON as described below:
+
+```json
+"json.schemas": [
+  {
+    "fileMatch": [
+      "/.applicationconfig.json",
+      "/applicationconfig.json"
+    ],
+    "url": "https://unpkg.com/@commercetools-frontend/application-config/schema.json"
```
emmenko, line 32: @tdeekens do you think it makes sense to reference/use this URL?

tdeekens: Yeah. Should be fine I think.

packages/application-config/package.json
```diff
@@ -0,0 +1,85 @@
...
+    "schema.json",
+    "package.json",
+    "LICENSE",
+    "README.md"
+  ],
+  "scripts": {
+    "prepare": "./../../scripts/version.js replace",
+    "prebuild": "rimraf build/**",
+    "build": "yarn build:schema && yarn build:src && yarn build:typings",
+    "build:schema": "json2ts schema.json src/schema.ts --style.singleQuote --bannerComment '/* eslint-disable prettier/prettier */\n// This file was automatically generated by json-schema-to-typescript.\n// DO NOT MODIFY IT BY HAND. Instead, modify the source schema.json file.'",
```
emmenko, line 39: This generates the TS types for the JSON schema

packages/application-config/schema.json
```diff
@@ -0,0 +1,133 @@
...
+    "mcApiUrl": {
+      "description": "The URL of the Merchant Center API. We usually recommend to use the `cloudIdentifier` option to avoid possible typos.",
+      "type": "string"
+    },
+    "env": {
+      "description": "This is an object of keys that represent different environments (for example `production`). The environment used depends on the environment variable MC_APP_ENV. If MC_APP_ENV isn't set then NODE_ENV will be used, if it's not set then it defaults to \"development\"",
+      "type": "object",
+      "properties": {
+        "production": {
+          "description": "Configuration for production only",
```
emmenko, line 59: This `env.production` is inspired by the babel config.

packages/application-config/src/substitute-env-variable-placeholders.ts
```diff
@@ -0,0 +1,69 @@
...
+  );
+};
+
+const getValueOfPlaceholder = (valueWithPlaceholder: string) =>
+  valueWithPlaceholder
+    .replace(variableSyntax, (_match, varName) => varName.trim())
+    .replace(/\s/g, '');
+
+const substituteEnvVariablePlaceholders = <T>(config: T): T =>
+  JSON.parse(JSON.stringify(config), (_key, value) => {
```
emmenko, line 48: FYI: Using JSON.parse and the callback function simplifies the nested recursion logic.

packages/application-config/test/fixtures/config-simple.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "avengers-app",
+  "entryPointUriPath": "avengers",
+  "cloudIdentifier": "gcp-eu",
+  "env": {
+    "production": {
+      "url": "https://avengers.app"
+    }
+  }
+}
```
emmenko, line 10: This is the simplest version that a Custom App can have. As you can see, most of the final values are inherited from this.

emmenko:
Short update: after experimenting with different combinations of the JSON schema, I'm finally settling in the current implemented version. See the `test/fixtures` examples to have a better mental picture.

Processing the config yields all the necessary data where most of it is inherited from e.g. the URLs or the `process.env`.

The next step would be to implement the backwards compatible logic for reading the config from `env.json` and `headers.json`. We can do that in a separate PR.
After that, we can migrate the other packages to use the new config package, while maintaining the backwards compatibility.
Then we test things in the MC and if all goes well, we update the documentation to recommend using the new config file.

tdeekens:
Would it be okay for you to add some inline comments to the PR before putting it to review. I glimpsed over it but it are many moving parts and it has grown quite a bit. Just so others are able to follow along.

emmenko:
Sure, I still need to do some testing but I'll document things as much as I can before putting it in review.

packages/application-config/README.md
```diff
@@ -0,0 +1,52 @@
...
+      "/.applicationconfigrc",
+      "/.applicationconfig.json",
+      "/applicationconfig.json"
+    ],
+    "url": "https://unpkg.com/@commercetools-frontend/application-config/schema.json"
+  }
+]
+```
+
+## API
```
emmenko, line 52: I'll document this as a follow up, as I don't know yet what we need to expose for the package to be used by the other packages in a backwards compatible way.

packages/application-config/package.json
```diff
@@ -0,0 +1,85 @@
...
+    "build:src": "babel src --out-dir build --extensions .ts --ignore src/types.ts,src/schema.ts",
```
emmenko, line 40: Note: we're transpiling with Babel here as we want to retain the single files, as opposed to Rollup where we get one single bundle file.

packages/application-config/schema.json
```diff
@@ -0,0 +1,133 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
```
emmenko, line 2: If you're unfamiliar with JSON schema, you can have a look at their documentation: https://json-schema.org/learn/getting-started-step-by-step.html

packages/application-config/schema.json
```diff
@@ -0,0 +1,133 @@
...
+  "$id": "https://unpkg.com/@commercetools-frontend/application-config/schema.json",
+  "title": "JSON schema for Custom Application configuration files",
+  "type": "object",
+  "definitions": {
```
emmenko, line 6: The definitions here are "reusable" schema properties that you can reference within the schema.

packages/application-config/schema.json
```diff
@@ -0,0 +1,133 @@
...
+      "type": "string"
+    },
+    "name": {
+      "description": "The name of the Custom Application",
+      "type": "string"
+    },
+    "entryPointUriPath": {
+      "description": "The unique route path of the Custom Application. This is the identifier that the Merchant Center Proxy uses to match the HTTP request and forward the request to the Custom Application. This value also needs to be used in the application client side routes. The value must be the same as the `navbarMenu.uriPath` value.",
+      "type": "string"
+    },
```
emmenko, line 31: The `entryPointUriPath` is a "special" property that intends to clear up possible confusions.
The property itself is not used in the config, it just serves an informative purpose and a reminder to the user that this is the entry point of the Custom Application.

Effectively, this value should be the same as the `navbarMenu.uriPath`, which is what the MC proxy uses to match the incoming request.

packages/application-config/schema.json
```diff
@@ -0,0 +1,133 @@
...
+            "url": {
+              "description": "The URL where the Custom Application is hosted",
+              "type": "string"
+            },
+            "cdnUrl": {
+              "description": "The URL where the Custom Application static assets are hosted, like an external CDN. If the static assets are hosted alongside the Custom Application, you can omit this option and the Custom Application URL will be used instead.",
+              "type": "string"
+            }
+          },
+          "additionalProperties": false,
```
emmenko, line 71: When you see this value within the schema, it means that the related schema object does not accept any extra properties that are not defined in the schema.

packages/application-config/src/load-config.ts
```diff
@@ -0,0 +1,32 @@
...
+import { cosmiconfigSync, defaultLoaders } from 'cosmiconfig';
+
+const moduleName = 'applicationconfig';
+const explorer = cosmiconfigSync(moduleName, {
+  // Restrict the supported file formats / names
+  searchPlaces: [
+    `.${moduleName}rc`,
+    `.${moduleName}.json`,
+    `${moduleName}.json`,
+  ],
```
emmenko, line 12: Here we say that we only support the files with those names.

packages/application-config/src/load-config.ts
```diff
@@ -0,0 +1,32 @@
...
+  loaders: {
+    noExt: defaultLoaders['.json'],
```
emmenko, line 14: This basically tells the library to load the `.applicationconfigrc` as JSON.

packages/application-config/src/load-config.ts
emmenko: It's all JSON so that we can always validate it with our JSON schema

packages/application-config/src/process-config.ts
```diff
@@ -0,0 +1,108 @@
...
+} from './utils';
+import substituteEnvVariablePlaceholders from './substitute-env-variable-placeholders';
+
+type ProcessConfigOptions = {
+  disableCache?: boolean;
+  processEnv?: NodeJS.ProcessEnv;
+};
+
+const developmentConfig: JSONSchemaForCustomApplicationConfigurationFiles['env']['production'] = {
+  url: 'http://localhost:3001',
```
emmenko, line 22: This assumes that the development URL of the app is always `http://localhost:3001`.

I can't think of a case where this value can be different, and therefore it's pointless to ask the user to provide this value.

If this needs to be configurable, we can refactor the implementation a bit to support that.

packages/application-config/src/process-config.ts
```diff
@@ -0,0 +1,108 @@
...
+
+const processConfig = async ({
+  // Options useful for testing
+  disableCache = false,
+  processEnv = process.env,
+}: ProcessConfigOptions = {}): Promise<ApplicationConfig> => {
+  if (cachedConfig && !disableCache) return cachedConfig;
+
+  const appEnvKey =
+    processEnv.MC_APP_ENV ?? processEnv.NODE_ENV ?? 'development';
```
emmenko, line 37: We always read first from `MC_APP_ENV` over `NODE_ENV`

packages/application-config/src/process-config.ts
```diff
@@ -0,0 +1,108 @@
...
+  const isProd = getIsProd(processEnv);
+
+  // TODO: handle legacy configs for backwards compatibility
```
emmenko, line 40: This will be done in a follow up PR

packages/application-config/src/process-config.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  const rawAppConfig = loadConfig();
+  await validateConfig(rawAppConfig);
```
emmenko, line 43: Here we load and validate the config.

packages/application-config/src/process-config.ts
```diff
@@ -0,0 +1,108 @@
...
+  const validatedRawAppConfig = rawAppConfig as JSONSchemaForCustomApplicationConfigurationFiles;
+
+  const appConfig = substituteEnvVariablePlaceholders<
+    JSONSchemaForCustomApplicationConfigurationFiles
+  >(validatedRawAppConfig, { processEnv });
```
emmenko, line 48: The application config also supports env variable placeholders, so we need to substitute them first.

packages/application-config/src/process-config.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  const appEnvConfig = isProd ? appConfig.env.production : developmentConfig;
```
emmenko, line 50: If we're in "production" mode, we read the config from `env.production` (for example like Babel does).

packages/application-config/src/process-config.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  // Parse all the supported URLs, which gets implicitly validated
+  const appUrl = getOrThrow(
+    () => new URL(appEnvConfig.url),
+    `Invalid application URL: "${appEnvConfig.url}"`
+  );
+  const cdnUrl = getOrThrow(
+    () => new URL(appEnvConfig.cdnUrl || appUrl.origin),
```
emmenko, line 58: The CDN url, if not explicitly specified, falls back to be the same as the application URL.

packages/application-config/src/process-config.ts
```diff
@@ -0,0 +1,108 @@
...
+    `Invalid application CDN URL: "${appEnvConfig.cdnUrl || appUrl.origin}"`
+  );
+  const mcApiUrl = getOrThrow(
+    () =>
+      new URL(
+        appConfig.mcApiUrl ??
```
emmenko, line 64: If the `mcApiUrl` is specified, we use that value, otherwise we infer the value from the cloud identifier. This is useful for internal purposes (e.g. for our staging env).

packages/application-config/src/process-config.ts
```diff
@@ -0,0 +1,108 @@
...
+      applicationId: appConfig.id,
+      applicationName: appConfig.name,
+      cdnUrl: cdnUrl.origin,
+      env: appEnvKey,
+      frontendHost: appUrl.host,
+      location: appConfig.cloudIdentifier,
+      mcApiUrl: mcApiUrl.origin,
+      revision: '', // TODO
+      servedByProxy: isProd,
+    },
```
emmenko, line 84: This is where we map the values and the inferred values to the format used by the AppShell.
Note that most of the values are inferred.

packages/application-config/src/process-config.ts
```diff
@@ -0,0 +1,108 @@
...
+          mapCloudIdentifierToApiUrl(
+            appConfig.cloudIdentifier as CloudIdentifier
+          )
+      ),
+    `Invalid MC API URL: "${appConfig.mcApiUrl}"`
+  );
+
+  cachedConfig = {
+    env: {
+      ...(appConfig.additionalEnv || {}),
```
emmenko, line 74: We spread the additional properties first, to ensure that the required values below are not being overwritten.

packages/application-config/src/process-config.ts
```diff
@@ -0,0 +1,108 @@
...
+          [mcApiUrl.hostname].concat(isProd ? appUrl.hostname : [])
+        ),
+        'script-src': getUniqueValues(
+          appConfig.headers?.csp['script-src'],
+          isProd ? [appUrl.hostname] : []
+        ),
+        'style-src': getUniqueValues(
+          appConfig.headers?.csp['style-src'],
+          isProd ? [appUrl.hostname] : []
+        ),
```
emmenko, line 100: Note that the required header values are all inferred from the application URL and MC API URL.

packages/application-config/src/process-config.ts
```diff
@@ -0,0 +1,109 @@
...
+          mapCloudIdentifierToApiUrl(
+            appConfig.cloudIdentifier as CloudIdentifier
+          )
+      ),
+    `Invalid MC API URL: "${appConfig.mcApiUrl}"`
+  );
+
+  cachedConfig = {
+    env: {
+      ...additionalAppEnv,
```
emmenko, line 75: We spread the additional properties first, to ensure that the required values below are not being overwritten.

packages/application-config/src/process-config.ts
```diff
@@ -0,0 +1,109 @@
...
+      applicationId: appConfig.id,
+      applicationName: appConfig.name,
+      cdnUrl: cdnUrl.origin,
+      env: appEnvKey,
+      frontendHost: appUrl.host,
+      location: appConfig.cloudIdentifier,
+      mcApiUrl: mcApiUrl.origin,
+      revision,
+      servedByProxy: isProd,
+    },
```
emmenko, line 85: This is where we map the values and the inferred values to the format used by the AppShell.
Note that most of the values are inferred.

packages/application-config/src/substitute-env-variable-placeholders.ts
```diff
@@ -0,0 +1,74 @@
+/**
```
emmenko, line 1: This file is the "same" as the one we currently have in the `mc-html-template` package

packages/application-config/src/load-config.ts
```diff
@@ -0,0 +1,32 @@
+import type { JSONSchemaForCustomApplicationConfigurationFiles } from './schema';
+
+import { cosmiconfigSync, defaultLoaders } from 'cosmiconfig';
+
+const moduleName = 'applicationconfig';
```
tdeekens, line 5: With the naming of this file. I wonder if `applicationconfig` is a bit too "wide". Maybe `merchantCenterCustomApplicationConfig`? Dang it that's long or `customApplicationConfig? Not 100% sure either.

packages/application-config/src/process-config.ts
tdeekens: I agree.

packages/application-config/src/process-config.ts
```diff
@@ -0,0 +1,109 @@
...
+}: ProcessConfigOptions = {}): Promise<ApplicationConfig> => {
+  if (cachedConfig && !disableCache) return cachedConfig;
+
+  const appEnvKey =
+    processEnv.MC_APP_ENV ?? processEnv.NODE_ENV ?? 'development';
+  const isProd = getIsProd(processEnv);
+
+  // TODO: handle legacy configs for backwards compatibility
+
+  const rawAppConfig = loadConfig();
```
tdeekens, line 42:
```suggestion
  const loadedAppConfig = loadConfig();
```

🤷

packages/application-config/src/substitute-env-variable-placeholders.ts
tdeekens: So we should keep them in sync, remove one or merge them from a utils package?

packages/application-config/test/json-schema.spec.js
```diff
@@ -0,0 +1,68 @@
...
+});
+
+describe('invalid configurations', () => {
+  it('should validate that "env" is defined', () =>
+    expect(
+      validateConfig({
+        ...fixtureConfigSimple,
+        env: undefined,
+      })
+    ).rejects.toEqual(
```
tdeekens, line 24: Maybe `toMatchInlineSnapshot` is also nice here. Just wondering. I think they're a bit underused at times :)

jest.text.config.js
```diff
@@ -2,6 +2,7 @@ module.exports = {
   runner: 'jest-runner-executor',
   displayName: 'vale',
   moduleFileExtensions: ['md', 'mdx'],
+  modulePathIgnorePatterns: ['build', 'dist', 'public'],
```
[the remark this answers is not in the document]
emmenko: It's related to the new package because the transpiled files are put into the `build` folder.

We do the same for other packages that run code in a node environment. I think it's helpful to differentiate it from the `dist` package, where we bundle the code into one file.

lint-staged.config.js
```diff
@@ -26,4 +26,7 @@ module.exports = {
     // For that reason, we move the `--onlyChanged` flag next to it.
     'yarn lint:css --reporters=jest-silent-reporter --onlyChanged',
   ],
+  'packages/application-config/schema.json': [
+    'bash -c "yarn --cwd packages/application-config build:schema"',
```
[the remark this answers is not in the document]
emmenko: Because otherwise the command will receive the file path from lint-staged. In this case, we just need to run a command, independently from the file.

It's the recommended approach from lint-staged if I remember.

packages/application-config/package.json
```diff
@@ -0,0 +1,85 @@
...
+    "node": ">=10",
+    "npm": ">=5"
+  },
+  "babel": {
+    "presets": [
+      [
+        "@babel/preset-env",
+        {
+          "targets": {
+            "node": "10"
```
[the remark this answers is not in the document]
emmenko: Yeah, for those env still running on node 10 (e.g. google cloud functions)

packages/application-config/package.json
```diff
@@ -0,0 +1,85 @@
...
+    "@babel/plugin-transform-runtime": "7.10.4",
+    "@babel/preset-env": "7.10.4",
+    "@babel/preset-typescript": "7.10.4",
+    "json-schema-to-typescript": "9.1.1"
+  },
+  "engines": {
+    "node": ">=10",
+    "npm": ">=5"
+  },
+  "babel": {
```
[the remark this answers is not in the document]
emmenko: Mostly to avoid confusing tools, as this config is only used by this package for development. We do the same in other packages (mc-html-template, mc-scripts)

packages/application-config/src/load-config.ts
emmenko: Yeah I though about that but the full name felt too long. Maybe `customapplicationconfig` if you prefer so.

packages/application-config/src/substitute-env-variable-placeholders.ts
emmenko: I'll migrate things in a follow up PR. The one in the mc-html-template package will then go away

packages/application-config/src/load-config.ts
tdeekens: Yeah. I would think having some "branding" can help to users not confuse things. E.g. `customapplicationconfig` or `custom-application-config`?

emmenko: Renamed to `custom-application-config` 👌

tdeekens:
From my side this looks good for now. Do you have any follow-up plans @emmenko?

packages/application-config/schema.json
```diff
@@ -0,0 +1,133 @@
...
+      ]
+    }
+  },
+  "additionalProperties": false,
+  "required": [
+    "name",
+    "entryPointUriPath",
+    "cloudIdentifier",
+    "env"
+  ]
```
mohib0306, line 132: I see that `id` is not required, is it on purpose. I think every custom app will have a `ID`

emmenko:
I have a branch with the migration changes (based on this branch). I'll open a new PR later.

packages/application-config/schema.json
emmenko: Good observation. The ID is not required (yet), it will be in future releases and when we (soon) implement a new workflow to register custom apps.

emmenko:
Alright, thanks for the review first of all!

I have opened another PR (#1633) that shows and migrates the codebase to use the new application config.

PS: should we review and merge the other PR into this one, and merge to master all of it together? Or should we keep the PRs and merges separated?

PR #1631, opened 2020-07-13
feat(babel-preset-mc-app): add logical assignment operator

tdeekens:
#### Summary

This pull request suggests to add the logical assignment operator to our babel config.

#### Description

The logical assignment operator [TC39 proporal](https://github.com/tc39/proposal-logical-assignment) can help us when assignment optional values based on a condition. 

A simple use case is:

```js
a ||= b;
obj.a.b ||= c;

a &&= b;
obj.a.b &&= c;
```

More real world would be

```js
const books = [
  {
    isbn: '123',
  },
  {
    title: 'ECMAScript Language Specification',
    isbn: '456',
  },
];

// Add property .title where it’s missing
for (const book of books) {
  book.title ??= '(Untitled)';
}
```

I think in general we have cases internally where we assign values if they're not assigned yet through ternaries. These case could be simplified through this.

1. https://2ality.com/2020/06/logical-assignment-operators.html
2. https://github.com/tc39/proposal-logical-assignment
3. https://babeljs.io/docs/en/next/babel-plugin-proposal-logical-assignment-operators.html

### DIFF
--- a/.changeset/chilled-kiwis-pretend.md
+++ b/.changeset/chilled-kiwis-pretend.md
@@ -1,0 +1,5 @@
+---
+"@commercetools-frontend/babel-preset-mc-app": minor
+---
+
+feat(babel-preset-mc-app): add logical assignment operator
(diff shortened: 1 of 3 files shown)


emmenko:
Snapshots approved (I added a new test project today, hence the project count of 2)

PR #1633, opened 2020-07-15
refactor: migrate to new application config

emmenko:
This PR is based off #1626 and migrates the appkit codebase to use the new application config.


packages/application-config/src/index.ts
```diff
@@ -1,2 +1 @@
 export { default as processConfig } from './process-config';
-export { default as validateConfig } from './validate-config';
```
emmenko, line 2: Not necessary to expose it

packages/application-config/src/load-deprecated-config.ts
```diff
@@ -0,0 +1,67 @@
...
+    });
+  } catch (error) {
+    // Ignore
+  }
+  return rawData ? JSON.parse(rawData) : {};
+};
+
+// For backwards compatibility, we still support the `env.json` and `headers.json` files.
+// TODO: remove in `v17`.
+const loadDeprecatedConfig = (
```
emmenko, line 36: This is the part where we load the `env.json` and `headers.json` for backwards compatibility

packages/application-config/src/process-config.ts
```diff
@@ -37,9 +42,21 @@ const processConfig = ({
...
+    console.warn(
+      `No custom application config found, attempting to load the config from env.json and headers.json.`
+    );
+    const loadedDeprecatedAppConfig = loadDeprecatedConfig(deprecatedOptions);
+    if (!loadedDeprecatedAppConfig) {
+      throw new Error(`No configuration for the Custom Application found.`);
+    }
+    // Return the legacy config as-is, no need to transform it.
+    return loadedDeprecatedAppConfig;
+  }
```
emmenko, line 58: This is the logic where we try to load the deprecated files, in case the new config file was not found.

If you have other suggestions for implementing this logic, please let me know.

packages/application-shell/src/components/redirect-to-project-create/redirect-to-project-create.tsx
```diff
@@ -27,19 +29,32 @@ export const RedirectToProjectCreate = () => {
   }
 
   return (
-    <Spacings.Stack>
-      <Text.Headline as="h3">Please create a project!</Text.Headline>
-      <Text.Body>
-        You are using the Merchant Center in development mode - it is not served
-        by a proxy (`env.json`). Moreover, you do not have any projects yet. As
```
emmenko, line 34: Since this message was referring to the `env.json`, I took the liberty to improve the UI a bit, as well as to adjust the information message.

Note that this message only appears in development, when the user has no projects.

packages/mc-html-template/README.md
```diff
@@ -22,49 +22,28 @@ This method will return the compiled HTML document with the CSS/JS scripts injec
...
 - `__LOADING_SCREEN_CSS__`: (_defined internally_) the CSS for the loading animation in case the page takes longer to load
 - `__LOADING_SCREEN_JS__`: (_defined internally_) the JS for the loading animation in case the page takes longer to load
-- `__APP_ENVIRONMENT__`: the sanitized `env.json`, which will be available at the global variable `window.app`
-- `__DATALAYER_JS__`: the initial configuration for GTM, in case the `trackingGtm` is defined in the `env.json`
-- `__GTM_SCRIPT__`: the actual GTM script, in case the `trackingGtm` is defined in the `env.json`
+- `__APP_ENVIRONMENT__`: the sanitized application config environment, which will be available at the global variable `window.app`
+- `__DATALAYER_JS__`: the initial configuration for GTM, in case the `trackingGtm` is defined in the `additionalEnv` property of the application config
+- `__GTM_SCRIPT__`: the actual GTM script, in case the `trackingGtm` is defined in the `additionalEnv` property of the application config
 
-#### `loadEnv(configPath: String): Object`
```
emmenko, line 39: The `loadEnv` export has been removed. I don't think we need to consider it a breaking change, as it's like an internal API.

packages/mc-html-template/README.md
```diff
@@ -22,49 +22,28 @@ This method will return the compiled HTML document with the CSS/JS scripts injec
...
-
-This method will attempt to load and parse the `env.json` file, performing some validation and returning the parsed JSON.
-
-More information about required values and references can be found in the [Runtime Configuration
-](https://docs.commercetools.com/custom-applications/deployment/runtime-configuration) documentation of Custom Applications.
-
-#### `loadHeaders(env: Object, { headersPath: String, cspPath?: String }): Object`
```
emmenko, line 46: Same here, the `loadHeaders` export has been removed. I don't think we need to consider it a breaking change, as it's like an internal API.

packages/mc-html-template/README.md
```diff
@@ -22,49 +22,28 @@ This method will return the compiled HTML document with the CSS/JS scripts injec
...
+#### `processHeaders(applicationConfig: Object, { env: Object, headers: Object }): Object`
```
emmenko, line 39: The `processHeaders` is the new export, which is basically the same as the `loadHeaders`, only with different arguments.

packages/mc-html-template/README.md
tdeekens: It is becomes we can re-add it silently if somebody asks for it.

packages/application-config/src/load-deprecated-config.ts
```diff
@@ -0,0 +1,70 @@
...
+import substituteEnvVariablePlaceholders from './substitute-env-variable-placeholders';
+import { parseJsonFile } from './utils';
+
+export type DeprecatedOptions = {
+  envPath: string;
+  headersPath: string;
+  cspPath?: string;
+};
+
+// List the required fields of `env.json`
```
mohib0306, line 14: `env.json` should be created by the users by themselves and no longer part of the repo. Right?

packages/application-config/test/process-config.spec.js
```diff
@@ -402,3 +405,71 @@ describe('when MC API URL is malformed', () => {
...
+      headers: {
+        csp: {
+          'connect-src': ['mc-api.europe-west1.gcp.commercetools.com'],
+          'script-src': [],
+          'style-src': [],
+        },
+      },
+    });
+    expect(console.warn).toHaveBeenCalledWith(
+      `No custom application config found, attempting to load the config from env.json and headers.json.`
```
mohib0306, line 456: can be wrapped in single quotes

packages/application-config/src/load-deprecated-config.ts
emmenko: I'm not 100% sure what you mean here but in general we want to start using the new config file (as we do now). I also need to update the documentation and write a release note (separate PR).
People can still use the `env.json` file, as long as it's supported. For example, in the next major version v17 we can drop support for that.

Does this answer your question?

mohib0306: yes. That answers the question

PR #1634, opened 2020-07-16
docs: add release notes and new doc page about new application config

### LINKED ISSUE #1440: [RFC] Unified and less error-prone custom apps configuration

emmenko:
This is the final step to close #1440


website/src/releases/2020-07/application-config.mdx
```diff
@@ -0,0 +1,168 @@
...
+
+This release introduces the usage of a new configuration file format and marks the deprecation of the `env.json` and `headers.json` files.
+
+<Info>
+
+The `env.json` and `headers.json` files will still keep working but they will be removed in the next major release.
+
+</Info>
+
+The new configuration format aims to drastically simplify how to configure the Custom Application, as well as to make the configuration process less error prone.
```
tdeekens, line 20:
```suggestion
The new configuration format aims to drastically simplify the configuration of a Custom Application. In addition, it also strives to make the configuration process less error prone.
```

website/src/releases/2020-07/application-config.mdx
```diff
@@ -0,0 +1,168 @@
...
+In fact, the new configuration file is backed by a JSON schema that is shipped together with the [new package](https://www.npmjs.com/package/@commercetools-frontend/application-config). The configuration file is then validated against the JSON schema.
```
tdeekens, line 21:
```suggestion
To achieve this, the new configuration file is backed by a JSON schema that is shipped together with the [new package](https://www.npmjs.com/package/@commercetools-frontend/application-config). The configuration file is then validated against the JSON schema.
```

website/src/releases/2020-07/application-config.mdx
```diff
@@ -0,0 +1,168 @@
...
+
+Furthermore, the new configuration file tries to infer as many required information as possible.
```
tdeekens, line 23:
```suggestion
Furthermore, the new configuration process tries to infer as much information as possible to reduce the amount of required fields.
```

website/src/releases/2020-07/application-config.mdx
```diff
@@ -0,0 +1,168 @@
...
+
+# Migrating to the new configuration file format
+
+The new configuration file is a JSON file with one of the following names:
+
+- `.custom-application-configrc`
+- `.custom-application-config.json`
+- `custom-application-config.json`
+
+The file is automatically loaded by the other packages, so you don't need to explicitly specify it in, for example, the `mc-scripts` commands.
```
tdeekens, line 35:
```suggestion
The file is automatically loaded by the packages depending on it, so you don't need to explicitly specify it anywhere. This applies for instance to the `mc-scripts` commands.
```

website/src/releases/2020-07/application-config.mdx
```diff
@@ -0,0 +1,168 @@
...
+
+<Info>
+The file needs to be located in the project root folder.
```
tdeekens, line 38:
```suggestion
The file needs to be present in the project root folder.
```

website/src/releases/2020-07/application-config.mdx
```diff
@@ -0,0 +1,168 @@
...
+    }
+  }
+}
+```
+
+That's it! All other values are inferred from the config, like the Content-Security-Policy (CSP) headers, etc.
+
+<Info>
+
+Note that the environment placeholder values `${env:VALUE}` are still working.
```
tdeekens, line 113:
```suggestion
Additionally, note that the environment placeholder syntax `${env:VALUE}` continues to work.
```

website/src/releases/2020-07/application-config.mdx
```diff
@@ -0,0 +1,168 @@
...
+
+</Info>
+
+# Migrating `mc-scripts` commands
+
+The new configuration file is automatically loaded and therefore there is no need to explicitly pass the file path to the `mc-scripts` commands in the `package.json`
```
tdeekens, line 119:
```suggestion
The new configuration file is automatically loaded. Therefore, there is no need to explicitly pass the file path to the `mc-scripts` commands in the `package.json`
```

website/src/releases/2020-07/application-config.mdx
```diff
@@ -0,0 +1,168 @@
...
+      "/.custom-application-configrc",
+      "/.custom-application-config.json",
+      "/custom-application-config.json"
+    ],
+    "url": "https://unpkg.com/@commercetools-frontend/application-config/schema.json"
+  }
+]
+```
+
+Then your editor can offer autocompletion and validation of the JSON properties.
```
tdeekens, line 166:
```suggestion
Through this your editor can offer autocompletion and validation of the JSON properties.
```

packages/application-config/schema.json
```diff
@@ -52,7 +52,7 @@
       "type": "string"
     },
     "env": {
-      "description": "This is an object of keys that represent different environments (for example `production`). The environment used depends on the environment variable MC_APP_ENV. If MC_APP_ENV isn't set then NODE_ENV will be used, if it's not set then it defaults to \"development\"",
+      "description": "This is an object of keys that represent different environments (for example `production`). The environment used depends on the environment variable `MC_APP_ENV`. If `MC_APP_ENV` isn't set then `NODE_ENV` will be used. If neither is set, it defaults to `development`",
```
mohib0306, line 55:
```suggestion
      "description": "This is an object of keys that represents different environments (for example `production`). The environment used depends on the environment variable `MC_APP_ENV`. If `MC_APP_ENV` isn't set then `NODE_ENV` will be used. If neither is set, it defaults to `development`",
```

packages/application-config/src/schema.ts
```diff
@@ -27,7 +27,7 @@ export interface JSONSchemaForCustomApplicationConfigurationFiles {
    */
   mcApiUrl?: string;
   /**
-   * This is an object of keys that represent different environments (for example `production`). The environment used depends on the environment variable MC_APP_ENV. If MC_APP_ENV isn't set then NODE_ENV will be used, if it's not set then it defaults to "development"
+   * This is an object of keys that represent different environments (for example `production`). The environment used depends on the environment variable `MC_APP_ENV`. If `MC_APP_ENV` isn't set then `NODE_ENV` will be used. If neither is set, it defaults to `development`
```
mohib0306, line 30:
```suggestion
   * This is an object of keys that represents different environments (for example `production`). The environment used depends on the environment variable `MC_APP_ENV`. If `MC_APP_ENV` isn't set then `NODE_ENV` will be used. If neither is set, it defaults to `development`
```

PR #1636, opened 2020-07-18
fix(config): preserve full URLs for inferred CSP header values

emmenko:


packages/application-config/src/process-config.ts
```diff
@@ -89,10 +90,10 @@ const processConfig = ({
 
   cachedConfig = {
     env: {
-      ...additionalAppEnv,
+      ...omitEmpty(additionalAppEnv),
```
emmenko, line 93: All empty fields, for example `{foo: ''}` will be removed for convenience.

packages/application-config/src/process-config.ts
```diff
@@ -106,15 +107,15 @@ const processConfig = ({
         ...appConfig.headers?.csp,
         'connect-src': getUniqueValues(
           appConfig.headers?.csp['connect-src'],
-          [mcApiUrl.hostname].concat(isProd ? appUrl.hostname : [])
+          [mcApiUrl.origin].concat(isProd ? [appUrl.href] : [])
         ),
         'script-src': getUniqueValues(
           appConfig.headers?.csp['script-src'],
-          isProd ? [appUrl.hostname] : []
+          isProd ? [appUrl.href, cdnUrl.href] : []
```
emmenko, line 114: I forgot to also assign the cdn url to the csp directive.

packages/application-config/src/substitute-env-variable-placeholders.ts
```diff
@@ -25,9 +25,9 @@ const substituteEnvVariablePlaceholder = (
   meta: { processEnv: NodeJS.ProcessEnv } = { processEnv: process.env }
 ) => {
   const [, requestedEnvVar] = valueOfPlaceholder.split(':');
-  const valueOfEnv = meta.processEnv[requestedEnvVar];
+  const hasEnvField = meta.processEnv.hasOwnProperty(requestedEnvVar);
 
-  if (!valueOfEnv) {
+  if (!hasEnvField) {
```
emmenko, line 30: This is to allow env variables that have an empty value

PR #1641, opened 2020-07-20
fix: unify login/logout dev routes for http servers

emmenko:
This PR aims to fix an issue that we currently have with one of our internal apps, due to the new application config.

TL;DR: the login/logout dev pages that we use in the dev server and in the mc-http-server should only be used when the property `disableAuthRoutesOfDevServer` is explicitly passed or when the application is not running behind a proxy.


packages/mc-http-server/routes/login.js
```diff
@@ -1,7 +0,0 @@
-module.exports = (env) => (request, response, next) => {
-  if (!env.servedByProxy) {
```
emmenko, line 2: This was the original issue, where the `disableAuthRoutesOfDevServer` wasn't taken into account.

With the new config, the `servedByProxy` value is inferred from the node env or app env.
In the case of the auth app, when starting the app locally in production mode (`NODE_ENV=production MC_APP_ENV=development`), the `servedByProxy` is set to false, causing the dev login page to be rendered. However, for the auth app, we want to render the "real" login page and for that we need to use the `disableAuthRoutesOfDevServer`.

packages/mc-dev-authentication/middlewares/create-login-middleware.js
```diff
@@ -0,0 +1,9 @@
+const createLoginMiddleware = (env) => (request, response, next) => {
+  if (Boolean(env.disableAuthRoutesOfDevServer) || env.servedByProxy) {
```
tdeekens, line 2: Just for clarify. Either of them are booleans of strings?

emmenko: They're booleans usually. In theory the `disableAuthRoutesOfDevServer` can also be a string. Let me amend that. Thanks for pointing it out.

PR #1645, opened 2020-07-21
refactor(config): to be more strict on deriving the prod env based on the MC_APP_ENV

emmenko:
TL;DR: in case the `MC_APP_ENV` is defined, we consider that it's a production environment unless it's one of `development` or `test`. This allows to use for example the `staging` value, which from the application perspective is still considered a production environment.

### DIFF
--- a/.github/workflows/publish-cloud-registry.yml
+++ b/.github/workflows/publish-cloud-registry.yml
@@ -3,7 +3,7 @@ name: Publish to Google Cloud Registry
 on:
   push:
     tags:
-      - "@commercetools-frontend/mc-http-server@*"
+      - "*"
 
 jobs:
   mc-http-server:
--- a/packages/application-config/test/process-config.spec.js
+++ b/packages/application-config/test/process-config.spec.js
@@ -118,6 +118,41 @@ describe('processing a simple config', () => {
       });
     });
   });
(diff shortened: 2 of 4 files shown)


packages/application-config/test/process-config.spec.js
```diff
@@ -118,6 +118,41 @@ describe('processing a simple config', () => {
...
+            MC_APP_ENV: 'staging',
+          },
+        })
+      );
+      expect(result).toEqual({
+        env: {
+          applicationId: undefined,
+          applicationName: 'avengers-app',
+          cdnUrl: 'https://avengers.app/',
+          env: 'staging',
```
emmenko, line 136: Yeah basically we derive the "production" value from the env variables but also use it for this `env` field.

90% of the cases this is always production. Only for our internal apps it's nice to differentiate it with staging.
Because of that, I didn't want to accept an additional value only for this use case.

.github/workflows/publish-cloud-registry.yml
```diff
@@ -3,7 +3,7 @@ name: Publish to Google Cloud Registry
 on:
   push:
     tags:
-      - "@commercetools-frontend/mc-http-server@*"
+      - "*"
```
emmenko, line 6: PS: let's see if this helps triggering the workflow. The job itself has its own condition anyway.

PR #1647, opened 2020-07-21
fix(ci): use the release event type to trigger the google cloud registry workflow

emmenko:
I've been trying to find out more info about the proper trigger for the google cloud registry workflow.

I think we've been using the "wrong" trigger `on.push.tags`, because the release bot uses the `createRelease` function (https://github.com/changesets/action/blob/bfeb9e077e6cf393e4c4ef17e2bbc75b308b4364/src/run.ts#L35-L40) to create the release and implicitly the new tag.

So that's (maybe) why the workflow never got triggered.

Anyway, let's see what happens with this one.

### DIFF
--- a/.github/workflows/publish-cloud-registry.yml
+++ b/.github/workflows/publish-cloud-registry.yml
@@ -1,9 +1,8 @@
 name: Publish to Google Cloud Registry
 
 on:
-  push:
-    tags:
-      - "*"
+  release:
+    types: [created]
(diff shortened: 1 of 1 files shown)


emmenko:
Cool, this approach seems to work!

image

The only little "downside" is that for every new release, an attempt to start the workflow is logged but at least only the one matching the http-server will run.

PR #1648, opened 2020-07-21
fix(app-shell): coercing values for stringified array with double escaped quotes

emmenko:
Nasty little edge case. I found this issue when passing a stringified array as an env variable, the injected value in the `window.app` has this value (note the double escaped quotes)

```
"foo":"[\\\"bar\\\"]"
```

This somehow causes the coercion to fail. The only solution I was able to find, was to do a `.replace` and remove the escaped character.

The mysteries of JS...

### DIFF
--- a/.changeset/ninety-dancers-pump.md
+++ b/.changeset/ninety-dancers-pump.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': patch
+---
+
+fix(app-shell): coercing values for stringified array with double escaped quotes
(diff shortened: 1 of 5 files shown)


emmenko:
I was debugging this issue for like 2 hours and almost lost my sanity...

mohib0306:
May we can add more fixture values so that most of the intended cases are covered. 

p.s. This change is not a blocker

emmenko:
@mohib0306 hmm I'm not sure I get what you mean. The example with `{ someArray: ["hi"] }` is covered in the test cases. Can you clarify if I missed something?

mohib0306:
I was just mentioning that the function will convert a `someField: [12]` (array of single element) to `someField: 12` (not array anymore), and if this is intended.

emmenko:
```
> String([1])
'1'
```

What the hell?? You're right...let me see what is going on here....

emmenko:
Ah right, the `toString` implementation of the array converts the array object into a comma separated list of the values...

emmenko:
Thanks for pointing it out, I'll open a new PR!

PR #1650, opened 2020-07-22
 fix(app-shell): make coerce values a react hook, bail out if value is not a string

emmenko:
Leftover of #1648 , thanks @mohib0306 !! 🙏


packages/application-shell/src/components/application-shell/use-coerced-environment-values.ts
```diff
@@ -32,4 +45,19 @@ const shallowlyCoerceValues = (uncoercedEnvironmentValues: ShallowJson) =>
...
+  environment: ShallowJson
+): TApplicationContext<AdditionalEnvironmentProperties>['environment'] => {
+  const coercedEnvironmentValues = React.useMemo(
+    () =>
+      shallowlyCoerceValues(environment) as TApplicationContext<
+        AdditionalEnvironmentProperties
+      >['environment'],
+    [environment]
+  );
+  return coercedEnvironmentValues;
```
tdeekens, line 60: Nice!

PR #1672, opened 2020-08-06
feat(react-notifications) add max resource limit exceeded error translation

tdeekens:
#### Summary

This pull request adds an error message to be translated for the `MaxResourceLimitExceeded` API error.

#### Description

The API error contains the message `You have exceeded the limit of $limit resources of type $resourceTypeId.`. I don't think it's feasible or valuable to translate the `resourceTypeId`. Lots of work in the MC goes into preventing users to run into the error itself.

### DIFF
--- a/packages/react-notifications/src/components/notification-kinds/api-error-message/api-error-message.tsx
+++ b/packages/react-notifications/src/components/notification-kinds/api-error-message/api-error-message.tsx
@@ -200,7 +200,7 @@ function getSpecialFormattedMessageByErrorCode(
   if (
     error.code === 'MaxResourceLimitExceeded'
   ) {
-    return intl.formatMessage(messages.TaxCategoryDuplicateCountry);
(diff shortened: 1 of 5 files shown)


packages/react-notifications/src/components/notification-kinds/api-error-message/api-error-message.tsx
```diff
@@ -197,5 +197,11 @@ function getSpecialFormattedMessageByErrorCode(
     });
   }
 
+  if (
+    error.code === 'MaxResourceLimitExceeded'
+  ) {
+    return intl.formatMessage(messages.TaxCategoryDuplicateCountry);
```
emmenko, line 203:
```suggestion
    return intl.formatMessage(messages.MaxResourceLimitExceeded);
```

tdeekens: Whoopsa. Thanks.

PR #1677, opened 2020-08-10
[Breaking] Migrate packages to new @apollo/client

emmenko:
https://www.apollographql.com/docs/react/migrating/apollo-client-3-migration

See release notes


packages/application-shell/src/configure-apollo.ts
```diff
@@ -100,7 +69,8 @@ export const createApolloClient = () =>
   new ApolloClient({
     link,
     cache: new InMemoryCache({
-      fragmentMatcher,
+      // https://www.apollographql.com/docs/react/data/fragments/#defining-possibletypes-manually
+      // possibleTypes: [],
```
emmenko, line 73: This is the only change that I don't know if "it just works" or not, and we probably need to "test" to ensure things work.

As [documented here](https://www.apollographql.com/docs/react/migrating/apollo-client-3-migration/#cache-improvements), the new Apollo client cache comes with some breaking changes.

packages/application-shell/src/configure-apollo.ts
```diff
@@ -44,49 +47,15 @@ const httpLink = createHttpLink({
...
- * The InMemoryCache itself uses the matcher to determine if caching is intended (https://github.com/apollographql/apollo-client/blob/0166da079da4a9f88eecc7c976ac9d34225f5bf8/packages/apollo-cache-inmemory/src/writeToStore.ts#L242-L266).
- */
-const introspectionQueryResultData = {
-  __schema: {
-    types: [],
-  },
-};
-const fragmentMatcher = new IntrospectionFragmentMatcher({
-  introspectionQueryResultData,
-});
```
tdeekens, line 88: 👼 Heaven 👼

packages/application-shell/src/configure-apollo.ts
tdeekens: As I understood it using the `IntrospectionFragmentMatcher` is similar to using `possibleTypes: []`. Should we maybe allow an application to pass the `possibleTypes` and merge them in with a static list of pre-defined static types (if ever defined) here?

tdeekens:
@emmenko I did some manual testing of most views internally. I can't spot major issues apart from us having to migrate some cache interaction I presume.

emmenko:
We already have some issues spotted by the tests:

image

We need to look a bit more into the cache stuff.

Thanks for doing some initial tests!

tdeekens:
Yeah, saw that one too. If I am not misreading it one has `key` the other one `key` and `id`. Maybe it's around the `dataIdFromObject` as it has been deprecated in v3 I think. One should use `keyFields` of the `TypePolicy` I think.

packages/application-shell-connectors/src/components/project-extension-image-regex/project-extension-image-regex.spec.tsx
```diff
@@ -83,7 +83,7 @@ describe('rendering', () => {
         {
           request: {
             query: FetchProjectExtensionImageRegex,
-            variables: {
+            context: {
```
emmenko, line 86: See release notes to know more about these changes.

FYI: one of the reasons to do that was because the generated graphql types now use an `Exact` object shape for query variables, which obviously does not include request metadata such as `target`, `projectKey`, etc.
This metadata can/should now be passed via the `context` object.

packages/application-shell/src/apollo-links/header-link.ts
```diff
@@ -16,35 +18,36 @@ import {
...
-type GraphQlTarget = typeof GRAPHQL_TARGETS[keyof typeof GRAPHQL_TARGETS];
-
 type ApolloContextWithInMemoryCache = TApolloContext & {
   cache: ApolloClient<NormalizedCacheObject>;
 };
 
 type QueryVariables = {
-  target: GraphQlTarget;
+  // Deprecated, use `{ context: { target } }`
+  target?: TGraphQLTargets;
```
emmenko, line 27: Those metadata options in the `variables` object are now deprecated (for backwards compatibility) and instead should be passed using the `context` object.

packages/application-shell/src/components/authenticated/am-i-logged-in.tsx
```diff
@@ -1,11 +1,12 @@
+import type { TQueryVariablesWithTarget } from '@commercetools-frontend/constants';
 import type {
   TAmILoggedInQuery,
   TAmILoggedInQueryVariables,
 } from '../../types/generated/mc';
 
 import React from 'react';
-import { useQuery } from 'react-apollo';
 import { GRAPHQL_TARGETS } from '@commercetools-frontend/constants';
+import { useMcQuery } from '../../hooks/apollo-hooks';
```
emmenko, line 9: These new hooks are just a wrapper around the original apollo hooks. The only difference is that the `context` object is now properly typed instead of being `any`.

packages/application-shell/src/components/quick-access/quick-access.ctp.graphql
```diff
@@ -20,7 +20,6 @@ query QuickAccess(
   }
 
   productById: product(id: $searchText) @include(if: $canViewProducts) {
-    key
```
emmenko, line 23: We don't know why but without this field the apollo cache does not complain about a merging problem of product objects.

This is the only solution we found that would not result in having a warning. I checked the code and these `key` fields are actually not used.

packages/application-shell/src/components/quick-access/quick-access.tsx
```diff
@@ -408,7 +414,7 @@ const QuickAccess = (props: Props) => {
         commands.push({
           id: `go/product-by-key/product(${productId})`,
           text: intl.formatMessage(messages.showProduct, {
-            productName: data.productByKey.key,
+            productName: searchText,
```
emmenko, line 417: This was the only place where the `key` was used, but given the query by key, we should be able here to use the `searchText` value (which is the same as the key).

packages/application-shell/src/configure-apollo.ts
```diff
@@ -1,34 +1,32 @@
...
-import { createHttpLink } from 'apollo-link-http';
 import {
+  ApolloClient,
+  from,
+  createHttpLink,
   InMemoryCache,
-  IntrospectionFragmentMatcher,
-} from 'apollo-cache-inmemory';
+} from '@apollo/client';
+// import * as apolloVersion from '@apollo/client/version';
```
emmenko, line 9: Apollo does not expose a `version` anymore.

Actually it does but it's in ES6 format, which breaks bundling.

website/src/releases/2020-08/apollo-3.0.mdx
```diff
@@ -0,0 +1,83 @@
...
+title: Migrating to the new Apollo Client 3.0
+description: |
+  With the new release of Apollo Client 3.0, certain breaking changes have been introduced which require some migration steps.\
+  As a consequence, the Application Kit packages have also been released as a new major version v17.\
+  In this release note we go through the important migration steps.
+type: feature
+topics:
+  - Breaking Changes
+  - Dependencies
+published: false
```
emmenko, line 12: This will prevent the page to be published when we merge the PR.

website/src/releases/2020-08/apollo-3.0.mdx
```diff
@@ -0,0 +1,83 @@
+---
+date: 2020-08-25
```
emmenko, line 2: Just a guess, we might need to change this.

website/src/releases/2020-08/apollo-3.0.mdx
```diff
@@ -0,0 +1,83 @@
...
+
+## About the new Apollo Cache
+
+With Apollo version `3.0`, the cache got some improvements as well, resulting in some breaking changes. Please make sure to [read this document](https://www.apollographql.com/docs/react/migrating/apollo-client-3-migration/#cache-improvements) to understand the changes.
+
+For Custom Applications this also might require some migration steps, depending on how you've been using the queries and cache in your application.
+
+The `@commercetools-frontend/application-shell` now exposes an option to allow users to configure the Apollo cache behavior according to the needs of their application.
+
+> FIXME: this is not implemented yet!!!
```
emmenko, line 39: Making the Apollo Cache configurable still needs to be implemented!

emmenko:
FYI: the only think missing here is to make the apollo cache configurable.

packages/application-shell/src/components/quick-access/quick-access.tsx
```diff
@@ -250,18 +251,23 @@ const QuickAccess = (props: Props) => {
...
-        includeProductsByIds: Boolean(
-          canViewProducts && idsOfProductsMatchingSearchText.length > 0
-        ),
-      });
+      const data = await execQuery<TQuickAccessQuery>(
+        QuickAccessQuery,
+        {
+          searchText: sanitize(searchText),
+          // Pass conditional arguments to disable some of the queries
+          canViewProducts,
```
adnasa, line 259: 👍

packages/application-shell/src/configure-apollo.ts
```diff
@@ -44,84 +42,37 @@ const httpLink = createHttpLink({
...
-        // they would clash with the referenced resource.
-        if (
-          customResult.__typename &&
-          referenceTypes.includes(customResult.__typename)
-        )
-          return `${customResult.__typename}:${customResult.id}`;
-
-        return customResult.id;
+      // https://www.apollographql.com/docs/react/caching/cache-configuration/#generating-unique-identifiers
+      typePolicies: {
```
adnasa, line 60: what is missing in the new set up is the reference type.
References can still same id but are of different `typeId`, right?

website/src/releases/2020-08/apollo-3.0.mdx
adnasa: Yes, let's postpone this to September.

adnasa (approved):

website/src/releases/2020-08/apollo-3.0.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+## Changes about Apollo query variables and context
+
+Previously, to perform GraphQL requests to the Merchant Center API Gateway, you would need to specify at least the _GraphQL Target_: https://docs.commercetools.com/custom-applications/main-concepts/graphql#request-format
+
+This value had to be specified in the query `variables` object, which the built-in Merchant Center Apollo HTTP Link would parse and assign it to the request as an HTTP header.
+
+Now that Apollo has a more stable support for the `context` option, it is time that we make use of that instead of "spamming" the query `variables` object with request metadata.
+
+Therefore, all the **metadata options** that you previously would pass to the `variables` object **can now be defined** in the `context` object.
```
adnasa, line 49: We can call this **request context**. WDYT?

emmenko: Sure, it sounds better. Feel free to adjust the terminology, thanks

emmenko:
@adnasa do you still want to work on this or should I take over?

adnasa:
Please take over, I'll focus on the docs of things

emmenko:
Update: I pushed a commit to make the apollo client/cache configurable: https://github.com/commercetools/merchant-center-application-kit/pull/1677/commits/cd204b55f8630ed674f8b646c7e34299c74c6173

website/src/releases/2020-10/migrating-apollo-v3.mdx
```diff
@@ -0,0 +1,103 @@
+---
+date: 2020-10-12
+title: Migrating to the new Apollo Client 3.0
```
emmenko, line 3: FYI: since we're planning other breaking changes to be included in v17, it's likely that we adjust this release note to be about migrating to v17 (which obviously includes the apollo client changes among other things).

packages/application-shell/src/configure-apollo.ts
adnasa: in previous implementation, we made sure that References are unique by leveraging `<typeId>:<id>`.

```js
        if (
          customResult.__typename &&	
          referenceTypes.includes(customResult.__typename)	
        )	
          return `${customResult.__typename}:${customResult.id}`;	

```

This logic is missing now, unless I'm blind

packages/application-shell/src/components/application-shell-provider/application-shell-provider.tsx
```diff
@@ -33,6 +37,13 @@ type Props<AdditionalEnvironmentProperties extends {}> = {
 const ApplicationShellProvider = <AdditionalEnvironmentProperties extends {}>(
   props: Props<AdditionalEnvironmentProperties>
 ) => {
+  const apolloClient = React.useMemo(
+    () => props.apolloClient ?? createApolloClient(),
+    [props.apolloClient]
+  );
```
adnasa, line 43: 🔥 👌

adnasa (commented):
Looks really gooood 🔥 

Still have this question
https://github.com/commercetools/merchant-center-application-kit/pull/1677#discussion_r499446040
wouldn't this be a regression?

packages/application-shell/src/configure-apollo.ts
emmenko: Yeah about that. From what I understand with the new cache implementation, one of the braking changes is that fields are [NOT merged by default](https://www.apollographql.com/docs/react/migrating/apollo-client-3-migration/#breaking-cache-changes).

> By default, the `InMemoryCache` no longer merges the fields of two objects unless those objects have the same unique identifier and that identifier is present in both objects. Additionally, the values of fields with the same name are no longer merged recursively by default. You can define a custom `merge` function for a field to handle both of these changes for a particular field.

So I assume (_which needs to be confirmed and tested in our internal apps_) that for such cases we need to provide a merging strategy on our own.

In theory we could implement the `dataIdFromObject` for this specific case (https://www.apollographql.com/docs/react/caching/cache-configuration/#customizing-identifier-generation-globally) but it might be that Apollo removes this functionality in the future, so I'd like to avoid doing that if possible.

adnasa: all right
Then we test this first then, and see if we have to add `dataIdFromObject`. I'd like to avoid that too

packages/application-shell/src/hooks/apollo-hooks/apollo-hooks.ts
```diff
@@ -0,0 +1,50 @@
...
+function useMcMutation<TData = unknown, TVariables = OperationVariables>(
+  mutation: DocumentNode,
+  options?: TMutationOptionsWithContext<TData, TVariables>
+): MutationTuple<TData, TVariables> {
+  return useMutation<TData, TVariables>(mutation, options);
+}
+
+// Custom Apollo query/mutation wrappers, useful to take advantage
+// of the `context` shape specific to MC queries.
+export { useMcQuery, useMcLazyQuery, useMcMutation };
```
tdeekens, line 50: I was thinking about this some lines above when reading the `context` change. Should we also expose `useSettingsQuery` and `useAdministration` query etc?

emmenko: Hmm you mean to pre-configure the target?
We would then need a 3-way combination for each target then: 

```
useSettingsQuery
useSettingsLazyQuery
useSettingsMutation

// ...
```

adnasa: I'm not sure if it adds bigger benefits besides better DX since the dev doesn't have to fiddle with targets.
If this adds more value in terms of convenience in DX, then maybe we should go for it. I don't think this adds further benefits on testing either.

emmenko: Let's reconsider this as a follow up.

tdeekens: > If this adds more value in terms of convenience in DX, then maybe we should go for it. I don't think this adds further benefits on testing either.

It encapsulates. Given we'd have that already we'd not have to:

1. Change any query or mutation in the MC after this migration to use `context` over `variables`
2. You could search the entire code base easily to see where what requests is being made

emmenko: FYI: I added a point about this in the main release issue tracker: #1769

emmenko:
@tdeekens are the changes regarding the configuration of the Apollo cache via `createApolloClient` ok for you?

tdeekens (approved):
Nope. Looks good.

PR #1685, opened 2020-08-18
refactor: use compiled-data from uikit, expose compiled-data from appkit, use new data table

emmenko:


packages/react-notifications/src/components/notification-kinds/api-error-message/messages.ts
```diff
@@ -21,7 +21,7 @@ export default defineMessages({
     description:
       'User does a bulk update but someone else has saved changes for that element',
     defaultMessage:
-      'Sorry, but we were unable to save your changes as someone else made changes to this same resource while you were editing.',
+      'Sorry, but we were unable to save your changes as someone else made changes to this same source while you were editing.',
```
emmenko, line 24: These changes simply reflect the values in the `core.json`, so that by running the extract command we don't overwrite/change the values.

emmenko:
Alright, I think this is good to go now. It might be that the mc apps need to be updated to latest uikit + appkit + react-intl versions, to have everything working fine.

tdeekens (approved):
🥳🤩😎

mohib0306 (approved):
These changes looks good.

mohib0306:
Just one thing about exporting data. If you remember the discussion about exporting countries lists (at least the english countries data) from our `l10n` package so that we could have better test data generation for MC-FE.

emmenko:
Yep, we can add that as well

jonnybel (approved):

PR #1687, opened 2020-08-20
test: rewrite some tests using msw for mocking graphql requests

emmenko:
I would like to start refactoring the tests to avoid using Apollo mocks and instead using the MSW network mocks: https://mswjs.io/

This should be used in combination with the test data. In this repo we already have some test data mocks but eventually we want to use the new ones we're developing internally.

### DIFF
--- a/packages/application-shell/src/components/fetch-project/fetch-project.spec.js
+++ b/packages/application-shell/src/components/fetch-project/fetch-project.spec.js
@@ -3,7 +3,11 @@ import { setupServer } from 'msw/node';
 import React from 'react';
 import { reportErrorToSentry } from '@commercetools-frontend/sentry';
 import { ProjectMock } from '../../../../../graphql-test-utils';
-import { experimentalRenderAppWithRedux, screen } from '../../test-utils';
+import {
+  experimentalRenderAppWithRedux,
+  screen,
+  waitForElementToBeRemoved,
+} from '../../test-utils';
 import FetchProject from './fetch-project';
 
 jest.mock('@commercetools-frontend/sentry');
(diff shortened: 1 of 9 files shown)


packages/application-shell/src/components/fetch-project/fetch-project.spec.js
```diff
@@ -20,132 +39,19 @@ const renderProject = (options) =>
...
-          },
-          result: {
-            data: createGraphqlResponseForProjectQuery(),
-          },
-        },
-      ],
-    });
-    await rendered.findByText(/Test 1/i);
-  });
-  it('should render loading state', async () => {
```
emmenko, line 111: I'm not sure how we can test the loading state now. The question also is if we actually need to or just skip that.

emmenko: Nvm, I figured that out

packages/application-shell/src/components/fetch-user/fetch-user.spec.js
```diff
@@ -1,13 +1,36 @@
...
+import {
+  experimentalRenderAppWithRedux,
+  screen,
+  waitForElementToBeRemoved,
+} from '../../test-utils';
 import FetchUser from './fetch-user';
 
 jest.mock('@commercetools-frontend/sentry');
 
+const mockServer = setupServer(
```
tdeekens, line 15: I was wondering if we should expose a msw helper/package library from app-kit which allows you to easily mock things also with variables and/or the target.

For now I think this is great but if we would have `msw.withVariables({ target}).withQuery('FetchLoggedInUser').respond(...)` it might keep setup a bit learner? WDYT?

emmenko: We can do something yes but I wouldn't worry about that now. I would first migrate all tests, then see what we can simplify setup-wise out of that.

What we can start thinking about though, is how we can expose/reuse our internal test data.

tdeekens: Totally agree. Just thinking/tinkering forward.

PR #1688, opened 2020-08-21
chore: update cypress to v5

emmenko:
https://docs.cypress.io/guides/references/migration-guide.html#Migrating-to-Cypress-5-0


yarn.lock
```diff
@@ -16,6 +16,24 @@
     call-me-maybe "^1.0.1"
     js-yaml "^3.13.1"
 
+"@apollo/client@^3.0.2":
```
tdeekens, line 19: Huh?

emmenko: I think it's just a nested dep of `graphql-tools`

PR #1698, opened 2020-08-25
test(app-shell): migrate tests to use msw, instead of custom graphql mock server

emmenko:


graphql-test-utils/index.js
```diff
@@ -1,14 +1,7 @@
...
-
-export { default as applyMocksForExternalNetworkRequests } from './apply-mocks-for-external-network-requests';
-export { default as createGraphqlMockServer } from './create-graphql-mock-server';
 export { default as UserMock } from './graphql-models/user';
 export { default as ProjectMock } from './graphql-models/project';
 export { default as ApplicationAppbarMenuMock } from './graphql-models/application-appbar-menu';
 export { default as ApplicationNavbarMenuMock } from './graphql-models/application-navbar-menu';
 export { default as ApplicationNavbarSubmenuMock } from './graphql-models/application-navbar-submenu';
 export { default as CustomApplicationMock } from './graphql-models/custom-application';
 export { default as ProjectExtensionMock } from './graphql-models/project-extension';
```
emmenko, line 7: FYI: As a follow up, we will be start open sourcing our internal test data, starting with the core library. With that, we can then build up the models here in appkit.

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -654,7 +654,10 @@ export const LoadingNavBar = () => {
               isActive={false}
               onClick={() => undefined}
             >
-              <div className={styles['loading-dot-container']}>
+              <div
+                className={styles['loading-dot-container']}
+                data-testid={`dot-container-${index}`}
```
emmenko, line 659: Needed to help with "waiting for the loading navbar to be removed"

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
```diff
@@ -64,6 +64,24 @@ export const SetupFlopFlipProvider = (props: Props) => {
     [flags, props.ldClientSideId, props.projectKey, props.user]
   );
 
+  if (process.env.NODE_ENV === 'test') {
```
tdeekens, line 67: Otherwise the ld adapter will initialise and make requests?

graphql-test-utils/index.js
tdeekens: Sure. That'd be cool.

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
```diff
@@ -64,6 +64,24 @@ export const SetupFlopFlipProvider = (props: Props) => {
...
+    const memoryAdapter = require('@flopflip/memory-adapter').default;
```
emmenko, line 68: Not sure if this the "best" way to do that, but we need to use the memory adapter when testing the app shell.
Previously we were kind of mocking the requests to launchdarkly, but that is a bit tricky and error prone.

tdeekens: Yeah, makes total sense. It's not possible to just wrap it in a `flopflip-adapter.js` which does export either or them based on the env?

emmenko: I double checked the rollup + webpack bundle, and it's correctly stripped from the webpack bundle:

image

adnasa (approved):

tdeekens (approved):

PR #1717, opened 2020-09-07
fix(sentry): update versions, use allowUrls option instead of denyUrls

emmenko:
Caused by #1614 

I mixed up the options, my bad 🤦‍♂️


packages/application-shell/package.json
```diff
@@ -74,7 +74,6 @@
     "@flopflip/types": "2.5.6",
     "@flopflip/memory-adapter": "1.9.2",
     "@flopflip/react-broadcast": "10.2.3",
-    "@sentry/browser": "5.21.3",
```
emmenko, line 77: Not sure why we had this dependency in the appshell package...

tdeekens (approved):

adnasa (approved):

Issue #1739, opened 2020-09-14
Improve documentation around exchange token format for external API


emmenko:
**Which part of the documentation does it affect and which topic does it refer to?**

https://docs.commercetools.com/custom-applications/main-concepts/proxy-to-external-api

**Describe what needs to be improved**

Document a bit more in-depth the format/fields of the JWT exchange token.

PR #1742, opened 2020-09-14
docs(proxy-external): add description of decoded JWT claims

adnasa:
#### Summary

- add description for decoded JWT and their claims
- closes https://github.com/commercetools/merchant-center-application-kit/issues/1739

### DIFF
--- a/website/src/content/main-concepts/proxy-to-external-api.mdx
+++ b/website/src/content/main-concepts/proxy-to-external-api.mdx
@@ -121,7 +121,7 @@ It's imperative that the external API is securely protected and that only users 
 
 </Warning>
 
-### Decoding the token
+### Exchange token structure
 
 You can decode the JWT token with [JWT.io](https://jwt.io/). In the decoded version, you will see the following fields.
 
--- a/website/src/content/main-concepts/proxy-to-external-api.mdx
+++ b/website/src/content/main-concepts/proxy-to-external-api.mdx
@@ -140,13 +140,13 @@ You can decode the JWT token with [JWT.io](https://jwt.io/). In the decoded vers
 These decoded JWT token consist of **Register Claim Names** and **Private claims** and have the following fields:
 
 
-- `exp`: This claim identifies the **expiration time**. The value of this claim requires to be the time before the current time to be valid and accepted.
-- `sub`: This claim identifies the **subject**. The value of `sub` issued by the Merchant Center API Gateway for Custom Applications is the user-ID for Merchant Center.
-- `iss`: This claim identifies the **issuer**. The value of `iss` is either a cloud identifier or a valid URL to the Merchant Center API Gateway. The value differs depending on the region as outlined above.
-- `aud`: This claim identifies the **audience**. The value of **audience** is the exact URL to the External API.
-- `type`: This is public claim identified by the Merchant Center API Gateway. The decoded JWT in the example above is a result of authorization exchange between Merchant Center API Gateway and the External API, hence the name.
-- `<hostname>/claims/project_key`: This is a claim identified by the Merchant Center API Gateway. This identifies the project key that the JWT was issued for.
-- `iat`: This claim identifies **issued at**. The value of **issued at** is the time that JWT was issued.
+- `exp`: The **expiration time** on or after which the exchange token must not be accepted for processing.
+- `sub`: The **subject identifier** is the unique user identifier of a Merchant Center account, specifically the ID of the logged in user.
+- `iss`: The **issuer identifier** is the URL identifier of the server that issued the JWT. In this particular case is the URL of the [Merchant Center API Gateway](/main-concepts/api-gateway).
+- `aud`: The **audience** is the value that the exchange token is intended for. In this particular case is the exact URL passed to the `X-Forward-To` HTTP header.
+- `iat`: This value represents the time at which the JWT was issued.
+- `type`: This value is a **private claim** and is always set to `exchange`.
+- `<iss>/claims/project_key`: This value represents the project key that the request is associated to. The claim is a **public claim** prefixed by a collision-resistant namespace, as per the JWT specification.
 
 ## Validating the JSON Web Token
 
--- a/website/src/content/main-concepts/proxy-to-external-api.mdx
+++ b/website/src/content/main-concepts/proxy-to-external-api.mdx
@@ -123,7 +123,10 @@ It's imperative that the external API is securely protected and that only users 
 
 ### Exchange token structure
 
-You can decode the JWT token with [JWT.io](https://jwt.io/). In the decoded version, you will see the following fields.
+The Exchange token follows the [JWT structure](https://auth0.com/docs/tokens/json-web-tokens/json-web-token-structure), specified by the [JWT specification](https://tools.ietf.org/html/rfc7519).
+This means that the Exchange token contains claims, which are statements about the token and is used to assert the token itself.
+
+The Exchange token structure consist of **Registered Claim Names** and **Private claims**:
 
 ```json
 {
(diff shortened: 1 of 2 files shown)


website/src/content/main-concepts/proxy-to-external-api.mdx
```diff
@@ -214,6 +214,32 @@ exports.handler = async function (request, response) {
...
+  "<hostname>/claims/project_key": "<project-key>",
+  "iat": 1600092341
+}
+```
+
+These fields are according to specification **Register Claim Names** and consist of the following:
+
+- `exp`: This claim identifies the **expiration time**. The claims requires to be before the current time in order to accepted for processing.
+- `sub`: This claim identifies the **subject**. The value of `sub` issued by the API Gateway for Custom Applications is the user-ID for Merchant Center.
+- `iss`: This claim identifies the **issuer**. The value of `iss` is the hostname to the API Gateway itself. This value differs depending on the region the JWT was issued. These are listed under API Gateway under **Hostnames**.
```
adnasa, line 237: I couldn't link to the **API Gateway page -> Hostnames**

website/src/content/main-concepts/proxy-to-external-api.mdx
```diff
@@ -214,6 +214,32 @@ exports.handler = async function (request, response) {
...
+  "type": "exchange",
+  "<hostname>/claims/project_key": "<project-key>",
+  "iat": 1600092341
+}
+```
+
+These fields are according to specification **Register Claim Names** and consist of the following:
+
+- `exp`: This claim identifies the **expiration time**. The claims requires to be before the current time in order to accepted for processing.
+- `sub`: This claim identifies the **subject**. The value of `sub` issued by the API Gateway for Custom Applications is the user-ID for Merchant Center.
```
emmenko, line 236: Please use the term `Merchant Center API Gateway` instead of `API Gateway`.

Maybe we can just say it's the "ID of the commercetools platform user".

tdeekens, line 236:
```suggestion
- `sub`: This claim identifies the **subject**. The value of `sub` issued by the API Gateway for Custom Applications is the ID of the respective commercetools Platform user using the Merchant Center.
```

website/src/content/main-concepts/proxy-to-external-api.mdx
```diff
@@ -214,6 +214,32 @@ exports.handler = async function (request, response) {
...
+- `iss`: This claim identifies the **issuer**. The value of `iss` is the hostname to the API Gateway itself. This value differs depending on the region the JWT was issued. These are listed under API Gateway under **Hostnames**.
+- `aud`: This claim identifies the **audience**. The value of **audience** is the exact URL to the External API.
+- `type`: This is public claim identified by the API Gateway. The decoded JWT in the example above is a result of authorization exchange between API Gateway and the External API, hence the name.
+- `iat`: This claim identifies **issued at**. The value of **issued at** is the time that JWT was issued.
+- `<hostname>/claims/project_key`: This is a public claim identified by the API Gateway. This identifies the project key that the JWT was issued.
```
emmenko, line 241: `...issued for.`

tdeekens, line 241: Maybe we can split this as Private Claims and the above as Public Claims?

website/src/content/main-concepts/proxy-to-external-api.mdx
```diff
@@ -214,6 +214,32 @@ exports.handler = async function (request, response) {
...
+}
+```
+
+These fields are according to specification **Register Claim Names** and consist of the following:
+
+- `exp`: This claim identifies the **expiration time**. The claims requires to be before the current time in order to accepted for processing.
+- `sub`: This claim identifies the **subject**. The value of `sub` issued by the API Gateway for Custom Applications is the user-ID for Merchant Center.
+- `iss`: This claim identifies the **issuer**. The value of `iss` is the hostname to the API Gateway itself. This value differs depending on the region the JWT was issued. These are listed under API Gateway under **Hostnames**.
+- `aud`: This claim identifies the **audience**. The value of **audience** is the exact URL to the External API.
+- `type`: This is public claim identified by the API Gateway. The decoded JWT in the example above is a result of authorization exchange between API Gateway and the External API, hence the name.
```
emmenko, line 239: This is a private claim.

Public claims should always be namespaced.

emmenko (changes requested):
Thanks, I left some comments as we're not quite there yet.

website/src/content/main-concepts/proxy-to-external-api.mdx
```diff
@@ -214,6 +214,32 @@ exports.handler = async function (request, response) {
 };
 ```
 
+### Decoding the token
```
emmenko, line 217: I think we should move this into the `Authenticating requests from the external API` section.
For example: https://auth0.com/docs/tokens/id-tokens/id-token-structure

website/src/content/main-concepts/proxy-to-external-api.mdx
adnasa: never mind. we already outlined **cloud regions** and **MC API url** above, so I point to that.

website/src/content/main-concepts/proxy-to-external-api.mdx
adnasa: could not find it under the spec, so I assumed it was public. will address

adnasa:
Addressed all feedback

tdeekens (approved):

website/src/content/main-concepts/proxy-to-external-api.mdx
```diff
@@ -121,6 +121,32 @@ It's imperative that the external API is securely protected and that only users
...
+
+
+- `exp`: This claim identifies the **expiration time**. The value of this claim requires to be the time before the current time to be valid and accepted.
+- `sub`: This claim identifies the **subject**. The value of `sub` issued by the Merchant Center API Gateway for Custom Applications is the user-ID for Merchant Center.
+- `iss`: This claim identifies the **issuer**. The value of `iss` is either a cloud identifier or a valid URL to the Merchant Center API Gateway. The value differs depending on the region as outlined above.
+- `aud`: This claim identifies the **audience**. The value of **audience** is the exact URL to the External API.
+- `type`: This is public claim identified by the Merchant Center API Gateway. The decoded JWT in the example above is a result of authorization exchange between Merchant Center API Gateway and the External API, hence the name.
+- `<hostname>/claims/project_key`: This is a claim identified by the Merchant Center API Gateway. This identifies the project key that the JWT was issued for.
+- `iat`: This claim identifies **issued at**. The value of **issued at** is the time that JWT was issued.
```
emmenko, line 149:
```suggestion
- `exp`: The **expiration time** on or after which the exchange token must not be accepted for processing.
- `sub`: The **subject identifier** is the unique user identifier of a Merchant Center account, specifically the ID of the logged in user.
- `iss`: The **issuer identifier** is the URL identifier of the server that issued the JWT. In this particular case is the URL of the [Merchant Center API Gateway](/main-concepts/api-gateway).
- `aud`: The **audience** is the value that the exchange token is intended for. In this particular case is the exact URL passed to the `X-Forward-To` HTTP header.
- `iat`: This value represents the time at which the JWT was issued.
- `type`: This value is a **private claim** and is always set to `exchange`.
- `<iss>/claims/project_key`: This value represents the project key that the request is associated to. The claim is a **public claim** prefixed by a collision-resistant namespace, as per the JWT specification.
```

emmenko (changes requested):

emmenko (approved):
Looking good, thanks a lot for jumping into it! 👌

tdeekens (approved):

PR #1743, opened 2020-09-15
Define test data for user and project, using new core package

emmenko:
Trying to play around and test the new `@commercetools-test-data/core` package for building test data models.

I have some questions regarding defining the generator/transformer for the user + project, but I need to have a chat with @tdeekens first to clarify some points.


tdeekens (commented):
Looks razor sharp. Thanks for pushing the idea here.

emmenko:
This now should correctly implement the transformers for the permissions -> allAppliedPermissions (etc).

As a result of this, there is a pending fix in the test-data repo: https://github.com/commercetools/test-data/pull/4

test-data/project/fields/applied-menu-visibilities/transformers.ts
```diff
@@ -0,0 +1,25 @@
...
+const upperFirst = (value: string): string =>
+  value.charAt(0).toUpperCase() + value.slice(1);
+
+const transformers = {
+  graphql: Transformer<TMenuVisibilities, TAppliedMenuVisibilitiesGraphql>(
+    'graphql',
+    {
+      replaceFields: ({ fields }) =>
+        Object.entries(fields).map(([key, value]) => ({
+          __typename: 'AppliedMenuVisibilities',
```
tdeekens, line 17: Can‘t this be an add field.

test-data/project/generator.ts
```diff
@@ -0,0 +1,34 @@
...
+    // The following fields are built in the builder
+    owner: null,
+    suspension: null,
+    expiry: null,
+    permissions: null,
+    actionRights: null,
+    menuVisibilities: null,
+    dataFences: null,
+  },
+  postBuild: (project) => {
```
tdeekens, line 28: Might be possible setting it to null and then replaceField. To not depend in test-data-bot for it.

tdeekens (commented):
Looks very very promising. Thanks for pushing the topic!

test-data/project/fields/applied-menu-visibilities/transformers.ts
emmenko: If you use `replaceFields` you can't (and does not make sense) to use `addFields` or `removeFields`.

emmenko:
Anything else to add/change here? @tdeekens @adnasa

tdeekens (approved):
Nope. This is awesome.

adnasa (approved):
love it

PR #1754, opened 2020-09-24
fix(application-shell): to support token retry on mc backend graphql …

tdeekens:
#### Summary

The Merchant Center backend has various GraphQL queries and mutations it supports. For instance project creation. For these queries our so called "token with context" is required. If the token is lacking the right scopes or is invalid we should also retry the request upon a `401`.

I noticed this bug internally when creating a project. Without a prior successful request to our platform APIs the request fails. It doesn't retry however even though it returns a `401` with `invalid_token`. I assume that this is the reason.


emmenko:
Thanks. Do you remember why we limited the retry to only certain graphql targets?

emmenko (approved):

tdeekens:
> Thanks. Do you remember why we limited the retry to only certain graphql targets?

Not fully. Maybe just to control it. My assumption is that when introduced we didn't have a powerful enough schema in the MC backend to need token retrying. Now we do and forgot to add it.

packages/application-shell/src/apollo-links/token-retry-link.spec.js
```diff
@@ -214,4 +214,12 @@ describe('supported GraphQL targets', () => {
       })
     ).toBe(true);
   });
+
+  it('should not support the `merchant center backend service` GraphQL target', () => {
```
adnasa, line 218: should not support?

adnasa (approved):

packages/application-shell/src/apollo-links/token-retry-link.spec.js
tdeekens: Thanks!

packages/application-shell/src/components/fetch-project/fetch-project.spec.js
```diff
@@ -52,7 +52,7 @@ describe('rendering', () => {
   it('should render error state', async () => {
     mockServer.use(
       graphql.query('FetchProject', (req, res, ctx) => {
-        return res.once(ctx.status(401));
+        return res.data(ctx.status(401));
```
tdeekens, line 55: @emmenko this is actually a very nice catch. The project query will fail once with 401 then result in data being loaded after.

So the fact is that the query can load twice until you get an error. I am trying to think if this is problematic but I think not. What do you think?

emmenko: Hmm let me think. Now that I see this, I'm actually wondering how is this even working...shouldn't we define the main mock in the `setupServer` as `res.once`? 🤔

tdeekens: Yeah, I think that's a bit what we talked about yesterday. `res.once` is actually quite powerful. We might want to use it as the default more. While having mock handlers local is often easier to maintain than this global catch-all stuff also if it's a bit more redundant.

I tried to improve it in ac18bf5. I find such tests incredibly powerful to be honest ⭐.

packages/application-shell/src/components/fetch-project/fetch-project.spec.js
```diff
@@ -44,20 +33,71 @@ const renderProject = (options) =>
...
 describe('rendering', () => {
-  it('should fetch project and pass data to children function', async () => {
-    renderProject();
-    await waitForElementToBeRemoved(() => screen.getByText('loading...'));
-    expect(screen.getByText(/Test 1/i)).toBeInTheDocument();
+  describe('when fetching project succeeds', () => {
+    it('should fetch project and pass data to children function', async () => {
+      mockServer.use(
+        graphql.query('FetchProject', (req, res, ctx) =>
+          res.once(
```
emmenko, line 40: If you're doing it like this (define the mock for each test), there is no need to use `res.once`.

Using `res.once` is only useful if you have multiple mocks for the same query in the same test.
After each test, the mocks are reset.

packages/application-shell/src/components/fetch-project/fetch-project.spec.js
```diff
@@ -12,18 +12,7 @@ import FetchProject from './fetch-project';
 
 jest.mock('@commercetools-frontend/sentry');
 
-const mockServer = setupServer(
-  graphql.query('FetchProject', (req, res, ctx) =>
```
emmenko, line 16: The idea of having this here was that in the tests we don't need to provide any mocks.

packages/application-shell/src/components/fetch-project/fetch-project.spec.js
tdeekens: Yes, but both mocks operate on the same query. So I want to make sure to respond with one first once and then the other second. 

Can you maybe commit the change your suggesting? I don't fully understand the change request being made.

packages/application-shell/src/components/fetch-project/fetch-project.spec.js
tdeekens: Okay, what's the change request being made? Can you commit it maybe if you've something in mind?

packages/application-shell/src/components/fetch-project/fetch-project.spec.js
emmenko: Sure, let me try

emmenko: Please have a look: https://github.com/commercetools/merchant-center-application-kit/pull/1754/commits/19659837f407a3ac356854342a2f35ca65dcd9d8

tdeekens: Got cha. Still not to fond of global setup or mocks but that is subjective:)

tdeekens:
Tests now fail continuously on CI. Having a look. So far hard to reproduce locally.

tdeekens:
I can't reproduce the failure so far. I also can't see why it happens. Maybe @emmenko you have a hunch?

PR #1756, opened 2020-09-25
fix(app-shell): error link to support opting out of token retry

tdeekens:
#### Summary

This pull request improve our `error-link` to allow opting out of token retrying.

#### Description

In a prior pull request we allowed token retrying on the MC API. That is fine. We also added opting out of token retrying via a `context.skipTokenRetry`. While integrating we noticed that token retry still happens on the `AmILoggedIn`-query. Opon investigation it was noticed that it's _not_ the `token-retry-link` which is triggering the token retry in the non authenticated case. But the `error-link`. As both links share little to no code this issue slipped through. This pull request builds up a shared utils between all links and fixes the error link to skip token retry when the respective context variable is set.


packages/application-shell/src/apollo-links/token-retry-link.spec.js
```diff
@@ -227,47 +225,3 @@ describe('tokenRetryLink', () => {
     });
   });
 });
-
-describe('supported GraphQL targets', () => {
```
tdeekens, line 231: Moved to util

packages/application-shell/src/apollo-links/token-retry-link.ts
```diff
@@ -1,55 +1,12 @@
...
 
 import { RetryLink } from 'apollo-link-retry';
+import { STATUS_CODES } from '@commercetools-frontend/constants';
 import {
-  STATUS_CODES,
-  GRAPHQL_TARGETS,
-} from '@commercetools-frontend/constants';
-import { SUPPORTED_HEADERS } from '../constants';
-
-type TTokenRetryGraphQlTarget =
```
tdeekens, line 10: Moved to util.

emmenko (approved):
Looks good, I also don't recall why we did it that way.
Thanks again for fixing that!

PR #1764, opened 2020-09-28
fix(app-shell): to handle errors without extentions property

tdeekens:
#### Summary

This pull request fixes token regeneration for GraphQL APIs which do not have an `extentions` property.

#### Description

Some internal APIs do not expose an `extensions` property but an error which looks e.g. like this

```json
{"statusCode":401,"message":"invalid_token","errors":[{"code":"invalid_token","message":"invalid_token"}],"error":"invalid_token"}
```

As a result I propose that we don't only assume a GraphQL error based on the presence of the `extensions` property but also based on the `message` property which is part of the `GraphQLError` type. Please note that `code` is actually not part of it.

### DIFF
--- a/packages/application-shell/src/apollo-links/error-link.spec.js
+++ b/packages/application-shell/src/apollo-links/error-link.spec.js
@@ -56,7 +56,7 @@ describe('with unauthenticated error', () => {
     let resolvedResponse;
     const responses = {
       success: { data: { sample: { id: 'sample-id' } } },
-      unauthenticated: {
+      unauthenticatedWithExtensionCode: {
         data: null,
         errors: [
           {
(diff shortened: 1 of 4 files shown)


packages/application-shell/src/apollo-links/error-link.spec.js
```diff
@@ -66,48 +66,102 @@ describe('with unauthenticated error', () => {
           },
         ],
       },
+      unauthenticatedWithStatusCode: {
+        data: null,
+        statusCode: 401,
+        errors: [
+          {
+            message: 'invalid_token',
```
tdeekens, line 74: The `code` is not standard so we should just use message I suppose.

packages/application-shell/src/apollo-links/error-link.spec.js
```diff
@@ -66,48 +66,102 @@ describe('with unauthenticated error', () => {
...
     };
     describe('when token retry is not skipped', () => {
-      beforeEach(async () => {
-        const debugLink = new ApolloLink((operation, forward) => {
-          context = operation.getContext();
-          operation.setContext({
-            headers: {
-              'X-Graphql-Target': GRAPHQL_TARGETS.COMMERCETOOLS_PLATFORM,
-            },
+      describe('when unauthorized via extension code', () => {
```
tdeekens, line 80: The old test.

packages/application-shell/src/apollo-links/error-link.spec.js
```diff
@@ -66,48 +66,102 @@ describe('with unauthenticated error', () => {
...
+              'X-Force-Token': 'true',
+            })
+          );
+        });
+        it('should eventually resolve with data', () => {
+          const [{ values }] = resolvedResponse;
+          expect(values).toEqual([responses.success]);
+        });
       });
+      describe('when unauthorized via status code', () => {
```
tdeekens, line 123: The new test.

packages/application-shell/src/apollo-links/error-link.ts
```diff
@@ -36,8 +36,12 @@ const errorLink = onError(
       const context = operation.getContext() as TApolloContext;
 
       for (const err of graphQLErrors) {
+        const isNonAuthenticatedViaExtensionCode =
+          err?.extensions?.code === 'UNAUTHENTICATED';
+        const isNonAuthenticatedViaCode = err?.message === 'invalid_token';
```
tdeekens, line 41: Just moved up for clarity.

packages/application-shell/src/apollo-links/utils.ts
```diff
@@ -60,7 +60,11 @@ const isGraphQLError = (
   error: ErrorResponse['graphQLErrors']
 ): error is GraphQLError[] =>
   Array.isArray(error) &&
-  error.some((err) => (err as GraphQLError).extensions !== undefined);
+  error.some(
+    (err) =>
+      (err as GraphQLError).message !== undefined ||
+      (err as GraphQLError).extensions !== undefined
```
tdeekens, line 66: I think message is even required on any `GraphQLError` according to the type. Maybe it alone is enough. Open for a discussion.

packages/application-shell/src/apollo-links/error-link.ts
emmenko, line 41: Should we mention here that this is a "workaround"? At least for us for future references or something.

packages/application-shell/src/apollo-links/error-link.spec.js
```diff
@@ -66,48 +66,102 @@ describe('with unauthenticated error', () => {
           },
         ],
       },
+      unauthenticatedWithStatusCode: {
+        data: null,
+        statusCode: 401,
```
emmenko, line 71: Is this correct?

emmenko (approved):
Thanks, looks good  and it should be "good enough" for now.

packages/application-shell/src/apollo-links/error-link.spec.js
tdeekens: Yeah, but we can remove it. Some APIs add a statusCode property.

tdeekens: Removed.

packages/application-shell/src/apollo-links/error-link.ts
tdeekens: Added a comment.

PR #1766, opened 2020-09-28
fix(website): links to other microsites to use the new paths

emmenko:


tdeekens (approved):
Nice one. I guess we wait until the other ones are live tomorrow?

PR #1767, opened 2020-09-29
chore(package.json): upgrade ui-kit

adnasa:
#### Summary

- upgrade ui-kit
- extracted from https://github.com/commercetools/merchant-center-application-kit/pull/1758


mohib0306 (approved):
Nice, no visual changes.

emmenko:
Changeset is missing

adnasa:
> Changeset is missing

I didn't think that changeset was needed for dep updates, but sure no problem :)

tdeekens:
> I didn't think that changeset was needed for dep updates, but sure no problem :)

It is. Otherwise no package will be published and no consumer will be able to consume the updated package with the dependency update. In tern following changes towards individual packages and their changesets will lead to packages being published with different dependency versions.

Assume updating ui-kit for A,B and C without a changeset. Then changing something in A while adding a changeset and publishing it. Now B and C have an old version of the dependency while A is updated as it was changed after the dependency update.

adnasa:
thanks for the example

tdeekens (approved):

emmenko (approved):

PR #1768, opened 2020-09-29
chore: setup linkinator for custom applications

davifantasia:


package.json
```diff
@@ -43,6 +43,8 @@
     "template-starter:start": "yarn --cwd application-templates/starter start",
     "template-starter:start:prod:local": "yarn --cwd application-templates/starter start:prod:local",
     "build": "NODE_ENV=production lerna run build --no-private",
+    "postbuild": "yarn check-links",
```
emmenko, line 46: This should only be used as post build of the website.

```suggestion
    "postbuild:website": "yarn check-links",
```

package.json
```diff
@@ -43,6 +43,8 @@
...
+    "check-links": "npx linkinator ./website/public/custom-applications/public",
```
emmenko, line 47:
```suggestion
    "check-links": "npx linkinator ./website/public/custom-applications",
```

package.json
emmenko: Better to include it in the postbuild hook of the website package

emmenko (changes requested):
Thanks. Can you move it to the `website` folder only though?
It does not need to run in the main repo.

website/linkinator.config.json
```diff
@@ -0,0 +1,28 @@
+{
+  "recurse": true,
+  "silent": true,
+  "skip": [
+    ".(json|js|png|jpg|svg|css)(\\?(.*)+)?$",
```
emmenko, line 5: I included this rule to exclude check for static assets (which are handled by Gatsby anyway).

adnasa:
This came out of nowhere for me.
What's beneficial with linkinator?

emmenko:
It's what we use in the docs websites to check that links are correct.

emmenko (approved):

Issue #1769, opened 2020-09-30
Upcoming breaking changes (v17)


emmenko:
This issue aims to give an overview of all the **breaking changes** we would like to introduce in the next major version **v17**.

- [x] #1677 Migrate to Apollo v3
- [x] #1790 Remove the `extract-intl` command
  - Remove the babel plugin for intl
- [x] #1787 Make the `--use-local-assets` the default behavior
  - The usage of this flag is confusing and it turns out that this is what you need 99% of the time so we should make it the default behavior.
  - Remove the `mc-http-server`.
- [x] #1791 Remove `experimentalRenderAppWithRedux`, instead expose a `RealApolloProvider` component wrapper that can be passed as `ApolloProviderComponent` option to `renderApp` and `renderAppWithRedux`
- [x] Remove deprecated CLI options
  - https://github.com/commercetools/merchant-center-application-kit/blob/96b3af7c3e8276e7a136e48c3d313fe6d504099d/packages/mc-html-template/src/compile-html.js#L9
  - https://github.com/commercetools/merchant-center-application-kit/blob/master/packages/application-config/src/load-deprecated-config.ts
- [x] #1781 Remove deprecated exports:
  - https://github.com/commercetools/merchant-center-application-kit/blob/74d43d2abe7efa085780692d8a364de4aaad7278/packages/application-shell/src/index.ts#L21-L23
- [x] #1780 Remove deprecated options, due to oppressive language renaming. 
  - https://github.com/commercetools/merchant-center-application-kit/blob/564cd9186d23ea34886d1c41718486e16d3ca90e/packages/application-shell/src/components/application-shell/application-shell.tsx#L77-L79

tdeekens:
>  Remove deprecated hostnames

I am not sure we can do that yet. Accordingly to my knowledge customers are still relying on them. The hostname migration date at least is postponed until January next year. We can remove them here if there is no change for old published versions though.

> React Router to v6 + History to v5

I fear this would entail a lot of work internally. Maybe we can smoothen the migration out internally but evaluating changes we need to make first.

> Remove experimentalRenderAppWithRedux, instead expose a RealApolloProvider component wrapper that can be passed as ApolloProviderComponent option to renderApp and renderAppWithRedux

Would it possible to do this in a two step migration. First expose the wrapper and allow passing it. Recommend to migrate until the then next major v18 and only then remove it?

adnasa:
I agree with Tobi's points above.

[...]

PR #1772, opened 2020-10-01
test(app-kit): remove use of `enzyme` in playground and application-templates/starter

adnasa:
#### Summary

- removes use of `enzyme` in `playground` and `application-templates`
- address one of https://github.com/commercetools/merchant-center-application-kit/issues/1769

#### Description

There is still use of `enzyme` in `SDK.Get`. I split that in a separate step.

EDIT:
I don't think it makes sense yet to update the docs.

### DIFF
--- a/playground/src/components/entry-point/entry-point.spec.js
+++ b/playground/src/components/entry-point/entry-point.spec.js
@@ -1,13 +1,56 @@
 import React from 'react';
-import { shallow } from 'enzyme';
+import { MC_API_PROXY_TARGETS } from '@commercetools-frontend/constants';
+import { renderAppWithRedux } from '@commercetools-frontend/application-shell/test-utils';
 import { ApplicationStateMachines } from './entry-point';
 
-describe('rendering', () => {
-  let wrapper;
-  beforeEach(() => {
-    wrapper = shallow(<ApplicationStateMachines />);
+const createStateMachinesListSdkMock = () => ({
+  action: {
+    type: 'SDK',
+    payload: {
+      method: 'GET',
+      service: 'states',
+      options: {
+        perPage: 25,
+        page: 1,
+      },
+      mcApiProxyTarget: MC_API_PROXY_TARGETS.COMMERCETOOLS_PLATFORM,
+    },
+  },
+  response: {
+    count: 25,
+    offset: 0,
+    total: 2,
+    results: [],
+  },
+});
+
+const render = (options) => {
+  return renderAppWithRedux(<ApplicationStateMachines />, {
+    sdkMocks: [createStateMachinesListSdkMock()],
+    permissions: {
+      canViewDeveloperSettings: true,
+      canManageDeveloperSettings: true,
+    },
+    ...options,
   });
-  it('should render state-machines route', () => {
-    expect(wrapper).toRender({ path: '/:projectKey/state-machines' });
+};
+
+describe('when route is /state-machines', () => {
+  let rendered;
+  it('should render state machines', async () => {
+    rendered = render({
+      route: '/project/state-machines',
+    });
+    await rendered.findByText(/State machines/i);
+  });
+});
+
+describe('when route is not /state-machines', () => {
+  let rendered;
+  it('should render catch all', async () => {
+    rendered = render({
+      route: '/project/xyz',
+    });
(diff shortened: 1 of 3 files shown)


playground/src/components/entry-point/entry-point.spec.js
```diff
@@ -1,13 +1,56 @@
...
+describe('when route is /state-machines', () => {
+  let rendered;
+  it('should render state machines', async () => {
+    rendered = render({
+      route: '/project/state-machines',
+    });
+    await rendered.findByText(/State machines/i);
   });
-  it('should render state-machines route', () => {
-    expect(wrapper).toRender({ path: '/:projectKey/state-machines' });
```
tdeekens, line 11: How dare you remove this meaningful test!

adnasa: my fears has come true (tomatoes on face)

tdeekens (approved):

playground/src/components/entry-point/entry-point.spec.js
```diff
@@ -1,13 +1,56 @@
...
+    count: 25,
+    offset: 0,
+    total: 2,
+    results: [],
+  },
+});
+
+const render = (options) => {
+  return renderAppWithRedux(<ApplicationStateMachines />, {
+    sdkMocks: [createStateMachinesListSdkMock()],
```
emmenko, line 29: Have you tried using msw?

emmenko (commented):
Nice 🙌

playground/src/components/entry-point/entry-point.spec.js
adnasa: ah good call.
actually no, I kept it minimal this time.
I can give it a try.

emmenko: I guess it's more of a question what best practices we want to teach our users.

For me something like msw is definitely easier to set up and use, compared to apollo/sdk mocks.

Together with our test-data initiative, I assume that this is the direction we are going to.

adnasa:
okay, so I gave `msw` a spin.

A little discovery however:
The `StateMachinesList` renders `StateMachinesListConnector -> Sdk.Get`.
in the background we call `sdk.get`, what happens then is

- the `sdk.get` is picked up by our redux `middleware` and tries to converts the action into a URI with the help of a `getProjectKey` util.
- The default definition for `getProjectKey` is [`selectProjectKeyFromUrl`](https://github.com/commercetools/merchant-center-application-kit/blob/master/packages/application-shell/src/utils/select-project-key-from-url/select-project-key-from-url.ts#L9:L9) where it picks the value from `window.location.pathname`

`window.location.pathname` is undefined in our tests. So we get an error when the middleware attempts to convert the action into a URI. We need to follow this up, where the `Sdk.Get` can succesfully build a request in tests.

Some approach I thought about:

1. in an SDK action, we pass a `getProjectKey` function that is forwarded `actionToUri` in our middleware
2. simply pass a `projectKey` in the SDK action (I believe we had this before)
3. mock the middleware (specifically `actionToUri`). Terrible idea, but I'm trying to turn all stones here.
4. *maybe* expose the [createTestMiddleware](https://github.com/commercetools/merchant-center-application-kit/blob/master/packages/application-shell/src/test-utils/test-utils.tsx#L494:L494) as a hook and internally uses a route hook.
**Assumption**: The route hook should be able to pick up the `projectKey` from the route context even in tests. If it does, we can forward a `getProjectKey` function

adnasa:
I created an issue to follow this up. I'm ready to merge this

tdeekens:
Thanks for investigating. That's really a bit unfortunate and surprising. For clarify for me: the apollo header-link uses the same function to determine the project key and pass it along as a header. Does it mean this header will also be null under test which wasn't noticed however as nothing throws? I would assume so, right?

adnasa:
if it is using `selectProjectKeyFromUrl` as well, then yes.
good catch

emmenko:
The apollo link uses that as a **fallback** (for the accounts app use case, which I don't remember why we need the project key there)

https://github.com/commercetools/merchant-center-application-kit/blob/c01697c142c94f5f16971738992aa2e1bd508849/packages/application-shell/src/apollo-links/header-link.ts#L52-L60

emmenko (approved):

Issue #1733, opened 2020-09-13
Speed up Custom Application deployment step on Firebase, Vercel & AWS


anilkk:
**Is your feature request related to a problem? Please describe.**

Currently, I need to create a `transformer-firebase.js` file manually and paste the code from docs.
Example: I would like to deploy **my starter custom application** on **Firebase**. 
I'm following the steps described in the [docs](https://docs.commercetools.com/custom-applications/deployment/example-deployment-firebase#generate-firebasejson-function-using-a-transformer-function).

**Steps** in the docs are create file  `./config/transformer-firebase.js` and copy code to this file.

Being a beginner of Custom Applications developer, I would like to get my first application up and running quickly, without wasting my time on manually creating files and copying the code. 

**Describe the solution you'd like**

It would be nice to have some `mc-scripts` to auto-generate the transformer-function.

`mc-scripts generate-transformer-function --vercel`
`mc-scripts generate-transformer-function --firebase`
`mc-scripts generate-transformer-function --aws`
 
Example: 
**Describe alternatives you've considered**

`mc-scripts generate-transformer-function-vercel`
`mc-scripts generate-transformer-function-firebase`
`mc-scripts generate-transformer-function-aws`

**Additional context**

As a Custom Application developer, I would like to get up and running my application quickly, without wasting or worrying time in the configs.

emmenko:
PS: @stephsprinkle you were mentioning some paint points regarding hosting and deployments of Custom Applications (as a general thing). Would something like this help easing that "pain" a bit?

stephsprinkle:
It's definitely a start! At least for Firebase and AWS, there are a bunch of steps after the transformer to actually getting the application deployed and I think that is where the majority of the "pain" is.

anilkk:
I noticed one more place to optimize. 
can we optimize these 3 steps to deploy into 1 step? 
[deployment vercel](https://docs.commercetools.com/custom-applications/deployment/example-deployment-vercel#deployment)

Expecting some script like this one.
`
mc-scripts deploy-to vercel
`
which basically runs all these three steps with just one script.  
`
$ yarn build
$ mc-scripts compile-html --use-local-assets --transformer $(pwd)/config/transformer-vercel.js
$ vercel
`

emmenko:
Hmm I wouldn't go that far, as then we're locking the user into our own opinionated configuration.

For now, I would only look into the "generation" of a transformer template file.

adnasa:
[...]

PR #1787, opened 2020-10-06
[Breaking] refactor: drop mc-http-server, make --use-local-assets the default behavior

emmenko:
Ref #1769 

TL;DR: The usage of the `--use-local-assets` flag is confusing and it turns out that this is what you need 99% of the time so we should make it the default behavior.
I also suggest to remove the `mc-http-server`. I don't see a reason to keep maintaining and having it. People can run `compile-html` first, then serve the static files however they would like to.


tdeekens (commented):
Thanks. Wondering if we should provide potential security fixes and e.g. dependency updates for the mc-http-server for a bit longer?

emmenko:
> Wondering if we should provide potential security fixes and e.g. dependency updates for the mc-http-server for a bit longer?

Hmm you mean in the Docker image?

tdeekens:
> Hmm you mean in the Docker image?

Yes, or is it pretty much shallow?

emmenko:
Btw, the current HTTP server version is actually broken (if you don't use the `--use-local-assets` and you use the Docker image). It's broken because the application config should be automatically loaded, but there is no application config in the Docker image, unless you mount your local project folder.

Given that we haven't had any bug report for the past months, I would assume that nobody is using that anyway.

tdeekens:
> Btw, the current HTTP server version is actually broken (if you don't use the `--use-local-assets` and you use the Docker image). It's broken because the application config should be automatically loaded, but there is no application config in the Docker image, unless you mount your local project folder.
> 
> Given that we haven't had any bug report for the past months, I would assume that nobody is using that anyway.

I don't think the conclusion of something broken should be to remove it entirely. I am open to remove the `mc-http-server` it it's not used widely but not using that argument. I would assume that most Custom Applications didn't migrate to the new application config. Would that be maybe a fair assumption why there are few bug tickets?

emmenko:
Yes and no. We can fix it (which is what I tried to do in the last hour) but doing so requires some (breaking) changes regardless (because of the way the application config is loaded). So I'd rather encourage people to use something else, or use another approach.

tdeekens (approved):
Okay then. The old versions used will continue to work if running. I guess we will educate people in the release notes how to migrate. Thanks.

adnasa (commented):
I'm open to removing the `mc-http-server`, just wasn't expecting to initiate this discussion over a PR.

If it is a breaking change regardless, then sure we can remove it.
We'll have to make sure to provide a migration path.

emmenko:
My initial intention was not to remove it obviously. But by looking at what needs to be done it's just not really worth the effort. Hence the suggestion that we simply remove the package.

We'll provide proper documentation to migrate of course.

adnasa (approved):

application-templates/starter/package.json
```diff
@@ -6,7 +6,8 @@
   "scripts": {
     "build": "mc-scripts build",
     "start": "dotenv -- mc-scripts start",
-    "start:prod:local": "NODE_ENV=production MC_APP_ENV=development dotenv -- mc-http-server --use-local-assets",
+    "compile-html": "NODE_ENV=production dotenv -- mc-scripts compile-html",
+    "start:prod:local": "NODE_ENV=production MC_APP_ENV=development dotenv -- mc-scripts compile-html && NODE_ENV=production npx serve -l 3001 public",
```
emmenko, line 10: To run it locally in production mode, we can do it like this: compile it first, then serve the static file using `serve` (or any other static server CLI).

application-templates/starter/package.json
```diff
@@ -6,7 +6,8 @@
   "scripts": {
     "build": "mc-scripts build",
     "start": "dotenv -- mc-scripts start",
-    "start:prod:local": "NODE_ENV=production MC_APP_ENV=development dotenv -- mc-http-server --use-local-assets",
+    "compile-html": "NODE_ENV=production dotenv -- mc-scripts compile-html",
```
emmenko, line 9: This is just added for convenience.

application-templates/starter/package.json
```diff
@@ -6,7 +6,9 @@
...
+    "compile-html:local": "NODE_ENV=production MC_APP_ENV=development dotenv -- mc-scripts compile-html --transformer @commercetools-frontend/mc-dev-authentication/transformer-local.js",
```
emmenko, line 10: This is turning out to be pretty nice. The `mc-dev-authentication` package now implements a `transformer-local.js` that compiles and copies the login/logout routes to the `public` folder of the application.

This way, everything is still static and we can start the app in production mode locally.

emmenko: What is also nice is that we can reference npm packages/modules.

```
--transformer @commercetools-frontend/mc-dev-authentication/transformer-local.js
```

emmenko: This is also a bit related to #1733.
We can have a package with pre-configured transformers, which people can also simply copy or extend.

website/src/releases/2020-10/migrating-to-v17.mdx
```diff
@@ -1,19 +1,16 @@
 ---
 date: 2020-10-12
-title: Migrating to the new Apollo Client 3.0
+title: Migrating to v17
```
emmenko, line 3: I've renamed the release note, as it's about migrating to v17 now (which of course includes the apollo v3 migration).

tdeekens (approved):
Really nice. The transform local is nice. Feels like it removes concepts and keeps things more concise. 

Maybe we should add a note to the application config page that it doesn’t work with the http-server which was deprecated in v17. Just a thought.

.changeset/poor-avocados-act.md
```diff
@@ -0,0 +1,30 @@
...
+You can decide to [deploy the Custom Application statically to one of the popular cloud providers](https://docs.commercetools.com/custom-applications/deployment/compiling-a-custom-application#deployment), or serve the files on your own using a static server.
+
+For example, to run locally the Custom Application using the production bundles:
+
+```console
+NODE_ENV=production MC_APP_ENV=development dotenv -- \
+  mc-scripts compile-html \
+  --transformer @commercetools-frontend/mc-dev-authentication/transformer-local.js
+
+mc-scripts serve
```
emmenko, line 29: I forgot that we need to serve the files as a SPA.
The `serve` package has a `--single` option that does that but then we loose the routes for the login/logout pages.

Therefore I decided to add a new `serve` command to `mc-scripts`, so that we can customize the rewrite routes.
A `serve` command is something that is also used in other tools, like Gatsby.

In our case the `mc-scripts serve` command starts an HTTP server to serve the files in the `public` folder. The main use case for this is to start the Custom Application locally to test the production bundles.

packages/mc-scripts/src/serve.js
```diff
@@ -0,0 +1,69 @@
...
+
+const appDirectory = fs.realpathSync(process.cwd());
+const resolveApp = (relativePath) => path.resolve(appDirectory, relativePath);
+
+const publicAssetsPath = resolveApp('public');
+
+const server = http.createServer((request, response) => {
+  // You pass two more arguments for config and middleware
+  // More details here: https://github.com/vercel/serve-handler#options
+  return serveHandler(request, response, {
```
tdeekens, line 40: 👍

PR #1794, opened 2020-10-09
feat(app-shell): to add hideMenuItems property

tdeekens:
#### Summary

This pull request proposes to add a `hideMenuItems` prop to the `ApplicationShell`.

#### Description

Each application currently has a non-documented feature which allows hiding menu items based on the `ctx.environment.disabledMenuItems`. We sometimes use this property internally. However, we would like to move this property to be controlled by our APIs instead of the environment to have better control over it.

The result would be to remove the `ctx.environment.disabledMenuItems` and have a non-documented `hideMenuItems` on the Application Shell. We could then feed this prop from our API.

We could also attempt to remove the `ctx.environment.disabledMenuItems` with this if our APIs and applications are already prepared for it. Open for discussions here.

### DIFF
--- a/packages/application-shell/src/components/application-shell/application-shell.tsx
+++ b/packages/application-shell/src/components/application-shell/application-shell.tsx
@@ -78,6 +78,7 @@ type Props<AdditionalEnvironmentProperties extends {}> = {
   // Use `trackingEventList` instead.
   trackingEventWhitelist?: TrackingList;
   trackingEventList?: TrackingList;
+  hideMenuItems?: string[];
   applicationMessages: TAsyncLocaleDataProps['applicationMessages'];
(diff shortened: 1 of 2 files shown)


emmenko:
I'm wondering if we can achieve the same by fetching this information from our API within our MC proxy, then attach this information in the graphql response that loads the applications menu links.

tdeekens:
> I'm wondering if we can achieve the same by fetching this information from our API within our MC proxy, then attach this information in the graphql response that loads the applications menu links.

Yes, that's another option true. A challenge however would be that the `alLFeatures` needs an `mcAccessToken`. From what I recall our proxy doesn't send it.

tdeekens:
We will move this into our infrastructure instead.

PR #1801, opened 2020-10-12
feat(app-shell): send graphql operation name as HTTP header, for debugging purposes

emmenko:


emmenko:
Putting as blocked as we need to allow this new header in the CORS config.

tdeekens (approved):

tdeekens:
We need to allow list the header on the MC API too. /cc @YahiaElTai as you were interested in this.

adnasa (approved):

PR #1809, opened 2020-10-19
fix(app-shell): default cache policy for apps menu

emmenko:


packages/application-shell/src/configure-apollo.ts
```diff
@@ -59,6 +61,27 @@ const link = from([
   httpLink,
 ]);
 
+// This custom merge function allows to merge two arrays of objects.
+// The incoming list is what we need to update to, but we still need
+// to ensure that existing cache elements are not removed but updated.
+// This is usually the case when elements get removed.
+const mergeArraysObjects: FieldMergeFunction = (
```
tdeekens, line 68: Should we expose this from app-kit to re-use it in apps or rather not?

emmenko: Not for now, at least until we are sure it does what we need.

tdeekens (approved):

PR #1811, opened 2020-10-20
fix(sentry): to use beforeSend to omit some data

tdeekens:
#### Summary

This pull request suggests to use the `beforeSend` function from Sentry to omit props from events. We already have very strict PII rules in Sentry itself but can take it one step further I think. Allowing maybe on the server to be more bit more permissive again.

1. https://docs.sentry.io/platforms/javascript/configuration/filtering/
2. https://github.com/odynvolk/omit-deep-lodash/blob/master/src/index.js#L25

### DIFF
--- a/packages/sentry/src/sentry.spec.ts
+++ b/packages/sentry/src/sentry.spec.ts
@@ -1,7 +1,7 @@
 import * as Sentry from '@sentry/browser';
 import sentryTestkit from 'sentry-testkit';
 import waitForExpect from 'wait-for-expect';
-import { reportErrorToSentry } from './sentry';
+import { reportErrorToSentry, removeUnsafeFields } from './sentry';
 
 const { testkit, sentryTransport } = sentryTestkit();
 
@@ -29,7 +29,7 @@ class PromiseRejectionEvent extends Event {
   }
 }
 
-describe('reportErrorToSentry', () => {
+describe('reporting to sentry', () => {
   let error;
 
   beforeAll(() => {
@@ -159,5 +159,39 @@ describe('reportErrorToSentry', () => {
         message: '{"message":"something went wrong"}',
       });
     });
+  });
+});
+
+describe('when using unsafe fields', () => {
+  it('should redact unsafe fields', () => {
+    let safeFields = removeUnsafeFields({
+      user: {
+        email: 'john@doeland.com',
+      },
+    });
+
+    expect(safeFields.user?.email).toEqual('[Redacted]');
+
+    safeFields = removeUnsafeFields({
+      extra: {
+        action: {
+          payload: {
+            firstName: 'John',
+          },
+        },
+      },
+    });
+
+    // @ts-expect-error
+    expect(safeFields.extra?.action.payload.firstName).toEqual('[Redacted]');
+
+    safeFields = removeUnsafeFields({
+      extra: {
+        actions: [{ lastName: 'Doe' }],
+      },
+    });
+
+    // @ts-expect-error
+    expect(safeFields.extra?.actions[0].lastName).toEqual('[Redacted]');
   });
 });
--- a/packages/sentry/src/sentry.spec.ts
+++ b/packages/sentry/src/sentry.spec.ts
@@ -1,7 +1,7 @@
 import * as Sentry from '@sentry/browser';
 import sentryTestkit from 'sentry-testkit';
 import waitForExpect from 'wait-for-expect';
-import { reportErrorToSentry, removeUnsafeFields } from './sentry';
+import { reportErrorToSentry, redactUnsafeEventFields } from './sentry';
 
 const { testkit, sentryTransport } = sentryTestkit();
 
@@ -164,7 +164,15 @@ describe('reporting to sentry', () => {
 
 describe('when using unsafe fields', () => {
   it('should redact unsafe fields', () => {
-    let safeFields = removeUnsafeFields({
+    type EventExtra = {
+      payload?: {
+        firstName: string;
+      };
+      actions?: Array<{
+        lastName: string;
+      }>;
+    };
+    let safeFields = redactUnsafeEventFields({
       user: {
         email: 'john@doeland.com',
       },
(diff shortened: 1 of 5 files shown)


packages/sentry/src/sentry.ts
```diff
@@ -38,6 +40,10 @@ export const boot = () => {
           onerror: false,
         }),
       ],
+      beforeSend(event) {
+        // @ts-expect-error
```
tdeekens, line 44: The types seem wrong. Looking at 

1. https://github.com/odynvolk/omit-deep-lodash/blob/master/src/index.js#L25

We can see that it's a dynamic arguments processing given we have more than two we silce.

The types do not respect that with `declare function omitDeep(input: object, props: string): object;`. Either we:

1. Path upstream (takes time)
2. Locally type
3. Know about the error and opt out

emmenko: I'm wondering if we should just implement this function on our own, the implementation is just a bunch a lines.
Then we also don't have the issue with the types.

emmenko (commented):
Thanks. Did you try this locally to see how an event gets patched?
Also, should we maybe still send the fields but having them redacted?

packages/sentry/src/sentry.ts
tdeekens: Did you have something like 3b9b6b1 in mind?

packages/sentry/src/sentry.ts
```diff
@@ -21,6 +22,38 @@ const sendErrorToSentry = (error: Reportable) => {
...
+      const hasPropReplacement = Object.keys(replacements).includes(prop);
+
+      switch (typeof event[prop]) {
+        case 'string':
+          if (hasPropReplacement) {
+            event[prop] = replacements[prop];
+          }
+
+          break;
+        case 'object':
```
emmenko, line 44: What about arrays?

emmenko (commented):
Yes something like this, thanks! 👌

Can you also add a test?

packages/sentry/src/sentry.ts
tdeekens: I'd say `[Redacted]`. Not like first element this, second element that.

emmenko: No I mean does the recursion work also for arrays or nested array objects?

tdeekens: Yes, arrays are objects. I added a test for the case.

packages/sentry/src/sentry.spec.ts
```diff
@@ -161,3 +161,37 @@ describe('reportErrorToSentry', () => {
...
+      extra: {
+        action: {
+          payload: {
+            firstName: 'John',
+          },
+        },
+      },
+    });
+
+    // @ts-expect-error
```
tdeekens, line 185: The `extra` is unknown. I am not sure how I would get around the error `Object is of type 'unknown'` here without jumping through hoops.

packages/sentry/src/sentry.ts
```diff
@@ -23,35 +23,47 @@ const sendErrorToSentry = (error: Reportable) => {
...
 
-const replaceEventValues = (
-  event: Event,
-  replacements: TReplacements
-): Event => {
-  for (const prop in event) {
-    if (event.hasOwnProperty(prop)) {
+const replaceValues = <T>(source: T, replacements: TReplacements): T => {
+  const replaceValue = (prop: string) => {
+    // @ts-expect-error
```
tdeekens, line 31: This is flat out fail type to avoid errors. Maybe somebody can have a look.

packages/sentry/src/sentry.spec.ts
```diff
@@ -161,3 +161,37 @@ describe('reportErrorToSentry', () => {
     });
   });
 });
+
+describe('when using unsafe fields', () => {
+  it('should redact unsafe fields', () => {
+    let safeFields = removeUnsafeFields({
```
tdeekens, line 167: You can't actually test this through testkit as it doesn't boot sentry it seems. At least `beforeSend` is never invoked. So I renamed things and code to make sense (hopefully).

packages/sentry/src/sentry.ts
```diff
@@ -21,6 +22,50 @@ const sendErrorToSentry = (error: Reportable) => {
   return Sentry.captureException(errorToCapture);
 };
 
+type TReplacements = {
+  [name: string]: string;
+};
+
+const replaceValues = <T>(source: T, replacements: TReplacements): T => {
```
emmenko, line 29: I don't think we need a generic here, we can declare `source` to be an `Event`. Maybe then it's easier to deal with the type errors?

packages/sentry/src/sentry.spec.ts
emmenko: I think we can just cast it

```ts
(safeFields.extra as MyType)?.action.payload.firstName
```

emmenko (commented):
Thanks for putting the extra effort 🙏

packages/sentry/src/sentry.ts
tdeekens: Nope. It won't simplify things. It it's the typing of Event which makes things hard as it's a lot of optionals and unknowns.

packages/sentry/src/sentry.spec.ts
```diff
@@ -164,34 +164,42 @@ describe('reporting to sentry', () => {
 
 describe('when using unsafe fields', () => {
   it('should redact unsafe fields', () => {
-    let safeFields = removeUnsafeFields({
+    type EventExtra = {
```
tdeekens, line 167: This works to cast. Still fells odd.

packages/sentry/src/sentry.ts
```diff
@@ -39,15 +42,15 @@ const replaceValues = <T>(source: T, replacements: TReplacements): T => {
       switch (typeof source[prop]) {
```
tdeekens, line 42: Removing the generic caused `No index signature with a parameter of type 'string' was found on type 'Event'.` which I don't fully understand as event has string as keys.

emmenko (approved):
💯 Thanks again!

PR #1812, opened 2020-10-21
feat(components/modal-pages): allow use of iconLeft on secondaryButton

dogayuksel:
#### Summary

In discount application's rule builder quick selection context we use the secondary button of modal pages to trigger 'revert changes'. Based on the decision by the ui-team, revert changes button should have the corresponding revert icon. These changes allow the use of an icon with the secondary button of the modal pages.

### DIFF
--- a/.changeset/mighty-ears-develop.md
+++ b/.changeset/mighty-ears-develop.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-components': minor
+---
+
+Allow iconLeft use on secondaryButton of modal-pages application component
--- a/packages/application-components/src/components/modal-pages/internals/default-form-buttons.tsx
+++ b/packages/application-components/src/components/modal-pages/internals/default-form-buttons.tsx
@@ -20,6 +20,7 @@ type Props = {
   label: Label;
   onClick: (event: React.SyntheticEvent) => void;
   isDisabled: boolean;
+  iconLeft?: React.ReactNode;
   dataAttributes: { [key: string]: string };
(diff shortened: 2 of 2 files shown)


emmenko (changes requested):
Thanks for the contribution.

First of all, it seems to me that this change is not complete. How do you want to expose this prop to the main components? You just added the prop to an internal component, so I'm a bit confused how this should be used.

Also, I would suggest to at least open an issue first, describing what feature/changes you would like to implement, as we need to make a decision first how to proceed.
For instance, the requirements here are not so clear, meaning for example if this is the only use case, how much we want to build as defaults into the components (also from a UX perspective), etc.

dogayuksel:
> First of all, it seems to me that this change is not complete. How do you want to expose this prop to the main components? You just added the prop to an internal component, so I'm a bit confused how this should be used. 

Hmm, i was working with `CustomFormModalPage` which directly exposes the buttons from internals. When you use it, you pass in the buttons to be renderd to "formControls". I guess you would also expect these changes do be added to `FormModalPage` which decorates the buttons itself, hence needs an update of props.

> Also, I would suggest to at least open an issue first, describing what feature/changes you would like to implement, as we need to make a decision first how to proceed.
> For instance, the requirements here are not so clear, meaning for example if this is the only use case, how much we want to build as defaults into the components (also from a UX perspective), etc.

I opened an issue as you suggested, added the details from the ui-team's findings and the dependent issue.
https://github.com/commercetools/merchant-center-application-kit/issues/1813

PR #1815, opened 2020-10-22
fix(test-utils): pass the apollo cache to the mocked provider

emmenko:
See https://www.apollographql.com/docs/react/development-testing/testing/#a-note-on-fragment-usage-typepolicies-and-possibletypes


tdeekens (approved):

emmenko:
Please have another look, this is a bit more complicated than it was initially.

TL;DR: the `addTypename` option in the test-utils has been removed, and it's inferred by the Apollo cache.
I don't consider this a breaking change but more of a leftover fix from the v17 release.

tdeekens (approved):

adnasa (approved):

PR #1816, opened 2020-10-22
fix(quick access): wrong navigation when using product commands

ahmehri:
#### Summary

Fix wrong navigation when using Quick Access Product commands.

#### Description

Fix wrong navigation when using Quick Access Product commands.

- Using `Open Product List` and `Open Products` commands navigate the user to the old product list. Using them now navigate the user to the PIM list instead.
- Using `Open Variant` command navigates the user to a "We could not find this Project" page because of the malformed URL. Using it now navigates the user to the variant details page.

### DIFF
--- a/.changeset/violet-fireants-hope.md
+++ b/.changeset/violet-fireants-hope.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': minor
+---
+
+Fix wrong navigation when using Quick Access Product commands
--- a/packages/application-shell/src/components/quick-access/create-commands.ts
+++ b/packages/application-shell/src/components/quick-access/create-commands.ts
@@ -57,7 +57,7 @@ const createCommands = ({
         keywords: ['Go to Products'],
         action: {
           type: actionTypes.go,
-          to: `/${applicationContext.project.key}/products`,
+          to: `/${applicationContext.project.key}/products/pim-search`,
         },
         subCommands: [
           hasSomePermissions(
@@ -68,7 +68,7 @@ const createCommands = ({
             text: intl.formatMessage(messages.openProductList),
             action: {
               type: actionTypes.go,
-              to: `/${applicationContext.project.key}/products`,
+              to: `/${applicationContext.project.key}/products/pim-search`,
             },
           },
           hasSomePermissions(
--- a/packages/application-shell/src/components/quick-access/quick-access.spec.js
+++ b/packages/application-shell/src/components/quick-access/quick-access.spec.js
@@ -705,7 +705,7 @@ describe('QuickAccess', () => {
     fireEvent.keyUp(searchInput, { key: 'Enter' });
     await waitFor(() => {
       expect(rendered.history.location.pathname).toBe(
-        '/test-with-big-data/products'
+        '/test-with-big-data/products/pim-search'
       );
     });
     // should close quick access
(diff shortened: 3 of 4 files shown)


.changeset/violet-fireants-hope.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': minor
```
emmenko, line 2:
```suggestion
'@commercetools-frontend/application-shell': patch
```

packages/application-shell/src/components/quick-access/create-commands.ts
```diff
@@ -57,7 +57,7 @@ const createCommands = ({
         keywords: ['Go to Products'],
         action: {
           type: actionTypes.go,
-          to: `/${applicationContext.project.key}/products`,
+          to: `/${applicationContext.project.key}/products/pim-search`,
```
emmenko, line 60: Shouldn't the products app have a redirect `products` -> `products/pim-search`? 🤔

tdeekens: Once the old list is removed yes. That was always my understanding. I always thought we have to alterantives when the old list is removed

1. Redirect from `products` to `products/pim-search`
2. Make `products/pim-search` the default instead of `products`

I somewhat like 2. as the routing is cleaner which would mean we wouldn't change anything here. If we do we'd have to change it back again.

ahmehri: Yes, and we will do that, except not now because we still didn't remove the old product list from the application, once we do that, we will rewrite the URLs and add the redirect. Meanwhile, we're doing a quick fix here that won't need to be changed later thanks to the redirect.

emmenko: Ok

emmenko (approved):

packages/application-shell/src/components/quick-access/create-commands.ts
ahmehri: > I somewhat like 2. as the routing is cleaner which would mean we wouldn't change anything here. If we do we'd have to change it back again.

We will go with 1. because `products/pim-search` was meant to be a temporary route until the PIM list feature is released, and `products` is more clean and straightforward name. In addition, we will create a redirect from `products/pim-search` to `products` (once we bind the PIM list to `products`) so that users used to `products/pim-search` won't be lost. And thanks to this redirect we won't need to change the fix implemented in this PR.

tdeekens: Ah, I think you mean by 1. what I meant by 2. I prefer the way you describe. So numbers don't matter :)

tdeekens (approved):

ahmehri:
The build failed, not sure what to do.

tdeekens:
I restarted it. If it fails again maybe rebase over `origin/master`.

ahmehri:
Rebasing didn't work, it's an infrastructure issue.

```
Error: Path does not exist /home/runner/work/merchant-center-application-kit/merchant-center-application-kit/cypress/videos
Error: Exit code 1 returned from process: file name '/home/runner/runners/2.273.5/bin/Runner.PluginHost', arguments 'action "GitHub.Runner.Plugins.Artifact.PublishArtifact, Runner.Plugins"'.
```

emmenko:
No, this is the error

emmenko:
You have to scroll up a bit in the CI logs.

ahmehri:
I fixed the issue and now there is one visual change, not related to my changes, that needs to be reviewed.

tdeekens:
Was just an SVG. Looks good.

PR #1818, opened 2020-10-24
feat(react-scripts): add support for experimental Fast Refresh

emmenko:
Now that `react-scripts` v4 is out, we can sync our custom `mc-scripts` implementation and add support for Fast Refresh.

PS: the new JSX transform can be done as a follow up.


tdeekens (approved):
Looks cool. Have you tried it?

I agree the JSX transform can be a follow up. We prolly habe to make it opt-in to support "older" versions of React which didn't get the back port. Maybe also via an env variable or babel config value.

emmenko:
I haven't fully tested it yet, just had some little time to make the PR 🙂

jonnybel (approved):

PR #1819, opened 2020-10-24
fix(babel): allow changing jsx runtime in babel

tdeekens:
#### Summary

This pull request adds a `JSX_RUNTIME` environment variable which can be used to set the JSX runtime to e.g. `automatic`.

The JSX runtime is configured in the mc-babel-preset through the `preset-react`. It would always continue to default to `classic`. However, this way we'd allow experimentation with the `automatic` runtime. Which is not yet supported in TypeScript (starting 4.1). 

The new runtime doesn't require React to be in scope. This means we could potentially internally already experiment with it.

There are other ways to allow the configuration. We could also pass an `jsxRuntime` argument through our webpack configs and jest config to the babel preset. However, that seems far more invasive and also at times not possible as e.g. a webpack config just does

```js
presets: [
  require.resolve('@commercetools-frontend/babel-preset-mc-app'),
]
```

While a jest config does 

```js
transform: {
  '^.+\\.js$': resolveRelativePath('./transform-babel-jest.js'),
  '^.+\\.graphql$': 'jest-transform-graphql',
},
```

which in all I fear makes argument passing hard. So the env variable might be an easy middle ground and could be eventually also default to `automatic` or be removed entirely again.

I would leave it as a `patch` change for now and not document it until we tested it internally.

[1]: https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

### DIFF
--- a/.changeset/plenty-hotels-suffer.md
+++ b/.changeset/plenty-hotels-suffer.md
@@ -9,7 +9,7 @@ feat: adds the ability to opt into the new `automatic` JSX runtime
 
 Recent React versions support a new JSX runtime. Read more about it [here](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html). 
 
-You may opt into the new runtime in `automatic` mode by setting the `ENABLE_NEW_JSX_TRANSFORM` environment variable. Please note you need at least React v17 or v16.14 in your application. 
+You may opt into the new runtime in `automatic` mode by setting the `ENABLE_NEW_JSX_TRANSFORM` environment variable to `true`. Please note you need at least React v17 or v16.14 in your application. 
 Opting into the new JSX transform automatically also changes the Babel, Jest and ESLint configurations. As a conssequence ESLint will warn whenever it discovers React being in scope by importing it as `import React from 'react`'. You have to remove those imports using the respective codemod by running `npx react-codemod update-react-imports`.
 
 Lastly, all code of the Merchant Center Application Kit will continue to be bundled in `classic` mode to support older versions of React.
--- a/.changeset/plenty-hotels-suffer.md
+++ b/.changeset/plenty-hotels-suffer.md
@@ -1,0 +1,15 @@
+---
+"@commercetools-frontend/babel-preset-mc-app": minor
+"@commercetools-frontend/eslint-config-mc-app": minor
+"@commercetools-frontend/jest-preset-mc-app": minor
+"@commercetools-frontend/mc-scripts": minor
+---
+
+feat: adds the ability to opt into the new `automatic` JSX runtime
+
+Recent React versions support a new JSX runtime. Read more about it [here](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html). 
+
+You may opt into the new runtime in `automatic` mode by setting the `ENABLE_NEW_JSX_TRANSFORM` environment variable. Please note you need at least React v17 or v16.14 in your application. 
+Opting into the new JSX transform automatically also changes the Babel, Jest and ESLint configurations. As a conssequence ESLint will warn whenever it discovers React being in scope by importing it as `import React from 'react`'. You have to remove those imports using the respective codemod by running `npx react-codemod update-react-imports`.
+
+Lastly, all code of the Merchant Center Application Kit will continue to be bundled in `classic` mode to support older versions of React.
--- a/.changeset/plenty-hotels-suffer.md
+++ b/.changeset/plenty-hotels-suffer.md
@@ -10,6 +10,6 @@ feat: adds the ability to opt into the new `automatic` JSX runtime
 Recent React versions support a new JSX runtime. Read more about it [here](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html). 
 
 You may opt into the new runtime in `automatic` mode by setting the `ENABLE_NEW_JSX_TRANSFORM` environment variable to `true`. Please note you need at least React v17 or v16.14 in your application. 
-Opting into the new JSX transform automatically also changes the Babel, Jest and ESLint configurations. As a conssequence ESLint will warn whenever it discovers React being in scope by importing it as `import React from 'react`'. You have to remove those imports using the respective codemod by running `npx react-codemod update-react-imports`.
+Opting into the new JSX transform automatically also changes the Babel, Jest and ESLint configurations. As a consequence ESLint will warn whenever it discovers React being in scope by importing it as `import React from 'react`'. You have to remove those imports using the respective codemod by running `npx react-codemod update-react-imports`.
 
 Lastly, all code of the Merchant Center Application Kit will continue to be bundled in `classic` mode to support older versions of React.
--- a/packages/babel-preset-mc-app/index.js
+++ b/packages/babel-preset-mc-app/index.js
@@ -1,5 +1,5 @@
 /* eslint-disable global-require */
-module.exports = function getBabePresetConfigForMcApp(api, opts) {
+module.exports = function getBabePresetConfigForMcApp(api, opts = {}) {
   // This is similar to how `env` works in Babel:
   // https://babeljs.io/docs/usage/babelrc/#env-option
   // We are not using `env` because it’s ignored in versions > babel-core@6.10.4:
--- a/packages/babel-preset-mc-app/index.js
+++ b/packages/babel-preset-mc-app/index.js
@@ -9,6 +9,7 @@ module.exports = function getBabePresetConfigForMcApp() {
   const env = process.env.BABEL_ENV || process.env.NODE_ENV;
   const isEnvDevelopment = env === 'development';
   const isEnvProduction = env === 'production';
+  const jsxRuntime = process.env.JSX_RUNTIME || 'classic';
   const isEnvTest = env === 'test';
   const isRollup = process.env.BUILD_ROLLUP === true;
 
@@ -60,6 +61,9 @@ module.exports = function getBabePresetConfigForMcApp() {
           // Will use the native built-in instead of trying to polyfill
           // behavior for any plugins that require one.
           useBuiltIns: true,
+          // Allows changing the JSX runtime.
+          // https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html
+          runtime: jsxRuntime,
         },
       ],
       [
--- a/packages/babel-preset-mc-app/index.js
+++ b/packages/babel-preset-mc-app/index.js
@@ -1,5 +1,5 @@
 /* eslint-disable global-require */
-module.exports = function getBabePresetConfigForMcApp() {
+module.exports = function getBabePresetConfigForMcApp(api, opts) {
   // This is similar to how `env` works in Babel:
   // https://babeljs.io/docs/usage/babelrc/#env-option
   // We are not using `env` because it’s ignored in versions > babel-core@6.10.4:
@@ -9,7 +9,6 @@ module.exports = function getBabePresetConfigForMcApp() {
   const env = process.env.BABEL_ENV || process.env.NODE_ENV;
   const isEnvDevelopment = env === 'development';
   const isEnvProduction = env === 'production';
-  const jsxRuntime = process.env.JSX_RUNTIME || 'classic';
   const isEnvTest = env === 'test';
   const isRollup = process.env.BUILD_ROLLUP === true;
 
@@ -60,10 +59,8 @@ module.exports = function getBabePresetConfigForMcApp() {
           development: isEnvDevelopment || isEnvTest,
           // Will use the native built-in instead of trying to polyfill
           // behavior for any plugins that require one.
-          ...(jsxRuntime !== 'automatic' ? { useBuiltIns: true } : {}),
-          // Allows changing the JSX runtime.
-          // https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html
-          runtime: jsxRuntime,
+          ...(opts.runtime !== 'automatic' ? { useBuiltIns: true } : {}),
+          runtime: opts.runtime || 'classic',
         },
       ],
       [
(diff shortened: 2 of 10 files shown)


packages/babel-preset-mc-app/index.js
```diff
@@ -9,6 +9,7 @@ module.exports = function getBabePresetConfigForMcApp() {
   const env = process.env.BABEL_ENV || process.env.NODE_ENV;
   const isEnvDevelopment = env === 'development';
   const isEnvProduction = env === 'production';
+  const jsxRuntime = process.JSX_RUNTIME || 'classic';
```
emmenko, line 12:
```suggestion
  const jsxRuntime = process.env.JSX_RUNTIME || 'classic';
```

packages/babel-preset-mc-app/index.js
```diff
@@ -60,6 +61,9 @@ module.exports = function getBabePresetConfigForMcApp() {
           // Will use the native built-in instead of trying to polyfill
           // behavior for any plugins that require one.
           useBuiltIns: true,
+          // Allows changing the JSX runtime.
+          // https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html
+          runtime: jsxRuntime,
```
emmenko, line 66: In react-scripts, they define `useBuiltIns` only when NOT using the runtime: https://github.com/facebook/create-react-app/blob/ed958938f642007645dd5ac3466db36202f8754e/packages/babel-preset-react-app/create.js#L98-L99

emmenko (commented):
This is how CRA configures the jest transformer: https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/config/jest/babelTransform.js

And for webpack they pass it as a babel option from webpack: https://github.com/facebook/create-react-app/blob/ed958938f642007645dd5ac3466db36202f8754e/packages/react-scripts/config/webpack.config.js#L413-L418

Unless you see other problem, I think we should try the same approach. What do you think?

packages/babel-preset-mc-app/index.js
tdeekens: Why on 🌎 did I remove the `env` before commiting 🤷

packages/babel-preset-mc-app/index.js
tdeekens: Thanks. Will add!

tdeekens:
Thanks for sharing the CRA approach. Also a good option. I felt we can silently allow opting in. Keeping most on the `classic` runtime for now. Their approach seems more like enabling it by default (checking react) while allowing an opt-out. I the transform was recommended to be used more carefully. If CRA goes with a "smart" default `automatic` we can do that too.

Let me commit that in one commit. Then we can compare approaches and decide.

packages/babel-preset-mc-app/index.js
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable global-require */
-module.exports = function getBabePresetConfigForMcApp() {
+module.exports = function getBabePresetConfigForMcApp(api, opts) {
```
tdeekens, line 2: We don't use `env`. The `opts` is what comes from e.g. webpack. Maybe we can use `env` in the future.

packages/jest-preset-mc-app/has-jsx-runtime.js
```diff
@@ -0,0 +1,14 @@
+function hasJsxRuntime() {
```
tdeekens, line 1: Just liked to have this separately. We have it once in the for webpack and once for babel as CRA.

packages/jest-preset-mc-app/transform-babel-jest.js
```diff
@@ -1,7 +1,10 @@
 const getBabePresetConfigForMcApp = require('@commercetools-frontend/babel-preset-mc-app');
 const getJestBabelPreset = require('babel-preset-jest');
+const hasJsxRuntime = require('./has-jsx-runtime');
 
-const mcAppBabelConfig = getBabePresetConfigForMcApp();
+const mcAppBabelConfig = getBabePresetConfigForMcApp(null, {
```
tdeekens, line 6: We spread in the `babelJestConfig` we we might have to do it this way here if we don't want to change that.

tdeekens (commented):
The last commits contain the change to have it dynamic while allowing the opt-out flag via `DISABLE_NEW_JSX_TRANSFORM`.

rollup.config.js
```diff
@@ -28,7 +28,9 @@ const { packageJson: pkg } = readPkgUp.sync({
 });
 const [, packageName] = pkg.name.split('@commercetools-frontend/');
 const extensions = ['.js', '.ts', '.tsx'];
-const babelOptions = getBabelPreset();
+const babelOptions = getBabelPreset(null, {
+  runtime: 'automatic',
```
tdeekens, line 32: I am not super sure here. This was sort of the initial reason for the soft opt-in. We could also bundle with `classic`. This would assume that any newly published version runs with a recent React version I'd assume.

tdeekens: The last CI run is interesting. It means you cannot have automatic and import React. Either we bundle classic but I am not sure if that would break consumers using automatic then again.

emmenko (commented):
What should we do with Rollup, meaning how should we bundle the appkit packages? 

Also, if we remove the React imports in the appkit repo, we would need first to update to TS 4.1 (https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#typescript).

tdeekens:
> What should we do with Rollup, meaning how should we bundle the appkit packages?
> 
> Also, if we remove the React imports in the appkit repo, we would need first to update to TS 4.1 (https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#typescript).

Yes, that is the interesting part to me too. We have to bundle classic until 4.1. After we can bundle with automatic. All under the assumption that a lib bundled with classic works with an app bundling with automatic. That is an assumption for me so far I am not certain.

packages/jest-preset-mc-app/has-jsx-runtime.js
```diff
@@ -0,0 +1,14 @@
...
+  if (process.env.DISABLE_NEW_JSX_TRANSFORM === 'true') {
```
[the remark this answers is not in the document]
tdeekens: Having it opt-in was my initial point of having the JSX_TRANSFORM env variable which defaulted to classic. We then wanted to do what CRA does which is opt-out. What do you prefer. If we opt-out we basically are almost at the initial step again apart from the variable name and how it is passed down.

tdeekens:
Memo to self: check if we can also automatically (dynamic) update ESLint rules or if consumers have to disable then manually for now.

packages/jest-preset-mc-app/has-jsx-runtime.js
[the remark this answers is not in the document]
emmenko: Yes I mean to follow the same implementation approach, but then "invert" the env variable value basically (if we want to have make it opt-in).

emmenko:
> This implies to me that both can work side by side. So we can bundle classic until TS 4.1 and our apps could use automatic already.

Yeah this would be my assumption as well, from what I understood.

So to recap:
- Rollup uses `classic` for now, at least until we can upgrade to TS 4.1
- Babel, Webpack, and Jest use `classic` as the default, so applications can continue to work with the normal transform
- Applications can decide to opt into the new JSX transform by passing an environment variable, so that application code (not libraries) can be written without React in scope.
  - Do we need to migrate all the application code at once or can we do it step-by-step?

emmenko:
> check if we can also automatically (dynamic) update ESLint rules

Good point. CRA does it in Webpack because they have the eslint config configured there: https://github.com/facebook/create-react-app/blob/e63de79cc4530efc7402e38923d875f27cdef0c3/packages/react-scripts/config/webpack.config.js#L760-L763

tdeekens:
> - Rollup uses `classic` for now, at least until we can upgrade to TS 4.1

Yes. Anything built with classic runs in an app using automatic as it should be backwards compatible and createElement goes nowhere. 

> - Babel, Webpack, and Jest use `classic` as the default, so applications can continue to work with the normal transform

Yes, we should make it opt-in instead of opt-out (as now). So ENABLE over DISABLE. Otherwise it would be breaking to apps as they would habe to either remove react imports or disable the transform which both is a breaking change. 

> - Applications can decide to opt into the new JSX transform by passing an environment variable, so that application code (not libraries) can be written without React in scope.

Yes. 

>   - Do we need to migrate all the application code at once or can we do it step-by-step?

An application, or more precisely the scope of bundling, needs to be migrated at once. We can use the codemod for it. As you can't use the new transform and import React to my understanding. This bears an interesting challenge to our all-in-one app and shared code. I feel we should just enable it across all apps and use the codemod. So in all it is not ideal but should be fine. 

> ESLint in CRA

Yeah, I thought about adding the rules behind the same logic in out eslint config here so it just works for anybody using it.

packages/eslint-config-mc-app/index.js
```diff
@@ -94,6 +96,10 @@ module.exports = {
     'lines-between-class-members': 0,
     // NOTE: The regular rule does not support do-expressions. The equivalent rule of babel does.
     'no-unused-expressions': 0,
+    ...(hasJsxRuntime() && {
```
tdeekens, line 99: This is what the React docs do.

packages/eslint-config-mc-app/has-jsx-runtime.js
```diff
@@ -1,5 +1,5 @@
 function hasJsxRuntime() {
-  if (process.env.DISABLE_NEW_JSX_TRANSFORM === 'true') {
+  if (process.env.ENABLE_NEW_JSX_TRANSFORM !== 'true') {
```
tdeekens, line 2: Otherwise I fear that we make this a breaking change. As either users have to opt-out or migrate their code to node have `import React from 'react'`. We might be able to swap it to `DISABLE_` (opt-out) in the future and then remove it entirely. At least that's my current preferred thinking towards introducing it.

emmenko: Perfect, this is what I had in mind as well.

.changeset/plenty-hotels-suffer.md
```diff
@@ -0,0 +1,15 @@
...
+"@commercetools-frontend/jest-preset-mc-app": minor
+"@commercetools-frontend/mc-scripts": minor
+---
+
+feat: adds the ability to opt into the new `automatic` JSX runtime
+
+Recent React versions support a new JSX runtime. Read more about it [here](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html). 
+
+You may opt into the new runtime in `automatic` mode by setting the `ENABLE_NEW_JSX_TRANSFORM` environment variable to `true`. Please note you need at least React v17 or v16.14 in your application. 
+Opting into the new JSX transform automatically also changes the Babel, Jest and ESLint configurations. As a conssequence ESLint will warn whenever it discovers React being in scope by importing it as `import React from 'react`'. You have to remove those imports using the respective codemod by running `npx react-codemod update-react-imports`.
```
emmenko, line 13:
```suggestion
Opting into the new JSX transform automatically also changes the Babel, Jest and ESLint configurations. As a consequence ESLint will warn whenever it discovers React being in scope by importing it as `import React from 'react`'. You have to remove those imports using the respective codemod by running `npx react-codemod update-react-imports`.
```

packages/babel-preset-mc-app/index.js
emmenko, line 2:
```suggestion
module.exports = function getBabePresetConfigForMcApp(api, opts = {}) {
```

packages/mc-scripts/src/config/create-webpack-config-for-production.js
```diff
@@ -257,6 +258,9 @@ module.exports = ({ distPath, entryPoint, sourceFolders, toggleFlags }) => {
                   require.resolve(
                     '@commercetools-frontend/babel-preset-mc-app'
                   ),
+                  {
+                    runtime: hasJsxRuntime ? 'automatic' : 'classic',
```
emmenko, line 262:
```suggestion
                    runtime: hasJsxRuntime() ? 'automatic' : 'classic',
```

emmenko (commented):
Looking good so far.

Have you tried testing this out in the playground app? Running the codemod, enabling the feature, etc.

tdeekens:
> Looking good so far.
> 
> 
> 
> Have you tried testing this out in the playground app? Running the codemod, enabling the feature, etc.

Not yet. Didn't have the time. Planned to tomorrow. Let's see where it doesn't work ;)

packages/mc-scripts/src/config/create-webpack-config-for-production.js
tdeekens: Oh god. Thanks.

tdeekens:
So a little testing in the playground revealed:

1. Setting the `ENABLE_NEW_JSX_TRANSFORM="true"` in the `.env` works as expected to enable the transform
2. When the transform runs the application runs in development but looks visually off without migrating it

CleanShot 2020-10-26 at 09 51 24@2x

3. The codemod is very nice: asks for directory and type of codebase (js, flow or ts) and also allows destructuring `import * from` imports
3.1: `React.use*` to `use*`
3.2: `React.lay` to `lazy`
3.3.: `React.Component` to `Component`

CleanShot 2020-10-26 at 09 56 03@2x

4. Linting
4.1  Changes to our public ESlint preset are hard to test as we don't use it ourselves
4.2 Once codemod ran linting in playground needs adjusting as it warns that React needs to be in scope
5. Once the codemod ran the application continues to look visually off but that's not due to our logic here I think

In summary I would argue that the logic is fine so far. We might want to add an intermediate migration step where we change the imports of React but keep it in scope so the final step is less noisy. The visual inconsistency is something to evaluate after to me.

emmenko:
Thanks for the feedback.

> looks visually off without migrating it

What do you mean that?

tdeekens:
> Thanks for the feedback.
> 
> > looks visually off without migrating it
> 
> What do you mean that?

Sorry, I wasn't clear. What I saw was a visual change in `automatic` mode (image above) compared to `classic` mode. However, I can't reproduce it now anymore. Maybe my local branch wasn't up-to-date or outdated. So think this change would be good now.

If you approve this changeset I would continue migrating the playground (incl. ESLint) in a separate PR.

emmenko (approved):
💯

PR #1827, opened 2020-10-26
refactor(playground): to use explicit imports of React

tdeekens:
#### Summary

As an outcome of adopting the new JSX transform https://github.com/commercetools/merchant-center-application-kit/pull/1819 I suggest to add a small intermediate refactor to adjust the imports according to the codemod.

The benefits of this are that chore on codebase is separate and small when adopting transform and that we adopt the recommended style before.

### DIFF
--- a/.changeset/moody-melons-arrive.md
+++ b/.changeset/moody-melons-arrive.md
@@ -1,0 +1,5 @@
+---
+"playground": patch
+---
+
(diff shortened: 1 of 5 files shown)


emmenko (approved):

adnasa:
same feedback as the other PR. please ping me on future PRs

PR #1836, opened 2020-10-30
fix: always infer MC API URL from origin when running behind the MC proxy

emmenko:


packages/application-shell-connectors/src/utils/get-mc-api-url/get-mc-api-url.ts
```diff
@@ -1,5 +1,8 @@
 import type { ApplicationWindow } from '@commercetools-frontend/constants';
 
+const removeMcPrefix = (host: string) =>
+  host.replace(/^mc(-(\d){4,})?\.(.*)$/, '$3');
```
emmenko, line 4: This now includes our preview URLs, so there is no need to have the `disableInferringOfMcApiUrlOnProduction` option anymore.

tdeekens (approved):
Thanks. This makes sense.

adnasa (approved):

PR #1838, opened 2020-10-30
fix(webpack): babel preset syntax

emmenko:
See https://github.com/facebook/create-react-app/blob/545d4607c81dd876c06f66b8104056a3570af242/packages/react-scripts/config/webpack.config.js#L409-L416

No idea how this wasn't caught here in this repo...


tdeekens (approved):
Strange. Thanks.

PR #1839, opened 2020-10-30
fix: how FAST_REFRESH variable is evaluated

emmenko:


playground/.env
```diff
@@ -3,3 +3,4 @@ APP_URL="https://<your_app_hostname>"
 ECHO_SERVER_URL="http://localhost:3000/api/echo"
 TRACKING_GTM="GTM-XXXXXX"
 ENABLE_NEW_JSX_TRANSFORM="true"
+FAST_REFRESH="true"
```
tdeekens, line 6: Assuming this is intended

tdeekens (approved):

playground/.env
emmenko: Yes

PR #1842, opened 2020-10-31
fix(mc-scripts): to add jsx runtime for js files not only svg

tdeekens:
#### Summary

This pull request intends to fix the mc-scripts to also pass the jsx runtime property to the babel preset on js files.

#### Description

The prior addition seems to have only passed the babel config to svg files which "makes sense" as they are also transformed.

However by not passing it ot js files internally the build would fail. What I fail to understand is why it works in this repo.

### DIFF
--- a/.changeset/old-carpets-accept.md
+++ b/.changeset/old-carpets-accept.md
@@ -1,0 +1,5 @@
+---
+"@commercetools-frontend/mc-scripts": patch
+---
+
+fix(mc-scripts): to add jsx runtime for js files not only svg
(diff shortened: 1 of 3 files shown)


emmenko (approved):
Oh god, luckily it's an experimental feature 😅
Thanks!

tdeekens:
Yup. It is not a bug if nobody noticed ;)

Having a look at CI now. At least it now wakes up ;)

emmenko:
Ah now it makes also sense why we didn't notice the wrong preset syntax, because it was configured for svg files and not js files, and the playground does not use svgs.

tdeekens:
Yup. Duble negation sort of 😄.

PR #1844, opened 2020-10-31
fix(babel): pragma frag to be Fragment not React.Fragment

tdeekens:
#### Summary

This pull request fixes the `pragmaFrag` of the babel config to be `Fragment` when using the `automatic` runtime.

#### Description

By default the babel transform seems to transform a `<></>` to `<React.Fragment></React.Fragment>` which will break in `automatic` mode as `React` is not in scope.

For instance at something like this.

CleanShot 2020-10-31 at 19 36 23@2x

It should just compile it to `Fragment` as `React` doesn't have to be in scope for then it to be picked up.

Reference: https://babeljs.io/docs/en/babel-plugin-transform-react-jsx#pragmafrag

### DIFF
--- a/.changeset/fifty-jeans-shave.md
+++ b/.changeset/fifty-jeans-shave.md
@@ -1,0 +1,5 @@
+---
+"@commercetools-frontend/babel-preset-mc-app": patch
+---
+
+fix(babel): pragma frag to be Fragment not React.Fragment
--- a/packages/babel-preset-mc-app/index.js
+++ b/packages/babel-preset-mc-app/index.js
@@ -59,7 +59,9 @@ module.exports = function getBabePresetConfigForMcApp(api, opts = {}) {
           development: isEnvDevelopment || isEnvTest,
           // Will use the native built-in instead of trying to polyfill
           // behavior for any plugins that require one.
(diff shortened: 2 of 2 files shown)


tdeekens:
I am not 100% sure yet that's the issue. I wouldn't understand why this wouldn't be handled internally.

emmenko (approved):
Nice!

emmenko:
> I am not 100% sure yet that's the issue. I wouldn't understand why this wouldn't be handled internally. Still, we always crash on fragments and have React.Fragment in our built code in development mode.

You could just test it by changing the value directly in the node_modules and see if this fixes the issue.

tdeekens:
> > I am not 100% sure yet that's the issue. I wouldn't understand why this wouldn't be handled internally. Still, we always crash on fragments and have React.Fragment in our built code in development mode.
> 
> 
> 
> You could just test it by changing the value directly in the node_modules and see if this fixes the issue.

I tried. The value isn't picked up in the transform. I however had the same issue with the runtime (being classic before at times). Now it is solved. It must be somehow rooted in the interplay of all deps. 

The transform itself also sets the pragma to React.Fragment of not passed in classic mode. It however doesn't set anything in automatic mode. So I assume babel passes it down and we have to overwrite. 

Anyway, I feel our change makes sense.

PR #1853, opened 2020-11-07
chore: add import export api target

tdeekens:
#### Summary

This adds an `IMPORT_EXPORT` HTTP target. This will become relevant when the respected API will be started to be integrated. 

This is not publicly supported yet. We will have to do some internal work to add proxying support. We can keep this on hold until then or prepare for it already.

### DIFF
--- a/.changeset/thick-wasps-hang.md
+++ b/.changeset/thick-wasps-hang.md
@@ -1,0 +1,5 @@
+---
+"@commercetools-frontend/constants": patch
+---
+
+chore: add import export api target
--- a/.changeset/thick-wasps-hang.md
+++ b/.changeset/thick-wasps-hang.md
@@ -2,4 +2,4 @@
 "@commercetools-frontend/constants": patch
 ---
 
-chore: add import export api target
+chore: add import api target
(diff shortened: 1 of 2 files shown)


emmenko (approved):

tdeekens:
I need to talk to the team again. Looking at the API docs the have one `import.*` API url. If they plan to add an `export.*` as a separate API we should also have two targets. That might also mean we need to permissions `View,Manage Imports` and `View,Manage Exports`.

To me this assumption makes sense but I am need to clarify. I am just trying to work ahead of them.

emmenko:
Ok thanks for clarifying. Let's put this as a draft then.

daern91:
> I need to talk to the team again. Looking at the API docs the have one `import.*` API url. If they plan to add an `export.*` as a separate API we should also have two targets. That might also mean we need to permissions `View,Manage Imports` and `View,Manage Exports`.
> 
> To me this assumption makes sense but I am need to clarify. I am just trying to work ahead of them.

Yup, `import.*` is the URL prefix for import-api. For export, we haven't really started yet but I can't see any reason why we would not use `export.*`, as I think it makes the most sense. And with regards to the permissions, you're right, they should have separate permissions and will also have different scopes on the CTP side.

tdeekens:
Thanks. Two APIs seem highly likely to me too. Then we will have two targets two. One for each regardless of permission and OAuth scopes. Lets take that discussion elsewhere. Will open it tomorrow.

PR #1854, opened 2020-11-07
chore: update deps

emmenko:
Alternative PR from #1846 to update deps incrementally to find out what's causing the build/test errors.


tdeekens (approved):
VRTs seem to bite but not other tests.

PR #1856, opened 2020-11-08
fix(babel): to use import source for emotion and automatic runtime

tdeekens:
#### Summary

This pull request intends to fix the setup of the automatic runtime in combination with emotion.

#### Description

Emotion has recently added support for the automatic runtime. Then changed a bit again how it works

From this

CleanShot 2020-11-08 at 15 55 51@2x

To this

CleanShot 2020-11-08 at 15 55 38@2x

Ref: https://emotion.sh/docs/css-prop#babel-preset

So I guess we're lucky to have been sitting that change out quietly ;)

### DIFF
--- a/packages/babel-preset-mc-app/index.js
+++ b/packages/babel-preset-mc-app/index.js
@@ -60,7 +60,10 @@ module.exports = function getBabePresetConfigForMcApp(api, opts = {}) {
           // Will use the native built-in instead of trying to polyfill
           // behavior for any plugins that require one.
           ...(opts.runtime !== 'automatic'
-            ? { useBuiltIns: true, pragmaFrag: 'Fragment' }
+            ? {
+                useBuiltIns: true,
+                importSource: '@emotion/core',
+              }
             : {}),
           runtime: opts.runtime || 'classic',
         },
--- a/packages/babel-preset-mc-app/index.js
+++ b/packages/babel-preset-mc-app/index.js
@@ -59,13 +59,21 @@ module.exports = function getBabePresetConfigForMcApp(api, opts = {}) {
           development: isEnvDevelopment || isEnvTest,
           // Will use the native built-in instead of trying to polyfill
           // behavior for any plugins that require one.
-          ...(opts.runtime !== 'automatic'
-            ? {
-                useBuiltIns: true,
-                importSource: '@emotion/core',
-              }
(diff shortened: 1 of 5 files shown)


emmenko (changes requested):
From what I understand we need to also remove `@emotion/babel-preset-css-prop` and use `babel-plugin-emotion` instead, right? 🧐

> If you want to configure `runtime: "automatic"` just replace `@emotion/babel-preset-css-prop` with `@babel/preset-react` and `babel-plugin-emotion`.

tdeekens:
You‘re right. Didn’t spot we use that. Will update this accordingly.

tdeekens:
This is how I'd understand it.

packages/babel-preset-mc-app/index.js
```diff
@@ -60,18 +60,14 @@ module.exports = function getBabePresetConfigForMcApp(api, opts = {}) {
           // Will use the native built-in instead of trying to polyfill
           // behavior for any plugins that require one.
           ...(opts.runtime !== 'automatic'
-            ? { useBuiltIns: true, pragmaFrag: 'Fragment' }
+            ? {
+                useBuiltIns: true,
+                importSource: '@emotion/core',
```
emmenko, line 65: image

If I understand this correctly, we should always use this option, right? 🤔

tdeekens: I think so. I read the changelogs of most emotion packages and find things quite interestingly confusing. I think we can always use this yes. 

Then however I am not sure why the build fails with the usual 'Can not find React'. Something doesn't kick in. Maybe an outdated version though.

packages/babel-preset-mc-app/index.js
```diff
@@ -60,18 +60,14 @@ module.exports = function getBabePresetConfigForMcApp(api, opts = {}) {
...
+              }
             : {}),
           runtime: opts.runtime || 'classic',
         },
       ],
-      [
-        '@emotion/babel-preset-css-prop',
```
emmenko, line 69: If we don't use the new runtime, we should still use this. I suppose we need to make this and the `babel-plugin-emotion` configurable as well?

tdeekens (commented):
🙏

emmenko (approved):

PR #1867, opened 2020-11-11
fix: run prettier after bumping changeset versions

emmenko:


tdeekens (approved):
Seen this before ;). Thanks.

PR #1871, opened 2020-11-15
refactor: compile and bundle monorepo using preconstruct cli

emmenko:
Use `@preconstruct/cli` to manage compiling and bundling monorepos, which means:
- no more rollup (precostruct does that)
- no more ts config files in each package

Note that we still need to bundle app-shell on our own because we still use css modules (#1884 )


packages/browser-history/package.json
```diff
@@ -15,18 +15,11 @@
   "publishConfig": {
     "access": "public"
   },
-  "main": "./dist/browser-history.cjs.js",
-  "module": "./dist/browser-history.es.js",
-  "typings": "./dist/typings/index.d.ts",
-  "types": "./dist/typings/index.d.ts",
```
tdeekens, line 21: Don't we need these for consumers of the package? At least types? 🤔

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.ts
```diff
@@ -70,22 +63,22 @@ const useIsAuthorized = ({
   demandedDataFences?: TDemandedDataFence[];
   selectDataFenceData?: TSelectDataFenceData;
   shouldMatchSomePermissions?: boolean;
-}) => {
+}): boolean => {
   const impliedPermissions = getImpliedPermissions(demandedPermissions);
 
-  useWarning(
```
tdeekens, line 76: Is this change related to what we're experimenting with on this PR? The goal here was to show the warning only once using the `useWarning` and to not emit it on each render.

emmenko: Yes the `tiny-invariant` dep was missing, and since we prefer to use that I removed the `tiny-warning` dep.

> The goal here was to show the warning only once

Ah that's a good point, I'll change that back. Thanks!

packages/browser-history/package.json
emmenko: Apparently TypeScript can still find the declarations, according to this: https://github.com/preconstruct/preconstruct/pull/93#pullrequestreview-285817702

packages/application-shell-connectors/package.json
```diff
@@ -15,31 +15,25 @@
...
-  "main": "./dist/application-shell-connectors.cjs.js",
-  "module": "./dist/application-shell-connectors.es.js",
-  "typings": "./dist/typings/index.d.ts",
-  "types": "./dist/typings/index.d.ts",
+  "main": "dist/commercetools-frontend-application-shell-connectors.cjs.js",
+  "module": "dist/commercetools-frontend-application-shell-connectors.esm.js",
   "files": ["dist", "test-utils", "package.json", "LICENSE", "README.md"],
+  "preconstruct": {
+    "entrypoints": ["./index.ts", "./test-utils/index.ts"]
+  },
```
emmenko, line 23: @adnasa FYI: this is how you would define multiple entrypoints.

packages/application-shell-connectors/test-utils/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "main": "dist/commercetools-frontend-application-shell-connectors-test-utils.cjs.js",
+  "module": "dist/commercetools-frontend-application-shell-connectors-test-utils.esm.js"
+}
```
emmenko, line 4: @adnasa ...and here.

adnasa (approved):
Looks good! 🔥

package.json
```diff
@@ -68,7 +68,27 @@
...
+      "packages/i18n",
+      "packages/l10n",
+      "packages/notifications",
+      "packages/permissions",
+      "packages/react-notifications",
+      "packages/sdk",
+      "packages/sentry",
+      "packages/url-utils",
+      "packages-backend/express",
+      "packages-backend/loggers"
```
tdeekens, line 88: We could also specify `packages/*` and `!packages/foo` to have exclusion. At least that's what I did in the flopper.

tdeekens (approved):

package.json
emmenko: Yes but there are several packages to exclude. For now I think being explicit is better

adnasa:
👏

PR #1879, opened 2020-11-16
refactor(actions-global/types): add `text` as optional to `TShowNotification`

### LINKED ISSUE #1870: Type error when using useShowNotification
**Describe the issue**

A type error occurred when using `useShowNotification`.

**To Reproduce**

Example code:

```js
const showNotification = useShowNotification();
showNotification({
  kind: 'success',
  domain: DOMAINS.SIDE,
  text: 'My text',
});
```

**Expected behavior**

Because the passed object is of type `Notification extends TShowNotification`, I don't expect any typing error. However, the following error occurred.

```
Argument of type '{ kind: "success"; domain: "side"; text: string; }' is not assignable to parameter of type 'TShowNotification'.
  Object literal may only specify known properties, and 'text' does not exist in type 'TShowNotification'.ts(2345)
```

**Screenshots**

Screenshot 2020-11-13 at 16 16 35

**Additional context**

adnasa:
#### Summary

- closes #1870

### DIFF
--- a/.changeset/long-cats-cheat.md
+++ b/.changeset/long-cats-cheat.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/actions-global': patch
+---
+
+extend `TShowNotification` with optional `text`
--- a/packages/actions-global/src/types.ts
+++ b/packages/actions-global/src/types.ts
@@ -11,6 +11,7 @@ import type {
 export type TShowNotification = TNotification & {
   domain: TAppNotificationDomain;
   kind: TAppNotificationKind;
+  text?: string;
 };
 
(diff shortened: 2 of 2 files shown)


ahmehri (approved):

tdeekens (approved):
Simplifies the typing I think.

emmenko (approved):
Ideally we should have a notification type specific to a text notification, which then you can import and pass it to the type parameter.

For now it should be fine like this.
Thanks

adnasa:
> Ideally we should have a notification type specific to a text notification, which then you can import and pass it to the type parameter.


yepp. it is also something I had in mind.

PR #1880, opened 2020-11-17
fix(config): include the entryPointUriPath in the applicationId header

emmenko:


packages/application-config/src/process-config.ts
```diff
@@ -68,11 +68,16 @@ const processConfig = ({
       ),
     `Invalid MC API URL: "${appConfig.mcApiUrl}"`
   );
+  // TODO: make the `id` required once we release the new model.
```
adnasa, line 71: the `id` wil come from DB and is not user provided, right?

adnasa (approved):

packages/application-config/src/process-config.ts
emmenko: You have to provide it in the app config.

tdeekens (approved):

packages/application-config/src/process-config.ts
adnasa: ok. clear

PR #1894, opened 2020-11-25
fix: nohoist gatsby emotion deps, as they have not been updated yet to v11

emmenko:
This should do the trick.

image


adnasa (approved):
nice trick. let's release

tdeekens (approved):
Oh boi.

emmenko:
Ok let me close this, as this is not going anywhere.

The `gatsby-plugin-emotion` support for v11 has been released yesterday. I'll do a release of docs kit and then update and release appkit as well.
https://github.com/commercetools/commercetools-docs-kit/pull/752

PR #1898, opened 2020-11-26
chore: add HOST_GCP_STAGING env var to playground

jonnybel:
#### Summary

This PR adds a missing flag to the .env of the `playground`, required in order to be able to run it.


emmenko:
This is not meant to be public.

The variable is already defined (in one of the last merged PR). https://github.com/commercetools/merchant-center-application-kit/blob/2287813c19a54b7edb7005d417c5f6a7b34cad1f/playground/.env#L4

PR #1904, opened 2020-11-27
refactor(l10n/localize): export transformLocalizedFieldToLocalizedString

adnasa:
#### Summary

- allow export of `transformLocalizedFieldToLocalizedString`
- allow `locale` to be passed as `null`, leverage type-guard for the "lookup"

---

This is cherry picked from a feature branch I'm working on

### DIFF
--- a/packages/l10n/src/localize.ts
+++ b/packages/l10n/src/localize.ts
@@ -111,18 +111,33 @@ export const formatLocalizedString = <Input extends Record<string, unknown>>(
     fallbackOrder = [],
     fallback = '',
   }: FormatLocalizedStringOptions<Input>
-): string => {
-  if (!entity || !entity[key] || !locale) return fallback;
+) => {
+  if (!entity || !entity[key]) return fallback;
   const localizedString = entity[key] as LocalizedString;
-  if (localizedString[locale]) return localizedString[locale];
-  // see if we can fallback to the primary locale, eg. de for de-AT
-  const primaryLocale = locale && getPrimaryLocale(locale);
-  if (localizedString[primaryLocale]) return localizedString[primaryLocale];
-  const fallbackLanguage = findFallbackLocale(localizedString, fallbackOrder);
-  return fallbackLanguage
+  const fallbackLocale = findFallbackLocale(localizedString, fallbackOrder);
+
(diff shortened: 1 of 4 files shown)


packages/l10n/src/types.ts
```diff
@@ -33,7 +33,7 @@ export type FieldNameTranformationMapping = {
 
 export type FormatLocalizedStringOptions<T> = {
   key: keyof T;
-  locale: string;
+  locale: string | null;
```
adnasa, line 36: I want to allow `null`  here.
Normally, the source of `dataLocale` is extracted from `application-context` and that value allows `null` 

[reference](https://github.com/commercetools/merchant-center-application-kit/blob/master/packages/application-shell-connectors/src/components/application-context/application-context.tsx#L117:L117)

tdeekens (approved):

packages/l10n/src/types.ts
emmenko: This is the case for applications that do not have a project context, for example auth and account.

I think we can be strict and avoid `null` here.

emmenko: After a second thought, it would be annoying indeed to always let the consumer check for the existence of `dataLocale`.
I suppose we would need to fix the types there first, but it's tricky to work with conditionals.

emmenko (approved):

packages/l10n/src/localize.ts
```diff
@@ -111,18 +111,33 @@ export const formatLocalizedString = <Input extends Record<string, unknown>>(
...
+  if (localizedString[locale]) return localizedString[locale];
+
+  // GIVEN locale
+  // AND there is a value on primary locale
+  // THEN return value on primary locale
+  const primaryLocale = locale && getPrimaryLocale(locale);
+  if (localizedString[primaryLocale]) return localizedString[primaryLocale];
+
+  // use formattedFallback by fallbackOrder as last resort
+  return formattedLocalizedFallback;
```
adnasa, line 142: this is done to not break the tests. the issue with my initial suggestion was that I fallback "too early".
the localize should try to do a lookup if there is any value at all in `localizedString` and just pick that.

PR #1913, opened 2020-12-02
fix(loggers): do not mutate meta object

emmenko:


packages-backend/loggers/src/loggers.spec.ts
```diff
@@ -43,4 +43,60 @@ describe('application logger', () => {
...
+      },
+    };
+    logger.info('Test log', {
+      meta: { req: { headers: fakeRequest.headers } },
+    });
+    expect(fakeRequest).toEqual({
+      headers: {
+        authorization: 'Bearer 123',
+      },
+    });
```
emmenko, line 87: This is the proof that the original object is not being mutated (with this fix).

adnasa (approved):
nice one!

PR #1914, opened 2020-12-02
feat(application-components): add experimental `product-picker-input`

adnasa:
#### Summary

- adds `debounce`, `use-load-with-prefix`hook
- adds `product-picker-input` 
- adds `product-picker-page`
  - This is for testing purposes is ought to be removed later

I intentionally didn't generate the types to keep the PR small.
per-commit review is recommended.

Keeping this as draft, for early review on implementation and to test internally.

#### TODO

- ~documentation [reference](https://docs.commercetools.com/custom-applications/components/info-dialog)~ (separate PR)
- [x] Add tests for `product-picker-input`

##### When approved

- [x] remove `playground/product-picker-page`
- [x] add `changeset`
- [x] generate CTP types


packages/application-components/src/experimental/components/product-picker-input/product-picker-input.tsx
```diff
@@ -0,0 +1,239 @@
...
+    skip: !props.value,
+  });
+
+  const { error: prefetchError } = prefetchSelectedProductQuery;
+  React.useEffect(() => {
+    setLoadingError(true);
+    if (prefetchError && onInitialLoadError) {
+      onInitialLoadError(prefetchError);
+    }
+  }, [prefetchError, onInitialLoadError, setLoadingError]);
```
emmenko, line 196: Be careful with this!

If the props change their reference on each re-render, you end up in an infinite loop here.

I bumped into a similar issue once: https://github.com/commercetools/merchant-center-application-kit/pull/1567

I suggest to exclude `onInitialLoadError` from the dependencies, as the other values are controlled by us.

packages/application-components/src/experimental/components/product-picker-input/product-picker-input.tsx
```diff
@@ -0,0 +1,239 @@
...
+
+  const prefetchSelectedProduct = prefetchSelectedProductQuery.data?.product
+    ? convertProductToOption(
+        flattenProduct(prefetchSelectedProductQuery.data.product)
+      )
+    : props.value;
+
+  return (
+    <Constraints.Horizontal constraint="scale">
```
emmenko, line 205: I'm not sure this component needs the horizontal constraint.

packages/application-components/src/experimental/components/product-picker-input/product-picker-input.tsx
```diff
@@ -0,0 +1,239 @@
...
+      {prefetchSelectedProductQuery.loading ? (
+        <CenteredLoadingSpinner />
+      ) : (
```
emmenko, line 208: Nit, in these cases I prefer to do

```ts
if (prefetchSelectedProductQuery.loading) {
  return ()
}

return ()
```

packages/application-components/src/experimental/components/shared/centered-loading-spinner/centered-loading-spinner.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { css } from '@emotion/react';
+import LoadingSpinner from '@commercetools-uikit/loading-spinner';
+
+const CenteredLoadingSpinner: React.FC = () => (
```
emmenko, line 5: Do we need `React.FC`?

packages/application-components/src/experimental/hooks/use-debounced-promise-callback/use-debounced-promise-callback.ts
```diff
@@ -0,0 +1,9 @@
+import debounce from 'debounce-promise';
+import { useCallback } from 'react';
+
+const useDebouncedPromiseCallback = <T extends (...args: any[]) => any>(
+  promiseCallback: T,
+  wait: number
+) => useCallback(debounce(promiseCallback, wait), [promiseCallback]);
```
emmenko, line 7: Shouldn't this be something like:

```suggestion
const useDebouncedPromiseCallback = <T>(
  promiseCallback: (...args: unknown[]) => Promise<T>,
  wait: number
) => useCallback(debounce(promiseCallback, wait), [promiseCallback]);
```

packages/application-components/src/experimental/hooks/use-load-with-prefix-search-fields/use-load-with-prefix-search-fields.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+    return comparison;
+  };
+  return prefixSearchFields.map(constructPrefixSearchExpression).join(' or ');
+};
+
+export const getPrefixSearchWherePredicate = (
+  inputValue: string,
+  prefixSearchFields: string[]
+) => {
```
emmenko, line 49:
```suggestion
): string => {
```

packages/application-components/src/experimental/hooks/use-load-with-prefix-search-fields/use-load-with-prefix-search-fields.ts
```diff
@@ -0,0 +1,86 @@
...
+  return constructPrefixSearchPredicate(
+    prefixSearchFields,
+    lowerBound,
+    upperBound
+  );
+};
+
+const useLoadWithPrefixSearchFields = (
+  options: TLoadingWithPrefixSearchFieldsOptions
+) => {
```
emmenko, line 60: Nit: We should provide the return type, as we export this function.

packages/application-components/src/experimental/hooks/use-load-with-prefix-search-fields/use-load-with-prefix-search-fields.ts
```diff
@@ -0,0 +1,86 @@
...
+  const { query, prefixSearchFields } = options;
+  const apolloClient = useApolloClient();
+  const loadWithPrefixSearchFields = React.useCallback(
+    ({
```
emmenko, line 64:
```suggestion
    async ({
```

To make it clear that it returns a Promise

packages/application-components/src/utils/is-uuid.ts
```diff
@@ -0,0 +1,6 @@
+const isUUID = (string: string): boolean =>
```
emmenko, line 1: Maybe we should use the `uuid.validate` function

https://www.npmjs.com/package/uuid#uuidvalidatestr

playground/src/routes.js
```diff
@@ -28,6 +29,9 @@ const ApplicationRoutes = (props) => {
         <Route path={`${props.match.path}/echo-server`}>
           <EchoServer />
         </Route>
+        <Route>
+          <ProductPickerPage />
+        </Route>
```
emmenko, line 34: We should remove this before merging.

emmenko (commented):
I did a first review. Looks good overall, just some improvements here and there.

Great job so far! 💯

packages/application-components/src/experimental/components/product-picker-input/product-picker-input.tsx
adnasa: I looked at this again. I think it makes sense to keep this for now.

packages/application-components/src/experimental/components/product-picker-input/product-picker-input.tsx
```diff
@@ -0,0 +1,239 @@
...
+  value: string;
+};
+
+type TPrefixProductGraphqlResults = {
+  results: TFetchedProductWithMasterData[];
+};
+
+type TPrefixProductGraphqlProducts = {
+  products: TPrefixProductGraphqlResults;
+};
```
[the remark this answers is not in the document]
adnasa: The thing about this, is that the generated types are in `application-shell` and they are not exported.
If we want to do this, we need to export the exports from `application-shell` then to be imported into `application-components`

I'm not really sure if we want that.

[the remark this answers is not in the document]
emmenko: No need for that, you simply generate the types for the components package. Check the graphql config file.

We already do that for other packages as we can't really share those types.

packages/application-components/src/experimental/components/product-picker-input/product-picker-input.tsx
```diff
@@ -0,0 +1,239 @@
+import React, { ChangeEvent } from 'react';
```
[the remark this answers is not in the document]
adnasa: hmm

```
Duplicate declaration "React"
🎁 error @commercetools-frontend/application-components   1 | import type { ChangeEvent } from 'react';
🎁 error @commercetools-frontend/application-components > 2 | import React from 'react';
🎁 error @commercetools-frontend/application-components     |        ^^^^^
```

packages/application-components/src/experimental/components/product-picker-input/product-picker-input.tsx
```diff
@@ -155,7 +145,7 @@ const ProductPickerInput = (props: TProps): JSX.Element => {
   const prefetchSelectedProductQuery = useQuery(PrefetchProduct, {
     variables: { id: props.value },
     context: { target: GRAPHQL_TARGETS.COMMERCETOOLS_PLATFORM },
-    skip: !props.value,
+    skip: !props.value || !isUUID(props.value),
```
adnasa, line 148: now we skip the prefetch of product if the initial value was invalid.
this is preventive action to getting an `onInitialLoadError`

packages/application-components/src/experimental/hooks/use-debounced-promise-callback/use-debounced-promise-callback.ts
```diff
@@ -1,9 +1,15 @@
 import debounce from 'debounce-promise';
 import { useCallback } from 'react';
 
-const useDebouncedPromiseCallback = <T extends (...args: any[]) => any>(
-  promiseCallback: T,
-  wait: number
-) => useCallback(debounce(promiseCallback, wait), [promiseCallback]);
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type DebouncedCallback<T> = (...args: any[]) => Promise<T>;
```
adnasa, line 5: I tried `unknown` but I got a pretty tricky TS error on the consumer side, which I spent too much time trying to resolve.
Saving this for another day

image

packages/application-components/src/experimental/components/product-picker-input/product-picker-input.spec.js
```diff
@@ -0,0 +1,214 @@
...
+    context: { target: GRAPHQL_TARGETS.COMMERCETOOLS_PLATFORM },
+    variables: { id: productId },
+  },
+  result: {
+    data: {
+      product: createTestProduct(),
+    },
+  },
+  ...custom,
+});
```
adnasa, line 58: I have a TODO to add `Product` into `test-data`.
it's about time

@types/commercetools-uikit__async-select-input/index.d.ts
```diff
@@ -18,7 +18,7 @@ declare module '@commercetools-uikit/async-select-input' {
     menuPortalZIndex: number;
     containerId?: string;
     name: string;
-    value: string | string[] | TAsyncOption;
+    value?: string | string[] | TAsyncOption | null;
```
adnasa, line 21: `null` is the fallback value when consumer is able to convert a nullable `data` field into an option.
I prefer this over `undefined`. `null` derives directly from the Apollo result

emmenko:
@adnasa FYI: I pushed a commit to improve a bunch of things.

packages/application-config/src/schema.ts
```diff
@@ -21,7 +21,7 @@ export interface JSONSchemaForCustomApplicationConfigurationFiles {
   /**
    * The cloud identifier where the Custom Application is running. This value is used to derive the Merchant Center API URL. Alternatively you can use the `mcApiUrl` property.
    */
-  cloudIdentifier: ('gcp-au' | 'gcp-eu' | 'gcp-us' | 'aws-fra' | 'aws-ohio') | EnvVariablePlaceholder;
+  cloudIdentifier: (('gcp-au' | 'gcp-eu' | 'gcp-us' | 'aws-fra' | 'aws-ohio') | EnvVariablePlaceholder) & string;
```
emmenko, line 24: ??

emmenko (commented):
I can take a look at the failing test

packages/application-components/src/experimental/components/internals/centered-loading-spinner/centered-loading-spinner.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { css } from '@emotion/react';
+import LoadingSpinner from '@commercetools-uikit/loading-spinner';
+
+const CenteredLoadingSpinner = (): JSX.Element => (
+  <div
+    data-testid="picker-centered-loading-spinner"
```
pa3, line 7: In the future (not sure worth doing now) the file can probably be moved outside of the `internals` scope and the `test-id` changed to be just `data-testid="centered-loading-spinner"`. I can foresee the need for this component in other places too.

packages/application-components/src/experimental/components/internals/option/option.tsx
```diff
@@ -0,0 +1,31 @@
...
+  key: string;
+  label: string;
+  value: string;
+};
+
+type TProps = {
+  data: TData;
+};
+
+const Option = <T extends TProps>(props: T): JSX.Element => {
```
pa3, line 17: Asking just out of personal curiosity: why not just 

```
const (props: TProps): JSX.Element => {
```

TS will complain that additional props are passed in this case?

packages/application-components/src/experimental/components/product-picker-input/product-picker-input.spec.tsx
```diff
@@ -0,0 +1,216 @@
+import type { TRenderAppOptions } from '@commercetools-frontend/application-shell/test-utils';
+
+// import { graphql } from 'msw';
+// import { setupServer } from 'msw/node';
```
pa3, line 4: Leftovers?

packages/application-components/src/experimental/hooks/use-load-with-prefix-search-fields/use-load-with-prefix-search-fields.spec.ts
```diff
@@ -0,0 +1,18 @@
+import { getPrefixSearchWherePredicate } from './use-load-with-prefix-search-fields';
+
+describe('getPrefixSearchWherePredicate', () => {
+  it.each`
+    prefixSearchFields                       | inputValue | predicate
+    ${['name.en', 'key']}                    | ${'s'}     | ${'name(en >= "s" and en < "t") or key >= "s" and key < "t"'}
+    ${['masterData.current.name.en', 'key']} | ${'s'}     | ${'masterData(current(name(en >= "s" and en < "t"))) or key >= "s" and key < "t"'}
+    ${['name.en']}                           | ${'s'}     | ${'name(en >= "s" and en < "t")'}
+    ${['key']}                               | ${'s'}     | ${'key >= "s" and key < "t"'}
```
pa3, line 9: :heart_eyes:

pa3 (approved):
LGTM! :+1:

packages/application-components/src/experimental/components/product-picker-input/product-picker-input.spec.tsx
adnasa: yeah my bad. 
They are not supposed to be there :)

packages/application-components/src/experimental/components/internals/option/option.tsx
adnasa: hey, thanks for this question.
What I want to achieve here is that the consumer can pass more props than `props.data`.

as you can see in `L19`, we spread the props to the underlying component.
If the consumer implements a select option that consumes this, they can define their own _required_ props top of `TProps`.

At least that was my aim.

packages/application-components/src/experimental/components/internals/centered-loading-spinner/centered-loading-spinner.tsx
adnasa: yes I have always considered this, but I came to the same conclusion as you that this is premature

packages/application-components/src/experimental/components/internals/option/option.tsx
tdeekens: React.ReactNode and JSX.Element always confused me as to when to use which. So I'll just share that here too. A ReactNode is a ReactElement, a ReactFragment, a string, a number or an array of ReactNodes, or null, or undefined, or a boolean. A SX.Element is a ReactElement, with the generic type for props and type being any.

This always drove me nuts. I kept the rule of thumb that a ReactNode is what render on a class component returns and is used if you e.g. allow to return a string A JSX.Element is only for Functional Components.

That said: if an option can be a string or undefined then this might want to be ReactNode.

emmenko: A `JSX.Element` is what usually TS uses/infers when you don't explicitly define the return type. Which is also how  `FC` is declared.

image

So it's fine here. A `ReactNode` is what you use for example for typing `children`.

emmenko: > If the consumer implements a select option that consumes this, they can define their own required props top of TProps.

This component is only used internally, so we don't really need to make it unnecessarily generic.
You can keep it simple maybe.
I'm also fine either way.

tdeekens: What I said: JSX.Element is for React.FC. ReactNode for ClassComponent.

emmenko: You can always "force it" to return a `JSX.Element` by wrapping it into a fragment, for example:

```jsx
const Text = () => <>Hello</>
```

packages/application-components/src/experimental/components/product-picker-input/product-picker-input.spec.tsx
```diff
@@ -0,0 +1,216 @@
...
+import ProductPickerInput from './product-picker-input';
+import FetchProductByPrefix from './fetch-product-by-prefix.ctp.graphql';
+import PrefetchProduct from './prefetch-selected-product.ctp.graphql';
+
+const productId = '6f69c022-b894-4deb-b19f-6335547cb5d7';
+const inputName = 'product-picker-input';
+
+const createTestProps = (custom = {}) => ({
+  name: inputName,
+  value: productId,
```
emmenko, line 21: Better maybe to consider the default props without a `value`. Then when you want to test with an existing value, you pass it explicitly.

packages/application-components/src/experimental/components/product-picker-input/product-picker-input.spec.tsx
```diff
@@ -0,0 +1,216 @@
...
+    },
+  },
+  ...custom,
+});
+
+const render = (
+  ui: React.ReactElement,
+  renderOptions: Partial<TRenderAppOptions<{}>>
+) => {
+  return renderApp(
```
emmenko, line 84: Note: as follow up we should maybe look into making test-utils more generic available. For instance, I don't like much that we use the appshell test-utils in this package.

packages/application-components/src/experimental/components/product-picker-input/product-picker-input.spec.tsx
```diff
@@ -0,0 +1,216 @@
...
+    it('should prefetch product', async () => {
+      const props = createTestProps();
+      render(<ProductPickerInput {...props} />, {
+        mocks: [createPrefetchedProductQueryMock()],
+      });
+      await screen.findByText(/milk/i);
+    });
+  });
+  describe('when `value` is not valid', () => {
+    it('should not prefetch product', async () => {
```
emmenko, line 107: Do we also expect to show an error message/state here?

packages/application-components/src/experimental/components/product-picker-input/product-picker-input.spec.tsx
```diff
@@ -0,0 +1,216 @@
...
+describe('loading products', () => {
+  describe('when product(s) is found', () => {
+    it('should display found product as option', async () => {
+      const props = createTestProps({ value: null });
+      render(<ProductPickerInput {...props} />, {
+        mocks: [createPrefixedSearchProductQueryMock()],
+      });
+
+      const input = await screen.findByLabelText('Product picker');
+      expect(input).toBeInTheDocument();
```
emmenko, line 152: I think this is not necessary, as it's implied by `findBy`

packages/application-components/src/experimental/components/product-picker-input/product-picker-input.spec.tsx
```diff
@@ -0,0 +1,216 @@
...
+                products: {
+                  results: [],
+                },
+              },
+            },
+          }),
+        ],
+      });
+      const input = await screen.findByLabelText('Product picker');
+      expect(input).toBeInTheDocument();
```
emmenko, line 179: Same

packages/application-components/src/experimental/components/product-picker-input/product-picker-input.spec.tsx
```diff
@@ -0,0 +1,216 @@
...
+                  code: 'InvalidRequest',
+                  message: 'invalid request',
+                },
+              ],
+            },
+          }),
+        ],
+      });
+      const input = await screen.findByLabelText('Product picker');
+      expect(input).toBeInTheDocument();
```
emmenko, line 208: Same

packages/application-components/src/experimental/components/product-picker-input/product-picker-input.spec.tsx
```diff
@@ -0,0 +1,216 @@
...
+      fireEvent.change(input, {
+        target: { value: 'H' },
+      });
+      await screen.findByText(/unable to load products/i);
+      expect(props.onLoadError).toHaveBeenCalled();
```
emmenko, line 213: `toHaveBeenCalledWith` ?

emmenko (commented):
Looking good, just left some suggestions for the tests.

packages/application-components/src/experimental/components/product-picker-input/product-picker-input.spec.tsx
adnasa: yes I thought about that too. let's move this to an issue.

packages/application-components/src/experimental/components/product-picker-input/product-picker-input.spec.tsx
adnasa: with the status quo, I would not recommend that to display to the MC user.
The cases where the value is invalid;

1. user pastes in a invalid ID => do not send a request. displaying an error wouldn't make sense and if anything the consumer should decide that through `onInputChange`
2. the initial value was `null` <= this is a side effect of transformations

packages/application-config/src/schema.ts
emmenko: Hmm this seems to be an artifact of the schema. Let's ignore this and keep it like this.

@types/commercetools-uikit__select-input/index.d.ts
```diff
@@ -62,13 +62,19 @@ declare module '@commercetools-uikit/select-input' {
...
       ): JSX.Element;
     };
     ValueContainer: {
-      <OptionType extends OptionTypeBase = { label: string; value: string }>(
-        props: ValueContainerProps<OptionType>
+      <
+        OptionType extends OptionTypeBase = { label: string; value: string },
+        IsMulti extends boolean = false
+      >(
+        props: ValueContainerProps<OptionType, IsMulti>
```
adnasa, line 77: 👍 
yeah I tried this too. works fine

emmenko (approved):
🚀

tdeekens (approved):

adnasa:
Finally!! 🚀 
aaawwww, my baby going into `master` 😭 😭 😭  (tears of joy)

PR #1915, opened 2020-12-05
fix(babel): regression, pass ROLLUP_BUILD env

emmenko:
Our babel preset uses a `ROLLUP_BUILD` env variable that we forgot to set when we moved to preconstruct.
It only affected prop types.


tdeekens (approved):
🥸

PR #1922, opened 2020-12-11
refactor: menu links to use their own permission data

emmenko:
As we are introducing some changes related to permissions in the MC API, we need to ensure that the menu links are not affected by that. This is because menu links are considered a "shared" resource across all applications.

~~**DO NOT MERGE THIS FOR NOW UNTIL THE CHANGES IN THE MC API ARE LIVE**~~

### DIFF
--- a/packages/application-shell-connectors/src/components/application-context/application-context.tsx
+++ b/packages/application-shell-connectors/src/components/application-context/application-context.tsx
@@ -9,7 +9,6 @@ import moment from 'moment-timezone';
 import getDisplayName from '../../utils/get-display-name';
 import {
   normalizeAllAppliedActionRights,
-  normalizeAllAppliedMenuVisibilities,
   normalizeAllAppliedPermissions,
   normalizeAllAppliedDataFences,
 } from './normalizers';
@@ -19,7 +18,6 @@ type TFetchedUser = TFetchLoggedInUserQuery['user'];
 type TFetchedProject = TFetchProjectQuery['project'];
 
 type TApplicationContextPermissions = { [key: string]: boolean };
-type TApplicationContextMenuVisibilities = { [key: string]: boolean };
 type TActionRight = {
   [key: string]: boolean;
 };
@@ -112,7 +110,6 @@ export type TApplicationContext<AdditionalEnvironmentProperties extends {}> = {
   project: ReturnType<typeof mapProjectToApplicationContextProject>;
   permissions: TApplicationContextPermissions | null;
   actionRights: TApplicationContextActionRights | null;
-  menuVisibilities: TApplicationContextMenuVisibilities | null;
   dataFences: TApplicationContextDataFences | null;
   dataLocale: string | null;
 };
@@ -146,7 +143,6 @@ const createApplicationContext: <AdditionalEnvironmentProperties extends {}>(
   project: mapProjectToApplicationContextProject(project),
   permissions: normalizeAllAppliedPermissions(project),
   actionRights: normalizeAllAppliedActionRights(project),
-  menuVisibilities: normalizeAllAppliedMenuVisibilities(project),
   dataFences: normalizeAllAppliedDataFences(project),
   dataLocale: projectDataLocale || null,
 });
--- a/packages/application-shell-connectors/src/components/application-context/application-context.tsx
+++ b/packages/application-shell-connectors/src/components/application-context/application-context.tsx
@@ -141,9 +141,11 @@ const createApplicationContext: <AdditionalEnvironmentProperties extends {}>(
   environment: mapEnvironmentToApplicationContextEnvironment(environment),
   user: mapUserToApplicationContextUser(user),
   project: mapProjectToApplicationContextProject(project),
-  permissions: normalizeAllAppliedPermissions(project),
-  actionRights: normalizeAllAppliedActionRights(project),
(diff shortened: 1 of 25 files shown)


tdeekens (approved):
Thanks. Looks good to me.

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -112,7 +110,6 @@ export type TApplicationContext<AdditionalEnvironmentProperties extends {}> = {
   project: ReturnType<typeof mapProjectToApplicationContextProject>;
   permissions: TApplicationContextPermissions | null;
   actionRights: TApplicationContextActionRights | null;
-  menuVisibilities: TApplicationContextMenuVisibilities | null;
```
emmenko, line 115: This now is not exposed anymore in the application context, as it's only used by the navbar anyway.

It's a breaking change, but more of an internal one.

packages/application-shell-connectors/src/components/application-context/index.ts
```diff
@@ -10,10 +16,21 @@ export type TApplicationContext<
   AdditionalEnvironmentProperties extends {}
 > = ApplicationContext<AdditionalEnvironmentProperties>;
 
+export type TNormalizedMenuVisibilities = NormalizedMenuVisibilities;
+export type TNormalizedPermissions = NormalizedPermissions;
+export type TNormalizedActionRights = NormalizedActionRights;
+export type TNormalizedDataFences = NormalizedDataFences;
```
emmenko, line 22: Exposing those types as they are re-use them in other packages.

packages/application-shell/src/components/application-shell/application-shell.tsx
```diff
@@ -366,12 +366,15 @@ export const RestrictedApplication = <
                                   return (
                                     <ApplicationContextProvider<AdditionalEnvironmentProperties>
                                       user={user}
-                                      project={project}
```
emmenko, line 369: This is important NOT to set it!

The idea is that the application context does not contain the "normal" project permissions within the navbar component, but instead evaluating the permissions for the menu links is done more explicitly and by using the specific permissions for the menu.

We do it this way to be able to use the same "isAuthorized" logic but with different permissions inputs.

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -329,6 +348,7 @@ const RestrictedMenuItem = (props: RestrictedMenuItemProps) => {
         }}
         // Always check that some of the given permissions match.
         shouldMatchSomePermissions={true}
+        projectPermissions={props.projectPermissions}
```
emmenko, line 351: Here we explicitly pass the project permissions (specific to the menu links) to the `RestrictedByPermissions` component.

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -555,11 +579,20 @@ const NavBar = <AdditionalEnvironmentProperties extends {}>(
...
+    () => ({
+      permissions: normalizeAllAppliedPermissions(props.project),
+      actionRights: normalizeAllAppliedActionRights(props.project),
+      dataFences: normalizeAllAppliedDataFences(props.project),
+    }),
+    [props.project]
+  );
+  const menuVisibilities = React.useMemo(
+    () => normalizeAllAppliedMenuVisibilities(props.project),
+    [props.project]
```
emmenko, line 594: Normalizing the data is usually done in the application context (connectors). We need to do it manually here since we're skipping the application context now.

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.ts
```diff
@@ -94,14 +86,17 @@ const useIsAuthorized = ({
...
-    (applicationContext) => applicationContext.actionRights
+  const actualActionRights = useApplicationContext<TNormalizedActionRights | null>(
+    (applicationContext) =>
+      projectPermissions?.actionRights ?? applicationContext.actionRights
   );
-  const actualDataFences = useApplicationContext<TDataFences | null>(
-    (applicationContext) => applicationContext.dataFences
+  const actualDataFences = useApplicationContext<TNormalizedDataFences | null>(
+    (applicationContext) =>
+      projectPermissions?.dataFences ?? applicationContext.dataFences
```
emmenko, line 99: As you can see here, we try to read/resolve the permissions passed as prop first, otherwise falling back to the application context.

packages/application-shell-connectors/src/components/application-context/normalizers.ts
```diff
@@ -1,9 +1,46 @@
...
-  Maybe,
-  TFetchProjectQuery,
+  TAppliedPermission,
+  TAppliedActionRight,
+  TAppliedMenuVisibilities,
   TStoreDataFence,
 } from '../../types/generated/mc';
 
+// Menu visibilities
+export type TMenuVisibilities = {
```
tdeekens, line 9: It seems to be a matter of subjective opinion but I like `Record<string, boolean>` for these type of things. Keep it as is if you want to :)

tdeekens (commented):
Looks good to me. Keeping this a comment as it's a draft but overall looks perfect. I like how few tests have to change :)

adnasa (approved):
👍

PR #1928, opened 2020-12-15
chore: add `react-select` to resolutions

adnasa:
> Note there is probably another way to do this, but this PR fixes the typecheck issue on master

#### Summary

- adds `react-select@3.1.0` to resolutions

#### Description

on `master` we get a typecheck error, where `OptionProps` and `CustomValueContainerProps` has a second required argument. I'm open to other approaches. I tried to fix the declaration file but spent too much time with it with no progress.

### DIFF
--- a/packages/application-config/src/schema.ts
+++ b/packages/application-config/src/schema.ts
@@ -21,7 +21,7 @@ export interface JSONSchemaForCustomApplicationConfigurationFiles {
   /**
    * The cloud identifier where the Custom Application is running. This value is used to derive the Merchant Center API URL. Alternatively you can use the `mcApiUrl` property.
    */
-  cloudIdentifier: ('gcp-au' | 'gcp-eu' | 'gcp-us' | 'aws-fra' | 'aws-ohio') | EnvVariablePlaceholder;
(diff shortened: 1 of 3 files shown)


adnasa:
I don't think a changeset is needed

packages/application-config/src/schema.ts
```diff
@@ -21,7 +21,7 @@ export interface JSONSchemaForCustomApplicationConfigurationFiles {
   /**
    * The cloud identifier where the Custom Application is running. This value is used to derive the Merchant Center API URL. Alternatively you can use the `mcApiUrl` property.
    */
-  cloudIdentifier: ('gcp-au' | 'gcp-eu' | 'gcp-us' | 'aws-fra' | 'aws-ohio') | EnvVariablePlaceholder;
+  cloudIdentifier: (('gcp-au' | 'gcp-eu' | 'gcp-us' | 'aws-fra' | 'aws-ohio') | EnvVariablePlaceholder) & string;
```
adnasa, line 24: yarn build 🤷

emmenko:
Not relevant anymore.

PR #1929, opened 2020-12-16
Add Northern Ireland country to the list of available countries list

mohib0306:
#### Summary

- `l10n` => Updated the `json` files that holds that countries and their codes. The CLDR has not yet release their latest data which would have held this change, therefore the list were updated manually.


.changeset/popular-apes-complain.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-frontend/l10n': patch
```
emmenko, line 2:
```suggestion
'@commercetools-frontend/l10n': minor
```

emmenko (commented):
Thanks but shouldn't we wait for the official release? 🤔

mohib0306:
> shouldn't we wait for the official release?

Its new country code is relevant for VAT purposes (tax categories used in the CTP). The official announcement by the EU members is already done, it is just that the CLDR has not yet release the latest dataset and is expected sometime next year.

emmenko:
Ok. Can you put it in alphabetical order maybe, so then it's easier to find and diffs won't be confusing when we update the cldr next time?

Thanks

adnasa:
I was expecting that we update the script to manually add that and not update the jsons directly.
the value of having that on script is that we can inline document that and easily revert once CLDR are up to date.

Can you update a README, create an issue on this repo to keep an eye on CLDR?
I would mark this as tech debt.

emmenko (approved):

PR #1933, opened 2020-12-18
chore(header-link): add operation name to missing target error

pa3:
I am investigating an error reported to Sentry from MC saying that `GraphQL target "undefined" is missing or is not supported`. Went manually through all the queries we sent in MC to try to find the one with a missing `target` param. No luck. 

So, I decided to extend the error message thrown from HttpLink so that it will include `operationName` to narrow down the scope of further investigation.

### DIFF
--- a/.changeset/seven-clocks-smell.md
+++ b/.changeset/seven-clocks-smell.md
@@ -1,0 +1,5 @@
+---
+"@commercetools-frontend/application-shell": patch
+---
+
+Improve error message when graphql target is missing.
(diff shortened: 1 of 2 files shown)


emmenko:
Thanks, it makes sense to improve the error message. Please also add a changeset, so that we can release this change.

PS: for debugging you can also manually adjust the error in the `node_modules` bundles, to avoid waiting for this release.

pa3:
That's the thing with `node_modules` monkey-patching: I can't reproduce it locally.

emmenko (approved):

PR #1934, opened 2020-12-18
feat: implement new experimental login workflow for local development

emmenko:
**Disclaimer: this is an opt-in experimental feature. Use it at your own risk.**

> We want to test this feature internally first. Until then, we discourage you to try it out.

TL;DR: this feature changes a bit how developers log in when developing the application. It also requires to provide additional information in the custom application config.
These changes are part of a bigger topic related to new changes and improvements around Custom Applications. So I'm not going to spoil too much.

Stay tuned for more upcoming information.


packages/application-config/schema.json
```diff
@@ -17,10 +17,6 @@
     }
   },
   "properties": {
-    "id": {
```
emmenko, line 20: This has moved to the `env.production.applicationId` field (it wasn't used before anyway).

packages/application-config/src/process-config.ts
```diff
@@ -49,15 +47,25 @@ const processConfig = ({
   const additionalAppEnv = appConfig.additionalEnv ?? {};
   const revision = (additionalAppEnv.revision as string) ?? '';
 
-  const appEnvConfig = isProd ? appConfig.env.production : developmentConfig;
+  // Feature flags
+  const isOidcForDevelopmentEnabled =
+    processEnv.ENABLE_OIDC_FOR_DEVELOPMENT === 'true' ||
+    // @ts-expect-error
+    processEnv.ENABLE_OIDC_FOR_DEVELOPMENT === true;
```
emmenko, line 54: Opt-in feature

packages/application-shell/src/apollo-links/header-link.ts
```diff
@@ -68,10 +68,16 @@ const headerLink = new ApolloLink((operation, forward) => {
...
+    STORAGE_KEYS.SESSION_TOKEN
+  );
 
   operation.setContext({
     credentials: 'include',
-    headers: omitEmpty({
+    headers: omitEmpty<Headers>({
+      [SUPPORTED_HEADERS.AUTHORIZATION]: sessionToken
+        ? `Bearer ${sessionToken}`
+        : undefined,
```
emmenko, line 80: Here we send the session token as  an Authorization header.

packages/application-shell/src/components/application-shell/redirect-to-login.tsx
```diff
@@ -0,0 +1,100 @@
...
+    `${STORAGE_KEYS.NONCE}_${nonce}`,
+    JSON.stringify(state)
+  );
+  return nonce;
+};
+
+const RedirectToLogin = () => {
+  const location = useLocation();
+
+  if (window.app.__DEVELOPMENT__) {
```
emmenko, line 36: For local development, we send the construct the authorize request and redirect to the real login page.

packages/application-shell/src/components/authenticated/auth-callback.tsx
```diff
@@ -0,0 +1,118 @@
...
+        console.warn(
+          `Cannot parse session state for "${key}".\n${unparsedSessionState}`
+        );
+      }
+    }
+  }
+  return null;
+};
+
+const AuthCallback = () => {
```
emmenko, line 48: This is the component that handles the auth callback redirect (after login). Most of the things here are very similar to what we also do for the SSO callback.

cypress/integration/playground/application-shell.js
```diff
@@ -1,51 +1,75 @@
...
-import { URL_BASE, URL_STATE_MACHINES } from '../../support/urls';
+import {
+  URL_BASE,
+  URL_STATE_MACHINES,
+  ENTRY_POINT_STATE_MACHINES,
+} from '../../support/urls';
 
 describe('when user is authenticated', () => {
+  beforeEach(() => {
+    cy.loginByOidc({ entryPointUriPath: ENTRY_POINT_STATE_MACHINES });
```
emmenko, line 11: This is a new command (see `@commercetools-frontend/cypress` package)

cypress/integration/playground/application-shell.js
```diff
@@ -1,51 +1,75 @@
...
+  });
   it('should log out with reason "user"', () => {
-    cy.login({ redirectToUri: URL_STATE_MACHINES });
-
     cy.findByRole('button', { name: /open user settings menu/i }).click();
-    cy.findByRole('link', { name: /logout/i }).click();
```
emmenko, line 11: We don't have a logout page anymore, so we simply check that the link is correct (as it redirects to another domain)

packages/application-shell/src/components/application-shell-provider/application-shell-provider.tsx
```diff
@@ -56,11 +57,12 @@ const ApplicationShellProvider = <AdditionalEnvironmentProperties extends {}>(
               <Router history={ApplicationShellProvider.history}>
                 <GtmBooter trackingEventList={props.trackingEventList || {}}>
                   <Authenticated
+                    locale={browserLocale}
+                    applicationMessages={props.applicationMessages}
```
emmenko, line 61: Needed for rendering the "error page" for the `/auth/callback` route.

packages/application-shell/src/components/authenticated/auth-callback-error-page.tsx
```diff
@@ -0,0 +1,95 @@
...
+  box-sizing: border-box;
+  width: 100%;
+  margin: 0;
+  border: 0;
+  border-style: solid;
+  border-top-width: 1px;
+  border-top-color: ${customProperties.colorNeutral};
+`;
+
+const AuthCallbackErrorPage = (props: TProps) => {
```
emmenko, line 42: The structure is very similar to our internal "PublicPageContainer" component.

packages/application-shell/src/components/authenticated/auth-callback-error-page.tsx
```diff
@@ -0,0 +1,95 @@
...
+                <Card>
+                  <Spacings.Stack scale="l">
+                    <Spacings.Inline justifyContent="center">
+                      <img
+                        src={FailedAuthenticationSVG}
+                        alt="Failed authentication"
+                      />
+                    </Spacings.Inline>
+                    <Text.Headline as="h2">
+                      {'Authentication error'}
```
emmenko, line 66: I don't think we need to translate these messages for now:

* it's still an experimental feature
* it's only used for development.

packages/application-shell/src/components/authenticated/auth-callback.tsx
```diff
@@ -0,0 +1,113 @@
...
+        console.warn(
+          `Cannot parse session state for "${key}".\n${unparsedSessionState}`
+        );
+      }
+    }
+  }
+  return null;
+};
+
+const AuthCallback = (props: TProps) => {
```
emmenko, line 50: If you're familiar with the MC SSO callback component, this is very similar: we parse the token from the hash fragment, decode it, validate the nonce, etc.

packages/application-shell/src/components/authenticated/has-cached-authentication-state.ts
```diff
@@ -1,6 +1,53 @@
+import type { ApplicationWindow } from '@commercetools-frontend/constants';
+
 import { STORAGE_KEYS } from '../../constants';
+import { buildOidcScope } from './helpers';
+
+declare let window: ApplicationWindow;
 
-const hasCachedAuthenticationState = () =>
-  window.localStorage.getItem(STORAGE_KEYS.IS_AUTHENTICATED) === 'true';
+const hasCachedAuthenticationState = (): boolean => {
```
emmenko, line 8: I should probably add some unit tests for this function.

packages/application-shell/src/components/project-switcher/project-switcher.tsx
```diff
@@ -168,11 +172,19 @@ const ProjectSwitcher = (props: Props) => {
         aria-labelledby="project-switcher"
         onChange={(event) => {
           const selectedProjectKey = event.target.value;
-          if (selectedProjectKey !== props.projectKey)
+          if (selectedProjectKey !== props.projectKey) {
+            if (window.app.__DEVELOPMENT__) {
+              window.localStorage.setItem(
+                STORAGE_KEYS.ACTIVE_PROJECT_KEY,
+                selectedProjectKey
+              );
```
emmenko, line 180: Doing this triggers a new login flow because the session scope has changed.

packages/application-shell/src/components/redirector/redirector.tsx
```diff
@@ -9,29 +9,41 @@ import { location } from '../../utils/location';
...
+  queryParams?: QueryParams;
 };
 
-const redirectTo = (targetUrl: string) => location.replace(targetUrl);
+const redirectTo = (targetUrl: string) => {
+  location.replace(targetUrl);
+};
 
 const Redirector = (props: Props) => {
-  React.useEffect(() => {
```
emmenko, line 22: We don't need all those variables to be defined inside the `useEffect`, hence this change.

packages/application-shell/src/configure-store.ts
```diff
@@ -51,10 +54,23 @@ const mergeObjectValues = <T>(object: { [key: string]: T }) =>
...
+    [SUPPORTED_HEADERS.X_APPLICATION_ID]: window.app.applicationId,
+    [SUPPORTED_HEADERS.X_TEAM_ID]: selectTeamIdFromLocalStorage(),
+  });
+};
+
 const sdkMiddleware = createSdkMiddleware({
   getCorrelationId: patchedGetCorrelationId,
   getProjectKey: selectProjectKeyFromUrl,
-  getTeamId: selectTeamIdFromLocalStorage,
+  getAdditionalHeaders,
```
emmenko, line 73: We're passing now a different function as it better encapsulates the required headers.

packages/cypress/src/add-commands/index.ts
```diff
@@ -0,0 +1,65 @@
...
+  'loginByOidc',
+  ({ entryPointUriPath }: { entryPointUriPath: string }) => {
+    Cypress.log({ name: 'loginByOidc' });
+
+    const projectKey = Cypress.env('PROJECT_KEY');
+    const sessionNonce = uuidv4();
+
+    cy.task('customApplicationConfig', {
+      entryPointUriPath,
+    }).then((appConfig: ApplicationConfig['env']) => {
```
emmenko, line 22: We're using a [task](https://docs.cypress.io/api/commands/task.html) to be able to perform nodejs operations, specifically using the filesystem.

Essentially we want to read up the runtime environment config (which is what is available at `window.app`) as the first thing, as we need to get some of the values to construct the request to fetch a token.

packages/cypress/src/add-commands/index.ts
```diff
@@ -0,0 +1,65 @@
...
+              `${STORAGE_KEYS.NONCE}_${sessionNonce}`,
+              JSON.stringify({ applicationId, query: {} })
+            );
+            win.sessionStorage.setItem(
+              STORAGE_KEYS.SESSION_SCOPE,
+              sessionScope
+            );
+          },
+        });
+      });
```
emmenko, line 62: The token response contains a redirect link. We're visiting that link, which contains the session token as hash fragments, etc.

Before doing that though, we need to prepare the browser storage with some values, as otherwise the auth callback will fail to verify that the request is legit.

playground/package.json
```diff
@@ -5,9 +5,9 @@
   "private": true,
   "scripts": {
     "build": "mc-scripts build",
-    "start": "dotenv -- mc-scripts start",
+    "start": "dotenv -c development -- mc-scripts start",
```
emmenko, line 8: https://www.npmjs.com/package/dotenv-cli#cascading-env-variables

packages/application-shell/src/components/authenticated/has-cached-authentication-state.ts
```diff
@@ -1,6 +1,53 @@
...
+        return false;
+      }
+      const activeProjectKey = window.localStorage.getItem(
+        STORAGE_KEYS.ACTIVE_PROJECT_KEY
+      );
+      if (activeProjectKey) {
+        // The application is not requesting a project key, therefore
+        // we assume that the application does not need a project context
+        // and we can remove the cached project key.
+        // This is usually the case for applications like account.
```
adnasa, line 25:
```suggestion
        // GIVEN The application is not requesting a project key,
        // THEN we assume that the application does not need a project context
        // GIVEN The application is not requesting a project key,
        // we remove the cached project key.
        // This is the case of an application like `account`.
```

packages/application-shell/src/components/authenticated/helpers.ts
```diff
@@ -0,0 +1,34 @@
...
+  if (options.projectKey) {
+    projectClaims.push(`${OIDC_CLAIMS.PROJECT_KEY}:${options.projectKey}`);
+    projectClaims.push(
+      ...(window.app.__DEVELOPMENT__?.permissions?.view ?? []).map(
+        (scope) => `${OIDC_CLAIMS.VIEW}:${scope}`
+      )
+    );
+    projectClaims.push(
+      ...(window.app.__DEVELOPMENT__?.permissions?.manage ?? []).map(
+        (scope) => `${OIDC_CLAIMS.MANAGE}:${scope}`
```
adnasa, line 20: I expected this to be `<project-key>:<oauth-scope>` not `<scope>:<oauth-scope>`.
it looks like this is by design so I could've mixed things up

adnasa (commented):
it's hard to follow the user flow on how works.
I doubt I can find time to test this out locally.

A walk through would be awesome.
thanks!

packages/application-config/src/process-config.ts
```diff
@@ -49,15 +49,27 @@ const processConfig = ({
...
```
tdeekens, line 56: You might be able to just `JSON.parse(processEnv.ENABLE...)` to get the `string | boolean` into what you want: boolean.

packages/application-config/src/process-config.ts
```diff
@@ -69,18 +81,45 @@ const processConfig = ({
...
       ...omitEmpty(additionalAppEnv),
-      ...(applicationId ? { applicationId } : {}),
+      // TODO: how else should we provide the app identifier?
+      applicationId,
       applicationName: appConfig.name,
       entryPointUriPath: appConfig.entryPointUriPath,
+      ...(!isProd && isOidcForDevelopmentEnabled
+        ? {
+            __DEVELOPMENT__: omitEmpty({
+              authorizeUrl: `${mcApiUrl.protocol}//${mcApiUrl.host.replace(
```
tdeekens, line 113: Should we explicitly define `authorizeUrl` or is there a benefit from inferring it.

packages/application-config/src/schema.ts
```diff
@@ -26,14 +22,44 @@ export interface JSONSchemaForCustomApplicationConfigurationFiles {
    * The URL of the Merchant Center API. We usually recommend to use the `cloudIdentifier` option to avoid possible typos.
    */
   mcApiUrl?: string;
+  /**
+   * The permissions configuration for the required OAuth Scopes for this Custom Application.
+   */
+  permissions?: {
```
tdeekens, line 28: Should we call this scopes? Just thinking out loud. Permissions infer scopes in the MC. This doesn't infer anything it's just scopes. Without the indirection of permissions.

packages/application-shell/src/apollo-links/header-link.ts
```diff
@@ -21,7 +21,7 @@ declare let window: ApplicationWindow;
 type ApolloContextWithInMemoryCache = TApolloContext & {
   cache: ApolloClient<NormalizedCacheObject>;
 };
-
+type Headers = { [key: string]: string };
```
tdeekens, line 24: Nit: `Record<string, string>`.

packages/application-shell/src/components/application-shell/redirect-to-login.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+    // According to the OIDC spec, the `state` parameter is recommended to be sent
+    // to the authorization server to help mitigating Cross-Site Request Forgery.
+    // https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
+    // Using `state` or `nonce` is very similar but with subtle differences.
+    // Here we follow the approach described by Auth0 on how to use both, where
+    // we generate a `nonce`, store it in session storage together with some state,
+    // and send it to the authorization server as the `state` parameter.
+    // https://auth0.com/docs/protocols/oauth2/redirect-users
+    // Additionally, we still send the `nonce` parameter as well.
```
tdeekens, line 52: 👍🏻

packages/application-shell/src/components/application-shell/redirect-to-login.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+    // Store session scopes, to be able to detect if requested scopes changed
+    // in the application config and invalidate the session.
+    // This is only valid for local development.
+    window.sessionStorage.setItem(STORAGE_KEYS.SESSION_SCOPE, requestedScope);
+
+    return (
+      <Redirector
+        to="login/authorize"
+        origin={window.app.__DEVELOPMENT__.authorizeUrl}
```
tdeekens, line 71: Can you elaborate why we scope this under `__DEVELOPERMENT__` and not e.g. `oidcConfig`?

packages/application-shell/src/components/application-shell/redirect-to-logout.tsx
```diff
@@ -1,22 +1,42 @@
...
   const servedByProxy = useApplicationContext(
     (context) => context.environment.servedByProxy
   );
+
+  if (window.app.__DEVELOPMENT__) {
+    // Remove the `sessionToken` from storage, so that the AppShell can initiate
+    // a new authorization flow.
+    window.sessionStorage.removeItem(STORAGE_KEYS.SESSION_TOKEN);
+    window.sessionStorage.removeItem(STORAGE_KEYS.SESSION_SCOPE);
+    window.sessionStorage.removeItem(STORAGE_KEYS.IS_AUTHENTICATED);
```
tdeekens, line 30: Nit: Maybe we can have a small `development-oidc-localstorage.ts` module which allows setting and unsetting things. Just to not spread knowledge o the keys into different parts of the shell.

packages/application-shell/src/components/authenticated/authenticated.tsx
```diff
@@ -1,23 +1,61 @@
...
+
+  // Fall back to check for cookie-based authentication.
   return <AmILoggedIn {...props} />;
 };
 Authenticated.displayName = 'Authenticated';
 
-export default Authenticated;
+const AuthenticationRoutes = (props: TProps) => (
+  <Switch>
+    <Route path={`/account/auth/callback`}>
```
tdeekens, line 42: `auth` or `authorize`? Is it the same or is `auth` preferred?

packages/application-shell/src/components/authenticated/helpers.ts
```diff
@@ -0,0 +1,34 @@
...
+  if (teamId) {
+    projectClaims.push(`${OIDC_CLAIMS.TEAM_ID}:${teamId}`);
+  }
+  return [
+    // This is required as per OIDC spec.
+    OIDC_CLAIMS.OPEN_ID,
+    // Custom claims
+    ...projectClaims,
+  ].join(' ');
+};
```
tdeekens, line 34: How would we test store scopes with this? Is it intended or a possible scenario?

packages/cypress/README.md
```diff
@@ -0,0 +1,39 @@
...
+## Usage
+
+This package extends Cypress' `cy` command.
+
+Add this line to your project's `cypress/plugins/index.js`:
+
+```javascript
+const customApplications = require('@commercetools-frontend/cypress/task');
+
+module.exports = (on, config) => {
```
tdeekens, line 24: Not sure it applies but I still would recommend to install like this:

```js
customApplicationsPlugin.install(on, config);
```

This gives more freedom in the future. You can pass config options potentially and have access to the whole "on".

tdeekens (commented):
Impressive! I did a first pass review but plan to do a second.

cypress/integration/playground/state-machines.js
```diff
@@ -1,18 +1,20 @@
-import { URL_STATE_MACHINES, URL_STATE_MACHINES_ID } from '../../support/urls';
+import {
+  URL_STATE_MACHINES_ID,
+  ENTRY_POINT_STATE_MACHINES,
+} from '../../support/urls';
 
 describe('State machines', () => {
+  beforeEach(() => {
+    cy.loginByOidc({ entryPointUriPath: ENTRY_POINT_STATE_MACHINES });
```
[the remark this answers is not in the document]
emmenko: The problem is when having multiple applications in the same repository.

packages/application-config/src/process-config.ts
emmenko: We should infer it, because the environment (staging, production, etc) has to be the same with the APIs that the application uses.

packages/application-config/src/schema.ts
emmenko: Initially I went with this because it's more or less what we wanted to show in the UI (registration form).
We then changed it and went with view/manage oauth scopes.

I'm fine remaining this, if you think it's confusing. Thanks for the reminder.

packages/application-shell/src/components/application-shell/redirect-to-login.tsx
emmenko: Because it's a feature that should only be used in development (for now). So having `__DEVELOPMENT__` makes things more clear (that was the intention anyway).

Do you see other reasons for not using `__DEVELOPMENT__`?

packages/application-shell/src/components/application-shell/redirect-to-logout.tsx
emmenko: Thanks, I'll see what I can do.

packages/application-config/src/schema.ts
tdeekens: I think it would help. The confusing bit could be otherwise that you might think that by adding "orders" here under "view" that I'd get the scopes the View Orders permission implies: e.g. types. That's not true tho.

packages/application-shell/src/components/application-shell/redirect-to-login.tsx
tdeekens: Not strictly. I just stumbled across it. Can't come up with a better indication of that either.

packages/application-shell/src/components/authenticated/authenticated.tsx
emmenko: Good question.

Usually `/authorize` is well defined in the spec as the endpoint to start the authentication flow.
The `/auth/callback` is usually left to the user to decide.

I don't have a particular opinion about the name, it's also something we can manage internally.

Do you have suggestions? We can also change it to `/oidc/callback`

packages/application-shell/src/components/authenticated/helpers.ts
emmenko: I think it can be defined as a normal scope.

tdeekens: I scope is `manage_orders:{projectKey}:{storeKey}`. So if I add `view.[0] = 'orders:storeA` would that work?

cypress/integration/playground/application-shell.js
```diff
@@ -1,51 +1,75 @@
...
+
+describe('failed OIDC authentication', () => {
+  describe('when sessionToken is missing', () => {
+    it('should show auth callback error page', () => {
+      cy.visit(`/${URL_STATE_MACHINES}/auth/callback`);
+      cy.findByText('Authentication error');
+      cy.findByText(/missing sessionToken/i);
+      cy.percySnapshot();
+    });
+  });
```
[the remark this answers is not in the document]
emmenko: Oh...well just check Percy 😅

packages/application-config/src/process-config.ts
```diff
@@ -69,18 +81,45 @@ const processConfig = ({
...
+
+  // The real application ID is only used in production.
+  // In development, we prefix the entry point with the "__local" prefix.
+  // This is important to determine to which URL the MC should redirect to
+  // after successful login.
+  let applicationId: string | undefined = isOidcForDevelopmentEnabled
+    ? `__local:${appConfig.entryPointUriPath}`
+    : undefined;
+  if (isProd) {
+    // TODO: decide if we do require the application ID or not.
```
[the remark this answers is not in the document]
emmenko: Yes that was the idea. I still think it's valid, we can still discuss it maybe just to be sure.

packages/application-config/test/process-config.spec.js
```diff
@@ -61,6 +62,7 @@ describe('processing a simple config', () => {
           applicationName: 'avengers-app',
           entryPointUriPath: 'avengers',
           cdnUrl: 'https://avengers.app/',
+          entryPointUriPath: 'avengers',
```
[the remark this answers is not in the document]
emmenko: Ah no, good catch

packages/application-shell/src/components/authenticated/authenticated.tsx
tdeekens: I like `oidc/calback` to be honest. It clearly shows what auth flow the callback is for. Also if that deviates a bit from the norm.

packages/application-shell/src/components/authenticated/helpers.ts
emmenko: Yes the idea is to have well-defined claims:
* `project_key`
* `view`
* `manage`

Having `<project-key>:<oauth-scope>` makes it harder to parse, because you always assume that the left-hand value is the project key.

packages/application-shell/src/components/authenticated/helpers.ts
emmenko: I double checked. It's not supported at the moment.

I think we can fix this in the backend.

So you would still pass the oauth scopes as `view_products` or `view_orders:storeA`.

packages/cypress/README.md
emmenko: I double checked this. We're exporting a `task`, which already implies that in the `cypress/plugins/index.js` you have to use `on('task', {...})`.

It does not make much sense to pass `on` to the task itself.

If we were to implement a plugin, then I guess it makes sense to do as you suggested.

packages/application-config/src/schema.ts
emmenko: Changed it: https://github.com/commercetools/merchant-center-application-kit/pull/1934/commits/43934feb54e6fd7d2ae7ddbbf5cb60be557a5aec

packages/application-shell/src/components/authenticated/authenticated.tsx
emmenko: https://github.com/commercetools/merchant-center-application-kit/pull/1934/commits/496d5add7b13982ba5e2847555bf08e95e75db18

packages/application-shell/src/components/application-shell/redirect-to-logout.tsx
emmenko: https://github.com/commercetools/merchant-center-application-kit/pull/1934/commits/05a7dd9668a0285e0a4fef80e637c6be1c4cf0c3

tdeekens (approved):
💯 🏩 ⭐

PR #1945, opened 2020-12-28
refactor(app-shell): remove updateFlags

tdeekens:
#### Summary

This removes the `updateFlags` re-export from the `application-shell` package.

The re-export was not used internally and I presume nobody used it externally. The alternative is to use `ldAdapter.updateFlags`. The stand alone export was only a "redirect" anyway.

This re-export will not be present anymore in the upcoming major release of flopflip. Please apply the changeset you see adequate for this.


emmenko (approved):
Thanks!

PR #1948, opened 2020-12-29
feat(app-shell): pre-configure app entry point routes when using children instead of render prop

### LINKED ISSUE #1886: [RFC] Let <ApplicationShell> pre-configure the application main route

All applications have this `entry-point.js` file where the `<ApplicationShell>` is rendered and so the application routes.

For example:

https://github.com/commercetools/merchant-center-application-kit/blob/63b684135d0e77ebdac9e44bb6e2ed2641ec7f92/application-templates/starter/src/components/entry-point/entry-point.js#L20-L40

This is the "usual" setup. There is a main route that maps to the application entry point, which then renders the actual application.
Here the main route should obviously match what the application `entryPointUriPath` is in the application config.

I think at this point we can make the `<ApplicationShell>` a bit smarter, in that we provide this preconfiguration which is based on the application config (source of truth).

To be backwards compatible, we can maybe introduce this feature using the `children` prop. So you can pass the application component and render it as a child, or keep using the `render` prop but then configure things on your own.

Taking the starter app example, it would look something like this:

```js
const AsyncApplicationRoutes = React.lazy(
  () => import('../../routes' /* webpackChunkName: "starter-routes" */)
);

class EntryPoint extends React.Component {
  static displayName = 'EntryPoint';
  render() {
    return (
      <ApplicationShell
        environment={window.app}
        onRegisterErrorListeners={({ dispatch }) => {
          Sdk.Get.errorHandler = (error) =>
            globalActions.handleActionError(error, 'sdk')(dispatch);
        }}
        applicationMessages={loadMessages}
        DEV_ONLY__loadNavbarMenuConfig={() =>
          import('../../../menu.json').then((data) => data.default || data)
        }
        featureFlags={FEATURE_FLAGS}
      >
        <AsyncApplicationRoutes />
      </ApplicationShell>
    );
  }
}
```

That's it.

emmenko:
Closes #1886

This PR implements the proposal described in #1886


.changeset/quick-dragons-deliver.md
```diff
@@ -0,0 +1,76 @@
...
+-    render={() => <ApplicationStarter />}
+-  />
++  >
++    <AsyncApplicationRoutes />
++  </ApplicationShell>
+);
+```
+
+Furthermore, the `test-utils` of the `@commercetools-frontend/application-shell` have now a new option to enable this opt-in behavior of rendering the application with pre-configured routes.
+To enable this option, pass the `renderEntryPointRoutes: true` to the `renderApp` or `renderAppWithRedux` functions.
```
tdeekens, line 63: Nitty: `renderEntryPointRoutes` sounds a bit like a render-prop to me while it obviously is a boolean option. I assume we would like to enable this behaviour by default at some point. I personally (subjective obviously) started to like `disableX` for such things. So this would be defaulting to `disableAutomaticEntryPointRoutes` which defaults to `true`. One can then opt-into the new behaviour while we with some breaking release only claim that the option was now removed and is always is true.

application-templates/starter/src/components/main-view/main-view.spec.js
```diff
@@ -1,14 +1,10 @@
...
-} from '@commercetools-frontend/application-shell/test-utils';
-import { ApplicationStarter } from '../entry-point';
+import { fireEvent } from '@commercetools-frontend/application-shell/test-utils';
+import { renderApplication } from '../../test-utils';
 
 describe('main view', () => {
   it('the user can click on the link to "one" and the page should show a text with "View one"', async () => {
     const initialRoute = '/my-project/examples-starter';
-    const rendered = renderApp(<ApplicationStarter />, {
+    const rendered = renderApplication(null, {
```
tdeekens, line 7: Null or empty fragment?

packages/application-shell/src/components/application-entry-point/application-entry-point.tsx
```diff
@@ -0,0 +1,58 @@
...
+          process.env.NODE_ENV === 'production' ? null : (
+            <Redirect
+              exact={true}
+              from="/:projectKey"
+              to={`/:projectKey/${entryPointUriPath}`}
+            />
+          )
+        }
+        <Route path={`/:projectKey/${entryPointUriPath}`}>
+          {props.children}
```
tdeekens, line 36: We could use React.Children.only<React.ReactNode>(props.children);` to verify that children has only one child (a React element) and return it otherwise throw an error.

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -179,18 +182,29 @@ afterEach(() => {
...
-        (context) => context.environment.applicationName
-      );
-      return <p>{`Application name: ${applicationName}`}</p>;
-    };
-    const rendered = renderApp(<TestComponent />);
-    await rendered.findByText('Application name: my-app');
-  });
-});
+describe.each`
+  renderNodeAsChildren | route
```
tdeekens, line 186: 💯

packages/application-shell/src/components/application-entry-point/application-entry-point.tsx
```diff
@@ -0,0 +1,58 @@
...
+        </Route>
+        {/* Catch-all route */}
+        <RouteCatchAll />
+      </Switch>
+    );
+  }
+
+  // We still support the `render` prop, for backwards compatibility
+  // and for having more control in certain cases (mostly to some of our internal apps).
+  if (props.render) {
```
tdeekens, line 46: Maybe `typeof props.render === 'function'`.

packages/application-shell/src/test-utils/test-utils.tsx
```diff
@@ -416,7 +426,16 @@ function renderApp<AdditionalEnvironmentProperties = {}>(
           >
             <Router history={history}>
               <React.Suspense fallback={<LoadingFallback />}>
-                {props.children}
+                <ApplicationEntryPoint
+                  environment={mergedEnvironment}
+                  render={
+                    renderEntryPointRoutes
+                      ? undefined
+                      : () => <>{props.children}</>
```
tdeekens, line 434: Should we exact this? It unfortunate to wrap `props.children` in a render-prop. Maybe `const renderOrChildren = { !renderEntryPointRoutes ? children : render: undefined } to then `...renderOrChildren`? Not entirely sure myself.

.changeset/quick-dragons-deliver.md
emmenko: Yeah I couldn't come up with a better name. Thanks for the suggestion, I like it!

packages/application-shell/src/components/application-entry-point/application-entry-point.tsx
```diff
@@ -0,0 +1,58 @@
...
+  render?: () => JSX.Element;
+  children?: React.ReactNode;
+};
+
+const ApplicationEntryPoint = <AdditionalEnvironmentProperties extends {}>(
+  props: Props<AdditionalEnvironmentProperties>
+) => {
+  // If the `children` prop is used (instead of the `render` prop),
+  // we pre-configure the application entry point routes to avoid
+  // users to do so on their own.
```
adnasa, line 19: 💯

adnasa (approved):
LGTM

.changeset/quick-dragons-deliver.md
emmenko: I changed it as per suggestion.

application-templates/starter/src/components/main-view/main-view.spec.js
emmenko: I think `null` is fine, as it's what you would return when you don't want to render something.

packages/application-shell/src/test-utils/test-utils.tsx
emmenko: I think I'll keep it like this, it's a bit more explicit and things stay simple.

tdeekens (approved):
💯

application-templates/starter/src/test-utils/index.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import {
+  renderApp,
+  renderAppWithRedux,
+} from '@commercetools-frontend/application-shell/test-utils';
+import { AsyncApplicationRoutes } from '../components/entry-point';
```
tdeekens, line 6: @emmenko follow-up: we shouldn't render async components here I think. We had issues with that internally.

PR #1950, opened 2020-12-30
chore: move renovate and kodiak config to `.github`

tdeekens:
#### Summary

Relatively recently `kodiak` and `renovate` both added support to move their configs to the `.github` folder. I think it makes sense. It clearly shows that it's not a "repository feature" but relates to GitHub.

While doing so the `renovaterc.json` as the be renamed to `renovate.json` as `*rc.json` is not supported anymore inside the `.github` folder. I think it's inline with recent trends to not name things `rc` files anymore when they are really run commands.

While moving the Renovate config I would kindly ask to split out flopflip updates from Renovate. It gives me personally some breathers at times when I experiment with flopflip.


emmenko (approved):
Looks good, thanks!

PR #1952, opened 2021-01-04
chore: add and fix manypkg

### LINKED ISSUE #1951: Add `@manypkg/cli` to validate and check mono-repo
# Summary

`@manypkg/cli` (https://github.com/Thinkmill/manypkg) is a tool to validate mono-repos. It checks dependency versions and the presence of a `devDependency` for each `peerDependency` among other things. 

# Description

I recently added `@manypkg/cli` to `@flopflip` and was quite surprised by the things I did wrong. Mainly lack of `devDependencies` or wrong ranges. `@manypkg/cli` is built by the same people as `preconstruct` from what I can tell while `preconstruct` uses it too.

If I run `@manypkg/cli` on this repo I get roughly this output.

CleanShot 2021-01-04 at 09 28 58@2x

A nice things about it is that most rules it currently has, which I think will become more, are auto-fixable.

tdeekens:
Closes #1951 

#### Summary

This pull request adds `@manypkg/cli` as suggested in #1951.

#### Old architecture

The dependencies and ranges of each workspace were not validated.

#### New architecture

The dependencies of all workspaces are checked. Please check the rules manypkg lists on their readme.


packages/application-shell/package.json
```diff
@@ -82,8 +93,8 @@
     "is-retina": "1.0.3",
     "lodash": "4.17.20",
     "memoize-one": "5.1.1",
-    "moment": "^2.24.0",
-    "moment-timezone": "^0.5.26",
+    "moment": "2.29.1",
+    "moment-timezone": "0.5.32",
```
emmenko, line 97: Any particular reason for removing the range?

packages/i18n/package.json
```diff
@@ -40,11 +40,12 @@
     "@commercetools-frontend/sentry": "17.7.1",
     "@commercetools-uikit/i18n": "10.42.2",
     "@emotion/react": "11.1.4",
-    "@types/prop-types": "15.7.3",
-    "moment": "^2.24.0",
+    "@types/prop-types": "^15.7.3",
+    "moment": "2.29.1",
```
emmenko, line 44: Same here

packages/l10n/package.json
```diff
@@ -27,13 +38,14 @@
     "@babel/runtime-corejs3": "7.12.5",
     "@commercetools-frontend/sentry": "17.7.1",
     "@emotion/react": "11.1.4",
-    "@types/prop-types": "15.7.3",
+    "@types/prop-types": "^15.7.3",
     "lodash": "4.17.20",
-    "moment": "^2.24.0",
+    "moment": "2.29.1",
```
emmenko, line 43: ⬆️

packages/react-notifications/package.json
```diff
@@ -35,16 +45,22 @@
...
     "@types/history": "4.7.6",
-    "@types/lodash": "4.14.167",
-    "@types/prop-types": "15.7.3",
+    "@types/lodash": "^4.14.167",
+    "@types/prop-types": "^15.7.3",
     "lodash": "4.17.20",
-    "moment": "^2.24.0",
-    "moment-timezone": "^0.5.26",
+    "moment": "2.29.1",
+    "moment-timezone": "0.5.32",
```
emmenko, line 52: ⬆️

emmenko (commented):
Thanks for giving this a shot. I tried this when we introduced preconstruct but there were some weird errors (I think most of them related to emotion).

I guess now things are looking better.

I just left some comments regarding changes of moment version range, otherwise the rest looks good 👌

packages/application-shell/package.json
tdeekens: I had quite big trouble when updating deps cause `@types` had ranges and resolved to different versions has a result. Those I decided to pin. We can unpin moment if you want. Do you remember what the reson was not to pin it?

emmenko: The only reason was to be able to use newer versions of moment from the consumer side, for example like we do with uikit etc.

If there are weird issues we can keep it pinned, we just need to be aware of it.

tdeekens:
@emmenko moment unpinned. Thanks for noticing.

emmenko (approved):
💯

PR #1953, opened 2021-01-04
refactor: to update @flopflip to v12

tdeekens:
#### Summary

This pull request updates `@flopflip` to v12. The holiday release 🎉.

#### Description

For us this release does not contain breaking changes apart from the already taken care of one in https://github.com/commercetools/merchant-center-application-kit/pull/1945.

This release mainly enables combining adapters which we can evaluate in another step. For now the flag state will be split as this

CleanShot 2021-01-04 at 10 53 28@2x

Note the `launchdarkly` namespacing which prior was like this

CleanShot 2021-01-04 at 10 51 44@2x

Additional benefit is that `@flopflip` now types the user of any adpater (e..g LaunchDarkly) depending on the underling types [here](https://github.com/launchdarkly/js-sdk-common/blob/912bf8c370362dd48b8745032bf62cee720fac45/typings.d.ts#L267).


packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
```diff
@@ -17,6 +18,15 @@ type Props = {
   shouldDeferAdapterConfiguration?: boolean;
 };
 
+type CustomLDUser = {
```
tdeekens, line 21: This is out type we pass to `omitEmpty` so it can infer the types correctly from how I understand the library.

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
```diff
@@ -44,21 +54,23 @@ export const SetupFlopFlipProvider = (props: Props) => {
...
+        // This is mostly useful for internal usage on our staging environments.
+        clientSideId: props.ldClientSideId ?? ldClientSideIdProduction,
+      },
       flags,
       user: {
-        key: props.user && props.user.id,
-        custom: {
-          id: props.user && props.user.launchdarklyTrackingId,
+        key: props.user?.id,
+        custom: omitEmpty<CustomLDUser, Partial<CustomLDUser>>({
```
tdeekens, line 66: The `Partial` makes all properties optional which is the `Input` type.

@types/omit-empty-es/index.d.ts
```diff
@@ -1,4 +0,0 @@
-declare module 'omit-empty-es' {
```
tdeekens, line 1: This is not needed anymore to how I understand it. The issue with having it is that TypeScript prefers the `@types` over any bundled type of the `omit-empty-es`-library. It shadows the actual types which have changed and uses these outdated types which do not support a generic and resolve to `Record<string, unknown>` which we can't assign to anything.

emmenko: Yes thanks

emmenko (approved):
Thanks, looks good!

emmenko:
@tdeekens ah, don't forget the changeset.

PR #1954, opened 2021-01-04
chore: update deps for release

tdeekens:
#### Summary

Updates dependencies. Let's see how much TypeSript and different `@types/*` packages likes us here :)


packages/jest-preset-mc-app/setup-tests.js
```diff
@@ -13,6 +14,7 @@ global.window.app = {
 };
 
 window.MutationObserver = MutationObserver;
+global.Headers = global.Headers || Headers;
```
tdeekens, line 17: Same as we have in the mc services. This is needed as a transient library assumes that headers are globally available but they're not with recent versions of jsdom.

Issue #1829, opened 2020-10-26
Unreleased breaking changes


emmenko:
This issue is meant to serve as a TODO list for possible future breaking changes, that have not been planned yet.

- [ ] TypeScript v4 (#1695)
- [ ] React Router to v6 + History to v5 (#1575) 
- [ ] GraphQL to v15
- [ ] Remove deprecated hostnames
  - https://github.com/commercetools/merchant-center-application-kit/blob/879b4d618569659335945cab4a5ad28c0d19f5c8/packages/mc-html-template/src/process-headers.js#L72-L73
- [ ] Expose apollo hooks abstractions for each specific target (see https://github.com/commercetools/merchant-center-application-kit/pull/1677#discussion_r500211542)
- [ ] Remove `enzyme` (also from presets and so on) as we want to encourage the usage of RTL + Cypress only
  - [ ] Update docs about testing
  - [ ] address https://github.com/commercetools/merchant-center-application-kit/issues/1773

PR #1961, opened 2021-01-09
[Breaking] refactor(eslint-mc-app): simplify and improve usage of eslint config

emmenko:
Our ESLint config hasn't been updated in a while. Furthermore, in this repo we were not using the mc-app config but a "custom" one specific for typescript.

Let's change that and use our own config here as well, with support for TypeScript out of the box.

So what changed here:
* I decided to drop the airbnb base config and use instead the react-app config. The react-app config comes with some good defaults (also regarding TypeScript) and it's anyway what we look for in terms of tooling (react-scripts, babel, etc). So it's more aligned on that front as well.
* No more many required peer dependencies to be installed. I think we should make it as simple as possible for users to set up their dev environment, and having to install >10 peer dependencies does not help much. Even though ESLint recommends to ship a sharable config with peer dependencies, I think the usability is still better if we only require `eslint` as a peer dependency.
  > Note that all the dependencies are defined with a caret `^` version range, which gives a bit of flexibility to the consumers in case of version conflicts.
* Cleaned up some of the rules.
* Included some additional TypeScript config that we were using before anyway (extending what is already specified in react-app config).


tdeekens:
Nice. Let's finish internal refactors by earlier releases before loading teams with more things I'd suggest. Can you anticipate how many changes this would cause and how many are auto-fixable?

emmenko:
> Can you anticipate how many changes this would cause and how many are auto-fixable?

I tested this in our internal repo. It's all good, and it actually found a bunch of legit errors! 🤓

packages/eslint-config-mc-app/README.md
```diff
@@ -4,7 +4,7 @@
   <a href="https://www.npmjs.com/package/@commercetools-frontend/eslint-config-mc-app"><img src="https://badgen.net/npm/v/@commercetools-frontend/eslint-config-mc-app" alt="Latest release (latest dist-tag)" /></a> <a href="https://www.npmjs.com/package/@commercetools-frontend/eslint-config-mc-app"><img src="https://badgen.net/npm/v/@commercetools-frontend/eslint-config-mc-app/next" alt="Latest release (next dist-tag)" /></a> <a href="https://bundlephobia.com/result?p=@commercetools-frontend/eslint-config-mc-app"><img src="https://badgen.net/bundlephobia/minzip/@commercetools-frontend/eslint-config-mc-app" alt="Minified + GZipped size" /></a> <a href="https://github.com/commercetools/merchant-center-application-kit/blob/master/LICENSE"><img src="https://badgen.net/github/license/commercetools/merchant-center-application-kit" alt="GitHub license" /></a>
 </p>
 
-ESLint config used by a MC application.
+ESLint config used by a (Merchant Center) Custom Application.
```
adnasa, line 7: We don't need `(Merchant Center)`

website-components-playground/src/pages/form-dialog.js
```diff
@@ -3,6 +3,7 @@ import { Formik } from 'formik';
 import { FormDialog } from '@commercetools-frontend/application-components';
 import Spacings from '@commercetools-uikit/spacings';
 import TextField from '@commercetools-uikit/text-field';
+import TextInput from '@commercetools-uikit/text-input';
```
adnasa, line 6: how is this related to the eslint udpate? 🤔 
or was this something you just discovered on the way?

adnasa (approved):

emmenko:
@tdeekens @adnasa I think we're good to go now with this (major) release.

We can also include a bunch more breaking changes, for example "Remove deprecated hostnames" (https://github.com/commercetools/merchant-center-application-kit/issues/1829).

tdeekens (approved):
Nice. Thanks for iterating on this!

PR #1970, opened 2021-01-11
ci(changesets): to use conventional commits

tdeekens:
#### Summary

Suggesting to have `changesets` use a conventional commit pr title and commit.

I noticed that this breaks a husky v5 update in flopflip over the weekend.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -71,6 +71,8 @@ jobs:
         with:
           publish: yarn changeset publish
           version: yarn changeset:version-and-format
+          title: 'ci(changesets): version packages'
+          commit: 'ci(changesets): version packages'
         env:
           GITHUB_TOKEN: ${{ secrets.RELEASE_GITHUB_TOKEN }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -71,6 +71,8 @@ jobs:
         with:
           publish: yarn changeset publish
           version: yarn changeset:version-and-format
+          title: 'ci(changesets): version packages'
```
emmenko, line 74: This is the PR title, right? If so, I would keep the default.

.github/workflows/release.yml
```diff
@@ -71,6 +71,8 @@ jobs:
...
+          commit: 'ci(changesets): version packages'
```
emmenko, line 75: Thanks

.github/workflows/release.yml
tdeekens: Yup. Wasn't sure what's used via a squash merge. The first commit or the title of the pr as the commit.

emmenko (approved):

PR #1971, opened 2021-01-11
refactor: to use combine adapters with one

tdeekens:
#### Summary

This pull request experiments with the `graphql-adapter` and `combine-adapters` from `@floplfip`. 

We don't have to use this functionality but I would like to propose it through this internally.


packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
```diff
@@ -95,8 +153,9 @@ export const SetupFlopFlipProvider = (props: Props) => {
   }
 
   return (
-    <ConfigureFlopFlip<typeof ldAdapter>
-      adapter={ldAdapter}
+    <ConfigureFlopFlip<typeof combineAdapters>
+      adapter={combineAdapters}
+      // @ts-expect-error
```
tdeekens, line 158: Will have to look into it. Typing flopflip has become rocket science recently when combining adapters :)

emmenko: What's the error you get?

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
```diff
@@ -54,23 +91,44 @@ export const SetupFlopFlipProvider = (props: Props) => {
...
+            tenant: props.user?.launchdarklyTrackingTenant,
+          }),
+        },
+      },
+      graphql: {
+        user: {
+          key: props.user?.id,
+        },
+        uri: `${getMcApiUrl()}/graphql`,
+        query: 'query AllFeatures { allFeatures { name value } }',
```
emmenko, line 127: Can we put this into a `.graphql` file?

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
```diff
@@ -54,23 +91,44 @@ export const SetupFlopFlipProvider = (props: Props) => {
...
+        pollingInteralMs: 1000,
+        getRequestHeaders,
+        parseFlags,
+        cacheIdentifier: 'local',
```
emmenko, line 131: What's this for?

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
```diff
@@ -54,23 +91,44 @@ export const SetupFlopFlipProvider = (props: Props) => {
...
+        },
+      },
+      graphql: {
+        user: {
+          key: props.user?.id,
+        },
+        uri: `${getMcApiUrl()}/graphql`,
+        query: 'query AllFeatures { allFeatures { name value } }',
+        pollingInteralMs: 1000,
+        getRequestHeaders,
```
emmenko, line 129: It would be cool maybe to be able to pass the `apolloClient` here, to avoid having to configure the url, the headers, etc.

emmenko (commented):
Looks very promising. Thanks for pushing this 🤗

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
tdeekens: The adapter will save and restore priorly fetched flags in local storage.

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
tdeekens: Sorry, I have a PRs welcome for this ;)

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
tdeekens: What would be the advantage or it? It's a "raw" fetch in the end.

emmenko: So that we can validate the query

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
emmenko: I can open a PR if you want, sure.

tdeekens: I thought about it over night a bit more. What I don't like about it that `flopflip` "binds" to Apollo. I don't want it to be biased towards any GraphQL Client. Nor do I want to maintain and release breaking changes if Apollo does (as it would be a `peerDependency`). 

I somewhat like the generic nature of `fetch` and the `url`. The alternative would be to inverse control entirely. E.g. via a `fetchFlags: async () => TFlags`. Then a user could use `fetch` or any client.

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
tdeekens: Oh well.

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
tdeekens: Right now, the document node is new as the fetch expects a string.

```
packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx:161:7 - error TS2322: Type '{ user: { key: string | undefined; }; memory: { user: { key: string | undefined; }; }; launchdarkly: { sdk: { clientSideId: string; }; flags: {}; user: { key: string | undefined; custom: CustomLDUser; }; }; graphql: { ...; }; }' is not assignable to type 'TCombinedAdapterArgs'.
  Type '{ user: { key: string | undefined; }; memory: { user: { key: string | undefined; }; }; launchdarkly: { sdk: { clientSideId: string; }; flags: {}; user: { key: string | undefined; custom: CustomLDUser; }; }; graphql: { ...; }; }' is not assignable to type '{ launchdarkly?: TLaunchDarklyAdapterArgs | undefined; localstorage?: TLocalStorageAdapterArgs | undefined; memory?: TBaseAdapterArgs<...> | undefined; splitio?: TSplitioAdapterArgs | undefined; graphql?: TGraphQLAdapterArgs | undefined; }'.
    Types of property 'graphql' are incompatible.
      Type '{ user: { key: string | undefined; }; uri: string; query: DocumentNode; pollingInteralMs: number; getRequestHeaders: (props: Props) => TGraphQLAdapterRequestHeaders; parseFlags: (fetchedFlags: TFetchedFlags) => { ...; }; cacheIdentifier: string; }' is not assignable to type 'TGraphQLAdapterArgs'.
        Type '{ user: { key: string | undefined; }; uri: string; query: DocumentNode; pollingInteralMs: number; getRequestHeaders: (props: Props) => TGraphQLAdapterRequestHeaders; parseFlags: (fetchedFlags: TFetchedFlags) => { ...; }; cacheIdentifier: string; }' is not assignable to type '{ fetcher?: ((input: RequestInfo, init?: RequestInit | undefined) => Promise<Response>) | undefined; uri: string; query: string; pollingInteralMs?: number | undefined; getQueryVariables?: ((adapterArgs: TGraphQLAdapterArgs) => unknown) | undefined; getRequestHeaders?: ((adapterArgs: TGraphQLAdapterArgs) => Record<.....'.
          Types of property 'query' are incompatible.
            Type 'DocumentNode' is not assignable to type 'string'.

161       adapterArgs={adapterArgs}
          ~~~~~~~~~~~

  node_modules/@flopflip/types/dist/declarations/src/index.d.ts:186:5
    186     adapterArgs: TConfigureAdapterArgs<TAdapterInstance>;
            ~~~~~~~~~~~
    The expected type comes from property 'adapterArgs' which is declared here on type 'IntrinsicAttributes & Pick<Props<CombineAdapters>, "children" | "adapter" | "adapterArgs"> & Partial<Pick<Props<CombineAdapters>, "shouldDeferAdapterConfiguration" | "defaultFlags">> & Partial<...>'
```

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
emmenko: Yes I was also thinking about the generic approach.
I also don't want to bind the adapter to Apollo.

The main point is that the adapter should probably not be concerned about with making and configuring the HTTP request.

emmenko: So I guess it can then also be an `http-adapter` instead of something specific to graphql.

tdeekens: Yeah in the end it could be renamed. Somehow it's not related to GraphQL after all.

adnasa:
I'm trying to review the implementation but I don't understand the application of `graphql-adapter`.
What do we get with this? 🤔

tdeekens:
> I'm trying to review the implementation but I don't understand the application of `graphql-adapter`.
> What do we get with this? 🤔

The idea is to homogenise the internal use of feature flags. We also have a fe-chapter issue about the topic. 

The status quo internally is that we have:

1. Long-lived feature configuration: e.g. if the dashboard is deployed to the given environment
2. Short-lived feature flags: e.g. if platform limits are enabled

Both are rightfully configured in different systems. 1. is in our k8s files as a ConfigMap. 2. is in LaunchDarkly.

Now for either system or type of flag we use a different "backing system". For 1. we use `@flopflip` for 2. we use a `useFeatureStatus` (as we call it). This evaluates if it's possible to migrate both to only use `@flopflip`. It would load both types of flags and exposes them both to the FE. With that we could throw away all code we have around `useFeatureStatus` and remove confusion when to use what system or library.

adnasa:
aaah clear. 
thanks for clarifying!

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
```diff
@@ -27,11 +32,46 @@ type CustomLDUser = {
...
+};
+
 // This value is hard-coded here because we want to make sure that the
 // app uses our account of LD. The value is meant to be public, so there
 // is no need to be concerned about security.
 const ldClientSideIdProduction = '5979d95f6040390cd07b5e01';
 
+combineAdapters.combine([ldAdapter, graphqlAdapter]);
+
+const getRequestHeaders = (props: Props) =>
```
adnasa, line 52: apologies if I confuse things.

- the `getRequestHeaders` is supposedly called with [`adapterArgs`](https://github.com/tdeekens/flopflip/blob/8fd148b724394887e99ed9136bb309776a2fc375/packages/types/src/index.ts#L69-L71). 
- Here, we accept `props`.. Where did props come from?

adnasa (commented):
Looks good overall. just a question

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
emmenko: Then let's maybe do that before proceeding here?

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
tdeekens: Thanks. That's a good catch!

adnasa: ping me when you resolve it ;)

tdeekens: Sure. I am doing this on the side. So might take some time :)

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
tdeekens: One of the few issues is that importing a GraphQl query results in a `DocumentNode` type. I don't know how to convert that into a string again or how Apollo does it when actually performing a fetch so far.

tdeekens: Seems like `print` from graphql does that. I don't want to push that into the flopflip adpater. Also if it was an http-adapter. How would you proceed if you want to lint this tiny query while concerting the import to a string?

emmenko: If we have an `http-adapter`, the adapter does not concern itself with such things anymore. 

The adapter args would then be something like:

```js
http: {
  user: {
    key: props.user?.id,
  },
  execute,
  pollingIntervalMs: 1000,
  parseFlags,
  cacheIdentifier: 'local',
},
```

We have then the `execute` callback function, which does the data fetching.

emmenko: In the `execute` function, you then construct and execute the request as you want. So you can use apollo client and import the query as normal.

emmenko: Let me know if you want me to work on that in flopflip.

tdeekens: Yeah. I agree. With an `http-adapter` the issue is gone. I like the abstraction too. If we ever need a GraphQl adapter it can compose or extend the http-adapter. Maybe one night I can work on it. I'll ping you for review.

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
```diff
@@ -141,17 +132,23 @@ export const SetupFlopFlipProvider = (props: Props) => {
...
-      graphql: {
-        user: getGraphQlUser(props.user, props.projectKey),
-        uri: `${getMcApiUrl()}/graphql`,
-        query: AllFeaturesQuery,
-        pollingInteralMs: 1000 * 60 * 30, // 30 minutes
-        getRequestHeaders,
-        parseFlags,
+      http: {
+        user: getHttpUser(props.user, props.projectKey),
+        execute: async () => {
```
tdeekens, line 137: @emmenko let me know what you think. This roughly would be the direction we talked about I think.

emmenko: Yep, looks much better now! 🎉

emmenko (commented):
Nice, I like it. It's also much simpler now!

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
```diff
@@ -18,21 +25,68 @@ type Props = {
...
+    subgroup: user.launchdarklyTrackingSubgroup,
+    tenant: user.launchdarklyTrackingTenant,
+  }
+
+  if (!project) return customFieldsOfUser
+
+  return {
+    project: projectKey,
+    ...customFieldsOfUser
+  };
```
tdeekens, line 71: @emmenko this is why I liked the `omitEmpty` also if it was, I admit you caught me, lying about the types. Kind of funny that you can with generics.

How would you like to make this more explicit? TypeScript doesn't like this I think. Thanks.

tdeekens:
I think I am done on the `@flopflip` side. Type checks should also pass. Next would be experimenting with it internally.

tdeekens:
So far so good

CleanShot 2021-01-15 at 00 54 26@2x

Need to fix 1-2 things in `@flopflip` I think how we can read these complex flag values now.

tdeekens:
⭐ Success. Using `useFlagVariation` now works as expected and `flopflip` manages state for two adapters like a breeze ⭐ 

CleanShot 2021-01-22 at 09 19 24@2x

emmenko:
Nice!! 🤩

Is this now ready for review?

tdeekens:
In theory yes. I wanted to ask how we release it. I see two options:

1. Add it and refactor all internal uses while updating to app-kit
2. Add it behind a `disableLongLivedFeaturesFetching` (default true) on `<ApplicationShell />` to give more freedom and not have Custom Apps load the flags needlessly

Do you have any thoughts on that already?

emmenko:
Good question. Maybe opt-in would be better first.

adnasa:
I think opt-in would be my preferred approach

tdeekens:
Would you prefer a generic approach here? I can see something like

```jsx
<ApplicationShell 
  disableFeatureConfigurationFetching={true}
/>
```

but also 

```jsx
<ApplicationShell 
  flags={{
    disableFeatureConfigurationFetching: true
  }}
/>
```

any preference?

emmenko:
I would maybe go with an application config environment value. This way we don't need to add any new prop to the ApplicationShell, as we may remove it again in the future.

tdeekens:
> I would maybe go with an application config environment value. This way we don't need to add any new prop to the ApplicationShell, as we may remove it again in the future.

Sure. Why not. Sounds good.

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
```diff
@@ -82,8 +81,16 @@ const parseFlags = (fetchedFlags: TFetchedFlags): TParsedHttpAdapterFlags =>
...
+type TAdditionalEnvironmentProperties = {
+  env: {
+    enableFeatureConfigurationFetching?: boolean;
+  }
+};
 export const SetupFlopFlipProvider = (props: Props) => {
   const apolloClient = useApolloClient();
+  const enableFeatureConfigurationFetching = useApplicationContext<TAdditionalEnvironmentProperties>(
+    (context) => context.environment.env.enableFeatureConfigurationFetching
+  );
```
tdeekens, line 93: @emmenko this is what you had in mind, right?

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
```diff
@@ -27,12 +35,62 @@ type CustomLDUser = {
...
+
+type TAdditionalEnvironmentProperties = {
+  env: {
+    enableFeatureConfigurationFetching?: boolean;
+  }
+};
 export const SetupFlopFlipProvider = (props: Props) => {
+  const apolloClient = useApolloClient();
+  const enableFeatureConfigurationFetching = useApplicationContext<TAdditionalEnvironmentProperties>(
+    (context) => context.environment.env.enableFeatureConfigurationFetching
```
emmenko, line 92:
```suggestion
    (context) => context.environment.enableFeatureConfigurationFetching
```

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
emmenko: Yes thanks

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
```diff
@@ -81,15 +81,10 @@ const parseFlags = (fetchedFlags: TFetchedFlags): TParsedHttpAdapterFlags =>
     ])
   );
 
-type TAdditionalEnvironmentProperties = {
```
tdeekens: I felt that anything that app-kit supports should be additional conceptually...

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
```diff
@@ -101,11 +96,11 @@ export const SetupFlopFlipProvider = (props: Props) => {
     [allMenuFeatureToggles.allFeatureToggles, props.flags]
   );
   React.useMemo(() => {
```
tdeekens, line 98: Simple code is good code I think.

packages/constants/src/constants.ts
```diff
@@ -169,5 +169,6 @@ export interface ApplicationWindow extends Window {
     useFullRedirectsForLinks?: boolean;
     // Properties for OIDC-like workflow for development
     __DEVELOPMENT__?: ApplicationOidcForDevelopmentConfig;
+    enableFeatureConfigurationFetching?: boolean;
```
tdeekens, line 172: ...that's why I added it here.

emmenko: Can you move it below `useFullRedirectsForLinks`?
This way we keep the props "grouped". Thanks

tdeekens: Sure. I moved it to `enableSignUp` cause those two look similar.

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
```diff
@@ -143,7 +143,7 @@ export const SetupFlopFlipProvider = (props: Props) => {
           });
 
           return parseFlags(response.data);
-        }
+        },
```
adnasa, line 155: 😄

adnasa (approved):

emmenko (commented):
Thanks, looks good to me now. Don't forget to add the changeset.

tdeekens:
Yup. I want to beta test this a bit more. So I'll not merge it for now. I need time for that.

tdeekens:
Ok. I did extensive testing internally. This works very nicely. Anything left here @emmenko otherwise I can start integrating.

.changeset/dull-bees-obey.md
```diff
@@ -0,0 +1,28 @@
...
+"@commercetools-frontend/mc-html-template": patch
+"@commercetools-frontend/mc-scripts": patch
+"@commercetools-frontend/notifications": patch
+"@commercetools-frontend/permissions": patch
+"@commercetools-frontend/react-notifications": patch
+"@commercetools-frontend/sdk": patch
+"@commercetools-frontend/sentry": patch
+---
+
+refactor: use `combine-adapter` to be able to use `http-adapter` and `launchdarkly-adapter`
```
emmenko, line 28: Nit: I think we can keep it a bit more generic, as external users are not very familiar with this and it's anyway for internal usage only.

```suggestion
Add internal opt-in support for combining feature flag adapters.
```

emmenko (approved):
Looks good! Looking forward to see this in action 🚀

PR #1989, opened 2021-01-22
chore: update uikit to 10.44, which comes with some packages exporting type declarations

emmenko:
Let's start doing some cleanup 🤓


tdeekens (approved):
Good stuff!

emmenko:
cc @adnasa for awareness

PR #1996, opened 2021-01-25
fix(mc-html-template): add no-referer policy to script-tags

tdeekens:
#### Summary

This pull request is very similar to the #1994 just focussed on `script`-tags.

I noticed that we could also apply the same directive to script tags.

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script


emmenko (approved):
Thanks!

PR #1998, opened 2021-01-26
fix(mc-html-template): to never send a referrer

tdeekens:
#### Summary

This pull request takes the idea of never sending a referrer a step further.

#### Description

Given we have a script which as a no-referrer. Given this script loads another script. That child script will be loaded in the context of the HTMLs referrer policy not the parent script's one.

Given we can also specify the `no-referrer` on the HTML as a meta-tag I propose to do that. 

I don't think we ever rely on the `referer` (the header is misspelled on purpose) we should be able to do that. Also the origin header is similar to the Referer header, but, unlike this header, it doesn't disclose the whole path.

### DIFF
--- a/packages/mc-html-template/src/generate-template.js
+++ b/packages/mc-html-template/src/generate-template.js
@@ -4,11 +4,10 @@ module.exports = function generateTemplate({
 }) {
   const cssImports = cssChunks.map(
     (chunkPath) =>
-      `<link href="__CDN_URL__${chunkPath}" rel='stylesheet' type='text/css' referrerpolicy='no-referrer'></link>`
+      `<link href="__CDN_URL__${chunkPath}" rel='stylesheet' type='text/css'></link>`
   );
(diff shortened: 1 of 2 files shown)


packages/mc-html-template/src/generate-template.js
```diff
@@ -17,8 +16,9 @@ module.exports = function generateTemplate({
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no">
-    <link rel="preconnect" href="__CDN_URL__" referrerpolicy="no-referrer">
-    <link rel="preconnect" href="__MC_API_URL__" referrerpolicy="no-referrer">
+    <meta name="referrer" content="no-referrer">
```
tdeekens, line 19: This would be the relevant change.

emmenko (approved):
Ah nice, that's simpler!

PR #1999, opened 2021-01-26
fix(app-shell): rendering of app with implicit routes when redirecting to /account route

emmenko:
Small regression related to our special `/account` route.


dogayuksel (approved):

adnasa (approved):
💯

tdeekens (approved):

PR #2008, opened 2021-01-28
fix(cypress): importing types into a .d.ts module

emmenko:


packages/cypress/add-commands/index.d.ts
```diff
@@ -1,12 +1,13 @@
-import type { CommandLoginByOidcOptions } from './dist/commercetools-frontend-cypress-add-commands.cjs';
```
emmenko, line 1: Apparently TS is not able to properly import this type, causing the entire declaration file to be unusable.

Using the inline `import()` is the only thing I could find to make it work.

adnasa (approved):

tdeekens (approved):

PR #2010, opened 2021-01-28
chore: upgrade graphql to v15

emmenko:


emmenko:
Changes included in #2041

PR #2014, opened 2021-01-29
refactor(test-utils): to use flopflip test provider

tdeekens:
#### Summary

This pull request refactors our test-utils to use the new `TestProviderFlopFlip`.

#### Description

The mentioned provider reduces the amount of complexity for the test setup:

1. No more `memory-adapter` and `reset` on each tets
2. No "async" state of flags which leak into tests (wait until flags are flushed otherwise act warnings)

### DIFF
--- a/.changeset/tidy-llamas-deliver.md
+++ b/.changeset/tidy-llamas-deliver.md
@@ -4,5 +4,7 @@
 
 refactor(test-utils): to use flopflip test provider
 
-In this release we use the `TestProviderFlopFlip` when using our `test-utils` or the `SetupFlipFlip` component. 
-This change just not affect consumers of those packages. You can remove any `adapter` you may have passed to any `react-testing-library` renderer. These are not needed anymore.
+When using `@commercetools-frontend/application-shell/test-utils`, we now render a `TestProviderFlopFlip` instead of the normal `ConfigureFlopFlip` with the `memory` adapter.<br/>
+This change simplifies how feature flags are propagated during tests and should not affect the usage of the test-utils.
+
+> In the very unlikely case that you have been passing the `adapter` option to the test-utils, you can remove that as it's not necessary anymore.
--- a/.changeset/tidy-llamas-deliver.md
+++ b/.changeset/tidy-llamas-deliver.md
@@ -1,0 +1,8 @@
+---
+"@commercetools-frontend/application-shell": patch
+---
+
+refactor(test-utils): to use flopflip test provider
+
+In this release we use the `TestProviderFlopFlip` when using our `test-utils` or the `SetupFlipFlip` component. 
+This change just not affect consumers of those packages. You can remove any `adapter` you may have passed to any `react-testing-library` renderer. These are not needed anymore.
--- a/packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
+++ b/packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
@@ -117,11 +117,6 @@ export const SetupFlopFlipProvider = (props: Props) => {
       user: {
         key: props.user?.id,
       },
-      memory: {
-        user: {
-          key: props.user?.id,
-        },
-      },
       launchdarkly: {
         sdk: {
           // Allow to overwrite the client ID, passed via the `additionalEnv` properties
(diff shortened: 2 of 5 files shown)


packages/application-shell/src/test-utils/test-utils.tsx
```diff
@@ -274,7 +256,6 @@ export type TRenderAppOptions<AdditionalEnvironmentProperties = {}> = {
   route: string;
   disableAutomaticEntryPointRoutes: boolean;
   history: ReturnType<typeof createEnhancedHistory>;
-  adapter: typeof memoryAdapter;
```
emmenko, line 277: Kind of a breaking change in theory but I assume that we would only use this internally (if any).

So no need to bump a major.

packages/application-shell/src/test-utils/test-utils.tsx
```diff
@@ -429,12 +408,7 @@ function renderApp<AdditionalEnvironmentProperties = {}>(
         apolloClient={apolloClient}
         mocks={mocks}
       >
-        <ConfigureFlopFlip
-          adapter={adapter}
-          defaultFlags={flags}
-          adapterArgs={defaultFlopflipAdapterArgs}
-          shouldDeferAdapterConfiguration={!hasFlags}
-        >
+        <TestProviderFlopFlip flags={flags}>
```
emmenko, line 411: 💯

emmenko (commented):
Thanks! Do we need to document any possible "side effects" of this change, or something that consumers might need to be aware of?

packages/application-shell/src/test-utils/test-utils.tsx
tdeekens: I don't think anybody used this. Not even we did.

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
```diff
@@ -9,7 +9,10 @@ import { useApplicationContext } from '@commercetools-frontend/application-shell
 import ldAdapter from '@flopflip/launchdarkly-adapter';
 import httpAdapter from '@flopflip/http-adapter';
 import combineAdapters from '@flopflip/combine-adapters';
-import { ConfigureFlopFlip } from '@flopflip/react-broadcast';
+import {
+  ConfigureFlopFlip,
+  TestProviderFlopFlip,
```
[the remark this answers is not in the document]
tdeekens: Good point. I fixed it to be `require'd in the `if`-block just like the adapter before.

tdeekens:
> Thanks! Do we need to document any possible "side effects" of this change, or something that consumers might need to be aware of?

I don't see any. I don't think anybody passed an `adapter` actually. Also all tests which were async to wait for UI to appear in accordance to flags will continue to work. In that I think this should be transparent to consumers.

.changeset/tidy-llamas-deliver.md
```diff
@@ -0,0 +1,8 @@
+---
+"@commercetools-frontend/application-shell": patch
+---
+
+refactor(test-utils): to use flopflip test provider
+
+In this release we use the `TestProviderFlopFlip` when using our `test-utils` or the `SetupFlipFlip` component. 
+This change just not affect consumers of those packages. You can remove any `adapter` you may have passed to any `react-testing-library` renderer. These are not needed anymore.
```
emmenko, line 8:
```suggestion
When using `@commercetools-frontend/application-shell/test-utils`, we now render a `TestProviderFlopFlip` instead of the normal `ConfigureFlopFlip` with the `memory` adapter.<br/>
This change simplifies how feature flags are propagated during tests and should not affect the usage of the test-utils.

> In the very unlikely case you have been passing the `adapter` option to the test-utils, you can remove that as it's not necessary anymore.
```

packages/application-shell/src/test-utils/test-utils.tsx
emmenko, line 411: Hmm wait, do we still need the change in `SetupFlopFlip` then? 🤔

tdeekens: Yes, otherwise our tests here fail of app-shell if this is removed. I think removing it, if possible, could also come after.

emmenko (approved):
🚀  Thanks

tdeekens:
Tested a build internally. This works like a charm.

Issue #1884, opened 2020-11-18
Re-implement <Navbar> component following new design guidelines


emmenko:
> Just so that we can keep track of it.

**Background**

The current implementation of the `<Navbar>` is pretty "old" and has not really been updated since years (layout and styles).

The design team did provide some improvements in the design and layout, also in regards to having more menu items (for Custom Applications). However, we never really prioritized the work here.

**Task**

The `<Navbar>` component should be re-implemented following the new design guidelines.
Furthermore, we should use emotion and not css-modules. This is important as it currently blocks the work in #1871 (related to loading css files via babel).

**Design guidelines**

_To be provided_

PR #2018, opened 2021-01-29
refactor(app-shell): drop Rollup, pre-compile css module and use a macro to load the styles

emmenko:
So I finally figured out a way to bundle the app-shell package using preconstruct.

Until now it was not possible because we loaded CSS modules, and thus babel does not know how to process that import.

Now we don't directly load the `.css` file anymore. Instead, we use `postcss` to compile it and load the styles using a macro. This allows the code to be bundled using Babel.


.gitignore
```diff
@@ -69,11 +69,11 @@ typings/
 .public
 
 # Build
+compiled
 compiled-data
 dist
 dist-tarballs
 public
-packages/application-shell/test-utils
```
emmenko, line 76: The `test-utils` is now a preconstruct entrypoint.

packages/application-shell/scripts/compile-styles.js
```diff
@@ -0,0 +1,83 @@
...
+
+        // This file provides the type declaration for the JSON file created above.
+        fs.writeFileSync(
+          `${path.join(compiledPaths.dir, compiledPaths.fileName)}.json.d.ts`,
+          `/* eslint-disable prettier/prettier */
+declare const styles: ${JSON.stringify(json)};
+export default styles;`,
+          { encoding: 'utf8' }
+        );
+      },
```
emmenko, line 54: This function essentially creates the JSON file that contains the mapping between the class names and the compiled selectors:

image

packages/application-shell/scripts/compile-styles.js
```diff
@@ -0,0 +1,83 @@
...
+    const css = fs.readFileSync(cssFilePath, { encoding: 'utf8' });
+    const compiled = await processor.process(css, {
+      ...options,
+      from: cssFilePath,
+    });
+    fs.writeFileSync(
+      path.join(compiledPaths.dir, compiledPaths.fileName),
+      compiled.css,
+      { encoding: 'utf8' }
+    );
```
emmenko, line 76: Here `compiled.css` is the actual compiled CSS code.

packages/application-shell/src/components/application-shell-provider/application-shell-provider.tsx
```diff
@@ -47,46 +47,49 @@ const ApplicationShellProvider = <AdditionalEnvironmentProperties extends {}>(
...
-              <Router history={ApplicationShellProvider.history}>
-                <GtmBooter trackingEventList={props.trackingEventList || {}}>
-                  <Authenticated
-                    locale={browserLocale}
-                    applicationMessages={props.applicationMessages}
-                    render={({ isAuthenticated }) => {
-                      if (isAuthenticated)
-                        return props.children({ isAuthenticated });
+    <>
+      <GlobalStyles />
```
emmenko, line 51: Injected using Emotion `Global`

packages/application-shell/src/components/application-shell/application-shell.tsx
```diff
@@ -1,4 +1,3 @@
-import type { SyntheticEvent } from 'react';
```
emmenko, line 1: Preconstruct doesn't like importing React types separately.

packages/application-shell/src/components/navbar/navbar.styles.js
```diff
@@ -0,0 +1,11 @@
+const fs = require('fs');
+const path = require('path');
+
+const compiledStyles = {
+  global: fs.readFileSync(path.join(__dirname, './compiled/navbar.css'), {
+    encoding: 'utf8',
+  }),
+  jsonMap: require('./compiled/navbar.css.json'),
+};
```
emmenko, line 9: This is the macro. We're simply returning the loaded CSS and JSON map.

Note that in theory we could have run `postcss` here as well (instead of using a custom script). However, the process is async and it would complicate things, as we want here things to be sync.

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -55,11 +55,14 @@ import { RestrictedByPermissions } from '@commercetools-frontend/permissions';
...
 import LoadingPlaceholder from '../loading-placeholder';
-import styles from './navbar.mod.css';
 import messages from './messages';
 import useLoadingMenuLayoutEffect from './use-loading-menu-layout-effect';
 import useNavbarStateManager from './use-navbar-state-manager';
 import nonNullable from './non-nullable';
+// https://babeljs.io/blog/2017/09/11/zero-config-with-babel-macros
+import compiledStyles from /* preval */ './navbar.styles';
+
+const styles = compiledStyles.jsonMap;
```
emmenko, line 65: In the code, we reference the class names using the JSON map.

Bonus now is that we also get autocompletion and type checks.

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -538,15 +541,22 @@ type NavBarLayoutProps = {
...
-      data-track-component="Navigation"
-    >
-      {props.children}
-    </nav>
+    <>
+      <Global
+        styles={css`
+          ${compiledStyles.global}
+        `}
+      />
```
emmenko, line 549: Since we pre-compiled the CSS styles, we need to inject them globally here.

emmenko:
The styles in the bundle are now like this:

image

Which is actually more or less the same as we had before with Rollup compiling CSS modules.

packages/application-shell/package.json
```diff
@@ -15,20 +15,15 @@
...
   "files": ["dist", "test-utils", "package.json", "LICENSE", "README.md"],
   "scripts": {
     "prepare": "./../../scripts/version.js replace",
-    "prebuild": "rimraf dist/** test-utils/**",
-    "build": "yarn build:bundles && yarn build:test-utils && yarn build:typings",
-    "build:bundles": "cross-env NODE_ENV=production rollup -c ../../rollup.config.js -i ./src/index.ts -d dist",
-    "build:bundles:watch": "yarn build:bundles -w",
-    "build:test-utils": "cross-env NODE_ENV=development rollup -c ../../rollup.config.js -i ./src/test-utils/index.ts",
-    "build:typings": "cross-env tsc -p tsconfig.declarations.json --emitDeclarationOnly --declarationDir dist/typings",
-    "postbuild:typings": "echo \"export * from '../dist/typings/test-utils';\" > test-utils/index.d.ts"
```
tdeekens, line 31: 💯

packages/application-shell/scripts/compile-styles.js
tdeekens: This will still be fingerprinted etc as it's loaded through webpack, right?

tdeekens (approved):
Thanks. I can follow along from a setup perspective. Macros are somewhat magical to me still :)

packages/application-shell/scripts/compile-styles.js
emmenko: No really,this script is what we execute independently.

adnasa (approved):

adnasa:
I don't even know what a macro is 😄

tdeekens:
> I don't even know what a macro is (except via Sass or Less) 😄

This article was quite nice: https://kentcdodds.com/blog/write-your-own-code-transform

adnasa:
It took a couple of more rounds of reading, and something I would have liked to see at the PR. but here is what I understood

1. We pre-compile the css
2. We expose the css via a macro as `compiledStyles`
3. the macro exposes two values, `jsonMap` and `global`
  - the `jsonMap` is a list of references to the compiled css classes (`.foo`)
  - the `global` are value that we inline into a global component, hence the the css classes and their rules will be applied

---

questions

- question regarding the `jsonMap`, since we reference the only the  css classes, where are their rules injected?
- when to use what? 
- Fo the moment, the `application-provider` hosts the global styles, namely `reset` and `grid`.
  - If we need to introduce more global styles further down the road, we add them into `<GlobalStyles />`?

side note

- is the `application-provider` the right place to host global styles?

emmenko:
> question regarding the jsonMap, since we reference the only the css classes, where are their rules injected? (can't find the code that does the injection pointing to a complied css file)

See https://github.com/commercetools/merchant-center-application-kit/pull/2018#issuecomment-770388070
The compiled CSS is injected using `Global`, which is what the `global` string contains. The code is [here](https://github.com/commercetools/merchant-center-application-kit/pull/2018/files#diff-a7455b26572be541aa005dcd6b6e0843bfff33052c8c0092d7daf11c4e885d54R545-R549).

> when to use what?

What what?

PS: note that this is a "temporary" solution, as we want to use Emotion eventually and no CSS modules. See #1884

> If we need to introduce more global styles further down the road, we add them into <GlobalStyles />?

Yes

> is the application-provider the right place to host global styles?

Yes, because it's the only component that is always rendered (see our internal auth app, which does not use AppShell).

adnasa:
> The compiled CSS is injected using Global, which is what the global string contains. The code is here.

aah... nice that is actually elegant.

---

PS. can't say it's genius, however 🤷🏽‍♂️  :trollface: #slap

PR #2030, opened 2021-02-02
fix(authenticated): cache lookup

adnasa:
#### Summary

- moves `cacheScope` logic to the bottom
- remove redundant key in `env`, already defined in `env.local`


emmenko (approved):
Thanks

PR #2033, opened 2021-02-03
refactor: to rename to enableLongLivedFeatureFlags

tdeekens:
#### Summary

This pull request suggests to rename `enableFeatureConfigurationFetching` to `enableLongLivedFeatureFlags`.

#### Description

While integrating this feature it became apparent that we will now naturally distinguish between short-lived flags (through LaunchDarkly) and long-running flags (through our GraphQL API).

This also means that we can keep this environment variable. Custom Applications will not have to fetch our long-running flags while they could (if e.g. an internal app) as an opt-in.

As a result, I think it makes sense to enable this with a more descriptive environment variable which indicates more clearly what it will cause.

<hr />

As a follow-up we can also discuss with we should have a `enableShortLivedFeatureFlags`. This would mean that Custom Applications would not receive feature flags from LaunchDarkly. Which could be nice as long as we do not use feature flags on the menu or we accept inconsistencies from Custom Applications with internal applications being under development. Note, that all internal applications can be hidden on a network layer (proxy) these days.

### DIFF
--- a/.changeset/cuddly-kangaroos-hope.md
+++ b/.changeset/cuddly-kangaroos-hope.md
@@ -1,0 +1,6 @@
+---
+"@commercetools-frontend/application-shell": patch
+"@commercetools-frontend/constants": patch
+---
+
+Rename to `enableFeatureConfigurationFetching` to `enableLongLivedFeatureFlags`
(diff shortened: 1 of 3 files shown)


.changeset/yellow-penguins-grab.md
```diff
@@ -0,0 +1,6 @@
+---
+"@commercetools-frontend/application-shell": patch
+"@commercetools-frontend/constants": patch
+---
+
+refactor: to rename to enableLogLivedFeatureFlags
```
adnasa, line 6:
```suggestion
rename enableFeatureConfigurationFetching to enableLogLivedFeatureFlags
```

adnasa (approved):

PR #2056, opened 2021-02-15
fix(components): remove unnecessary apollo client dependency

emmenko:
I guess this was a leftover from removing the "experimental" entry point from the components package.

### DIFF
--- a/packages/application-components/package.json
+++ b/packages/application-components/package.json
@@ -31,7 +31,6 @@
     "prepare": "./../../scripts/version.js replace"
   },
   "dependencies": {
-    "@apollo/client": "3.3.8",
     "@babel/runtime": "7.12.13",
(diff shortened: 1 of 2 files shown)


packages/application-components/package.json
```diff
@@ -31,7 +31,6 @@
     "prepare": "./../../scripts/version.js replace"
   },
   "dependencies": {
-    "@apollo/client": "3.3.8",
```
emmenko, line 34: FYI: in the other packages where we do use it, it's declared as a peer dependency

adnasa (approved):
good catch

adnasa:
> I guess this was a leftover from removing the "experimental" entry point from the components package.


good catch. yes it was

tdeekens (approved):

PR #2060, opened 2021-02-19
fix(sdk/middleware): support `apply` endpoint for `orderEdits`

### LINKED ISSUE #2058: applyOrderEditTo applied as a query string parameter
Hi I encountered an issue when using post requests with the frontend sdk.
I am attempting to post and 'apply' OrderEdit but when I apply it I get a malformed url resulting in a 400

```
  return sdkActions.post({
    service: SERVICEPATH.OrdersEdits,
    mcApiProxyTarget: MC_API_PROXY_TARGETS.COMMERCETOOLS_PLATFORM,
    options: { applyOrderEditTo: orderEdit?.id },
    payload: JSON.stringify(applyOrderEdit)
 })
```
Like
orders/edits?/acc7e3ca-afb4-490c-b20c-65910a06ee35/apply

It seems to originate here: https://github.com/commercetools/nodejs/blame/master/packages/api-request-builder/src/create-service.js#L141 where order edit it always false and so its appended as a query string. 

image

I can work around it doing this:
```
 let applyOrderEdit: OrderEditApply = { editVersion: orderEdit?.version, resourceVersion: orderVersion  }
  return sdkActions.post({
    service: SERVICEPATH.OrdersEdits,
    mcApiProxyTarget: MC_API_PROXY_TARGETS.COMMERCETOOLS_PLATFORM,
    options: { id: orderEdit?.id + "/apply" },
    payload: JSON.stringify(applyOrderEdit)
  });
};
```

adnasa:
#### Summary

-  support `apply` endpoint for `orderEdits`
- closes #2058

#### Description

In `sdk/middleware`, we `build()` request [for the respective service](https://github.com/commercetools/nodejs/blob/master/packages/api-request-builder/src/default-services.js#L1:L1) in `sdk`.

It turned out that the `service.build()` is coupled to an option for `orderEdits` that determines whether an `order/edits/${id}/apply` endpoint should be build.

Here is how we reason when building the `order/edits/${id}/apply` endpoint in `sdk/middleware`:

- GIVEN that `service=orderEdits` and `applyOrderEditTo` has a value: [we build the apply endpoint](https://docs.commercetools.com/api/projects/order-edits#apply-an-orderedit)
- GIVEN that `service=orderEdits` and `applyOrderEditTo` has no value: [we build the update endpoint](https://docs.commercetools.com/api/projects/order-edits#update-orderedit-by-id)

### DIFF
--- a/packages/sdk/src/middleware/middleware.ts
+++ b/packages/sdk/src/middleware/middleware.ts
@@ -37,7 +37,17 @@ const actionToUri = (action: TSdkAction, projectKey?: string): string => {
   const service = requestBuilder[action.payload.service];
   if (action.payload.options) service.parse(action.payload.options);
 
-  return service.build();
+  return service.build({
+    // given `service=orderEdits` and given `applyOrderEditTo`, we build an apply endpoint
+    // given `service=orderEdits` and no `applyOrderEditTo`, we build an update endpoint
+    // https://docs.commercetools.com/api/projects/order-edits
+    applyOrderEdit:
+      action.payload.service === 'orderEdits' &&
+      typeof action.payload.options?.applyOrderEditTo === 'string',
+
+    // at this stage, the `projectKey` should be available already.
+    withProjectKey: true,
+  });
 };
 
 // Force TS cast of generic action to TNotificationAction
--- a/packages/sdk/src/middleware/middleware.ts
+++ b/packages/sdk/src/middleware/middleware.ts
@@ -20,6 +20,9 @@ const isSdkActionForUri = (
 ): actionPayload is TSdkActionPayloadForUri =>
   (actionPayload as TSdkActionPayloadForUri).uri !== undefined;
 
+// https://github.com/commercetools/nodejs/blob/master/packages/api-request-builder/src/default-services.js#L200:L200
+const ORDER_EDIT_SERVICE = 'orderEdits';
+
 const actionToUri = (action: TSdkAction, projectKey?: string): string => {
   if (isSdkActionForUri(action.payload)) return action.payload.uri;
 
@@ -42,9 +45,8 @@ const actionToUri = (action: TSdkAction, projectKey?: string): string => {
     // given `service=orderEdits` and no `applyOrderEditTo`, we build an update endpoint
     // https://docs.commercetools.com/api/projects/order-edits
     applyOrderEdit:
-      action.payload.service === 'orderEdits' &&
+      action.payload.service === ORDER_EDIT_SERVICE &&
       typeof action.payload.options?.applyOrderEditTo === 'string',
-
     // at this stage, the `projectKey` should be available already.
     withProjectKey: true,
   });
--- a/.changeset/breezy-taxis-tie.md
+++ b/.changeset/breezy-taxis-tie.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/sdk': minor
+---
+
+support order/edits/apply in sdk/middleware
--- a/@types/commercetools__api-request-builder/index.d.ts
+++ b/@types/commercetools__api-request-builder/index.d.ts
@@ -1,6 +1,10 @@
 declare module '@commercetools/api-request-builder' {
+  type TUriOptions = {
+    withProjectKey?: boolean;
+    applyOrderEdit?: boolean;
+  };
   export type ServiceBuilder = {
-    build(): string;
+    build(uriOptions?: TUriOptions): string;
     parse(options: { [key: string]: unknown }): string;
   };
   export type ApiRequestBuilder = {
--- a/packages/sdk/src/middleware/middleware.spec.ts
+++ b/packages/sdk/src/middleware/middleware.spec.ts
@@ -283,6 +283,87 @@ describe('when the action is of type SDK', () => {
     });
   });
 
+  describe('when `service=orderEdits` and `applyOrderEditTo`', () => {
+    const dispatch = jest.fn();
+    const orderId = 'order-to-edit-id-1';
+    const action = sdkActions.post({
+      service: 'orderEdits',
+      options: { applyOrderEditTo: orderId },
+      payload: {},
+    });
+    const next = jest.fn();
+    const response = { body: { foo: 'bar' }, headers: {}, statusCode: 200 };
+    beforeEach(() => {
+      execute = jest.fn(() => Promise.resolve(response));
+      mocked(createClient).mockReturnValue({
+        execute,
+      });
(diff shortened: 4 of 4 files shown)


packages/sdk/src/middleware/middleware.ts
```diff
@@ -37,7 +37,17 @@ const actionToUri = (action: TSdkAction, projectKey?: string): string => {
   const service = requestBuilder[action.payload.service];
   if (action.payload.options) service.parse(action.payload.options);
 
-  return service.build();
+  return service.build({
+    // given `service=orderEdits` and given `applyOrderEditTo`, we build an apply endpoint
+    // given `service=orderEdits` and no `applyOrderEditTo`, we build an update endpoint
+    // https://docs.commercetools.com/api/projects/order-edits
+    applyOrderEdit:
+      action.payload.service === 'orderEdits' &&
```
adnasa, line 45: This is a bit hacky, but the request-builder doesn't expose any type that I can make use of.

packages/sdk/src/middleware/middleware.ts
```diff
@@ -37,7 +38,16 @@ const actionToUri = (action: TSdkAction, projectKey?: string): string => {
...
   if (action.payload.options) service.parse(action.payload.options);
 
-  return service.build();
+  return service.build({
+    // given `service=orderEdits` and given `applyOrderEditTo`, we build an apply endpoint
+    // given `service=orderEdits` and no `applyOrderEditTo`, we build an update endpoint
+    // https://docs.commercetools.com/api/projects/order-edits
+    applyOrderEdit:
+      action.payload.service === SUPPORTED_SERVICES.orderEdits &&
+      typeof action.payload.options?.applyOrderEditTo === 'string',
```
adnasa, line 47: we want to ensure that it is a string, not `null` nor a number

packages/sdk/src/middleware/middleware.ts
```diff
@@ -37,7 +38,16 @@ const actionToUri = (action: TSdkAction, projectKey?: string): string => {
...
-  return service.build();
+  return service.build({
+    // given `service=orderEdits` and given `applyOrderEditTo`, we build an apply endpoint
+    // given `service=orderEdits` and no `applyOrderEditTo`, we build an update endpoint
+    // https://docs.commercetools.com/api/projects/order-edits
+    applyOrderEdit:
+      action.payload.service === SUPPORTED_SERVICES.ORDER_EDITS &&
+      typeof action.payload.options?.applyOrderEditTo === 'string',
+    // at this stage, the `projectKey` should be available already.
+    withProjectKey: true,
```
qmateub, line 49: Why it was not there before? 🤔

packages/sdk/src/middleware/middleware.ts
qmateub: There is one thing that I'm not sure I follow 100%. Is this parameter then going to be send for all the services and not just for `order-edits` (to `false` obviously) or is this part of the code only related to the interaction with the `order-edits` endpoint?

qmateub (commented):
Got what the problem is 👍 . I think I'm maybe missing some context 😅

packages/sdk/src/middleware/middleware.ts
adnasa: because it is defaulted https://github.com/commercetools/nodejs/blob/master/packages/api-request-builder/src/create-service.js#L128

I added it here to be more explicit.

packages/sdk/src/middleware/middleware.ts
adnasa: This will always be sent to `service.build()`
https://github.com/commercetools/nodejs/blob/master/packages/api-request-builder/src/create-service.js#L128

Which by default is `false`.
We continue sending false as long as it doesn't fulfil the requirements discussed in the PR description

mohib0306 (approved):
code looks good to me. I was wondering if we need to specify all the supported services now. I see that only one is used and the rest are just defined but not used as of now

packages/sdk/src/middleware/middleware.ts
```diff
@@ -37,7 +38,16 @@ const actionToUri = (action: TSdkAction, projectKey?: string): string => {
...
   if (action.payload.options) service.parse(action.payload.options);
 
-  return service.build();
+  return service.build({
+    // given `service=orderEdits` and given `applyOrderEditTo`, we build an apply endpoint
+    // given `service=orderEdits` and no `applyOrderEditTo`, we build an update endpoint
+    // https://docs.commercetools.com/api/projects/order-edits
+    applyOrderEdit:
+      action.payload.service === SUPPORTED_SERVICES.ORDER_EDITS &&
+      typeof action.payload.options?.applyOrderEditTo === 'string',
```
tdeekens, line 47:
```suggestion
    applyOrderEdit: getShouldApplyOrderEdit(action)
```

Somewhere above:

```js
function getShouldApplyOrderEdit(action) {
   return (
      action.payload.service === SUPPORTED_SERVICES.ORDER_EDITS &&
      typeof action.payload.options?.applyOrderEditTo === 'string'
   );
}
```

Generally I wonder what would happen if we demand to pass it in as an option on the payload? So `action.payload.options.applyOrderEditTo`. Would that work or is it against the API?

adnasa: no, we can not really demand it, unless we use unions.
But using unions wouldn't make that the much sense to implement since the consumer can't necessarily be TS

packages/sdk/src/constants.ts
```diff
@@ -0,0 +1,44 @@
+const SUPPORTED_SERVICES = {
```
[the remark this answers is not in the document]
adnasa: yeah we could.
I want to be verbose that `service` implies an endpoint.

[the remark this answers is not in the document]
tdeekens: Yup, I can see that just wondering if there is a true benefit in this case. Having this "work-around" isolated might also be nice.

packages/sdk/src/middleware/middleware.ts
tdeekens: Got yah. Thanks for clarifying. To combine with my point above. Code looks sharp, tests suberb. I'd subjectively not have the constants and slam the check into a small function.

qmateub (approved):
🚀

adnasa:
@tdeekens I dropped the services `constants` file and extracted the string into a constant with reference to the original `default-services` in `nodejs`

tdeekens (approved):
Thanks!

adnasa:
these snapshots emits the following diff.

image

I'm going to take a gamble here and approve them

PR #2072, opened 2021-03-01
fix(webpack): to pass parallelism to thread-loader

tdeekens:
#### Summary

Internally CircleCI fails at times. The assumption is that it doesn't report the CPUs right. We already pass this value to Terser for the same reason. We now would also pass it to the thread lower to limit the CPUs used.

### DIFF
--- a/packages/mc-scripts/src/config/create-webpack-config-for-production.js
+++ b/packages/mc-scripts/src/config/create-webpack-config-for-production.js
@@ -416,6 +416,9 @@ module.exports = function createWebpackConfigForProduction(options = {}) {
             {
               loader: require.resolve('babel-loader'),
               options: {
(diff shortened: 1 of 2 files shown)


packages/mc-scripts/src/config/create-webpack-config-for-production.js
```diff
@@ -416,6 +416,9 @@ module.exports = function createWebpackConfigForProduction(options = {}) {
             {
               loader: require.resolve('babel-loader'),
               options: {
+                ...(Number.isInteger(mergedOptions.toggleFlags.parallelism)
```
tdeekens, line 419: We check for a number cause for Terser we accept a Boolean which thread-loader doesn't accept.

adnasa (approved):

PR #2073, opened 2021-03-01
chore: fix dep update

tdeekens:
#### Summary

Something is wrong on `main` with the last update. I went through each dep one by one.


tdeekens:
Lol. So I by one by one updating the deps while running the failing test I ended up int eh same spot but without a failing test.

PR #2074, opened 2021-03-01
fix(webpack): options passing to thread-loader

tdeekens:
#### Summary

A bad on my end passing the options to the wrong loader.


tdeekens:
I will go ahead and merge this when CI passes as I don't want to bother anybody with the stupidity of mine.

PR #2076, opened 2021-03-02
refactor: to remove lerna and only use many-pkg

tdeekens:
#### Summary

This pull request removes `lerna` and replaces it with `many-pkg` or `yarn` itself.

#### Description

Lerna is a tool to administrate mono-repos but also more. It is build around publishing packages mainly. We only used lerna to run commands in different workspace packages. For that we can use `many-pkg` and `yarn` itself and condense the tool usage we have.

Some things to consider:

1. `lerna run` silently fails if a workspace package doesn't have the run script defined
   - This is different to `many-pkg` and `yarn` both fail so we define a few `exit 0` scripts
2. `manypkg run` only runs a run script in one package of any workspace
3. `yarn workspaces run` runs a run script in all packages of all workspaces
4. `manypkg exec` runs a shell command in all packages across all workspaces

In this we can also drop usages of `yarn --cwd` cause that's what `many-pkg run` does nicely too and even a bit better cause it's not bound to a folder (as cwd is) but a package name independent of its workspace.


package.json
```diff
@@ -36,34 +35,33 @@
...
-    "build:watch": "preconstruct watch & NODE_ENV=development lerna run --no-private --parallel build:bundles:watch",
-    "build:website": "yarn build && yarn --cwd website build && yarn --cwd website-components-playground build",
+    "playground:build": "yarn manypkg run playground build",
+    "playground:start": "yarn manypkg run playground start",
+    "playground:start:prod:local": "yarn manypkg run playground start:prod:local",
+    "visual-testing-app:start": "yarn manypkg run visual-testing-app start",
+    "visual-testing-app:build": "yarn manypkg run visual-testing-app build",
+    "visual-testing-app:serve": "yarn manypkg run visual-testing-app serve",
+    "template-starter:build": "yarn manypkg run application-template-starter build",
+    "template-starter:start": "yarn manypkg run application-template-starter start",
```
tdeekens, line 45: This works due to many-pkg's nice substring matching of package names.

package.json
```diff
@@ -36,34 +35,33 @@
...
     "versions:uikit": "manypkg upgrade @commercetools-uikit",
     "versions:docskit": "manypkg upgrade @commercetools-docs"
   },
   "workspaces": {
     "packages": [
-      "application-templates/*",
-      "packages-backend/*",
       "packages/*",
+      "packages-backend/*",
+      "application-templates/*",
```
tdeekens, line 64: This ordering actually determines in what order yarn runs things. Which makes total sense. We want to build our actual packages first before using them in e.g. the template.

package.json
```diff
@@ -36,19 +35,18 @@
...
+    "playground:start": "yarn manypkg run playground start",
+    "playground:start:prod:local": "yarn manypkg run playground start:prod:local",
+    "visual-testing-app:start": "yarn manypkg run visual-testing-app start",
+    "visual-testing-app:build": "yarn manypkg run visual-testing-app build",
+    "visual-testing-app:serve": "yarn manypkg run visual-testing-app serve",
+    "template-starter:build": "yarn manypkg run application-template-starter build",
+    "template-starter:start": "yarn manypkg run application-template-starter start",
+    "template-starter:start:prod:local": "yarn manypkg run application-template-starter start:prod:local",
+    "build": "preconstruct build && NODE_ENV=production yarn manypkg run application-template-starter build && NODE_ENV=production yarn manypkg run mc-html-template build && NODE_ENV=production yarn manypkg run mc-scripts build && NODE_ENV=production yarn manypkg run playground build && NODE_ENV=production yarn manypkg run visual-testing-app build && NODE_ENV=production yarn manypkg run website build && NODE_ENV=production yarn manypkg run website-components-playground build",
```
tdeekens, line 47: This is the explicit version of this confusing bit here

https://github.com/commercetools/merchant-center-application-kit/blob/main/package.json#L48-L49

We run prebuild in all packages. To run build with manypkg which runs prebuild again in some packages. Then we run build in all packages using lerna. Lerna only runs them in the packages which have it explicitly defined. Which runs prebuild again.

application-templates/starter/package.json
```diff
@@ -4,6 +4,7 @@
   "description": "A starter example for the bare minimum setup to develop a Merchant Center Application",
   "private": true,
   "scripts": {
+    "prebuild": "exit 0",
```
emmenko, line 7: A bit weird to ship a starter template with this. Is there another way to avoid doing this?

package.json
emmenko: The applications shouldn't be built here. Only the "nodejs" packages.

The `build` command is only meant to build library packages (see the "no private" option from the usage with lerna).

emmenko (changes requested):
Thanks for looking into this and further simplify our setup.

application-templates/starter/package.json
tdeekens: Yes, you were looking at an old commit. I removed it. It's was needed cause we had a `prebuild` in the global `package.json` which called into each sub-package to `prebuild`. Which I think should happen anyway if manypkg invokes `build` in a package or we do via `yarn`.

tdeekens:
> The applications shouldn't be built here. Only the "nodejs" packages.

I will try to deduct what the applications are. The prior setup was quite indirect in that to me at least.

emmenko:
Only public packages need to be built when running `yarn build`.
Most of them are built with `preconstruct` but there are some that are simply compiled with Babel (mc-scripts etc).
What lerna did was essentially build those remaining packages.

The private packages (which are basically applications) do not need to be built in this step.

tdeekens:
> Only public packages need to be built when running `yarn build`.
> Most of them are built with `preconstruct` but there are some that are simply compiled with Babel (mc-scripts etc).
> What lerna did was essentially build those remaining packages.
> 
> The private packages (which are basically applications) do not need to be built in this step.

Have a look at the last commit if I got it right. Otherwise let me know which ones I need to remove or add.

emmenko:
Hmm all public packages that have a `build` script should be "migrated"

tdeekens:
> Hmm all public packages that have a build script should be "migrated"

Migrated? What do you have in mind? Please check 94ced36.

emmenko:
I would try to avoid explicitly listing those packages if possible.
It might be that we would need a custom nodejs script to run the build command only on those packages.

Essentially we want something like: "run the `build` command in the public packages that have it defined".

emmenko:
Maybe we can also contribute to `manypkg` to add support for this more refined functionality.

tdeekens:
> I would try to avoid explicitly listing those packages if possible.
> It might be that we would need a custom nodejs script to run the build command only on those packages.
> 
> Essentially we want something like: "run the `build` command in the public packages that have it defined".

Yes, I checked the code to know if it was possible but it isn't. I wouldn't do it in this PR. What we could need is:

1. Allow running commands `manypkg run` in a workspace or a glob
2. Allow running commands based on a value in the `package.json` the fact that there is a `build` run script

package.json
emmenko: Ah didn't know about that. Thanks!

package.json
```diff
@@ -36,19 +35,19 @@
...
+    "playground:start": "manypkg run playground start",
+    "playground:start:prod:local": "manypkg run playground start:prod:local",
+    "visual-testing-app:start": "manypkg run visual-testing-app start",
+    "visual-testing-app:build": "manypkg run visual-testing-app build",
+    "visual-testing-app:serve": "manypkg run visual-testing-app serve",
+    "template-starter:build": "manypkg run application-template-starter build",
+    "template-starter:start": "manypkg run application-template-starter start",
+    "template-starter:start:prod:local": "manypkg run application-template-starter start:prod:local",
+    "prebuild": "yarn compile-intl",
+    "build": "preconstruct build && NODE_ENV=production NODE_ENV=production manypkg run mc-html-template build && NODE_ENV=production manypkg run mc-scripts build",
```
emmenko, line 48: Ok so we only have those 2 packages that are compiled through babel.

Let's keep them hardcoded here for now and follow up on this. We can look into contributing to manypkg to have a more flexible `run` command, or have a custom script that automatically detects which packages should be built using babel.

emmenko (approved):
This is good from my side now. Thanks again! 🙌

package.json
tdeekens: I opened and issue with them to propose it. I somehow find this setup loads easier than our prior mix of tools which did stuff.

PR #2079, opened 2021-03-04
Add navigating to a project's stores to quick access commandlist

mohib0306:
#### Summary

- `QuickAccess` => add navigating to stores of a project to the quick access's command (suggestions) list


tdeekens (approved):

packages/application-shell/src/components/quick-access/messages.ts
```diff
@@ -158,6 +158,10 @@ export default defineMessages({
     id: 'QuickAccess.openProjectSettingsChannelsTab',
     defaultMessage: 'Open Project Settings • Channels',
   },
+  openProjectSettingsStoresTab: {
+    id: 'QuickAccess.openProjectSettingsStoresTab',
+    defaultMessage: 'Open Project Settings • Stores',
+  },
```
adnasa, line 164: Will this be translated?

adnasa (approved):

packages/application-shell/src/components/quick-access/messages.ts
mohib0306: I was misled by the different naming of the command compare to MC-FE and hence missed. It pushed them here https://github.com/commercetools/merchant-center-application-kit/pull/2079/commits/172a016d5d261398caf4bdd681f0bc01ffba486d

PR #2080, opened 2021-03-04
fix(packages.json): fix the yarn:clean command

mohib0306:
#### Summary

While developing locally, the `yarn clean` command was failing due to unable to find the directories listed in the command. Reading the `manypkg` [readme](https://github.com/Thinkmill/manypkg#manypkg-exec-cli-command), it seems that we do not need to wrap the directories in single-quotes.

May be another dev can double check and confirm this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "scripts": {
     "postinstall": "manypkg check && preconstruct dev && yarn compile-css-modules",
     "auth": "npm_config_registry=https://registry.npmjs.org npm whoami",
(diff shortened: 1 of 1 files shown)


mohib0306:
I think this do not need a change-set, right?

tdeekens (approved):

tdeekens:
> I think this do not need a change-set, right?

No, it doesn't affect published packages.

Thanks for fixing.

adnasa (approved):

tdeekens:
The VRTs didn't finish but aren't affected by this PR. I'll merge it cause I need to see if master has the same issue.

PR #2084, opened 2021-03-04
fix(ci): percy to not hang

tdeekens:


tdeekens:
Also merging this to unblock CI.

PR #2041, opened 2021-02-04
[Breaking changes] Bump several major dependencies

emmenko:
Follow up of #1874 

---

This PR now contains several major dependency updates, resulting in releasing a new major of appkit due to the nature of peer dependencies and the major library upgrades.

More specifically:
* React v17 is required now
* Several other peer deps cleanups, mainly to drop old versions.
* Using uikit v12
* Using webpack v5
* Using graphql v15
* Using docskit v16 (gatsby v3)

The appkit packages don't contain any breaking changes.

PR #2093, opened 2021-03-11
chore: upgrade docs-kit to 15

emmenko:


emmenko:
Waiting for upstream fix: https://github.com/commercetools/commercetools-docs-kit/pull/903

emmenko:
So this is a bit f*ed up. Gatsby v3 ships with Webpack v5 and Graphql v15. This unblocks the Graphql v15 upgrade #2010 but it also requires the Webpack v5 upgrade #2041 (which is currently blocked by react-dev-utils).

I don't see a way of upgrading those majors separately, unless we start doing crazy things with dep hoisting and resolution (which I'd like to avoid).

emmenko:
Closing in favor of #2095

PR #2095, opened 2021-03-12
chore: upgrade to docs-kit v15

emmenko:
~~So this is a bit f*ed up. Gatsby v3 ships with Webpack v5 and Graphql v15. This unblocks the Graphql v15 upgrade #2010 but it also requires the Webpack v5 upgrade #2041 (which is currently blocked by react-dev-utils).~~

~~I don't see a way of upgrading those majors separately, unless we start doing crazy things with dep hoisting and resolution (which I'd like to avoid).~~

The websites now have separate lockfiles. This allows us to upgrade docs-kit deps without conflicts.
Once we are able to upgrade to Webpack v5, we can include the websites into the workspaces again.


emmenko:
After many try and error, until Webpack v5 is fully supported we are going to keep the website packages split from yarn workspaces, so each needs to be installed separately.

Hopefully CRA gets the Webpack v5 support released soon.

tdeekens:
Let's do it all together in one pr. I don't see a reason not to. Do you? Thanks for pushing this!

emmenko:
Alright, I removed the graphql v15 change now, since the websites have different lockfiles. We can upgrade to graphql v15 separately.

This PR now only enables upgrading to docs-kit without conflicts. Once we are able to upgrade to Webpack v5, we can include the websites into the workspaces again.

tdeekens (approved):
Maybe open an issue to remind us of adding it tp the workspace later.

adnasa:
my bad for missing out on this :/

Issue #1930, opened 2020-12-16
[Tech Debt]: Revert the changes regarding manually updated countries list


mohib0306:
#### Summary 
- `l10n` => Revert the changes from #1929. Revert the edited `json` files and re-run the script to update the files. Whenever the new `CLDR` data is release from http://cldr.unicode.org/index/downloads/cldr-39, the changes from #1929 should be reverted.

adnasa:
@mohib0306 
Did we set a date on when to do this?

mohib0306:
The CLDRv39 is not released yet, which might have the changes for the `Northern Ireland` country code http://cldr.unicode.org/index/downloads/cldr-39. The CLDRv38.1 is still missing the changes for the `Northern Ireland`. See discussion here https://github.com/commercetools/merchant-center-application-kit/pull/2054.

We can't revert these changes unless we have `Northern Ireland` country code in the CLDR data

PR #2105, opened 2021-03-16
feat(l10n): update generate-data script

mohib0306:
#### Summary

The PR updates the  `generate-l10n-data` script of `l10n` package which is responsible for generating CLDR data. i.e. countries, currencies, timezones etc. 
Back in December 2020, we have to update manually the auto-generated file of countries data and add `Northern Ireland` to them. It was needed due to Brexit. Doing this was kinda of hot fix but it also blocks updating the `cldr` package.

The latest versions of CLDR data is still missing territory data for `Northern Ireland`. It seems that it might take a while for them to add this. You can check the discussion on https://github.com/commercetools/merchant-center-application-kit/pull/2054.

 In order to unblock the update of the `cldr` package and also avoid manual update of the auto-generated file by `generate-l10n-data` script, we would simply update the script and inject the `Northern Ireland` country code to the available countries and from the there the script will take care of it and add it to the auto-generated file by it self.

This PR also closes https://github.com/commercetools/merchant-center-application-kit/issues/1930

### DIFF
--- a/packages/l10n/scripts/generate-l10n-data.js
+++ b/packages/l10n/scripts/generate-l10n-data.js
@@ -24,6 +24,27 @@ const supportedLocales = ['en', 'de', 'es', 'fr-FR', 'zh-CN', 'ja'];
 // https://www.drupal.org/project/drupal/issues/2036219
 const excludedCountries = ['QO', 'UN', 'ZZ'];
 
+// getNorthernIrelandCountryCode returns the Northern Ireland country code (key: name) depending on the passed locale
+// "XI" is the countryCode used to identify the Northern Ireland
+const getNorthernIrelandCountryCode = (locale) => {
+  switch (locale) {
+    case 'en':
+      return { XI: 'Northern Ireland' };
+    case 'de':
+      return { XI: 'Nordirland' };
+    case 'es':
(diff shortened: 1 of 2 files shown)


packages/l10n/scripts/generate-l10n-data.js
```diff
@@ -24,6 +24,27 @@ const supportedLocales = ['en', 'de', 'es', 'fr-FR', 'zh-CN', 'ja'];
...
+      return { XI: 'Northern Ireland' };
+    case 'de':
+      return { XI: 'Nordirland' };
+    case 'es':
+      return { XI: 'Irlanda del Norte' };
+    case 'fr-Fr':
+      return { XI: 'Irlande du Nord' };
+    case 'zh-CN':
+      return { XI: '北爱尔兰' };
+    case 'ja':
```
tdeekens, line 41: Only challenge is not forgetting to add a locale here given we introduce another one but it's not avoidable until it's fixed upstream.

tdeekens (approved):
As discussed 😄. I like it simple and unblocks potentially outdated other things which is more important I think.

adnasa (approved):

Issue #2115, opened 2021-03-24
Audience problem when calling External API from custom applications


michaelashe:
**Describe the issue**

We use this approach to call our external API from custom applications https://docs.commercetools.com/custom-applications/main-concepts/proxy-to-external-api

It works great for post and get requests, except for the case when we add ?query to the GET endpoint.

on the backend of external api we're checking session via createSessionAuthVerifier from the @commercetools-backend/express package.

when adding ?query it throws error: UnauthorizedError: jwt audience invalid.

The audience is set correctly and works great for all kind of requests / routes except when the route has any query param in the url (https://docs.commercetools.com/custom-applications/main-concepts/proxy-to-external-api#validating-the-json-web-token)

Are we doing something wrong? or it this is a bug? if it's a bug, could you guide when approximately it could be resolved?

emmenko:
Ah good point. I think it's a bug. We'll try to get a fix out tomorrow.

cc @adnasa let's pair on this.

PR #2118, opened 2021-03-26
fix(express/auth): omit search params when reconstrucing audience URL

adnasa:
Ref #2115 

Fix issue with invalid `audience` in case the request contains query parameters.

We now omit the search params when validating the audience (PR in MC API also needs to be merged first).

### DIFF
--- a/.changeset/few-snails-hunt.md
+++ b/.changeset/few-snails-hunt.md
@@ -2,4 +2,4 @@
 '@commercetools-backend/express': patch
 ---
 
-Fix issue with invalid `audience` in case the request contains query parameters.
+Fix issue with invalid JWT `aud` in case the request contains query string parameters. Now the constructed audience url omits the query string parameters.
--- a/.changeset/few-snails-hunt.md
+++ b/.changeset/few-snails-hunt.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-backend/express': minor
+---
+
+update `getConfiguredAudience` with explicit return and not use `toString()`
--- a/.changeset/few-snails-hunt.md
+++ b/.changeset/few-snails-hunt.md
@@ -1,5 +1,5 @@
 ---
-'@commercetools-backend/express': minor
+'@commercetools-backend/express': patch
 ---
 
-update `getConfiguredAudience` with explicit return and not use `toString()`
+Fix issue with invalid `audience` in case the request contains query parameters.
--- a/packages-backend/express/src/auth.ts
+++ b/packages-backend/express/src/auth.ts
@@ -110,9 +110,9 @@ const getConfiguredAudience = (
 ) => {
   const url = new URL(options.audience);
   if (requestPath !== '/') {
-    url.pathname = requestPath;
+    return `${url.origin}${requestPath}${url.search}`;
   }
-  return url.toString();
+  return `${url.origin}${url.search}`;
 };
 
 function createSessionAuthVerifier<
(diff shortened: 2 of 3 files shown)


adnasa:
@emmenko I'm setting this to draft so I update the test once the approach is approved.
Sorry for being slow on this bud 🙌🏽

.changeset/few-snails-hunt.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-backend/express': minor
+---
+
+update `getConfiguredAudience` with explicit return and not use `toString()`
```
emmenko, line 5:
```suggestion
---
'@commercetools-backend/express': patch
---

Fix issue with invalid `audience` in case the request contains query parameters.
```

packages-backend/express/src/auth.ts
```diff
@@ -104,15 +104,15 @@ const getConfiguredDefaultIssuer = (options: TSessionMiddlewareOptions) => {
...
 // the token validation might fail because of mismatching audiences.
-const getConfiguredAudience = (
+export const getConfiguredAudience = (
   options: TSessionMiddlewareOptions,
   requestPath: string
 ) => {
   const url = new URL(options.audience);
   if (requestPath !== '/') {
-    url.pathname = requestPath;
+    return `${url.origin}${requestPath}${url.search}`;
```
emmenko, line 113: The `requestPath` already contains the search query

```suggestion
    return `${url.origin}${requestPath}`;
```

packages-backend/express/src/auth.ts
```diff
@@ -104,15 +104,15 @@ const getConfiguredDefaultIssuer = (options: TSessionMiddlewareOptions) => {
...
   }
-  return url.toString();
+  return `${url.origin}${url.search}`;
```
emmenko, line 115: I'm actually considering to completely remove the search params from the `audience` in the first place (when we forward the request and sign the token).

So we should fix this in both places (MC API and here).

emmenko (changes requested):

packages-backend/express/src/auth.ts
adnasa: that is actually a good call! let's go for it! 🙌

packages-backend/express/src/auth.ts
```diff
@@ -104,15 +104,15 @@ const getConfiguredDefaultIssuer = (options: TSessionMiddlewareOptions) => {
...
   options: TSessionMiddlewareOptions,
   requestPath: string
 ) => {
   const url = new URL(options.audience);
   if (requestPath !== '/') {
-    url.pathname = requestPath;
+    return `${url.origin}${requestPath}`;
   }
-  return url.toString();
+  return url.origin;
```
emmenko, line 115: We have to change this now to omit the search params. I suggest to parse everything into the `URL` constructor and then return `origin` and `pathname`.

```js
const url = new URL(`${options.audience.replace(/\/?$/, '')}${requestPath}`);
if (requestPath === '/') {
  return url.origin;
}
return `${url.origin}${url.pathname}`;
```

adnasa: 👍

adnasa: https://github.com/commercetools/merchant-center-application-kit/pull/2118/commits/2758c9d4bce4e1a7eeef36be517abd69c683d730

emmenko (approved):

PR #2131, opened 2021-04-08
chore: upgrade ui-kit to v12

adnasa:
#### Update

> ℹ️ This now targets https://github.com/commercetools/merchant-center-application-kit/pull/2041

#### Summary

- upgrades ui-kit to canary version of v12 (canary)
- adjusts components to remove use of deprecated props

#### Description

I opened this PR to check that the build passes and inspect the Percy snapshots.
If things are fine, I will:

1. release ui-kit v12
2. update this PR (versions and title)
3. changeset


@types/commercetools-uikit__design-system/index.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module '@commercetools-uikit/design-system' {
+  export const customProperties: {
+    [key: string]: string;
+  };
+}
```
adnasa, line 5: doing a `yarn typecheck` yielded errors, so I added this.
My intuition is that it doesn't make sense to do this, when the package itself was written in TS.
but I had a brainfart and couldn't figure out where to start addressing this properly.

packages/application-components/src/components/modal-pages/internals/modal-page-top-bar.tsx
```diff
@@ -96,8 +96,8 @@ const ModalPageTopBar = (props: Props) => {
         />
         {props.currentPathLabel && (
           <React.Fragment>
-            <Text.Detail isInline>/</Text.Detail>
-            <Text.Detail title={props.currentPathLabel} isInline truncate>
```
adnasa, line 100: ooooops! 
this has to be `as="span"`. 
what do you think?

packages/application-components/src/components/modal-pages/internals/modal-page-top-bar.tsx
```diff
@@ -96,8 +96,8 @@ const ModalPageTopBar = (props: Props) => {
         />
         {props.currentPathLabel && (
           <React.Fragment>
-            <Text.Detail isInline>/</Text.Detail>
```
adnasa, line 99: wait. why is this here?

adnasa:
this now points of `v19`

packages/application-components/src/components/modal-pages/internals/modal-page-top-bar.tsx
adnasa: fixed via https://github.com/commercetools/ui-kit/pull/1854

.changeset/small-grapes-cross.md
```diff
@@ -0,0 +1,26 @@
+---
```
adnasa, line 1: some of these packages were still on v17, @emmenko 
Shall we generate a "final" changeset, setting all packages to be v19?
what do you think?

adnasa:
oohh yeah, build passed 😎

.changeset/small-grapes-cross.md
emmenko: Sure, we can do that.

@types/commercetools-uikit__design-system/index.d.ts
emmenko: I can take a look, but this file should be removed.

emmenko (approved):

PR #2138, opened 2021-04-12
chore(package): add script to install canary version of ui-kit

adnasa:
#### Summary

Something I discovered we could do, in conjunction with the v12 migration of UI-kit.
I thought we should save the command under `scripts`


package.json
```diff
@@ -56,6 +56,7 @@
     "generate-types:settings": "./scripts/graphql-cli.sh codegen --project settings",
     "generate-types:proxy": "./scripts/graphql-cli.sh codegen --project proxy",
     "generate-types": "yarn generate-types:mc && yarn generate-types:ctp && yarn generate-types:settings && yarn generate-types:proxy",
+    "canary:uikit": "manypkg upgrade @commercetools-uikit canary",
```
emmenko, line 59: Thanks but I personally wouldn't add this script. You can always do it like `yarn versions:uikit canary`.

If you want, you can mention this in the README or something.

emmenko (changes requested):

PR #2139, opened 2021-04-13
fix(jest-preset): use Enzyme adapter for React v17

emmenko:
This is a quick fix for using a compatible adapter.

As a follow up we can drop Enzyme support and make it opt-in.


tdeekens (approved):

adnasa:
😍

PR #2140, opened 2021-04-13
refactor(create-mc-app): only check for latest version for outdated version hint

emmenko:
Just something I noticed.


adnasa (approved):

adnasa:
let's include it in #2141 ?

PR #2144, opened 2021-04-14
feat(test-utils): add renderHook wrapper

pa3:
The more logic we move to hooks in MC the more we observe the need to pass a `wrapper` combined from all the `application-shell` providers used by a hook under test in `renderHook` function in corresponding tests. For instance, whenever we want to test a hook which use Apollo under the hood, we have to wrap rendered hook in `<ApolloProvider />`. When testing logic related to tracking we wrap everything in `<GtmContext.Provider />`. And so on. 

Given that our `renderApp*` functions from `test-utils` already nicely mock all the providers one option is to test hooks via these `render*` function and not rely on `renderHooks`. Add people do it indeed. But that's not really nice, because of all the boilerplate needed to map hook inputs/outputs to a wrapping test component, so that it can be checked via rtl assertions.

Why not provide our own custom `renderHooks` wrapped in all the mocked providers in the same fashion we do with `renderApp*` utils and get rid off all the boilerplate?

That's basically what this PR is about. =)

### DIFF
--- a/.changeset/heavy-ligers-admire.md
+++ b/.changeset/heavy-ligers-admire.md
@@ -2,4 +2,25 @@
 '@commercetools-frontend/application-shell': minor
 ---
 
-feat(test-utils): add wrapped renderHook util
+Re-export `@testing-library/react-hooks` from `test-utils` with `renderHook` functions wrapped in all the app-kit providers.
+
+`renderHooks` is wrapped the same way the existing `renderAppWithRedux` helper is:
+
+- all the `options` which can be passed to `renderAppWithRedux`can be passed to `renderHooks` as well
+- all the additional poperties returned as a result of `renderAppWithRedux` invokation (like `store` or `history`) is returend from the wrapped `renderHooks` too.
+
+All the `@testing-library/react-hooks` content is exported under the namespace `hooks` from the package '@commercetools-frontend/application-shell/test-utils'.
+
+**Usage example**
+
+```jsx
+import { hooks } from '@commercetools-frontend/application-shell/test-utils';
+
+const { act, renderHook } = hooks;
+
+it('should navigate to a route', () => {
+  const { result, history } = renderHook(useRoutes);
+  act(() => result.current.someRoute.go());
+  expect(history.location.pathname).toBe('/some-route');
+});
+```
--- a/.changeset/heavy-ligers-admire.md
+++ b/.changeset/heavy-ligers-admire.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': minor
+---
+
+feat(test-utils): add wrapped renderHook util
--- a/.changeset/heavy-ligers-admire.md
+++ b/.changeset/heavy-ligers-admire.md
@@ -2,14 +2,14 @@
 '@commercetools-frontend/application-shell': minor
 ---
 
-Re-export `@testing-library/react-hooks` from `test-utils` with `renderHook` functions wrapped in all the app-kit providers.
+Adds re-export of `@testing-library/react-hooks` from `test-utils` with `renderHook` functions wrapped with app-kit providers.
 
-`renderHooks` is wrapped the same way the existing `renderAppWithRedux` helper is:
+The `renderHook` function is wrapped similarily to the existing `renderAppWithRedux`:
 
-- all the `options` which can be passed to `renderAppWithRedux`can be passed to `renderHooks` as well
-- all the additional poperties returned as a result of `renderAppWithRedux` invokation (like `store` or `history`) is returend from the wrapped `renderHooks` too.
+- All `options` which can be passed to `renderAppWithRedux` can be passed to `renderHooks`
+- All additional poperties returned as a result of an `renderAppWithRedux` call (like `store` or `history`) are returend from the wrapped `renderHook` too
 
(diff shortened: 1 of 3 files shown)


emmenko:
Thanks @pa3, I'll take a look at this later.

pa3:
@emmenko there is actually a sever issue in this implementation I just realized. `@testing-library/react-hooks` has it's own `act` and `waitFor` which have to be exported separately from `@testing-library/react`. At least intermixing `act`s leads to warnings. That's why I removed you from reviewers until I find a better organization for that.

adnasa:
ping me too when this is ready for review.
Thanks

adnasa (commented):
The diffs can be misleading when reviewing this, but let me see if I understood the steps you took.

1. We extract the common params from `renderApp` and `renderAppWithRedux` into a factory function `createApplicationProviders`
2. use `createApplicationProviders` inside the `renderHook`
2.1 `renderHook` includes the redux context.

.changeset/heavy-ligers-admire.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': minor
+---
+
+feat(test-utils): add wrapped renderHook util
```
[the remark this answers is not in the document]
adnasa: a small reference, here is how I did it in another PR

tdeekens (approved):
From how I read the code it's good.

We need to document the usage eventually in our documentation so others can read it on docs.ct.com. This can be a follow-up (please don't forget) once we tested it out internally.

adnasa:
> We need to document the usage eventually in our documentation so others can read it on docs.ct.com. 

Agree. What would be appreciated is that the Changelog includes the usage examples for now.
this will be good reference

tdeekens:
At best here https://docs.commercetools.com/custom-applications/development/testing#usage. Still, I don't think it has to be now but I trust us that we follow-up and want to urge us to if we merge this before.

packages/application-shell/src/test-utils/test-utils.tsx
```diff
@@ -595,7 +625,85 @@ function renderAppWithRedux<
...
+}
+
 // re-export everything
 export * from '@testing-library/react';
 
-export { renderApp, renderAppWithRedux };
+// namespace for hooks related helpers
+const hooks = {
+  ...rtlHooks,
+  renderHook,
```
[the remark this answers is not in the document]
pa3: I'd personally prefer to keep original naming not to introduce any confusion.

tdeekens (approved):

.changeset/heavy-ligers-admire.md
```diff
@@ -2,4 +2,25 @@
...
+import { hooks } from '@commercetools-frontend/application-shell/test-utils';
+
+const { act, renderHook } = hooks;
+
+it('should navigate to a route', () => {
+  const { result, history } = renderHook(useRoutes);
+  act(() => result.current.someRoute.go());
+  expect(history.location.pathname).toBe('/some-route');
+});
+```
```
adnasa: yeaaah boooooi 😍

adnasa (approved):
🚀 
thanks a lot for this, @pa3 !

.changeset/heavy-ligers-admire.md
```diff
@@ -0,0 +1,26 @@
...
+---
+
+Adds re-export of `@testing-library/react-hooks` from `test-utils` with `renderHook` functions wrapped with app-kit providers.
+
+The `renderHook` function is wrapped similarily to the existing `renderAppWithRedux`:
+
+- All `options` which can be passed to `renderAppWithRedux` can be passed to `renderHooks`
+- All additional poperties returned as a result of an `renderAppWithRedux` call (like `store` or `history`) are returend from the wrapped `renderHook` too
+
+All `@testing-library/react-hooks` functionality is exported under the namespace `hooks` from the package '@commercetools-frontend/application-shell/test-utils'.
```
emmenko, line 12:
```suggestion
```

.changeset/heavy-ligers-admire.md
```diff
@@ -0,0 +1,26 @@
+---
+'@commercetools-frontend/application-shell': minor
+---
+
+Adds re-export of `@testing-library/react-hooks` from `test-utils` with `renderHook` functions wrapped with app-kit providers.
```
emmenko, line 5:
```suggestion
The `@commercetools-frontend/application-shell/test-utils` now exports a new utility function for testing hooks: `renderHook`. This function wraps the original `renderHook` function from `@testing-library/react-hooks` but it comes pre-configured with providers for testing Custom Applications.
```

.changeset/heavy-ligers-admire.md
```diff
@@ -0,0 +1,26 @@
...
+
+The `renderHook` function is wrapped similarily to the existing `renderAppWithRedux`:
+
+- All `options` which can be passed to `renderAppWithRedux` can be passed to `renderHooks`
+- All additional poperties returned as a result of an `renderAppWithRedux` call (like `store` or `history`) are returend from the wrapped `renderHook` too
```
emmenko, line 10:
```suggestion
The new `renderHook` function also accepts the same options as other test utils, such as `renderApp`, and returns some useful properties like `store`, `history`, etc.
```

emmenko (approved):
Thanks, looks good! 👌

PR #2150, opened 2021-04-21
fix(jest-preset-mc-app): to use default v16 enzyme adapter

tdeekens:
#### Summary

This pull request suggests to default again to v16 of the enzyme adapter while offering a config option to opt into a newer version.

We noticed that the v17 adapter can't be as easily used as expeceted.


packages/jest-preset-mc-app/load-config.js
```diff
@@ -3,6 +3,12 @@ const { cosmiconfigSync } = require('cosmiconfig');
 const moduleName = 'jest-preset-mc-app';
 const explorer = cosmiconfigSync(moduleName);
 
+function createListMergerWithDefaults(key) {
```
tdeekens, line 6: Unrelated: hoisting fanciness is not needed I think.

packages/jest-preset-mc-app/setup-test-framework.js
```diff
@@ -13,7 +13,12 @@ const loadConfig = require('./load-config');
 
 const jestConfig = loadConfig();
 
-Enzyme.configure({ adapter: new Adapter(), disableLifecycleMethods: true });
+Enzyme.configure({
+  adapter: jestConfig.EnzymeAdapter
```
tdeekens, line 17: So we default again the the old version and offer an option to pass in any adapter you want. An alternative is a config string e.g. v17 and v16 but then both dependencies need to reside here and it gives the impression we support any version.

adnasa (approved):
thanks for the follow up!

tdeekens:
As another step the `EnzymeAdapter` should be required if it's desired and totally opt-in. This would be a breaking change though.

PR #2155, opened 2021-04-22
refactor(jest-preset-mc-app): remove relying on enzyme

tdeekens:
⚠️ Please consider this a discussion not a solution ⚠️  

#### Summary

This pull request tries to discuss in code how we can remove our jest config to rely on enzyme.

#### Description

Internally we don't write enzyme based tests anymore. However, we have a fair amount to run and wait for until they're replaced.

As a result, I suggest that we reverse that enzyme is an opt-in to the config.

Given, that this is a breaking change and we can potentially change anything in our config setup. 

The solution proposes:

1. `jestConfig.enzymeConfig`: similar to `rtlConfig` allowing to pass in an Enzyme version and associated config
2. `jestConfig.expectExtend`: similar allows to expand the matchers. We might need this as our matchers are not bound to support enzyme anymore
3. `jestConfig.preset` allows to pass in any part of the preset. Can be generally valuable but in this instance helps to overwrite and support the snapshot serializer

The above is the baseline to remove enzyme while keeping it an opt-in as one possible solution.

Additionally, I felt a `disableThorwingConsole` could help. I've seen internally that at times a throwing console might not be strictly needed. E.g. when working locally as it pollutes the stacktrace.

### DIFF
--- a/application-templates/starter/package.json
+++ b/application-templates/starter/package.json
@@ -43,7 +43,6 @@
     "@manypkg/cli": "0.17.0",
     "@testing-library/react": "11.2.6",
     "dotenv-cli": "4.0.0",
-    "enzyme": "3.11.0",
     "jest": "26.6.3"
   }
 }
--- a/package.json
+++ b/package.json
@@ -140,7 +140,6 @@
     "cross-env": "7.0.3",
     "cypress": "7.1.0",
     "dotenv": "8.2.0",
-    "enzyme": "3.11.0",
     "eslint": "7.24.0",
     "eslint-formatter-pretty": "4.0.0",
     "eslint-plugin-graphql": "4.0.0",
--- a/packages/jest-preset-mc-app/jest-preset.js
+++ b/packages/jest-preset-mc-app/jest-preset.js
@@ -1,9 +1,12 @@
 const path = require('path');
+const loadConfig = require('./load-config');
 
 // Resolve the absolute path of the caller location.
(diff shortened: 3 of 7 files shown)


emmenko:
Thanks, I actually started on this last week but didn't find the time to wrap up the PR.
You can take a look at the direction I was going to: https://github.com/commercetools/merchant-center-application-kit/compare/nm-enzyme-opt-in?expand=1

tdeekens:
> Thanks, I actually started on this last week but didn't find the time to wrap up the PR.
> You can take a look at the direction I was going to: [`nm-enzyme-opt-in`?expand=1 (compare)](https://github.com/commercetools/merchant-center-application-kit/compare/nm-enzyme-opt-in?expand=1)

Oh cool. Didn't know. Looks like a working other approach you took. We just push it into a preset. How would you take care of the snapshot serializer?

Do you generally have any preference? I can't fully make up my mind. I like your approach cause it pushes it into a corner and avoids supporting it using configuration. While I like mine as it gives more flexibility also for other matchers/expects/config without always having to adjust presets or configs here. Any opinion?

emmenko:
From a first look, I'm a bit hesitant to provide support via config, as it seems more work from our side.

tdeekens:
> From a first look, I'm a bit hesitant to provide support via config, as it seems more work from our side.

Ok, then I'll close and let you take it on :)

PR #2163, opened 2021-04-26
fix: handle HTTP responses with header 'x-refreshed-session-token' when using OIDC workflow

emmenko:


packages/application-shell/src/apollo-links/header-link.ts
```diff
@@ -103,7 +103,18 @@ const headerLink = new ApolloLink((operation, forward) => {
         : undefined,
     }),
   });
-  return forward(operation);
+  return forward(operation).map((response) => {
```
tdeekens, line 106: Can we add a test? We have tests for our links I believe.

tdeekens (approved):

PR #2165, opened 2021-04-27
feat(mc-scripts): add support for webpack config extensions

tdeekens:
#### Summary

This pull request intends to add support for webpack configs with different file endings such as `webpack.config.prod.cjs`

#### Description

The original reason for this pull request is to support webpack configs in projects running on ES modules. Those packages at times have a `type: "module"` in their `package.json`. 

Whenever, our `mc-scripts build` then loads the webpack configs from those modules the `package.json` config is read. Node.js will then throw as the e.g. `webpack.config.prod.js` contains `require` statements not only `import`.

ESLint solves this by supporting `.cjs` extension for their configurations while not supporting ES module configurations. This is a good middle ground to allow the ecosystem to migrate while not blocking them.

As a side-effect this pull request allows `ts` configurations which should be fine. It's a bit odd to support `jsx` but I am not sure if we want to prohibit it. Is so we could split our supported extension list more meaningfully.

### DIFF
--- a/.changeset/ninety-ravens-beam.md
+++ b/.changeset/ninety-ravens-beam.md
@@ -1,7 +1,9 @@
 ---
-"@commercetools-frontend/mc-scripts": patch
+"@commercetools-frontend/eslint-config-mc-app": minor
+"@commercetools-frontend/jest-preset-mc-app": minor
+"@commercetools-frontend/mc-scripts": minor
 ---
 
-Add support for webpack configurations with different file extensions.
+Adds support for the `*.mjs` and `*.cjs` JavaScript file extensions.
 
-Previously webpack configurations could only have the `.js` extension. In order to support ES modules we now also support the `*.cjs` extension.
+Updates the webpack configurations, Jest and ESLint presets to support the `*.mjs` and `*.cjs` extensions. This allows better integration with packages using ES modules.
--- a/.changeset/ninety-ravens-beam.md
+++ b/.changeset/ninety-ravens-beam.md
@@ -1,0 +1,7 @@
+---
+"@commercetools-frontend/mc-scripts": patch
+---
+
+Add support for webpack configurations with different file extensions.
+
+Previously webpack configurations could only have the `.js` extension. In order to support ES modules we now also support the `*.cjs` extension.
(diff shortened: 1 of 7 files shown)


.changeset/ninety-ravens-beam.md
```diff
@@ -0,0 +1,7 @@
+---
```
emmenko, line 1: We should add other changesets for the other packages as well (eslint, jest, etc).

.changeset/ninety-ravens-beam.md
```diff
@@ -0,0 +1,7 @@
...
+"@commercetools-frontend/mc-scripts": patch
+---
+
+Add support for webpack configurations with different file extensions.
+
+Previously webpack configurations could only have the `.js` extension. In order to support ES modules we now also support the `*.cjs` extension.
```
emmenko, line 7: I guess also `.mjs`?

packages/mc-scripts/src/config/paths.js
```diff
@@ -1,14 +1,15 @@
...
 
-const moduleFileExtensions = ['js', 'jsx', 'ts', 'tsx'];
+const moduleFileExtensions = ['js', 'mjs', 'cjs', 'jsx', 'ts', 'tsx'];
 
 // Make sure any symlinks in the project folder are resolved:
 // https://github.com/facebook/create-react-app/issues/637
 const appDirectory = fs.realpathSync(process.cwd());
-const resolveApp = (relativePath) => path.resolve(appDirectory, relativePath);
+const resolveApp = (relativePath, extentions) =>
+  path.resolve(appDirectory, relativePath);
```
emmenko, line 10:
```suggestion
const resolveApp = (relativePath) => path.resolve(appDirectory, relativePath);
```

emmenko (commented):
If you need help with the linters let me know.

.changeset/ninety-ravens-beam.md
tdeekens: Yeap. I was just pilling up the internal patches. Thanks for the reminder.

tdeekens (commented):
Ready for review @emmenko :)

emmenko (approved):
Lovely ❤️

islam3zzat (approved):

PR #2170, opened 2021-04-29
fix(mc-html-template): permission policy to be joined with comma

tdeekens:
#### Summary

The permission policy header is joined by a `', '` separator in the specification. Not a `': '`.

CleanShot 2021-04-29 at 14 04 11@2x

### DIFF
--- a/packages/mc-html-template/src/process-headers.js
+++ b/packages/mc-html-template/src/process-headers.js
@@ -24,7 +24,7 @@ const mergeCspDirectives = (...csps) =>
       ),
     {}
   );
-const toHeaderString = (directives = {}) =>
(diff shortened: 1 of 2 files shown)


packages/mc-html-template/src/process-headers.js
```diff
@@ -24,7 +24,7 @@ const mergeCspDirectives = (...csps) =>
       ),
     {}
   );
-const toHeaderString = (directives = {}) =>
+const toHeaderString = (directives = {}, seperator = '; ') =>
```
emmenko, line 27: The `seperator` does not seem to be used.

islam3zzat (approved):

PR #2172, opened 2021-04-29
fix(mc-html-template): to use separator

tdeekens:
#### Summary

No words needed.


emmenko (approved):
Thanks

tdeekens:
No, thank you for catching it.

PR #2183, opened 2021-05-03
fix(mc-html-template): fix processing of structured headers

tdeekens:
#### Summary

This fixes the processing of structured headers like the permissions policy.

A structured header format is `key=(), key2=()` not `key (),`.

In this fix I removed the seperator and created a second function. I subjectively found that easier to understand.

### DIFF
--- a/packages/mc-html-template/src/process-headers.js
+++ b/packages/mc-html-template/src/process-headers.js
@@ -24,13 +24,20 @@ const mergeCspDirectives = (...csps) =>
       ),
     {}
   );
-const toHeaderString = (directives = {}, seperator = '; ') =>
+const toHeaderString = (directives = {}) =>
   Object.entries(directives)
     .map(
       ([directive, value]) =>
         `${directive} ${Array.isArray(value) ? value.join(' ') : value}`
     )
-    .join(seperator);
+    .join('; ');
+const toStructuredHeaderString = (directives = {}) =>
+  Object.entries(directives)
+    .map(
+      ([directive, value]) =>
+        `${directive} ${Array.isArray(value) ? value.join('=') : value}`
+    )
+    .join(', ');
 
 const processHeaders = (applicationConfig) => {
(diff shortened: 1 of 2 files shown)


packages/mc-html-template/src/process-headers.js
```diff
@@ -24,13 +24,20 @@ const mergeCspDirectives = (...csps) =>
...
-const toHeaderString = (directives = {}, seperator = '; ') =>
+const toHeaderString = (directives = {}) =>
   Object.entries(directives)
     .map(
       ([directive, value]) =>
         `${directive} ${Array.isArray(value) ? value.join(' ') : value}`
     )
-    .join(seperator);
+    .join('; ');
+const toStructuredHeaderString = (directives = {}) =>
```
emmenko, line 34: I wouldn't mind to have a specific function for mapping the permissions policy directives. Then it's clear what it does, especially the part with the `=`.

emmenko: And by that I mean naming the function specific for the permissions policy header.

tdeekens: Yes and no. The format isn't specific to it. It's a structured header to how I understand the spec. I think having a mapper per header type not value makes sense. At least to me. Now I know that a permissions policy header is a structured header by reading the code.

emmenko: I didn't read the spec. What does "structured header" mean exactly?

tdeekens: To what I understand it's the format: `key=(), key2=(), key3=()`.

emmenko: Ok got it, I did some googling and that's the name of this data structure. Thanks

https://www.w3.org/TR/permissions-policy-1/#permissions-policy-http-header-field
image

emmenko (approved):

emmenko:
Before merging this, let me release the Jest changes first

PR #2175, opened 2021-04-29
[Breaking] Drop Enzyme support from Jest preset and make it opt-in

emmenko:
Ref #1829

PR #2184, opened 2021-05-03
refactor: improve Jest setup for Enzyme support

emmenko:
Follow up of #2175

### DIFF
--- a/packages/jest-preset-mc-app/README.md
+++ b/packages/jest-preset-mc-app/README.md
@@ -75,7 +75,10 @@ const jestPreset = require('@commercetools-frontend/jest-preset-mc-app');
 const createJestPresetWithEnzyme = require('@commercetools-frontend/jest-preset-mc-app/enzyme/create-jest-preset-with-enzyme');
 
 module.exports = {
-  ...createJestPresetWithEnzyme(16, jestPreset),
+  ...createJestPresetWithEnzyme({
+    enzymeAdapterVersion: 16,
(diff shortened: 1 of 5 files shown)


packages/jest-preset-mc-app/README.md
```diff
@@ -66,22 +66,21 @@ yarn add \
...
 const jestPreset = require('@commercetools-frontend/jest-preset-mc-app');
-const jestPresetForEnzyme16 = require('@commercetools-frontend/jest-preset-mc-app/enzyme/jest-preset-for-enzyme-16');
+const createJestPresetWithEnzyme = require('@commercetools-frontend/jest-preset-mc-app/enzyme/create-jest-preset-with-enzyme');
 
 module.exports = {
-  preset: {
-    ...jestPreset,
-    ...jestPresetForEnzyme16,
-  },
+  ...createJestPresetWithEnzyme(16, jestPreset),
```
tdeekens, line 78: Should we do named object arguments? I mean I know what `16` is but can I do `23`? :)

emmenko: You can try but an error will be thrown that the file does not exist.

How would you expect this to be with named arguments?

tdeekens: Something like. `createJestPresetWithEnzyme({ enzypeAdapterVersion: 16, jestPreset })`

tdeekens (approved):

PR #2186, opened 2021-05-04
test(mc-html-template): add test for processing headers

tdeekens:
#### Summary

This adds tests for processing the headers of an application config.

I felt we should have some tests for this logic given it didn't work as we wanted. While writing the tests I noticed that it still doesn't 🤕.

### DIFF
--- a/packages/mc-html-template/src/process-headers.spec.js
+++ b/packages/mc-html-template/src/process-headers.spec.js
@@ -1,0 +1,98 @@
+import processConfig from './process-headers';
+
+const defaultApplicationConfig = {
+  name: 'Application test',
+  entryPointUriPath: 'test',
+  cloudIdentifier: 'test',
+  env: {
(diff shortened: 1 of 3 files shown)


packages/mc-html-template/src/process-headers.spec.js
```diff
@@ -0,0 +1,98 @@
...
+            camera: ["'https://example.com'"],
+          },
+        },
+      };
+
+      const processedApplicationConfig = processConfig(testApplicationConfig);
+
+      expect(
+        processedApplicationConfig['Feature-Policy']
+      ).toMatchInlineSnapshot(
```
tdeekens, line 30: I find this more convenient than `toEqual` cause it would mean copy pasting while this nicely shows the diff to accept under test.

emmenko (approved):
Thanks!

Issue #2188, opened 2021-05-04
Build errors do not display correctly using `mc-scripts` to build


dgoemans:
**Describe the issue**
When using the MC application starter template with `mc-scripts`, any build errors generate the error message: `message.split is not a function`.

**To Reproduce**

1. Create a project from the starter template (`create-mc-app my-new-custom-application-project --template starter`)
2. Have a blatant compilation error (for example,  change `import loadMessages from '../../load-messages';` to `import . from '../../load-messages';` in `entry-point.js`)
3. Run `yarn build`
4. Note that the error message from the build output is `message.split is not a function` instead of indicating the error in your code.

**Expected behavior**

The build output displays the error in your code

**Screenshots**

**Additional context**
This seems to be related to an issue where `react-dev-utils` doesn't work with Webpack 5's stats. I have a PR incoming with a fix until it is supported.

tdeekens:
Thanks for opening the issue. We actually ran into it internally too. Basically, noticing that (as you) `react-dev-utils` doesn't yet support `webpack@5`. We noticed that it's related to error message formats. Not sure if we can fix it on our side but interested in how you would approach it. 🙇🏼

dgoemans:
@tdeekens, the approach that I did locally was to write some adaptation code in the `mc-scripts` interaction with `react-dev-utils`. This will work until `react-dev-utils` updates to using webpack 5.
PR is open here:
https://github.com/commercetools/merchant-center-application-kit/pull/2189

tdeekens:
Cool. Let's continue discussing on the PR :)

emmenko:
Fix has been released: `@commercetools-frontend/mc-scripts@20.0.3`.

Thanks again!

PR #2189, opened 2021-05-04
Bugfix for error messages not showing in `mc-build`

dgoemans:
#### Summary
React-dev-utils doesn't support the webpack 5 stats, so this is an adapter to allow errors to propogate properly to `mc-scripts build`.

#### Description
Resolves https://github.com/commercetools/merchant-center-application-kit/issues/2188

### DIFF
--- a/.changeset/tricky-crews-breathe.md
+++ b/.changeset/tricky-crews-breathe.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/mc-scripts': patch
+---
+
+Fix error messages not showing in mc-build
--- a/.changeset/tricky-crews-breathe.md
+++ b/.changeset/tricky-crews-breathe.md
@@ -2,4 +2,4 @@
 '@commercetools-frontend/mc-scripts': patch
 ---
 
-Fix error messages not showing in mc-build
+Temporary fix for parsing Webpack error messages until `react-dev-utils` supports Webpack v5.
--- a/packages/mc-scripts/src/build.js
+++ b/packages/mc-scripts/src/build.js
@@ -118,6 +118,7 @@ function build(previousFileSizes) {
           warnings: [],
         });
       } else {
+        //TODO: Remove after 'react-dev-utils' natively supports webpack@5
         const rawMessages = stats.toJson({
           all: false,
           warnings: true,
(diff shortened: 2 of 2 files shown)


packages/mc-scripts/src/build.js
```diff
@@ -118,9 +118,16 @@ function build(previousFileSizes) {
...
+        const rawMessages = stats.toJson({
+          all: false,
+          warnings: true,
+          errors: true,
+          moduleTrace: false,
+        });
+        messages = formatWebpackMessages({
+          errors: rawMessages.errors.map((e) => e.message),
+          warnings: rawMessages.warnings.map((e) => e.message),
+        });
```
tdeekens, line 130: Cool approach. Thinking about it "out load":

Do you think it would be possible to "sniff" the format. So see if we need to map the errors? Should we actually do that? Wondering what happens when `react-dev-utils` gets fixed 🤔.

dgoemans: We probably could with some inspection on fields of the `stats.toJson` result. The only thing i think is that when `react-dev-utils` does move to webpack 5, it's likely going to be part of a very large update.

Another proposal, since it's one tiny function, could be to pull `formatWebpackMessages` into this repo and support webpack 5 out of the box?

dgoemans: Interestingly, they have a PR open to fix it on their side with the detection, maybe we can lean on that instead:
https://github.com/facebook/create-react-app/pull/10098/files

tdeekens: Interesting. Seems like it would continue to work. Maybe you can add a `//TODO: Remove after `react-dev-utils` natively supports webpack@5` to your code so we don't forget :)

dgoemans: Done 😄

tdeekens (approved):
Nice. Thanks 💯. Totally fine by me given it's temporary and actually solves a quite annoying issue.

Waiting for @emmenko to also confirm.

.changeset/tricky-crews-breathe.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-frontend/mc-scripts': patch
+---
+
+Fix error messages not showing in mc-build
```
emmenko, line 5:
```suggestion
Temporary fix for parsing Webpack error messages until `react-dev-utils` supports Webpack v5.
```

emmenko (commented):
Ah thanks for finding a workaround!

emmenko (approved):

PR #2196, opened 2021-05-06
fix(mc-scripts): dotenv path resolution, remove experimental `--match` flag

emmenko:
The `--match` flag is not that helpful at the end, because you would then need to provide the dotenv files etc.

PS: I think that prompting for packages could be something more suited for something else like `@manypkg`.


adnasa (approved):
👍🏽

PR #2203, opened 2021-05-11
fix(babel-preset-mc-app): to add loose private methods

tdeekens:
#### Summary

We're receiving a lot of warnings internally when building with the recent babel update with about 40k lines of this on CI:

CleanShot 2021-05-11 at 09 42 25@2x

```
Though the "loose" option was set to "false" in your @babel/preset-env config, it will not be used for @babel/plugin-proposal-private-methods since the "loose" mode option was set to "true" for @babel/plugin-proposal-class-properties.
The "loose" option must be the same for @babel/plugin-proposal-class-properties, @babel/plugin-proposal-private-methods and @babel/plugin-proposal-private-property-in-object (when they are enabled): you can silence this warning by explicitly adding
```


emmenko:
Thanks but unfortunately that does not fully work.

Last time I tried to fix that, Cypress has problems: https://github.com/commercetools/merchant-center-application-kit/runs/2509665796

I haven't looked too much into it but apparently something else needs to be fixed/changed as well.
If you have the time, feel free to look into it.

tdeekens:
For others and me to document:

We use `loose` on class properties as the otherwise used `defineProperty` comes with a performance penalty. Currently `create-react-app` doesn't have and special configuration for private methods.

When true, private methods will be assigned directly on its parent via Object.defineProperty rather than a WeakSet. This results in improved performance and debugging (normal property access vs .get()) at the expense of potentially leaking "privates" via things like Object.getOwnPropertyNames.

So I don't see why Cypress chokes on this as both syntaxes should work. Will investigate further.

Reference: https://babeljs.io/docs/en/babel-plugin-proposal-private-methods

babel.config.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
         // TODO: change this to `automatic` when using the new runtime
         runtime: 'classic',
         keepPropTypes: true,
+        enableLooseMode: process.env.CYPRESS_CI === 'true' ? false : true,
```
tdeekens, line 9: Cypress hat its own webpack config. Even when enabling all three plugins to be loose Cypress seems to overwrite it as it doesn't want to run in loose mode. We can however disable loose mode. It's even better for debugging but slower for performance. There won't be a browser compatibility difference.

emmenko: I would probably make this opt out.

```suggestion
        disableLooseMode: process.env.CYPRESS_CI === 'true',
```

.changeset/tiny-kids-wait.md
```diff
@@ -0,0 +1,5 @@
+---
+"@commercetools-frontend/babel-preset-mc-app": patch
+---
+
+Fixes `babel-preset-mc-app` warnings by adding loose configuration to private methods
```
emmenko, line 5:
```suggestion
Fixes `babel-preset-mc-app` warnings about `loose` mode. You can now pass a `disableLooseMode` option to this preset to turn it off (defaults to `true`).
```

packages/babel-preset-mc-app/index.js
```diff
@@ -5,6 +5,11 @@ const defaultOptions = {
   // Usually when bundling packages we want to keep the prop types
   // but when building the final application we can remove them.
   keepPropTypes: false,
+  enableLooseMode: true,
```
emmenko, line 8:
```suggestion
  disableLooseMode: false,
```

packages/babel-preset-mc-app/index.js
```diff
@@ -108,7 +113,19 @@ module.exports = function getBabePresetConfigForMcApp(api, opts = {}) {
...
+      [
+        require('@babel/plugin-proposal-private-methods').default,
+        {
+          loose: options.enableLooseMode,
+        },
+      ],
+      [
+        require('@babel/plugin-proposal-private-property-in-object').default,
+        {
+          loose: options.enableLooseMode,
```
emmenko, line 128:
```suggestion
          loose: options.disableLooseMode,
        },
      ],
      [
        require('@babel/plugin-proposal-private-methods').default,
        {
          loose: options.disableLooseMode,
        },
      ],
      [
        require('@babel/plugin-proposal-private-property-in-object').default,
        {
          loose: options.disableLooseMode,
```

emmenko (commented):
Good idea!

emmenko (approved):

emmenko (approved):

PR #2204, opened 2021-05-11
chore: downgrade cypress to 7.2.0

emmenko:
It seems that Cypress `7.3.0` has issues being downloaded. Let's downgrade to `7.2.0` in the meantime.


tdeekens:
Hm. I had issues with both 404ing. Not 100% sure it's related to the version.

PR #2210, opened 2021-05-11
fix(mc-scripts): regression when updating webpack-dev-server to 4.0.0-beta.3

emmenko:
Caused by #2199 

This slipped through in the latest dependency upgrade

image

causing local dev to not start:

image


tdeekens (approved):
Wow. Sneaky.

tdeekens:
Yes, that's what I thought too. Wasn't meaning to imply that we should replace the production bundle testing with it.

PR #2211, opened 2021-05-12
chore: add tsc-files to pre-commit hook

tdeekens:
#### Summary

The project [tsc-files](https://github.com/gustavopch/tsc-files) allows to type check only files that changed. I wanted to suggest it as a pre-commit hook while CI still does the whole typecheck. I found out about it as the `lint-staged` owner recommended it online.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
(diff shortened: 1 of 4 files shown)


emmenko (approved):
Ah perfect, I was looking for something like that. Thanks! 💯

adnasa (approved):
👌🏽

PR #2228, opened 2021-05-24
feat: new placeholder for missing values (i.e. NO_VALUE_FALLBACK)

pa3:
There is a design decision to go from `- - - -` to `- -` across the MC. As `NO_VALUE_FALLBACK` is used in Application Shell too we have to change it in this repo. 

I guess that should be considered a braking change (custom apps relies on current value), hence there is a major version bump.

### DIFF
--- a/.changeset/curvy-ears-move.md
+++ b/.changeset/curvy-ears-move.md
@@ -1,5 +1,5 @@
 ---
-'@commercetools-frontend/constants': major
+'@commercetools-frontend/constants': minor
 ---
 
 Change missing value placholder (i.e. `NO_VALUE_FALLBACK`) from `- - - -` to `- -`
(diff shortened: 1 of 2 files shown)


tdeekens:
Thanks. In what way to you mean rely? As this is "only" a visual change I am not entirely sure it has to be considered breaking. At least such changes haven't been in the past to my knowledge.

pa3:
@tdeekens yeah, it's only about visual change. And, as you said some tests may start breaking after the change on app-kit consumers side. I was not sure what was general app-kit's policy about that kind of changes. But if it's OK to mark that as a minor version bump, I will happily change changeset file. =)

tdeekens:
Thanks for considering that! From a purist stand point I would agree that it can be breaking. However, from a pragmatic view point I'd say this is a minor change. I'd just ship it as that to be honest :-)

tdeekens (approved):

tdeekens:
I expect this to require some internal refactorings though.

PR #2232, opened 2021-05-25
chore: nvmrc to node.js v16

tdeekens:


tdeekens:
Will take the liberty to merge this as everything passes and I can't find any information why we're on node.js v14.

PR #2237, opened 2021-05-31
fix(eslint-config): to not error on new rtl rules

tdeekens:
#### Summary

Updating and hence enabling these rules is a breaking change.


tdeekens:
Merging to unblock internal updates.

PR #2242, opened 2021-06-04
fix(mc-html-template): add preload and includeSubDomains to header

tdeekens:
#### Summary

This adds the `preload` and `includeSubDomains` attributes to the `Strict-Transport-Security` header.

#### Description

Using [hstspreload.org](https://hstspreload.org) it is advised to `preload` (not part of the spec) and add the `includeSubDomains` attributes. More about HTS [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security).

CleanShot 2021-06-04 at 09 08 46@2x

I don't see a strict reason against adding those.

### DIFF
--- a/.changeset/fuzzy-hounds-agree.md
+++ b/.changeset/fuzzy-hounds-agree.md
@@ -1,0 +1,5 @@
+---
+"@commercetools-frontend/mc-html-template": patch
+---
+
+Add `preload` and `includeSubDomains` to `Strict-Transport-Security` header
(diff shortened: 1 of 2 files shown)


islam3zzat (approved):

mohib0306 (approved):

tdeekens:
I was hesitant of the `includeSubDomain` directive. However, I checked and all subdomains under `ct.com` have HTTPs enabled so there shouldn't be any issues.

tdeekens:
Confirmed with SRE, using `includeSubDomains` is fine.

PR #2244, opened 2021-06-04
feat(mc-html-template): make transport security configurable

tdeekens:
#### Summary

Instead of changing the default let's make it configurable.

### DIFF
--- a/packages/application-config/schema.json
+++ b/packages/application-config/schema.json
@@ -158,10 +158,7 @@
         },
         "strictTransportSecurity": {
           "description": "Additional configuration for the HTTP Strict-Transport-Security header (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security)",
-          "type": "array",
-          "items": {
-            "$ref": "#/definitions/hstsDirective"
-          }
+          "$ref": "#/definitions/hstsDirective"
         }
       },
       "additionalProperties": false,
--- a/packages/application-config/schema.json
+++ b/packages/application-config/schema.json
@@ -14,6 +14,13 @@
         "type": "string"
       },
       "uniqueItems": true
+    },
+    "hstsDirective": {
+      "type": "array",
+      "items": {
+        "enum": ["includeSubDomains", "preload"]
+      },
+      "uniqueItems": true
     }
   },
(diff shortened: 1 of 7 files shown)


pa3 (approved):

packages/application-config/schema.json
```diff
@@ -148,6 +155,10 @@
         "permissionsPolicies": {
           "description": "Configuration for the HTTP Permissions-Policy header (https://github.com/w3c/webappsec-permissions-policy/blob/main/permissions-policy-explainer.md)",
           "type": "object"
+        },
+        "strictTransportSecurity": {
+          "description": "Additional configuration for the HTTP Strict-Transport-Security header (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security)",
+          "$ref": "#/definitions/hstsDirective"
```
emmenko, line 161: Nit: a ref is only useful if you reuse the object within the schema. In this case you could just "inline" it.

packages/mc-html-template/src/process-headers.js
```diff
@@ -120,7 +120,9 @@ const processHeaders = (applicationConfig) => {
   );
 
   return {
-    'Strict-Transport-Security': 'max-age=31536000; includeSubDomains; preload',
```
emmenko, line 123: So the `preload` option is now NOT a default anymore? Any particular reason for that?

packages/application-config/schema.json
tdeekens: Not sure I recall 100% but inlining it might have produced different typings.

packages/mc-html-template/src/process-headers.js
tdeekens: The preload wasn't default before. The steps we took where while we only had `max-age`:

1. Add includeSubDomains and preload
2. Reflect and consider making it configurable and hence remove the just added defaults

All cause we wanted to integrate this change slowly one app first. Now that it runs for a week or two we could remove the configuration option or change the default.

emmenko: Ah, I didn't see the other PR #2242 🤦‍♂️

PR #2245, opened 2021-06-04
fix(docs): to allow docs embedding themselves in iframe

tdeekens:
#### Summary

This is an attempt to work "against" an internal change to allow embedding the docs inside an iframe for interactive content.


tdeekens:
In hindsight I am not sure this will work. Not 100% sure of the set up.

PR #2246, opened 2021-06-04
Custom application custom object tutorial

anilkk:
#### Summary
Custom Objects Custom Application tutorial.


anilkk:
Here is the demo application based on the tutorial.
https://github.com/anilkk/custom-objects-custom-application-project

PR #2257, opened 2021-06-09
test: disable core-js in test environment

pa3:
`core-js` causes troubles it tests when running together with `jest@27`. 

Beacuse `core-js` adds polyfills for `Promises` (and maybe `Date`, I'm not sure), that breaks some assumptions under the hood of resent version of `jest`. Which in turn leads to all tests relying on jest's fake timers to break.

I ran both app-kit and MC tests with this new setup and it seems to work fine.

### DIFF
--- a/packages/babel-preset-mc-app/index.js
+++ b/packages/babel-preset-mc-app/index.js
@@ -141,7 +141,9 @@ module.exports = function getBabePresetConfigForMcApp(api, opts = {}) {
       [
         require('@babel/plugin-transform-runtime').default,
         {
-          corejs: 3,
+          // corejs messes with jest@27 in tests, due to some
+          // Promise/Date related polyfills it provides.
+          corejs: !isEnvTest && 3,
           // To be able to use `runtime` in Rollup babel plugin
           helpers: true,
           regenerator: true,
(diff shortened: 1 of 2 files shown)


packages/babel-preset-mc-app/index.js
```diff
@@ -141,7 +141,9 @@ module.exports = function getBabePresetConfigForMcApp(api, opts = {}) {
       [
         require('@babel/plugin-transform-runtime').default,
         {
-          corejs: 3,
+          // corejs messes with jest@27 in tests, due to some
+          // Promise/Date related polyfills it provides.
+          corejs: !isEnvTest && 3,
```
tdeekens, line 146: Do we know if there are Node.js versions require `core-js` under test?

Alternatively we could have a `disableCoreJs` flag which however might be hard to pass from a `jest.test.config.js`. However, it could be a good idea if we want to make this change opt-in.

pa3: Good point. Will try with node@12. Is there any commitment related lowest node version we want to support?

tdeekens (approved):
Thanks. All good under the assumption that it works under Node.js v16 (CI) and v14. 

Generally the env should be set correctly and can yield side-effects like this.

tdeekens:
Should we also bump the Node.js version to at least v14 or current (process.env.version) in the babel-preset-env when running under test?

PR #2260, opened 2021-06-10
fix(babel-preset-mc-app): to set node.js version to current

tdeekens:
#### Summary

This sets the Node.js version to `current` instead of hard-coding it to `8`.

#### Description

Current implies the the version of the hosting system will be used which under test is often exactly what you want given everything runs in the version prescribed by the `.nvmrc` if you ask me.

CleanShot 2021-06-10 at 08 02 22@2x

### DIFF
--- a/.changeset/proud-cows-own.md
+++ b/.changeset/proud-cows-own.md
@@ -1,0 +1,7 @@
+---
+"@commercetools-frontend/babel-preset-mc-app": patch
+---
+
+Set Node.js version to `current` in the `babel-preset-mc-app` for `babel-preset-env`
+
+The previous version was hard-coded to `8` while `current` whill always use the version of the running process.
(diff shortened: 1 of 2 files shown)


mohib0306 (approved):

.changeset/proud-cows-own.md
```diff
@@ -0,0 +1,7 @@
+---
+"@commercetools-frontend/babel-preset-mc-app": patch
+---
+
+Set Node.js version to `current` in the `babel-preset-mc-app` for `babel-preset-env`
+
+The previous version was hard-coded to `8` while `current` whill always use the version of the running process.
```
emmenko, line 7: This is only valid for the test environment, right?

tdeekens: Yes, it's within a block of `isEnvTest`. If means less transpilation for Node.js as otherwise we transpile for 8 to run in 16 which is wasteful.

PR #2261, opened 2021-06-10
feat(babel-preset-mc-app): add ability to disable core-js

tdeekens:
#### Summary

This adds the ability to explicitly disable `core-js`.

#### Description

Some environments or build targets might not require `core-js`. In order to opt-out of it you can specify a `disableCoreJs` option now.

### DIFF
--- a/packages/babel-preset-mc-app/index.js
+++ b/packages/babel-preset-mc-app/index.js
@@ -5,11 +5,15 @@ const defaultOptions = {
   // Usually when bundling packages we want to keep the prop types
   // but when building the final application we can remove them.
   keepPropTypes: false,
-  disableLooseMode: false,
   // plugin-proposal-class-properties, plugin-proposal-private-methods,
   // plugin-proposal-private-property-in-object have a loose option which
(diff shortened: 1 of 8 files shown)


packages/babel-preset-mc-app/index.js
```diff
@@ -143,7 +149,7 @@ module.exports = function getBabePresetConfigForMcApp(api, opts = {}) {
         {
           // corejs messes with jest@27 in tests, due to some
           // Promise/Date related polyfills it provides.
-          corejs: !isEnvTest && 3,
+          corejs: isEnvTest || options.disableCoreJs ? false : 3,
```
mohib0306, line 152: This would enable `corejs` in test env all the time regardless of the `disableCoreJs` option.

mohib0306 (approved):

packages/babel-preset-mc-app/index.js
tdeekens: You mean disable, right? If so yes, that's what @pa3 implemented prior and intended.

PR #2280, opened 2021-06-29
feat(i18n): i18n messages from `@commercetools-community-kit/i18n`

emmenko:
Note (**tech debt**): we might want/need to rethink how to ship and consume i18n messages from packages/repositories. At the moment, they are all loaded from the i18n package in appkit, but it's not really scalable. Ideally, packages should ship they own messages instead of relying on a single i18n package that aggregates all messages. See also https://formatjs.io/docs/getting-started/message-distribution

### DIFF
--- a/packages/i18n/src/load-i18n.ts
+++ b/packages/i18n/src/load-i18n.ts
@@ -111,49 +111,31 @@ const getCommunityKitChunkImport = async (
   locale: string
 ): Promise<I18NImportData> => {
   const intlLocale = mapLocaleToIntlLocale(locale);
-  try {
-    /* eslint-disable import/no-unresolved */
-    switch (intlLocale) {
-      case 'de':
-        return await import(
-          // @ts-expect-error: it's an optional dependency.
-          /* webpackChunkName: "i18n-community-kit-locale-de" */ '@commercetools-community-kit/i18n/compiled-data/de.json'
-        );
-      case 'es':
-        return await import(
-          // @ts-expect-error: it's an optional dependency.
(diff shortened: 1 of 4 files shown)


emmenko:
Dang it, this is harder than I thought...

packages/i18n/src/load-i18n.ts
```diff
@@ -110,6 +107,56 @@ const getAppKitChunkImport = (locale: string): Promise<AppKitImportData> => {
   }
 };
 
+const getCommunityKitChunkImport = async (
```
tdeekens, line 110: Just thinking out loud: in the future each package has it's own loadTranslationsChunkByLocale. While then AppKit offers a registerTranslationsChunk which is then called in each application for each loader? Breaking change, more on consumers but decouples things. Or did you already think of something else?

tdeekens (commented):
Optional dependencies are hard.

packages/i18n/src/load-i18n.ts
emmenko: Maybe something like that yes.

I don't have any specific idea at the moment. We need to come up with a scalable and easy-to-use approach I would say.

emmenko:
So I guess we would need to include the dependency as a normal one for now. This means the i18n messages of community kit will be included in the bundle.

tdeekens:
> So I guess we would need to include the dependency as a normal one for now. This means the i18n messages of community kit will be included in the bundle.

What if bind the loading to `process.env.ENABLE_COMMUNITY_KIT` or sth like that?

emmenko:
I think for now we can include the dependency. The bundle size difference is not a problem for now. It was more for the future as we scale further. There is a chapter issue to discuss about this topic.

tdeekens (approved):

PR #2281, opened 2021-06-30
fix(navbar): keep main menu item a link even when submenu group is expanded

### LINKED ISSUE #2164: Navbar main application link only works if menu is collapsed
**Describe the issue**

Each Custom Application has a main route, determined by the `entryPointUriPath`. This is also reflected in the `menu.json`.

For example, the application `avengers` would have the main menu link to be:

```
/<project_key>/avengers
```

However, this link is only used when the navbar component is in the collapsed state.

image

When the navbar is expanded, there is no main link anymore, because clicking on the item expands/collapses the submenu links.

image

PS: This is also an issue with our internal applications, although we "mask" the issue because we also have sublinks that point to the same main route (for example in products)

image

**Expected behavior**

The main application route link should be respected and there should be no need to have sublinks pointing to the same main link.

emmenko:
Fixes #2164

### DIFF
--- a/packages/application-shell/src/components/navbar/navbar.tsx
+++ b/packages/application-shell/src/components/navbar/navbar.tsx
@@ -180,7 +180,7 @@ const MenuGroup = (props: MenuGroupProps) => {
     <ul
       id={`${props.id}-group`}
       data-testid={`${props.id}-group`}
-      role="menu"
+      role="link"
       aria-expanded={
         isSublistActiveWhileIsMenuExpanded ||
         isSublistActiveWhileIsMenuCollapsed
@@ -221,7 +221,7 @@ type MenuItemProps = {
 };
 const MenuItem = (props: MenuItemProps) => (
   <li
-    role="menu-item"
+    role="link"
     className={classnames(styles['list-item'], {
       [styles.item__active]: props.isActive,
       [styles['item_menu-collapsed']]: !props.isMenuOpen,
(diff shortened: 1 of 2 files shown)


packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -180,7 +180,7 @@ const MenuGroup = (props: MenuGroupProps) => {
     <ul
       id={`${props.id}-group`}
       data-testid={`${props.id}-group`}
-      role="menu"
+      role="link"
```
emmenko, line 183: Unrelated, eslint a11y was complaining.

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -221,7 +221,7 @@ type MenuItemProps = {
 };
 const MenuItem = (props: MenuItemProps) => (
   <li
-    role="menu-item"
+    role="link"
```
emmenko, line 224: Unrelated, eslint a11y was complaining.

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -428,6 +428,14 @@ const ApplicationMenu = (props: ApplicationMenuProps) => {
     return Boolean(match);
   };
 
+  const isMainMenuItemALink =
+    // 1. When the navbar is collapsed
+    !props.isMenuOpen ||
+    // 2. When there is no submenu
+    !hasSubmenu ||
+    // 3. When the submenu group is active/visible
+    props.isActive;
```
emmenko, line 437: This is the case we're fixing here.

packages/application-shell/src/components/navbar/navbar.tsx
tdeekens: Do you remember why we don't put a regular `a` into the `li`?

tdeekens (commented):
💯

packages/application-shell/src/components/navbar/navbar.tsx
emmenko: We do, it's a bit nested.

tdeekens: Gotcha. Didn't expand the diff enough then :)

tdeekens (approved):

PR #2284, opened 2021-07-01
fix(constants): add order-search to graphql proxy targets

tdeekens:
#### Summary

Adds the `order-search` as a constants of the `GRAPHQL_TARGETS`. Could be a feature but I'd rather mark it as a fix cause I assume it was more forgotten than not added purposefully before.


tdeekens:
My bad. There is sadly no GraphQL on that service.

PR #2316, opened 2021-07-29
chore: update to uikit v12.2.2

emmenko:


package.json
```diff
@@ -154,6 +154,7 @@
     "jest-watch-typeahead": "0.6.4",
     "lint-staged": "11.1.1",
     "mri": "1.1.6",
+    "patch-package": "6.4.7",
```
emmenko, line 157: Let's use the babel patch here as well.

PR #2317, opened 2021-07-30
feat(config): add support for .{js,cjs,mjs,ts} file extensions

emmenko:


.eslintrc.js
```diff
@@ -1,5 +1,8 @@
 process.env.ENABLE_NEW_JSX_TRANSFORM = 'true';
 
+/**
+ * @type {import("eslint").Linter.Config}
+ */
```
emmenko, line 5: Slightly unrelated but as I discovered that we can use the JSDoc `@type` to provide type declarations for an object (and thus intellisense), I've added this in most of the config files of the repo.

packages/application-config/src/load-config.ts
```diff
@@ -1,32 +1,51 @@
...
+  // Load JS modules using Babel, as we need to load
+  // the config synchronously with `require`, no `await import`.
+  require('@babel/register')({
+    babelrc: false,
+    extensions: ['.js', '.cjs', '.mjs', '.ts'],
+    presets: ['@commercetools-frontend/babel-preset-mc-app'],
+  });
+  const result = require(filePath);
+  return get(result, 'default', result);
+};
```
emmenko, line 17: This is the custom loader, as cosmiconfig does not handle `.mjs` and `.ts`. For simplicity and consistency, I decided to use the same loader also for `.js` and `.cjs` files.

Note that we use our own babel preset, so it can handle both JS and TS files.

And we use babel because we want to load the file synchronously, otherwise if we use `import` we need to make things `async`, which is not possible for the app config.

packages/application-config/src/load-config.ts
```diff
@@ -1,32 +1,51 @@
...
   loaders: {
     noExt: defaultLoaders['.json'],
+    '.js': loadJsModule,
+    '.cjs': loadJsModule,
+    '.mjs': loadJsModule,
+    '.ts': loadJsModule,
   },
 });
 
-let cachedConfig: JSONSchemaForCustomApplicationConfigurationFiles | undefined;
```
emmenko, line 18: The cache here is not needed, as we already do this in the `process-config.ts` file.

packages/application-config/src/process-config.ts
```diff
@@ -1,5 +1,5 @@
 import type { JSONSchemaForCustomApplicationConfigurationFiles } from './schema';
-import type { ApplicationConfig, CloudIdentifier } from './types';
+import type { ApplicationRuntimeConfig, CloudIdentifier } from './types';
```
emmenko, line 2: Renamed to make it less confusing.

packages/application-config/src/types.ts
```diff
@@ -3,17 +3,13 @@ import type { JSONSchemaForCustomApplicationConfigurationFiles } from './schema'
 
 import { CLOUD_IDENTIFIERS } from './constants';
 
-export type TJSONSchemaForCustomApplicationConfigurationFiles =
-  JSONSchemaForCustomApplicationConfigurationFiles;
+export type ConfigOptions = JSONSchemaForCustomApplicationConfigurationFiles;
```
emmenko, line 6: It's now exported as a different and shorter name (see JSDoc `@type` etc.)

packages/babel-preset-mc-app/index.js
```diff
@@ -68,8 +68,6 @@ module.exports = function getBabePresetConfigForMcApp(api, opts = {}) {
           // This is probably a fine default to help trim down bundles when
           // end-users inevitably import '@babel/polyfill'.
           useBuiltIns: !options.disableCoreJs ? 'entry' : false,
-          // Do not transform modules to CJS
-          modules: false,
```
emmenko, line 72: Otherwise loading `.js` files with `import` statements fail.

I double checked with the CRA babel preset (since we base our preset from theirs) and they have also removed it since a couple of years: https://github.com/facebook/create-react-app/blob/a179b05300c24dd54349c28a94005999492c2567/packages/babel-preset-react-app/create.js#L78-L89

packages/cypress/src/task/index.ts
```diff
@@ -55,23 +55,12 @@ const loadAllCustomApplicationConfigs = async (
...
-      );
-      if (!fs.existsSync(customAppConfigPath)) {
-        return allConfigs;
-      }
-
       console.log(
         `Reading Custom Application config for ${packageInfo.packageJson.name}`
       );
-      const customAppConfigJson: TJSONSchemaForCustomApplicationConfigurationFiles =
-        JSON.parse(fs.readFileSync(customAppConfigPath, { encoding: 'utf8' }));
```
emmenko, line 70: Not needed anymore, the config now is always loaded from the app folder.

.eslintrc.js
islam3zzat: Nice!

packages/application-config/src/load-config.ts
```diff
@@ -1,32 +1,51 @@
 import type { JSONSchemaForCustomApplicationConfigurationFiles } from './schema';
+// import type { LoaderSync } from 'cosmiconfig';
 
 import { cosmiconfigSync, defaultLoaders } from 'cosmiconfig';
+// import get from 'lodash/get';
```
islam3zzat, line 5: leftover?

packages/eslint-config-mc-app/index.js
```diff
@@ -30,7 +33,7 @@ module.exports = {
   ],
   rules: {
     // NOTE: The regular rule does not support do-expressions. The equivalent rule of babel does.
-    'no-unused-expressions': 0,
+    'no-unused-expressions': statusCode.off,
```
islam3zzat, line 36: 👍

tdeekens (approved):
Very cool. Had a look on the train and it looks good to me. The JSDoc based typing for non JSON files was a nice learning! 🤗

emmenko:
Yeah, I'm trying to fix an issue with Babel register that only seems to occur on CI.

Otherwise the approach stays like this.

PS: thanks for taking a look! 😉

packages/application-config/src/load-config.ts
emmenko: Just debugging some issues on CI.

islam3zzat (approved):

pa3 (approved):

PR #2333, opened 2021-08-05
refactor(config): localized label fields of menu links is optional and defaults to empty array

emmenko:


packages/application-config/schema.json
```diff
@@ -196,30 +176,21 @@
         "labelAllLocales": {
           "description": "Localized label based on the application locales available from the user profile.",
           "type": "array",
+          "default": [],
```
emmenko, line 179: Default value

packages/application-config/schema.json
```diff
@@ -248,30 +219,21 @@
               "labelAllLocales": {
                 "description": "Localized label based on the application locales available from the user profile.",
                 "type": "array",
+                "default": [],
```
emmenko, line 222: Default value

PR #2335, opened 2021-08-05
Rewrite starter template

emmenko:
### It's time to pimp up the starter template! 💅

TL;DR of the important changes:

* The template now has more built-in dev tools, like ESLint and Prettier.
* The template README has been cleaned up a bit.
* The template uses the new [React JSX transform](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html).
* The template uses the new [menu links in application config](https://docs.commercetools.com/custom-applications/development/menu-links#menu-links-in-application-config).
* The template has a nicer landing page now (`welcome.js`).

image

* The template has a `channels` page, that fetches channels and renders them in a table. This is a good showcase of some of our best practices, for example to use Hooks, to use GraphQL, to use notifications, to render data in a table (with pagination), etc.

image

* The template has RTL tests, including mocking GraphQL and using test data.

Furthermore, the `create-mc-app` package now does the following things when installing the template:

* Removes unnecessary files such as `CHANGELOG.md`,
* Derives the `entryPointUriPath` from the folder name, then replaces it in the template in the custom app config and in the constants file.


FFawzy:
Great Work man! love the new homepage and the best practices for how to use the UiKit hooks and others.

The only drawback here would be documentation, I'm familiar with the concepts so I can work my way around but for someone who is using the starterkit for the first time it would be harder to grab everything.
but again it's nice that it is there because people might be able to grab the best practices from that exercise rather than showing an empty page.

1- the old bug that was in the installation script still persists .. if you have a space in the folder name the installation crashes (from the first step)
2- made another folder ..tried it again .. it worked until installing dependencies step .. for some reason it failed see picture… the script fails but the project was downloaded and ran as expected with yarn start

PR #2346, opened 2021-08-20
chore(README.md): add more comprehensive instructions for developing application kit bundles in readme

ByronDWall:
#### Summary
Updates readme with more detailed instructions for setting up the development environment for application kit bundles.

#### Description
Adds more detailed direction regarding initially installing this repo and setting it up to watch the app-kit bundles and the playground

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,13 +23,70 @@ $ npx @commercetools-frontend/create-mc-app my-new-custom-application-project --
 
 ## Developing application-kit packages
 
-Install the dependencies (uses yarn workspaces):
+### Initial Installation
 
-```bash
-$ yarn
-```
+1. Clone this repo
 
-To run the tests:
+   ```bash
+   $ git clone https://github.com/commercetools/merchant-center-application-kit.git
+   ```
+
+2. Add necessary environment variables
+
+   > In the vscode file tree or the terminal, navigate to `merchant_center_application_kit/playground`, open `.env.local.template`, and add these values:
+
+   ```bash
+     MC_API_URL="https://mc-api.europe-west1.gcp.commercetools.com" # for prod
+     APP_ID="" # can be an empty string for dev
+     CTP_INITIAL_PRODUCT_KEY=<your-project-name> # the name of any project you have access to on prod/stage
+   ```
+
+   > In the vscode filetree or the terminal, change `.env.local.template` filename to `.env.local`
+
+3. Build and run the application kit
+
+   > In a new terminal window, navigate to the project root directory - `/merchant_center_application_kit` - and run:
+
+   ```bash
+   $ yarn && yarn prebuild && yarn build && yarn build:watch
+   ```
+
+   > Confirm that the watch is running successfully, you should see:
+
+   ```bash
+   > "info @commercetools-frontend/application-shell waiting for changes..."
+   >...
+   ```
+
+4. Build and run the [playground application](./playground):
+
+   > In a new terminal window, navigate to the project root directory - `/merchant_center_application_kit`, and run:
+
+   ```bash
+   $ `yarn playground:build && yarn playground:start`
+   ```
+
+   > This should open a browser window and the standard merchant cernter login prompt, login using your account, and then the playground should load in the browser if permissions are set correctly (I am unclear on what those perms are or how to set them, so this is completely theoretical on my part at this point)
+
+### RUNNING THE APP KIT AND PLAYGROUND AFTER INITIAL INSTALLATION
+
+- Open 2 terminal windows
+
+  - In the first terminal run
+
+    ```bash
+      $ yarn build:watch
+    ```
+
+  - Once the watch process has completed and is listening,in the second terminalm window run
+
+         ```bash
+           $ yarn playground:start
+         ```
+
+    > The playground application consumes the app-kit dependencies as es modules, which means you need to bundle the packages first. We recommend to bundle the packages in watch mode in one terminal process and start the playground app in another terminal process.
+
+## To run the tests:
 
 ```bash
 $ yarn test
@@ -38,25 +95,13 @@ $ yarn test
 $ yarn test:watch
 ```
 
-Build the application bundles
+## To Build the application bundles:
 
 ```bash
 $ yarn build
 
 # or
 $ yarn build:bundles:watch
-```
-
-Start the [playground application](./playground):
-
-> NOTE: the playground application consumes the app-kit dependencies's es modules, which means you need to bundle the packages first. We recommend to bundle the packages in watch mode in one terminal process and start the playground app in another terminal process.
-
-```bash
-// Terminal process 1
-$ yarn build:bundles:watch
-
-// Terminal process 2
-$ yarn playground:start
 ```
 
 ## Documentation


README.md
```diff
@@ -23,13 +23,70 @@ $ npx @commercetools-frontend/create-mc-app my-new-custom-application-project --
...
-```
+1. Clone this repo
+
+   ```bash
+   $ git clone https://github.com/commercetools/merchant-center-application-kit.git
+   ```
+
+2. Add necessary environment variables
+
+   > In the vscode file tree or the terminal, navigate to `merchant_center_application_kit/playground`, open `.env.local.template`, and add these values:
```
wrsenn, line 36: This file shouldn't be modified––you'll want to duplicate it and rename to `.env.local` instead. No need to specify the method you'll use to do it, either; I'm sure there's an Emacs user in the org somewhere 😉

ByronDWall, line 36:
```suggestion
   > Navigate to `merchant_center_application_kit/playground`, duplicate `.env.local.template`, name the duplicate `.env.local` and add these values:
```

ByronDWall: good point, changed in most recent commit

README.md
```diff
@@ -23,13 +23,68 @@ $ npx @commercetools-frontend/create-mc-app my-new-custom-application-project --
...
+   ```
+
+2. Add necessary environment variables
+
+   > Navigate to `merchant_center_application_kit/playground`, duplicate `.env.local.template`, name the duplicate `.env.local` and add these values:
+
+   ```bash
+     MC_API_URL="https://mc-api.europe-west1.gcp.commercetools.com" # for prod
+     APP_ID="" # can be an empty string for dev
+     CTP_INITIAL_PRODUCT_KEY=<your-project-name> # the name of any project you have access to on prod/stage
```
mohib0306, line 41: this should be `CTP_INITIAL_PROJECT_KEY`

ByronDWall: indeed it should be, thanks!

README.md
```diff
@@ -23,40 +23,119 @@ $ npx @commercetools-frontend/create-mc-app my-new-custom-application-project --
...
+
+   ```bash
+     MC_API_URL="https://mc-api.europe-west1.gcp.commercetools.com" # for prod
+     APP_ID="" # can be an empty string for dev
+     CTP_INITIAL_PROJECT_KEY=<your-project-name> # the name of any project you have access to on prod/stage
+   ```
+
+3. Build and run the application kit
+
+   > In a new terminal window, navigate to the project root directory - `/merchant_center_application_kit` - and run:
```
emmenko, line 46:
```suggestion
   > In a new terminal window, navigate to the project root directory and run:
```

README.md
```diff
@@ -23,40 +23,119 @@ $ npx @commercetools-frontend/create-mc-app my-new-custom-application-project --
...
+
+   ```bash
+   $ yarn && yarn prebuild && yarn build && yarn build:watch
```
emmenko, line 49: This command does not make much sense. Simply run `build`.

```suggestion
   $ yarn build
```

FYI: `prebuild` is automatically executed when the `build` command is executed. Every npm script that starts with `pre<name of command>` does that.

README.md
```diff
@@ -23,40 +23,119 @@ $ npx @commercetools-frontend/create-mc-app my-new-custom-application-project --
...
 ## Developing application-kit packages
 
-Install the dependencies (uses yarn workspaces):
+### Initial Installation
+
+1. Clone this repo
+
+   ```bash
+   $ git clone https://github.com/commercetools/merchant-center-application-kit.git
+   ```
```
emmenko, line 32: I would omit this part to be honest, it does not seem relevant.

README.md
```diff
@@ -23,40 +23,119 @@ $ npx @commercetools-frontend/create-mc-app my-new-custom-application-project --
...
+   > "info @commercetools-frontend/application-shell waiting for changes..."
+   >...
+   ```
+
+4. Remove fake permission:
+   > `PERMISSIONS.ViewPlaygroundStateMachines` which is required by playground is a fake permission and you might need to remove it
+   > in order to be able to run and view the application. You'll need to remove it twice in `custom-application-config.mjs` and once in
+   > `routes.js` so that `permissions`/`demandedPermissions` are just empty objects:
+
+`custom-application-config.mjs`
```
emmenko, line 64: I'm not sure I understand this point. Why fake?

README.md
```diff
@@ -23,40 +23,119 @@ $ npx @commercetools-frontend/create-mc-app my-new-custom-application-project --
...
+          locale: 'en',
+          value: '${intl:en:Menu.EchoServer}',
+        },
+        {
+          locale: 'de',
+          value: '${intl:de:Menu.EchoServer}',
+        },
+      ],
+    },
+  ],
```
emmenko, line 84: The custom app config does not need to change. You can omit this.

README.md
```diff
@@ -23,40 +23,119 @@ $ npx @commercetools-frontend/create-mc-app my-new-custom-application-project --
...
+```
 
-# or
-$ yarn test:watch
+5. Build and run the [playground application](./playground):
+
+> In a new terminal window, navigate to the project root directory - `/merchant_center_application_kit`, and run:
+
+```bash
+$ yarn playground:build && yarn playground:start
```
emmenko, line 100: It's either/or. If you want to develop with it, you just `start` it, no need to build it.

README.md
```diff
@@ -23,40 +23,119 @@ $ npx @commercetools-frontend/create-mc-app my-new-custom-application-project --
...
 ```
 
-Build the application bundles
+> This should open a browser window and the standard merchant cernter login prompt, login using your account, and then the playground should load in the browser if permissions are set correctly (I am unclear on what those perms are or how to set them, so this is completely theoretical on my part at this point)
```
emmenko, line 103: Let's remove this.

README.md
```diff
@@ -23,40 +23,119 @@ $ npx @commercetools-frontend/create-mc-app my-new-custom-application-project --
...
+
+### RUNNING THE APP KIT AND PLAYGROUND AFTER INITIAL INSTALLATION
```
emmenko, line 105: Why uppercase?

README.md
```diff
@@ -23,40 +23,119 @@ $ npx @commercetools-frontend/create-mc-app my-new-custom-application-project --
 
 ## Developing application-kit packages
 
-Install the dependencies (uses yarn workspaces):
+### Initial Installation
```
emmenko, line 26: This all section can be removed to be honest. We can mention to check and ensure to have the `.env.local` file but I wouldn't document its content. Instead you can add more information and give hints in the template file itself.

emmenko (changes requested):
Thanks for trying to improve the docs. I find some of the suggestions confusing though and I think we only need to mention a couple of things. That should be enough.

Can you take another look? Thanks

README.md
jmcreasman: Screen Shot 2021-09-21 at 2 44 09 PM

@emmenko 
When you initially pull down the repo this is the result when you run yarn build:watch after adding the environment variables without running yarn build first.

README.md
jmcreasman: Screen Shot 2021-09-21 at 3 00 10 PM

@emmenko 
"Fake" might not be the best terminology but without removing them in those two files you cannot access the playground (see screenshot). Fedor helped me discover this and he called them fake permissions which is why I used that term but I can change it if there's a better alternative. We were curious though why they were there in the first place and if they needed to be removed permanently.

README.md
ByronDWall: Good catch, changed in latest commit

README.md
ByronDWall: removed

README.md
ByronDWall: In our initial attempt starting the playground, we were under the impression that it was necessary to run the build step once before the watch step, or the watch would fail out.  With further investigation, apparently we were incorrect, the build step has been removed.

emmenko:
@jmcreasman @ByronDWall I see that there is still some confusion about this. Let's maybe have a 1:1 chat and I'll explain a bit more how things are supposed to work.

ByronDWall:
> I see that there is still some confusion about this. Let's maybe have a 1:1 chat and I'll explain a bit more how things are supposed to work.

@emmenko  I think a 1:1 chat would be helpful, thanks for taking the time to look things over and review

PR #2358, opened 2021-09-10
PAN-1031/DCY-123: Menu item and submenu of currently loaded application active in navbar

### LINKED ISSUE #2352: PAN-1031/DCY-123: Navbar menu item for current application set to active and displays submenus on initial mount

[PAN-1031](https://jira.commercetools.com/browse/PANGOLIN-1031) , a clone of [DCY-123](https://jira.commercetools.com/browse/DCY-123), describes implementing showing the currently viewed application's navbar menu item and submenus in an active state on initial load of the application.

I am currently exploring implementing a workflow that checks the URL pathname for the potential name of the active 'application' (i.e. dashboard, products, etc), and sets the navbar menu item to active and displays submenus if the application name in the url matches a given menu's key.

**Additional context**

Currently when an application is loaded the corresponding nav menu item and submenus are not active/visible on initial mount - as shown by this screenshot of the app-kit playground on `main`:
Screen Shot 2021-08-31 at 1 32 46 PM

This feature would make it such that when an application is loaded the corresponding nav menu item and submenus are active/visible on initial mount - as shown by this screenshot of the app-kit playground on `PAN-1031-left-menu-expansion`:
Screen Shot 2021-08-31 at 1 32 03 PM

ByronDWall:
Summary

Issue: When a user opens a link from another MC App, the left menu submenu is not expanded upon initial load. From a users perspective though, there should be no difference in left menu behavior compared to when navigating between pages within an app (where the submenu remains open)

Suggestion: By default, have the submenu open when an app is opened so it works similar to when navigating between pages inside an app.
Description

This PR uses the isMainMenuRouteActive method in ApplicationMenu component in navbar.tsx to ascertain whether the currently loaded route matches the menu item, and sets the menu item to active and the submenus to be visible if the current route matches the menu item's url.

Tests

- [x] unit tests 
- [x] integration tests

Browser Compat

- [x] - firefox
- [x] - chrome
- [x] - safari

closes #2352

### DIFF
--- a/cypress/integration/template-starter/channels.js
+++ b/cypress/integration/template-starter/channels.js
@@ -2,6 +2,7 @@ import { URL_EXAMPLES_STARTER_CHANNELS } from '../../support/urls';
 
 describe('Channels', () => {
   it('should render page', () => {
+    cy.viewport(1024, 1024);
     cy.login({ redirectToUri: URL_EXAMPLES_STARTER_CHANNELS });
     cy.findByText('Channels list').should('exist');
     cy.findByText('Store Berlin').should('exist');
--- a/cypress/integration/template-starter/channels.js
+++ b/cypress/integration/template-starter/channels.js
@@ -4,7 +4,6 @@ describe('Channels', () => {
   it('should render page', () => {
     cy.login({
       redirectToUri: URL_EXAMPLES_STARTER_CHANNELS,
-      isForcedMenuOpen: false,
     });
     cy.findByText('Channels list').should('exist');
     cy.findByText('Store Berlin').should('exist');
--- a/cypress/integration/template-starter/channels.js
+++ b/cypress/integration/template-starter/channels.js
@@ -2,9 +2,10 @@ import { URL_EXAMPLES_STARTER_CHANNELS } from '../../support/urls';
 
 describe('Channels', () => {
   it('should render page', () => {
-    cy.setDesktopViewport();
-    cy.viewport(2560, 1440);
-    cy.login({ redirectToUri: URL_EXAMPLES_STARTER_CHANNELS });
+    cy.login({
+      redirectToUri: URL_EXAMPLES_STARTER_CHANNELS,
+      isForcedMenuOpen: false,
+    });
 
     cy.findByText('Channels list').should('exist');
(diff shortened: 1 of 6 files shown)


cypress/integration/template-starter/channels.js
```diff
@@ -2,7 +2,9 @@ import { URL_EXAMPLES_STARTER_CHANNELS } from '../../support/urls';
 
 describe('Channels', () => {
   it('should render page', () => {
-    cy.login({ redirectToUri: URL_EXAMPLES_STARTER_CHANNELS });
+    cy.login({
+      redirectToUri: URL_EXAMPLES_STARTER_CHANNELS,
+    });
```
ByronDWall, line 7: The e2e test output has changed, but the output as it now stands (expanding the menu and activating the current menu item if the screen width is greater than 1024px) has been approved by UX, so I believe the chages in these two tests should be accepted.

jmcreasman (approved):
Looks good and makes sense to me!

emmenko (changes requested):
Thanks for looking into it. I think I understand the point and I would agree with that.

However, I noticed from the VRT that the navbar opens when it's supposed to be collapsed. We should respect that, and expand the submenu only when the navbar is in an expanded state. Can you check that please?

Also FYI, we do have a "task" to refactor the navbar at some point with an improved UX (#1884 ).

ByronDWall:
> However, I noticed from the VRT that the navbar opens when it's supposed to be collapsed. We should respect that, and expand the submenu only when the navbar is in an expanded state. Can you check that please?

I was unclear whether it would be best to force it into an expanded state or keep it closed based on the UX spec.  Thanks for the clarification, I will update that shortly.

emmenko (approved):
@ByronDWall I pushed a bunch of commits to clean things up a bit. From my side this should be good to go now.
Thanks again for the pushing this 👌

ByronDWall:
> @ByronDWall I pushed a bunch of commits to clean things up a bit. From my side this should be good to go now. Thanks again for the pushing this 👌

Thank you for the cleanup, should I go through and approve the e2e test changes, or leave that to you?

PR #2359, opened 2021-09-13
chore(eslint-config-mc-app deps): fix versions

ahmehri:
#### Summary

Use exact versions instead of range versions.

#### Description

Using range versions is not ideal, as it will cause updating versions without passing by Renovate. This update can even occur as a [side effect of painting the lock file](https://github.com/commercetools/merchant-center-frontend/pull/11133#issuecomment-918067352). In order to avoid this, versions have been fixed to get aligned with how it's being done with dependencies in other packages.


tdeekens (approved):
Thanks. Makes sense to me. Can' recall why we used ranges here. Given packages can leak in breaking changes in minor of fix releases this can be dangerous.

emmenko (changes requested):
Thanks for looking into it. The caret `^` version was actually done on purpose, to avoid possible conflicts with eslint packages installed by other packages.

Can't we just use a version resolution to fix the issue on our side?

ahmehri:
> Thanks for looking into it. The caret `^` version was actually done on purpose, to avoid possible conflicts with eslint packages installed by other packages.
> 
> Can't we just use a version resolution to fix the issue on our side?

I think it needs to be done the other way around. The implemented solution to solve the issue of the eslint conflicts is causing another issue as a side effect in the consumer land. This side effect is so bad and unpredictable, it's not logical at all that maintaining the lock file will result in dependencies update.

What needs to be done instead, in my opinion, is to fix the versions so that the side effect is gone, and fix properly the issue of the eslint conflicts without introducing any unwanted side effects, maybe using the proposed resolution fix here instead?

ahmehri:
After having a discussion with @emmenko, the use case is now clear. The dependencies with the range versions used to be defined as peer dependency, then the goal was to make it easier to use the app-kit by consumers by being able to use whatever versions of the dependencies they wish to use without having to wait for a new release of app-kit. In order to achieve this goal, the dependencies are now defined as normal dependencies without pinning the versions, to also avoid possible conflicts as @emmenko explained.

Therefore, there is no more need for the changes in this PR and they can be discarded, closing the issue.

PR #2364, opened 2021-09-29
PCM-2389: Overlapping portal container and notification bar

jaikumar-tj:
#### Summary

The error notification overlaps with the modal container but the components in the background adjust itself whenever notification appears.

### Fix

Moving the Portal container inside the same common parent div fixes the problem of overlapping.

#### Before

Screenshot 2021-09-29 at 10 04 12

#### After

Screenshot 2021-09-29 at 10 03 39

### DIFF
--- a/.changeset/dry-cherries-whisper.md
+++ b/.changeset/dry-cherries-whisper.md
@@ -1,5 +1,5 @@
 ---
-'@commercetools-frontend/application-shell': minor
+'@commercetools-frontend/application-shell': patch
 ---
 
-Change the DOM structure for portal container
+Page notifications (like errors) now correctly push the page content below them, even for modal pages.
--- a/.changeset/dry-cherries-whisper.md
+++ b/.changeset/dry-cherries-whisper.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': minor
+---
+
+Change the DOM structure for portal container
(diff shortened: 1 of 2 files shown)


jaikumar-tj:
@emmenko This is my first PR in application-kit and I'm curious if it is ok to change the root level DOM structure? Also let me know if I need to do anything else apart from changeset.

.changeset/dry-cherries-whisper.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': minor
```
emmenko, line 2: This looks like a fix, so we should use `patch`.

```suggestion
'@commercetools-frontend/application-shell': patch
```

.changeset/dry-cherries-whisper.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Change the DOM structure for portal container
```
emmenko, line 5: Note that the information we write here should ideally help users that read the changelog to understand what changed.

In this case, this sentence essentially tells what technically changed but without any further context it's hard to make a sense of it.

Instead, we can mention what kind of issue we fixed. For example:

```suggestion
Page notifications (like errors) now correctly push the page content below them, even for modal pages.
```

emmenko (commented):
Thanks, I like the solution: nice and simple. I can't think of any undesired side-effects at the moment, so I'm fine going ahead with that.

.changeset/dry-cherries-whisper.md
jaikumar-tj: Thanks for the explanation😊.

emmenko (approved):

pa3 (approved):
:+1:

PR #2387, opened 2021-11-05
refactor(application-components): add optional prop to modal page transform property

fuchodeveloper:
#### Summary

Adds an optional prop (`afterOpenStyles`) to `ModalPage` component
#### Description

The `<*ModalPage>` components now accept a prop `afterOpenStyles` to overwrite the default styles. You can pass a "class name" or a CSS-in-JS style object.

This fixes a `jumpy` behavior observed when attempting to drag-n-drop an element on a modal page.


packages/application-components/src/components/modal-pages/internals/modal-page.tsx
```diff
@@ -59,6 +59,7 @@ type Props = {
   baseZIndex: number;
   getParentSelector: typeof getDefaultParentSelector;
   shouldDelayOnClose: boolean;
+  afterOpenTransform?: string;
```
emmenko, line 62: I would suggest to use a more generic prop, something like `afterOpenStyles`.

So you can pass either a class name or a style object. Then in the `afterOpen` prop you do:

```js
afterOpen: typeof props.afterOpenStyles === 'string'
  ? props.afterOpenStyles
  : makeClassName(props.afterOpenStyles ?? getAfterOpenContainerAnimation())
```

The point here is that if we make the "after open" styles customizable, we should make the style itself customizable and not only a single css value.

emmenko (changes requested):
Thanks for the contribution.

I left a suggestion to make this a bit more customizable.

PS: also please provide a changeset.

packages/application-components/src/components/modal-pages/internals/modal-page.tsx
```diff
@@ -59,6 +59,7 @@ type Props = {
   baseZIndex: number;
   getParentSelector: typeof getDefaultParentSelector;
   shouldDelayOnClose: boolean;
+  afterOpenStyles?: string | { [key: string]: string | number };
```
emmenko, line 62:
```suggestion
  afterOpenStyles?: string | CSSProperties;
```

fuchodeveloper: Using `CSSProperties` returns the following error on line 125:
```
 Overload 1 of 2, '(template: TemplateStringsArray, ...args: CSSInterpolation[]): string', gave the following error.
    Argument of type 'CSSProperties | SerializedStyles' is not assignable to parameter of type 'TemplateStringsArray'.
      Type 'PropertiesFallback<string | number, string & {}>' is missing the following properties from type 'TemplateStringsArray': raw, length, concat, join, and 18 more.
  Overload 2 of 2, '(...args: CSSInterpolation[]): string', gave the following error.
    Argument of type 'CSSProperties | SerializedStyles' is not assignable to parameter of type 'CSSInterpolation'.
      Type 'CSSProperties' is not assignable to type 'CSSInterpolation'.
        Type 'PropertiesFallback<string | number, string & {}>' is not assignable to type 'CSSObject'.
```

Using `CSSObject` like so works for typing the prop:
```
import { CSSObject } from '@emotion/react';

...
afterOpenStyles?: string | CSSObject;
```

emmenko: Ah yes, thanks!

packages/application-components/src/components/modal-pages/tabular-modal-page/tabular-modal-page.mdx
```diff
@@ -44,6 +44,7 @@ import { TabularModalPage } from '@commercetools-frontend/application-components
...
 | `shouldDelayOnClose`      | `bool`             |    -     | `true`                   | Sets whether the ModalPage should delay calling its `onClose` function to allow the closing animation time to finish. This can be turned off if the developer is controlling the ModalPage only through the `isOpen` prop, and not abruptly mounting/unmounting it or one of its parent elements. You might also want to turn this off if you need to display a Prompt (for example to save changes) on the ModalPage before navigating out of it, as this option makes the Modal close itself before `onClose` is called. |
+| `afterOpenStyles`               | `string` or `object`       |    -     | -                        | The CSS styles applied to `afterOpen` class after the modal has been opened.                                                                                                                                                                                                                                                                                                                                                                                                                                   |
```
emmenko, line 47:
```suggestion
| `afterOpenStyles`               | `string` or `object`       |    -     | -                        | Overwrite the default styles of `afterOpen`. You can pass a "class name" or a CSS-in-JS style object. This should be used only in cases the default styles are causing some layout issues.                                                                                                                                                                                                                                                                                                                                                                                                                                   |
```

emmenko (commented):
We should add this prop to all modal pages, to be consistent.

emmenko (approved):
Awesome, thanks!!

packages/application-components/src/components/modal-pages/custom-form-modal-page/custom-form-modal-page.tsx
```diff
@@ -1,4 +1,5 @@
 import { SyntheticEvent, ReactNode, ReactElement } from 'react';
+import { CSSObject } from '@emotion/react';
```
Rhotimee, line 2:
```suggestion
import type { CSSObject } from '@emotion/react';
```

Rhotimee (commented):
A minor feedback

emmenko (approved):

Rhotimee (approved):

PR #2395, opened 2021-11-15
refactor(jest): use babel-jest instead of ts-jest

emmenko:
It seems that using `ts-jest` causes issues with Emotion whenever we use [selectors as styled components](https://emotion.sh/docs/styled#targeting-another-emotion-component).

The error we get is

```
Component selectors can only be used in conjunction with @emotion/babel-plugin
```

Apparently we can just "go back" to use `babel-jest`.


Rhotimee (approved):
Looks good.

PR #2397, opened 2021-11-17
fix(navbar): passing defaultLabel to submenu

emmenko:


packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -541,6 +548,7 @@ const ApplicationMenu = (props: ApplicationMenuProps) => {
                         >
                           <MenuLabel
                             labelAllLocales={submenu.labelAllLocales}
+                            defaultLabel={submenu.defaultLabel}
```
emmenko, line 551: This is the fix

Rhotimee (approved):

PR #2404, opened 2021-11-23
[experiment] Vercel monorepo

emmenko:


emmenko:
**Outcome of the experiment**

Vercel multi projects (monorepo) works with the following caveats:

* Yarn install is run in the subfolder specified in each project settings. To make the `postinstall` hook work correctly, we need to patch the install command like `yarn --cwd .. --immutable`, to run it from the repository root.
* Vercel caches framework specific folders, for example for Gatsby is `.cache` and `public`. However, the Yarn `node_modules` from the repository root is not cached, causing each build to re-fetch and link all packages. I haven't found an option to configure it.
* [Getting preview URLs of sibling projects is not supported at the moment](https://github.com/vercel/vercel/discussions/5231). There is hacky workaround but it's not a 100% guarantee.

Overall I don't think it's worth the effort to have multiple Vercel projects. The main reason for doing that would be to improve build time on CI to have things running in parallel.
However, concurrent builds are still limited and the dependencies installation time is not an improvements due to the build cache issue.

We could in theory move the build part into github actions and deploy to Vercel using the CLI. Main advantage here would be shorter dependencies installation time. We would then need to implement some extra logic to provide PR status checks and preview link.

So for now we won't change anything of the current setup.

PR #2415, opened 2021-11-30
fix(api-errors): migrate to `error.extensions.code` for notifications

### LINKED ISSUE #168: Support extension code for graphql errors
A recent [spec change](https://github.com/facebook/graphql/pull/407) introduced an optional field `extensions` which is used by the graphql server to pass e.g. an error `code`.

The standard GraphQL error fields include `message`, `path` and `locations`. They are still available directly in the error object. Custom fields, like `code`, now can be found under `extensions` field.

We need to support this (our CTP API currently supports this as well, as well as our Settings API).

Our API errors mapper should handle this case and try to handle it first by looking for `extensions.code`, otherwise falling back to `code` (which is what our CTP REST API uses).

Example of graphql error from our CTP graphql API:

```js
{
  "data": null,
  "errors": [
    {
      "message": "'password' should not be empty.",
      "path": ["customerSignUp"],
      "locations": [{"line": 2, "column": 3}],
      "extensions": {
        "code": "InvalidOperation"
      }
    }
  ]
}
```

Example of graphql error from our Settings API:

```js
{
  "errors": [
    {
      "message": "Some graphql error message",
      "locations": [{ "line": 1, "column": 1 }],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR"
      }
    }
  ]
}

```

jaikamat:
Closes #168 

#### Summary

This PR extends support for the the GraphQL API error object shape, `error.extensions.code`.

#### Description

While migrating Discounts, I noticed that migrating all test cases did not pass cleanly. @ByronDWall and I discovered that the `useShowApiErrorNotification` hook's return value created an action (that we pass `errors` to) that `<ApplicationShell />`'s Redux context consumes, and it is here where `error.code` is expected.

This PR changes the `TAppNotificationApiError` type and all tests that rely on it.

#### Considerations
I'm unsure if there are backwards compatibility issues / side effects for other projects given the changes here - should we extend the type as needed as a safeguard?

### DIFF
--- a/packages/constants/src/constants.ts
+++ b/packages/constants/src/constants.ts
@@ -66,6 +66,8 @@ export type TAppNotificationOfKind<T extends TAppNotificationOfDomain> =
   };
 export type TAppNotificationApiError<ExtraFields extends {} = {}> = {
   message: string;
+  /** @deprecated Use `extensions.code` */
+  code?: string;
   extensions?: {
     code?: string;
   };
--- a/packages/constants/src/constants.ts
+++ b/packages/constants/src/constants.ts
@@ -66,7 +66,9 @@ export type TAppNotificationOfKind<T extends TAppNotificationOfDomain> =
   };
 export type TAppNotificationApiError<ExtraFields extends {} = {}> = {
   message: string;
-  code?: string;
+  extensions?: {
+    code?: string;
+  };
 } & ExtraFields;
 export type TAppNotificationValuesApiError<ExtraFields extends {} = {}> = {
   errors: TAppNotificationApiError<ExtraFields>[];
--- a/.changeset/twelve-rabbits-look.md
+++ b/.changeset/twelve-rabbits-look.md
@@ -1,0 +1,6 @@
+---
+'@commercetools-frontend/constants': patch
+'@commercetools-frontend/react-notifications': patch
+---
+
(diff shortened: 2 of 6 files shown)


packages/constants/src/constants.ts
```diff
@@ -66,7 +66,9 @@ export type TAppNotificationOfKind<T extends TAppNotificationOfDomain> =
   };
 export type TAppNotificationApiError<ExtraFields extends {} = {}> = {
   message: string;
-  code?: string;
+  extensions?: {
+    code?: string;
+  };
```
jaikamat, line 71: should we optionally supply `code?: string` for backwards compatibility, and handle it downstream in the function with something like `const extensionErrorCode = props.error.code || props.error.extensions?.code;`?

emmenko: Yes we should support both for backwards compatibility. Can you update the PR with that change? Thanks!

emmenko: I have some time now, I can do that.

emmenko (approved):

emmenko (approved):

packages/constants/src/constants.ts
jaikamat: Thank you!

Rhotimee (approved):

PR #2422, opened 2021-12-06
refactor(app-shell): redirect to `/login` and `/logout` using a full origin URL

### LINKED ISSUE #2406: Render "fallback" page automatically in case the application is not served from MC domain

Custom Applications are only meant to be working within the Merchant Center domain, mostly for security reasons.

For instance, if you have your app deployed somewhere and you try to access the application directly, the app won't render. Actually first the app tries to "redirect" you to the login page. The login page though is assumed to be in the same domain, so the app reloads itself causing the request to theoretically go through the Merchant Center Proxy router that handles the login request.

To avoid this "infinite" redirect loop, we recommend to handle this route in your hosting provider and simply return a static page with a text similar to this:

```
This is not a real route. If you are seeing this, you most likely are accessing the custom application
directly from the hosted domain. Instead, you need to access the custom application from within the Merchant Center
domain, as custom applications are served behind a proxy router.
To do so, you need to first register the custom application in Merchant Center > Settings > Custom Applications.
```

That's fine, but we could do better.

Instead of leaving this logic to the user we could try to handle this case within the Application Shell as a built-in behavior.
Therefore, users don't need to worry about optionally handle this behavior, making the hosting part even simpler.

For instance, the ApplicationShell can try to handle the login route and return a nicer message in case the origin URL does not match a MC-like domain (and the app is configured to run behind the MC proxy).

emmenko:
Closes #2406 

Currently when you try to access a Custom Application directly to its URL, the application tries to reload itself to go to the `/login` route.

This ends up in an "infinite" loop as the Custom Application is not being served within the Merchant Center Proxy Router and thus there is no way to handle the `/login` request.

One way to avoid this is to manually configure a route handler for `/login` (and `/logout`) in your cloud provider and for example render a message.

To help simplifying the deployment setup, we can redirect to the `/login` and `/logout` pages using a full origin URL redirect.
For Custom Applications, this means that if you try to access the application directly, it will redirect to the login page of the configured MC environment, so there is no need to handle the routes to avoid the infinite loop.

### DIFF
--- a/packages/application-shell/src/components/application-shell/redirect-to-login.tsx
+++ b/packages/application-shell/src/components/application-shell/redirect-to-login.tsx
@@ -3,6 +3,7 @@ import type { AuthorizeSessionState } from '../authenticated/types';
 
 import { v4 as uuidv4 } from 'uuid';
 import { useLocation } from 'react-router-dom';
+import { useApplicationContext } from '@commercetools-frontend/application-shell-connectors';
 import {
   joinPaths,
   trimLeadingAndTrailingSlashes,
@@ -12,6 +13,7 @@ import { buildOidcScope } from '../authenticated/helpers';
 import { OIDC_RESPONSE_TYPES } from '../../constants';
 import * as oidcStorage from '../../utils/oidc-storage';
 import Redirector from '../redirector';
+import { getMcOrigin } from './helpers';
 
 declare let window: ApplicationWindow;
 
@@ -29,6 +31,9 @@ const generateAndCacheNonceWithState = (state: AuthorizeSessionState) => {
 
 const RedirectToLogin = () => {
   const location = useLocation();
+  const servedByProxy = useApplicationContext(
+    (context) => context.environment.servedByProxy
+  );
 
   if (window.app.__DEVELOPMENT__?.oidc?.authorizeUrl) {
     // We pick the project key from local storage. This assumes that the value
@@ -81,9 +86,13 @@ const RedirectToLogin = () => {
       />
     );
   }
+
+  const mcOrigin = servedByProxy ? getMcOrigin(window.app.mcApiUrl) : undefined;
+
   return (
     <Redirector
       to="login"
+      origin={mcOrigin}
       location={location}
       queryParams={{
         reason: LOGOUT_REASONS.UNAUTHORIZED,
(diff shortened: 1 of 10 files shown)


packages/application-shell/src/components/application-shell/redirect-to-login.tsx
```diff
@@ -29,6 +31,9 @@ const generateAndCacheNonceWithState = (state: AuthorizeSessionState) => {
 
 const RedirectToLogin = () => {
   const location = useLocation();
+  const servedByProxy = useApplicationContext(
```
tdeekens, line 34: Nit: we use this pattern four times already with this being the fifth. Maybe we can extract it now into a `useIsServerdByProxy` (in this or a follow up PR). As it's quite impactful logic resulting from this it might be good not to duplicate this too much even if it's simple.

packages/application-shell/src/components/application-shell/redirect-to-login.tsx
```diff
@@ -81,9 +86,13 @@ const RedirectToLogin = () => {
       />
     );
   }
+
+  const mcOrigin = servedByProxy ? getMcOrigin(window.app.mcApiUrl) : undefined;
```
tdeekens, line 90: I would consider moving this into a hook as `useMcOrigin` to be able to memoize the computation a bit using a `useMemo`. We're now creating URL objects on each render. As the resulting value is primitive if will have a stable reference but I wonder if it could make sense to avoid obviously needless computations here as the `window.app.mcApiUrl` won't change.

tdeekens: To elaborate:

```js
function useMcOrigin() {
   const isServedByProxy = useIsServerdByProxy();

   return useMemo(() => servedByProxy ? getMcOrigin(window.app.mcApiUrl) : undefined, [window.app.mcApiUrl, servedByProxy]);
}
```

I like keeping computation in a small getter as it's easier to test and allow logic to be lifted into different contexts.

tdeekens (commented):
Quick review as I was pinged :) That's a nice improvement! Only suggestions but nothing standing in the way.

Rhotimee (approved):
Looks good.

packages/application-shell/src/components/application-shell/redirect-to-login.tsx
tdeekens: Ok. Sounds good as long as we don't create 100s or URL objects.

tdeekens (approved):
Cool. Please consider a `useIsServedByProxy` extraction as a followup. Should be a good morning warm up task :)

PR #2425, opened 2021-12-09
refactor(application-shell): to useFlagVariation over ToggleFeature

tdeekens:
#### Summary

The application-shell supports passing a `featureToggle` as a configuration option on each menu item. A functionality only used internally to hide certain applications under certain conditions.

In this change we replace `ToggleFeature` with `useFlagVariation` to allow using structurally complex flag values such as `{ value: boolean, reason?: string }` which are internally used as long-running feature configurations.

#### Description

Prior to this change the resulting flag value corresponding to the passed `props.featureToggle` would have to be a `boolean`. Feature flags however can be more complex in structure. In fact in LaunchDarkly they could be a JSON value. In our particular case they're of the shape `{ value: boolean, reason?: string }`. If we use `ToggleFeature` to resolve the flag value we will always get a `boolean` and render a component as a result even if the flag is of the previously described shape.

By replacing `ToggleFeature` with `useFlagVariation` we can use long-running flags and evaluate them too for our menu configuration. As an example we could hide a menu item based on `{ enableSettings: { value: false, reason: 'not-installed' } }`.


.yarnrc.yml
```diff
@@ -5,7 +5,9 @@ plugins:
     spec: "@yarnpkg/plugin-constraints"
   - path: .yarn/plugins/@yarnpkg/plugin-postinstall.cjs
     spec: "https://raw.githubusercontent.com/gravitywelluk/yarn-plugin-postinstall/master/bundles/%40yarnpkg/plugin-postinstall.js"
-
-yarnPath: .yarn/releases/yarn-3.1.1.cjs
+  - path: .yarn/plugins/@yarnpkg/plugin-interactive-tools.cjs
```
tdeekens, line 8: Could be dropped. I find this tool very nice to selectively update dependencies interactively just like with `yarn@1`.

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -322,15 +318,25 @@ type RestrictedMenuItemProps = {
   dataFences?: TDataFence[];
   children: ReactNode;
 };
+type TLongLivedFlag = {
```
tdeekens, line 321: We defined our internal flag structure here was we don't yet need a shared type.

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -322,15 +318,25 @@ type RestrictedMenuItemProps = {
...
+  value: boolean;
+  reason?: string;
+};
 const restrictedMenuItemDefaultProps: Pick<
   RestrictedMenuItemProps,
   'permissions'
 > = {
   permissions: [],
 };
+function isLongLivedFlag(
```
tdeekens, line 331: As flopflip can't know what a service such as LaunchDarkly returns and as a result is exposed to an application it feels safest to type guard it instead of using generics only.

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -322,15 +318,25 @@ type RestrictedMenuItemProps = {
...
+  flag: TFlagVariation | TLongLivedFlag
+): flag is TLongLivedFlag {
+  return typeof (flag as TLongLivedFlag)?.value === 'boolean';
+}
 const RestrictedMenuItem = (props: RestrictedMenuItemProps) => {
   // NOTE: Custom application are activated/deactivated while their
   // visibility is not controlled via a visibiility overwrite.
+  const flagVariation = useFlagVariation(props.featureToggle);
```
tdeekens, line 339: Note, that with `useFlagVariation(undefined)` as `props.featureToggle` not being defined `@flopflip` will return `false` which in turn on L374 we don't use. A culprit to not being able to use hooks conditionally as we know.

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -365,12 +371,15 @@ const RestrictedMenuItem = (props: RestrictedMenuItemProps) => {
     );
 
   if (props.featureToggle) {
-    return (
-      <ToggleFeature flag={props.featureToggle}>
-        {permissionsWrapper}
-      </ToggleFeature>
-    );
+    // A regular short-lived feature toggle
+    if (flagVariation === true) return permissionsWrapper;
```
tdeekens, line 375: So either it's a boolean then we can use it. Could also be another type guard or more sophisticated check but this should do.

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -365,12 +371,15 @@ const RestrictedMenuItem = (props: RestrictedMenuItemProps) => {
...
+    // A long-lived feature toggle with `{ value: boolean, string: reason }`
+    if (isLongLivedFlag(flagVariation) && flagVariation.value === true)
```
tdeekens, line 377: Here we check the long-lived flag value.

tdeekens (commented):
This correlates to an internal change to which will reference this from there.

.yarnrc.yml
emmenko: Sure, no problem. Thanks.

emmenko (approved):
Top 💯 , Danke!

PR #2430, opened 2021-12-15
[Breaking] Changes for upcoming v21 release

emmenko:
TODO:
- [ ] ~~(Optional) Upgrade CLI to CRA v5 (https://github.com/facebook/create-react-app/releases/tag/v5.0.0)~~

  NOTE: Updating to CRA v5 does not seem so straight-forward. After the first try, I noticed that HMR (react refresh) does not seem to work properly. However, when trying to use the default setup of CRA, it works fine. Something with our webpack and babel config seems to "get in the way". I think we can look into this separately. Ideally, we would start to use as much as possible of the default setup with our customizations on top.


.changeset/hungry-bobcats-relax.md
```diff
@@ -0,0 +1,15 @@
...
+'@commercetools-frontend/create-mc-app': major
+'@commercetools-frontend/cypress': major
+'@commercetools-frontend/mc-dev-authentication': major
+'@commercetools-frontend/mc-html-template': major
+'@commercetools-frontend/mc-scripts': major
+'playground': major
+'@commercetools-local/visual-testing-app': major
+---
+
+Drop Node.js `v12`. Recommended min Node.js version is `v14` or `v16`.
```
tdeekens, line 15:
```suggestion
Drop Node.js `v12`. Recommended Node.js version are `v14` or `v16`.
```

.changeset/many-tables-appear.md
```diff
@@ -0,0 +1,18 @@
...
+'@commercetools-frontend/mc-html-template': major
+'@commercetools-frontend/mc-scripts': major
+'playground': major
+'@commercetools-local/visual-testing-app': major
+---
+
+Following breaking changes were introduced:
+
+- In `mc-scripts`, the `build` command additionally compiles the `index.html` by default.
+- Running the `compile-html` command by default should not be necessary anymore. However, you can pass `--build-only` to the `build` command to opt-out of the compilation step, in case you want to run it separately, for example to use the `--transformer`.
```
tdeekens, line 13: Maybe me but I rad this four times until I got it 😄 

```suggestion
- Running `build` now includes compiling the HTML. As a result running the `compile-html` command should not be necessary anymore. However, if you do not need HTML to be compiled with the `build` command pass the `--build-only` to opt-out of the compilation step. A use case for this is when using a `--transformer` to run compilation separately.
```

.changeset/many-tables-appear.md
```diff
@@ -0,0 +1,18 @@
...
+- Running the `compile-html` command by default does not print to `stdout` the JSON string with the security headers. You can opt into the old behavior by passing the `--print-security-headers` option.
+- The `--inline-csp` of `compile-html` has been dropped, as it's now the built-in behavior.
+- The `dist` folder created by the `build` command has been removed. Instead, the `build` command writes the production bundles directly into the `public` folder.
```
tdeekens, line 16:
```suggestion
- The `dist` folder created by the `build` command has been removed. Instead, the `build` command writes the production bundles directly into the `public` folder. If any tool relies on picking up the production bundles from `dist` make sure to reconfigure it to rely on the `public` folder.
```

application-templates/starter/custom-application-config.mjs
```diff
@@ -0,0 +1,39 @@
+import { PERMISSIONS, entryPointUriPath } from './src/constants';
+
+/**
+ * @type {import('@commercetools-frontend/application-config').ConfigOptions}
```
tdeekens, line 4: 💯 Nice!

application-templates/starter/package.json
```diff
@@ -7,8 +7,8 @@
     "postinstall": "manypkg check",
     "build": "mc-scripts build",
     "start": "mc-scripts start",
-    "compile-html": "NODE_ENV=production mc-scripts compile-html",
-    "compile-html:local": "NODE_ENV=production MC_APP_ENV=development mc-scripts compile-html --transformer @commercetools-frontend/mc-dev-authentication/transformer-local.js",
+    "compile-html": "mc-scripts compile-html",
+    "compile-html:local": "MC_APP_ENV=development mc-scripts compile-html --transformer @commercetools-frontend/mc-dev-authentication/transformer-local.js",
```
tdeekens, line 11: Making sure I follow along with the new `build` behaviour: why don't we `--build-only` here?

packages/application-shell/src/components/application-shell/application-shell.tsx
```diff
@@ -540,12 +530,6 @@ const ApplicationShell = <AdditionalEnvironmentProperties extends {}>(
                     render={props.render}
                     applicationMessages={props.applicationMessages}
                     onMenuItemClick={props.onMenuItemClick}
-                    DEV_ONLY__loadAppbarMenuConfig={
-                      props.DEV_ONLY__loadAppbarMenuConfig
-                    }
-                    DEV_ONLY__loadNavbarMenuConfig={
-                      props.DEV_ONLY__loadNavbarMenuConfig
-                    }
```
tdeekens, line 548: 👍🏼

packages/application-shell/src/test-utils/test-utils.tsx
```diff
@@ -220,12 +223,12 @@ const wrapIfNeeded = (
 type TApolloProviderWrapperProps = {
   apolloClient?: ApolloClient<NormalizedCacheObject>;
   mocks: ReadonlyArray<MockedResponse>;
-  disableApolloMocks: boolean;
+  enableApolloMocks: boolean;
```
tdeekens, line 226: I thought this was called `useApolloMocks`. At lease the change log stated so.

.changeset/many-tables-appear.md
emmenko: Thanks :)

application-templates/starter/package.json
emmenko: You can always run `compile-html` after `build`, even if you didn't use `--build-only`. What happens is that the new `compile-html` overrides the previous version.

packages/application-shell/src/test-utils/test-utils.tsx
emmenko: Ah thanks, it should be `enableApolloMocks`.

I find `useApolloMocks` to be confusing with some kind of react hook...

.changeset/tall-dolls-study.md
```diff
@@ -0,0 +1,145 @@
+---
+'merchant-center-application-template-starter': major
+'@commercetools-frontend/application-shell': major
+'playground': major
+---
+
+Following breaking changes were introduced in `test-utils`:
+
+- The deprecated `project.allAppliedMenuVisibilities` option has been removed.
+- The `disableApolloMocks` option has been renamed to `useApolloMocks`. The behavior is now inverted. By default, the Apollo mocks are disabled. This is to encourage mocking via [MSW](https://mswjs.io/).
```
emmenko, line 10:
```suggestion
- The `disableApolloMocks` option has been renamed to `enableApolloMocks`. The behavior is now inverted. By default, the Apollo mocks are disabled. This is to encourage mocking via [MSW](https://mswjs.io/).
```

packages/application-shell/src/test-utils/test-utils.tsx
tdeekens: Yeap, I also disable the `use*` toggle naming.

.github/workflows/main.yml
```diff
@@ -118,6 +118,8 @@ jobs:
 
       - name: Building Starter template application
         run: yarn template-starter:build
+        env:
+          CTP_INITIAL_PROJECT_KEY: ${{ secrets. CYPRESS_PROJECT_KEY }}
```
tdeekens, line 122:
```suggestion
          CTP_INITIAL_PROJECT_KEY: ${{ secrets.CYPRESS_PROJECT_KEY }}
```

emmenko:
This is now ready for final review/approval.

emmenko:
FYI: I think I'll include stylelint v14 as well in this release, as there might be some breaking changes on that side as well.

tdeekens (approved):

PR #2440, opened 2021-12-22
fix(app-shell): container styles for locale switcher

emmenko:
Before
image

After
image


Rhotimee (approved):

emmenko:
FYI: I updated the project languages of the project we use for e2e tests, so that the locale switcher is available in the VRTs, and thus catch these issues sooner.

PR #2465, opened 2022-02-01
Upgrade to Stylelint v14

emmenko:


CarlosCortizasCT (approved):
LGTM

Rhotimee (approved):

tdeekens (approved):
Thanks for including this!

emmenko:
FYI: note that css-in-js support from stylelint is not available anymore.

I tried different things but I wasn't happy with the results so I just decided to drop it as well.

PR #2476, opened 2022-02-07
refactor(website): replace linkinator with linkcheck, fix broken links

emmenko:
The `linkinator` package is not able to detect/check broken anchor links.

We decided to change the links check tool and found this https://github.com/filiph/linkcheck


emmenko:
We might not use this tool after all, as it has its own limitations.

PR #2485, opened 2022-02-16
fix(create-mc-app): reading user input in Node v14

### LINKED ISSUE #2484: Problem with reading user input when `create-mc-app` executed in Node v14
Hey everyone 👋 
I'm just learning the ropes here but I encountered a problem so I thought about reporting that.
I hope you don't consider it too forward that I make a PR with changes that seemed to be fixing that issue for me. 

**Describe the issue**
I noticed that in case of executing `create-mc-app` script in node 14 user’s input for `entryPointUriPath`:
```js
// packages/create-mc-app/src/parse-arguments.js
const answerEntryPointUriPath = await question(
    `Provide the Custom Application entryPointUriPath (default "${randomEntryPointUriPath}"): `
  );
```
 and `initialProjectKey`
```js
// packages/create-mc-app/src/parse-arguments.js
const initialProjectKey = await question(
    `Provide the initial project key for local development: `
  );
```
 is not stored but resolves to `undefined`. As a result the script fails silently.

Sidenote:
- `create-mc-app` executed with variables predefined via flags like 
```bash
npx @commercetools-frontend/create-mc-app \
  my-new-custom-application-project \
  --template starter \
  --entry-point-uri-path foo \
  --initial-project-key bar
```
works as expected.
- `create-mc-app` executed in Node v16 and Node v17 works as expected.

**To Reproduce**

Depending on how Node is installed in your system switch to Node v14 (in case of using `nvm` run `nvm use 14`).
Proceed with instructions of [installing a starter template](https://docs.commercetools.com/custom-applications/getting-started#installing-a-starter-template).

kark:
Fix reading user input to `entryPointUriPath` and `initialProjectKey`
variables in case of `create-mc-app` executed in Node v14.

closes #2484

The way I tested it:
```
cd packages/create-mc-app
npm pack
```
Then installed the package globally and run the script
```
npm i -g path_to_tarball
create-mc-app …
```

### DIFF
--- a/.changeset/fair-emus-hammer.md
+++ b/.changeset/fair-emus-hammer.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/create-mc-app': patch
+---
+
+Fix reading user input to `entryPointUriPath` and `initialProjectKey` variables in case of `create-mc-app` executed in Node v14.
--- a/packages/create-mc-app/src/parse-arguments.js
+++ b/packages/create-mc-app/src/parse-arguments.js
@@ -1,6 +1,5 @@
 /* eslint-disable no-console */
 const path = require('path');
-const util = require('util');
 const readline = require('readline');
 const crypto = require('crypto');
 const {
@@ -14,7 +13,9 @@ const rl = readline.createInterface({
   input: process.stdin,
   output: process.stdout,
 });
-const question = util.promisify(rl.question).bind(rl);
+
+const question = (query) =>
+  new Promise((resolve) => rl.question(query, resolve));
 
 const getTemplateName = (flags) => flags.template || 'starter';
 const getEntryPointUriPath = async (flags) => {


.changeset/fair-emus-hammer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-frontend/create-mc-app': patch
+---
+
+Fix reading user input to `entryPointUriPath` and `initialProjectKey` variables in case of `create-mc-app` executed in Node v14.
```
emmenko, line 5:
```suggestion
Fix compatibility with Node `< v14.18` for reading user input.
```

emmenko (commented):
Alright, we had a look at this together and found out that the problem appears when using Node `< 14.18`.

So we decided to use the `Promise` wrapper and avoid using `util.promisify`.

emmenko:
@kark I think it's better if you open a new PR from the repository directly, instead of using a fork, as you won't have access to the github secrets.

kark:
Replaced with #2491

PR #2487, opened 2022-02-18
Dependency problems when initializing custom-app template

### LINKED ISSUE #2486: Custom Application - Getting Started (Package installation, starting local dev server)
I tried out installing the starter template for the custom application, based on the documentation `https://docs.commercetools.com/custom-applications/getting-started`.

**1.** Upon executing the `npx @commercetools-frontend/create-mc-app my-new-custom-application-project --template starter`, I encountered errors during the installation of the dependencies:
[2022-02-09T04_02_45_177Z-debug (CREATE-MC-APP).txt](https://github.com/commercetools/merchant-center-application-kit/files/8089699/2022-02-09T04_02_45_177Z-debug.CREATE-MC-APP.txt)

**2. (A).** I retried installing the packages using `npm install --force` && `npm install --legacy-peer-deps`. Both allowed the dependencies to be installed.
**2. (B).** Downgrading to eslint@7.32.0 (from version 8.7.0) also solved the packages installation problem.

**3.** Running `npm start` will then lead to errors:
[2022-02-09T08_29_34_140Z-debug (NPM START).txt](https://github.com/commercetools/merchant-center-application-kit/files/8089701/2022-02-09T08_29_34_140Z-debug.NPM.START.txt)

Running on node v.14.19, npm v.8.1.2, Windows 10.

CarlosCortizasCT:
### Summary

closes #2486

When trying to create a new custom application, initialization command fails because an incompatibility between `jest-runner-eslint` and `eslint`.

### Description

We're using `jest-runner-eslint` version **0.11.1** which has a peer dependency on `eslint` version 6 or 7, but in the project we're using version 8.

`jest-runner-eslint` released a new version ([1.0.0](https://github.com/jest-community/jest-runner-eslint/compare/v0.11.1...v1.0.0)) just to add support for `eslint` version 8 as its peer dependency so we need to update our `package.json` to use it.


application-templates/starter/package.json
```diff
@@ -53,7 +53,7 @@
     "eslint": "8.7.0",
     "eslint-formatter-pretty": "4.1.0",
     "jest": "27.4.7",
-    "jest-runner-eslint": "0.11.1",
+    "jest-runner-eslint": "^1.0.0",
```
tdeekens, line 56: Let's pin it. We have bots running so no need to have it flexible.

```suggestion
    "jest-runner-eslint": "1.0.0",
```

package.json
```diff
@@ -131,7 +131,7 @@
     "jest": "27.4.7",
     "jest-each": "27.4.6",
     "jest-puppeteer": "6.0.3",
-    "jest-runner-eslint": "0.11.1",
+    "jest-runner-eslint": "^1.0.0",
```
tdeekens, line 134:
```suggestion
    "jest-runner-eslint": "1.0.0",
```

tdeekens (commented):
Let's pin it. Make suggestions but it would chnage the yarn.lock so just do it on your machine :)

application-templates/starter/package.json
tdeekens: Strange I am not getting the error locally. You're sure all versions of the dependency didn't have the caret?

CarlosCortizasCT: Yes, you're right. I reviewed my first commit and I messed with the caret in one of the package.json files.

I've just update both of them to use a fixed version.

kark:
Thank you Carlos. I am still having problems with running the `create-mc-app` in v15.14.0 (npm v7.7.6) - all works fine in v16 and v17.
I am not sure if this is a real problem or not, but I wanted to be vocal about it.
My method of testing involves global installation of the `@commercetools-frontend/create-mc-app` from tarball.

**Step by step**
```bash
nvm use 15
# in merchant-center-application-kit repo folder branch `cm-fix-starter-dependency` commit `5934b37b`
yarn
yarn build
cd packages/create-mc-app
npm pack
npm i -g full_path_to_tarball
create-mc-app my-new-custom-application-project --template starter
```

This is the output I am getting.
[output.txt](https://github.com/commercetools/merchant-center-application-kit/files/8096711/output.txt)

CarlosCortizasCT:
> Thank you Carlos. I am still having problems with running the `create-mc-app` in node v15.14.0 (npm v7.7.6) - all works fine in v16 and v17. I am not sure if this is a real problem or not, but I wanted to be vocal about it. My method of testing involves global installation of the `@commercetools-frontend/create-mc-app` from tarball.
> 
> **Step by step**
> 
> ```shell
> nvm use 15
> # in merchant-center-application-kit repo folder branch `cm-fix-starter-dependency` commit `5934b37b`
> yarn
> yarn build
> cd packages/create-mc-app
> npm pack
> npm i -g full_path_to_tarball
> create-mc-app my-new-custom-application-project --template starter
> ```
> 
> This is the output I am getting. [output.txt](https://github.com/commercetools/merchant-center-application-kit/files/8096711/output.txt)

Well, I think there are two issues here:
1. You're running the local `create-mc-app` build but what it is downloading the starter template from the [`main` branch](https://github.com/commercetools/merchant-center-application-kit/blob/main/packages/create-mc-app/src/tasks/download-template.js#L36) by default so the dependencies installation is not using the change in this PR. I think you should use this command to actually use the template from this branch:
```
npx @commercetools-frontend/create-mc-app@latest my-new-custom-application-project --template starter --template-version cm-fix-starter-dependency
```
2. The issue in the output file you uploaded is not with the `jest-runner-eslint` module but with `eslint-plugin-jest`. This is happening because that packages [does not support node v15](https://github.com/jest-community/eslint-plugin-jest/blob/main/package.json#L142). Actually, there are many other packages with the same situation (see screenshot below). So I don't think we are in a position to address that version. I guess it's on purpose because of the nature of nodejs [odd version numbers](https://nodesource.com/blog/understanding-how-node-js-release-lines-work/).
@tdeekens what do you think about this?

image

application-templates/starter/package.json
tdeekens: Lovely, thanks a lot!

tdeekens (approved):

CarlosCortizasCT:
@tdeekens 

You mean updating main `package.json` **engines** attribute, right?

image

tdeekens:
Yes, that's the engines field I meant. Sorry for not being precise. Good point. v17 should be included. The last point was to make sure v18 (when released) isn't automatically included as we don't/can't know if it would work.

CarlosCortizasCT:
> Yes, that's the engines field I meant. Sorry for not being precise. Good point. v17 should be included. The last point was to make sure v18 (when released) isn't automatically included as we don't/can't know if it would work.

Updated in this commit: c433178acac5b38c522436a6606870433e3bbf50

Thanks for your help 🙌

packages/create-mc-app/package.json
```diff
@@ -25,6 +25,6 @@
     "semver": "7.3.5"
   },
   "engines": {
-    "node": ">=14"
+    "node": "^14 || ^16 || ^17"
```
tdeekens, line 28: Thanks. Nit 
```suggestion
    "node": "14.x || 16.x || 17.x"
```
As that's the notation we often use in the repo. Mainly to keep it simple. Anybody understands a `14.x` while a  `^` usually leads to people going "Hot dang what was `~` and `^` again?" :)

tdeekens (approved):

PR #2490, opened 2022-02-21
Added missing changeset and update documentation

CarlosCortizasCT:
### Summary

This is a follow up from this PR where I missed adding the changeset for the `create-mc-app` package.

Also, I updated public documentation so we state nodejs versions we support in the prerequisites section of Custom Applications.

### DIFF
--- a/website/src/content/getting-started.mdx
+++ b/website/src/content/getting-started.mdx
@@ -9,7 +9,7 @@ Get up and running in a few steps and start developing Custom Applications for t
 Before you start development, make sure to have the following:
 
 - A [commercetools](https://docs.commercetools.com/tutorials/getting-started) account and a Project.
-- [Node.js](https://nodejs.org/en/) installed (version `>=14`, recommended `>=16`). Either [`yarn`](https://yarnpkg.com/) or [`npm`](https://www.npmjs.com/get-npm) as a package manager.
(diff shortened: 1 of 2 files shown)


website/src/content/getting-started.mdx
```diff
@@ -9,7 +9,7 @@ Get up and running in a few steps and start developing Custom Applications for t
 Before you start development, make sure to have the following:
 
 - A [commercetools](https://docs.commercetools.com/tutorials/getting-started) account and a Project.
-- [Node.js](https://nodejs.org/en/) installed (version `>=14`, recommended `>=16`). Either [`yarn`](https://yarnpkg.com/) or [`npm`](https://www.npmjs.com/get-npm) as a package manager.
+- [Node.js](https://nodejs.org/en/) installed (version `14.x`, `16.x` or `17.x`; recommended `>=16`). Either [`yarn`](https://yarnpkg.com/) or [`npm`](https://www.npmjs.com/get-npm) as a package manager.
```
emmenko, line 12: Nit: I think it might be enough to say it like this.

```suggestion
- [Node.js](https://nodejs.org/en/) installed (version `14.x`, recommended `>=16`). Either [`yarn`](https://yarnpkg.com/) or [`npm`](https://www.npmjs.com/get-npm) as a package manager.
```

This way it's clear that we exclude `v15`, but we leave it "open" for any versions greater than `16`.

emmenko (approved):

PR #2492, opened 2022-02-23
refactor: use linkcheck (for local server) to additionally check anchor links

emmenko:
Another try at using `linkcheck` (previous try #2476 )


.yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable */
+//prettier-ignore
+module.exports = {
```
CarlosCortizasCT, line 3: Just curious about what is this plugin for 🤔

CarlosCortizasCT (approved):
lgtm

.yarn/plugins/@yarnpkg/plugin-workspace-tools.cjs
emmenko: To be able to use yarn commands such as `yarn workspaces foreach`.

https://github.com/yarnpkg/berry/blob/HEAD/packages/plugin-workspace-tools/README.md

I had a command using that in this PR but then changed it back. The plugin is anyway useful and we also use it in other repos.

tdeekens (approved):
Looks good to me. Thanks.

PR #2494, opened 2022-02-23
fix(express): fix auth middleware in serverless functions

### LINKED ISSUE #2147: adapt createSessionAuthVerifier for aws api gateway http api
Hi! we're using this library `@commercetools-backend/express` to verify requests coming from custom applications. it works on express apps, but does not work when used in combination of `aws api gateway + lambda` (exact example of using it in google cloud function https://docs.commercetools.com/custom-applications/main-concepts/proxy-to-external-api#usage-for-serverless-functions does not work in aws)

**the cause**: request object in lambda does not have `url` and `originalUrl` fields. 

our current workaround it this:
```
request.originalUrl = request.rawPath + (
    request.rawQueryString 
      ? `?${request.rawQueryString}`
      : ''
  );
```

would be great to have this check built in in the library.

CarlosCortizasCT:
### Summary

When using the session auth verifier in a serverless environment we can get an error because of the shape the _request-like_ object the handler receives and passes to the middleware is not the expected.

closes #2147 

### Description
The middleware expects the object to be an `IncomingMessage` but that's not always the case. For instance, in AWS lambdas the handler will receive an `Event` object  with [this shape](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a1260a1f3d40f239b53fd29effba594b0d1bee08/types/aws-lambda/handler.d.ts#L84) (the specific type is [this one](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a1260a1f3d40f239b53fd29effba594b0d1bee08/types/aws-lambda/trigger/api-gateway-proxy.d.ts#L12)).

This is important because the middleware needs to find the original URL of the request to know about the `audience` it will use to validate the `JWT`.

Instead of trying to figure out how to resolve the URL in all different environments we will provide a new option to the middleware builder which will expect a function that will return that URL. This function will receive as parameter, the `request` like object which is passed to the middleware from the lambda function handler.

This is an example for AWS lambda function using its [payload v2](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html).
```
const {
  createSessionAuthVerifier,
  CLOUD_IDENTIFIERS,
} = require('@commercetools-backend/express');

const sessionAuthVerifier = createSessionAuthVerifier({
  audience: 'https://my-api-server.com',
  issuer: CLOUD_IDENTIFIERS.GCP_EU,
  // HERE IS WHERE WE RESOLVE THE URL <-----------
  urlResolver: (request) => `${request.rawPath}${request.rawQueryString ? '?' + request.rawQueryString : ''}`
});

exports.handler = async function (request, response) {
  try {
    await sessionAuthVerifier(request, response);
  } catch (error) {
    response.status(401).send(JSON.stringify({ message: 'Unauthorized' }));
    return;
  }
  // `request.session` contains the useful information
};
```


packages-backend/express/src/middlewares/session-middleware.spec.ts
```diff
@@ -51,8 +56,49 @@ describe.each`
...
+
+      return { sessionMiddleware, fakeRequest, fakeResponse };
+    }
+
+    it('should verify the token and attach the session info to the request', async () => {
+      const { sessionMiddleware, fakeRequest, fakeResponse } = setupTest();
+
+      await new Promise<void>((resolve, reject) => {
+        // @ts-ignore
+        sessionMiddleware(fakeRequest, fakeResponse, (error) => {
```
tdeekens, line 71: Nit: maybe it would make it easier to read in tests if we'd have a `function promisifyMiddleware` which takes a middleware does that we do here. Then we could use it here to await a `promisifiedMiddleware` and below. Just thinking out loud while reading the code :)

packages-backend/express/src/types.ts
```diff
@@ -25,6 +25,16 @@ export type TSessionMiddlewareOptions = {
...
   jwks?: Omit<ExpressJwtOptions, 'jwksUri'>;
+
+  // This middleware needs to resolve the request URL to know about the
+  // 'audience' it needs to use for the JWT.
+  // By default it assumes a nodejs IncomingMessage object as its first parameter,
+  // however this is not always the case (eg; AWS lambda functions).
+  // When providing this function you will be delegated to resolve the URL from
+  // the first parameter the middleware receives.
+  // Form example, in AWS lambda functions it will be an Event object
+  // (https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a1260a1f3d40f239b53fd29effba594b0d1bee08/types/aws-lambda/trigger/api-gateway-proxy.d.ts#L12)
```
tdeekens, line 36: Rewording to understand the issue :). Drop if not better.

```suggestion
  // In order to determine the 'audience' for the JWT the middleware needs to resolve the request URL.
  // As a default the middleware assumes takes the IncomingMessage of Node.js as its first parameter.
  // This default needs to be overwritten at times.  For instance in the case of AWS lambda functions.
  // Do to so this function, when provided, allows delegation of resolving the URL from the middleware's request.
  // In the case of an AWS lambda function an  Event object will be passed as defined here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a1260a1f3d40f239b53fd29effba594b0d1bee08/types/aws-lambda/trigger/api-gateway-proxy.d.ts#L12
```

packages-backend/express/src/middlewares/session-middleware.spec.ts
CarlosCortizasCT: Thanks for the suggestions, @tdeekens.

I had thought about it but wasn't sure it was needed.
```
function runSessionMiddleware(middleware: unknown, request: unknown, response: unknown) {
  return new Promise<void>((resolve, reject) => {
    // @ts-ignore
    middleware(request, response, (error) => {
      if (error) reject(error);
      else resolve();
    });
  });
}

it('...', async () => {
  const { sessionMiddleware, fakeRequest, fakeResponse } = setupTest();

  await runSessionMiddleware(sessionMiddleware, fakeRequest, fakeResponse);

  expect(...);
});
```
Do you think something like that would be more readable?

packages-backend/express/src/types.ts
CarlosCortizasCT: I ended up with this text when trying to merge both proposals:
```

  // In order to determine the 'audience' for JWT validation, the middleware needs
  // to resolve the request url.
  // By default, the middleware assumes it receives a Node.js IncomingMessage object
  // as its first parameter.
  // This default behaviour might not work for every scenario as, for example,
  // AWS lambda functions, where that first parameter is an Event object.
  // (Reference: https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html)
  // To handle these situations, this option allows the delegation of the URL resolution
  // from the request-like object provided to the middleware as its first parameter.
```

What do you think about it?

packages-backend/express/src/middlewares/session-middleware.spec.ts
tdeekens: Subjectively to me when I _first_ read the code yes. Upon _second_ reading not hat much anymore. In the end it's just moving stuff around and the "complexity" is somewhere. Still a `await runSessionMiddleware` or `await waitForSessionMiddleware` would make the `it` test case itself more learn which is something I personally like.

Totally up to you :)

packages-backend/express/src/types.ts
tdeekens: Best of both worlds. I like it. Thanks!

CarlosCortizasCT: Updated in this commit 8142081dec38c0af7bd1d5c81e0b84a73c1d30a4

packages-backend/express/src/middlewares/session-middleware.spec.ts
CarlosCortizasCT: I like the `waitForSessionMiddleware` naming so I went that path.

Updated in this commit f41c00b7da421e517ed093abfe4c1d357062e552

packages-backend/express/src/types.ts
```diff
@@ -25,6 +25,17 @@ export type TSessionMiddlewareOptions = {
...
+  // In order to determine the 'audience' for JWT validation, the middleware needs
+  // to resolve the request url.
+  // By default, the middleware assumes it receives a Node.js IncomingMessage object
+  // as its first parameter.
+  // This default behaviour might not work for every scenario as, for example,
+  // AWS lambda functions, where that first parameter is an Event object.
+  // (Reference: https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html)
+  // To handle these situations, this option allows the delegation of the URL resolution
+  // from the request-like object provided to the middleware as its first parameter.
+  urlResolver?: (request: unknown) => string;
```
emmenko, line 38: I would keep it a bit simpler. Also I would maybe name the function something like `rewriteRequestUrl` (which we also use internally in the MC API).

```suggestion
  // By default we assume that the `request` is a Node.js-like object having either a `originalUrl` or `url` properties.
  // If that's not the case (for example in AWS Lambda functions) you need to correctly map the URL of
  // the `request` to return the full URI path, including query string parameters. 
  rewriteRequestUrl?: (request: unknown) => string;
```

On another note, instead of using `unknown` we can (re)use the generic type that we defined here:
https://github.com/commercetools/merchant-center-application-kit/blob/39a387cc507f244b23507a0efd80166e36038f68/packages-backend/express/src/auth.ts#L120

Then the `TSessionMiddlewareOptions` would also need to accept the generic type argument

```ts
export type TSessionMiddlewareOptions<Request extends IncomingMessage> {
  rewriteRequestUrl?: (request: Request) => string;
}
```

emmenko (commented):
Thanks, looking good so far.

Please also update the docs to document this new functionality (and add a changeset). Let me know if you need help with that.

packages-backend/express/src/types.ts
CarlosCortizasCT: I'm not sure about providing a type which needs to extend `IncomingMessage`. In the case of AWS lambda function, the type of the object received will have [this shape](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a1260a1f3d40f239b53fd29effba594b0d1bee08/types/aws-lambda/trigger/api-gateway-proxy.d.ts#L169) which does not extend [`IncomingMessage`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/http.d.ts#L785) so I guess consumers using TS for their source code will get a typing error.

The problem is that different providers could use totally different types so I'm unsure if we can provide any type here.

emmenko: > I'm not sure about providing a type which needs to extend `IncomingMessage`.

Ah yes, you're right. I'm wondering then if it would make sense to adjust that as well. So instead of requiring `Request extends IncomingMessage` we can create our own "base" request type with the minimal properties that we expect to be used.

```ts
interface TBaseRequest {
  headers: Record<string, undefined>,
  originalUrl?: string,
  url?: string,
}

interface TBaseResponse {}
```

> The way it is right now, this function is not expected to do that but return the URL.

True, you're right. How about `getRequestUrl`?

CarlosCortizasCT: I think that the nature of this optional function is to deal with contexts we don't know about. I mean, if the received first middleware parameter is what we expect, it will include know attributes (`url` or `originalUrl`). But if it's not, then we can't know what the shape of the object would be, unless we prepare our code to all different providers we support: in that case, we wouldn't need this option because our code would handle it.

That's why I think this function parameter shouldn't be typed. We want to support scenarios we don't know about. 
I mean, for know scenarios, our code will work. For unknown scenarios, we just don't have a clue about the received object so we totally delegate the URL resolution up to the consumers of the middleware.

Does it make sense?



> True, you're right. How about getRequestUrl?

Sure! I'll change it.

packages-backend/express/src/middlewares/session-middleware.spec.ts
tdeekens: 🙇🏼 thank you!

packages-backend/express/src/auth.ts
```diff
@@ -149,9 +155,15 @@ function createSessionAuthVerifier<
...
 
-    // @ts-ignore: the node HTTP request does not know about `originalUrl`
-    const requestUrlPath = request.originalUrl ?? request.url;
-    const audience = getConfiguredAudience(options, requestUrlPath);
+    const requestUrlPath = options.getRequestUrl
+      ? options.getRequestUrl(request)
+      : request.originalUrl ?? request.url;
+
+    if (!requestUrlPath || requestUrlPath === 'undefined') {
+      throw new Error('Could not get the request URL.');
```
CarlosCortizasCT, line 163: Not sure if this message is enough or we should be more verbose (maybe a link to the docs?).

packages-backend/express/src/auth.ts
```diff
@@ -149,9 +155,15 @@ function createSessionAuthVerifier<
...
     }
 
-    // @ts-ignore: the node HTTP request does not know about `originalUrl`
-    const requestUrlPath = request.originalUrl ?? request.url;
-    const audience = getConfiguredAudience(options, requestUrlPath);
+    const requestUrlPath = options.getRequestUrl
+      ? options.getRequestUrl(request)
+      : request.originalUrl ?? request.url;
+
+    if (!requestUrlPath || requestUrlPath === 'undefined') {
```
CarlosCortizasCT, line 162: Comparision with `undefined` string is to take into account `getRequestUrl` implementations like this:
```
const sessionAuthVerifier = createSessionAuthVerifier({
  audience: 'https://my-api-server.com',
  issuer: CLOUD_IDENTIFIERS.GCP_EU,
  getRequestUrl: (event) => `${event.rawPath}${event.rawQueryString ? '?' + event.rawQueryString : ''}`
});
```
It would return that `'undefined'` string and I think that implementation would be common.

packages-backend/express/src/types.ts
CarlosCortizasCT: Yes, you're right. I didn't noticed we're at least expecting the request-like object to have the `headers`.

I updated the types in this commit 51c8b2e915793a334a26f9d9c8b3db796e95fac7

website/src/content/concepts/integrate-with-your-own-api.mdx
```diff
@@ -151,6 +151,8 @@ The middleware requires some options:
 
 - `jwks` (_object_): See [options of `jwks-rsa`](https://github.com/auth0/node-jwks-rsa/).
 
+- `getRequestUrl` (_(request) => string_): In case the middleware don't get an Express Request object as first parameter (see AWS Lambda example below), you need to provide this function to extract the full URL from the received object.
```
emmenko, line 154:
```suggestion
- `getRequestUrl` (_function_): Returns the URL (URI path + query string) of the `request` to validate the audience. Use this in case the `request` object does not contain either a property `originalUrl` or `url` (see serverless example below for using AWS Lambda functions).

  ```ts
  type function getRequestUrl(request: Request): string;
  ```
```

website/src/content/concepts/integrate-with-your-own-api.mdx
```diff
@@ -202,6 +204,39 @@ exports.handler = async function (request, response) {
 };
 ```
 
+There are other providers that do not pass a Node.js/Express `Request` to the handler as AWS lambda functions do.
+In that context, the handler will receive an `Event` ([reference](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-create-api-as-simple-proxy-for-lambda.html)) object with a different structure.
+As the verifier needs to access the original URL to validate the JWT and it will not find it in the expected attribute,
+you would need to provide a custom option when creating the verifier to resolve it.
+
+This is an example of validating the JWT for AWS Lambda Functions ([v2 payload](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html))
```
emmenko, line 212:
```suggestion
Other cloud providers might not use a standard Node.js/Express `request` object. For example AWS Lambda functions  receive an [`Event` object](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-create-api-as-simple-proxy-for-lambda.html) with different properties.

In these cases you need to implement the `getRequestUrl` function to correctly map the expected URL (URI path + query string). The following example shows how to map an AWS Lambda function [event](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html).
```

emmenko (commented):
Looking good now, thanks!

packages-backend/express/src/auth.ts
CarlosCortizasCT: This is something I realized when writing the tests.

Let's say you `request` _like_ object has the properties `rawPath` and `rawQueryPath` but you incorrectly think the properties are called `path` and `queryPath`.
Also, let's say you implement a simple `getRequestUrl` function like this (because it seems good enough):
```
getRequestUrl: (event) => `${event.path}${event.queryString ? '?' + event.queryString : ''}`
```
Now the problem is that executing that function will yield the string `'undefined'` as its result.

Instead of raising the "Invalid request path." error, as the url is not _falsy_, this value will be passed to the `express-jwt` package and you will get an "jwt audience invalid" error.

I thought that simple implementation would be quite common and that's why I opted for extending the condition.

I know this is a controversial and maybe personal taste so I can change it if we don't feel comfortable with it.

packages-backend/express/src/auth.ts
```diff
@@ -148,9 +154,17 @@ function createSessionAuthVerifier<
       issuer = mapToLegacyIssuer(cloudIdentifierHeader) ?? issuer;
     }
 
-    // @ts-ignore: the node HTTP request does not know about `originalUrl`
-    const requestUrlPath = request.originalUrl ?? request.url;
-    const audience = getConfiguredAudience(options, requestUrlPath);
+    const requestUrlPath = options.getRequestUrl
+      ? options.getRequestUrl(request)
+      : request.originalUrl ?? request.url;
```
tdeekens, line 159: Just the right amount of complexity for the ternary to not explode 😄

tdeekens (approved):
Thanks. Looks really nice to me now.

website/src/content/concepts/integrate-with-your-own-api.mdx
```diff
@@ -151,6 +151,11 @@ The middleware requires some options:
 
 - `jwks` (_object_): See [options of `jwks-rsa`](https://github.com/auth0/node-jwks-rsa/).
 
+- `getRequestUrl` (_function_): Returns the URL (URI path + query string) of the `request` to validate the audience. Use this in case the `request` object does not contain either an `originalUrl` or an `url` property (see serverless example below for using AWS Lambda functions).
+
+  ```ts
+  type function getRequestUrl(request: Request): string;
```
emmenko, line 157: Missing closing code block backticks (github suggestion ignore them apparently.

packages-backend/express/src/auth.ts
```diff
@@ -148,9 +154,17 @@ function createSessionAuthVerifier<
...
+
+    if (!requestUrlPath) {
+      throw new Error(
+        'Invalid request path. Please make sure that the request object has either a property `originalUrl` or `url`. If not, you should implement the `getRequestUrl` function. More info at https://docs.commercetools.com/custom-applications/concepts/integrate-with-your-own-api#validating-the-json-web-token'
```
emmenko, line 163:
```suggestion
        'Invalid request URI path. Please make sure that the `request` object has either a property `originalUrl` or `url`. If not, you should implement the `getRequestUrl` function. More info at https://docs.commercetools.com/custom-applications/concepts/integrate-with-your-own-api#validating-the-json-web-token'
```

packages-backend/express/src/types.ts
```diff
@@ -25,6 +31,17 @@ export type TSessionMiddlewareOptions = {
...
+
+  // In order to determine the 'audience' for JWT validation, the middleware needs
+  // to resolve the request url.
+  // By default, the middleware assumes it receives a Node.js IncomingMessage object
+  // as its first parameter.
+  // This default behaviour might not work for every scenario as, for example,
+  // AWS lambda functions, where that first parameter is an Event object.
+  // (Reference: https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html)
+  // To handle these situations, this option allows the delegation of the URL resolution
+  // from the request-like object provided to the middleware as its first parameter.
```
emmenko, line 43: We already explain the AWS example in the docs, I think we can keep this comment simple as to what the function is for. 

```suggestion
  // By default we assume that the `request` is a Node.js-like object having either
  // a `originalUrl` or `url` properties.
  // If that's not the case (for example in AWS Lambda functions) you need to correctly
  // map the URL (URI path + query string) of the `request`.
```

.changeset/early-files-shave.md
```diff
@@ -0,0 +1,22 @@
...
+
+Example for AWS Lambda Funtions:
+
+```
+const sessionAuthVerifier = createSessionAuthVerifier({
+  audience: 'https://my-api-server.com',
+  issuer: CLOUD_IDENTIFIERS.GCP_EU,
+  getRequestUrl: (event) => `${event.rawPath}${event.rawQueryString ? '?' + event.rawQueryString : ''}`
+});
+```
```
emmenko, line 22: Since we have already documented this in the docs, we can keep the changelog short and link to the docs.

```suggestion
Allow to map a `request` URL (URI path + query string) using the `getRequestUrl` function in case the `request` object does not contain either a `originalUrl` or `url` properties.

More info at https://docs.commercetools.com/custom-applications/concepts/integrate-with-your-own-api#validating-the-json-web-token.
```

emmenko (commented):
Thanks! Looks good now, just some minor docs improvements and then we're good to go.

website/src/content/concepts/integrate-with-your-own-api.mdx
CarlosCortizasCT: Updated in this commit e4be0ae17a4678626bf426a1c7ab237e458a00d1

packages-backend/express/src/auth.ts
CarlosCortizasCT: Updated in this commit e4be0ae17a4678626bf426a1c7ab237e458a00d1

packages-backend/express/src/types.ts
CarlosCortizasCT: Updated in this commit e4be0ae17a4678626bf426a1c7ab237e458a00d1

.changeset/early-files-shave.md
CarlosCortizasCT: Updated in this commit e4be0ae17a4678626bf426a1c7ab237e458a00d1

emmenko (approved):
Thanks! 💯

website/src/content/concepts/integrate-with-your-own-api.mdx
```diff
@@ -202,6 +208,36 @@ exports.handler = async function (request, response) {
...
+    await sessionAuthVerifier(event, context);
+  } catch (error) {
+    return {
+      statusCode: 401,
+      body: JSON.stringify({ message: 'Unauthorized' }),
+    };
+  }
+  // `event.session` contains the useful information
+};
+```
```
emmenko, line 238: FYI @commercetools/tech-writers

website/src/content/concepts/integrate-with-your-own-api.mdx
```diff
@@ -202,6 +208,36 @@ exports.handler = async function (request, response) {
 };
 ```
 
+Other cloud providers might not use a standard Node.js/Express `request` object. For example AWS Lambda functions receive an [`Event` object](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-create-api-as-simple-proxy-for-lambda.html) with different properties.
```
CarlosCortizasCT, line 211: FYI @commercetools/tech-writers

adinakleine, line 211:
```suggestion
Other cloud providers might not use a standard Node.js/Express `request` object. For example AWS Lambda functions receive an [`event` object](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-create-api-as-simple-proxy-for-lambda.html) with different properties.
```

CarlosCortizasCT: Thanks for the suggestion.

Updated in this commit 8991f1078f37d0b2ea7f6e8e8dbc389387c8a64b

PR #2495, opened 2022-02-24
docs: developing forms

emmenko:
Document developing forms in Custom Applications


website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,540 @@
+---
+title: Forms
+---
+
+In the previous sections we learned how to start building the actual User Interface by using UI components and fetching data. One of the most common thing to implement in a User Interface is forms.
```
tdeekens, line 5:
```suggestion
In the previous sections we learned how to start building the actual User Interface by using UI components and fetching data. One of the most common requirements when implementing in a User Interface is centered around forms.
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,540 @@
...
+
+To facilitate the development of UI forms, we provide many different ready-to-use input and field components based on [commercetools Design System](https://github.com/commercetools/ui-kit).
+
+# Form state management
+
+Forms generally consist of a group of inputs that users interact with and eventually some actions will be triggered.
```
tdeekens, line 11:
```suggestion
Forms generally consist of a group of inputs that users interact with to eventually trigger some actions.
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,540 @@
...
+All the user interactions need to be tracked, managed, and reflected in the UI. This is where the form needs to have some state to be able to handle all of that.
```
tdeekens, line 12:
```suggestion
In this process all user interactions need to be tracked, managed, and reflected in the UI. To achieve this a form needs to maintain its state.
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,540 @@
...
+
+At commercetools we have settled on using [Formik](https://formik.org) as the preferred form state management library when building User Interfaces.
+
+Formik comes with several built-in features such as validations, array fields, input states, async submission, etc.
```
tdeekens, line 16:
```suggestion
Formik comes with several built-in features such as validations, array fields, input states, async submission and many more.
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,540 @@
...
+    validate,
+    onSubmit: async (formikValues) => {
+      alert(`key: ${formikValues.key}`);
+      // Do something async
+    },
+  });
+
+  return (
+    <form onSubmit={formik.handleSubmit}>
+      <TextField
```
tdeekens, line 41: Should we recommend using `useField` to spread in the props easier?

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,540 @@
...
+
+<Info>
+
+All UIKit field components have a related input component, for example `<TextField>` -> `<TextInput>`.
+
+</Info>
+
+## Accessibility support
+
+All field and input components have built-in support for accessibility features such as aria labels, keyboard navigation and focus, error messaging.
```
tdeekens, line 81:
```suggestion
All field and input components have built-in support for accessibility features such as aria labels, keyboard navigation and focus management and error messaging.
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,540 @@
...
+
+<Info>
+
+Testing these field components with [React Testing Library](https://github.com/testing-library/react-testing-library) can be easily done using selectors such as `*ByLabelText`, `*ByRole`, etc.
+
+</Info>
+
+# Form validation
+
+One of the important parts of forms is validation. It's important to validate things like required fields, or to check that the value is a valid URL, or even perform asynchronous checks against an API.
```
tdeekens, line 91:
```suggestion
One of the important parts of forms is validation. It is important to validate constraints of forms such as required fields. Moreover, some fields might have semantic requirements which for instance that a value is a valid URL. Lastly, a form can perform not only client-side but also asynchronous validation against an API to ensure data correctness before submission.
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,540 @@
...
+For that, Formik allows you to implement a [`validate` function](https://formik.org/docs/api/formik#validate-values-values--formikerrorsvalues--promiseany) that should return an object of errors.
+
+```ts
+type FieldErrors = Record<string, boolean>;
+type FormikErrors = Record<string, FieldErrors>;
+
+declare const validate = (values: FormikValues) => FormikErrors;
+```
+
+Specifically, the `FormikErrors` object should contain properties as the name of the form fields with values being objects with error keys. Let's take a look at an example:
```
tdeekens, line 102:
```suggestion
Specifically, the `FormikErrors` object should contain properties correlating to fields names with their values being objects with reasons for the given error. Let's take a look at an example:
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,540 @@
...
+
+```js title="validate.js"
+import TextInput from '@commercetools-uikit/text-input';
+
+const validate = (values) => {
+  const errors = {};
+
+  if (TextInput.isEmpty(values.key)) {
+    errors.key = { missing: true };
```
tdeekens, line 111: Nit: I would not use `key` as an example field. It might easily be confusing to reference from text as `The object's key" and "the key field" mean different things. Maybe `sku` or something e-comm specific is easier for readers to follow along with.

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,540 @@
...
+  errors={formik.errors.key}
+  touched={formik.touched.key}
+  onChange={formik.handleChange}
+  onBlur={formik.handleBlur}
+/>
+```
+
+<Info>
+
+Error messages are only shown when the `touched` value for the specific field is set to `true`. This happens whenever the user stops interacting with a field (loses `focus`). The point here is that as long as the user is interacting with a field for the first time there is no need to show a validation.
```
tdeekens, line 137:
```suggestion
Error messages are only shown when the `touched` value for the specific field is set to `true`. This happens whenever the user stops interacting with a field and moves to another so that the prior field loses `focus`. The point here is that as long as the user is interacting with a field for the first time there is no need to show a validation.
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,540 @@
...
+
+```js title="conversions.js"
+export const formValuesToDoc = (formValues) => ({
+  key: formValues.key,
+})
+```
+
+<Info>
+
+Most of the time you would have a simple 1:1 mapping. However, we still recommend to use these conversion functions as a best practice and to help decoupling the data transformation logic from the form component.
```
tdeekens, line 225: Maybe add that it also helps not sending uninteresting fields to a form down. People otherwise might send a while product of the API into a form which intends to work on two fields only.

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,540 @@
...
+
+</Info>
+
+# Building a form page
+
+Let's apply what we just learned in our `Channels` application, as we might want to add a page to see and manage a Channel's details.
+
+Given that we might want to allow creating new channels and updating existing ones, there will be 2 different pages with the same form components. Therefore, we can implement the form as a component and re-use it in both the create and details pages. The only difference is that the form for the create page will be initially empty and the form for the details page will have some data.
```
tdeekens, line 233:
```suggestion
Given that we might want to allow creating new channels and updating existing ones, there will be two different pages with the same form components. Therefore, we can implement the form as a component and re-use it in both the create and details pages. The only difference is that the form for the create page will be initially empty and the form for the details page will have some data.
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,540 @@
...
+    </Spacings.Stack>
+  );
+
+  return props.children({
+    formElements,
+    isDirty: formik.dirty,
+    isSubmitting: formik.isSubmitting,
+    submitForm: formik.handleSubmit,
+    handleCancel: formik.handleReset,
+  });
```
tdeekens, line 439: Couldn't this be a hook? `useChannelsForm`? 🤔

website/src/content/development/forms.mdx
emmenko: Yeah, I had `name` initially, but Channel names are localized and I felt like the example would be too complex. Maybe it's good enough, as localized strings are common types in our API.

website/src/content/development/forms.mdx
emmenko: That's not always true. Clicking outside of the input also loses focus.

I think I'll keep it like this for now, if you don't mind.

website/src/content/development/forms.mdx
emmenko: Good point, thanks!

website/src/content/development/forms.mdx
emmenko: Hmm could be, although it's weird to return React elements from a hook.

Also, you can't always use `useFormik`, as using field arrays requires to use the `<Formik>` component.

emmenko:
@tdeekens Thanks for the initial review. I have addressed your suggestions (thanks) and I'll probably improve a couple more things before making it ready for review.

Is there any other particular thing that I missed or are we happy with the points for now?

website/src/content/development/forms.mdx
tdeekens: Yeah, we could have to define the a `FormKeyField` or something as a child to be able to do that.

website/src/content/development/forms.mdx
tdeekens: Yip. Was just thinking out loud case FaaC feels odd these days at times.

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,673 @@
...
+        onSubmit={handleSubmit}
+      />
+    </Spacings.Stack>
+  );
+}
+```
+
+## Using modal pages
+
+Most of the time a create or details page with a form can be implemented using either the [FormModalPage](/api-reference/commercetools-frontend-application-components#formmodalpage) or the [CustomFormModalPage](/api-reference/commercetools-frontend-application-components#customformmodalpage) components.
```
FilPob, line 431: I was pretty confused with the "a create or details page" formulation as I thought "create" was a verb. Now I get it that it's a create page or a details page. But maybe this could be formulated a bit better.

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,673 @@
...
+
+Using these components has the advantage of providing the form control buttons (for example `Cancel` and `Save`) in the correct place, according to our design guidelines.
+
+However, the form component must be defined "outside" of the modal page to be able to pass the necessary functions to the modal page to interact with the form.
+
+Therefore, our Channels form must be refactored to defined all the form elements but return them using the function-as-child component pattern.
```
CarlosCortizasCT, line 437:
```suggestion
Therefore, our Channels form must be refactored to define all the form elements but return them using the function-as-child component pattern.
```

I think there is a "d" that souldn't be there

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,673 @@
...
+type TFieldErrors = Record<string, boolean>;
+// Similar shape of `FormikErrors` but values are `TFieldErrors` objects.
+type TCustomFormErrors<Values> = {
+  [K in keyof Values]?: TFieldErrors;
+};
+
+declare const validate = (values: FormValues) => TCustomFormErrors<FormValues>;
+```
+
+Specifically, the `FormikErrors` object should contain properties correlating to field names with their values being objects with reasons for the given error. Let's take a look at an example:
```
CarlosCortizasCT, line 116: Given the example and that `FormikErrors` is just used in a comment, I guess it would be more clear if we state we're referring to `validate` function result.

```suggestion
Specifically, the object returned from the `validate` function should contain properties correlating to field names with their values being objects with reasons for the given error. Let's take a look at an example:
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,673 @@
...
+
+const ChannelsForm = () => {
+  const { dataLocale, languages } = useApplicationContext((context) => ({
+    dataLocale: context.dataLocale,
+    languages: context.project.languages,
+  }));
+  const formik = useFormik({
+    initialValues: {
+      // A Channel's `name`: https://docs.commercetools.com/api/projects/channels
+      name: LocalizedTextInput.createLocalizedString(languages),
```
CarlosCortizasCT, line 36: This was confusing to me because we're initializing form data but we're not using anything about the channels.
As this is the first example of this page and the `initialValues` is explained more in detail later, I think we maybe could omit it here.

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,673 @@
...
+  const handleSubmit = useCallback(
+    async (formValues) => {
+      const data = formValuesToDoc(formValues);
+      // This would trigger the request, for example a mutation.
+      const result = await updateChannel(data);
+      // If successful, show a notification.
+      // If errored, show an error notification.
+    },
+    [updateChannel]
+  );
```
CarlosCortizasCT, line 409: Since this is a detailed example and all the dependencies are defined, I'm missing where this `updateChannel` function comes from.

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,673 @@
...
+  const { data: channel } = useChannels(match.params.id)
+  const handleSubmit = useCallback(
+    async (formValues) => {
+      const data = formValuesToDoc(formValues);
+      // This would trigger the request, for example a mutation.
+      const result = await updateChannel(data);
+      // If successful, show a notification.
+      // If errored, show an error notification.
+    },
+    [updateChannel]
```
CarlosCortizasCT, line 571: Since this is a detailed example and all the dependencies are defined, I'm missing where this `updateChannel` function comes from.

website/src/content/development/forms.mdx
emmenko: Hmm I can add a comment making it more explicit that it's the initial data but I would keep the code as-is because technically is correct. The point here is also to show that the props you pass to the field related to the form values.

CarlosCortizasCT (approved):
lgtm

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,678 @@
...
+        isDisabled={formik.isSubmitting}
+      />
+    </form>
+  );
+}
+```
+
+# Field and input components
+
+The commercetools UIKit library provides several UI components for working with forms. Most of the time you want to use the field components as they provide all the recommended features for rendering form elements.
```
industrian, line 75:
```suggestion
The commercetools UIKit library provides several UI components for working with forms. Most of the time you should use the field components as they provide all the recommended features for rendering form elements.
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,678 @@
...
+
+![Form field example](/images/form-field.png 'Form field example')
+
+A field component consists of an input element wrapped with other field elements such as label, description, error message, hint, badge, etc.
```
industrian, line 79:
```suggestion
A field component consists of an input element wrapped with other field elements (such as label, description, error message, hint, and badge).
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,678 @@
...
+
+Depending on the use case, you might want to use a date field, a text field, or a select field. In the UIKit library, we provide a variety of components to cover different use cases. For more information about field and input components check their related packages in the [commercetools/ui-kit](https://github.com/commercetools/ui-kit) repository.
```
industrian, line 81:
```suggestion
Depending on the use case, you might want to use a date field, a text field, or a select field. In the UIKit library, there are many components that cover different use cases. For more information about field and input components check their related packages in the [commercetools/ui-kit](https://github.com/commercetools/ui-kit) repository.
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,678 @@
...
+
+<Info>
+
+All UIKit field components have a related input component, for example `<TextField>` -> `<TextInput>`.
+
+</Info>
+
+## Accessibility support
+
+All field and input components have built-in support for accessibility features such as aria labels, keyboard navigation and focus management, error messaging, and more.
```
industrian, line 91:
```suggestion
All field and input components have built-in support for accessibility features such as aria labels, keyboard navigation, focus management, and error messaging.
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,678 @@
...
+
+<Info>
+
+Testing these field components with [React Testing Library](https://github.com/testing-library/react-testing-library) can be easily done using selectors such as `*ByLabelText`, `*ByRole`, etc.
```
industrian, line 95:
```suggestion
Testing these field components with [React Testing Library](https://github.com/testing-library/react-testing-library) can be easily done using selectors such as `*ByLabelText` and `*ByRole`.
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,678 @@
...
+
+</Info>
+
+# Form validation
+
+One of the important parts of forms is validation. It is important to validate constraints of forms such as required fields but also to check additional semantic requirements, for example checking that the value is a valid URL.
```
industrian, line 101:
```suggestion
One of the important parts of forms is validation. It is important to validate constraints of forms such as required fields and to check additional semantic requirements (for example, checking that the value is a valid URL).
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,678 @@
...
+
+Asides from client-side validation, forms can also perform asynchronous validation against an API to ensure data correctness before submitting the form.
```
industrian, line 103:
```suggestion
Aside from client-side validation, forms can also perform asynchronous validation against an API to ensure data correctness before form submission.
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,678 @@
...
+
+For that, Formik allows you to implement a [`validate` function](https://formik.org/docs/api/formik#validate-values-values--formikerrorsvalues--promiseany) that should return an object of errors.
```
industrian, line 105:
```suggestion
For that, Formik allows you to implement a [`validate` function](https://formik.org/docs/api/formik#validate-values-values--formikerrorsvalues--promiseany) that returns an object of errors.
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,678 @@
...
+type TFieldErrors = Record<string, boolean>;
+// Similar shape of `FormikErrors` but values are `TFieldErrors` objects.
+type TCustomFormErrors<Values> = {
+  [K in keyof Values]?: TFieldErrors;
+};
+
+declare const validate = (values: FormValues) => TCustomFormErrors<FormValues>;
+```
+
+Specifically, the object returned from the `validate` function should contain properties correlating to field names with their values being objects with reasons for the given error. Let's take a look at an example:
```
industrian, line 117:
```suggestion
Specifically, the object returned from the `validate` function should contain properties correlating to field names with their values being objects with reasons for the given error. For example:
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,678 @@
...
+    errors.name = { missing: true };
+  }
+
+  return errors;
+}
+```
+
+In this example, we are validating that the `name` field has a required value. If the value is empty (no localized values have been provided), we assign to the `errors.name` property an error object with the error key `missing` set to `true`.
+
+In the field component, you need to make sure to assign the `errors` and `touched` props which will make the field component render an error message, in case one was returned from the `validate` function.
```
industrian, line 135:
```suggestion
In the field component, you must assign the `errors` and `touched` props which make the field component render an error message, in case one was returned from the `validate` function.
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,678 @@
...
+  }
+  touched={formik.touched.name}
+  onChange={formik.handleChange}
+  onBlur={formik.handleBlur}
+/>
+```
+
+<Info>
+
+Error messages are only shown when the `touched` value for the specific field is set to `true`. This happens whenever the user stops interacting with a field (loses `focus`). The point here is that as long as the user is interacting with a field for the first time there is no need to show validation.
```
industrian, line 155:
```suggestion
Error messages are only shown when the `touched` value for the specific field is `true`. This happens whenever the user stops interacting with a field (loses focus). The point here is that as long as the user is interacting with a field for the first time there is no need to show validation.
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,678 @@
...
+export const docToFormValues = (doc, languages) => ({
+  name: LocalizedTextInput.createLocalizedString(languages, doc?.name),
+})
+```
+
+<Info>
+
+Most UI-Kit field and input components expose some static methods to ease data conversion and validation. Make sure to check if these methods are available in either the field or the input component.
+
+For example, to validate if a text field is empty you can use `TextInput.isEmpty()`.
```
industrian, line 227:
```suggestion
For example, to validate if a text field is empty, use `TextInput.isEmpty()`.
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,678 @@
...
+  name: transformLocalizedStringToLocalizedField(
+    LocalizedTextInput.omitEmptyTranslations(formValues.name)
+  ),
+  // ...
+})
+```
+
+<Info>
+
+Most of the time you would have a simple 1:1 mapping. However, we still recommend to use these conversion functions as a best practice and to help decoupling the data transformation logic from the form component.
```
industrian, line 264:
```suggestion
Most of the time you would have simple 1:1 mapping. However, we still recommend using these conversion functions as a best practice and to help decoupling the data transformation logic from the form component.
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,678 @@
...
+
+For instance, the form might only need a couple of fields even though the data object has many mores. By being explicit in the conversion, we ensure that only the necessary data is passed to the form.
```
industrian, line 266:
```suggestion
For instance, the form might only need a couple of fields even though the data object has many more. By being explicit in the conversion, we ensure that only the necessary data is passed to the form.
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,678 @@
...
+
+</Info>
+
+# Building a form page
+
+Let's apply what we just learned in our `Channels` application, as we might want to add a page to see and manage a Channel's details.
```
industrian, line 272:
```suggestion
Let's apply what we just learned in our `Channels` application, as we might want to add a page to view and manage a Channel's details.
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,678 @@
...
+      <ChannelsForm
+        initialValues={docToFormValues(null, languages)}
+        onSubmit={handleSubmit}
+      />
+    </Spacings.Stack>
+  );
+}
+```
+
+On the "details" page, we would need to fetch the data first, then initialize the form with the data.
```
industrian, line 384:
```suggestion
On the "details" page, we need to fetch the data first, then initialize the form with the data.
```

website/src/content/development/forms.mdx
```diff
@@ -0,0 +1,678 @@
...
+    formElements,
+    isDirty: formik.dirty,
+    isSubmitting: formik.isSubmitting,
+    submitForm: formik.handleSubmit,
+    handleCancel: formik.handleReset,
+  });
+}
+```
+
+One of the advantages of splitting up the form fields is the encapsulation of the logic. You might notice that we don't explicitly pass any props to these components. Instead, each component can access the form data that they need from the form context, using Formik's `useField` hook.
```
industrian, line 652:
```suggestion
One of the advantages of splitting up the form fields is to encapsulate the logic. You might notice that we don't explicitly pass any props to these components. Instead, each component can access the form data that they need from the form context, using Formik's `useField` hook.
```

industrian (approved):
Some suggestions for improving the wording of the forms page.

emmenko:
@industrian Thanks a lot for the review! 💯

PR #2503, opened 2022-03-02
feat: add TabHeader component

### LINKED ISSUE #2498: Implement tab components to be used in `TabularModalPage`

Using the `TabularModalPage` component requires to provide some tab components. Currently there are no such components available to be used.

Implement and expose from `@commercetools-frontend/application-components` a tab component that can be used in the `TabularModalPage`.

**Additional context**

[InVision designs](https://commercetools.invisionapp.com/overview/SHIELD-411-Tabs-component-ckyear79f02pq0118hrbjbvhl/screens?sortBy=1&sortOrder=1&viewLayout=2)

kark:
#### Summary

The aim of this PR was to implement a new `TabHeader` component to be used within `TabularModalPage `.
I decided not to change the name of the component used in `MC-frontend`.

closes #2498


packages/application-components/src/components/tab-header/tab-header.tsx
```diff
@@ -0,0 +1,159 @@
...
+          `,
+      ]}
+      data-track-event="click"
+      data-track-component={startCase(props.name)}
+      data-testid={`header-list-item-${props.name}`}
+      onClick={props.isDisabled ? noop : props.onClick}
+    >
+      <LinkWrapper
+        to={props.isDisabled ? null : props.to}
+        // @ts-ignore
```
kark, line 132: Unfortunately the following line causes `Property 'css' does not exist on type 'IntrinsicAttributes & TLinkWrapperProps'.` error. The solution is changing `tsconfig.ts` to:
```
"typeRoots": [
      "@types",
      "@types-extensions",
      "node_modules/@types",
      "@emotion/react/types/css-prop"
    ]
```
but I didn't want to do that without your approval

website-components-playground/gatsby-node.js
```diff
@@ -31,3 +31,21 @@ exports.onCreateBabelConfig = ({ actions }) => {
...
+    redirectInBrowser: true,
+    isPermanent: true,
+  });
+
+  createRedirect({
+    fromPath: `/tabular-modal-page`,
+    toPath: `/tabular-modal-page/tab-one`,
+    redirectInBrowser: true,
+    isPermanent: true,
+  });
```
kark, line 50: This is strange, but `react-router-dom`'s `<Redirect />` doesn't seem to be working in Gatsby (at least I couldn't make it work). That is how I solved. There is a Gatsby plugin to trail the ending slash, but I wasn't sure if this solution is acceptable at all so I didn't add that package at this point

website/src/content/api-reference/commercetools-frontend-application-components.mdx
```diff
@@ -597,6 +598,57 @@ const AccountPage = () => {
   />
   ```
 
+## TabHeader
+
+<Playground urlPath="tabular-modal-page" />
```
kark, line 603: I added the same visual as in the `TabularModalPage` section since it's hard to showcase it on it's own

emmenko (commented):
I left some comments as references for the things we discussed.

Let me know if you need anything else for now.

packages/application-components/src/components/tab-header/tab.styles.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+  `,
+  ...getTabHeaderStateStyles(isActive, isDisabled),
+];
+
+export const getLinkWrapperStyles = (
+  isDisabled: boolean
+): Interpolation<Theme> => [
+  css`
+    font-size: ${vars.fontSizeDefault};
```
kark, line 79: @FilPob I used the default font size here (1rem), which in case of custom applications docs Gatsby app is set in the root element to 13px. I'm not quite sure if that is ok

packages/application-components/src/components/tab-header/tab-header.tsx
```diff
@@ -0,0 +1,159 @@
...
+const pathWithoutSearch = (path: string) => path?.split('?')[0];
+const noop = () => {};
+
+type TLinkWrapperProps = {
+  to?: LinkProps['to'] | null;
+  children: ReactNode;
+};
+
+const LinkWrapper = (props: TLinkWrapperProps) => {
+  const Component = (props.to ? Link : 'a') as ElementType;
```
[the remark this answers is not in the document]
kark: All in all it seems the `as` prop might not be needed after I figured out this wrapper's purpose. I added a comment in the code above.

packages/application-components/src/components/tab-header/tab.styles.ts
FilPob: It is meant to be the H4 with 16px font size

packages/application-components/src/components/tab-header/tab.styles.ts
```diff
@@ -0,0 +1,89 @@
...
+  return combinedStyles;
+};
+
+export const getTabHeaderStyles = (
+  isActive: boolean,
+  isDisabled: boolean
+): SerializedStyles[] => [
+  css`
+    color: ${vars.colorAccent};
+    min-width: 50px;
```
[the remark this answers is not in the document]
FilPob: Is this the min-width of a single Tab?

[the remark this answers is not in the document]
kark: Yes, that's right
Screenshot 2022-03-07 at 15 21 49

packages/application-components/src/components/tab-header/tab.styles.ts
```diff
@@ -0,0 +1,89 @@
...
+];
+
+export const getLinkWrapperStyles = (
+  isDisabled: boolean
+): Interpolation<Theme> => [
+  css`
+    font-size: ${vars.fontSizeDefault};
+    padding-bottom: ${vars.spacingXs};
+    display: block;
+    padding: ${vars.spacingM} ${vars.spacingS} ${vars.spacingM} ${vars.spacingS};
```
emmenko, line 82:
```suggestion
    padding: ${vars.spacingM} ${vars.spacingS};
```

packages/application-components/src/components/tab-header/tab.styles.ts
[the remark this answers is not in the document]
emmenko: I don't think we need this, it was only used in the dummy placeholder.

packages/application-components/src/components/tab-header/tab-header.tsx
```diff
@@ -0,0 +1,128 @@
...
+  /**
+   * Path that the TabHeader directs to when not disabled.
+   */
+  to: string | LocationDescriptor;
+  /**
+   * Any React node passed to TabHeader.
+   * * <br />
+   * Required if `intlMessage` is not provided.
+   */
+  children?: ReactNode;
```
emmenko, line 47: On second thoughts, I think we don't even need `children` as I don't see a use case where you need to render anything other than a string.
I would suggest to add instead a new prop `label?: string`.

So you essentially pass a string label to either `label` or as `intlMessage` (when using translations). We don't allow any other React element.

The `name` prop then can also be removed, as we can then simply use the string value of label (or intlMessage).

website-components-playground/src/pages/tabular-modal-page.js
```diff
@@ -50,105 +40,131 @@ const exampleCustomTitleRow = (
...
+                    </>
+                  }
+                  hideControls={values.hideControls}
+                >
+                  <Switch>
+                    <Route
+                      path={[
+                        '/custom-applications/playground/tabular-modal-page/', // production
+                        '/tabular-modal-page/', // development
+                      ]}
```
emmenko, line 148: Instead of this, I think we should be able to use Gatsby's `withPrefix` function. https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-link/#add-the-path-prefix-to-paths-using-withprefix

```suggestion
                      path={withPrefix('/tabular-modal-page/')}
```

website-components-playground/src/pages/tabular-modal-page.js
```diff
@@ -50,105 +40,131 @@ const exampleCustomTitleRow = (
...
+                    >
+                      <Redirect to="/tab-one" />
+                    </Route>
+                    <Route path="/tab-one">
+                      <Text.Body>{values['tab-one-content']}</Text.Body>
+                    </Route>
+                    <Route path="/tab-two">
```
emmenko, line 155: These paths should be updated as well.

website/src/content/api-reference/commercetools-frontend-application-components.mdx
```diff
@@ -531,7 +532,7 @@ const AccountPage = () => {
       isOpen={tabsModalState.isModalOpen}
       onClose={tabsModalState.closeModal}
       tabControls={
-        // Tabs components
+        // TabHeader components
```
emmenko, line 535: Now that we have the component, we should use it in the examples.

website/src/content/api-reference/commercetools-frontend-application-components.mdx
```diff
@@ -597,6 +598,44 @@ const AccountPage = () => {
   />
   ```
 
+## TabHeader
```
emmenko, line 601: I'm not sure I would document this separately. It should be enough for now to document it within `TabularModalPage` component context.

emmenko (commented):
Looking better now. I left some suggestions to simplify and improve things a bit more.

packages/application-components/package.json
```diff
@@ -46,26 +46,32 @@
...
+    "@types/history": "4.7.11",
     "@types/lodash": "^4.14.178",
+    "@types/lodash.startcase": "^4.4.6",
     "@types/prop-types": "^15.7.4",
     "@types/react": "^17.0.38",
     "@types/react-dom": "^17.0.11",
     "@types/react-modal": "^3.13.1",
+    "history": "4.10.1",
     "lodash": "4.17.21",
+    "lodash.startcase": "^4.4.0",
```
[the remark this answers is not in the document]
kark: 🤦‍♂️

packages/application-components/src/components/tab-list/tab-list.tsx
```diff
@@ -0,0 +1,29 @@
...
+  orientation?: 'horizontal' | 'vertical';
+};
+
+export const TabList = (props: TTabListProps) => {
+  return (
+    <div role="tablist" aria-orientation={props.orientation}>
+      {props.children}
+    </div>
+  );
+};
```
emmenko, line 20: We shouldn't need this component. I meant to add the `role` to the `TabularModalPage` itself

https://github.com/commercetools/merchant-center-application-kit/blob/23b39a40ae671e4e49288f038f65c8b14ac3fbbd/packages/application-components/src/components/modal-pages/tabular-modal-page/tabular-modal-page.tsx#L88-L94

packages/application-components/src/components/tab-list/tab-list.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export type TTabListProps = {
+  /**
+   * Any React node passed to TabHeader.
+   */
+  children: ReactNode;
+  /**
+   * Sets `aria-orientation` attribute. Defaults to `horizontal`
+   */
+  orientation?: 'horizontal' | 'vertical';
```
emmenko, line 11: Why do we need this? The `tablist` has the default horizontal orientation already.

emmenko (commented):
One small remark, otherwise this looks already much better now. Thanks

packages/application-components/src/components/tab-list/tab-list.tsx
```diff
@@ -0,0 +1,29 @@
...
+};
+
+export const TabList = (props: TTabListProps) => {
+  return (
+    <div role="tablist" aria-orientation={props.orientation}>
```
kark, line 16: Would you consider all the logic with setting `aria-orientation` useful or should I hardcode it to `horizontal`?

packages/application-components/src/components/tab-list/tab-list.tsx
kark: That is true, however other design system libraries I checked set this explicitly, so I included it. I will delete it.

website-components-playground/src/pages/tabular-modal-page.js
kark: I'm not sure if I got it right but when I changed this to `path={withPrefix("/tab-one")}` it stopped working for production builds

emmenko: I think the path should be like `withPrefix('/tabular-modal-page/tab-one')`

website-components-playground/src/pages/tabular-modal-page.js
```diff
@@ -50,105 +41,126 @@ const exampleCustomTitleRow = (
...
+                  }
+                  tabControls={
+                    <>
+                      <TabHeader to="/tab-one" label="Tab One" />
+                      <TabHeader to="/tab-two" label="Tab Two" />
+                      <TabHeader
+                        to="/tab-three"
+                        label="Disabled tab"
+                        isDisabled
+                      />
```
emmenko, line 118: I think these should be updated as well

```suggestion
                      <TabHeader to={withPrefix('/tabular-modal-page/tab-one')} label="Tab One" />
                      <TabHeader to={withPrefix('/tabular-modal-page/tab-two')} label="Tab Two" />
                      <TabHeader
                        to={withPrefix('/tabular-modal-page/tab-three')}
                        label="Disabled tab"
                        isDisabled
                      />
```

website/src/content/api-reference/commercetools-frontend-application-components.mdx
```diff
@@ -561,7 +566,7 @@ const AccountPage = () => {
...
-| `formControls`            | `node`             |    -     | -                        | Pass a React.node with form controls. You can use the pre-made form buttons exposed by this component or you can use your own.                                                                                                                                                                                                                                                                                                                                                                                      |
+| `formControls`                  | `node`                     |    -     | -                        | Pass a React.node with form controls. You can use a composition of `TabHeader` components or you can use your own.                                                                                                                                                                                                                                                                                                                                                                                           |
```
emmenko, line 569: I think you wanted to update the `tabControls` prop instead?

packages/application-components/src/components/tab-header/tab-header.tsx
emmenko: Do you still get an TS error now that the wrapper is gone?

packages/application-components/src/components/tab-header/tab-header.tsx
```diff
@@ -0,0 +1,114 @@
...
+      strict: false,
+    })
+  );
+  const isDisabled = Boolean(props.isDisabled);
+
+  const dataAttributeProps = {
+    'data-track-event': 'click',
+    ...(props.label && {
+      'data-track-component': startCase(props.label),
+    }),
```
emmenko, line 74: Since you can also use the `intlMessage`, I would suggest to compute the label value first.

```ts
warnIfMissingContent(props);

const label = intl.formatMessage(props.intlMessage) ?? props.label;

const dataAttributeProps = {
  'data-track-event': 'click',
  ...(props.label && {
    'data-track-component': startCase(label),
  }),
}
```

and then render the actual `label` below (no need to use the `<FormattedMessage>` component).

emmenko:
@kark Can you check the VRTs? It seems that there is no "active" tab selected. Something to do with routing maybe?

If you need help we can take a look at it together.

kark:
> @kark Can you check the VRTs? It seems that there is no "active" tab selected. Something to do with routing maybe?
> 
> If you need help we can take a look at it together.

At this point I really don't know 😕 both prod and dev build seem to be working locally and when I download the html file from percy it is "active"
https://percy.io/api/v1/snapshots/922689683/assets/head.html
So any help would be most welcome

packages/application-components/src/components/modal-pages/tabular-modal-page/tabular-modal-page.tsx
```diff
@@ -71,7 +71,6 @@ const TabularModalPage = (props: Props) => (
       css={css`
         background-color: ${customProperties.colorNeutral95};
         padding: ${customProperties.spacingM} ${customProperties.spacingM} 0;
-        border-bottom: 1px ${customProperties.colorNeutral} solid;
```
emmenko, line 74: Why was the border removed?

packages/application-components/src/components/tab-header/tab-header.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+    screen.getByText(/tab one/i);
+  });
+  it('should render when "to" is of LocationProvider type', () => {
+    renderTabHeader({ to: { pathname: '/tab-one' } });
+
+    screen.getByText(/tab one/i);
+  });
+  it('should display a custom intl message instead of a label', () => {
+    const intlMessage = { id: 'localized', defaultMessage: 'localized text' };
+    renderTabHeader({ intlMessage });
```
emmenko, line 28: In this case the `label` prop should not be passed.

packages/application-components/src/components/tab-header/tab-header.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+    screen.getByText('localized text');
+  });
+});
+describe('navigation', () => {
+  it('should navigate to link when clicked', async () => {
+    const { history } = renderTabHeader();
+
+    fireEvent.click(screen.getByRole('link', { name: /tab one/i }));
```
emmenko, line 36: Maybe use `tab` as the role?

packages/application-components/src/components/tab-header/tab-header.tsx
```diff
@@ -0,0 +1,114 @@
...
+const getDisabledTabHeaderAriaAttributes = (
+  isDisabled: TTabHeaderProps['isDisabled']
+) => (isDisabled ? { 'aria-disabled': true } : {});
+
+const getDisabledLinkAtributes = (isDisabled: TTabHeaderProps['isDisabled']) =>
+  isDisabled ? { tabIndex: -1 } : {};
+
+export type TTabHeaderProps = {
+  /**
+   * Path that the TabHeader directs to when not disabled.
```
emmenko, line 31:
```suggestion
   * A route path to redirect to when the tab is clicked.
```

packages/application-components/src/components/tab-header/tab-header.tsx
```diff
@@ -0,0 +1,114 @@
...
+   */
+  to: string | LocationDescriptor;
+  /**
+   * A label for the TabHeader.
```
emmenko, line 35:
```suggestion
   * The label of the tab.
```

packages/application-components/src/components/modal-pages/tabular-modal-page/tabular-modal-page.tsx
kark: @FilPob asked me to do so 🙂

packages/application-components/src/components/tab-header/tab-header.tsx
```diff
@@ -0,0 +1,114 @@
...
+   * * <br />
+   * Required if `intlMessage` is not provided.
+   */
+  label?: string;
+  /**
+   * An `intl` message object that will be rendered with `FormattedMessage`.
```
emmenko, line 41:
```suggestion
   * The label of the tab, using an `intl` message object.
```

packages/application-components/src/components/tab-header/tab-header.tsx
```diff
@@ -0,0 +1,114 @@
...
+    values?: Record<string, ReactNode>;
+  };
+  /**
+   * If `true`, indicates that the element is in a disabled state.
+   */
+  isDisabled?: boolean;
+  /**
+   * If set to `true` TabHeader will be visible as active only upon exact match of route.
+   */
+  exact?: boolean;
```
emmenko, line 55: Nit: maybe we should give it a more specific name, like `exactPathMatch`

packages/application-components/src/components/tab-header/tab-header.tsx
```diff
@@ -0,0 +1,114 @@
...
+   */
+  intlMessage?: MessageDescriptor & {
+    values?: Record<string, ReactNode>;
+  };
+  /**
+   * If `true`, indicates that the element is in a disabled state.
+   */
+  isDisabled?: boolean;
+  /**
+   * If set to `true` TabHeader will be visible as active only upon exact match of route.
```
emmenko, line 53:
```suggestion
   * If `true`, marks the tab as active if the link matches exactly the route.
```

packages/application-components/src/components/tab-header/tab-header.tsx
```diff
@@ -0,0 +1,114 @@
...
+      </Link>
+    </div>
+  );
+};
+
+TabHeader.displayName = 'TabHeader';
+
+const defaultProps: Pick<TTabHeaderProps, 'isDisabled' | 'exact'> = {
+  isDisabled: false,
+  exact: false,
```
emmenko, line 110: With default props, the prop itself can be marked as required.

Then you don't need to do things like `Boolean(props.isDisabled)`

website-components-playground/src/pages/tabular-modal-page.js
```diff
@@ -50,105 +41,134 @@ const exampleCustomTitleRow = (
...
-        },
-      ]}
-    >
-      {({ values }) => (
-        <ModalController
-          title={values.title}
-          buttonLabel="Open Tabular Modal Page"
-          containerId={containerId}
+const isBrowser = typeof window !== 'undefined';
+const Router = isBrowser ? BrowserRouter : StaticRouter;
```
emmenko, line 45: Is this is necessary for SSR? If so, maybe you can leave a comment explaining the reason.

emmenko (commented):
Nice, looking very good now! 🙌

I left a couple of suggestions to improve some minor things.

packages/application-components/src/components/tab-header/tab-header.spec.tsx
kark: How about leaving it to show that `intlMessage` takes precedence over `label` when both are passed and adding another test with `intlMessage` and without passing the `label` prop?

emmenko: Sure, that works too. Thanks

packages/application-components/src/components/tab-header/tab-header.spec.tsx
kark: Unfortunately `fireEvent.click(screen.getByRole('tab'));` doesn't seem to work currently (doesn't trigger the underlaying `<Link>`'s action). 
We would need to set the `<Link>`'s role to `tab` to make it work.
Are you OK with that change?

packages/application-components/src/components/modal-pages/tabular-modal-page/tabular-modal-page.tsx
kark: Sorry for that, I wasn't aware of the consequences. Reverted.

packages/application-components/src/components/tab-header/tab-header.tsx
```diff
@@ -0,0 +1,114 @@
...
+  if (props.intlMessage) {
+    label = intl.formatMessage(props.intlMessage);
+  }
+
+  const dataAttributeProps = {
+    'data-track-event': 'click',
+    ...(props.label && {
+      'data-track-component': startCase(label),
+    }),
+  };
```
CarlosCortizasCT, line 81: Just wanted to double check if is it ok to only add the `data-track-component` attribute when the consumer used the `label` property.
Perhaps we could also check, in case the consumer used the `intlMessage` prop, if `id` attribute exists in it.

packages/application-components/src/components/tab-header/tab-header.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  warnIfMissingContent(props);
```
CarlosCortizasCT, line 83: nit: just a personal opinion: I understand extracting to a helper function the warnings in case we need to add more in the future, but in the ui-kit we  don't extract it. Maybe we could discuss if we want to have a _best practise_ about this at some point.

.changeset/olive-cheetahs-boil.md
```diff
@@ -0,0 +1,20 @@
+---
+'@commercetools-frontend/application-components': minor
+'@commercetools-local/visual-testing-app': minor
+'@commercetools-website/custom-applications': minor
+'@commercetools-website/components-playground': minor
+---
+
+Add `TabHeader` component. A composition of those components is meant to be used in `Tabular Modal Page` passed as `tabControls` prop, e.g.
```
emmenko, line 8:
```suggestion
Add a new `<TabHeader>` component.

A composition of those components is meant to be used in the `<TabularModalPage>` component passed as `tabControls` prop. See example below:
```

packages/application-components/src/components/tab-header/tab-header.tsx
emmenko: This was meant to be using the `label` variable. However, the `label` itself can be technically `undefined` even though we do log a warning if that's the case.

```suggestion
  const dataAttributeProps = {
    'data-track-event': 'click',
    ...(label && {
      'data-track-component': startCase(label),
    }),
  };
```

packages/application-components/src/components/tab-header/tab-header.tsx
```diff
@@ -0,0 +1,114 @@
...
+    'TabHeader: one of either `label` or `intlMessage` is required but their values are `undefined`'
+  );
+};
+
+const getDisabledTabHeaderAriaAttributes = (
+  isDisabled: TTabHeaderProps['isDisabled']
+) => (isDisabled ? { 'aria-disabled': true } : {});
+
+const getDisabledLinkAtributes = (isDisabled: TTabHeaderProps['isDisabled']) =>
+  isDisabled ? { tabIndex: -1 } : {};
```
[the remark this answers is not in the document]
kark: Thanks, exactly as you pointed out, this was prettier's default formatting

packages/application-components/src/components/tab-header/tab-header.tsx
kark: Thanks, I wasn't sure if the consensus is to change it or not, so I left it as it was considering that it is a pattern from the ui-kit (in link, content-notification and text components). But I am up for changing that if needed.

emmenko (approved):
Awesome, this it now good to go from my side. Great work!

CarlosCortizasCT (approved):
💯

emmenko:
@kark feel free to merge it, I approved the VRTs 🚀

PR #2506, opened 2022-03-07
feat: add new `mc-scripts login` command

Rhotimee:
#### Summary
Implement login command for mc-scripts package


packages/mc-scripts/src/commands/login.js
```diff
@@ -0,0 +1,98 @@
...
+function validateEmail(email) {
+  const re = /\S+@\S+\.\S+/;
+  return re.test(email);
+}
+
+const getEmail = async () => {
+  let email = await read({ prompt: 'Email: ' });
+  if (!validateEmail(email)) {
+    console.log('Please input a valid email');
+    await getEmail();
```
kark, line 19: This is looking good 🙌 One suggestion:
```suggestion
    email = await getEmail();
```

packages/application-config/src/process-config.ts
```diff
@@ -144,6 +144,7 @@ const processConfig = ({
       applicationId,
       applicationName: appConfig.name,
       entryPointUriPath: appConfig.entryPointUriPath,
+      cloudIdentifier: appConfig.cloudIdentifier,
```
emmenko, line 147: We don't need to expose this.

What you want to use from the login command is the `mcApiUrl` value.

packages/mc-scripts/src/bin/cli.js
```diff
@@ -94,6 +94,10 @@ const applicationDirectory = fs.realpathSync(process.cwd());
         proxyCommand(command);
         break;
       }
+      case 'login': {
+        proxyCommand(command);
+        break;
+      }
```
emmenko, line 100: Please also set the `NODE_ENV` variable as in the other commands. I think `production` should be fine.

packages/mc-scripts/src/commands/login.js
```diff
@@ -0,0 +1,98 @@
...
+  const cookieData = response.headers.get('set-cookie');
+  const cookie = cookieData.split(';')[0].split('=')[1];
+  updateCredential(cloudIdentifier, cookie);
+  console.log(
+    `You've successfully logged into the ${cloudIdentifier} environment.`
+  );
+  process.exit(0);
+};
+
+login();
```
emmenko, line 98: Normally we do error handling on the top level function. This means that you always throw errors and let them bubble up.

```suggestion
login().catch(error => {
  console.error(error);
  process.exit(1);
});
```

packages/mc-scripts/src/commands/login.js
```diff
@@ -0,0 +1,98 @@
...
+      const credentials = getUpdatedCredentials(
+        JSON.parse(data),
+        cloudIdentifier,
+        token
+      );
+      fs.writeFileSync(filePath, credentials);
+    }
+  } catch (err) {
+    console.error('An error occurred');
+  }
```
emmenko, line 60: No need for this try-catch. Let the error bubble up

packages/mc-scripts/src/commands/login.js
```diff
@@ -0,0 +1,98 @@
...
+
+const getUpdatedCredentials = (oldCredentials, cloudIdentifier, token) => {
+  oldCredentials[cloudIdentifier] = token;
+  return JSON.stringify(oldCredentials);
+};
+
+const updateCredential = (cloudIdentifier, token) => {
+  const dirpath = `${homedir}/.mcscriptsrc`;
+  const filename = 'credentials.json';
+  const filePath = `${dirpath}/${filename}`;
```
emmenko, line 36: These can be defined statically outside of this function.

packages/mc-scripts/src/commands/login.js
```diff
@@ -0,0 +1,98 @@
...
+const getPassword = async () => {
+  return await read({ prompt: 'Password: ', silent: true });
+};
+
+const getUpdatedCredentials = (oldCredentials, cloudIdentifier, token) => {
+  oldCredentials[cloudIdentifier] = token;
+  return JSON.stringify(oldCredentials);
+};
+
+const updateCredential = (cloudIdentifier, token) => {
```
emmenko, line 33: I think it would be nice to have a "storage" abstraction object, something like `storage.{get,set}`.
So all the logic to write/read from file is handled within that object/module.

packages/mc-scripts/src/commands/login.js
```diff
@@ -0,0 +1,98 @@
...
+
+  try {
+    if (!fs.existsSync(dirpath)) {
+      fs.mkdirSync(dirpath);
+      const credentials = getUpdatedCredentials(
+        DEFAULT_CREDENTIALS,
+        cloudIdentifier,
+        token
+      );
+      console.log({ credentials });
```
emmenko, line 46: Let's not print out the credentials.

packages/mc-scripts/src/commands/login.js
```diff
@@ -0,0 +1,98 @@
...
+      fs.writeFileSync(filePath, credentials);
+    } else {
+      const data = fs.readFileSync(filePath, 'utf8');
+      console.log({ data });
```
emmenko, line 50: Remember to remove this as well.

packages/mc-scripts/src/commands/login.js
```diff
@@ -0,0 +1,98 @@
...
+      fs.writeFileSync(filePath, credentials);
+    }
+  } catch (err) {
+    console.error('An error occurred');
+  }
+};
+
+const login = async () => {
+  const applicationConfig = processConfig({
+    applicationPath: path.join(__dirname, 'fixtures'),
```
emmenko, line 65: ? 🤔

packages/mc-scripts/src/commands/login.js
```diff
@@ -0,0 +1,98 @@
...
+  });
+  const cloudIdentifier = applicationConfig.env.cloudIdentifier.toUpperCase();
```
emmenko, line 67: As mentioned before, we should use the computed `mcApiUrl`

packages/mc-scripts/src/commands/login.js
```diff
@@ -0,0 +1,98 @@
...
+  const email = await getEmail();
+  const password = await getPassword();
+
+  if (!MC_API_URLS[cloudIdentifier]) {
+    console.log('cloudIdentifier in the applicationConfig is invalid');
+    process.exit(0);
+  }
```
emmenko, line 74: No need for this, as the application config already does the validation.

packages/mc-scripts/src/commands/login.js
```diff
@@ -0,0 +1,98 @@
...
+    headers: {
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify({ email, password }),
+  });
+
+  if (response.status !== 200) {
+    console.log('Invalid email or password');
+    process.exit(0);
+  }
```
emmenko, line 87: We should parse the response and throw an error with the error message.

packages/mc-scripts/src/commands/login.js
```diff
@@ -0,0 +1,98 @@
...
+
+  const cookieData = response.headers.get('set-cookie');
+  const cookie = cookieData.split(';')[0].split('=')[1];
```
emmenko, line 90: Nit: I think we can also use the `cookie` library to do the parsing: https://www.npmjs.com/package/cookie

packages/mc-scripts/src/commands/login.js
```diff
@@ -0,0 +1,98 @@
...
+  }
+};
+
+const login = async () => {
+  const applicationConfig = processConfig({
+    applicationPath: path.join(__dirname, 'fixtures'),
+  });
+  const cloudIdentifier = applicationConfig.env.cloudIdentifier.toUpperCase();
+  const email = await getEmail();
+  const password = await getPassword();
```
emmenko, line 69: In case there is already a token saved, we probably don't need to log in again.

We can also check if the token is still valid and only in case it's not, we should log the user in again.

Can also be a follow up, just wanted to share this first.

emmenko (commented):
Thanks for looking into that. I left some early feedback as there are a couple of things to change.

Let me know if we should discuss some of the points in case they are not so clear.

packages/mc-scripts/src/commands/login.js
CarlosCortizasCT: I know there are different ways of implementing such an abstraction but I wanted to leave here a proposal in case it may help:
```
const { existsSync, writeFileSync, readFileSync } = require('fs');

class CredentialsStorage {
  constructor(configFilePath) {
    this.configFilePath = configFilePath;
  }

  _updateConfigFile(configuration) {
    writeFileSync(this.configFilePath, JSON.stringify(configuration), { encoding: 'utf8' });
  }

  _getAllCredentials() {
    if (!existsSync(this.configFilePath)) {
      this._updateConfigFile({ credentials: [] });
    }
    const data = readFileSync(this.configFilePath, { encoding: 'utf8' });
    return JSON.parse(data);
  }

  getCredentials(cloudIdentifier) {
    const allCredentials = this._getAllCredentials();
    return allCredentials[cloudIdentifier];
  }

  storeCredentials(cloudIdentifier, token) {
    const allCredentials = this._getAllCredentials();
    allCredentials[cloudIdentifier] = token;
    this._updateConfigFile(allCredentials);
  }
}

export default CredentialsStorage;
```

Rhotimee:
Hey guys @commercetools/shield-team-fe, 
Thanks for the feedback. I have implemented them.
Kindly take another look.

packages/mc-scripts/src/utils/credentialsStorage.js
```diff
@@ -0,0 +1,54 @@
+const { existsSync, writeFileSync, readFileSync, mkdirSync } = require('fs');
```
tdeekens, line 1: Before we forget: please kebab case filenames. :)

Rhotimee: Ah!.. yes. Thanks, @tdeekens

packages/mc-scripts/src/commands/login.js
```diff
@@ -0,0 +1,52 @@
...
+  console.log(
+    `You've successfully logged into the ${cloudIdentifier} environment.`
+  );
+  process.exit(0);
+};
+
+const login = async () => {
+  const applicationConfig = processConfig();
+  const cloudIdentifier = applicationConfig.env.location;
+  const credentialsStorage = new CredentialsStorage(cloudIdentifier);
```
emmenko, line 31: Nit: in general I would prefer to instantiate such objects on the module level rather than on each function execution.

The storage key can be passed to each instance method (like what you do with e.g. `localStorage`).

While in this specific case it's not a big deal, as this module is executed once anyway, I would still recommend to refactor it to get more in the habit of following these kind of best practices.

packages/mc-scripts/src/commands/login.js
```diff
@@ -0,0 +1,52 @@
...
+    email = await getEmail();
+  }
+  return email;
+};
+
+const getPassword = async () => {
+  return await read({ prompt: 'Password: ', silent: true });
+};
+
+const successfulLoginMessage = (cloudIdentifier) => {
```
emmenko, line 21: I think we need to provide different messages. Therefore I would remove this function and just inline the logs when needed (see other comments).

packages/mc-scripts/src/commands/login.js
```diff
@@ -0,0 +1,52 @@
...
+  process.exit(0);
+};
+
+const login = async () => {
+  const applicationConfig = processConfig();
+  const cloudIdentifier = applicationConfig.env.location;
+  const credentialsStorage = new CredentialsStorage(cloudIdentifier);
+
+  if (credentialsStorage.isSessionValid()) {
+    successfulLoginMessage(cloudIdentifier);
```
emmenko, line 34: We should say that there is already a valid session.
Also we should do an early return.

```suggestion
    console.log(`You already have a valid session for the ${mcApiUrl} environment.`);
    return;
```

packages/mc-scripts/src/commands/login.js
```diff
@@ -0,0 +1,52 @@
...
+
+  const email = await getEmail();
+  const password = await getPassword();
+
+  const cookieData = await authenticator(
+    email,
+    password,
+    applicationConfig.env.mcApiUrl
+  );
+  credentialsStorage.update(cookieData);
```
emmenko, line 45: I think I would prefer to have a "standard" storage API with getters and setters.

```js
credentialsStorage.getToken(mcApiUrl);

credentialsStorage.setToken(mcApiUrl, token);
```

packages/mc-scripts/src/commands/login.js
```diff
@@ -0,0 +1,52 @@
...
+  const credentialsStorage = new CredentialsStorage(cloudIdentifier);
+
+  if (credentialsStorage.isSessionValid()) {
+    successfulLoginMessage(cloudIdentifier);
+  }
+
+  const email = await getEmail();
+  const password = await getPassword();
+
+  const cookieData = await authenticator(
```
emmenko, line 40: I would expect this function to return the parsed token.

```suggestion
  const token = await getToken(
```

Note that the token should probably be an object like

```js
{
  sessionToken: "", // the JWT value
  expiresAt: "", // the cookie expiration date
}
```

packages/mc-scripts/src/commands/login.js
```diff
@@ -0,0 +1,52 @@
...
+    email,
+    password,
+    applicationConfig.env.mcApiUrl
+  );
+  credentialsStorage.update(cookieData);
+  successfulLoginMessage(cloudIdentifier);
```
emmenko, line 46:
```suggestion
  console.log(`Login successful for the ${mcApiUrl} environment.`);
```

packages/mc-scripts/src/utils/auth.js
```diff
@@ -0,0 +1,20 @@
+const fetch = require('node-fetch');
+
+const authenticator = async (email, password, mcApiUrl) => {
```
emmenko, line 3: As mentioned in the other comment, I would name this `getToken`.

Also, we might want to use names arguments

```suggestion
const getToken = async ({ email, password, mcApiUrl }) => {
```

packages/mc-scripts/src/utils/auth.js
```diff
@@ -0,0 +1,20 @@
...
+  const response = await fetch(`${mcApiUrl}/tokens`, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify({ email, password }),
+  });
+
+  if (response.status !== 200) {
```
emmenko, line 12: I think it's better to use `response.ok`

packages/mc-scripts/src/utils/auth.js
```diff
@@ -0,0 +1,20 @@
...
+    const res = await response.json();
```
emmenko, line 13: Nit: note that attempting to parse JSON is not always safe. We can't assume that the response body is always JSON.

Usually we have a function `processResponse` where we resolve the response using `response.text()` and attempt to safely parse it to JSON.

packages/mc-scripts/src/utils/auth.js
```diff
@@ -0,0 +1,20 @@
...
+    throw new Error(res.message);
+  }
+
+  return response.headers.get('set-cookie');
```
emmenko, line 17: As suggested in the other comment, we should parse the cookie and return the token object.

packages/mc-scripts/src/utils/credentials-storage.js
```diff
@@ -0,0 +1,54 @@
+const { existsSync, writeFileSync, readFileSync, mkdirSync } = require('fs');
+const cookie = require('cookie');
+const homedir = require('os').homedir();
+
+const dirPath = `${homedir}/.mcscriptsrc`;
```
emmenko, line 5: I think we can name this `.commercetools`.

packages/mc-scripts/src/utils/credentials-storage.js
```diff
@@ -0,0 +1,54 @@
...
+const filename = 'credentials.json';
```
emmenko, line 6: I think this can be named something like `mc-credentials.json`

emmenko (changes requested):
Alright, I like where this is going. Overall looks already much better.

I left different suggestions to restructure things a bit, as I noticed that things are unnecessarily mixed up with each other. The main purpose is to keep things simple, clean, and with clear separation of concerns.

As always, if you need to clarify things or want to discuss the comments, I'm happy to do so.

packages/mc-scripts/src/utils/credentials-storage.js
```diff
@@ -0,0 +1,54 @@
...
+const filePath = `${dirPath}/${filename}`;
+
+class CredentialsStorage {
+  constructor(cloudIdentifier) {
+    this.cloudIdentifier = cloudIdentifier;
+  }
+
+  _updateCredentialsFile(credentials) {
+    writeFileSync(filePath, JSON.stringify(credentials), { encoding: 'utf8' });
```
kark, line 15: How about stringifying it with formatting like `JSON.stringify(credentials, null, 2)` or `JSON.stringify(credentials, null, 4)`?

Rhotimee: Is there a reason you are suggesting adding extra white spaces?

kark: Just for readability. What do you reckon?

Rhotimee: No one is expected to read the file though. It's just for storing and retrieving the users session data.

kark: Agreed. It was just an idea and only meant for the rare cases when something requires debugging.
Every configuration json file saved in my `~` is formatted.

packages/mc-scripts/src/utils/auth.js
Rhotimee: I renamed this `getSessionData` instead of `getToken`. Since we are returning the expiry and token.

packages/mc-scripts/src/utils/credentials-storage.spec.js
```diff
@@ -0,0 +1,94 @@
...
+    });
+  });
+
+  it('should return token', () => {
+    expect(credentialsStorage.getToken(mcApiUrl)).toBe('hello-world');
+  });
+
+  it('should return true', () => {
+    expect(credentialsStorage.isSessionValid(mcApiUrl)).toBe(true);
+  });
```
emmenko, line 40: Nit: I would combine these into one `it` test, as we're only doing assertions.

```suggestion
  it('should load credentials', () => {
    expect(credentialsStorage.getToken(mcApiUrl)).toBe('hello-world');
    expect(credentialsStorage.isSessionValid(mcApiUrl)).toBe(true);
  });
```

packages/mc-scripts/src/utils/credentials-storage.spec.js
```diff
@@ -0,0 +1,94 @@
...
+    const newSessionData = {
+      sessionToken: 'fizz-buzz',
+      expiresAt: getExpiryDate(36000),
+    };
+    credentialsStorage.setToken(mcApiUrl, newSessionData);
+    expect(credentialsStorage.getToken(mcApiUrl)).toBe('fizz-buzz');
+  });
+});
+
+describe('When cookie is expired', () => {
```
emmenko, line 52:
```suggestion
describe('when session is expired', () => {
```

packages/mc-scripts/src/utils/credentials-storage.spec.js
```diff
@@ -0,0 +1,94 @@
...
+    });
+  });
+
+  it('should not return token', () => {
+    expect(credentialsStorage.getToken(mcApiUrl)).toBe(null);
+  });
+
+  it('should return false', () => {
+    expect(credentialsStorage.isSessionValid(mcApiUrl)).toBe(false);
+  });
```
emmenko, line 70: Same here

```suggestion
  it('should not load credentials', () => {
    expect(credentialsStorage.getToken(mcApiUrl)).toBe(null);
    expect(credentialsStorage.isSessionValid(mcApiUrl)).toBe(false);
  });
```

packages/mc-scripts/src/utils/credentials-storage.spec.js
```diff
@@ -0,0 +1,94 @@
...
+});
+
+describe('When credentials folder is not present', () => {
```
emmenko, line 73:
```suggestion
describe('when credentials file is missing', () => {
```

packages/mc-scripts/src/utils/credentials-storage.spec.js
```diff
@@ -0,0 +1,94 @@
...
+    mock({});
+  });
+
+  it('should not return token', () => {
+    expect(credentialsStorage.getToken(mcApiUrl)).toBe(null);
+  });
+
+  it('should return false', () => {
+    expect(credentialsStorage.isSessionValid(mcApiUrl)).toBe(false);
+  });
```
emmenko, line 84:
```suggestion
  it('should not load credentials', () => {
    expect(credentialsStorage.getToken(mcApiUrl)).toBe(null);
    expect(credentialsStorage.isSessionValid(mcApiUrl)).toBe(false);
  });
```

packages/mc-scripts/src/utils/get-session-data.js
```diff
@@ -0,0 +1,26 @@
...
+    headers: {
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify({ email, password }),
+  });
+
+  if (!response.ok) {
+    const rawResponseError = await response.text();
+    const errorMessage = JSON.parse(rawResponseError)['message'];
+    throw new Error(errorMessage);
```
emmenko, line 16: What I meant was to safely JSON parse, like this:

```suggestion
    const text = await response.text();
    let parsed;
    try {
      parsed = JSON.parse(text);
    } catch (error) {
    }
    const errorMessage = parsed ? parsed.message : text
    throw new Error(errorMessage);
```

packages/mc-scripts/src/utils/get-session-data.spec.js
```diff
@@ -0,0 +1,63 @@
...
+  beforeEach(() => {
+    mockServer.use(
+      rest.post(`${mcApiUrl}/tokens`, (req, res, ctx) => {
+        return res(ctx.status(200), ctx.cookie('mcAccessToken', 'hello-world'));
+      })
+    );
+  });
+  it('should match returned credentials', async () => {
+    const sessionData = await getSessionData({
+      email: 'lorem@commercetools.co',
```
emmenko, line 29:
```suggestion
      email: 'user@email.com',
```

packages/mc-scripts/src/utils/get-session-data.spec.js
```diff
@@ -0,0 +1,63 @@
...
+      password: 'pass',
```
emmenko, line 30:
```suggestion
      password: 'secret',
```

packages/mc-scripts/src/utils/get-session-data.spec.js
```diff
@@ -0,0 +1,63 @@
...
+);
+afterAll(() => mockServer.close());
+
+const mcApiUrl = 'https://mc-api.europe-west1.gcp.commercetools.com';
+
+describe('When login details are correct', () => {
+  beforeEach(() => {
+    mockServer.use(
+      rest.post(`${mcApiUrl}/tokens`, (req, res, ctx) => {
+        return res(ctx.status(200), ctx.cookie('mcAccessToken', 'hello-world'));
```
emmenko, line 23: Does this work? According to the docs, it does not seem like accessing `set-cookie` header works like this 🤔

https://mswjs.io/docs/recipes/cookies#mock-response-cookies

packages/mc-scripts/src/utils/get-session-data.spec.js
```diff
@@ -0,0 +1,63 @@
...
+  });
+  it('should match returned credentials', async () => {
+    const sessionData = await getSessionData({
+      email: 'lorem@commercetools.co',
+      password: 'pass',
+      mcApiUrl,
+    });
+    expect(sessionData).toEqual({
+      sessionToken: 'hello-world',
+      expiresAt: undefined,
```
emmenko, line 35: I would expect that this value is defined

.changeset/stupid-kangaroos-film.md
```diff
@@ -0,0 +1,5 @@
+---
+"@commercetools-frontend/mc-scripts": patch
```
emmenko, line 2:
```suggestion
"@commercetools-frontend/mc-scripts": minor
```

packages/mc-scripts/src/bin/cli.js
```diff
@@ -94,6 +94,12 @@ const applicationDirectory = fs.realpathSync(process.cwd());
         proxyCommand(command);
         break;
       }
+      case 'login': {
```
emmenko, line 97: Please document this in the CLI usage itself (see above L22).

emmenko (commented):
This is looking much better now. I left some small suggestions here and there but overall this looks good. Documentation is also missing a bit.

Thanks

packages/mc-scripts/src/commands/login.js
```diff
@@ -0,0 +1,31 @@
+const { promisify } = require('util');
+const read = promisify(require('read'));
+const { processConfig } = require('@commercetools-frontend/application-config');
+const CredentialsStorage = require('../utils/credentials-storage');
+const getSessionData = require('../utils/get-session-data');
+
+const credentialsStorage = new CredentialsStorage();
+
+const login = async () => {
+  const applicationConfig = processConfig();
```
CarlosCortizasCT, line 10: I was wondering how does the `processConfig` know about the `.env` file in case the user provides one:
```
mc-scripts login --env .env.ctp-gcp-production-eu
```

emmenko: It's implicitly done by the `proxyCommand` on the `cli.js` file

https://github.com/commercetools/merchant-center-application-kit/blob/dd5e33ea58b15e32a2d9aeb1d95cb4a4aec8d069/packages/mc-scripts/src/bin/cli.js#L120-L123

packages/mc-scripts/src/bin/cli.js
```diff
@@ -41,6 +41,8 @@ Commands:
 
   serve                       Serves previously built and compiled application from the "public" folder.
 
+  login                       Authenticate the user to the merchant center.
```
emmenko, line 44:
```suggestion
  login                       Log in to your Merchant Center account through the CLI. The session token is stored in your home directory and will be used by other CLI commands that require a valid session token.
```

.changeset/stupid-kangaroos-film.md
```diff
@@ -1,5 +1,8 @@
 ---
-"@commercetools-frontend/mc-scripts": patch
+'@commercetools-frontend/mc-scripts': minor
 ---
 
-feat: add `mc-scripts login` command
+Add `mc-scripts login` command
+This command enables users to log in to the merchant center service from the command line.
+Upon login, an authentication token is stored on the user's computer's home directory. The token can be used to perform some operations in the merchant center through the CLI.
+If no valid token is available on the computer, the login command requires an email and password to authenticate.
```
emmenko, line 8:
```suggestion
Add a new CLI command `mc-scripts login`.

This command enables users to log in to their Merchant Center account through the CLI. An interactive prompt will be displayed asking the user to enter the login credentials.

Upon login, the user session token is stored in the user's home directory `$HOME/.commercetools/mc-credentials.json`. The session token will be used by other CLI commands that require a valid session token.
```

packages/mc-scripts/src/utils/get-session-data.js
```diff
@@ -11,8 +11,14 @@ const getSessionData = async ({ email, password, mcApiUrl }) => {
...
 
   if (!response.ok) {
-    const rawResponseError = await response.text();
-    const errorMessage = JSON.parse(rawResponseError)['message'];
+    const text = await response.text();
+    let parsed;
+    try {
+      parsed = JSON.parse(text);
+    } catch (error) {
+      throw new Error(error);
```
emmenko, line 19: We shouldn't throw this error.

```suggestion
```

packages/mc-scripts/src/utils/get-session-data.spec.js
```diff
@@ -20,19 +20,25 @@ describe('When login details are correct', () => {
   beforeEach(() => {
     mockServer.use(
       rest.post(`${mcApiUrl}/tokens`, (req, res, ctx) => {
-        return res(ctx.status(200), ctx.cookie('mcAccessToken', 'hello-world'));
+        return res(
+          ctx.status(200),
+          ctx.set(
+            'set-cookie',
+            'mcAccessToken=hello-world; Expires=Tue, 15 Mar 2022 10:01:09 GMT'
```
emmenko, line 27: I would suggest to use `cookie.serialize` to have a realistic cookie value. Something like

```suggestion
            cookie.serialize('mcAccessToken', 'hello-world', { expires: new Date('2022-01-01'), httpOnly: true, domain: 'mc.europe-west1.gcp.commercetools.com' })
```

website/src/content/api-reference/cli.mdx
```diff
@@ -103,6 +103,16 @@ We recommend to set up the following scripts in your `package.json`:
 
 By default the application starts at `http://localhost:3001`.
 
+## login
+
+This command enables users to log in to the merchant center service from the command line. 
+Upon login, an authentication token is stored on the user's computer's home directory. The token can be used to perform some operations in the merchant center through the CLI. 
+If no valid token is available on the computer, the login command requires an email and password to authenticate.
```
emmenko, line 110:
```suggestion
This command enables users to log in to their Merchant Center account through the CLI. An interactive prompt will be displayed asking the user to enter the login credentials.

Upon login, the user session token is stored in the user's home directory `$HOME/.commercetools/mc-credentials.json`. The session token will be used by other CLI commands that require a valid session token.
```

emmenko (commented):
Thanks for adding some docs. I left some suggestions to improve the wording a bit.

packages/mc-scripts/src/utils/credentials-storage.js
```diff
@@ -1,24 +1,35 @@
...
+  static location = credentialsFilePath;
+
+  constructor() {
+    // Ensure the credentials file is present
+    if (!fs.existsSync(credentialsFilePath)) {
+      fs.mkdirSync(credentialsFolderPath, { recursive: true });
+      // Initialize with an empty object
+      this._writeCredentials({});
+    }
+  }
```
emmenko, line 21: There was a small issue when the folder `.commercetools` was present but not the file.

So I decided to move the logic of initializing the file to the constructor.

CarlosCortizasCT (approved):
Great job! 💯

PR #2510, opened 2022-03-12
chore: update dotenv dependencies to v16

emmenko:


tdeekens (approved):
Should be good. Did the same internally.

PR #2517, opened 2022-03-16
feat: add `TabularMainPage` and `TabularDetailPage` components

### LINKED ISSUE #2499: Implement `TabularPage` component

Currently there is a `TabularModalPage` component that renders as a modal page.

In the Merchant Center we also have "normal" pages that use tab navigation, for example a details page or as a "landing" page.

Expose the tabular page as a separate component, which is not rendered using a modal.

**Additional context**

Blocked by #2498

kark:
#### Summary

The aim of this PR was to implement new `TabularMainPage` and `TabularDetailPage` components.
closes #2499


packages/application-components/src/components/internals/page-top-bar.tsx
```diff
@@ -0,0 +1,63 @@
...
+import messages from './messages';
+
+// NOTE: the `MessageDescriptor` type is exposed by `react-intl`.
+// However, we need to explicitly define this otherwise the prop-types babel plugin
+// does not recognize the object shape.
+type MessageDescriptor = {
+  id: string;
+  description?: string | object;
+  defaultMessage?: string;
+};
```
CarlosCortizasCT, line 16: I think we should also add the `values` property to allow for messages that contain parameters as we had to do in ui-kit.

```suggestion
type MessageDescriptor = {
  id: string;
  description?: string | object;
  defaultMessage?: string;
  values?; Record<string, React.ReactNode>
};
```

packages/application-components/src/components/internals/tabular-page.tsx
```diff
@@ -0,0 +1,65 @@
...
+import type { ReactNode } from 'react';
+
+type TTabControlsProps = {
+  children: ReactNode;
+};
+
+const TabControls = (props: TTabControlsProps) => (
+  <div
+    css={css`
+      margin-top: 8px;
```
CarlosCortizasCT, line 12: Perhaps here we could use a css variable instead.

```suggestion
      margin-top: customProperties.spacingS;
```

packages/application-components/src/components/tabular-detail-page/tabular-detail-page.tsx
```diff
@@ -0,0 +1,82 @@
...
+import { ContentWrapper, PageWrapper } from '../internals/page.styles';
+
+// NOTE: the `MessageDescriptor` type is exposed by `react-intl`.
+// However, we need to explicitly define this otherwise the prop-types babel plugin
+// does not recognize the object shape.
+type MessageDescriptor = {
+  id: string;
+  description?: string | object;
+  defaultMessage?: string;
+};
```
CarlosCortizasCT, line 19: I think we should also add the `values` property to allow for messages that contain parameters as we had to do in ui-kit.

```suggestion
type MessageDescriptor = {
  id: string;
  description?: string | object;
  defaultMessage?: string;
  values?; Record<string, React.ReactNode>
};
```

packages/application-components/src/components/tabular-detail-page/tabular-detail-page.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+type TTabularDetailPageProps = {
+  /**
+   * The title of the page.
+   */
+  title: string;
+  /**
+   * The subtitle of the page.
+   */
+  subtitle?: string | ReactElement;
```
CarlosCortizasCT, line 29: nit: maybe we could use `ReactNode` instead as we did in the ui-kit.

packages/application-components/src/components/tabular-main-page/tabular-main-page.tsx
```diff
@@ -0,0 +1,63 @@
...
+  children: ReactNode;
+  /**
+   * A composition of tab components.
+   */
+  tabControls: ReactNode;
+};
+
+const TabularMainPage = (props: TTabularMainPageProps) => (
+  <PageWrapper>
+    <TabularPageContainer color="surface">
```
CarlosCortizasCT, line 38: Maybe we need to allow the consumers to provide the color.
What do you think?

kark: Thanks. Let's ask @FilPob when he's back 🙂

emmenko: This `TabularPageContainer` is only used internally. We discussed that we will have 2 components to cover the 2 use cases: main page and details page. Those have predefined colors but the user does not have to do anything about it.

CarlosCortizasCT: Ok, I think I got something wrong because I understood that we wanted to have a design where the background colors could be inverted.
Default:
Screenshot 2022-03-18 at 10 01 22


Alternative:
Screenshot 2022-03-18 at 10 02 17

Apologize for my mistake.

packages/application-components/src/components/tabular-detail-page/tabular-detail-page.tsx
kark: Thanks. The idea was to keep the API of all page components coherent. Would you be ok with leaving the type as it is?

.changeset/rare-cameras-bow.md
```diff
@@ -0,0 +1,56 @@
+---
+'@commercetools-frontend/application-components': minor
+'@commercetools-local/visual-testing-app': minor
+'@commercetools-website/custom-applications': minor
+'@commercetools-website/components-playground': minor
+---
+
+Add 2 new components: `<TabularMainPage>` and `<TabularDetailPage>` providing navigation through, for instance, `<TabHeader>` components. Both pages, as opposed to `<TabularModalPage>`, are not rendered within a modal.
```
emmenko, line 8: I would also mention where these pages are supposed to be used.

packages/application-components/src/components/tabular-detail-page/tabular-detail-page.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  // PageTopBar props:
+  /**
+   * A return route path label.
+   */
+  previousPathLabel?: string | MessageDescriptor;
+  /**
+   * Function called when back button is pressed.
+   */
+  onClose: (
```
emmenko, line 52: I'm not sure `onClose` is the appropriate name here. I would maybe name this prop `onPreviousPathClick`

packages/application-components/src/components/tabular-detail-page/tabular-detail-page.tsx
```diff
@@ -0,0 +1,83 @@
...
+// However, we need to explicitly define this otherwise the prop-types babel plugin
+// does not recognize the object shape.
+type MessageDescriptor = {
+  id: string;
+  description?: string | object;
+  defaultMessage?: string;
+  values?: Record<string, ReactNode>;
+};
+
+type TTabularDetailPageProps = {
```
emmenko, line 22: We should  also support the `formControls` prop (same as in the TabularModalPage).

packages/application-components/src/components/tabular-detail-page/tabular-detail-page.tsx
CarlosCortizasCT: Both forms are valid and the one proposed its more compact but one could argue that it could be too flexible (I don't see it that way).
I proposed it because we changed this same use case in many PRs form the ui-kit migration to Typescript.

Anyway, if we want to keep it like it is now, it's ok from my side.

packages/application-components/src/components/tabular-detail-page/tabular-detail-page.tsx
kark: updated in cdb034ded

.changeset/rare-cameras-bow.md
kark: Updated in cdb034d. However, I'm not sure if what I wrote is what you had in mind.

CarlosCortizasCT (approved):
lgtm

emmenko:
Overall looks good, thanks!

I was looking at the VRTs though and noticed a couple of things that we need to discuss with @FilPob , from a UX perspective.

packages/application-components/src/components/internals/page-header-title.tsx
```diff
@@ -7,30 +7,34 @@ import styled from '@emotion/styled';
...
   title: string;
   titleSize: 'big' | 'small';
+  truncate: boolean;
   subtitle?: string | ReactElement;
   children?: never;
 };
-const defaultProps: Pick<Props, 'titleSize'> = {
+const defaultProps: Pick<Props, 'titleSize' | 'truncate'> = {
   titleSize: 'small',
+  truncate: true,
```
emmenko, line 16: The default value should be `false`. So when you do need to use it, you can simply pass it as `truncate`. It's the same in the `<Text>` components.

Of course then we need to explicitly set it in the other components that expect it.

emmenko (commented):
Nice, I think we should be good to go now. All main use cases are covered.

Just one little suggestion to invert the default value for `truncate`, otherwise looks good.

emmenko (approved):
🚀

PR #2521, opened 2022-03-22
fix(eslint): provide default BABEL_ENV variable to allow VSCode to parse the Babel config

emmenko:
I noticed that VSCode has difficulties understanding that ESLint uses our Babel preset, even though the Babel preset is referenced to the production config (notice the `/production` import).

https://github.com/commercetools/merchant-center-application-kit/blob/90d5a2308942e232c2f671be182dbb4fbf07a718/packages/eslint-config-mc-app/index.js#L11-L19

This results in VSCode showing this ESLint error:

image

To amend this, we can explicitly define the `BABEL_ENV` variable within the ESLint config, so that VSCode can correctly parse the Babel preset.

### DIFF
--- a/.changeset/pink-steaks-lick.md
+++ b/.changeset/pink-steaks-lick.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/eslint-config-mc-app': patch
+---
+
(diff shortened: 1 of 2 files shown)


CarlosCortizasCT (approved):
lgtm

ddouglasz (approved):

Rhotimee (approved):

PR #2522, opened 2022-03-22
refactor(config): expose transformed custom app data from `processConfig`

emmenko:
This is part of the work for the new `sync` command.


packages/application-config/src/validate-config.ts
```diff
@@ -1,7 +1,5 @@
-import type { ErrorObject } from 'ajv';
 import type { JSONSchemaForCustomApplicationConfigurationFiles } from './schema';
-
-import Ajv from 'ajv';
+import Ajv, { type ErrorObject } from 'ajv';
```
Rhotimee, line 2: NIT: external module imports should come first.

Rhotimee (approved):
💯

PR #2524, opened 2022-03-24
feat: add new `mc-scripts config:sync` command 

Rhotimee:
Implement a new `config:sync` command for `mc-scripts`.

See changelog for more information


packages/mc-scripts/src/commands/config:sync.js
```diff
@@ -0,0 +1,138 @@
...
+    applicationId: fetchedCustomApplication.application.id,
+  });
+  const customAppLink = getMcUrlLink(
+    mcApiUrl,
+    organizationId,
+    fetchedCustomApplication.application.id
+  );
+  console.log(
+    chalk.green(
+      `You have successfully updated the configuration in the "${organizationName}" organization in environment ${location}.\nLink to the created custom application is ${customAppLink}`
```
emmenko, line 130:
```suggestion
      `You have successfully updated the configuration in the "${organizationName}" organization in environment ${location}.\nLink to the created custom application is ${customAppLink}`
      `Custom Application updated.\nYou can see the Custom Application data in the Merchant Center at ${customAppLink}.`
```

emmenko (commented):
Thanks, functionality wise looks good. I left some feedback to improve different minor things.

packages/mc-scripts/src/commands/config:sync.js
```diff
@@ -0,0 +1,138 @@
...
+  } else {
+    organizationId = userOrganizations.results[0].id;
+    organizationName = userOrganizations.results[0].name;
+  }
+
+  if (!fetchedCustomApplication) {
+    const { confirmation } = await prompts({
+      type: 'text',
+      name: 'confirmation',
+      message: `You are about to create the configuration in the "${organizationName}" organization in environment ${location}. Are you sure you want to proceed?`,
```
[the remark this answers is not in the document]
CarlosCortizasCT: Nicola recommendation is great, but bear in mind that perhaps we have an extra "environment" word over there:
> You are about to create a new Custom Application in the "${organizationName}" organization for the **environment** ${mcApiUrl} **environment**. Are you sure you want to proceed?

Maybe we could use only one:
> You are about to create a new Custom Application in the "${organizationName}" organization for the ${mcApiUrl} environment. Are you sure you want to proceed?

packages/mc-scripts/src/commands/config:sync.js
```diff
@@ -0,0 +1,138 @@
+const omit = require('lodash/omit');
+const prompts = require('prompts');
+const chalk = require('react-dev-utils/chalk');
```
CarlosCortizasCT, line 3: Here we are relying in the fact that `react-dev-utils` dependency currently export `chalk`.
Given that we are going to use [that dependency](https://www.npmjs.com/package/chalk) directly I guess it would be safer to explicitly depend on it by itself declaring it in our `package.json`.

packages/mc-scripts/src/commands/config:sync.js
```diff
@@ -0,0 +1,138 @@
...
+
+  if (!fetchedCustomApplication) {
+    const { confirmation } = await prompts({
+      type: 'text',
+      name: 'confirmation',
+      message: `You are about to create the configuration in the "${organizationName}" organization in environment ${location}. Are you sure you want to proceed?`,
+      initial: 'yes',
+    });
+    if (confirmation.toLowerCase().charAt(0) !== 'y') {
+      console.log('Aborted.');
```
CarlosCortizasCT, line 76: Perhaps we could use `chalk` (red?) here to be more explicit with the message.

packages/mc-scripts/src/commands/config:sync.js
```diff
@@ -0,0 +1,138 @@
...
+  // TODO: show diff (followup task)
+
+  const { confirmation } = await prompts({
+    type: 'text',
+    name: 'confirmation',
+    message: `You are about to update the configuration in the "${organizationName}" organization in environment ${location}. Are you sure you want to proceed?`,
+    initial: 'yes',
+  });
+  if (confirmation.toLowerCase().charAt(0) !== 'y') {
+    console.log('Aborted.');
```
CarlosCortizasCT, line 113: Perhaps we could use `chalk` (_red_?) here to be more explicit with the message.

packages/mc-scripts/src/utils/update-application-id-in-custom-application-config.js
```diff
@@ -0,0 +1,42 @@
+const fs = require('fs');
+const rcfile = require('rcfile');
+const prettier = require('prettier');
+const babel = require('@babel/core');
+const {
+  getCustomApplicationConfigPath,
+} = require('@commercetools-frontend/application-config');
+
+function updateApplicationIdInCustomApplicationConfig(applicationId) {
```
CarlosCortizasCT, line 9: I was wondering if we could maybe just update the applicationId using a regular expression to make this process simpler.

Maybe something like:
```
configFileContents.replace(/("env":.*"production":.*"applicationId":\s?)"(\w*)"/, '$1"[REPLACER]"')
```

(_perhaps this is more complex than I can see_)

Rhotimee:
@emmenko @CarlosCortizasCT, thanks for the feedback. All feedback above has been Implemented. Kindly take another look.

.changeset/tidy-zoos-applaud.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-frontend/application-config': minor
+---
+
+Implement and expose getConfigPath
```
emmenko, line 5:
```suggestion
Expose a new function `getConfigPath` to return the absolute file path of the Custom Application config file.
```

packages/mc-scripts/src/commands/config:sync.js
```diff
@@ -0,0 +1,136 @@
...
+        initial: 0,
+      })['organizationId'];
+
+      organizationName = organizationChoices.find(
+        ({ value }) => value === organizationId
+      ).title;
+    } else {
+      organizationId = userOrganizations.results[0].id;
+      organizationName = userOrganizations.results[0].name;
+    }
```
emmenko, line 66: I would also check if there are no organizations. In this case we need to throw an error.

And maybe invert the condition.
```suggestion
    if (userOrganizations.total === 0) {
      throw new Error(`It seems you are not an admin of any Organization. Please make sure to be part of the Administrators team of the Organization you want the Custom Application to be configured to.`);
    }
    if (userOrganizations.total === 1) {
      organizationId = userOrganizations.results[0].id;
      organizationName = userOrganizations.results[0].name;
    } else {
      const organizationChoices = userOrganizations.results.map(
        (organization) => ({
          title: organization.name,
          value: organization.id,
        })
      );

      organizationId = await prompts({
        type: 'select',
        name: 'organizationId',
        message: 'Select Organization',
        choices: organizationChoices,
        initial: 0,
      })['organizationId'];

      organizationName = organizationChoices.find(
        ({ value }) => value === organizationId
      ).title;
    }
```

packages/mc-scripts/src/utils/graphql-requests.js
```diff
@@ -0,0 +1,113 @@
...
+const FetchMyOrganizationsFromCli = requireGraphql(
+  './fetch-user-organizations.core.graphql'
+);
+
+const graphQLClient = (uri, token, target = GRAPHQL_TARGETS.SETTINGS_SERVICE) =>
+  new GraphQLClient(`${uri}/graphql`, {
+    headers: {
+      'x-graphql-target': target,
+      'x-mc-cli-access-token': token,
+      'x-user-agent': userAgent,
```
emmenko, line 25:
```suggestion
      Accept: 'application/json', 
      'Content-Type': 'application/json',
      'x-graphql-target': target,
      'x-mc-cli-access-token': token,
      'x-user-agent': userAgent,
```

packages/mc-scripts/src/utils/graphql-requests.js
```diff
@@ -0,0 +1,113 @@
+const { GraphQLClient } = require('graphql-request');
+const { GRAPHQL_TARGETS } = require('@commercetools-frontend/constants');
```
emmenko, line 2: Please define this dependency in the `package.json`.

packages/mc-scripts/src/utils/update-application-id-in-custom-application-config.js
```diff
@@ -0,0 +1,53 @@
...
+const { getConfigPath } = require('@commercetools-frontend/application-config');
+
+function updateApplicationIdInCustomApplicationConfig(applicationId) {
+  const filePath = getConfigPath();
+  if (filePath.endsWith('.json')) {
+    const customApplicationConfig = require(filePath);
+    customApplicationConfig.env.production.applicationId = applicationId;
+    fs.writeFileSync(
+      filePath,
+      JSON.stringify(customApplicationConfig, null, 2),
```
emmenko, line 14: Let's format this with Prettier as well.

website/src/content/api-reference/cli.mdx
```diff
@@ -120,6 +120,23 @@ The command uses the [cloud Region](/concepts/merchant-center-api#cloud-regions)
...
 
+<!-- ## config:sync
+
+This command allows users to synchronize the local Custom Application config with the Merchant Center. The sync implies that a new Custom Application will be created or an existing one will be updated.
+
+Developers can use the `config:sync` command to automatically manage the configuration of a Custom Application from the config file instead of having to manually fill out the information in the Merchant Center.
+
+If a new Custom Application needs to be created, an interactive prompt will ask the user to select the Organization where the Custom Application should be configured to.
+
+Additionally, the Custom Application ID is automatically synced when running the `config:sync` command.
```
emmenko, line 131:
```suggestion
Additionally, the [Custom Application ID](/api-reference/application-config#envproductionapplicationid) is automatically synced when running the `config:sync` command.
```

website/src/content/api-reference/cli.mdx
```diff
@@ -120,6 +120,23 @@ The command uses the [cloud Region](/concepts/merchant-center-api#cloud-regions)
...
+
+Note that this command requires a valid API token. You can get one by using the `mc-scripts login` command.
+
+```console
+mc-scripts config:sync
+```
+-->
```
emmenko, line 138:
```suggestion

<Info>
The command uses the [cloud Region](/concepts/merchant-center-api#cloud-regions) environment information from the Custom Application config.
</Info>
-->
```

emmenko (commented):
Awesome, looks much better now thanks!

Just a few minor adjustments and then we're good to go 🚀

packages/mc-scripts/src/commands/config:sync.js
```diff
@@ -0,0 +1,136 @@
...
+          `Custom Application created.\nThe "applicationId" in your local Custom Application config file has been updated with the application ID.\nYou can see the Custom Application data in the Merchant Center at ${customAppLink}.`
+        )
+      );
+    }
+    return;
+  }
+  // update applicationID in the custom-application-config file
+  updateApplicationIdInCustomApplicationConfig(
+    fetchedCustomApplication.application.id
+  );
```
CarlosCortizasCT, line 101: Is it actually possible for the ID of a Custom Application to change after its creation?

tdeekens: Not to my knowledge. It's a Prisma generated cuid.

Rhotimee: No. But, it is possible that the user created the custom app on the merchant center and not with the cli.

Rhotimee: The function is to update the local custom-application-config with the applicationId.

CarlosCortizasCT (approved):
lgtm

packages/mc-scripts/src/commands/config:sync.js
emmenko: Speaking of the `applicationId`, we need to consider this use case: the config might use an env placeholder for the application id, e.g. `${env:APPLICATION_ID}`.

In this case, we can't really replace the value.

I suppose we should only do that when we create a new Custom App. When updating, we shouldn't rewrite the value.

@Rhotimee Could you adjust the logic? Thanks

Rhotimee: sure 👍🏽

emmenko (approved):
💯 🚀

packages/mc-scripts/src/commands/config:sync.js
```diff
@@ -0,0 +1,175 @@
...
+      ).title;
+    }
+
+    const { confirmation } = await prompts({
+      type: 'text',
+      name: 'confirmation',
+      message: `You are about to create a new Custom Application in the "${organizationName}" organization for the ${mcApiUrl} environment. Are you sure you want to proceed?`,
+      initial: 'yes',
+    });
+    if (confirmation.toLowerCase().charAt(0) !== 'y') {
```
kark, line 92: How about:
```suggestion
    if (if (!confirmation || confirmation.toLowerCase().charAt(0) !== 'y') {
```

Rhotimee: Updated. Thanks for this.

kark (approved):
Awesome work 👏

PR #2525, opened 2022-03-24
refactor: use Vite for VRT app

emmenko:
Same as https://github.com/commercetools/ui-kit/pull/2153


visual-testing-app/src/application.tsx
```diff
@@ -1,72 +1,55 @@
-import { ComponentType, useEffect } from 'react';
-import { BrowserRouter as Router, Route, Switch } from 'react-router-dom';
+/// <reference types="vite/types/importMeta" />
```
CarlosCortizasCT, line 1: Is this line needed to use the `import.meta.globEager` function?

emmenko: Yes correct

visual-testing-app/src/application.tsx
```diff
@@ -1,72 +1,55 @@
-import { ComponentType, useEffect } from 'react';
-import { BrowserRouter as Router, Route, Switch } from 'react-router-dom';
+/// <reference path="../../node_modules/vite/types/importMeta.d.ts" />
```
emmenko, line 1: FYI: we can also use

```js
/// <reference types="vite/client" />
```

But it contains additional declarations like for `*.svg` that conflicts with what we also use in the repo. So I opted to only import the type declarations for `import.meta`, but we need to use a relative import path for that.

CarlosCortizasCT (approved):
lgtm

PR #2526, opened 2022-03-25
refactor(eslint-mc-app): merge CRA config to avoid possible ESLint merge errors

emmenko:
Recently we've been seeing this kind of errors when we had to update dependencies.

image

To try avoiding this as much as possible, I thought it would make sense to remove the `eslint-config-react-app` dependency and include most of its configuration into our own config.

So most of the things are copied over.

### DIFF
--- a/packages/eslint-config-mc-app/helpers/rules-presets.js
+++ b/packages/eslint-config-mc-app/helpers/rules-presets.js
@@ -1,0 +1,293 @@
+/**
+ * This file contains rules copied from other libraries, to make it easier to differentiate with
+ * our own rules and to better maintain them.
+ */
+
+// The ESLint browser environment defines all browser globals as valid,
+// even though most people don't know some of them exist (e.g. `name` or `status`).
+// This is dangerous as it hides accidentally undefined variables.
+// We deny the globals that we deem potentially confusing.
+// To use them, explicitly reference them, e.g. `window.name` or `window.status`.
+const restrictedGlobals = require('confusing-browser-globals');
+
+const { statusCode } = require('./eslint');
+
+// NOTE: When adding rules here, you need to make sure they are compatible with
+// `typescript-eslint`, as some rules such as `no-array-constructor` aren't compatible.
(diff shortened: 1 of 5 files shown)


packages/eslint-config-mc-app/helpers/rules-presets.js
```diff
@@ -0,0 +1,293 @@
...
+ * This file contains rules copied from other libraries, to make it easier to differentiate with
+ * our own rules and to better maintain them.
+ */
+
+// The ESLint browser environment defines all browser globals as valid,
+// even though most people don't know some of them exist (e.g. `name` or `status`).
+// This is dangerous as it hides accidentally undefined variables.
+// We blacklist the globals that we deem potentially confusing.
+// To use them, explicitly reference them, e.g. `window.name` or `window.status`.
+const restrictedGlobals = require('confusing-browser-globals');
```
CarlosCortizasCT, line 11: 💯

CarlosCortizasCT (approved):
👍

packages/eslint-config-mc-app/helpers/rules-presets.js
```diff
@@ -0,0 +1,293 @@
+/**
+ * This file contains rules copied from other libraries, to make it easier to differentiate with
+ * our own rules and to better maintain them.
+ */
+
+// The ESLint browser environment defines all browser globals as valid,
+// even though most people don't know some of them exist (e.g. `name` or `status`).
+// This is dangerous as it hides accidentally undefined variables.
+// We blacklist the globals that we deem potentially confusing.
```
tdeekens, line 9: We enforce a deny list of the globals...

tdeekens (approved):
Good idea. Copy paste is often underrated :-). 

Just wondering if we want to periodically sync with upstream?

packages/eslint-config-mc-app/helpers/rules-presets.js
emmenko: Yeah thanks, noticed that too.

PR #2527, opened 2022-03-27
feat: add new `eslint-config-node` for Node.js projects

emmenko:
This is a slightly simpler/different ESLint config for Node.js projects. We can use this in our backend services.

### DIFF
--- a/packages-backend/eslint-config-node/LICENSE
+++ b/packages-backend/eslint-config-node/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 commercetools GmbH
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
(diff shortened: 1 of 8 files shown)


packages-backend/eslint-config-node/index.js
```diff
@@ -0,0 +1,222 @@
...
+    /**
+     * @type {import('@babel/core').TransformOptions}
+     */
+    babelOptions: {
+      presets: [['@babel/preset-env', { targets: { node: 'current' } }]],
+    },
+  },
+
+  env: {
+    browser: false,
```
tdeekens, line 27: 💯

packages-backend/eslint-config-node/index.js
```diff
@@ -0,0 +1,222 @@
...
+  parser: '@babel/eslint-parser',
+
+  parserOptions: {
+    sourceType: 'module',
+    requireConfigFile: false,
+    /**
+     * @type {import('@babel/core').TransformOptions}
+     */
+    babelOptions: {
+      presets: [['@babel/preset-env', { targets: { node: 'current' } }]],
```
tdeekens, line 22: Should we maybe allow passing this as an option to not force anybody using this config to use the current version of Node.js? 🤔

emmenko: We can make it configurable yes.

emmenko: Hmm actually I'm not sure how to make it configurable, it's not like Babel.

Btw, the `current` node version is the same as `process.versions.node`. This respects the active node version, like when using `.nvmrc` file.

image

I think it should be fine like this. Or do you have a more specific use case in mind?

tdeekens: Ah okay. I understood differently what `current` means. If it's `process.env.node` it should be fine. As long as people respect the nvmrc.

emmenko: Yes that's what it says in the Babel docs: https://babeljs.io/docs/en/options#targetsnode

image

tdeekens: Sounds good. I think in internal services I just wanted to be more careful when it's around building.

packages-backend/eslint-config-node/index.js
```diff
@@ -0,0 +1,222 @@
+// This is a workaround for https://github.com/eslint/eslint/issues/3458
+require('@rushstack/eslint-patch/modern-module-resolution');
+
+const { statusCode, allSupportedExtensions } = require('./helpers/eslint');
+// const { craRules } = require('./helpers/rules-presets');
```
CarlosCortizasCT, line 5: Maybe this line is a leftover?

packages-backend/eslint-config-node/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 commercetools GmbH
```
tdeekens, line 3:
```suggestion
Copyright (c) 2022 commercetools GmbH
```

tdeekens (approved):
Looks good now. Let's start using it and adjust when needed.

CarlosCortizasCT (approved):
lgtm

packages-backend/eslint-config-node/LICENSE
emmenko: Yeah I'll do this as a follow up.

PS: I'm wondering if we can automate this with a github action or something that runs periodically once a year.

Issue #2059, opened 2021-02-17
Make <Notification> accessible


ahmehri:
**Is your feature request related to a problem? Please describe.**

Currently, closing notifications in tests is not ideal, we need to click on the close icon by selecting it using the following selector

```js
screen.getByLabelText(/Hide notification/i);
```

This is not ideal when we have more than one notification in the test, if I want to close the second notification, for example, I need to select the related close icon.

```js
screen.getAllByLabelText(/Hide notification/i)[1];
``` 

It's clear that the above selector is brittle and that the test will fail once the order of notifications changes.

**Describe the solution you'd like**

I would like to be able to select the close icon by first selecting the actual notification. This can be done by adding the a11y role `alertdialog` to the wrapper DOM element. `aria-describedby` property should be used, according to the [specification](https://www.w3.org/TR/wai-aria-1.1/#alertdialog), to reference the alert message element in the dialog.

```js
<div role="alertdialog" aria-describedby="content">
  <div id="content">{props.children}</div>
  {/* ... */}
</div>
```

The close icon can then be selected as follows.

```js
within(screen.findByRole('alertdialog', { description: /the message of the notification/i }).getByLabelText(/Hide notification/i);
```

**Describe alternatives you've considered**

I've considered using the role `dialog` but I ended up suggesting `alertdialog` instead because the latter is more suitable for the use case. A notification is [at the end](https://www.w3.org/TR/wai-aria-1.1/#alertdialog):

> A type of dialog that contains an alert message, where initial focus goes to an element within the dialog. See related alert and dialog.

**Additional context**

I tried to implement the proposed solution and I faced two challenges:

- The ID of the DOM element whose content will be the notification message should be unique. Otherwise, the selection won't be precise, in the above example of selecting the second notification, the result will be that the first notification is selected instead. I think this can be solved by passing a new prop `name` to the component and append it to the ID, `content-${name}`.
- The `ByRoleOptions` type of `@testing-library/dom` only supports the property `name` when selecting an element. I ignored the type warning and used `description` anyway and it worked, but still having the type warning is a problem. I also checked and couldn't find a way to set the name of an `alertdialog`.

```
// no type warning but it doesn't work
screen.findByRole('alertdialog', { name: /the message of the notification/i });

// type warning but it works
screen.findByRole('alertdialog', { description: /the message of the notification/i }
```

emmenko:
[...]

PR #2530, opened 2022-03-31
Add accessible attributes to notifications

CarlosCortizasCT:
#### Summary

Add accessible attributes to `Notification` component.

#### Description

Currently we are not using any attribute in `Notification` component that can help assistive technologies to properly be aware of this kind of elements when rendering them in the DOM.

The idea is to provide a proper _aria_ role and related attributes following [the spec](https://www.w3.org/TR/wai-aria-1.1/#alertdialog).

_(This is part of what we intend to work on related to #2059)_


packages/react-notifications/src/components/notification/notification.tsx
```diff
@@ -25,6 +27,8 @@ import {
   getStylesForNotification,
 } from './notification.styles';
 
+const sequentialId = createSequentialId('notification-');
```
emmenko, line 30: FYI: for us to aware.

With React v18 there is a new hook `useId` that we should be able to use as a replacement for our custom implementation.

https://reactjs.org/blog/2022/03/29/react-v18.html#useid

emmenko (approved):
Nice! 💯

packages/react-notifications/src/components/notification/notification.tsx
CarlosCortizasCT: Our id generators was the first thing that came to my mind when I read about that new hook 😄

kark (approved):
Spot on 🙌

ahmehri (approved):
💯

PR #2532, opened 2022-04-04
fix(eslint): regression of overriding test specific rules

emmenko:
Caused by #2526


CarlosCortizasCT (approved):
lgtm

packages/eslint-config-mc-app/index.js
```diff
@@ -140,9 +127,23 @@ module.exports = {
         'jest/globals': true,
       },
       rules: {
+        ...craRules.jest,
```
emmenko, line 130: The fix is that the jest/rtl rule overrides must be specified after the default CRA rules.

ddouglasz (approved):

tdeekens (approved):

PR #2534, opened 2022-04-04
fix(mc-scripts): downgrade react-refresh to v0.11.0

emmenko:
Caused by #2528

Ref: https://github.com/facebook/react/pull/24235

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -65,6 +65,14 @@ jobs:
           CI: true
           RTL_ASYNC_UTIL_TIMEOUT: 5000
 
+      - name: Testing Starter application development server
+        run: yarn start-server-and-test 'yarn template-starter:start' 3001 'exit 0'
+        env:
+          HOST_GCP_STAGING: ${{ secrets.HOST_GCP_STAGING }}
+          APP_ID: ${{ secrets.APP_ID_PLAYGROUND }}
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -65,6 +65,14 @@ jobs:
           CI: true
           RTL_ASYNC_UTIL_TIMEOUT: 5000
 
+      - name: Testing Starter application development server
+        run: yarn start-server-and-test 'yarn template-starter:start' 3001 'exit 0'
+        env:
+          HOST_GCP_STAGING: ${{ secrets.HOST_GCP_STAGING }}
+          APP_ID: ${{ secrets.APP_ID_PLAYGROUND }}
+          MC_API_URL: ${{ secrets.MC_API_URL }}
+          CTP_INITIAL_PROJECT_KEY: ${{ secrets.CYPRESS_PROJECT_KEY }}
```
emmenko, line 74: This should now catch these kind of errors for using the development server.

Otherwise on CI we have only been testing the production builds.

Rhotimee (approved):
💯

CarlosCortizasCT (approved):
lgtm

.github/workflows/main.yml
CarlosCortizasCT: awesome! 👍

tdeekens (approved):
Good follow-up!

PR #2535, opened 2022-04-04
feat(mc-script): show diff on CLI when updating custom app using config-sync

Rhotimee:
#### Summary
As a follow-up to implementing the config-sync command, a diff is displayed to show users the difference between config on the merchant center and config available locally before an update is made.

#### Screenshot
Below is an example of the displayed diff
image


packages/mc-scripts/src/commands/config:sync.js
```diff
@@ -2,6 +2,7 @@ const omit = require('lodash/omit');
 const prompts = require('prompts');
 const mri = require('mri');
 const chalk = require('chalk');
+const jsonDiff = require('json-diff');
```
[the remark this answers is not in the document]
Rhotimee: True. I agree with you. I think it's better to write a custom function to do the diffing since the data structure is known. 
I will update this. Thanks.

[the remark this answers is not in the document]
emmenko: Mainly we were looking for a library to show diff changes in a helpful way in the terminal. Computing the differences can also be done "manually" in this case yes.

packages/mc-scripts/src/commands/config-sync.js
```diff
@@ -126,12 +128,35 @@ const configSync = async () => {
...
 
-  // TODO: show diff (followup task)
+  const customAppLink = getMcUrlLink(
+    mcApiUrl,
+    fetchedCustomApplication.organizationId,
+    fetchedCustomApplication.application.id
+  );
+
+  // Sanitize SVG to be consistent with the one in MC
+  localCustomAppData.icon = sanitizeSvg(localCustomAppData.icon);
```
CarlosCortizasCT, line 138: Do we really need to sanitize the icon?
Since we will be just displaying it in the console, I can't see the security problem.

When creating the custom application in the Settings API, we are already [sanitizing its value](https://github.com/commercetools/merchant-center-services/blob/main/packages/settings/lib/resolvers/mutations/organization-extension/create-custom-application.ts#L69) before saving the data.

Rhotimee: I understand that the icon SVG is sanitised in the settings API. However, I'm sanitising the local config icon so that the diffing helper can differentiate if the icon is a new one or the previously sanitized icon saved in the mc.

.changeset/little-papayas-attend.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-frontend/mc-scripts': minor
+---
+
+As a follow-up to implementing the config-sync command, a diff is displayed to show users the difference between config on the merchant center and config available locally before an update is made.
```
emmenko, line 5:
```suggestion
The `config:sync` command now shows the diff changes before updating an existing Custom Application config.
```

packages/mc-scripts/src/commands/config-sync.js
emmenko, line 138: Maybe we can do that when we parse the config instead of here.

https://github.com/commercetools/merchant-center-application-kit/blob/238bd34a6045fd4cda443585b714ac0689ecd335/packages/application-config/src/transformers.ts#L43

packages/mc-scripts/src/utils/get-config-diff.js
```diff
@@ -0,0 +1,246 @@
+const { red, green } = require('chalk');
+
+const notNestedKeySet = new Set(['name', 'description', 'url', 'icon']);
+
+const t = (tabs) => '\t'.repeat(tabs);
```
emmenko, line 5: Nit: I'm not a big fan of using tabs indentation. Can we use 2 spaces?

packages/mc-scripts/src/utils/get-config-diff.js
```diff
@@ -0,0 +1,246 @@
...
+
+const getStringDiff = (oldStr, newStr, key, tabs = 0) => {
```
emmenko, line 7: Nit: use clearer variable names.

```suggestion
const getStringDiff = (prevValue, nextValue, label, indentLevel = 0) => {
```

packages/mc-scripts/src/utils/get-config-diff.js
```diff
@@ -0,0 +1,246 @@
...
+
+  // if there are old permissions left, it means they were deleted in the new Permissions.
+  Object.keys(mappedOldPermission).forEach((oldPermissionName) => {
+    permissionDiff.push(`\t"${red(oldPermissionName)}" was removed`);
+  });
+
+  if (permissionDiff.length > 1) return permissionDiff.join('\n');
+};
+
+const getArrayDiff = (oldArray, newArray, tabs = 0) => {
```
emmenko, line 56: Nit: move this after `getStringDiff`

emmenko, line 56:
```suggestion
const getArrayDiff = (prevValue, nextValue, indentLevel = 0) => {
```

packages/mc-scripts/src/utils/get-config-diff.js
```diff
@@ -0,0 +1,246 @@
...
+  const oldArraySet = new Set(oldArray);
+  const arrayDiff = [];
+
+  newArray?.forEach((item) => {
+    if (oldArraySet.has(item)) {
+      oldArraySet.delete(item);
+    } else {
+      arrayDiff.push(`${t(tabs)}Item with "${green(item)}" added`);
```
emmenko, line 64: Nit: can't we use the `getStringDiff` function?

```suggestion
      arrayDiff.push(getStringDiff(null, item, 'Item with', indentLevel));
```

packages/mc-scripts/src/utils/get-config-diff.js
```diff
@@ -0,0 +1,246 @@
...
+const getPermissionsDiff = (oldPermissions, newPermissions) => {
+  const permissionDiff = ['permissions changed'];
+  const mappedOldPermission = {};
+
+  oldPermissions.forEach(({ name, oAuthScopes }) => {
+    mappedOldPermission[name] = oAuthScopes;
+  });
+
+  newPermissions.forEach((newPermission) => {
+    const currDiff = [`\t${newPermission.name} changed`];
```
emmenko, line 26: I noticed that sometimes you use `t(tabs)` and for example here you use `\t` directly. Let's try to be a bit more consistent maybe.

packages/mc-scripts/src/utils/get-config-diff.js
```diff
@@ -0,0 +1,246 @@
...
+          getSubmenuLinksDiff(
+            oldConfig['submenuLinks'],
+            newConfig['submenuLinks']
+          )
+        );
+        break;
+      default:
+        break;
+    }
+  });
```
emmenko, line 241: I'm wondering if we can implement this in a more generic and scalable way.

For example, instead of "hardcoding" the fields that we are intend to look for, I would image that we can implement a recursive function that checks e.g. the `typeof` of each field and determines which "diff" function to execute.

This would have the advantage of not having to change this implementation every time we might add/change some of the fields in the config.

packages/mc-scripts/src/utils/get-config-diff.spec.js
```diff
@@ -0,0 +1,235 @@
...
+  it('should display no diff', () => {
+    expect(getConfigDiff(oldConfig, newConfig)).toBe('');
+  });
+
+  it('should display name diff', () => {
+    const updatedNewConfig = {
+      ...newConfig,
+      name: 'updated test name',
+    };
+    expect(getConfigDiff(oldConfig, updatedNewConfig)).toBe(
```
emmenko, line 34: I think for these kind of tests we can use `toMatchInlineSnapshot`

emmenko (changes requested):
Thanks a lot for giving this a try! The output is definitely much readable now.

In terms of implementation though I think we should improve some things to make it more maintainable for the future.

Let me know what you think about it, happy to have an in-depth chat about it 🤗

packages/mc-scripts/src/utils/get-config-diff.js
Rhotimee: The wording for getStringDiff is different. That's why I didn't use it.

packages/mc-scripts/src/utils/get-config-diff.js
Rhotimee: As discussed, we concluded to leave the hardcoded implementation for now and figure out how we can improve this in a scalable way for when we add/update the fields.

packages/mc-scripts/src/utils/get-config-diff.js
```diff
@@ -0,0 +1,252 @@
...
+  return arrayDiff.join('\n');
+};
+
+const getPermissionsDiff = (oldPermissions, newPermissions) => {
+  const permissionDiff = ['permissions changed'];
+  const mappedOldPermission = {};
+
+  oldPermissions.forEach(({ name, oAuthScopes }) => {
+    mappedOldPermission[name] = oAuthScopes;
+  });
```
CarlosCortizasCT, line 43: nit: Perhaps we could use `reduce` here instead of `forEach`.
```suggestion
const mappedOldPermissions = oldPermissions.reduce(
  (acc, { name, oAuthScopes }) => ({
    ...acc,
    [name]: oAuthScopes,
  }),
  {}
);
```

packages/mc-scripts/src/utils/get-config-diff.js
```diff
@@ -0,0 +1,252 @@
...
+
+  if (labelAllLocalesDiff.length > 1) return labelAllLocalesDiff.join('\n');
+};
+
+const getMainMenuLinkDiff = (oldMainMenuLink, newMainMenuLink) => {
+  const mainMenuLinkDiff = ['mainMenuLink changed'];
+
+  mainMenuLinkDiff.push(
+    getStringDiff(
+      oldMainMenuLink['defaultLabel'],
```
CarlosCortizasCT, line 124: I'm curious if we have any linting rule that enforces using square brackets for accessing object properties (`oldMainMenuLink['defaultLabel']`) instead of dots (`oldMainMenuLink.defaultLabel`).

CarlosCortizasCT (commented):
Impressive work here 💪 .

I just left a couple of comments for better understanding.

packages/mc-scripts/src/utils/get-config-diff.js
Rhotimee: Updated.

CarlosCortizasCT (approved):
Good to go from my side once tests pass again 🚀

packages/mc-scripts/src/utils/get-config-diff.js
```diff
@@ -0,0 +1,269 @@
+const chalk = require('chalk');
+
+// Since not all terminal supports colors, to make things more consistent for testing purposes,
+// during tests the color used is appended before the string instead of coloring it.
+const isTest = process.env.NODE_ENV === 'test';
+const red = (str) => {
+  if (isTest) return 'red-' + str;
```
emmenko, line 7: Maybe we can wrap the string like

```suggestion
  if (isTest) return `<color-red>${str}</color-red>`
```

packages/mc-scripts/src/utils/get-config-diff.js
```diff
@@ -0,0 +1,269 @@
...
+  return chalk.red(str);
+};
+const green = (str) => {
+  if (isTest) return 'green-' + str;
+  return chalk.green(str);
+};
+
+const keysToSkip = new Set(['id']);
+
+const t = (indentLevel) => '  '.repeat(indentLevel);
```
emmenko, line 17: Nit: maybe add a comment that we are using 2 spaces

packages/mc-scripts/src/utils/get-config-diff.js
```diff
@@ -0,0 +1,269 @@
...
+  });
+
+  return arrayDiff.join('\n');
+};
+
+const getPermissionsDiff = (oldPermissions, newPermissions) => {
+  const permissionDiff = ['permissions changed'];
+
+  const mappedOldPermissions = oldPermissions.reduce(
+    (acc, { name, oAuthScopes }) => ({
```
emmenko, line 53: Nit: when using `reduce` we try to provide meaningful names to the variables, for example `acc` is not super clear what it refers to.

packages/mc-scripts/src/utils/get-config-diff.spec.js
```diff
@@ -0,0 +1,224 @@
...
+
+const newConfig = {
+  name: 'test name',
+  description: 'test description',
+  url: 'https://test.com',
+  icon: '<svg><path fill="#000000"></path></svg>',
+  permissions: [],
+  mainMenuLink: {},
+  submenuLinks: [],
+};
```
emmenko, line 21: I would suggest to define a function to return an object

```js
const createTestConfig = (custom) => ({
  name: 'test name',
  // ...
  ...custom,
})
```

Then you can use the function to initialize e.g. old/new config values etc.

emmenko (commented):
Suggestions to make things a bit more readable.

packages/mc-scripts/src/utils/get-config-diff.js
```diff
@@ -0,0 +1,270 @@
+const chalk = require('chalk');
+
+// Since not all terminal supports colors, to make things more consistent for testing purposes,
+// during tests the color used is appended before the string instead of coloring it.
```
CarlosCortizasCT, line 4: Perhaps this comment is now outdated since we don't just append something before the strings but we wrap them.

Maybe we could update it to something like this:
```suggestion
// We use `chalk` to help users identify the changes we display in the terminal by using
// different colors for added/removed/updated values.
// There is a problem though since CI terminal does not support colors and that makes tests
// fail when we validate the output.
// To overcome that situation, we use a different approach in testing environment where we wrap
// relevant changes with custom tags so we can have consistent expectations
// during tests the color used is appended before the string instead of coloring it.
```

nit: we could even have a third color to distinguish between deletions and updates

packages/mc-scripts/src/utils/get-config-diff.spec.js
```diff
@@ -0,0 +1,223 @@
...
+const createNewTestConfig = (customConfig) => ({
+  name: 'test name',
+  description: 'test description',
+  url: 'https://test.com',
+  icon: '<svg><path fill="#000000"></path></svg>',
+  permissions: [],
+  mainMenuLink: {},
+  submenuLinks: [],
+  ...customConfig,
+});
```
CarlosCortizasCT, line 23: I guess we only need one function here:

```suggestion
const createTestConfig = (customConfig) => ({
  name: 'test name',
  description: 'test description',
  url: 'https://test.com',
  icon: '<svg><path fill="#000000"></path></svg>',
  permissions: [],
  mainMenuLink: {},
  submenuLinks: [],
  ...customConfig,
});
```

CarlosCortizasCT (commented):
Looking pretty good 👍 

Just left a couple of comments with some suggestions.

packages/mc-scripts/src/utils/get-config-diff.spec.js
emmenko: Yes only one function

Rhotimee: ahh.. True.

packages/mc-scripts/src/utils/get-config-diff.js
Rhotimee: Yes, thanks. We needed to update the comment.

Rhotimee: For updates, red signifies what was removed and green for what was added. 
image


I don't know if I understand you correctly, though.

CarlosCortizasCT: Oh! sorry. My bad.
Forget about it.

emmenko (approved):
This is good to go from my side. Great work! 🙌

PR #2536, opened 2022-04-05
Make non required props optional in typescript definitions

CarlosCortizasCT:
#### Summary

When using `TabularMainPage` component I noticed the `title` prop is marked as required in its typescript definition but that prop is actually optional since there is another one (`customTitleRow`) which you can use to tailor the header section of that component instead of just using a _string_.

I also found the same situation with `TabularDetailPage` component.

#### Description

This PR updates both components props TS definition to set `title` as optional and have am internal check so either that one or `customTitleRow` are provided when using the components.


.changeset/neat-trees-itch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-frontend/application-components': patch
+---
+
+Update both `TabularMainPage` and `TabularDetailPage` components so its `title` prop is not required in their typescript definitions as they are actually optional in their implementation.
```
emmenko, line 5:
```suggestion
Make the `title` prop optional for the components `<TabularMainPage>` and `<TabularDetailPage>`.

A warning will be logged if neither the `title` or the `customTitleRow` props are provided.
```

packages/application-components/src/components/tabular-detail-page/tabular-detail-page.tsx
```diff
@@ -29,7 +30,7 @@ type TTabularDetailPageProps = {
   /**
    * The title of the page.
    */
-  title: string;
+  title?: string;
```
emmenko, line 33: Can you also update the docs (prop types table) to make this optional?

We don't have this automated in appkit yet. Thanks

emmenko (commented):
Thanks, I'm fine then with this.

packages/application-components/src/components/tabular-main-page/tabular-main-page.tsx
```diff
@@ -51,40 +52,47 @@ const defaultProps: Pick<TTabularMainPageProps, 'hideControls'> = {
...
-  <PageWrapper>
-    <TabularPageContainer color="surface">
-      <Spacings.Stack>
-        {props.customTitleRow || (
-          <PageHeaderTitle
-            title={props.title}
-            subtitle={props.subtitle}
-            titleSize="big"
+const TabularMainPage = (props: TTabularMainPageProps) => {
+  warning(
```
kark, line 56: nit: we could also add tests for this case - I attempted doing that [at some point](https://github.com/commercetools/merchant-center-application-kit/blob/bf93463ebd4cb8cf80294dcaaedb115e7adeb8f5/packages/application-components/src/components/tabular-detail-page/tabular-detail-page.spec.tsx#L112:L121)

kark (commented):
Looking good! 👍

packages/application-components/src/components/tabular-main-page/tabular-main-page.tsx
CarlosCortizasCT: Sure.
I added a new test to both components.

Thanks for the suggestion 👍

emmenko (approved):

kark (approved):

PR #2537, opened 2022-04-05
fix(components): tab page itself should not scroll

emmenko:
Follow up of #2517

### DIFF
--- a/packages/application-components/src/components/internals/page.styles.ts
+++ b/packages/application-components/src/components/internals/page.styles.ts
@@ -3,6 +3,7 @@ import { customProperties } from '@commercetools-uikit/design-system';
 
 export const ContentWrapper = styled.div`
   flex: 1;
+  flex-basis: 0;
   padding: ${customProperties.spacingM};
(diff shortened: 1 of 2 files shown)


packages/application-components/src/components/internals/page.styles.ts
```diff
@@ -3,6 +3,7 @@ import { customProperties } from '@commercetools-uikit/design-system';
 
 export const ContentWrapper = styled.div`
   flex: 1;
+  flex-basis: 0;
```
tdeekens, line 6: Flex-basis overrules any specified CSS width value, so if you set flex-basis to 0, it doesn't fall back to the CSS "width" value.

tdeekens (approved):

CarlosCortizasCT (approved):
It seems to be the right solution, but I must admit I don't fully understand why `flex-basis` rule solves the issue here.

kark (approved):
Same as @CarlosCortizasCT. Seems to be fixing!

PR #2538, opened 2022-04-05
fix(babel): flaky resolution of emotion babel preset

emmenko:
Sometimes we get an error like `Cannot find module @emotion/babel-preset-css-prop`.

Usually this relates to how dependencies are hoisted in `node_modules`.

I noticed though that we weren't using `require()` to resolve the babel preset. This now should hopefully fix that problem.

### DIFF
--- a/.changeset/funny-crabs-applaud.md
+++ b/.changeset/funny-crabs-applaud.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/babel-preset-mc-app': patch
+---
+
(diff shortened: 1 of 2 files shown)


tdeekens (approved):
Good catch. So it's a mix of Node.js module resolution and yarn's hoisting I guess?

emmenko:
I suppose so yes

CarlosCortizasCT (approved):

PR #2541, opened 2022-04-07
fix(components): style position for `<PortalsContainer>` layout

emmenko:
Fixes #2050

PR #2554, opened 2022-04-14
fix: offset calculation of modal pages when notifications are opened

emmenko:
Follow up of #2549 and #2541

Using fixed positioned layout for modal pages had obviously some consequences, for instance how the layout adjusts itself when notifications are rendered on a page.

With the fixed position we now need to adjust the offset/top positions, which are based on the heights of the notifications (global + page).

To calculate the heights we're using now a `ResizeObserver` to update the position whenever the notifications are rendered/removed.

To avoid possible regressions in the future I also added some e2e + percy tests using some new custom pages in the playground app.

### DIFF
--- a/cypress/integration/playground/notifications.js
+++ b/cypress/integration/playground/notifications.js
@@ -1,0 +1,50 @@
+import {
+  URL_APP_KIT_PLAYGROUND_NOTIFICATIONS,
+  ENTRY_POINT_APP_KIT_PLAYGROUND,
+} from '../../support/urls';
+
+describe('Notifications', () => {
+  beforeEach(() => {
+    cy.loginByOidc({
+      entryPointUriPath: ENTRY_POINT_APP_KIT_PLAYGROUND,
+      initialRoute: URL_APP_KIT_PLAYGROUND_NOTIFICATIONS,
+    });
+  });
+  it('should adjust layout when notifications are open', () => {
+    cy.findByLabelText('Open modal 1');
+    cy.findByLabelText('Global notification').click();
+    cy.findByText('hello').should('exist');
+
+    cy.findByLabelText('Page notification').click();
+    cy.findByLabelText('Page notification').click();
+    cy.findAllByText('oops').should('have.length', 2);
+
+    cy.findByLabelText('Side notification').click();
+    cy.findByLabelText('Side notification').click();
+    cy.findAllByText('ok').should('have.length', 2);
+
+    cy.percySnapshot();
+  });
+
+  it('should adjust layout for modals when notifications are open', () => {
+    // Open modal
+    cy.findByLabelText('Open modal 1').click();
+    // Open a second modal
+    cy.findByLabelText('Open modal 2').click();
+
+    cy.findByLabelText('Modal page 2').within(() => {
+      cy.findByLabelText('Global notification').click();
+
+      cy.findByLabelText('Page notification').click();
+      cy.findByLabelText('Page notification').click();
+
+      cy.findByLabelText('Side notification').click();
+      cy.findByLabelText('Side notification').click();
+    });
+
+    cy.findByText('hello').should('exist');
+    cy.findAllByText('oops').should('have.length', 2);
+    cy.findAllByText('ok').should('have.length', 2);
+    cy.percySnapshot();
+  });
+});
--- a/packages/application-components/src/components/portals-container/portals-container.tsx
+++ b/packages/application-components/src/components/portals-container/portals-container.tsx
@@ -1,6 +1,21 @@
+import {
+  type MutableRefObject,
+  type RefObject,
+  forwardRef,
+  useState,
+} from 'react';
 import { css, Global } from '@emotion/react';
+import useResizeObserver from '@react-hook/resize-observer';
 import { PORTALS_CONTAINER_ID } from '@commercetools-frontend/constants';
 
+type TLayoutRefs = {
+  notificationsGlobalRef: MutableRefObject<HTMLDivElement>;
+  notificationsPageRef: MutableRefObject<HTMLDivElement>;
+};
+type TObservableElementDimensions = {
+  height: number;
+  width: number;
+};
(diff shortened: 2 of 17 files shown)


cypress/integration/playground/notifications.js
```diff
@@ -0,0 +1,50 @@
...
+      cy.findByLabelText('Side notification').click();
+      cy.findByLabelText('Side notification').click();
+    });
+
+    cy.findByText('hello').should('exist');
+    cy.findAllByText('oops').should('have.length', 2);
+    cy.findAllByText('ok').should('have.length', 2);
+    cy.percySnapshot();
+  });
+});
```
tdeekens, line 50: Good testing. This should catch such regressions earlier.

packages/react-notifications/src/components/notifications-list/notifications-list.styles.ts
```diff
@@ -13,7 +13,7 @@ const getStyles = (props: StyleProps) => {
     color: ${customProperties.colorSurface};
     position: relative;
     width: 100%;
-    z-index: 10000;
+    z-index: 19999;
```
emmenko, line 16: This is the "fix" for the notifications.

playground/src/components/notifications-playground/notifications-playground.js
```diff
@@ -0,0 +1,92 @@
...
+          onClick={modalState.openModal}
+        />
+
+        <InfoModalPage
+          isOpen={modalState.isModalOpen}
+          title={`Modal page ${props.level}`}
+          onClose={modalState.closeModal}
+          level={props.level}
+        >
+          <NotificationsPlayground level={props.level + 1} />
```
emmenko, line 79: FYI: it's a recursive rendering 🤓

packages/application-shell/src/components/application-shell/application-shell.tsx
```diff
@@ -149,6 +155,17 @@ export const RestrictedApplication = <
...
+
+  const notificationsGlobalRef = useRef<HTMLDivElement>(null);
+  const notificationsPageRef = useRef<HTMLDivElement>(null);
+  const layoutRefs = useRef<{
+    notificationsGlobalRef: RefObject<HTMLDivElement>;
+    notificationsPageRef: RefObject<HTMLDivElement>;
+  }>({
+    notificationsGlobalRef,
+    notificationsPageRef,
+  });
```
emmenko, line 167: I tried different things but eventually ended up with this.

We're "gathering" different refs into one and pass it to the `PortalsContainer`.

packages/application-components/src/components/portals-container/portals-container.tsx
```diff
@@ -36,49 +51,86 @@ const defaultProps: Pick<
...
-  offsetLeft: '0',
-  offsetLeftOnExpandedMenu: '0',
+  offsetTop: '0px',
+  offsetLeft: '0px',
+  offsetLeftOnExpandedMenu: '0px',
   containerSelectorToPreventScrollingOnOpen: 'main',
   zIndex: 10000,
 };
 
+const useObserverElementDimensions = (
```
tdeekens, line 61: Will be used twice. Once for page and once for global nofications. Both wrapper divs will receive a ref so we can target them.

packages/application-components/src/components/portals-container/portals-container.tsx
```diff
@@ -36,49 +51,86 @@ const defaultProps: Pick<
...
-          bottom: 0;
-          z-index: ${props.zIndex};
-        }
+            .ReactModal__Body--open #${PORTALS_CONTAINER_ID} {
+              position: fixed;
+              height: calc(
+                100% - ${props.offsetTop} -
+                  ${globalNotificationsElementDimensions.height}px -
+                  ${pageNotificationsElementDimensions.height}px
+              );
```
emmenko, line 105: This is where we calculate the height based on the notifications height.

packages/application-shell/src/components/application-shell/application-shell.tsx
```diff
@@ -149,6 +155,17 @@ export const RestrictedApplication = <
   // However, every route change will trigger a re-render. This is probably
   // ok-ish but we might want to look into a more performant solution.
   const location = useLocation();
+
+  const notificationsGlobalRef = useRef<HTMLDivElement>(null);
+  const notificationsPageRef = useRef<HTMLDivElement>(null);
```
tdeekens, line 160: These two are passed into the elements we intend to measure.

packages/application-shell/src/components/application-shell/application-shell.tsx
```diff
@@ -149,6 +155,17 @@ export const RestrictedApplication = <
...
+  const layoutRefs = useRef<{
```
tdeekens, line 161: This is auxiliary and ends up in the portals container so that there we can pass it on to the mutation observer.

packages/application-shell/src/components/application-shell/application-shell.tsx
```diff
@@ -404,10 +424,16 @@ export const RestrictedApplication = <
                               `}
                             >
                               <PortalsContainer
+                                // @ts-ignore
+                                ref={layoutRefs}
                                 offsetTop={DIMENSIONS.header}
-                                offsetLeft={DIMENSIONS.navMenu}
+                                offsetLeft={
+                                  projectKeyFromUrl ? DIMENSIONS.navMenu : '0px'
+                                }
```
CarlosCortizasCT, line 432: I would like to ask about this check for `projectKeyFromUrl`.
What use case is this?

emmenko: In the "account" app there is no left side menu. We can derive this check from the presence of the project key in the url, or not.

CarlosCortizasCT (approved):
lgtm

tdeekens (approved):
Thanks. Reviewed again. Looks good!

PR #2558, opened 2022-04-19
New documentation page for `actions-global`

kark:
#### Summary

A new documentation page added to document the usage of `@commercetools-frontend/actions-global` package.
Additionally, a part of form docs about [splitting form fields](https://docs.commercetools.com/custom-applications/development/forms#splitting-form-fields) was changed.


website/src/content/api-reference/commercetools-frontend-actions-global.mdx
```diff
@@ -0,0 +1,319 @@
...
+
+## Action creators and hooks
+
+`"@commercetools-frontend/actions-global"` package exposes both action creators (for instance `showNotification`) and hooks (for instance `useShowNotification`) that invoke the corresponding action creators under the hood.
+
+<Info>
+
+Whenever possible, we recommend using the hooks which have Redux store's `dispatch` method bound within the hook rather than the action creators directly. 
+
+</Info>
```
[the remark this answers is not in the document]
emmenko: Also, I think we can structure the page like this:

```
# Action creators

## showNotification

# Hooks

## useShowNotification
```

So in the action creators we document and explain what happens. In the hooks we simply mention how to use the hook and which action creator is used.

website/src/content/development/forms.mdx
```diff
@@ -610,43 +610,47 @@ Sometimes a form contains many different fields, resulting in the form component
...
+      }}
+    </Formik>
   );
-
-  return props.children({
-    formElements,
-    isDirty: formik.dirty,
-    isSubmitting: formik.isSubmitting,
-    submitForm: formik.handleSubmit,
-    handleCancel: formik.handleReset,
```
emmenko, line 647: I think this was meant to be named `handleCancel`

emmenko (commented):
Thanks for giving it a shot.

I left some comments and suggestions to work on the structure of the page as there are also some things that we can omit for now. Let me know if it's clear enough or we need to clarify a couple of things. 🤗

website/src/content/api-reference/commercetools-frontend-actions-global.mdx
```diff
@@ -0,0 +1,139 @@
...
+
+The `@commercetools-frontend/actions-global` package provides Redux action creators and React hooks to facilitate dispatching notifications within Custom Applications.
+
+# Action creators
+
+<Info>
+
+Rather than using the action creators directly we recommend using the corresponding React hooks. The hooks have the `dispatch` method of the Redux store bound internally. 
+
+</Info>
```
emmenko, line 43: Nice

website/src/content/api-reference/commercetools-frontend-actions-global.mdx
```diff
@@ -0,0 +1,139 @@
...
+
+## showNotification
+
+An action creator to dispatch any kind of notification.
+
```
emmenko, line 48: Maybe we can also document the function options/signature.

This includes for example the notification options like `domain`, `kind`, `text` and the "meta options" like `dismissAfter`.

website/src/content/development/forms.mdx
```diff
@@ -610,43 +610,47 @@ Sometimes a form contains many different fields, resulting in the form component
 
 One way to improve that is to split the form component into multiple smaller components. As a rule of thumb you can create one separate component for each form field. For example a `<FormChannelNameField>`, a `<FormChannelKeyField>`, and so on.
 
-These components are simple wrapper components of the UI Kit field components but they contain and encapsulate all the logic related to the specific field. As a result, your form component can look like this:
+These components are simple wrapper components of the UI Kit field components but they contain and encapsulate all the logic related to the specific field. 
+
+Beforehand, our Channels form must be refactored to use `<Formik>` component as a context provider. As a result, the form component can look like this:
```
emmenko, line 615:
```suggestion
<Info>

To use the `useField` hook the form must be initialized using the `<Formik>` component instead of using the `useFormik` hook, so that the React context is properly defined.

</Info>

As a result, the form component can look like this:
```

emmenko (commented):
Nice, looks very good.

Just a couple of small suggestions to make things a bit more clear.

website/src/content/api-reference/commercetools-frontend-actions-global.mdx
```diff
@@ -0,0 +1,157 @@
...
+An action creator to dispatch an (page) error notification for [API errors](https://docs.commercetools.com/api/errors) of the commercetools platform API.
+
+## handleActionError
+
+An action creator that attempts to handle a generic error and dispatch the appropriate (page) notification error if possible. Use this if you simply need to ensure a notification error is shown without knowing what the exact error is.
+
+## hideAllPageNotifications
+
+An action creator to hide all active page notifications.
```
emmenko, line 77: What about the signatures of these functions?

website/src/content/api-reference/commercetools-frontend-actions-global.mdx
```diff
@@ -0,0 +1,157 @@
...
+
+- `notification` object properties
+
+| Property            | Type       | Required | Values                                                                                | Default | Description                                        |
+| ------------------- | ---------- | :------: | ------------------------------------------------------------------------------------- | ------- | -------------------------------------------------- |
+| `kind`              | `string`   |    ✅    | `success`, `error`, `warning`, `info`, `unexpected-error`, `api-error`                | -       | The kind of the notification.                      |
+| `domain`            | `string`   |    ✅    | `side`, `global`, `page`                                                              | -       | The domain of the notification.                    |
+| `text`              | `string`   |    -     | -                                                                                     | -       | The text content of the notification.              |
+
+
```
emmenko, line 58: Maybe we can mention here as well that the `kind` and `domain` values can be defined using the `constants` package.

emmenko (commented):
Looking good, just some minor suggestions 👌

website/src/content/api-reference/commercetools-frontend-actions-global.mdx
```diff
@@ -0,0 +1,240 @@
...
+Custom Applications come with a built-in notification system built on top of [Redux](https://redux.js.org).
+
+The `@commercetools-frontend/actions-global` package provides Redux action creators and React hooks to facilitate dispatching notifications within Custom Applications.
+
+# Action creators
+
+<Info>
+
+Rather than using the action creators directly we recommend using the corresponding React hooks. The hooks have the `dispatch` method of the Redux store bound internally.
```
CarlosCortizasCT, line 41: Would that be ok to emphasize our recommendation on using React hooks?

```suggestion
Rather than using the action creators directly **we recommend using the corresponding React hooks**. The hooks have the `dispatch` method of the Redux store bound internally. 
```

website/src/content/api-reference/commercetools-frontend-actions-global.mdx
```diff
@@ -0,0 +1,240 @@
...
+npm --save install react redux react-redux
+```
+
+# Notification system
+
+Custom Applications come with a built-in notification system built on top of [Redux](https://redux.js.org).
+
+The `@commercetools-frontend/actions-global` package provides Redux action creators and React hooks to facilitate dispatching notifications within Custom Applications.
+
+# Action creators
```
CarlosCortizasCT, line 37: Since we recommend using React hooks, perhaps it would help to have its docs before the actions creators but I know it might not be feasible because hooks descriptions depend on actions creators ones.

website/src/content/api-reference/commercetools-frontend-actions-global.mdx
```diff
@@ -0,0 +1,240 @@
...
+<Info>
+
+Rather than using the action creators directly we recommend using the corresponding React hooks. The hooks have the `dispatch` method of the Redux store bound internally. 
+
+</Info>
+
+## showNotification
+
+An action creator to dispatch any kind of notification.<br />
+Takes 2 arguments: `notification` object (required) and `meta` object (optional).
```
CarlosCortizasCT, line 48: I'm confused about `text` not being required since I can't see a use case when we want to show a notification with no text.

website/src/content/api-reference/commercetools-frontend-actions-global.mdx
```diff
@@ -0,0 +1,240 @@
...
+| `dismissAfter`      | `number`   |     -    | -              | `5000` (only for the `success` notification)      | Delay (in milliseconds) between the appearing and the closing of the notification.     |
+| `onDismiss`         | `function` |     -    | -              | -                                                 | Called when the notification closes. Receives notification `id` as an argument.        |
+
+
+## showApiErrorNotification
+
+An action creator to dispatch an (page) error notification for [API errors](https://docs.commercetools.com/api/errors) of the commercetools platform API.
+Takes as an argument an object with `errors` property being an error object or an array of error objects.
+
+Some sources or errors (for instance GraphQL APIs) return an array or object. `showApiErrorNotification` action creator accepts either as the value of `errors` property.
```
CarlosCortizasCT, line 89: Not sure about this sentence.
Perhaps there is a typo?

```suggestion
Some sources of errors (for instance GraphQL APIs) return an array or object. `showApiErrorNotification` action creator accepts either as the value of `errors` property. 
```

website/src/content/api-reference/commercetools-frontend-actions-global.mdx
```diff
@@ -0,0 +1,240 @@
...
+        message:
+          "This is an error",
+        code: "Invalid",
+      },
+    ],
+  },
+}
+// ...
+dispatch(handleActionError(error)); // dispatch method from Redux store
+```
```
CarlosCortizasCT, line 130: I think it would he helpful to explain which property is used for the notification error.
For instance, I'm wondering whether `error.body.message` or a concatenation of `error.body.errors[n].message`.

website/src/content/api-reference/commercetools-frontend-actions-global.mdx
```diff
@@ -0,0 +1,240 @@
...
+];
+// ...
+dispatch(showApiErrorNotification({ errors })); // dispatch method from Redux store
+```
+
+## handleActionError
+
+An action creator that attempts to handle a generic error and dispatch the appropriate (page) notification error if possible. Use this if you simply need to ensure a notification error is shown without knowing what the exact error is.
+
+Takes an error object as an argument:
```
CarlosCortizasCT, line 110: Suggestions:
```suggestion
It takes only one parameter that can be any of these types:
```

website/src/content/development/forms.mdx
```diff
@@ -610,43 +610,51 @@ Sometimes a form contains many different fields, resulting in the form component
 
 One way to improve that is to split the form component into multiple smaller components. As a rule of thumb you can create one separate component for each form field. For example a `<FormChannelNameField>`, a `<FormChannelKeyField>`, and so on.
 
-These components are simple wrapper components of the UI Kit field components but they contain and encapsulate all the logic related to the specific field. As a result, your form component can look like this:
+<Info>
+
+To use the `useField` hook the form must be initialized using the `<Formik>` component instead of using the `useFormik` hook, so that the React context is properly defined.
```
CarlosCortizasCT, line 615: Suggestion:
```suggestion
To use the `useField` hook the form must be wrapped using the `<Formik>` component instead of using the `useFormik` hook, so that the React context is properly defined.
```

website/src/content/development/forms.mdx
```diff
@@ -610,43 +610,51 @@ Sometimes a form contains many different fields, resulting in the form component
...
-  });
-
-  // Only contains the form elements, no buttons.
-  const formElements = (
-    <Spacings.Stack scale="l">
-      <FormChannelNameField />
-      <FormChannelKeyField />
-    </Spacings.Stack>
+  return (
+    <Formik
```
CarlosCortizasCT, line 630: I would suggest to also wrap the field with a `form` element as suggested in [Formik's docs](https://formik.org/docs/api/formik) as it is more semantic.
Also, perhaps we could add a submit button for a better complete form example.

CarlosCortizasCT (commented):
Great work over here 👍 

I left some comments suggestions.

website/src/content/api-reference/commercetools-frontend-actions-global.mdx
kark: +1

website/src/content/development/forms.mdx
kark: I totally agree, but it seems to me the whole page deserves revisiting and adding `form` and the submit button wherever needed. There is a `// Only contains the form elements, no buttons.` comment in multiple places and either it should be addressed everywhere or left as it is. But it seems it's out of the scope of this PR. I only wanted to fix the issue with Formik context not being present when we use `useField`

CarlosCortizasCT: Yes, I also agree then that it is out of this PR's scope.
Perhaps something to look into as a follow-up.

website/src/content/api-reference/commercetools-frontend-actions-global.mdx
CarlosCortizasCT: Ah! understood.
I was expecting the user to always provide the message, but if we already provide that behaviour, me might want to keep it.

Thanks for the explanation 👍

CarlosCortizasCT (approved):
💯

website/src/content/development/forms.mdx
```diff
@@ -610,43 +610,51 @@ Sometimes a form contains many different fields, resulting in the form component
 
 One way to improve that is to split the form component into multiple smaller components. As a rule of thumb you can create one separate component for each form field. For example a `<FormChannelNameField>`, a `<FormChannelKeyField>`, and so on.
 
-These components are simple wrapper components of the UI Kit field components but they contain and encapsulate all the logic related to the specific field. As a result, your form component can look like this:
+<Info>
+
+To use the `useField` hook the form must be wrapped with the `<Formik>` component instead of using the `useFormik` hook, so that the React context is properly defined.
```
adinakleine, line 615:
```suggestion
To use the `useField` hook, the form must be wrapped with the `<Formik>` component instead of using the `useFormik` hook, so that the React context is properly defined.
```

adinakleine (approved):
Find some writing style suggestions. Thanks @kark 🙌

PR #2563, opened 2022-04-21
chore: update `react-intl` to `^5.25.0` 

kark:
#### Summary

This PR bumps `react-intl` to the newest **caret** version.
Additionally, `react-intl` is removed from `resolutions` section in the main `package.json`

### DIFF
--- a/.changeset/beige-eyes-agree.md
+++ b/.changeset/beige-eyes-agree.md
@@ -1,0 +1,13 @@
+---
+'merchant-center-application-template-starter': patch
+'@commercetools-frontend/application-components': patch
+'@commercetools-frontend/application-shell': patch
+'@commercetools-frontend/i18n': patch
+'@commercetools-frontend/react-notifications': patch
+'playground': patch
+'@commercetools-local/visual-testing-app': patch
+'@commercetools-website/custom-applications': patch
+'@commercetools-website/components-playground': patch
+---
+
+Update `react-intl` to version `^5.25.0`
--- a/application-templates/starter/package.json
+++ b/application-templates/starter/package.json
@@ -61,7 +61,7 @@
     "prop-types": "15.8.1",
     "react": "17.0.2",
     "react-dom": "17.0.2",
-    "react-intl": "5.24.8",
+    "react-intl": "^5.25.0",
     "react-redux": "7.2.8",
     "react-router-dom": "5.3.0",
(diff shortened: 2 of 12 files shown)


CarlosCortizasCT:
I agree we need to update app-kit's `react-intl` version to use a **caret** range indicator, but I'm not sure if it could cause problems for users when updating.
If I currently have both `ui-kit` and 'app-kit` dependencies and update only `app-kit`, it will use version `5.25.0` but will not make `ui-kit` dependencies to update to this new versions so maybe I end up with different versions in my _node_modules_ directory and I might need to run `yarn up -R react-intl` to solve it.
If this is the case, perhaps we could add some instructions in the changeset file.

Or maybe I'm seeing a problem that actually would not exist.

kark:
> I agree we need to update app-kit's `react-intl` version to use a **caret** range indicator, but I'm not sure if it could cause problems for users when updating. If I currently have both `ui-kit` and 'app-kit`dependencies and update only`app-kit`, it will use version `5.25.0`but will not make`ui-kit`dependencies to update to this new versions so maybe I end up with different versions in my _node_modules_ directory and I might need to run`yarn up -R react-intl` to solve it. If this is the case, perhaps we could add some instructions in the changeset file.
> 
> Or maybe I'm seeing a problem that actually would not exist.

Thanks, I guess adding some instructions would be a nice help for the users but perhaps it will be unnoticed in this changeset in the long run. This has been brought up on several occasions in the github issues, i.e. [today](https://github.com/commercetools/ui-kit/issues/2158#issuecomment-1104817278). I can only guess that users will see it there, but I can add it also here it's totally up to you

emmenko:
We can create a github discussion and provide the answers.

CarlosCortizasCT:
Just because this already showed up a couple of times in the issues I was thinking about having some kind of instructions to our users which we could use as reference.

I think @emmenko's idea about having a github discussion with some instructions on how to debug this kind of problem in the best approach.

CarlosCortizasCT (approved):

emmenko (approved):

emmenko:
@CarlosCortizasCT @kark Do you want to work on the Q/A part in the GitHub Discussions? 

It can be something similar to this: https://github.com/commercetools/merchant-center-application-kit/discussions/2479

Essentially, you ask the question and answer to your own question 😄

PR #2565, opened 2022-04-22
feat(codemod): add codemod CLI with transform `rename-js-to-jsx`

emmenko:
(_tech time work_)

**Background**

From time to time, we need to programmatically migrate some stuff, mostly related to Custom Application changes.

In the past, we would do these migrations manually but in certain occasions we can try to automate them. This is usually done using a Codemod.

**Proposal**

Have a new package `codemod` (Next.js has something similar https://github.com/vercel/next.js/tree/canary/packages/next-codemod and https://nextjs.org/docs/advanced-features/codemods).

The first transform is `rename-js-to-jsx`, which helps renaming JS files using React (JSX) to `.jsx` extension. This is required for some future work that will require having proper file extensions `.jsx`.


packages/codemod/bin/cli.js
```diff
@@ -0,0 +1,50 @@
...
+  process.exit(0);
+}
+
+const [transform, globPattern] = commands;
+
+const files = glob.sync(globPattern);
+
+switch (transform) {
+  case 'rename-js-to-jsx':
+    break;
```
tdeekens, line 33: Caught me for a minute here but I guess it's just a `throwIfUnknownTransform` sort of.

tdeekens (approved):

packages/codemod/bin/cli.js
CarlosCortizasCT: I was also scratching my head for a couple of minutes 😄

tdeekens: Then maybe dumping this into such named function isn't the worst idea 😄.

emmenko: Yes it's just to validate that the command is valid, since we're using the name to read from file system.

Should I make this more explicit?

CarlosCortizasCT: My suggestion would be something like this:

```
const KNOWN_TRANSFORMATIONS = [
  'rename-js-to-jsx'
];

if (!KNOWN_TRANSFORMATIONS.includes(transform) {
  throw new Error(`Unknown transform ${transform}.`);
}

Comment (tdeekens) [reply packages/codemod/bin/cli.js:33]: Yes, like

```js
function throwIfTransformIsUnknown() {
const KNOWN_TRANSFORMATIONS = [
  'rename-js-to-jsx'
];

if (!KNOWN_TRANSFORMATIONS.includes(transform) {
  throw new Error(`Unknown transform ${transform}.`);
}
}
```

packages/codemod/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "publishConfig": {
+    "access": "public"
+  },
+  "files": ["build", "package.json", "LICENSE", "README.md"],
+  "bin": {
+    "mc-codemod": "./bin/cli.js"
+  },
+  "dependencies": {
+    "glob": "8.0.1",
+    "is-git-clean": "1.1.0",
```
kark, line 23: I am wondering why are we adding this as a dependency?

emmenko: Ah yeah I wanted to implement this check but decided to omit it for now. I'll remove the dependency. Thanks for checking.

.changeset/config.json
```diff
@@ -7,7 +7,13 @@
     }
   ],
   "commit": false,
-  "linked": [["@commercetools-backend/*", "@commercetools-frontend/*"]],
+  "linked": [
+    [
+      "@commercetools-backend/*",
+      "@commercetools-frontend/*",
+      "!@commercetools-frontend/codemod"
```
emmenko, line 14: Btw, this essentially decouples the codemod package versions from the rest.

I think for codemods we can keep its own version.

package.json
```diff
@@ -40,8 +40,8 @@
     "template-starter:build": "yarn workspace merchant-center-application-template-starter build",
     "template-starter:start": "yarn workspace merchant-center-application-template-starter start",
     "template-starter:start:prod:local": "yarn workspace merchant-center-application-template-starter start:prod:local",
-    "build": "yarn compile-intl && preconstruct build && NODE_ENV=production yarn workspace @commercetools-frontend/mc-html-template build && NODE_ENV=production yarn workspace @commercetools-frontend/mc-scripts build",
-    "build:watch": "preconstruct watch & NODE_ENV=development yarn workspace @commercetools-frontend/mc-html-template build:bundles:watch && NODE_ENV=development yarn workspace @commercetools-frontend/mc-scripts build:bundles:watch",
```
emmenko, line 44: It's getting a bit messy here, I moved this into a bash script now

packages/codemod/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "publishConfig": {
+    "access": "public"
+  },
+  "main": "./build/cli.js",
+  "files": ["bin", "build", "package.json", "LICENSE", "README.md"],
+  "bin": {
+    "mc-codemod": "./bin/mc-codemod.js"
+  },
+  "scripts": {
+    "build": "rimraf build && tsc",
```
emmenko, line 23: We now compile the CLI using `tsc`.

packages/codemod/src/cli.ts
```diff
@@ -0,0 +1,72 @@
...
+    transformPath: string,
+    filePaths: string[],
+    options: TRunnerOptions
+  ) => {
+    await Runner.run(transformPath, filePaths, options);
+  };
+
+  const execute = async () => {
+    switch (transform) {
+      case 'rename-js-to-jsx': {
```
emmenko, line 44: FYI: I kept the switch, every new command can be simply added to this case.

```js
case 'rename-js-to-jsx'
case 'another':
case 'something-else': {
}
```

emmenko:
FYI: the package now is written in TS and compiled to JS using `tsc`.

We can't use preconstruct (to create one single bundle) because jscodeshift expects the transformer files to be actual files.

CarlosCortizasCT (approved):
🚀

kark (approved):

PR #2567, opened 2022-04-25
feat(application-config): Enforce uniqueness for uriPath of the submenu links.

Rhotimee:
#### Summary
SubmenuLinks are meant to be "unique" based on the `uriPath`
An error is thrown if there are duplicate uriPath in the submenuLinks. 

[SHIELD-469](https://jira.commercetools.com/browse/SHIELD-469)

### DIFF
--- a/.changeset/bright-hornets-enjoy.md
+++ b/.changeset/bright-hornets-enjoy.md
@@ -2,4 +2,4 @@
 '@commercetools-frontend/application-config': patch
 ---
 
-- Enforce uniqueness for `uriPath` of the submenu links.
+Enforce uniqueness for `uriPath` of the submenu links.
--- a/.changeset/bright-hornets-enjoy.md
+++ b/.changeset/bright-hornets-enjoy.md
@@ -1,6 +1,5 @@
 ---
-'@commercetools-frontend/application-config': minor
+'@commercetools-frontend/application-config': patch
 ---
 
-SubmenuLinks are meant to be "unique" based on the `uriPath`
-An error is thrown if there are duplicate uriPath in the submenuLinks.
+- Enforce uniqueness for `uriPath` of the submenu links.
--- a/.changeset/bright-hornets-enjoy.md
+++ b/.changeset/bright-hornets-enjoy.md
@@ -1,0 +1,6 @@
+---
+'@commercetools-frontend/application-config': minor
+---
+
+SubmenuLinks are meant to be "unique" based on the `uriPath`
+An error is thrown if there are duplicate uriPath in the submenuLinks.
--- a/packages/application-config/src/validate-config.ts
+++ b/packages/application-config/src/validate-config.ts
@@ -1,6 +1,7 @@
 import type { JSONSchemaForCustomApplicationConfigurationFiles } from './schema';
 import Ajv, { type ErrorObject } from 'ajv';
 import schemaJson from '../schema.json';
+import type { CustomApplicationData } from './types';
 
 type ErrorAdditionalProperty = ErrorObject<
   'additionalProperty',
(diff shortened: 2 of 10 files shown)


packages/application-config/src/validate-config.ts
```diff
@@ -9,6 +9,8 @@ type ErrorAdditionalProperty = ErrorObject<
 type ErrorEnum = ErrorObject<'enum', { allowedValues: string[] }>;
 
 const ajv = new Ajv({ strict: true, useDefaults: true });
+require('ajv-keywords')(ajv, 'uniqueItemProperties');
```
emmenko, line 12: Hmm I'm not sure we can 100% rely on this validation at the schema level, especially when values are provided using environment variables (env placeholders).

I think we need to validate these kind of things after parsing the config and resolving the values. For instance, in the `transformCustomApplicationConfigToData`.

emmenko (commented):
Nice idea extending the schema with the `ajv-keywords` (didn't know about it).

However, I believe it does not fully cover our requirements.

tdeekens:
Maybe worthwhile double checking internally. I think one app alternates between the same sub menu right now using toggles.

emmenko:
> I think one app alternates between the same sub menu right now using toggles.

Hmm I double checked. I guess you are referring to product types / attribute groups. From what I understand, the main reason for using the same link with different toggles is to have different labels.

We want to enforce uniqueness so I guess we would need to change the workaround. I'll check with the team responsible for that.

tdeekens:
> different toggles is to have different labels.

Yes. It is temporary. Probably best to have a chat with them.

packages/application-config/src/validate-config.ts
Rhotimee: Oh. I didn't think of that. 
I believe it makes more sense to do the validation in `transformCustomApplicationConfigToData `then`.

.changeset/bright-hornets-enjoy.md
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools-frontend/application-config': minor
```
emmenko, line 2: I wouldn't consider this a feature but rather a fix.

```suggestion
'@commercetools-frontend/application-config': patch
```

Nit: also adjust the PR title please, thanks

.changeset/bright-hornets-enjoy.md
```diff
@@ -0,0 +1,6 @@
...
+---
+
+SubmenuLinks are meant to be "unique" based on the `uriPath`
+An error is thrown if there are duplicate uriPath in the submenuLinks.
```
emmenko, line 6:
```suggestion
Enforce uniqueness for `uriPath` of the submenu links.
```

packages/application-config/src/validate-config.ts
```diff
@@ -46,3 +48,17 @@ const validateConfig = (
 };
 
 export default validateConfig;
+
+export const validateSubmenuLinks = (
```
emmenko, line 52: Nit, if we export multiple things it's better to avoid using `default`. Please adjust the export for `validateConfig` to be `export const validateConfig`.

Also, we can maybe rename the file to `validations.ts` instead of `validate-config.ts`.

emmenko (commented):
Thanks!

CarlosCortizasCT (approved):
💯

packages/application-config/src/validations.ts
```diff
@@ -45,4 +47,16 @@ const validateConfig = (
   }
 };
 
-export default validateConfig;
+export const validateSubmenuLinks = (
+  submenuLinks: CustomApplicationData['submenuLinks']
```
emmenko, line 51: Nit: maybe you can apply the changes here as well, to pass the Custom App object instead of the submenu links (same as in the services)

emmenko (approved):
Thanks, looks good now 👌

packages/application-config/src/validations.ts
Rhotimee: updated

kark (approved):

ddouglasz (approved):

PR #2568, opened 2022-04-25
feat(cli): add opt-in experimental support for Vite bundler

emmenko:
(tech time work)

**Background**

The `mc-scripts` CLI uses Webpack as the bundler for development and for building the production assets.

Webpack traditionally requires a lot of configuration and it's also not one of the fastest and more performant tools, especially in big projects. This usually results in a frustrating developer experience.

**Proposal**

Vite takes a different and more modern approach in bundling your application, primarily around new features like native ES modules.

Why Vite? Read here: https://vitejs.dev/guide/why.html

This PR adds opt-in support (by setting the environment variable `ENABLE_EXPERIMENTAL_VITE_BUNDLER="true"`) to switch the bundler used for developing and building production bundles.

All the existing CLI commands fully work the same as before, so there is no migration step required from users to use the new bundle.

If you want to try it out, simply turn on the flag and enjoy 🚀

---

For comparison: build time seems to be cut in half on average (left with Webpack, right with Vite). 😎

image


.changeset/fair-poets-watch.md
```diff
@@ -0,0 +1,11 @@
+---
+'@commercetools-frontend/mc-dev-authentication': minor
+'@commercetools-frontend/mc-html-template': minor
+'@commercetools-frontend/mc-scripts': minor
+---
+
+Enable opt-in support for experimental bundler (using [Vite.js](https://vitejs.dev/)). To enable it, set the environment variable `ENABLE_EXPERIMENTAL_BUNDLER="true"` in your dotenv file.
```
tdeekens, line 7:
```suggestion
Enable opt-in support for experimental bundler (using [Vite.js](https://vitejs.dev/)). To enable it, set the environment variable `ENABLE_EXPERIMENTAL_VITE_BUNDLING=true"` in your dotenv file.
```

Just as we might have the flag for a bit to make it really clear what it turns on.

tdeekens (commented):
Really promising. Thanks for spending time on this!

Just as a general question: is your vision for the `mc-scripts` and bundling in general that app-kit always supports _the_ (or one) most commonly used bundler while from time to time we transition to it or would you also see supporting different bundlers at the same time on the long-term?

emmenko:
> is your vision for the `mc-scripts` and bundling in general that app-kit always supports the (or one) most commonly used bundler while from time to time we transition to it or would you also see supporting different bundlers at the same time on the long-term?

Good question. I probably wouldn't support multiple different bundles. At least I don't see a need for that.
As long as we can abstract away (most of) the implementation details (the `mc-scripts` CLI usage remains the same) it does not matter much what bundler we use internally. 

So in this case, if we are happy with Vite we will then make it the default and remove the Webpack stuff.

tdeekens:
> > is your vision for the `mc-scripts` and bundling in general that app-kit always supports the (or one) most commonly used bundler while from time to time we transition to it or would you also see supporting different bundlers at the same time on the long-term?
> 
> 
> 
> Good question. I probably wouldn't support multiple different bundles. At least I don't see a need for that.
> 
> As long as we can abstract away (most of) the implementation details (the `mc-scripts` CLI usage remains the same) it does not matter much what bundler we use internally. 
> 
> 
> 
> So in this case, if we are happy with Vite we will then make it the default and remove the Webpack stuff.
> 
> 
> 
> 

I frankly also do not see the need. I just remembered us having a discussion around it being an option a while ago. Supporting multiple in the end costs maintenance and it would be hard to weigh the trade offs and allow users make informed decisions. So I think recommending one is probably easiest for both sides.

packages/mc-dev-authentication/middlewares/create-mc-dev-authentication-middleware.js
```diff
@@ -0,0 +1,84 @@
...
+const { logout } = require('../routes');
+
+const compileLoginView = pug.compileFile(
+  path.join(__dirname, '../views/login.pug')
+);
+const compileLogoutView = pug.compileFile(
+  path.join(__dirname, '../views/logout.pug')
+);
+
+function createMcDevAuthenticationMiddleware(applicationConfig) {
```
emmenko, line 12: This is an "aggregated" middleware for handling login/logout requests from developing against the local MC API.

This is only useful for internal teams.

packages/mc-dev-authentication/middlewares/create-mc-dev-authentication-middleware.js
```diff
@@ -0,0 +1,84 @@
...
+  const htmlLogin = compileLoginView({ env: applicationConfig.env });
+  const htmlLogout = compileLogoutView({ env: applicationConfig.env });
```
emmenko, line 14: Instead of relying of writing/reading these pages from disk, we can simply handle it in memory.

Here we compile the pages and serve them when needed.

packages/mc-dev-authentication/transformer-local.js
```diff
@@ -24,18 +24,6 @@ module.exports = ({ env }) => {
...
-    path.join(paths.appBuild, 'login.css')
-  );
-  fs.copyFileSync(
-    path.join(__dirname, 'views', 'login.js'),
-    path.join(paths.appBuild, 'login.js')
-  );
-  fs.copyFileSync(
-    path.join(__dirname, 'views', 'logout.js'),
-    path.join(paths.appBuild, 'logout.js')
-  );
```
emmenko, line 38: This was never really necessary, as the CSS and JS are **included** in the HTML page.

packages/mc-html-template/html-docs/application.html
```diff
@@ -0,0 +1,72 @@
+<!DOCTYPE html>
```
emmenko, line 1: Moving this in to a proper HTML file makes it a bit "easier" to maintain and read (syntax highlighting) instead of being a huge block of string.

packages/mc-html-template/html-scripts/public-path.js
```diff
@@ -0,0 +1,11 @@
...
+  return `${window.app.cdnUrl.replace(/\/$/, '')}/${importer.replace(
+    /^(\.\/)?/,
+    ''
+  )}`;
+};
+window.__dynamicImportPreload__ = function (preloads) {
+  return preloads.map(
+    (preload) => `${window.app.cdnUrl.replace(/\/$/, '')}/${preload}`
+  );
+};
```
emmenko, line 11: This is the equivalent of Webpack's `__webpack_public_path__`.

packages/mc-html-template/src/generate-template.js
```diff
@@ -1,82 +1,16 @@
...
-  const cssImports = cssChunks.map(
-    (chunkPath) =>
-      `<link href="__CDN_URL__${chunkPath}" rel='stylesheet' type='text/css'></link>`
-  );
-  const scriptImports = scriptChunks.map(
-    (chunkPath) => `<script src="__CDN_URL__${chunkPath}"></script>`
-  );
-
-  return `
-<!DOCTYPE html>
```
emmenko, line 14: Move to a separate HTML file.

packages/mc-html-template/src/load-html-docs.js
```diff
@@ -0,0 +1,21 @@
+// @preval
+/**
+ * https://github.com/kentcdodds/babel-plugin-preval#preval-file-comment--preval
+ * NOTE: This file is pre-evaluated during build time, using `babel-plugin-preval`.
+ * This is ok as the loaded files are static anyway and it prevents possible
+ * loading issues when files are required through Webpack own context.
+ */
```
emmenko, line 7: This is important and it's something we could have done some time ago.

Essentially instead of shipping the actual HTML, CSS, JS snippets, we bundle them inline on compile time.

This is how this file looks:

image

packages/mc-scripts/src/commands/build-vite.js
```diff
@@ -0,0 +1,134 @@
...
+          {
+            tag: 'script',
+            // Inject the functions to dynamically change the public path.
+            // This is also used for the `cdnUrl` and is the equivalent
+            // of Webpack's `__webpack_public_path__`.
+            children: htmlScripts.publicPath,
+            injectTo: 'body',
+          },
+        ],
+      };
```
emmenko, line 45: This transformer is important and it allows to prepend the asset URLs with the `cdnUrl` value.

packages/mc-scripts/src/commands/build-vite.js
```diff
@@ -0,0 +1,134 @@
...
+    // Define the module entry point (path relative from the `/public` folder).
+    // NOTE: that this is different from the development configuration.
+    scriptImports: [`<script type="module" src="/${appEntryPoint}"></script>`],
+  });
+  // Write `index.html` (template) into the `/public` folder.
+  fs.writeFileSync(paths.appIndexHtml, html, { encoding: 'utf8' });
+
+  // TODO: allow to pass additional config options.
+  // * `define`
+  // * `plugins`
```
emmenko, line 66: This can be done as a follow up.

packages/mc-scripts/src/commands/start.js
```diff
@@ -66,8 +66,6 @@ choosePort(HOST, DEFAULT_PORT)
     });
     // Serve webpack assets generated by the compiler over a web sever.
     const serverConfig = createDevServerConfig({
-      allowedHost: urls.localUrlForBrowser,
-      contentBase: paths.appBuild,
```
emmenko, line 70: Those weren't really used.

tdeekens (commented):
Super nice. Did a first review. Not much to add. Will however need another round with a fresh mind tomorrow.

emmenko:
TODOs:
* Rebase PR and resolve conflicts.
* Some assets are not correctly loaded in production build (see VRTs).

packages/mc-scripts/src/commands/build-vite.js
```diff
@@ -0,0 +1,134 @@
...
+  return {
+    name: 'custom-application',
+    /**
+     * @type {import('vite').IndexHtmlTransformHook}
+     */
+    transformIndexHtml(rawHtml, _ctx) {
+      // Ensure to use the `cdnUrl` value when loading the entry point.
+      // NOTE: with Webpack this is done by setting the `__webpack_public_path__`.
+      const html = rawHtml.replace(
+        new RegExp(`<script type="module"(.*) src="(.*)">`, 'g'),
```
tdeekens, line 29: I guess alternatively we could parse the HTML which might be more expensive. Maybe defining `function prependCdnUrlToScriptSrc` would help. At least I got stuck here deducting that :)

tdeekens (commented):
Thanks again. Had a second look:

1. I really like the addition of `preval` it makes a lot of sense
2. I wonder if we want to drop `pug` I don't see why we need it
3. We might use es6-template-strings - no direct benefit but avoids the `__` replacements while we're here

As a note: We don't use JavaScript modules as `type="module"` yet. Support wise (not IE) we should. Still, it's to how I know and tell a change with Vite which might be worthwhile also documenting.

Lastly, I think the hardest part of understand in the vite setup is the asset and index.html re-routing but from what I can tell there is little we can do about it. I wonder just if it's fragile and could have bugs we don't know about yet.

packages/mc-dev-authentication/middlewares/create-mc-dev-authentication-middleware.js
```diff
@@ -0,0 +1,84 @@
+const path = require('path');
+const pug = require('pug');
+const { logout } = require('../routes');
+
+const compileLoginView = pug.compileFile(
```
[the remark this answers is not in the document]
emmenko: I don't have a strong reason to keep using `pug` either. What would be the alternative? Plain old HTML?

emmenko:
> I really like the addition of preval it makes a lot of sense

Yeah that's kind of neat.

> As a note: We don't use JavaScript modules as type="module" yet. Support wise (not IE) we should. Still, it's to how I know and tell a change with Vite which might be worthwhile also documenting.

You mean mentioning (changelog, etc.) that it uses `<script type="module"`?

> Lastly, I think the hardest part of understand in the vite setup is the asset and index.html re-routing but from what I can tell there is little we can do about it. I wonder just if it's fragile and could have bugs we don't know about yet.

Yes unfortunately there is no built-in support for this ([yet](https://github.com/vitejs/vite/issues/3522#issuecomment-874377284)). I thought about actually dropping support for the `cdnUrl` property but it's still something useful to have I suppose.

The alternative is to set the `config.base` to point to the `cdnUrl` but it would mean no more "compile-html" step, so you need to rebuild the app every time something in the config changes.

tdeekens:
> You mean mentioning (changelog, etc.) that it uses <script type="module"?

Yes. Just so people are aware. I don't think app-kit has a strict browser policy at least I am not aware of any. However, all existing script tags are not of `type=module` to my knowledge and with using Vite would become unless I am mistaken.

> Yes unfortunately there is no built-in support for this (https://github.com/vitejs/vite/issues/3522#issuecomment-874377284). I thought about actually dropping support for the cdnUrl property but it's still something useful to have I suppose.

I think it makes sense to keep the status quo even if that means the internals of the commands are a bit wonky. As a later step we could discuss removing some of that functionality if possible if we want to simplify those internals then.

emmenko:
Actually there is an issue with the setup for the `cdnUrl`, as I had to set the `config.base` to `''` (empty string) but this is causing other issues (see VRTs).

So I'll have to rethink how to handle that 🧐

emmenko:
So it seems that this is more tricky than I thought. The good news is that there is a general consensus on supporting this feature (_configuring the `base` path on runtime_). You can follow a bit the thread here: https://github.com/vitejs/vite/discussions/7611#discussioncomment-2500939

So for now, I would keep the setup simpler and avoid using the `cdnUrl`. This means that when using Vite, all assets must be served relative to the application `index.html`.
A warning will be logged to mention this.

tdeekens:
> So for now, I would keep the setup simpler and avoid using the cdnUrl. This means that when using Vite, all assets must be served relative to the application index.html.

I think that's a reasonable restriction then until Vite supports it natively.

emmenko:
Alright, see changes in the last commit (https://github.com/commercetools/merchant-center-application-kit/pull/2568/commits/e2396e115f627215d182954f05e08df0353a9840) with the fixes and improvements.

packages/mc-dev-authentication/middlewares/create-mc-dev-authentication-middleware.js
```diff
@@ -0,0 +1,76 @@
...
+      ) {
+        if (request.originalUrl.startsWith('/login/authorize')) {
+          if (isDevAuthenticationMiddlewareDisabled) {
+            next();
+          } else {
+            response.end(htmlLogin);
+          }
+          return;
+        }
+      }
```
CarlosCortizasCT, line 50: Maybe I'm missing something but it seems here if we do have an `authorizeUrl` in the application configuration but it does not start with `http://localhost` or the `originalUrl` does not start with `/login/authorize` we will not call `next()` nor finish the `response` so we could end up with a hanging request.

Perhaps we should just let the request continue in those cases?
```suggestion
      if (
        applicationConfig.env.__DEVELOPMENT__?.oidc?.authorizeUrl.startsWith(
          'http://localhost'
        )
      ) {
        if (request.originalUrl.startsWith('/login/authorize')) {
          if (isDevAuthenticationMiddlewareDisabled) {
            next();
          } else {
            response.end(htmlLogin);
          }
          return;
        } else {
          next();
        }
      } else  {
        next();
      }
```

emmenko: In line L72 we do call `next()` as a fallback. Isn't that enough? 🤔

CarlosCortizasCT: Yes, it is.

Sorry.

emmenko: All good, thanks for double checking 🤗

packages/mc-scripts/src/bin/cli.js
```diff
@@ -61,19 +61,34 @@ const applicationDirectory = fs.realpathSync(process.cwd());
...
         process.env.NODE_ENV = 'production';
 
         const shouldAlsoCompile = !flags['build-only'];
+        const shouldUseExperimentalBundler =
+          process.env.ENABLE_EXPERIMENTAL_VITE_BUNDLER === 'true';
+        if (shouldUseExperimentalBundler) {
+          console.log('Experimental Vite bundler enabled! 🚀');
+          console.warn(
+            'NOTE that the "cdnURL" value is not supported at the moment when using Vite.'
+          );
```
CarlosCortizasCT, line 81: 💯

PR #2578, opened 2022-04-29
refactor: migrate `mc-dev-authentication` package to TypeScript

emmenko:
(_tech time work_)

The package now is now written in TypeScript and being built using preconstruct.


packages/mc-dev-authentication/html-docs/login.html
```diff
@@ -0,0 +1,228 @@
+<html>
```
emmenko, line 1: As suggested by @tdeekens in another PR, we can using a plain old HTML file for these pages.

packages/mc-dev-authentication/html-docs/login.html
```diff
@@ -0,0 +1,228 @@
...
+            payload.nonce = queryParams.get('nonce');
+          }
+
+          var container = document.getElementById('errors');
+          // Clean up error message elements
+          while (container.firstChild) {
+            container.removeChild(container.firstChild);
+          }
+
+          const url = '__MC_API_URL__/tokens';
```
emmenko, line 171: This is the only part that "changed". We now simply do a string replacement for this variable.

packages/mc-dev-authentication/html-docs/login.html
tdeekens: Thanks a lot. I think this is easier at least for my lost soul :)

packages/mc-dev-authentication/src/transformer-local.ts
```diff
@@ -0,0 +1,36 @@
...
+const paths = {
+  appBuild: resolveApp('public'),
+};
+
+// This transformer will generate a development `login` and `logout` HTML files
+// and copy them to the application public foder.
+// This is necessary to run the application locally in production mode.
+const transformerLocal = (compiledHtml: TCompiledHtml) => {
+  const htmlLogin = pages.loginPage.replace(
+    new RegExp('__MC_API_URL__', 'g'),
```
tdeekens, line 23: Will always work but still wonder if eventually we should use a very tiny templating library. I guess benefits are almost not there yet though.

tdeekens (approved):

tdeekens (approved):

packages/mc-dev-authentication/src/pages.js
```diff
@@ -0,0 +1,17 @@
...
+      encoding: 'utf8',
+    }
+  );
+  return content;
+};
+
+module.exports = {
+  loginPage: loadHtmlDocAsString('login.html'),
+  logoutPage: loadHtmlDocAsString('logout.html'),
+};
```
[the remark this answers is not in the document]
CarlosCortizasCT: I guess this file was not migrated to TS because babel macro don't support precompiling typescript files. Is that right?

packages/mc-dev-authentication/src/transformer-local.ts
```diff
@@ -0,0 +1,36 @@
...
+// https://github.com/facebook/create-react-app/issues/637
+const appDirectory = fs.realpathSync(process.cwd());
+const resolveApp = (relativePath: string) =>
+  path.resolve(appDirectory, relativePath);
+const paths = {
+  appBuild: resolveApp('public'),
+};
+
+// This transformer will generate a development `login` and `logout` HTML files
+// and copy them to the application public foder.
```
CarlosCortizasCT, line 19:
```suggestion
// and copy them to the application public folder.
```

CarlosCortizasCT (approved):
💯

packages/mc-dev-authentication/src/pages.js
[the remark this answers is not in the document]
emmenko: Yes correct

PR #2579, opened 2022-04-29
refactor: migrate `mc-html-template` package to TypeScript

emmenko:
(tech time work)

The package now is now written in TypeScript and being built using preconstruct.


tdeekens (approved):
Thanks. I assume the `js` files where mainly renamed to `ts` and types got added and nothing changed internally much.

emmenko:
Yes, and the preval setup changed slightly (instead of `@preval` annotation, we use `/* preval */` before the `import`).

packages/mc-html-template/src/replace-html-placeholders.ts
```diff
@@ -1,22 +1,33 @@
...
+// https://babeljs.io/blog/2017/09/11/zero-config-with-babel-macros
+import htmlScripts from /* preval */ './load-html-scripts';
+// https://babeljs.io/blog/2017/09/11/zero-config-with-babel-macros
+import htmlStyles from /* preval */ './load-html-styles';
 
-const trimTrailingSlash = (value) => value.replace(/\/$/, '');
+type TReplaceHtmlPlaceholdersOptions = {
+  env: ApplicationRuntimeConfig['env'];
+  headers: Record<string, string | undefined>;
+};
```
CarlosCortizasCT, line 11: Previously we supported another prop called `cliFlags` I guess we remove because it wasn't actually used.
Perhaps we need to update the README since it was mentioned [over there](https://github.com/commercetools/merchant-center-application-kit/blob/ef163df49d3ba04f05e2b745a5b8b7f7203a1dc5/packages/mc-html-template/README.md?plain=1#L23).

emmenko: Good point about the README, I'll update that.

PR #2581, opened 2022-05-02
feat: introducing the Stacking Layer System

### LINKED ISSUE #864: Define constants for the z-index of Modals and Dialogs
### Is your feature request related to a problem? Please describe
Following the implementation and usage of the new ModalPages, already some cases have been found where Dialogs are being displayed behind ModalPages, because they have a lower `z-index`, and they need to be set manually for each case in order to be fixed, but because there are a lot of different combinations of stacked ModalPages and Dialogs that can show in one place or another, setting the perfect values everywhere may be a difficult task.

### Describe the solution you'd like
I was thinking of defining constants for these values, shared across all apps of our MC, but I've realized this should first be defined here on the app-kit, as it could save a lot of work just by **having good defaults**.

Currently, both ModalPages and Dialogs have a default `z-index` of `1000`, and ModalPages will stay on top because they increment `1` to the base z-index per each unit of `level`, which means a level 1 ModalPage (the default) will start with z-index of `1001` and that's why it stays on top of the default Dialog.

I think it's reasonable to say that a Dialog should be on top of a ModalPage in the vast majority of cases, as a ModalPage is simply a page which can trigger other ModalPages or Dialogs on top of it, while a Dialog usually is something that appears on top of everything and usually doesn't trigger opening another Page on top of it (though it might happen, and one can set the prop manually for such cases).
That being said, I believe we should change the default values of the `z-index` for ModalPages and Dialogs so that the latter have a much higher value than the former.
* ModalPages could stay at `1000`, and with their stacking mechanic they increase by a unit with each level (1001...1002...1003, etc). So this stays the same.
* Dialogs should have a default considerably higher than that, to make sure they are rendered on top of ModalPages even if there are several ones stacked on top of each other.
Could be a z-index of `2000`.

What do you think?

emmenko:
Closes #864 

Introducing the **Stacking Layer System**.

# Background

Components such as modal pages, dialogs, etc. are rendered using a "modal" container. These containers are then rendered within a special container called `portals-container`.

Up until now, rendering these components required to define things like `zIndex` or `level` props, to imperatively determine how the component will be visible.
This was required as the modal containers are positioned `absolute` and finding the correct `z-index` value is important.

However, it's the responsibility of the developer to "pick" the correct values which is error prone. In fact, choosing a wrong `z-index` results in the modal to not be visible and thus leading to UI bugs.

A better and more reliable approach would be for the Custom Application to automatically determine the correct `z-index` values for every modal container rendered on the page.

# Stacking Layer System

To solve this issue, a Custom Application now implements a **Stacking Layer System** to automatically determine and apply the correct `z-index` values for every modal container.

Therefore, it is not necessary anymore to explicitly provide the `zIndex` and `level` props to the modal pages or dialog components. The following props have been deprecated: `level` and `baseZIndex` (modal pages).

For backwards compatibility, the `zIndex` prop is still supported and, if defined, it will overwrite the `z-index` value using `!important`. Therefore we recommend to only define it if absolutely necessary, otherwise it's safe to remove it.

# Implementation

The system works by calculating the layer values based on the child nodes of the portals container.

For instance, there are two values that need to be calculated:
* `stackingLevel`: a sequential number used to apply the `z-index` value of each new child node. Example:
  image
  image

* `indentationLevel`: a sequential number based on the number of modal pages rendered, to apply the indentation width of each modal page. Example:
  image


packages/application-components/src/components/dialogs/internals/dialog-container.tsx
```diff
@@ -27,18 +27,28 @@ const getDefaultParentSelector = () =>
...
+    {contentElement}
+  </div>
+);
+
 type Props = {
   isOpen: boolean;
   onClose?: (event: SyntheticEvent) => void;
   size: 'm' | 'l' | 7 | 8 | 9 | 10 | 'scale';
-  zIndex: number;
+  zIndex?: number;
```
tdeekens, line 45: Are there actual use cases to set it explicitly given the automatic mechanisms implemented here? Should we deprecate it maybe?

packages/application-components/src/components/modal-pages/custom-form-modal-page/custom-form-modal-page.tsx
```diff
@@ -22,12 +22,18 @@ type MessageDescriptor = {
 };
 type Label = string | MessageDescriptor;
 type Props = {
+  /**
```
tdeekens, line 25: Ah, sorry. Disregard comment above.

packages/application-components/src/components/portals-container/portals-container.tsx
```diff
@@ -16,6 +18,7 @@ type TObservableElementDimensions = {
   height: number;
   width: number;
 };
+type TStackingLayer = { stackingLevel: number; indentationLevel: number };
```
tdeekens, line 21: Nit: Maybe add a comment what each intends to do. Otherwise requires some information gathering in code.

tdeekens (commented):
Thanks. Had a really brief look in between things. I really like that this would be managed internally!

A question as we here use mutation observers to "be notified" of when a new modal appears to be stacked. Would this somehow we also in the mounting of reach modal to inform a parent context that a new level has been added?

emmenko:
> A question as we here use mutation observers to "be notified" of when a new modal appears to be stacked. Would this somehow we also in the mounting of reach modal to inform a parent context that a new level has been added?

Initially I went for a React Context approach where components will subscribe to the context and render the `z-index` value according to some logic.
However, it's not that simple because the component needs to know the "stacking level" specific for that component. Which means that the context needs to know somehow of the component or have some kind of reference in order to select the correct value.

At the end of the day it's also a matter of responsibility and we can think of it as a parent-child relation: the parent has all the logic and the child simply consumes the injected values.

Since most of the logic of layouts for modals is anyway defined in the portals container, I opted of using a more global approach using CSS so that components don't need to deal with the context and the logic part.

Hopefully this makes a bit more sense.

tdeekens:
Thanks for sharing. Makes sense. I see the issues with a context and agree that this approach might be more fitting.

packages/application-components/src/components/portals-container/portals-container.tsx
```diff
@@ -79,48 +114,110 @@ const useObserverElementDimensions = (
...
+                }
+                .ReactModalPortal[data-level='${stackingLayer.stackingLevel}']
+                  [data-role='modal-overlay']
+                  [role='dialog'] {
+                  width: calc(
+                    100% -
+                      (${indentationSize} * ${stackingLayer.indentationLevel})
+                  );
+                }
+              `
```
CarlosCortizasCT, line 214: Something that confused me is that we are creating and so updating some global css rules in the document every time we add/remove a new `overlay` to then link them to the actual HTML elements with a class.
My first impression was that this could be done inline. Instead of adding a new data attribute to the overlay elements, just add to them the `z-index` and `width` css rules inline. Perhaps with that approach we could also remove the local state since we are resetting it on every element added to the `PortalsContainer`.

Maybe I could try to explain myself better when if have a talk about this PR.

emmenko: Yes I know what you mean. Initially I tried to do that but ended up here instead.

I'll give it another try, thanks for suggesting that!

emmenko: Ok I double checked and then I remembered the reason why I used CSS.

There are cases were modal components are rendered but they are not "opened/active" yet, for example a confirmation dialog is rendered on the page but it hasn't been opened yet.

However, the portal container of these components is rendered into the DOM even though it's empty. Then when the dialog is opened, it renders inside its portal container.

image


Because the container was already rendered, the mutation observer won't be notified again when the dialog actually opens. In theory we could also subscribe to the `subtree` changes but it will trigger a mutation for the entire subtree of nodes.

So by using pre-defined CSS selectors, when the dialog opens, the styles gets automaticaly applied.

CarlosCortizasCT: Ah! I see.

Thanks for the explanation.

packages/application-components/src/components/portals-container/portals-container.tsx
```diff
@@ -79,48 +114,110 @@ const useObserverElementDimensions = (
...
+            // Apply styles for stacking layers.
+            ...stackingLayers.map(
+              /**
+               * Every "overlay" container gets assigned a sequential `z-index` level.
+               * Furthermore, the "modal overlay" containers (not "dialog") get assigned
+               * the correct indentation level width.
+               */
+              (stackingLayer) => css`
+                .ReactModalPortal[data-level='${stackingLayer.stackingLevel}']
+                  [data-role$='overlay'] {
```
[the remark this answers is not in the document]
emmenko: What typo? 🤔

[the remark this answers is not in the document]
emmenko: In this case we're matching the `data-role` attributes that end with the value `overlay`, as we have `modal-overlay` and `dialog-overlay`.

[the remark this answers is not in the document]
kark: Right, thanks for the explanation :)

emmenko:
I think this is ready for a final review now, if we're all fine with the approach. Thanks

emmenko:
Ah, I have to fix the static VRTs also.

CarlosCortizasCT (approved):
👏 👏 👏

website/src/content/api-reference/commercetools-frontend-application-components.mdx
```diff
@@ -79,7 +79,7 @@ const ChannelsInfo = () => {
...
-| `zIndex`            | `number`   |    -     | -                 | `1000`  | The `z-index` value to be applied to the overlay container. It is useful if you have stacking modals.                                                                                                                                                                                                           |
+| `zIndex`            | `number`   |    -     | -                 | -  | The `z-index` value to be applied to the overlay container. By default the all modal containers are automatically assigned a correct `z-index` value. If this prop is defined, it will overwrite the assigned value.                                                                                                                                                                                                           |
```
ddouglasz, line 82: Really cool work to follow 🙌🏾. Just a tiny observation here and some other repeated places. 
```suggestion
| `zIndex`            | `number`   |    -     | -                 | -  | The `z-index` value to be applied to the overlay container. By default, all the modal containers are automatically assigned a correct `z-index` value. If this prop is defined, it will overwrite the assigned value.                                                                                                                                                                                                           |
```

tdeekens (approved):

ddouglasz (approved):
👏👏

kark (approved):

PR #2585, opened 2022-05-04
refactor: remove support for updating application Id

### LINKED ISSUE #2560: CLI `config:sync` command fails to update `.ts` version of the config file
**Describe the issue**

Using the CLI command `config:sync` may result in creating a new Custom Application. After that, the `applicationId` value in the Custom Application config file is updated with the newly created ID.

If the config file is a `.ts` file, the CLI fails as the file cannot be properly parsed.

**To Reproduce**

1. Update the config file to be a `.ts` file.
2. Change the `entryPointUriPath` to something random, so that a new Custom Application will be created.
3. Run the `mc-scripts config:sync` command and follow the prompts.
4. The CLI fails with a `SyntaxError`.

```
SyntaxError: /path/to/project/custom-application-config.ts: Unexpected token, expected "from" (2:12)

  1 | import { entryPointUriPath, PERMISSIONS } from './src/constants';
> 2 | import type { ConfigOptions } from '@commercetools-frontend/application-config';
    |             ^
  3 |
  4 | const config: ConfigOptions = {
```

**Expected behavior**

The CLI command works also for `.ts` files.

**Additional context**

Parsing the `.ts` file can be easily solved by making sure the Babel transform uses the typescript plugin. However, writing the updated AST to the config file results in loosing the TS syntax, as Babel transformed the TS to JS when parsing the AST.

Ideally, when parsing/writing `.ts` files we should use a TS parser that preserve the syntax.

If this is not possible, we would need to drop this functionality and print out to stdout the instructions to update the `applicationId`.

Rhotimee:
#### Summary
Remove support for updating application Id in the custom-application file. 

Closes #2560


emmenko (changes requested):
Thanks, I think it makes things simpler to drop this.

However, we need to also update the messages of the CLI to instruct the user to put the ID into the config.

website/src/content/api-reference/cli.mdx
```diff
@@ -128,7 +128,7 @@ Developers can use the `config:sync` command to automatically manage the configu
 
 If a new Custom Application needs to be created, an interactive prompt will ask the user to select the Organization where the Custom Application should be configured to.
 
-Additionally, the [Custom Application ID](/api-reference/application-config#envproductionapplicationid) is automatically synced when running the `config:sync` command.
+Additionally, after creating a Custom Application using the `config:sync` command, you should add the [Custom Application ID](/api-reference/application-config#envproductionapplicationid) generated to the application config file.
```
emmenko, line 131:
```suggestion
Additionally, after creating a new Custom Application using the `config:sync` command, you must add the generated [Custom Application ID](/api-reference/application-config#envproductionapplicationid) to the application config file, following the instructions in the terminal.
```

emmenko (commented):
Thanks, just some wording improvements otherwise 👌

emmenko (approved):
💯

CarlosCortizasCT (approved):
💯

PR #2586, opened 2022-05-05
fix(components): type declarations for `<PortalsContainer>`

emmenko:
Using `defaultProps` with `forwardRef` results in the type declarations to ignore the `defaultProps`. Therefore, the default props are typed as optional and we initialize the values with the default values within the component.


tdeekens (approved):

Rhotimee (approved):

CarlosCortizasCT (approved):
👍

PR #2588, opened 2022-05-05
fix(components): make props optional

emmenko:
Damn, I forgot to actually make the props optional.

Leftover of https://github.com/commercetools/merchant-center-application-kit/pull/2586

### DIFF
--- a/packages/application-components/src/components/portals-container/portals-container.tsx
+++ b/packages/application-components/src/components/portals-container/portals-container.tsx
@@ -21,27 +21,27 @@ type TPortalsContainerProps = {
    * The offset value for positioning the container from the top, when opened.
    * Usually this corresponds to the height of the header section.
    */
-  offsetTop: string;
+  offsetTop?: string;
   /**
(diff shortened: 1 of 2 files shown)


packages/application-components/src/components/portals-container/portals-container.tsx
```diff
@@ -21,27 +21,27 @@ type TPortalsContainerProps = {
...
    * Usually this corresponds to the width of the expanded nav menu.
    */
-  offsetLeftOnExpandedMenu: string;
+  offsetLeftOnExpandedMenu?: string;
   /**
    * The CSS selector to apply the `overflow: hidden` style to (preventing scrolling)
    * when a modal container is open.
    */
-  containerSelectorToPreventScrollingOnOpen: string;
+  containerSelectorToPreventScrollingOnOpen?: string;
```
CarlosCortizasCT, line 40: nit: Perhaps we could also include the default values in the jsdocs for these props (as it is done for the `zIndex` prop).

CarlosCortizasCT (approved):

PR #2593, opened 2022-05-05
PANGOLIN-1665-timezone-blocklist

ByronDWall:
#### Summary

This PR adds a `constants.ts` file to `packages/l10n/src`, and adds a `EXCLUDED_TIME_ZONES` const to that file.

#### Description
This is the first PR in a series to update the timezone list exposed by `packages/l10n` based on UX research/requirements.  

It is PR'ed into the `PANGOLIN-1253` branch, which is meant to be a feature trunk branch for the overall timezone feature-set.

For reference, the basis for this blocklist based on UX research [can be found here](https://docs.google.com/spreadsheets/d/1okm8x53VMdVV9G6joA6Eksb2369s0nepDbFfUuwpLZU/edit#gid=0)


packages/l10n/src/constants.ts
```diff
@@ -0,0 +1,298 @@
+import type { timeZone_blocklist } from './types';
+
+export const TIMEZONE_BLOCKLIST: timeZone_blocklist = [
```
emmenko, line 3: I would omit the `timeZone_blocklist` and keep it like this

```suggestion
export const TIMEZONE_BLOCKLIST = [
```

PS: type names usually are in PascalCase, so `timeZone_blocklist` should have been `TimeZoneBlockList`.

emmenko: Also, can you provide a comment explaining what this blocklist is for? Thanks

emmenko (commented):
Thanks, I left some minor suggestions.

packages/l10n/src/constants.ts
tdeekens: I think the type would be inferred. Also maybe naming it listOfUnsupportedTimeZones or excludedTimeZones might be clearer while adding a comment. 

@emmenko our idea was to not detach ourselves from moment's list entirely. As that also has drawbacks. As a result excluding felt more explicit and good common ground. I think a comment can help.

emmenko: Have you consider the other way around? So keeping an allow list only?

tdeekens: Yes couldn't find much yet. The flow after this is intended to be if you pull/generate TZs that you are notified via a shell output to either accept the new TZ or add it to the deny list. As such we want to make denying a conscious choice not accepting a new TZ.

ByronDWall: I've updated the constant to be name `EXCLUDED_TIME_ZONES`, removed the type, and added a comment as to the intention behind this constant.

Let me know if we need to talk further about this implementation.

emmenko (approved):

packages/l10n/src/constants.ts
```diff
@@ -0,0 +1,310 @@
...
+ * consumes this list of excluded timezones, and an object of locale-specific
+ * long-form translations.  If a timezone is present in the output of
+ * `moment.tz.names()`, but either is not in the excluded list or does not
+ * have a translation, that timezone is considered a new timezone.
+ *
+ * If a new timezone is encountered, the user running the CLI command
+ * will be prompted via the shell whether they would like to accept
+ * the new timezone, or add that timezone to this excluded list.
+ */
+export const EXCLUDED_TIME_ZONES = [
```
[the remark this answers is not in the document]
ByronDWall: currently they are sorted by gmt offset, but I can run an alphabetic sort on them if you'd prefer

tdeekens:
> Should I add a changeset to this PR, or to the larger PR that will occur when I PR the feature branch into main?

Both should work. You could also merge into a base to avoid the issue maybe.

packages/l10n/src/constants.ts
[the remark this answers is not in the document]
tdeekens: Let's move forward with this. We can reiterate on sorting later.

stephsprinkle (approved):
Confirmed that the excluded time zones list matches the list provided by UX

packages/l10n/src/constants.ts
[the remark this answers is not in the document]
emmenko: I think having them sorted alphabetically makes it easier to manage the list. Can be a follow up yes, just wanted to mention it.

jmcreasman (commented):
Confirmed the list too, looks like everything was included.

jmcreasman (approved):

Issue #2551, opened 2022-04-13
Implement `{Info,Form,CustomForm}DetailPage`


kark:
**Is your feature request related to a problem? Please describe.**

As a followup to #2499 we should implement components corresponding to `InfoModalPage`, `FormModalPage` and `CustomFormModalPage` but rendered without using a modal.

**Describe the solution you'd like**

Expose from the `@commercetools-frontend/application-components` package and document components:
- `InfoDetailPage`
- `FormDetailPage`
- `CustomFormDetailPage`

PR #2594, opened 2022-05-06
feat(detail-pages): add <{Info,Form,CustomForm}DetailPage> components

kark:
#### Summary

The aim of this PR was to implement new `<InfoDetailPage>`, `<FormDetailPage>` and `<CustomFormDetailPage>` components.
closes https://github.com/commercetools/merchant-center-application-kit/issues/2551


.changeset/old-otters-arrive.md
```diff
@@ -0,0 +1,154 @@
...
+These components are similar to the `<InfoModalPage>`, `<FormModalPage>` and `<CustomFormModalPage>` respectively but they are not rendered as modals.
+
+# Usage
+
+The detail pages are supposed to be used as a direct child of one of the main pages.
+The layout of those pages can be recognized by the gray background header and the white content background. A back link in the header section of each of the pages is optional.
+
+## InfoDetailPage
+
+Form Detail pages are controlled components used to render a page with detailed data.
```
emmenko, line 18:
```suggestion
Info Detail pages are controlled components used to render a page with detailed data.
```

packages/application-components/src/components/detail-pages/custom-form-detail-page/custom-form-detail-page.tsx
```diff
@@ -0,0 +1,145 @@
...
+  /**
+   * Determines if the form controls should be rendered.
+   */
+  hideControls?: boolean;
+
+  // PageTopBar props:
+  /**
+   * Makes page top bar visible
+   */
+  showPageTopBar?: boolean;
```
emmenko, line 69: Is this a new prop? What's the purpose of it? 🤔

packages/application-components/src/index.ts
```diff
@@ -16,6 +16,11 @@ export { default as FormModalPage } from './components/modal-pages/form-modal-pa
 export { default as TabularModalPage } from './components/modal-pages/tabular-modal-page';
 export { default as CustomFormModalPage } from './components/modal-pages/custom-form-modal-page';
 
+// Detail Pages
+export { default as CustomFormDetailPage } from './components/detail-pages/custom-form-detail-page';
+export { default as FormDetailPage } from './components/detail-pages/form-detail-page';
+export { default as InfoDetailPage } from './components/detail-pages/info-detail-page';
+
```
emmenko, line 23: The `tabular-detail-page` should also be moved into the `detail-pages` folder

emmenko (commented):
Thanks for working on this 🙌

Had a quick look with some questions. Will have another look later.

packages/application-components/src/components/detail-pages/custom-form-detail-page/custom-form-detail-page.tsx
kark: Thanks. Looking at what is expected at `/customers/new` in the MC:
Screenshot 2022-05-09 at 17 29 05
the top bar has to be optional.

The other though I had was to render it optionally based on `onPreviousPathClick`

.changeset/old-otters-arrive.md
```diff
@@ -0,0 +1,154 @@
...
+        name="email"
+        title="Email"
+        isRequired={true}
+        value={formik.values.email}
+        errors={formik.errors.email}
+        touched={formik.touched.email}
+        onChange={formik.handleChange}
+        onBlur={formik.handleBlur}
+      />
+    </FormDetailPage>
```
CarlosCortizasCT, line 88: For both `FormDetailPage` and `CustomFormDetailPage` I'm missing the `form` html tag wrapping the input fields.

Personally, I would expect those components to internally add that element but, if we don't want that solution, at least I find it necessary to add it into the examples (both in the changelog and in the docs).

If we include the form tag in our components, perhaps we would also want to link the primary button to it using the form attribute ([HTML attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes); [form attribute explained](https://www.impressivewebs.com/html5-form-attribute/)). Although this can be tricky with the current implementation.

CarlosCortizasCT (commented):
Awesome work 💪 

I just left a comment with something to maybe discuss about.

.changeset/old-otters-arrive.md
CarlosCortizasCT: I pushed a branch (**cm-details-pages**) with an implementation idea (not a working solution) of what I meant with my previous comment in case we want to discuss adding the form tag to the new form components.

Changes can be seen [here](https://github.com/commercetools/merchant-center-application-kit/commit/14bc21dfa08356ebd73a834af9e1e112b3280a24).

packages/application-components/src/components/detail-pages/custom-form-detail-page/custom-form-detail-page.tsx
emmenko: > The other though I had was to render it conditionally based on the `onPreviousPathClick`

Yeah maybe that's better so we don't need an extra prop for that.

.changeset/old-otters-arrive.md
emmenko: FYI: I think using the `form` element is a valid point and we should discuss this separately, as we're also not really using it in many parts of the MC at the moment.

@CarlosCortizasCT Can you maybe open a discussion or issue to talk about it and how we plan to use it? We should also check with the other teams if there are other concerns about having it set by default in the components.

Thanks

tdeekens: I'd also value this as I know many locations inside the MC where it's not used. Let's discuss it separately. I think we can either scan the next child and if it's a `form` not wrap it or have a `disableWrapByForm` prop which is `false` by default but can be disabled for instance. Similar to Next.js `Link` migration strategy which has a `oldBehaviour` prop.

.changeset/old-otters-arrive.md
```diff
@@ -0,0 +1,151 @@
...
+  return (
+    <CustomFormDetailPage
+      title="Manage your account"
+      onPreviousPathClick={() => history.push('/starting-page')}
+      formControls={
+        <>
+          <CustomFormDetailPage.FormSecondaryButton
+            onClick={() => {
+              formik.resetForm();
+            }}
```
tdeekens, line 129: Nit as we do that below
```suggestion
            onClick={formik.resetForm}
```

visual-testing-app/src/components/info-detail-page/info-detail-page.visualroute.tsx
```diff
@@ -0,0 +1,59 @@
+import Text from '@commercetools-uikit/text';
```
tdeekens, line 1: GitHub doesn't have syntax highlighting on this file but I guess that's unrelated to the file itself.

visual-testing-app/src/components/info-detail-page/info-detail-page.visualroute.tsx
```diff
@@ -0,0 +1,59 @@
...
+  </div>
+);
+DetailPageContainer.displayName = 'DetailPageContainer';
+
+export const Component = () => (
+  <Suite>
+    <Spec label="InfoDetailPage" size="xl">
+      <DetailPageContainer onPreviousPathClick={() => alert('Go back clicked')}>
+        <Text.Body>
+          {`Lorem ipsum dolor sit amet, consectetur adipiscing elit. Curabitur nec turpis in risus elementum fringilla. Vestibulum nec vulputate metus, fringilla luctus nisl. Vestibulum mattis ultricies augue sagittis vestibulum. Nulla facilisi. Quisque tempor pulvinar efficitur. Praesent interdum ultrices leo. Vivamus non ex maximus justo egestas suscipit eget sed purus. Aliquam ut venenatis nulla. Fusce ac ligula viverra, blandit augue eget, congue turpis. Curabitur a sagittis leo. Nunc sed quam dictum, placerat nunc quis, luctus erat.`}
```
tdeekens, line 27: I don't think we need tow rap these in `{``}`. There are some places in this PR we do so and others we don't.

tdeekens (commented):
Super cool. Can't wait to get rid of more homegrown internal things inside the MC with this 💟.

tdeekens (approved):
Thanks. This looks good from my view.

emmenko:
@kark Looks good, thanks!

One minor thing: the list of components got longer now and I think it would be nice to group things a bit (e.g. Dialogs, Modal pages, Detail pages, etc.)

image

Can also be a follow up.

emmenko (approved):
🚀

Let's do the docs improvements as a follow up, as discussed. Thanks!

website-components-playground/src/pages/custom-form-detail-page.js
```diff
@@ -136,9 +136,7 @@ const CustomFormDetailPageExample = (props) => {
                 formControls={
                   <>
                     <CustomFormDetailPage.FormSecondaryButton
-                      onClick={() => {
-                        formikProps.resetForm();
-                      }}
```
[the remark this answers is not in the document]
kark: Thanks 😊

PR #2597, opened 2022-05-09
feat(codemod): add transform `remove-deprecated-modal-level-props`

emmenko:
Relates to #2581 

To help with the "migration" related to the Stacking Layer System.

### DIFF
--- a/packages/codemod/src/transforms/rename-js-to-jsx.ts
+++ b/packages/codemod/src/transforms/rename-js-to-jsx.ts
@@ -20,7 +20,7 @@ function renameJsToJsx(file: FileInfo, api: API, options: TRunnerOptions) {
     }
   }
 
-  return root.toSource(options);
+  return null;
 }
 
 export default renameJsToJsx;
--- a/.changeset/kind-impalas-tickle.md
+++ b/.changeset/kind-impalas-tickle.md
@@ -3,5 +3,5 @@
 '@commercetools-website/custom-applications': patch
 ---
 
-Remove automatically updating local config application ID after a custom application is created.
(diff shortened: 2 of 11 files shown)


packages/codemod/src/transforms/rename-js-to-jsx.ts
```diff
@@ -20,7 +20,7 @@ function renameJsToJsx(file: FileInfo, api: API, options: TRunnerOptions) {
     }
   }
 
-  return root.toSource(options);
+  return null;
```
CarlosCortizasCT, line 23: I'm curious about the nature of this change.
Was this causing a problem?

CarlosCortizasCT (commented):
💯 

I just left a question for better understanding.

packages/codemod/src/transforms/rename-js-to-jsx.ts
emmenko: I saw doing this (conditionally return `null` or the actual transformed content) in some of the example codemods. Essentially it's meant to return something only if something actually changed.

This transformer is actually kind of a fake one, because we're just renaming the file and we're not transforming anything. So for consistency I changed that to return `null` as it's more correct technically.

CarlosCortizasCT: Ok, so I guess the tool only updates the source file if the transformer returns something.

Thanks for the explanation 👍 .

CarlosCortizasCT (approved):
💯

Rhotimee (approved):

kark (approved):
🙌

Issue #2577, opened 2022-04-29
Vite support: missing features


emmenko:
This is an umbrella issue to gather missing requirements/features when using the Vite bundler (#2568 ).

- [ ] Allow to configure the [`define` variables](https://vitejs.dev/config/#define).
- [ ] Allow to pass additional [plugins](https://vitejs.dev/config/#plugins).
- [ ] Support for missing babel plugins (opt-in)
  - [ ] `@babel/plugin-proposal-do-expressions`

PR #2599, opened 2022-05-09
refactor(config): load the config as a JS module instead of parsing it as JSON

emmenko:
Basic work for continuing with #2577


packages/application-config/scripts/load-js-module.js
```diff
@@ -1,34 +0,0 @@
...
-const [filePath] = process.argv.slice(2);
-
-// Require the module. It's expected that the module exports the application config
-const moduleExport = require(filePath);
-
-// In case we are loading an ES module, we need to pick the `default` export.
-const result = get(moduleExport, 'default', moduleExport);
-
-// Write the application config to `stdout`so that the main program can read it.
-process.stdout.write(JSON.stringify(result));
```
emmenko, line 34: Before this was executed as a Node script. The config was then printed to stdout and parsed from the child process.

packages/application-config/loaders/load-js-module.js
```diff
@@ -0,0 +1,29 @@
...
+  require('@babel/register')({
+    babelrc: false,
+    extensions: ['.js', '.cjs', '.mjs', '.ts'],
+    presets: [
+      require.resolve('@commercetools-frontend/babel-preset-mc-app/production'),
+    ],
+  });
+
+  // Require the module. It's expected that the module exports the application config
+  const moduleExport = require(filePath);
```
emmenko, line 24: Now we simply `require` the JS module, no JSON serialization.

This allows to define and preserve JS functions into the application config (useful for example for dev server plugins)

emmenko:
So this is not going anywhere, as I don't like any of the approaches I tried.

The main driver for this was to be able to define functions within the config and be able to use them. For example, for Vite to pass extra plugins and use them when starting the dev server.

I'm opting now for a different approach, to have plugins defined in a separate file and simply reference the file path in the custom app config.

Small downside is that there would be another file but I think it's a reasonable tradeoff.

emmenko:
Closing as there would be a different approach.

PR #2601, opened 2022-05-10
Improve regex for Entry Point Uri Path to accept at least 2 characters

ddouglasz:
Improve regex for Entry Path Url to accept at least 2 alphanumeric characters.
Currently, the regex to validate the entryPointUriPath seems to prevent at least 2 characters. Find more information [here](https://jira.commercetools.com/browse/SHIELD-474)

### DIFF
--- a/packages/application-config/schema.json
+++ b/packages/application-config/schema.json
@@ -30,7 +30,7 @@
       "oneOf": [
         {
           "type": "string",
-          "pattern": "^[^\\-_]([0-9a-z]|[\\-_](?![\\-_])){2,64}[^\\-_]$"
+          "pattern": "^[^\\-_]([0-9a-z]|[\\-_](?![\\-_])){0,62}[^\\-_]$"
         },
         {
           "$ref": "#/definitions/envVariablePlaceholder"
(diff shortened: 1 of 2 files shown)


packages/application-config/schema.json
```diff
@@ -30,7 +30,7 @@
       "oneOf": [
         {
           "type": "string",
-          "pattern": "^[^\\-_]([0-9a-z]|[\\-_](?![\\-_])){2,64}[^\\-_]$"
+          "pattern": "^[^\\-_#]([0-9a-z]|[\\-_](?![\\-_])){0,62}[^\\-_#]$"
```
ddouglasz, line 33: I assume that since # is kind of like a special url character, maybe we want to check in the character class `[^-_#]` beginning and end to be sure the entryPathUriPath does not start or end with it too.
Any thoughts on this?

emmenko: Good idea, thanks for noticing.

Let's add it to the regex as you suggested, thanks!

CarlosCortizasCT (approved):
💯

kark (approved):

tdeekens (approved):

Rhotimee (approved):

PR #2604, opened 2022-05-10
refactor(l10n): add translations for included timezones

ByronDWall:
#### Summary

This PR adds a `time-zone-messages.js` file to `l10n/src`.

#### Description

This messages file is meant to be the single source of truth for translations of long-form timezone names.

The objects are keyed by IANA timezone names, and include an `offset` value that is either a number, or `false`.  The reasoning behind this is to include the GMT offset number for timezones that have no daylight/offset times, and `moment-timezone` to return the correct GMT offset for timezones that have daylight/offset time.

### DIFF
--- a/packages/l10n/src/time-zone-messages.ts
+++ b/packages/l10n/src/time-zone-messages.ts
@@ -1,0 +1,1419 @@
+// import { defineMessages } from 'react-intl';
+
+export const TIME_ZONE_MESSAGES = {
+  'Pacific/Niue': {
+    id: 'Pacific/Niue',
+    defaultMessage: 'Niue Time',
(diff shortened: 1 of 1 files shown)


emmenko:
I'm still confused about this.

I would expect the source translations to be defined in the `data/time-zones/core.json` file.

tdeekens:
Not 100% sure I manage to follow yet. My thinking was that we either add a `core.json` file to `data/time-zones` or have a `message.js` with a `defineMessages` all from `react-intl` to have a basis for the translations yielding them being part of a `core.json`. Maybe you can schedule a quick chat with @emmenko and me some time later this week.

PR #2608, opened 2022-05-11
ci: build and deploy docs from gh actions

emmenko:
Similarly to how we deploy the docs in the other repos, we now do the same for appkit.

Main reason is that Vercel has a size limit of 1GB of the cache, which we've been running very close to that limit.

image


CarlosCortizasCT (approved):
💯

Rhotimee (approved):

.github/workflows/docs.yml
```diff
@@ -0,0 +1,137 @@
...
+
+      - name: Building docs website
+        run: yarn build:website
+
+      - name: Detect if is a production deployment (documentation)
+        if: github.event_name == 'push' && github.ref_name == 'main'
+        run: |
+          echo "DEPLOY_DOCS_ENV=Production" >> $GITHUB_ENV
+
+      - name: Deploy to Vercel (Production)
```
CarlosCortizasCT, line 63: I'm just curious about how does the Vercel step knows which folder we want to deploy since we don't specify it.
Does it automatically search for a `vercel-public` directory (we create it in the `yarn build:website` command) to deploy from?

emmenko: Yes it's defined in Vercel

image

ddouglasz (approved):

PR #2609, opened 2022-05-11
ci: fix cache config for docs

emmenko:
Leftover of #2608

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -34,16 +34,17 @@ jobs:
         with:
           node-version-file: '.nvmrc'
 
-      - name: Get yarn cache
-        id: yarn-cache
-        run: echo "::set-output name=dir::$(yarn config get cacheFolder)"
-
       - uses: actions/cache@v3
         with:
-          path: ${{ steps.yarn-cache.outputs.dir }}
-          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          path: |
+            .cache
+            public
+          key: ${{ runner.os }}-yarn-docs-${{ hashFiles('**/yarn.lock') }}
           restore-keys: |
-            ${{ runner.os }}-yarn-
+            ${{ runner.os }}-yarn-docs-
+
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -34,16 +34,17 @@ jobs:
...
-        id: yarn-cache
-        run: echo "::set-output name=dir::$(yarn config get cacheFolder)"
-
       - uses: actions/cache@v3
         with:
-          path: ${{ steps.yarn-cache.outputs.dir }}
-          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          path: |
+            .cache
+            public
```
emmenko, line 41: I forgot that for the Gatsby cache setup, we need to do this.

CarlosCortizasCT: Sorry that I'm not very familiar with this but it seems we're caching both `.cache` and `public` directories and assign a key tied to a specific version of the `yarn.lock` file.

Perhaps those directories contain previous builds with Gatsby so we would like them to not be reused when we change some content, but changing content will not make the `yarn.lock` file to be updated so the previous build would be fetched from the cache.

I guess I'm missing something so this is just a question for me to better understand the flow.

Thanks.

tdeekens: The other day I learned that `setup-node` actually has good default caching for most package managers. It's just hidden in their docs under "Adavanced Features". You can find it [here](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-data). Maybe we can use some of it. Not entirely sure why we would cache `public`.

emmenko: It's required for enabling Gatsby incremental builds, otherwise Gatsby will rebuild the entire app again.

See also our custom script for managing multiple docs apps: https://github.com/commercetools/merchant-center-application-kit/blob/main/scripts/gatsby-cache.mjs

emmenko: Btw, this is the same setup we use for all other docs-related repos.

PR #2612, opened 2022-05-13
refactor(l10n): add time zone translations and update time zone data generation

ByronDWall:
#### Summary

This PR adds translations for long form time zone names to `l10n/data/time-zones/core.json` and `l10n/data/time-zones/en.json`.  And updates the `generate-l10n-data` and `getTimeZonesForLocale` method.

#### Description

The `generate-l10n-data` script will get the most recent IANA timezone list by running `moment.tz.names()`, remove any time zones that are in the `EXCLUDED_TIME_ZONES` constant from that array, then make sure that every remaining timezone has a corresponding translation string in `core.json`.  If a timezone is not in `EXCLUDED_TIME_ZONES` or `core.json`, the script will prompt the user to either add any "new" time zones that should be translated to `core.json`, or to exclude them from being added to the time zone list by adding them to the `EXCLUDED_TIME_ZONES` constant in `src/constants.ts`.

The `getTimeZonesForLocale` function in `src/time-zone-information.ts` builds a normalized object using the locale-specific translation from `<locale>.json` for the `label` value, and the `offset` and `abbr` from `moment-timezone`.  This allows for us to generate the `offset` value at runtime, which will insure that the `offset` for all time zones that have daylight/offset time is accurate.  Previously, the `offset` value was only set when the `generate-l10n-data` script was run, leading to some timezones showing inaccurate offset values when they moved from daylight to standard time or vice-versa.

A SCREEN RECORDING OF THE FINAL CLI WORKFLOW IS HERE:

https://user-images.githubusercontent.com/87667330/173929422-06ccd6aa-1cf9-4c5a-9e0e-b211a3e9c267.mov


packages/l10n/data/time-zones/core.json
```diff
@@ -0,0 +1,285 @@
...
+  "America/Grand_Turk": "Eastern Time - Grand Turk",
+  "America/Indiana/Indianapolis": "Eastern Time - Indianapolis",
+  "America/Iqaluit": "Eastern Time - Iqaluit",
+  "America/Kentucky/Louisville": "Eastern Time - Louisville",
+  "America/New_York": "Eastern Time - New York",
+  "America/Port-au-Prince": "Eastern Time - Port-au-Prince",
+  "America/Thunder_Bay": "Eastern Time - Thunder Bay",
+  "America/Toronto": "Eastern Time - Toronto",
+  "America/Guyana": "Guyana Time ",
+  "America/Asuncion": "Paraguay Time - Asuncion",
```
stephsprinkle, line 72:
```suggestion
  "America/Asuncion": "Paraguay Time - Asunción",
```

Not sure about this one, there are others too that have been standardized out in the clean data column.

packages/l10n/data/time-zones/core.json
```diff
@@ -0,0 +1,285 @@
...
+  "Europe/Luxembourg": "Central European Time - Luxembourg",
+  "Europe/Madrid": "Central European Time - Madrid",
+  "Europe/Malta": "Central European Time - Malta",
+  "Europe/Monaco": "Central European Time - Monaco",
+  "Europe/Oslo": "Central European Time - Oslo",
+  "Europe/Paris": "Central European Time - Paris",
+  "Europe/Prague": "Central European Time - Prague",
+  "Europe/Rome": "Central European Time - Rome",
+  "Europe/Stockholm": "Central European Time - Stockholm",
+  "Europe/Tirana": "Central European Time - Tirane",
```
stephsprinkle, line 136: The city name on this one is `Tirana` in the doc

packages/l10n/data/time-zones/core.json
```diff
@@ -0,0 +1,285 @@
...
+  "Asia/Hebron": "Eastern European Time - Hebron",
+  "Europe/Helsinki": "Eastern European Time - Helsinki",
+  "Europe/Kiev": "Eastern European Time - Kyiv",
+  "Asia/Nicosia": "Eastern European Time - Nicosia",
+  "Europe/Riga": "Eastern European Time - Riga",
+  "Europe/Sofia": "Eastern European Time - Sofia",
+  "Europe/Tallinn": "Eastern European Time - Tallinn",
+  "Europe/Uzhgorod": "Eastern European Time - Uzhgorod",
+  "Europe/Vilnius": "Eastern European Time - Vilnius",
+  "Europe/Zaporozhye": "Eastern European Time - Zaporozhskoye",
```
stephsprinkle, line 165: The city name on this one is `Zaporozhye` in the doc

packages/l10n/data/time-zones/core.json
```diff
@@ -0,0 +1,285 @@
...
+  "Israel": "Israel Time - Tel Aviv",
+  "Europe/Kirov": "Moscow Standard Time - Kirov",
+  "Europe/Minsk": "Moscow Standard Time - Minsk",
+  "Europe/Moscow": "Moscow Standard Time - Moscow",
+  "Europe/Simferopol": "Moscow Standard Time - Simferopol",
+  "Europe/Volgograd": "Moscow Standard Time - Volgograd",
+  "Turkey": "Turkey Time - Istanbul",
+  "Asia/Yerevan": "Armenia Standard Time - Yerevan",
+  "Europe/Astrakhan": "Europe/Astrakhan",
```
stephsprinkle, line 174:
```suggestion
  "Europe/Astrakhan": "Astrakhan Time - Astrakhan",
```

packages/l10n/data/time-zones/core.json
```diff
@@ -0,0 +1,285 @@
...
+  "Asia/Baku": "Azerbaijan Time - Baku",
+  "Asia/Tbilisi": "Georgia Standard Time - Tbilisi",
+  "Indian/Mauritius": "Mauritius Time",
+  "Indian/Reunion": "Réunion Time",
```
stephsprinkle, line 178: We should definitely align on accent marks

packages/l10n/data/time-zones/core.json
```diff
@@ -0,0 +1,285 @@
...
+  "Asia/Ho_Chi_Minh": "Indochina Time - Ho Chi Minh City",
+  "Asia/Krasnoyarsk": "Krasnoyarsk Time",
+  "Asia/Novosibirsk": "Novosibirsk Time",
+  "Asia/Tomsk": "Tomsk Time",
+  "Asia/Jakarta": "Western Indonesian Time - Jakarta",
+  "Asia/Pontianak": "Western Indonesian Time - Pontianak",
+  "Australia/West": "Australian Western Standard Time - Perth",
+  "Asia/Brunei": "Brunei Darussalam Time",
+  "Asia/Makassar": "Central Indonesian Time - Makassar",
+  "Asia/Macau": "China Standard Time- Macau",
```
stephsprinkle, line 226:
```suggestion
  "Asia/Macau": "China Standard Time - Macau",
```

stephsprinkle (commented):
I erred on the side of caution here, feel free to disregard if too nitpicky

tdeekens (commented):
Nice the shape of the file looks good. Wondering how: was this generated by the script or hand? 🤔

packages/l10n/data/time-zones/core.json
```diff
@@ -0,0 +1,285 @@
...
+  "America/Merida": "Central Time - Merida",
+  "America/Mexico_City": "Central Time - Mexico City",
+  "America/Monterrey": "Central Time - Monterrey",
+  "America/Rankin_Inlet": "Central Time - Rankin Inlet",
+  "America/Indiana/Tell_City": "Central Time - Tell City, Indiana",
+  "America/Winnipeg": "Central Time - Winnipeg",
+  "America/Cancun": "Eastern Standard Time - Cancun",
+  "America/Jamaica": "Eastern Standard Time - Jamaica",
+  "America/Panama": "Eastern Standard Time - Panama",
+  "America/Guayaquil": "Eastern Standard Time - Guayaquil",
```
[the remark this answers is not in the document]
ByronDWall: This is something I will have to ask Diane - we've found a few other duplicated or missed time zones, so I will add this to the list

packages/l10n/data/time-zones/core.json
ByronDWall: I have asked Diane about accent marks as well, and will add/remove them as she requires once she gets back to me

ByronDWall:
> Nice the shape of the file looks good. Wondering how: was this generated by the script or hand? 🤔

the original constant was, `core.json` was created by running a reduce over that constant.

packages/l10n/scripts/generate-l10n-data.js
```diff
@@ -294,6 +295,79 @@ const mapDiffToWarnings = (oldData, newData) => {
     .filter(Boolean);
 };
 
+const rl = readline.createInterface({
+  input: process.stdin,
+  output: process.stdout,
+});
```
emmenko, line 301: Is this supposed to be used somewhere? 🤔

packages/l10n/scripts/generate-l10n-data.js
```diff
@@ -294,6 +295,79 @@ const mapDiffToWarnings = (oldData, newData) => {
...
+  const targetExcludedTimeZonesFile = path.join(
+    targetTranslationsFolder,
+    'excluded_time_zones.json'
+  );
+
+  moment.tz.setDefault('Etc/GMT');
+
+  // const allTimeZoneIds = moment.tz.names();
+  const fakeTimeZones = ['space/moon', 'space/mars', 'space/jupiter'];
+  const allTimeZoneIds = moment.tz.names().concat(fakeTimeZones);
```
emmenko, line 336: Can you clarify this?

packages/l10n/scripts/generate-l10n-data.js
```diff
@@ -294,6 +295,79 @@ const mapDiffToWarnings = (oldData, newData) => {
...
+
+  console.log(
+    `[${key}]: Checking moment-timezone for new IANA timezone identifiers`
+  );
+  try {
+    fs.accessSync(targetTranslationsFile, fs.F_OK);
+    fs.accessSync(targetExcludedTimeZonesFile, fs.F_OK);
```
emmenko, line 343:
```suggestion
    fs.accessSync(targetTranslationsFile, fs.constants.F_OK);
    fs.accessSync(targetExcludedTimeZonesFile, fs.constants.F_OK);
```

emmenko (commented):
Thanks for working on this!
I left some early feedback, also to clarify a couple of things.

packages/l10n/scripts/generate-l10n-data.js
```diff
@@ -294,6 +295,79 @@ const mapDiffToWarnings = (oldData, newData) => {
...
+   * have a translation, that timezone is considered a new timezone.
+   *
+   * If a new timezone is encountered, the user running the CLI command
+   * will be prompted via the shell whether they would like to accept
+   * the new timezone, or add that timezone to this excluded list.
+   */
+  const targetExcludedTimeZonesFile = path.join(
+    targetTranslationsFolder,
+    'excluded_time_zones.json'
+  );
```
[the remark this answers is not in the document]
ByronDWall: Putting it in `scripts` works for me.  

The only reason I have it as a json file is because there is a potential that we might be writing values to the file programmatically via the cli, but I think you're right that `.js` makes more sense

packages/l10n/scripts/generate-l10n-data.js
ByronDWall: This won't be in the final, it's just there so I can have some 'new' timezones to use while developing the cli tool.

packages/l10n/scripts/generate-l10n-data.js
ByronDWall: should I update this in other methods (ie `updateLocaleData`)?

packages/l10n/scripts/parse-new-time-zones.js
```diff
@@ -0,0 +1,58 @@
+const readline = require('readline');
```
emmenko, line 1: I would suggest to use the [prompts](https://npmjs.com/package/prompts) library, it's easier to handle user prompts.

packages/l10n/src/time-zone-information.ts
```diff
@@ -1,40 +1,67 @@
...
 type ImportData = {
   default: TimeZones;
 };
 
+type TranslationData = {
+  default: TimeZoneTranslations;
+};
+/**
+ * TODO:
+ * - handle fallback to 'default'/core.json when there are no translations for a locale
```
emmenko, line 17: There are different ways to handle this.

For example, we could remove all "empty" locale files and omit the locale in the switch-case, so that the fallback/default `en` locale is used.

This also means that at least the `en.json` should be defined

packages/l10n/src/time-zone-information.ts
```diff
@@ -1,40 +1,67 @@
...
+};
+
 const getImportChunk = (locale: string): Promise<ImportData> => {
   const intlLocale = mapLocaleToIntlLocale(locale);
   switch (intlLocale) {
     case 'de':
       return import(
         /* webpackChunkName: "timezone-data-de" */ '../data/time-zones/de.json'
-      );
+      ).then((module) => getTimeZoneData(module));
```
emmenko, line 44: I wouldn't do the mapping here but in the `getTimeZonesForLocale` function.

emmenko (commented):
Some quick feedback to simplify things a bit more.

packages/l10n/scripts/generate-l10n-data.js
```diff
@@ -294,6 +296,114 @@ const mapDiffToWarnings = (oldData, newData) => {
...
 };
 
+const updateTimeZoneData = async (key) => {
+  const allTimeZoneIds = moment.tz.names();
+  // uncomment for testing/review purposes, will be removed once PR is out of draft
+  // const fakeTimeZones = ['zzspace/moon', 'zzspace/mars', 'zzspace/jupiter'];
+  // const allTimeZoneIds = moment.tz.names().concat(fakeTimeZones);
+  const dataFolderPath = path.join(__dirname, '..', DATA_DIR[key].path);
+  const sourceDataFilePath = path.join(dataFolderPath, 'core.json');
+  await fs.accessSync(sourceDataFilePath, fs.constants.F_OK);
```
tdeekens, line 306: awaiting something sync is odd to me. Maybe we should extract it into `function exitIfSourceFolderDoesNotExist` as that's what we want I think.

packages/l10n/scripts/generate-l10n-data.js
```diff
@@ -294,6 +296,114 @@ const mapDiffToWarnings = (oldData, newData) => {
...
+  const translatedTimeZones = JSON.parse(
```
tdeekens, line 307: Coming in as a reviewer I struggle with naming and flow a bit. Just as a note. Will read the code multiple times to suggest something to improve it.

packages/l10n/scripts/generate-l10n-data.js
```diff
@@ -294,6 +296,114 @@ const mapDiffToWarnings = (oldData, newData) => {
...
+      const updatedTranslationsFile = Object.keys(updatedTranslationData)
+        .sort((a, b) => a.toLowerCase().localeCompare(b.toLowerCase()))
+        .reduce((acc, key) => {
+          return {
+            ...acc,
+            [key]: updatedTranslationData[key],
+          };
+        }, {});
+      // write new translations file
+      shell.mkdir('-p', dataFolderPath);
```
tdeekens, line 357: I don't think this is needed to use this package. We can just `fs.mkdirSync(dir);`

packages/l10n/scripts/generate-l10n-data.js
```diff
@@ -294,6 +296,114 @@ const mapDiffToWarnings = (oldData, newData) => {
...
+  const sourceDataFilePath = path.join(dataFolderPath, 'core.json');
+  await fs.accessSync(sourceDataFilePath, fs.constants.F_OK);
+  const translatedTimeZones = JSON.parse(
+    fs.readFileSync(sourceDataFilePath, { encoding: 'utf8' })
+  );
+  const translatedTimeZoneIds = Object.keys(translatedTimeZones);
+  console.log(
+    `[${key}]: Checking moment-timezone for new IANA timezone identifiers`
+  );
+  const newTimeZoneIds = allTimeZoneIds.filter(
```
tdeekens, line 314:
```suggestion
  const nonExistingTimeZoneIds = allTimeZoneIds.filter(
```

packages/l10n/scripts/generate-l10n-data.js
```diff
@@ -294,6 +296,114 @@ const mapDiffToWarnings = (oldData, newData) => {
...
+      // get path for file to write
+      const excludedTimeZonesFilePath = path.join(
+        __dirname,
+        'excluded-time-zones.js'
+      );
+      // write new exclusions file
+      shell.mkdir('-p', './');
+      fs.writeFileSync(
+        excludedTimeZonesFilePath,
+        `const EXCLUDED_TIME_ZONES = ${JSON.stringify(
```
tdeekens, line 386: Can't we write a JSON file here we just import and parse?

packages/l10n/scripts/generate-l10n-data.js
```diff
@@ -294,6 +296,114 @@ const mapDiffToWarnings = (oldData, newData) => {
...
+        (translationData, newTranslationId) => {
+          return {
+            ...translationData,
+            [newTranslationId]: 'ADD NEW TRANSLATION',
+          };
+        },
+        translatedTimeZones
+      );
+      // sort translations alphabetically
+      const updatedTranslationsFile = Object.keys(updatedTranslationData)
```
tdeekens, line 348: Why do we reduce twice?

packages/l10n/src/time-zone-information.ts
ByronDWall: I was thinking that we could lazy-load the locale first, and if the length `locale.default` is zero then lazy-load `en`, that way we wouldn't _necessarily_ always have to define `en.json`

packages/l10n/scripts/generate-l10n-data.js
ByronDWall: I don't think this actually needs an `await`, I'll remove it.

packages/l10n/scripts/generate-l10n-data.js
ByronDWall: fwiw, I freely admit that naming and flow could be improved, I'll think about it as well.

packages/l10n/scripts/generate-l10n-data.js
ByronDWall: `fs.mkdirSync(dir)` errors out if the directory already exists, `shell.mkdir('-p', dir)` does not

packages/l10n/scripts/generate-l10n-data.js
tdeekens: Ok. Forgot about that comment. Sorry.

packages/l10n/scripts/generate-l10n-data.js
ByronDWall: Sure, I can check for existence using `fs`.

I'm following this pattern because it's what's used in `updateLocaleData`, would you like me to remove `shell` there as well so that we can remove the dependency altogether?

packages/l10n/scripts/parse-unhandled-time-zones.js
```diff
@@ -0,0 +1,65 @@
...
+module.exports = async function parseUnhandledTimeZones(unhandledTimeZones) {
+  const questions = unhandledTimeZones.map((timeZone) => ({
+    type: 'select',
+    name: timeZone,
+    message: `Choose option for ${timeZone}: `,
+    choices: [
+      {
+        title: 'Translate',
+        description: `Add ${timeZone} to core.json for translation`,
+        value: ANSWER_TYPES.ADD,
```
stephsprinkle, line 48: Could be interesting to add the option for the user to enter the text to be added in `core.json` for the new/unhandled time zone.

packages/l10n/scripts/generate-l10n-data.js
tdeekens: That's for another time I think. A decent goal though to avoid such bulky things for so little benefit.

packages/l10n/src/time-zone-information.ts
```diff
@@ -50,12 +88,21 @@ export const timeZonesShape = PropTypes.objectOf(
...
-  const timeZones = await getImportChunk(supportedLocale);
+  let localeTranslations = await getImportChunk(supportedLocale);
+  // if there are no translations for the desired locale, fall back to en/default
+  if (
+    !localeTranslations.default ||
+    Object.keys(localeTranslations.default).length === 0
+  ) {
+    localeTranslations = await getImportChunk('en');
+  }
+  // create time zone object with abbreviations and offsets
```
tdeekens, line 104: Don't we have to handle this on a per timezone basis also and not on a chunk basis? What if we also add the new timezone name (EN version) to any `<locale.>json` file if the timezone is new. That would have it default to EN until we pull in something else from transifex.

packages/l10n/scripts/generate-l10n-data.js
```diff
@@ -294,6 +269,130 @@ const mapDiffToWarnings = (oldData, newData) => {
...
+    if (timeZonesToTranslate.length) {
+      const updatedTranslations = translatedTimeZoneIds
+        .concat(timeZonesToTranslate)
+        .sort(timeZoneCompareFn)
+        .reduce((previousTimeZones, timeZoneId) => {
+          return {
+            ...previousTimeZones,
+            [timeZoneId]: translatedTimeZones[timeZoneId]
+              ? translatedTimeZones[timeZoneId]
+              : 'ADD TRANSLATION',
```
tdeekens, line 363: 👋🏼 When can this actually happen?

tdeekens:
Just to aggregate open discussion for me

1. https://github.com/commercetools/merchant-center-application-kit/pull/2612#discussion_r874073635
2. https://github.com/commercetools/merchant-center-application-kit/pull/2612/files/a2e614d429085909dbd3545131e31ecef5c5725c#diff-ad1a3ae97e85f30b06d0f27cf78a4aac888375337bcdfc1088ff0130064f858e
3. https://github.com/commercetools/merchant-center-application-kit/pull/2612/files/67b17b86194a43748da97eadb8c460efd1633560#diff-3250963761c21e63195cdc5c49c8a950e1526a7a2bf100e5ca68a0ec44899abf

packages/l10n/src/time-zone-information.ts
ByronDWall: This works for me if it works for you

tdeekens: Otherwise I don't see how we can show anything if we have say 22 translations in en.json and 20 in de.json (as two are new). I don't see how de.json being loaded then (as it's not empty) would then mean that it can never show anything for the two new TZs if I am not mistaken.

ByronDWall: I didn't know whether it'd be better to wait for a translation before making a 'new' timezone available in the final tz list, or to add the tz to the locales immediately.

We can talk more about the workflow when we meet about transifex, but I think there are pros and cons to each approach.  If we decide to add the translation using the CLI tool, I will refactor it so that it writes to all locale files as well as `core.json`.

tdeekens: > I didn't know whether it'd be better to wait for a translation before making a 'new' timezone available in a locale's tz list, or to add the tz to the locale files immediately.

I think that's a 🐔 🥚 problem if I understand your concern right. Let's talk tomorrow.

stephsprinkle (commented):
One thing I noticed and I'm not sure how much we care about this: with two new time zones, if I opt to translate one, I cannot use the newly translated time zone as a fallback for the second (I had to select an IANA time zone name that was already translated).

packages/l10n/scripts/generate-l10n-data.js
```diff
@@ -294,6 +274,206 @@ const mapDiffToWarnings = (oldData, newData) => {
...
+}
+
+async function updateTimeZoneData(key, supportedLocales) {
+  const allTimeZoneIds = moment.tz.names();
+  // uncomment for testing/review purposes, will be removed once PR is out of draft
+  // const fakeTimeZones = ['zzspace/moon', 'zzspace/mars', 'zzspace/jupiter'];
+  // const allTimeZoneIds = moment.tz.names().concat(fakeTimeZones);
+  const dataFolderPath = path.join(__dirname, '..', DATA_DIR[key].path);
+  const sourceDataFilePath = path.join(dataFolderPath, 'core.json');
+  /**TO CONSIDER:
```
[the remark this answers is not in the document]
ByronDWall: yes, once I'm told what the mapping will be

stephsprinkle (approved):
Looks good to me and does what it's supposed to do! Incredible effort 👏

.changeset/dirty-ravens-judge.md
```diff
@@ -0,0 +1,43 @@
...
+    },
+    'America/New_York': {
+        id: 'America/New_York',
+        abbr: 'EST',
+        label: 'Eastern Time - New York',
+        offset: '-05:00',
+        translationFor: ['EST', 'EDT']
+    }
+}
+```
```
emmenko, line 43: Thanks for the detailed explanation.

For a changeset though, this information is not really relevant and is more for us maintainers to know.

Can you move this to the README of the package maybe? There is a section about "Generating localization data".

In the changeset I would simply say the following:

```suggestion
The localized data has been updated. In particular, the time zones data has been updated and restructured to be more usable and readable.
```

packages/l10n/package.json
```diff
@@ -29,18 +29,18 @@
     "@types/react": "^17.0.44",
     "lodash": "4.17.21",
     "moment": "^2.29.2",
+    "moment-timezone": "^0.5.34",
+    "prompts": "^2.4.2",
```
emmenko, line 33: Nit: this should be a dev dependency

emmenko (commented):
Thanks for the huge work!

I did a first review and left a couple of suggestions (I also pushed a commit https://github.com/commercetools/merchant-center-application-kit/pull/2612/commits/39d454fbf9238a7fd63b516ce451d9b045d8842e).

Will do a final review next week.

packages/l10n/package.json
ByronDWall: When I move `moment-timezone` to `devDependencies`, I get this error from `yarn build`:
```
🎁 error @commercetools-frontend/l10n "moment-timezone" is imported by "src/time-zone-information.ts" but the package is not specified in dependencies or peerDependencies
```

But I've added `prompts` to `devDependencies`

.changeset/dirty-ravens-judge.md
ByronDWall: changeset and readme have been updated

tdeekens (approved):
Tried to get back into the groove of this as much as possible. Thanks for all the effort and time spent. To me this looks really nice now, allowing us to achieve anything we envisioned.

emmenko (approved):
Thanks for the great work and contribution!

From my side this is good to go as well 🙌

PR #2613, opened 2022-05-13
refactor: migrate `mc-scripts` package to TypeScript

emmenko:
Following up some previous work #2578 and #2579

See inline comments for more information.

TODO: 
- [x] Refactor commands to be loaded via dynamic `import`, instead of spawning a child process. This allows us to bundle the package normally without relying on relative files from the file system.
- [x] Bundle with Preconstruct
- [x] Migrate all configs (Webpack, etc.)
- [x] Migrate all commands
  - [x] `start`, `build`, `compile-html`, `serve`
  - [ ] `login`, 'config:sync'
- [x] Replace `mri` with `cac` (used by Vite CLI). This allows to define the CLI commands and options more declarative and to better handle long running commands (e.g. starting server).
- [x] Generate GraphQL types for CLI requests
- [ ] Migrate remaining utils for `login` and `config:sync` commands


packages/mc-scripts/package.json
```diff
@@ -14,8 +14,20 @@
   "publishConfig": {
     "access": "public"
   },
-  "main": "./build/index.js",
-  "files": ["build", "config", "package.json", "LICENSE", "README.md"],
+  "main": "dist/commercetools-frontend-mc-scripts.cjs.js",
+  "module": "dist/commercetools-frontend-mc-scripts.esm.js",
+  "preconstruct": {
+    "entrypoints": ["./index.ts", "./application-runtime.ts"]
+  },
```
emmenko, line 21: The `application-runtime` is a separate entry point so that the file can be referenced from the web pack config.

packages/mc-scripts/package.json
```diff
@@ -42,6 +54,7 @@
     "autoprefixer": "^10.4.4",
     "babel-loader": "8.2.5",
     "browserslist": "^4.20.2",
+    "cac": "6.7.12",
```
emmenko, line 57: New CLI library as a replacement for `mri`, as it has some nicer features.

https://www.npmjs.com/package/cac

packages/mc-scripts/src/bin/cli.ts
```diff
@@ -0,0 +1,244 @@
...
+        process.env.ENABLE_EXPERIMENTAL_VITE_BUNDLER === 'true';
+      if (shouldUseExperimentalBundler) {
+        console.log('Experimental Vite bundler enabled! 🚀');
+        console.log('');
+      }
+
+      const startCommand = shouldUseExperimentalBundler
+        ? await import('../commands/start-vite')
+        : await import('../commands/start');
+      await startCommand.default();
```
emmenko, line 63: This is now how we load and execute the commands. The bundle then contains code-splitter chunks.

packages/mc-scripts/src/bin/cli.ts
```diff
@@ -0,0 +1,244 @@
...
+
+  cli.help();
+  cli.version(pkgJson.version);
+  cli.parse();
+};
+
+// Load dotenv files into the process environment.
+// This is essentially what `dotenv-cli` does, but it's now built into this CLI.
+// Inspired also by https://create-react-app.dev/docs/adding-custom-environment-variables/#what-other-env-files-can-be-used
+function loadDotEnvFiles(globalOptions: TCliGlobalOptions) {
```
emmenko, line 205: Nothing changed here

packages/mc-scripts/src/commands/build.ts
```diff
@@ -0,0 +1,152 @@
...
+
+const measureFileSizesBeforeBuild =
+  FileSizeReporter.measureFileSizesBeforeBuild;
+const printFileSizesAfterBuild = FileSizeReporter.printFileSizesAfterBuild;
+
+// These sizes are pretty large. We'll warn for bundles exceeding them.
+const WARN_AFTER_BUNDLE_GZIP_SIZE = 512 * 1024;
+const WARN_AFTER_CHUNK_GZIP_SIZE = 1024 * 1024;
+
+async function run() {
```
emmenko, line 21: Nothing much changed here, besides wrapping the logic into the `run` function and using more async/await

packages/mc-scripts/src/commands/compile-html.ts
```diff
@@ -0,0 +1,38 @@
+import fs from 'fs';
+import chalk from 'chalk';
+import { compileHtml } from '@commercetools-frontend/mc-html-template';
+import paths from '../config/paths';
+import { TCliCommandCompileHtmlOptions } from '../types';
+
+const appDirectory = fs.realpathSync(process.cwd());
+
+async function run(options: TCliCommandCompileHtmlOptions = {}) {
```
emmenko, line 9: Nothing changed here

packages/mc-scripts/src/commands/serve.ts
```diff
@@ -0,0 +1,43 @@
+import http from 'http';
+import serveHandler from 'serve-handler';
+import paths from '../config/paths';
+
+const port = 3001;
+
+async function run() {
```
emmenko, line 7: Nothing changed here

packages/mc-scripts/src/commands/start.ts
```diff
@@ -0,0 +1,129 @@
...
+  createCompiler,
+  prepareUrls,
+  type CreateCompilerOptions,
+} from 'react-dev-utils/WebpackDevServerUtils';
+import openBrowser from 'react-dev-utils/openBrowser';
+import paths from '../config/paths';
+import createDevServerConfig from '../config/webpack-dev-server.config';
+import createWebpackConfigForDevelopment from '../config/create-webpack-config-for-development';
+
+async function run() {
```
emmenko, line 18: Nothing much changed here, besides wrapping into `run` function, using async/await and cleaning up a bit.

packages/mc-scripts/src/commands/start.ts
```diff
@@ -0,0 +1,129 @@
...
+  } as CreateCompilerOptions);
+
+  // Serve webpack assets generated by the compiler over a web sever.
+  const serverConfig = createDevServerConfig({
+    port,
+    publicPath: config.output?.publicPath as string,
+  });
+  const devServer = new WebpackDevServer(serverConfig, compiler);
+
+  await devServer.start();
```
emmenko, line 72: We're using a promise now

packages/mc-scripts/src/utils/graphql-requests.ts
```diff
@@ -0,0 +1,165 @@
...
+  TCustomApplicationDraftDataInput,
+  TFetchCustomApplicationFromCliQuery,
+  TFetchCustomApplicationFromCliQueryVariables,
+  TUpdateCustomApplicationFromCliMutation,
+  TUpdateCustomApplicationFromCliMutationVariables,
+} from '../generated/settings';
+import type {
+  TFetchMyOrganizationsFromCliQuery,
+  TFetchMyOrganizationsFromCliQueryVariables,
+} from '../generated/core';
```
emmenko, line 18: Using the generated types.

packages/mc-scripts/src/commands/config-sync.ts
```diff
@@ -110,6 +115,8 @@ const configSync = async () => {
       data,
     });
 
+    if (!createdCustomApplication) return;
```
CarlosCortizasCT, line 118: Not sure about this line. I don't know when the `createCustomApplication` can return a null value and have not raised an error.
Even in that scenario, we would just exit the process and not tell the use what happened.

I guess we would need to know when this use case can happen and, if it really exists, add an explanatory message before exiting the process.

packages/mc-scripts/src/utils/user-agent.ts
```diff
@@ -1,12 +1,12 @@
-const createHttpUserAgent = require('@commercetools/http-user-agent');
-const pkgJson = require('../../package.json');
+import createHttpUserAgent from '@commercetools/http-user-agent';
+import pkgJson from '../../package.json';
 
 const userAgent = createHttpUserAgent({
-  name: 'cli-login',
+  name: 'graphql-request',
```
CarlosCortizasCT, line 5: I think the previous name was more meaningful from the debugging perspective.

CarlosCortizasCT (commented):
What a huge work in this PR 😮 💪 

I just left a single comment to double check something I didn't fully understand.

packages/mc-scripts/src/commands/config-sync.ts
emmenko: The graphql client throws an error, so we don't need to explicitly handle this case.
Not sure if this check is really necessary. Maybe TS complains about something? Could you double check?

packages/mc-scripts/src/utils/user-agent.ts
emmenko: The "cli-login" value is not correct because the user agent is for the entire CLI and not one single command.
The value here refers to the HTTP library we are using for sending requests, so in this case `graphql-request`. The user agent also contains the library name `mc-scripts`, so we know it's the CLI.

packages/mc-scripts/src/commands/config-sync.ts
CarlosCortizasCT: So, that line was included in [this commit](https://github.com/commercetools/merchant-center-application-kit/commit/2edba6e21e25d93d3370f881e52518b8b81bb779) due to some Typescript checking;

`createdCustomApplication` is the result of calling `createCustomApplication` function which returns an optional object.

This is the function implementation:
```

const createCustomApplication = async ({
  mcApiUrl,
  token,
  organizationId,
  data,
}: TCreateCustomApplicationOptions) => {
  const variables = {
    organizationId,
    data,
  };

  try {
    const createdCustomAppData = await graphQLClient(mcApiUrl, token).request<
      TCreateCustomApplicationFromCliMutation,
      TCreateCustomApplicationFromCliMutationVariables
    >(CreateCustomApplicationFromCli, variables);
    return createdCustomAppData.createCustomApplication;
  } catch (error) {
    if (error instanceof Error) {
      throw new Error(error.message);
    }
    throw error;
  }
};
```

And this is the definition for the `TCreateCustomApplicationFromCliMutation` type:
```
export type TCreateCustomApplicationFromCliMutation = {
  __typename?: 'Mutation';
  createCustomApplication?: {
    __typename?: 'RestrictedCustomApplicationForOrganization';
    id: string;
  } | null;
};
```

So, the return statement `createdCustomAppData.createCustomApplication` can be technically _null_.

If we are sure that sending the create mutation will either work or raise an error, perhaps we can change the type definition to state `createCustomApplication` field is not optional.

This is the mutation sent:
```
mutation CreateCustomApplicationFromCli(
  $organizationId: String!
  $data: CustomApplicationDraftDataInput!
) {
  createCustomApplication(organizationId: $organizationId, data: $data) {
    id
  }
}
```

CarlosCortizasCT: @Rhotimee What's your take about updating the `TCreateCustomApplicationFromCliMutation` so the `createCustomApplication` is not optional?

emmenko: These types are generated so we can't change them. I'll check again if we can make the return type non optional.

Rhotimee: > @Rhotimee What's your take about updating the `TCreateCustomApplicationFromCliMutation` so the `createCustomApplication` is not optional?

Sorry, I missed the tagged comment. 

The line is not necessary, IMO. It was just added because of the type check error. 
I don't think it matters what we return there because an error will be thrown anyway if the custom application is not created.

CarlosCortizasCT (approved):
💯

ddouglasz (approved):

packages/mc-scripts/src/commands/build.ts
```diff
@@ -0,0 +1,153 @@
+import path from 'path';
+import fs from 'fs-extra';
+import webpack, { type Configuration, type Stats } from 'webpack';
+import chalk from 'chalk';
+import checkRequiredFiles from 'react-dev-utils/checkRequiredFiles';
```
kark, line 5: nit: I noticed that the type declaration file was not recognized in IDE until I added to `tsconfig.json`:
```json
{
  "extends": "@tsconfig/node16/tsconfig.json",
  "compilerOptions": {
    "typeRoots": ["../../@types", "../../@types-extensions"],
  }
}
```

packages/mc-scripts/src/cli.ts
```diff
@@ -0,0 +1,247 @@
+import fs from 'fs';
+import path from 'path';
+import { cac } from 'cac';
+import dotenv from 'dotenv';
+import dotenvExpand from 'dotenv-expand';
+import pkgJson from '../package.json';
```
kark, line 6: nit: perhaps we could also use
```json
{
  "extends": "@tsconfig/node16/tsconfig.json",
  "compilerOptions": {
    "resolveJsonModule": true
  }
}
```

packages/mc-scripts/src/commands/compile-html.ts
```diff
@@ -0,0 +1,38 @@
+import fs from 'fs';
+import chalk from 'chalk';
+import { compileHtml } from '@commercetools-frontend/mc-html-template';
+import paths from '../config/paths';
+import { TCliCommandCompileHtmlOptions } from '../types';
```
kark, line 5: nit: 
```suggestion
import type { TCliCommandCompileHtmlOptions } from '../types';
```

kark (approved):
Wowzers! 🎉

Rhotimee (approved):
Looks great 💯

emmenko:
Thanks @kark

PR #2620, opened 2022-05-16
docs: prefer using formik `handleReset` instead of `resetForm`

emmenko:
I've seen multiple times using `resetForm` passed as a callback function to some click handlers. For example

```js
<button onClick={formik.resetForm}>Click</button>
```

The `resetForm` method from Formik accepts an argument as the `nextState` and that could lead to unexpected issues if the function is called with the event `resetForm(event)`, like in the example above `onClick={formik.resetForm}`.

https://formik.org/docs/api/formik#resetform-nextstate-partialformikstatevalues--void

However, Formik also exposes a `handleReset` function (https://formik.org/docs/api/formik#handlereset---void) that does not accept any arguments.
This should be the preferred method to use as callback handlers like `onClick={formik.handleReset}`.

### DIFF
--- a/website-components-playground/src/pages/custom-form-modal-page.js
+++ b/website-components-playground/src/pages/custom-form-modal-page.js
@@ -66,10 +66,7 @@ const CustomFormModalPageExample = () => (
                   formControls={
                     <>
                       <CustomFormModalPage.FormSecondaryButton
-                        onClick={() => {
-                          formikProps.resetForm();
-                          setIsOpen(false);
-                        }}
+                        onClick={formikProps.handleReset}
                       />
                       <CustomFormModalPage.FormPrimaryButton
                         onClick={formikProps.handleSubmit}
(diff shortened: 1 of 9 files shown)


website-components-playground/src/pages/custom-form-modal-page.js
```diff
@@ -66,10 +66,7 @@ const CustomFormModalPageExample = () => (
                   formControls={
                     <>
                       <CustomFormModalPage.FormSecondaryButton
-                        onClick={() => {
-                          formikProps.resetForm();
-                          setIsOpen(false);
-                        }}
+                        onClick={formikProps.handleReset}
```
CarlosCortizasCT, line 69: Is this because we don't intentionally want the secondary button to close the modal in this _playground_? 🤔

CarlosCortizasCT: Ok. Just wanted to understand the change.

Thanks 👍

CarlosCortizasCT (approved):
👌

Issue #1668, opened 2020-08-06
Document "TypeScript" page


emmenko:
**Which part of the documentation does it affect and which topic does it refer to?**

https://docs.commercetools.com/custom-applications/development/typescript

**Describe what needs to be improved**

Document how to use the Custom Application with TypeScript.

> We might need to wait for the ui-kit to fully support TS first, so that we can provide more concrete examples.
> In the meantime, we can just mention that the appkit packages ship with TS declaration files.

PR #2624, opened 2022-05-24
docs: TypeScript support in Custom Applications and release `v21.8`

kark:
Ref #1668 

The aim of this PR was to document TypeScript support in Custom Applications.


website/src/content/development/adding-typescript.mdx
```diff
@@ -0,0 +1,95 @@
...
+
+```console noPromptLines="2-3"
+npx @commercetools-frontend/create-mc-app@latest \
+  my-new-custom-application-project \
+  --template starter-typescript
+```
+
+The TypeScript starter template is the same as the standard JS starter template in terms of functionality but it includes the additional TypeScript setup.
+
+# Configuration
```
CarlosCortizasCT, line 25: Here we explain about the needed configuration but, at least for me, it seems like this is something that I would need to do when creating a new Custom Applications.

Perhaps we could add a note at the end of the section explaining all of these steps are already preconfigured in our starter template and users don't need to worry about setting them up.

website/src/content/development/adding-typescript.mdx
```diff
@@ -0,0 +1,95 @@
...
+      "strict": false
+    }
+  }
+  ```
+
+  Please remember to revert that once the migration is over, as using `any` should be avoided.
+
+  </Info>
+
+* Add a script command `"typecheck": "tsc --noEmit"` in your `package.json` to compile the application.
```
CarlosCortizasCT, line 95: For me this is a little confusing since the command is called `typecheck`, which I understand as a command to validate the types in the application but we say this is to compile your application.

I thought our `mc-scripts build` command would also support building the TS application and the new `typecheck` command is only for "lintin" purposes.

website/src/releases/2022-07/release-v21.8.mdx
```diff
@@ -0,0 +1,48 @@
+---
```
CarlosCortizasCT, line 1: 👌

website/src/content/development/adding-typescript.mdx
emmenko: Good point. Technically "compiling" is what TypeScript does. In fact, running `tsc` does compile the project but the option `--noEmit` does not emit any files, which makes it look like it only checks the types.

To be honest this is confusing yes, a more explicit command would have been better, something like `tsc check` or something.

Let's see if we can try rewording this:

```suggestion
* Add a script command `"typecheck": "tsc --noEmit"` in your `package.json` to run the type checks of the application.
```

website/src/content/development/adding-typescript.mdx
emmenko: Thanks, updated here: https://github.com/commercetools/merchant-center-application-kit/pull/2624/commits/785d879b8aa392d7235f38076fef2b6d3e5adbee

website/src/content/getting-started.mdx
```diff
@@ -24,6 +24,8 @@ npx @commercetools-frontend/create-mc-app \
   --template starter
 ```
 
+If you want to start with a [TypeScript](https://www.typescriptlang.org/) project you can use the `--template=starter-typescript` option. Read more about TypeScript in the [Adding TypeScript](/development/adding-typescript) page.
```
Rhotimee, line 27: NIT

```suggestion
If you want to start with a [TypeScript](https://www.typescriptlang.org/) project you can use the `--template=starter-typescript` flag. Read more about TypeScript in the [Adding TypeScript](/development/adding-typescript) page.
```

Rhotimee (approved):
Looks great 🙌🏽

emmenko:
@adinakleine As soon as possible, as it's blocking the release 😇

Thanks 🙏

industrian:
> @adinakleine As soon as possible, as it's blocking the release 😇
> 
> Thanks 🙏

I'll review it now.

website/src/content/getting-started.mdx
industrian:
```suggestion
To start with a [TypeScript](https://www.typescriptlang.org/) project, use the `--template=starter-typescript` flag. Read more about TypeScript in the [Adding TypeScript](/development/adding-typescript) page.
```

website/src/content/development/adding-typescript.mdx
```diff
@@ -0,0 +1,101 @@
...
+title: Adding TypeScript
+---
+
+[TypeScript](https://www.typescriptlang.org/) is a typed superset of JavaScript that compiles to plain JavaScript.
+
+Custom Applications have full support for developing applications in TypeScript.
+
+<Info>
+
+This feature is available from version `>= 21.8.0`.
```
industrian, line 11:
```suggestion
This feature is available from version `21.8.0`.
```

Using the word "from" should imply greater than or equal to the stated version.

industrian (approved):
I have added some suggestions to improve readability, but it LGTM.

Also, I'm not sure when you wish to go live, but the date of the release note is currently `2022-07-04`

website/src/content/development/adding-typescript.mdx
emmenko: Thanks for the suggestion. If we change this, there are also other places in the docs that would need to be updated.

Can you confirm that it's clear that the "equal" part is implied? Not every user is a native english speaker, not I'm not sure if that's really obvious... 🤔

To be on the safe side, would it still be ok with you if we keep the `>=`?

emmenko:
@industrian Thanks Stephen, I'll integrate the feedback 🙌

website/src/content/development/adding-typescript.mdx
industrian: Sure, you could keep `>=` for now, or you could consider the following as well:

```suggestion
This feature is available from version `21.8.0` onwards.
```

emmenko: Thanks, I'll use the `onwards` version.

emmenko (approved):

emmenko (approved):

PR #2626, opened 2022-05-27
fix(mc-scripts): local login dev for internal usage

emmenko:
Caused by #2568

### DIFF
--- a/packages/mc-scripts/src/config/webpack-dev-server.config.js
+++ b/packages/mc-scripts/src/config/webpack-dev-server.config.js
@@ -45,14 +45,12 @@ module.exports = ({ port, publicPath }) => ({
   hot: true,
   https: false,
   port,
-  // Enable HTTPS if the HTTPS environment variable is set to 'true'
-  // `proxy` is run between `before` and `after` `webpack-dev-server` hooks
   setupMiddlewares(middlewares, devServer) {
     if (!devServer) {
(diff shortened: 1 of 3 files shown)


packages/mc-scripts/src/config/webpack-dev-server.config.js
```diff
@@ -45,14 +45,12 @@ module.exports = ({ port, publicPath }) => ({
...
   port,
-  // Enable HTTPS if the HTTPS environment variable is set to 'true'
-  // `proxy` is run between `before` and `after` `webpack-dev-server` hooks
   setupMiddlewares(middlewares, devServer) {
     if (!devServer) {
       throw new Error('webpack-dev-server is not defined');
     }
 
-    middlewares.push(createMcDevAuthenticationMiddleware(applicationConfig));
+    devServer.app.use(createMcDevAuthenticationMiddleware(applicationConfig));
```
emmenko, line 53: This is the fix. Apparently using the `middlewares` list didn't really work as expected.

tdeekens: Woah, I read the documentation to verify but it also does this. When debugging the production build on the middlewares, the only thing I noticed it that it wasn't called.

tdeekens (approved):

CarlosCortizasCT (approved):
👍

PR #2628, opened 2022-05-27
feat(application-config): shared `tsconfig.json`

kark:
Ref #1668 

#### Summary

Exposes a base `tsconfig.json` file to extend in Custom Applications.


packages/tsconfig/README.md
```diff
@@ -0,0 +1,23 @@
...
+$ npm install --save @commercetools-frontend/tsconfig
+```
+
+## Usage
+
+Add to your `tsconfig.json`:
+
+```json
+{
+  "extends": "@commercetools-frontend/tsconfig"
```
emmenko, line 21: This should be

```suggestion
  "extends": "@commercetools-frontend/tsconfig/tsconfig.json"
```

However, seeing this makes me wonder if we really need an extra package for it...maybe we can include the `tsconfig.json` file in one of the required packages like `application-config` or `application-shell`... 🤔

CarlosCortizasCT: If we update this example here, I believe we should also update it in the changeset.

Regarding the question raised about whether creating or not a new package, personally I think the `application-config` package would be a good fit to export this configuration file.

kark: Done in 6dbbc90aaca54fc841c6aa007e16372be4e3323a

CarlosCortizasCT (approved):
👍

emmenko (approved):
Having a file `tsconfig/tsconfig.json` seemed a bit unnecessary to me, so I renamed it to `tsconfig-mc-app.json`.

I also cleaned up a bit the default options and removed the `types` option, as this is usually project/repo specific and should be provided on demand.

PR #2633, opened 2022-06-01
Align children and legal message of PublicPageLayout component

CarlosCortizasCT:
#### Summary
Children in PublicPageLayout component are not correctly aligned when using a long legal message.

#### Description
Provided children to PublicPageLayout component have a container with a fixed width but the legal message that can be rendered below does not. So if that message is too long, both contents do not align well.

Example:
image

We used the same with for the legal message to avoid this problem.

An alternative could be to allow different widths and use a center alignment between the children and the legal message.

### DIFF
--- a/.changeset/bright-numbers-check.md
+++ b/.changeset/bright-numbers-check.md
@@ -1,0 +1,6 @@
+---
+"@commercetools-frontend/application-components": patch
+"@commercetools-local/visual-testing-app": patch
+---
+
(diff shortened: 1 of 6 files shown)


FilPob (approved):

tdeekens (approved):

kark (approved):

Rhotimee (approved):

emmenko (approved):
Thanks

PR #2635, opened 2022-06-02
feat: add .licrc for licensebat

tdeekens:
#### Summary

The licensebat is installed to the repository. This should enable the tool.


tdeekens:
Too bad. Seems like the tool fails to analyses things.

PR #2645, opened 2022-06-15
Update MC Proxy Router docs to explain URL rewrites

CarlosCortizasCT:
#### Summary

This is a proposal to update the docs regarding the MC Proxy Router on how it forwards request to external deployed Custom Applications since we had several issues about it.

#### Description

The idea is to further explain how MC Proxy Router forwards request to Custom Applications so users understand why they need to configure URL forwarding in their end.
We need their deployments to have this kind of setup where every request to the Custom Application configured URL must return the `index.html` file of their app, no matter what pathname is in the URL request.

[This is the preview link](https://appkit-pr-2645.commercetools.vercel.app/custom-applications/concepts/merchant-center-proxy-router) for the updated page.

### DIFF
--- a/website/src/content/concepts/merchant-center-proxy-router.mdx
+++ b/website/src/content/concepts/merchant-center-proxy-router.mdx
@@ -39,7 +39,7 @@ This is because the Custom Application handles client-side routing and it needs 
 
 You need to setup your deployment service to return the `index.html` file no matter which pathname exists in the Custom Application URL.
 
-### Example
+## Example
 
 Let's say you have deployed a Custom Application with these info:
 * Application URL: `https://my-app.netlify.app/`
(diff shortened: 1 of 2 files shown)


Rhotimee (approved):

ddouglasz (approved):

kark (approved):

CarlosCortizasCT:
> @CarlosCortizasCT Thanks a lot for pushing to improve the docs. I pushed a commit with some extra changes to clarify things a bit more in a more structured way, so that we send the same message in different places. [95732ac](https://github.com/commercetools/merchant-center-application-kit/commit/95732ac3575242e3c05cb0e0229b8da63a02ecee)
> 
> Have a look and let me know if that's ok with you as well.

Yes, it looks better.

Thanks for the improvement contribution 👍

emmenko (approved):

PR #2651, opened 2022-06-22
Fix ApplicationShell component aria roles issues

CarlosCortizasCT:
### Summary

There are some invalid accessibility aria roles being used in the `ApplicationShell` component.

### Description

* We are using a role named **application-layout** which does not exist. The most similar actual role id **application** but I don't think it applies to this use case as it is suitable for elements ["that do not follow a standard interaction pattern"](https://www.w3.org/TR/wai-aria/#application) which I don't think it is the case. So I think we should remove this role altogether.
* We are using a role named **header** (which does not exist) in a **header** element. We must remove that role.
* We are using a role named **aside** (which does not exist) in a **aside** element. We must remove that role.
* We are using a role named **global-notifications** which does not exist. Since the element using this role is supposed to contain dynamic notifications that can be shown to the user, I think we should update the role to be **region** and to add the **aria-live** attribute ([reference](https://www.w3.org/TR/wai-aria/#aria-live)). For the value of this attribute, I would use **polite** as we can't know for sure what type of notifications will be shown but it does not seem we need to break the speech of a screen reader.

### DIFF
--- a/.changeset/quick-plums-wonder.md
+++ b/.changeset/quick-plums-wonder.md
@@ -1,0 +1,9 @@
+---
+'@commercetools-frontend/application-shell': patch
+---
+
+Fixed aria role names.
+
+We were using invalid role names in some of the component's elements.
+
+Special mention to the element wrapping the notifications as it now uses the aria-live [attribute](https://www.w3.org/TR/wai-aria/#aria-live) (with **polite** value).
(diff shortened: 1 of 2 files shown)


emmenko (commented):
Can you double check that we are not using these attributes as selectors in any tests, both here and in mc-fe?

Just to make sure and to know what we would need to adjust in case.

CarlosCortizasCT:
> Can you double check that we are not using these attributes as selectors in any tests, both here and in mc-fe?
> 
> Just to make sure and to know what we would need to adjust in case.

I double checked and didn't find any test affected by this change.

ddouglasz (approved):

PR #2652, opened 2022-06-22
Update viewport meta in main html template for applications

CarlosCortizasCT:
### Summary

Remove *user-scalable=no* property from `viewport` meta tag as it is a problem from the accessibility point of view.

### Description

One of the findings of the accessibility report is about having a meta tag with *viewport* name that contains a value of *no* for *user-scalable* property.

(_This property only affects zooming in touch screen devices_).

As stated [in the documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Viewport_meta_tag#viewport_basics), this is an accessibility issue we should take care of.

### DIFF
--- a/.changeset/slimy-apples-grab.md
+++ b/.changeset/slimy-apples-grab.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/mc-html-template': patch
+---
+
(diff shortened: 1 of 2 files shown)


emmenko (commented):
Thanks!

PS: Remember to add a changeset before merging.

emmenko (approved):
💯

kark (approved):

PR #2654, opened 2022-06-23
`TabHeader` component accessibility improvements

kark:
#### Summary

Accessibility improvements of the `TabHeader` component.
> Every tab must have a tab aria role and all of them must be direct children of another element with tablist role.

`Lighthouse` benchmark changes (measured for the `visual-testing-app` at `/tabular-main-page`)
Before: 
screencapture-googlechrome-github-io-lighthouse-viewer-2022-06-23-14_46_28
After:
screencapture-googlechrome-github-io-lighthouse-viewer-2022-06-23-14_48_58

### DIFF
--- a/packages/application-components/src/components/tab-header/tab-header.tsx
+++ b/packages/application-components/src/components/tab-header/tab-header.tsx
@@ -19,10 +19,6 @@ const warnIfMissingContent = (props: TTabHeaderProps) => {
   );
 };
 
-const getDisabledTabHeaderAriaAttributes = (
-  isDisabled: TTabHeaderProps['isDisabled']
-) => (isDisabled ? { 'aria-disabled': true } : {});
-
 const getDisabledLinkAtributes = (isDisabled: TTabHeaderProps['isDisabled']) =>
   isDisabled ? { tabIndex: -1, 'aria-disabled': true } : {};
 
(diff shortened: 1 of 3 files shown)


packages/application-components/src/components/tab-header/tab.styles.ts
```diff
@@ -42,7 +42,7 @@ export const getLinkStyles = (
   isActive &&
     css`
       ${getBottomBorderStyles(customProperties.colorPrimary)}
-      > * > * {
+      > * {
```
CarlosCortizasCT, line 45: nitpick: perhaps we can take this opportunity to be more specific with this rule.
Since we now we are using an `h4` element within the link (and that's the element we want to address the color rule to), we could update it to be something like this:
```
& h4 {
```

packages/application-components/src/components/tab-header/tab.styles.ts
```diff
@@ -63,7 +63,7 @@ export const getLinkStyles = (
       :hover,
       :focus,
       :active {
-        > * > * {
+        > * {
```
CarlosCortizasCT, line 66: nitpick: perhaps we can take this opportunity to be more specific with this rule.
Since we now we are using an `h4` element within the link (and that's the element we want to address the color rule to), we could update it to be something like this:
```
& h4 {
```

CarlosCortizasCT (approved):
Look very good to me 👍 

Just left a minor suggestion in case you want to use it.

packages/application-components/src/components/tab-header/tab-header.tsx
```diff
@@ -84,20 +80,16 @@ export const TabHeader = (props: TTabHeaderProps) => {
...
       to={props.to}
       css={getLinkStyles(isActive, isDisabled)}
       {...getDisabledLinkAtributes(isDisabled)}
       {...dataAttributeProps}
     >
-      <div
-        role="tab"
-        aria-selected={isActive}
-        {...getDisabledTabHeaderAriaAttributes(isDisabled)}
-      >
```
emmenko, line 96: Can't remember why we implemented it like this 🤔

emmenko (approved):
Much better now, thanks for picking this up! 🙌

Rhotimee (approved):

ddouglasz (approved):

packages/application-components/src/components/tab-header/tab-header.tsx
kark: I suppose it's a leftover from [this refactor](https://github.com/commercetools/merchant-center-application-kit/pull/2503#discussion_r826076176)

packages/application-components/src/components/tab-header/tab.styles.ts
kark: thanks 👍 done in 8a752503b89e5e7125b3540d8d9174ed62840486

PR #2655, opened 2022-06-23
Fix `UserSettingsMenu` a11y issues

CarlosCortizasCT:
#### Summary

This PR addresses some accessibility issues found.

#### Description

We are using the Downshift external package to implement the dropdown for the user settings menu and the accessibility report we got complains about having an element with combobox role that does not have a corresponding accessible name.

Also, there is another warning about an element with the [`combobox`](https://www.w3.org/TR/wai-aria/#combobox) not having a direct element with a `listbox` role.

A11y issues explanation:
* [non-accessible names](https://web.dev/aria-name/?utm_source=lighthouse&utm_medium=devtools)
* [invalid child roles](https://web.dev/aria-required-children/?utm_source=lighthouse&utm_medium=devtools)


packages/application-shell/src/components/user-settings-menu/user-settings-menu.tsx
```diff
@@ -261,14 +266,19 @@ UserSettingsMenuBody.displayName = 'UserSettingsMenuBody';
...
+      <AccessibleHidden>
+        <span id={menuLabelId}>
+          <FormattedMessage {...messages.menuLabel} />
+        </span>
+      </AccessibleHidden>
+      <Downshift stateReducer={stateReducer} labelId={menuLabelId}>
         {(downshiftProps) => (
           <div>
             <button
-              role="user-menu-toggle"
```
CarlosCortizasCT, line 271: This is an invalid role so I removed it.
I also searched this and MC frontend repositories and didn't find any test using this role name.

CarlosCortizasCT:
Regarding the `combobox` not having a `listbox` as a direct child, [reading the docs](https://www.w3.org/TR/wai-aria-1.1/#combobox) it seems that's only needed when the dropdown is expanded.
This is controlled internally by `downshift` and, although it does not create the context menu as a direct child of the `combobox` element, it uses an [`aria-owns`](https://www.w3.org/TR/wai-aria-1.1/#aria-owns) property which is used when the parent-child HTML structure cannot be achieved to also create this _owning_ relationship.
Also, I run the lighthouse report several times and this warning is not present anymore.

packages/application-shell/src/components/user-settings-menu/user-settings-menu.tsx
```diff
@@ -261,14 +266,19 @@ UserSettingsMenuBody.displayName = 'UserSettingsMenuBody';
 
 const UserSettingsMenu = (props: Props) => {
   const intl = useIntl();
+  const menuLabelId = useFieldId(undefined, menuLabelSequentialId);
 
   return (
     <div data-test="user-settings-menu">
-      <Downshift stateReducer={stateReducer}>
+      <AccessibleHidden>
+        <span id={menuLabelId}>
```
emmenko, line 274: There is only one user menu in the page, so we can provide a fixed id here, no need for generating a random one.

emmenko (commented):
Thanks!!

One small remark otherwise looks good.

packages/application-shell/src/components/user-settings-menu/user-settings-menu.tsx
CarlosCortizasCT: Updated in this commit: eb558206564c5ed2d2277d5634f240c2b6764d97

emmenko (approved):
Very nice, thanks!

Rhotimee (approved):
Looks good.

kark (approved):
Works great 👍

PR #2660, opened 2022-06-24
🚀 Speed up CI time by splitting up jobs

emmenko:
As our CI pipeline is growing with many different things that we need to build and test, having one single workflow takes longer and longer. For example, having a second starter template (#2615 ) will increase the time even more.

As of now, a CI run takes between 15 and 20 minutes.

image

Using one workflow has some benefits, like installing dependencies and building packages once. However, it's highly sequential even though many steps don't relate with each other.

To speed things up and make our CI more resilient and useful, I experimented a bit with using multiple smaller jobs:
* `lint_and_test`: runs the main typechecking on the repo as well as all linters and tests
* `test_visual`: runs only the VRTs of the UI components
* `test_starter_template`: runs only the build and tests of the starter template application
* `test_starter_template_installation`: runs the installation workflow of the starter template application
* `test_playground`: runs only the build and tests of the playground application

As you can see, the grouping is done based on the different applications or use cases.

All these jobs share a common CI setup (installing dependencies and building packages) which has been extracted into a so-called [composite action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action).

One drawback is that every job now installs the dependencies and builds the packages on their own. However, it seems that the time it takes is less than 3 minutes (considering the cache) so it's overall acceptable considering that these jobs now take between 5 and 10 minutes.

Compared to the previous setup with a single workflow, it appears that by splitting up the jobs a CI run now takes less than a half of the time it took before 🚀

Additionally, all jobs now have a reduced timeout to 10 minutes.

NOTE: I split one more job into a separate workflow for deploying the playground app to Vercel. This workflow only runs on the main branch.

### DIFF
--- a/.github/actions/ci/action.yml
+++ b/.github/actions/ci/action.yml
@@ -1,0 +1,36 @@
+name: CI
+
+description: Shared action with basic setup for each job
+
+runs:
+  using: composite
+
+  steps:
+    - name: Setup Node (uses version in .nvmrc)
+      uses: actions/setup-node@v3
+      with:
+        node-version-file: '.nvmrc'
+
+    - name: Get yarn cache
+      id: yarn-cache
+      run: echo "::set-output name=dir::$(yarn config get cacheFolder)"
+      shell: bash
+
+    - uses: actions/cache@v3
+      with:
+        path: ${{ steps.yarn-cache.outputs.dir }}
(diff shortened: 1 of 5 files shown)


CarlosCortizasCT (approved):
💯

kark (approved):
🙌

tdeekens (approved):
Nice!

PR #2665, opened 2022-06-27
Fix `ApplicationShell` header selectors a11y issues

CarlosCortizasCT:
#### Summary

This PR addresses some accessibility issues found in locale and project selectors.

#### Description

The latests accessibility report complains about mentioned selectors (in the header of an application) as they don't have any label explaining what they serve for.

A11y issue explanation:
* [Form elements without labels](https://web.dev/label/)


packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -663,9 +663,8 @@ describe('when user does not have permissions to access the project', () => {
 });
 describe('when switching project', () => {
   it('should render app for new project', async () => {
-    renderApp(<label id="project-switcher">{'Project switcher'}</label>);
-    await screen.findByText('Project switcher');
-    const input = await screen.findByLabelText('Project switcher');
+    renderApp();
+    const input = await screen.findByLabelText('Projects menu');
```
CarlosCortizasCT, line 667: We don't need to add a label anymore since now the component renders one.

packages/application-shell/src/components/locale-switcher/locale-switcher.tsx
```diff
@@ -14,6 +17,8 @@ type Props = {
   availableLocales: string[];
 };
 
+const LOCALE_SWITCHER_LABEL_ID = 'locale-switcher-label';
```
CarlosCortizasCT, line 20: We have a constant field since there is only one instance of this component in the page.

packages/application-shell/src/components/locale-switcher/messages.ts
```diff
@@ -0,0 +1,9 @@
+import { defineMessages } from 'react-intl';
+
+export default defineMessages({
+  localesLabel: {
+    id: 'LocaleSwitcher.localesLabel',
+    description: 'The label for project dropdown switcher',
+    defaultMessage: 'Locales menu',
```
CarlosCortizasCT, line 7: I used the `menu` word instead of `switcher` or `selector` as that's the word we use for the user menu dropdown.
Let me know if you think we should change it for the switchers.

packages/application-shell/src/components/project-switcher/messages.ts
```diff
@@ -1,10 +1,10 @@
 import { defineMessages } from 'react-intl';
 
 export default defineMessages({
-  projects: {
-    id: 'ProjectSwitcher.projects',
+  projectsLabel: {
```
CarlosCortizasCT, line 4: `projects` key message was not used so I decided to remove it and create a new one more specific to the context where it's used.

packages/application-shell/src/components/project-switcher/project-switcher.spec.tsx
```diff
@@ -13,16 +13,10 @@ const render = () => {
...
-    </>,
-    {
-      disableRoutePermissionCheck: true,
-      mocks: mockedRequest,
-    }
-  );
+  return renderApp(<ProjectSwitcher projectKey="key-0" />, {
+    disableRoutePermissionCheck: true,
+    mocks: mockedRequest,
+  });
```
CarlosCortizasCT, line 19: We don't need to render a label anymore since now the component itself renders one.

packages/application-shell/src/components/project-switcher/project-switcher.spec.tsx
```diff
@@ -32,8 +26,7 @@ describe('rendering', () => {
 
   it('should search and select a project', async () => {
     render();
-    await screen.findByLabelText('Project switcher');
-    const input = screen.getByLabelText('Project switcher');
+    const input = await screen.findByLabelText('Projects menu');
```
CarlosCortizasCT, line 29: We were finding and element and then getting it but it seems more straightforward just to do it in one line.

packages/application-shell/src/components/project-switcher/project-switcher.tsx
```diff
@@ -32,6 +33,8 @@ type CustomValueContainerProps = ValueContainerProps & {
   projectCount: number;
 };
 
+const PROJECT_SWITCHER_LABEL_ID = 'project-switcher-label';
```
CarlosCortizasCT, line 36: We have a constant field since there is only one instance of this component in the page.

packages/application-shell/src/components/locale-switcher/messages.ts
emmenko: Remember to run this command to extract the new messages:

https://github.com/commercetools/merchant-center-application-kit/blob/6cf96ee64d59177675a1d0bdea9a1ef953596198/package.json#L9

CarlosCortizasCT: Thanks for the reminder.

Updated here: 212e4d4e65f410075bf84a2090ef428372995829

kark (approved):
Neat!

Rhotimee (approved):
Looks great. 
Thanks

PR #2666, opened 2022-06-27
refactor(create-mc-app): migrate to TypeScript

emmenko:
Similarly to `mc-script` CLI, the `create-mc-app` CLI has been migrated to TS. And we're also using the `cac` library for the parsing logic.


packages/create-mc-app/src/cli.ts
```diff
@@ -0,0 +1,109 @@
+import { cac } from 'cac';
+import { Listr, ListrTask } from 'listr2';
+import * as tasks from './tasks';
+import { throwIfNodeVersionIsNotSupported } from './validations';
+import { shouldUseYarn } from './utils';
+import hintOutdatedVersion from './hint-outdated-version';
+import processOptions from './process-options';
+import { TCliCommandOptions } from './types';
```
kark, line 8: nit: 
```suggestion
import type { TCliCommandOptions } from './types';
```

packages/create-mc-app/src/validations.ts
```diff
@@ -1,18 +1,30 @@
-const fs = require('fs');
-const path = require('path');
-const { isSemVer } = require('./utils');
+import fs from 'fs';
+import path from 'path';
+import semver from 'semver';
+import { TCliCommandOptions } from './types';
```
kark, line 4: nit: 
```suggestion
import type { TCliCommandOptions } from './types';
```

packages/create-mc-app/src/validations.ts
```diff
@@ -1,18 +1,30 @@
...
+import { isSemVer } from './utils';
 
-const availableTemplates = ['starter', 'starter-typescript'];
+const availableTemplates = {
+  starter: 'starter',
+  'starter-typescript': 'starter-typescript',
+} as const;
```
kark, line 10: 🎉

packages/create-mc-app/src/cli.ts
```diff
@@ -0,0 +1,109 @@
+import { cac } from 'cac';
+import { Listr, ListrTask } from 'listr2';
```
kark, line 2: It's more a question than a suggestion - I was wondering if you'd see any downside of importing `ListrTask` interface as type here? (It's also imported in a few other places). I'm asking since it seems to me it was done like this for a reason.

kark (approved):
Works great 🙌

packages/create-mc-app/src/cli.ts
tdeekens: `import { type TCliCommandOptions } from './types';` would also work in newer TS versions so you can mix types and regular imports.

packages/create-mc-app/src/cli.ts
```diff
@@ -0,0 +1,109 @@
...
+        return;
+      }
+
+      await hintOutdatedVersion(pkgJson.version);
+
+      console.log('');
+      console.log(
+        `Documentation available at https://docs.commercetools.com/custom-applications`
+      );
+      console.log('');
```
Rhotimee, line 69: why not use `\n` for new lines instead?

packages/create-mc-app/src/cli.ts
```diff
@@ -0,0 +1,109 @@
...
+      );
+
+      await taskList.run();
+      const useYarn = shouldUseYarn();
+
+      console.log('');
+      console.log(
+        `🎉 🎉 🎉 The Custom Application has been created in the "${taskOptions.projectDirectoryName}" folder.`
+      );
+      console.log('');
```
Rhotimee, line 90: Any reason we are not using `\n` for new lines instead?

packages/create-mc-app/src/tasks/update-application-constants.ts
```diff
@@ -0,0 +1,61 @@
+import os from 'os';
+import fs from 'fs';
+import path from 'path';
+// @ts-ignore
+import rcfile from 'rcfile';
+import prettier from 'prettier';
+import { type PluginItem, transformFileSync, types } from '@babel/core';
+import { ListrTask } from 'listr2';
```
Rhotimee, line 8:
```suggestion
import { type ListrTask } from 'listr2';
```

packages/create-mc-app/src/tasks/update-custom-application-config.ts
```diff
@@ -0,0 +1,85 @@
+import os from 'os';
+import fs from 'fs';
+import path from 'path';
+// @ts-ignore
+import rcfile from 'rcfile';
+import prettier from 'prettier';
+import { transformFileSync, types, type PluginItem } from '@babel/core';
+import { ListrTask } from 'listr2';
```
Rhotimee, line 8:
```suggestion
import { type ListrTask } from 'listr2';
```

packages/create-mc-app/src/cli.ts
emmenko: Thanks @kark.

@tdeekens If we're only importing types, we kind of prefer to use `import type`, as otherwise you need to declare the `type` for every named import.

packages/create-mc-app/src/cli.ts
emmenko: We should use `type ListrTask` yes, thanks for pointing it out.

packages/create-mc-app/src/cli.ts
emmenko: No particular reason, in this case I just find it a bit easier to read that there is a new line.

Issue #2634, opened 2022-06-02
Custom Application - proxying requests to the external API


alochowski:
I have a question about integrating custom application with my own API - especially genrated JWT token there.
https://docs.commercetools.com/custom-applications/concepts/integrate-with-your-own-api#exchange-token-structure

Is it possible to set **`aud`** param in payload on my own?

Because now this **audience** param is set as provided **`uri`** param in SDK, but our **uri** contain dynamic parts.

We are using https://oauth2-proxy.github.io/oauth2-proxy/ and we expect constant **audience** but for now we have the same as **uri**.

Better for us will be set **audience** as constant value representing e.g. service name or origin URL (protocol, hostname, port) without all path.

Is there any chance to set somehow this **`aud`** in JWT token? Or you are going to add such as possibilities by your SDK?

Regards,
Arek

CarlosCortizasCT:
Hi @alochowski 

Thanks for opening this issue.

I guess we would need a little more background to actually understand what is the problem you are facing.

We know you will send a request from the browser (Custom Application) to the Merchant Center API (configuring it to be forwarded) and we will forward that request to you backend from our gateway sending the JWT token where the `aud` claim will contain the value of the full URL of your endpoint.

It seems that request will first hit you `oauth2-proxy` and then it will get to your backend service where, if you use our [express packages](https://docs.commercetools.com/custom-applications/concepts/integrate-with-your-own-api#validating-the-json-web-token), you could validate the token.

It would be helpful to really understand the steps of your flow, what happens in every step, what causes the problem and what would you need to be different.

Thanks.

alochowski:
Hi @CarlosCortizasCT ,

Thanks for your response.

Everything is as you wrote, except last step - backend service in which we should do validate token.

We don't have implementation on backend side, our backend is security-agnostic.
This is why we have proxy according to the [specification](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3) 
_Each principal intended to process the JWT MUST identify itself with a value in the audience claim._ 

It would be great for us if we will have opportunities to set audience not only per resource (_uri_), but also e.g.  per API or custom string.

CarlosCortizasCT:
Hi @alochowski 

Thanks for your reply.

So, currently we don't support modifying the `aud` claim our gateway uses to create the JWT token that it's used in the proxied request to your backend.

The flow for a proxied request would be something like this example:
[...]

PR #2667, opened 2022-06-28
feat: add support for setting the audience policy

emmenko:
Ref #2634 

To be merged once the functionality is available in the MC API.


packages-backend/express/src/auth.ts
```diff
@@ -119,10 +119,17 @@ export const getConfiguredAudience = <Request extends TBaseRequest>(
...
+
+  switch (options.audiencePolicy) {
+    case 'forward-url-origin':
+      return url.origin;
+    default: {
+      if (requestPath === '/') {
+        return url.origin;
+      }
+      return `${url.origin}${url.pathname}`;
+    }
```
emmenko, line 131: This is the main logic to actually apply the policy.

.changeset/perfect-drinks-teach.md
```diff
@@ -0,0 +1,19 @@
...
+Supported values are:
+
+- `forward-url-full-path`: This is the default policy. It sets the `audience` using the full URL (origin + pathname).
+- `forward-url-origin`: This is the alternative policy. It sets the `audience` using only the origin URL part.
+
+```js
+createSessionAuthVerifier({
+  audiencePolicy: 'forward-url-origin',
+  // ...
+});
```
CarlosCortizasCT, line 16: Since we offer both `createSessionAuthVerifier` and `createSessionMiddleware` from this package, perhaps we could also include an example for the latter.

```suggestion
createSessionAuthVerifier({
  audiencePolicy: 'forward-url-origin',
  // ...
});

createSessionMiddleware({
  audiencePolicy: 'forward-url-origin',
  // ...
});
```

.changeset/perfect-drinks-teach.md
```diff
@@ -0,0 +1,19 @@
...
+
+// "X-Forward-To-Audience-Policy": "forward-url-origin"
```
CarlosCortizasCT, line 18: Not sure about this header example since it made sense in the `application-shell` changeset because that's the header attached to the request with the context created by the helper but here it seems less clear.
Maybe we want to say that these backend helpers will use that header?

CarlosCortizasCT (approved):
This looks great to me. 💯 

Just left a couple of nitpick suggestions.

packages-backend/express/src/types.ts
```diff
@@ -9,33 +9,48 @@ export type TIssuer = string;
...
+  /**
+   * The public-facing URL used to connect to the server / serverless function.
+   * The value should only contain the origin URL (protocol, hostname, port),
+   * the request path is inferred from the incoming request.
+   */
   audience: TAudience;
-  // The cloud identifier (see `CLOUD_IDENTIFIERS`) that maps to the MC API URL
-  // of the related cloud region or the MC API URL.
+  /**
+   * How the audience value should be exchange between the server and the client.
```
tdeekens, line 28:
```suggestion
   * How the audience value should be exchanged between the server and the client.
```

tdeekens (commented):
Looks very nice.

kark (approved):
🚀

ddouglasz (approved):
💯

.changeset/perfect-drinks-teach.md
emmenko: True, good catch!

emmenko:
```suggestion
```

emmenko:
The API changes have been deployed. This can be merged now.

PR #2676, opened 2022-07-01
fix(create-mc-app): formatting of patched constants file

emmenko:
To prevent lint errors when installing the template.


CarlosCortizasCT (approved):
👌

PR #2677, opened 2022-07-01
Fix i18n extract command in TS starter template

CarlosCortizasCT:
Fixes the pattern used to extract i18n messages in the `extract-intl` command (script in the `package.json`) for the Typescript starter template.


emmenko (approved):
Thanks!

kark (approved):

PR #2679, opened 2022-07-04
Add support for dynamic loading moment.js locales

CarlosCortizasCT:
### Summary
MC users can select from a wide variety of options when choosing the interface locale from their user settings.
Selected locale not only affects the language in which the texts are displayed but it also is used to format dates and numbers across MC.

When a user selects some of the available locales (ex: en-AU) dates are not using the expected format.

The problem here is that we are using moment in many places ([FormattedDateTime](https://github.com/commercetools/merchant-center-frontend/tree/main/packages-shared/core/components/formatted-date-time) component) to format dates but we are not loading the right locale metadata needed for that library.

### Description
When a MC application is loaded, we use the [AsyncDataLocale](https://github.com/commercetools/merchant-center-application-kit/blob/main/packages/i18n/src/async-locale-data/async-locale-data.tsx) component to load all locale related data: we load the application translations, app-kit and ui-kit translations and also we try to load the moment metadata file corresponding to selected locale.

The problem is that we don't allow to load any metadata file but [only the ones corresponding to the main locales](https://github.com/commercetools/merchant-center-application-kit/blob/main/packages/i18n/src/load-i18n.ts#L20) officially supported (en, de, es, fr, ja, zh-CN). So, if the user selects any regional locale, only the main locale metadata file will be loaded.

For the example with the en-AU locale, this means that moment will use the en metadata file (american english) so dates will be formatted with an american pattern, which is different than the australian one.

We would need to support all different combinations `moment.js` [already supports](https://github.com/moment/moment/tree/develop/src/locale) but we need to be aware that _dynamic import expressions_ have [these limitations](https://github.com/rollup/plugins/tree/master/packages/dynamic-import-vars#limitations) in rollup (which is the library preconstruct uses under the hood when building packages).

### Proposal
The idea here is to copy into our `i18n` package all the moment locale files so we can reference them with a relative path when dynamically loading them from the `AsyncDataLocale` component.

Since we already have a folder called `compile-data` in this package to hold the translations, I think it's also a good fit for these `moment.js` locale metadata files.

When building the packages we call the `compile-data` command from `i18n` package so I extended that one to include importing `moment.js` files.


packages/i18n/src/utils.ts
```diff
@@ -11,15 +11,6 @@ export const mergeMessages = (
   ...messages: TMessageTranslations[]
 ): TMessageTranslations => Object.assign({}, ...messages);
 
-export const mapLocaleToMomentLocale = (locale: string): string => {
```
CarlosCortizasCT, line 14: This is not needed anymore since we now try to load the whole locale and have some fallback logic to deal with it in the loader file (`load-i18n.ts`).

packages/i18n/src/load-i18n.ts
```diff
@@ -1,49 +1,34 @@
...
+
+  try {
+    await import(
+      /* webpackChunkName: "i18n-moment-locale-[name]" */ `../compiled-data/moment/locales/${locale}.js`
+    );
+  } catch (error) {
+    // If not-found locale was a regional one, we use the main language by default,
+    // otherwise we use the default locale (already included).
+    const [language, region] = locale.split('-');
+    if (region && language !== MOMENT_INCLUDED_METADATA_LOCALE) {
```
CarlosCortizasCT, line 27: If `locale` is an english regional one (eg: `en-CO`) and it does not exist, we don't want to load the main one ((it would be `en`) as that one is already bundled with `moment.js`.

packages/i18n/scripts/import-moment-locales.sh
```diff
@@ -0,0 +1,5 @@
+echo "Importing moment.js locale metadata files."
+mkdir -p ./compiled-data/moment/locales
+cp -R ../../node_modules/moment/dist/locale/*.js ./compiled-data/moment/locales
+# copied files have a relative import to 'moment' so we need to update it to be global
```
CarlosCortizasCT, line 4: Imported files look like this:
image

We need to update the import to be:
```
import moment from `moment`;
```
_(`moment` is a direct dependency of this package)_.

jest.eslint.config.js
```diff
@@ -12,6 +12,7 @@ module.exports = {
     'dist/',
     'public/',
     'generated',
+    'compiled-data',
```
CarlosCortizasCT, line 15: We don't want to validate generated/imported code.

packages/mc-scripts/src/config/create-webpack-config-for-development.ts
```diff
@@ -179,10 +181,9 @@ function createWebpackConfigForDevelopment(
         }),
       mergedOptions.toggleFlags.generateIndexHtml &&
         new LocalHtmlWebpackPlugin(),
-      // Strip all locales except `en`, `de`
-      // (`en` is built into Moment and can't be removed).
+      // Only keep locales that are available in the Merchant Center.
       new MomentLocalesPlugin({
-        localesToKeep: ['de', 'es', 'fr', 'zh-cn', 'ja'],
+        localesToKeep: momentLocalesToKeep,
```
emmenko, line 186: This is the result in the bundle

image

emmenko (approved):
As discussed with @CarlosCortizasCT , we are now generating the moment imports using code generation, based on the list of languages/locales (from the `l10n` package) that is used in the user profile.

The main reason here is to avoid "messing up" with the internal files in the `moment` package.

In addition to loading the locales, there was also the moment webpack plugin to include the list of moment locales to keep, which is now based on the same logic.

emmenko:
@CarlosCortizasCT Please have another look at the new changes. Thanks 🤗

packages/l10n/moment-utils.js
```diff
@@ -0,0 +1,113 @@
...
+
+async function loadMomentLocales(locale: string): Promise<void> {
+  const lowercaseLocale = locale.toLowerCase();
+
+  switch (lowercaseLocale) {
+  ${allAvailableLocalesWithMatchingMomentLocale
+    .map(
+      ({ locale, momentLocale }) => `
+    case '${locale}':
+      import('moment/locale/${momentLocale}');
```
CarlosCortizasCT, line 95: Since this is an `async` function, I think we would need to `await` for the `import` so we know for sure that the locale has been loaded after `loadMomentLocales` has been called.

```suggestion
      await import('moment/locale/${momentLocale}');
```

CarlosCortizasCT (commented):
This looks better now; thanks Nicola 🙌 

I just left a minor comment.

packages/l10n/moment-utils.js
emmenko: Yes good point, thanks

CarlosCortizasCT:
@emmenko maybe we could also add a new view to the playground where we use some date, number and currency just to showcase. (I don't see we-re testing the `ApplicationShell` or related components in the `visual-test-app` but maybe we should add it over there instead).

Is it ok if I try to do it?

emmenko:
@CarlosCortizasCT Sure, good idea thanks!

packages/l10n/moment-utils.js
```diff
@@ -0,0 +1,115 @@
...
+  // Assign to the available locales in the Merchant Center the corresponding
+  // matching moment locale, to make it easier to generate the code.
+  const allAvailableLocalesWithMatchingMomentLocale = allAvailableLocales
+    .map((availableLocale) => {
+      // Attempt to match a moment locale as-is.
+      const matchingLocale = momentLocaleDataFiles.find(
+        (momentLocale) => momentLocale === availableLocale
+      );
+      // Attempt to match a moment locale using only the language tag.
+      // This is used as a fallback in case the exact match didn't work.
```
ddouglasz, line 50: Oh, this makes sense, I was a bit curious about duplicates  but I see it has been taken care of below 🙌🏾. Thanks for the comment.

ddouglasz (approved):
👏

CarlosCortizasCT:
I added a new view (this is the commit: bf2d858cb7b2cc7a8ed0ef7203a87a6c6ddcaeb5) in the playground to showcase date and number formatting but it is not working properly: number is correctly formatted (we're using `formatjs` for that) but dates are not (we use `moment`):

image

Inspecting the network calls I can see the right `moment` locale file is being loaded, but I also noticed we have two instances of `moment` library loaded in the browser.
When the locale file is loaded, it registers itself in one of the `moment` instances (both are the same version) but when formatting the dates from the components, the other instance is used so it does not have the expected locale info.

image

One of the instances is in `esm` format and the other one is `cjs`.

I also tried to build and run the prod versions of the playground but the problem still exists.

I'm still trying to figure out what the root cause is but I wanted to push the change and share this comment in case anyone has an idea about what the problem could be.

packages/l10n/moment-utils.js
```diff
@@ -0,0 +1,115 @@
...
+    );
+
+  return supportedMomentLocales;
+}
+
+function generateMomentLocaleImports() {
+  const allAvailableLocalesWithMatchingMomentLocale =
+    getListOfAvailableLocalesWithMatchingMomentLocale();
+
+  const code = `/// <reference path="../moment-locales.d.ts" />
```
CarlosCortizasCT, line 85: This line seems to be the cause of [the error in CI](https://github.com/commercetools/merchant-center-application-kit/runs/7231505280?check_suite_focus=true) when typechecking:

image

I'm not sure why we need it.
I tried locally to remove that reference and building and typechecking seem to work.

Maybe @emmenko you know how to tackle the issue.

emmenko: I added this as otherwise lint-staged (using `tsc-files`) wouldn't work. I'll take another look at this then.

emmenko:
@CarlosCortizasCT Thanks for working on the playground. Let me know if we need to have a look together at the issue.

CarlosCortizasCT:
> I added a new view (this is the commit: [bf2d858](https://github.com/commercetools/merchant-center-application-kit/commit/bf2d858cb7b2cc7a8ed0ef7203a87a6c6ddcaeb5)) in the playground to showcase date and number formatting but it is not working properly: number is correctly formatted (we're using `formatjs` for that) but dates are not (we use `moment`):
> 
> image
> 
> Inspecting the network calls I can see the right `moment` locale file is being loaded, but I also noticed we have two instances of `moment` library loaded in the browser. When the locale file is loaded, it registers itself in one of the `moment` instances (both are the same version) but when formatting the dates from the components, the other instance is used so it does not have the expected locale info.
> 
> image
> 
> One of the instances is in `esm` format and the other one is `cjs`.
> 
> I also tried to build and run the prod versions of the playground but the problem still exists.
> 
> I'm still trying to figure out what the root cause is but I wanted to push the change and share this comment in case anyone has an idea about what the problem could be.

Regarding this issue, I found the problem was the way we imported `moment` locale files.

We were importing files like this:
```
await import('moment/locale/${momentLocale}');
```
and those files have a relative reference to the actual `moment` library like this:
```
factory(require('../moment'))
```
so they were requiring the file `node_modules/moment/moment.js` but the rest of the dependencies in the app, they use the library exported from the `dist` directory (`node_modules/dist/moment/moment.js`) so we had two instances:
* `node_modules/dist/moment/moment.js`: this is the one used by the application and `ui-kit` components
* `node_modules/moment/moment.js`: this is the one used by the moment locale files (which auto-register themselves in that instance)

The problem is that the locale data files were registering themselves in one instance, but the app used the other one for the actual formatting, so no locale data was found when formatting.

The solution is to require the locale data files that also exist in the `dist` directory of the package (ex: `node_modules/moment/dist/locale/en-GB.js`).

The issue was addressed in this commit: 945b7559428100078c977f386dea67c944d03fc6

CarlosCortizasCT:
There was [another issue](https://github.com/commercetools/merchant-center-application-kit/runs/7234653367?check_suite_focus=true) with one of the tests in the `ApplicationShellProvider` component:

image

The problem here is about that test using internally the `AsyncLocaleData` which in turn was trying to load one of the `moment` locale data files. As this files are ESM based, the test failed because `jest` does not support it.

The two alternatives I found were:
1. Update jest configuration to transpile `moment` locale files
2. Mock the locale data file loaded in the test (its contents are not actually used)

I opted to implement the second option as it seems cleaner to me.
Commit: db2be1c6fb3f4bb4a8b18e862193b185d10736e2

CarlosCortizasCT:
I updated the new playground view to showcase formatted dates and numbers in several locales at the same time following @emmenko idea.

image

Updated in this commit: c7833b4bc685af2b07c014beecd4b9181a4c9e46

emmenko:
Super nice, thanks a lot!! 🥳

tdeekens:
This is turning out super nice! Thanks all!

emmenko:
@CarlosCortizasCT As a follow up maybe, I was wondering if we can also add some E2E tests or VRTs to make sure we don't introduce regressions to the formattings.

What do you think?

emmenko:
@CarlosCortizasCT Also, we forgot to add the changeset 🙈 Can you add a PR with that? Thanks

CarlosCortizasCT:
> @CarlosCortizasCT Also, we forgot to add the changeset 🙈 Can you add a PR with that? Thanks

Sorry, my bad 😞 

I created this PR: #2687 2687

CarlosCortizasCT:
> @CarlosCortizasCT As a follow up maybe, I was wondering if we can also add some E2E tests or VRTs to make sure we don't introduce regressions to the formattings.
> 
> What do you think?

So that was what I tried to ask in my [previous comment](https://github.com/commercetools/merchant-center-application-kit/pull/2679#issuecomment-1176405056).

I think it would also make sense to have a VRT for this, but then maybe I would just move it from the playground as I don't see the need for having it in both apps since I guess our primary goal is validation more than visualization.

emmenko:
> I would just move it from the playground as I don't see the need for having it in both apps

Not sure I follow. We can keep it in the playground and take a snapshot in the E2E test, as we [already do](https://github.com/commercetools/merchant-center-application-kit/tree/main/cypress/integration/playground) with other tests in the playground app.

If you meant to have "normal" integration tests, then I guess we can also test it with RTL as well.

So yeah, maybe VRTs are not needed in this case.

PR #2680, opened 2022-07-04
fix(mc-scripts): parsing of CLI options casing

emmenko:
Small regression of #2613 

The CLI options are now parsed as camelCase with the [parsing library we're using now](https://www.npmjs.com/package/cac).

I totally missed that part of the docs:

image


CarlosCortizasCT (approved):
LGTM 👌

tdeekens (approved):

PR #2682, opened 2022-07-04
Add workflow to handle staled issues

CarlosCortizasCT:
#### Summary

Add a new GitHub workflow so we can automatically handle those opened issues which we already replied to but the author do not reply back for a long time.

#### Description

Configuration:
* Requires the issue to have a specific label
* 14 days to mark it as stale
* another 14 days (on top of the previous ones) to close it

### DIFF
--- a/.github-labelsrc.json
+++ b/.github-labelsrc.json
@@ -1,7 +1,7 @@
 [
   {
     "id": 4294819708,
-    "name": "⌛ Status: Waiting for Customer",
+    "name": "⌛ Status: Awaiting for feedback",
     "color": "000000",
     "description": "Waiting for the answer from customer"
   },
--- a/.github-labelsrc.json
+++ b/.github-labelsrc.json
@@ -1,5 +1,11 @@
 [
   {
+    "id": 4294819708,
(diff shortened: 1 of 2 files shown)


.github-labelsrc.json
```diff
@@ -1,4 +1,10 @@
 [
+  {
+    "id": 4294819708,
+    "name": "⌛ Status: Waiting for Customer",
```
Rhotimee, line 4: I don't think the author of the issue necessarily needs to be the customer. 
I feel like we can use `user` instead of `customer`. 

I wonder what the rest of the team thinks about this though.

emmenko: We can keep it a bit more generic:

```suggestion
    "name": "⌛ Status: Awaiting for feedback",
```

emmenko (commented):
Thanks!

.github-labelsrc.json
CarlosCortizasCT: Updated in this commit: 5bada16b028c5bdb1af5e803162edc1c44c496f3

Rhotimee (approved):
💯

emmenko (approved):
💯

PR #2691, opened 2022-07-12
refactor: add support in Vite for loading `.react.svg` files as React components

emmenko:
Does the same thing [as the Webpack setup](https://github.com/commercetools/merchant-center-application-kit/blob/702a7b26b050c46215eed273a2d47b885deda611/packages/mc-scripts/src/config/create-webpack-config-for-development.ts#L238-L256).


packages/application-config/client.d.ts
```diff
@@ -1,5 +1,13 @@
 type CSSModuleClasses = { readonly [key: string]: string };
 
+declare namespace NodeJS {
+  // Used for global variable replacements.
+  interface ProcessEnv {
+    readonly DEBUG: 'true' | 'false';
+    readonly NODE_ENV: 'development' | 'production' | 'test';
+  }
+}
```
emmenko, line 9: Slightly unrelated but noticed this in the CRA setup: https://github.com/facebook/create-react-app/blob/f34d88e30c7d8be7181f728d1abc4fd8d5cd07d3/packages/react-scripts/lib/react-app.d.ts#L5-L10

packages/application-config/client.d.ts
```diff
@@ -1,5 +1,13 @@
 type CSSModuleClasses = { readonly [key: string]: string };
 
+declare namespace NodeJS {
+  // Used for global variable replacements.
+  interface ProcessEnv {
+    readonly DEBUG: 'true' | 'false';
+    readonly NODE_ENV: 'development' | 'production' | 'test';
```
tdeekens, line 7: Don't we assign a value to this somewhere? Maybe then it can't be readonly.

emmenko: Yes right, let me remove it then as it's then more confusing.

packages/mc-scripts/src/commands/start-vite.ts
```diff
@@ -111,6 +112,26 @@ async function run() {
           plugins: ['@emotion/babel-plugin'],
         },
       }),
+      pluginSvgr({
```
CarlosCortizasCT, line 115: I was wondering if we also need to include this plugin in the `build-vite` command as well.

emmenko: Good point, yes we do. I forgot.

tdeekens (approved):

PR #2693, opened 2022-07-13
feat(codemod): add transform `rename-mod-css-to-module-css`

emmenko:
This codemod renames `.mod.css` files to `.module.css` and updates imports.

The file extension for CSS modules is officially `.module.css` and thus some tools rely on that naming (e.g. Vite).
This is the same as the `.js` to `.jsx` renaming.


packages/codemod/src/cli.ts
```diff
@@ -1,35 +1,38 @@
 import path from 'path';
-import mri from 'mri';
+import cac from 'cac';
```
emmenko, line 2: I took the chance to use the `cac` library for configuring the CLI.

packages/codemod/src/transforms/rename-mod-css-to-module-css.ts
```diff
@@ -0,0 +1,53 @@
...
+          '.mod.css',
+          '.module.css'
+        );
+        node.source.value = nextCssModulesFilePath;
+        hasModifications = true;
+
+        fs.renameSync(
+          path.join(path.dirname(file.path), previousCssModulesFilePath),
+          path.join(path.dirname(file.path), nextCssModulesFilePath)
+        );
```
emmenko, line 36: Side effect of the transform (which updates the import) to also rename the actual CSS module file.

application-templates/starter/src/components/welcome/welcome.jsx
```diff
@@ -7,7 +7,7 @@ import { AngleRightIcon } from '@commercetools-uikit/icons';
 import Spacings from '@commercetools-uikit/spacings';
 import Text from '@commercetools-uikit/text';
 import messages from './messages';
-import styles from './welcome.mod.css';
+import styles from "./welcome.module.css";
```
CarlosCortizasCT, line 10: Maybe we should keep single quotes here.

packages/codemod/test/__snapshots__/transforms.spec.ts.snap
```diff
@@ -42,3 +42,8 @@ function Modal() {
 
 export default Modal;"
 `;
+
+exports[`testing transform "rename-mod-css-to-module-css" transforms correctly 1`] = `
+"// eslint-disable-next-line import/extensions, import/no-unresolved, no-unused-vars
+import styles from \\"./styles.module.css\\";"
```
CarlosCortizasCT, line 48: nit: It seems the import is changed from single quotes to double quotes. Maybe we could try to keep single quotes.

application-templates/starter/src/components/welcome/welcome.jsx
emmenko: Fixed here: https://github.com/commercetools/merchant-center-application-kit/pull/2693/commits/acef0769b03e4723df46a357538fc38d9dd4a71e

Lint staged didn't include `.jsx` file extensions for running prettier.

packages/codemod/test/__snapshots__/transforms.spec.ts.snap
emmenko: This should do the trick: https://github.com/commercetools/merchant-center-application-kit/pull/2693/commits/3c0251ae5069dfb92da882ac9696a24b8ec1bec1

tdeekens (approved):

PR #2695, opened 2022-07-13
chore: update to graphql `v16`

emmenko:


codegen.core.yml
```diff
@@ -1,5 +1,5 @@
 schema:
-  - '${MC_GRAPHQL_API_URL}/graphql':
```
emmenko, line 2: Some leftovers

tdeekens (approved):
Looks good. Let's see how it integrates internally.

PR #2701, opened 2022-07-14
refactor(config): simplify schema and validate entry point separately

emmenko:
Something I noticed some time ago. The current JSON schema contains some validation regex rules for some of the fields, for example the entry point, but also to allow using variable placeholders.

This is not really working as one would expect. Therefore, I think it's simpler to keep the JSON schema "simple" (only requiring a `string` value) and perform validation as we parse and process the config (as we already do for the most part).


packages/application-config/schema.json
```diff
@@ -27,26 +23,11 @@
...
     "cloudIdentifier": {
       "description": "See https://docs.commercetools.com/custom-applications/api-reference/application-config#cloudidentifier",
-      "oneOf": [
-        {
-          "enum": ["gcp-au", "gcp-eu", "gcp-us", "aws-fra", "aws-ohio"]
-        },
-        {
-          "$ref": "#/definitions/envVariablePlaceholder"
-        }
-      ]
```
CarlosCortizasCT, line 49: `entryPointUriPath` validation was extracted to the `validations` file but it seems we don't validate `cloudIdentifier` value anymore. Shouldn't we also extract this validation?

emmenko: Actually we already kind of validated it here, when we process the value: https://github.com/commercetools/merchant-center-application-kit/blob/66c2067cf9a30ce624fe8f98c9a616c185115498/packages/application-config/src/utils.ts#L5-L28

CarlosCortizasCT: Ah! Ok 👍 
It seems the comment is now outdated 😅

CarlosCortizasCT (approved):

kark (approved):

Rhotimee (approved):

ddouglasz (approved):

PR #2710, opened 2022-07-14
chore: update PostCSS and Stylelint dependencies

emmenko:
Extracted from #2623


packages/jest-stylelint-runner/__mocks__/css-functions-list.js
```diff
@@ -0,0 +1,15 @@
...
+// of this package `css-functions-list` relies on some properties in
+// the `document` object that are not available.
+// Therefore, to keep things simple, we just mock the module, which
+// returns the location of a JSON file.
+const cssFunctionsListJsonFilePath = path.join(
+  path.dirname(require.resolve('css-functions-list')),
+  'index.json'
+);
+
+const location = cssFunctionsListJsonFilePath;
```
emmenko, line 13: Kind of "ugly" but it's the "simplest" solution I can think of.

The original error is: `The URL must be of scheme file`.

If we look at the `css-functions-list` package, we can see how it's implemented:

image

So wait, that's it? Where is this error coming from?

Well, the screenshot shows the ESM version, if we look at the CJS version we see this:

image

Notice how it's checking for the existence of the global object `document`.

Since the tests in appkit run in `jsdom` environment, this `document` object is defined but with very limited properties.
In fact, the `document.currentScript` and `document.baseURI` properties are not defined, which is what causes the `The URL must be of scheme file` error.

Therefore, let's just mock this module.

On another note, we might want to look into our testing setup in appkit to maybe have 2 group of tests: one for `jsdom` environments and another one for `node` environments, as we do have some packages that are Node-only packages.

Rhotimee (approved):

ddouglasz (approved):

PR #2711, opened 2022-07-14
chore: update Vite to `v3`

emmenko:
Vite v3 came out this week: https://vitejs.dev/blog/announcing-vite3.html

There is one feature that we were missing in v2, which is about configuring the public/assets path dynamically on runtime.

In Webpack we already do that using `__webpack_public_path__`

https://github.com/commercetools/merchant-center-application-kit/blob/main/packages/mc-scripts/src/application-runtime.ts

However, in Vite v2 it wasn't really well supported so that using the `cdnUrl` field in the Custom Application config wouldn't really work.

Now in Vite v3 they released better support for achieving the same result: https://vitejs.dev/guide/build.html#advanced-base-options

So here we go. Since Vite support in Custom Applications is still experimental, we can simply upgrade (I'm pretty sure nobody is using this yet).


packages/mc-scripts/src/commands/build-vite.ts
```diff
@@ -26,13 +27,13 @@ async function run() {
...
   await build({
     root: paths.appRoot,
+    base: './', // <-- Important to allow configuring the runtime base path.
     define: {
       'process.env.DEBUG': JSON.stringify(false),
       'process.env.NODE_ENV': JSON.stringify('production'),
     },
     build: {
       outDir: 'public',
-      assetsDir: '.',
```
emmenko, line 35: Currently we have all public files (`index.html`, favicons, etc.) AND the assets files (JS / CSS chunks) as a flat list inside the `public` folder.

I don't see any specific benefit from doing that (it was kind of a back and forth with some previous changes) so I think at least for Vite we can stick with the default behavior, which is having an `assets` folder.

This would make it easier to move this `assets` folder to a CDN or something, while leaving the other "public" files inside the `public` root folder.

image

packages/mc-scripts/src/commands/start-vite.ts
```diff
@@ -1,78 +1,14 @@
...
-      const html = replaceHtmlPlaceholders(rawHtml, {
-        env: enhancedLocalEnv,
-        headers: compiledHeaders,
-      });
-      return html;
-    },
-  };
-};
+import pluginCustomApplication from '../vite-plugins/vite-plugin-custom-application';
+import pluginSvgr from '../vite-plugins/vite-plugin-svgr';
```
emmenko, line 11: These custom plugins are now extracted into a `vite-plugins` folder, similarly to the `webpack-plugins` folder.

visual-testing-app/src/application.tsx
```diff
@@ -8,8 +8,9 @@ type TVisualRouteSpec = {
   Component: ComponentType;
 };
 
-const visualRoutesModules = import.meta.globEager<TVisualRouteSpec>(
-  './components/**/*.visualroute.tsx'
+const visualRoutesModules = import.meta.glob<TVisualRouteSpec>(
+  './components/**/*.visualroute.tsx',
+  { eager: true }
```
tdeekens, line 13: Interesting. Had too look this option up.

packages/mc-scripts/src/vite-plugins/vite-plugin-svgr.ts
```diff
@@ -0,0 +1,57 @@
+/**
+ * COPIED FROM https://github.com/pd4d10/vite-plugin-svgr
```
CarlosCortizasCT, line 2: Just curious about why we copied the file instead of importing the dependency into the project. Is it because we don't want to maintain another dependency when it only contains one file? would we need to manually check for new versions?

CarlosCortizasCT (approved):

tdeekens (approved):

packages/mc-scripts/src/vite-plugins/vite-plugin-svgr.ts
emmenko: Yes so mostly I wanted to make sure the plugin implementation was compatible with Vite 3. Since the implementation is pretty simple, and we only need specific functionality of it, I thought to just copy it over and remove some of the unused parts.

We can always switch back to using the package if we want to.

ddouglasz (approved):

PR #2713, opened 2022-07-14
chore: Cypress `v10`

emmenko:
Announcement: https://www.cypress.io/blog/2022/06/01/cypress-10-release/
Migration guide: https://docs.cypress.io/guides/references/migration-guide#Migrating-to-Cypress-version-10-0

Note: Cypress v10 adds (experimental) support for Component testing. It might be something interesting to look into as a follow up.


cypress.config.ts
```diff
@@ -0,0 +1,35 @@
...
+export default defineConfig({
+  e2e: {
+    baseUrl: 'http://localhost:3001',
+    retries: 1,
+    supportFile: 'cypress/support/e2e.ts',
+    async setupNodeEvents(on, cypressConfig) {
+      // Load the config
+      if (!process.env.CI) {
+        const envPath = path.join(__dirname, 'cypress/.env');
+        console.log('Loading environment variables from', envPath);
```
Rhotimee, line 13: I think you forgot the console.log

Rhotimee (approved):
🙌🏽

cypress.config.ts
```diff
@@ -0,0 +1,35 @@
+import path from 'path';
+import { defineConfig } from 'cypress';
+
+export default defineConfig({
+  e2e: {
+    baseUrl: 'http://localhost:3001',
+    retries: 1,
+    supportFile: 'cypress/support/e2e.ts',
```
CarlosCortizasCT, line 8: e2e are failing because not tests are found.
It seems that's because the default pattern for finding test files is this: `cypress/e2e/**/*.cy.{js,jsx,ts,tsx}` ([reference](https://docs.cypress.io/guides/references/configuration#Testing-Type-Specific-Options)).

We could add something like this:
```suggestion
    supportFile: 'cypress/support/e2e.ts',
    specPattern: 'cypress/integration/**/*.{js,jsx,ts,tsx}',
```

emmenko: Thanks!

Actually let me push a commit to restructure the files as well, following the structure and naming conventions.

emmenko: Restructured things here: https://github.com/commercetools/merchant-center-application-kit/pull/2713/commits/a8b9f9672f617540898a295778cc266d25344a07

This is also nicer as we get proper icons in VSCode:
image

emmenko:
Superseeded by #2766

PR #2723, opened 2022-07-22
refactor(app-shell): better reuse of `getMcApiUrl` util

emmenko:
The `getMcApiUrl` function was kind of duplicated across several packages. Now is defined and exported from the `application-shell-connectors` package.

The function itself has been refactored a bit accordingly, but the functionality remains the same.

In the `sdk` package we also have it defined, but I would keep it duplicated there for now, as otherwise it gets a bit tricky to share it between these 3 packages.


packages/application-shell-connectors/src/utils/get-mc-api-url/get-mc-api-url.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+  ${'true'}     | ${'https://mc-12345.us-central1.gcp.commercetools.com'}                                        | ${'https://mc-api.europe-west1.gcp.commercetools.com'} | ${'https://mc-api.us-central1.gcp.commercetools.com'}
+  ${'true'}     | ${'https://123.mc-preview.europe-west1.gcp.commercetools.com'}                                 | ${'https://mc-api.europe-west1.gcp.commercetools.com'} | ${'https://mc-api.europe-west1.gcp.commercetools.com'}
+  ${'true'}     | ${'https://123.mc-preview.us-central1.gcp.commercetools.com'}                                  | ${'https://mc-api.europe-west1.gcp.commercetools.com'} | ${'https://mc-api.us-central1.gcp.commercetools.com'}
+  ${'true'}     | ${'https://test.mc-preview.us-central1.gcp.commercetools.com'}                                 | ${'https://mc-api.europe-west1.gcp.commercetools.com'} | ${'https://mc-api.us-central1.gcp.commercetools.com'}
+  ${'true'}     | ${'https://this-is-not-what-you-are-looking-for.mc-preview.us-central1.gcp.commercetools.com'} | ${'https://mc-api.europe-west1.gcp.commercetools.com'} | ${'https://mc-api.us-central1.gcp.commercetools.com'}
```
emmenko, line 15: The tests have been rewritten using `describe.each` and should cover more cases:

image

packages/application-shell-connectors/src/utils/get-mc-api-url/get-mc-api-url.ts
```diff
@@ -1,45 +1,41 @@
...
-type TApplicationContextEnvironment = ApplicationWindow['app'];
-type TPartialWindow = Pick<Window, 'origin'>;
-
-const getMcApiUrlFromOrigin = (partialWindow: TPartialWindow) => {
-  const url = new URL(partialWindow.origin);
-  const originTld = removeMcPrefix(url.host);
+declare let window: ApplicationWindow;
+
+const mcHostnameRegex = /^mc(-(\d){4,})?\.(.*)$/;
+const mcPreviewHostnameRegex = /^.*\.mc-preview\.(.*)$/;
```
emmenko, line 6: This is new and we need it to test some new functionality.

packages/application-shell/src/utils/get-mc-api-url/get-mc-api-url.ts
```diff
@@ -1,55 +1,45 @@
-export interface ApplicationWindow extends Window {
-  app: {
-    servedByProxy: string | boolean;
-    mcApiUrl: string;
-  };
-}
+// NOTE: Copied from `@application-shell-connectors`.
+// TODO: Remove this version once the change in the connectors has been released.
```
emmenko, line 2: So the reason for this is that is this:

image

It seems that when we build/package the packages for attempting to install a starter template in isolation, there are some issues with cross-dependencies, maybe related to how packages are linked together.

To be honest, this is surprising and it means that either that particular setup is somehow wrong or it's not really possible to do that within a monorepo. I don't know at this point how to tackle that.
Definitely something to investigate but in the meantime I only have this solution.

PS: as a follow up, after merging/releasing this, we can be remove this again and the issue should be gone.

CarlosCortizasCT (approved):
💯

PR #2733, opened 2022-07-28
refactor: add basic support for `apollo-link-rest` and provide example implementation

emmenko:
I've seen a couple of apps using the [apollo-link-rest](https://www.apollographql.com/docs/react/api/link/apollo-link-rest) to use Apollo but for REST requests via the `@rest` directive.

However, most of our Apollo configuration is managed internally and not exposed to, for example, let users create their own Apollo client. It's also important that the Apollo requests are properly configured, for example to include the necessary HTTP headers.

To making it easier to use the `apollo-link-rest`, we can allow users to pass this link as an option to [createApolloClient](https://docs.commercetools.com/custom-applications/api-reference/commercetools-frontend-application-shell#createapolloclient).

The main reason for being so specific is that this link (and the other internal links) must be defined in a specific order.

In the playground app, there is a reference implementation using the `@rest` directive.

As a follow up, I'll update the documentation to describe how to set it up.

### DIFF
--- a/packages/application-shell/src/apollo-links/header-link.ts
+++ b/packages/application-shell/src/apollo-links/header-link.ts
@@ -17,6 +17,10 @@ import * as oidcStorage from '../utils/oidc-storage';
 
 declare let window: ApplicationWindow;
 
+type ApolloContextAfterResponse = TApolloContext & {
+  response?: Response;
+  restResponses?: Response[];
+};
 type QueryVariables = {
   /**
    * @deprecated: Use `{ context: { target } }`
@@ -60,6 +64,30 @@ const getAppliedForwardToHeaders = (
   }
 
   return {};
+};
+
+const extractSessionTokenFromResponse = (
+  context: ApolloContextAfterResponse
+): string | null => {
+  const refreshedSessionToken = context.response?.headers?.get(
+    'x-refreshed-session-token'
+  );
+  if (refreshedSessionToken) {
+    return refreshedSessionToken ?? null;
+  }
+
+  const restResponseWithRefreshTokenHeader = context.restResponses?.find(
+    (response) => response.headers?.has('x-refreshed-session-token')
+  );
+  if (restResponseWithRefreshTokenHeader) {
+    return (
+      restResponseWithRefreshTokenHeader.headers.get(
+        'x-refreshed-session-token'
+      ) ?? null
+    );
+  }
+
+  return null;
 };
 
 /* eslint-disable import/prefer-default-export */
@@ -119,11 +147,10 @@ const headerLink = new ApolloLink((operation, forward) => {
     }),
   });
   return forward(operation).map((response) => {
-    const context = operation.getContext();
+    const context = operation.getContext() as ApolloContextAfterResponse;
 
-    const refreshedSessionToken = context.response?.headers?.get(
-      'x-refreshed-session-token'
-    );
+    const refreshedSessionToken = extractSessionTokenFromResponse(context);
+
     if (refreshedSessionToken) {
       oidcStorage.setActiveSession(refreshedSessionToken);
     }
--- a/playground/src/apollo-client.js
+++ b/playground/src/apollo-client.js
@@ -1,0 +1,58 @@
+import omit from 'lodash/omit';
+import { RestLink } from 'apollo-link-rest';
+import { createApolloClient } from '@commercetools-frontend/application-shell';
+import { transformLocalizedStringToLocalizedField } from '@commercetools-frontend/l10n';
+
+// TODO: move to `l10n` package.
+const applyTransformedLocalizedStrings = (
+  objectWithLocalizedStrings,
+  fieldNames
+) => {
+  const transformedFieldDefinitions = fieldNames.reduce(
+    (nextTransformed, fieldName) => ({
+      ...nextTransformed,
+      [fieldName.to]: transformLocalizedStringToLocalizedField(
+        objectWithLocalizedStrings[fieldName.from]
+      ),
+    }),
+    {}
+  );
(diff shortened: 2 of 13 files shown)


packages/application-shell/src/apollo-links/header-link.ts
```diff
@@ -19,8 +17,10 @@ import * as oidcStorage from '../utils/oidc-storage';
 
 declare let window: ApplicationWindow;
 
-type ApolloContextWithInMemoryCache = TApolloContext & {
-  cache: ApolloClient<NormalizedCacheObject>;
```
emmenko, line 23: This wasn't really used

packages/application-shell/src/apollo-links/header-link.ts
```diff
@@ -19,8 +17,10 @@ import * as oidcStorage from '../utils/oidc-storage';
...
+type ApolloContextWithInMemoryCache = TApolloContext;
+type ApolloContextAfterResponse = TApolloContext & {
+  response?: Response;
+  restResponses?: Response[];
```
emmenko, line 23: To help handling the response

packages/application-shell/src/apollo-links/header-link.ts
```diff
@@ -67,6 +67,30 @@ const getAppliedForwardToHeaders = (
   return {};
 };
 
+const extractSessionTokenFromResponse = (
```
emmenko, line 70: Extracted some logic into this function.

playground/src/apollo-client.js
```diff
@@ -0,0 +1,58 @@
...
+    namesToOmit
+  );
+  return {
+    ...objectWithouLocalizedFields,
+    ...transformedFieldDefinitions,
+  };
+};
+
+const restLink = new RestLink({
+  uri: window.app.mcApiUrl,
```
CarlosCortizasCT, line 32: nit: maybe we could receive the URI as a parameter of the `configureApollo` function so this module don't need to rely on some global property, but maybe this is just a personal preference.

packages/application-shell/src/apollo-links/header-link.ts
```diff
@@ -19,8 +17,10 @@ import * as oidcStorage from '../utils/oidc-storage';
 
 declare let window: ApplicationWindow;
 
-type ApolloContextWithInMemoryCache = TApolloContext & {
-  cache: ApolloClient<NormalizedCacheObject>;
+type ApolloContextWithInMemoryCache = TApolloContext;
```
CarlosCortizasCT, line 20: Just wanted to double check here because the new type is named as it has some kind of cache property but we're not extending the `TApolloContext` withe the `cache` property anymore and [it doesn't seem to have it](https://github.com/commercetools/merchant-center-application-kit/blob/main/packages/application-shell/src/utils/apollo-context.ts#L11) by itself.

CarlosCortizasCT (approved):
Looks really good to me 👍 
(great to learn about the `apollo-link-rest`).

I left a couple of minor comments.

playground/src/apollo-client.js
emmenko: Yes I agree, I'm also not super happy about this.

In fact, there is another PR #2723 where the function `getMcApiUrl` is exported and can be used for these cases. I bumped into other issues there though...let's see.

> maybe we could receive the URI as a parameter of the `configureApollo` function

Could you provide an example of what you had in mind?

packages/application-shell/src/apollo-links/header-link.ts
emmenko: Good catch, I'll remove this type yes.

playground/src/apollo-client.js
CarlosCortizasCT: Sure!
Since this module is used from the `entry-point` (and we already need to access global state over there) I would expect to provide the value from there.

Here we could have something like this:
```
const createRestLink = (uri: string) => new RestLink({
  uri,
  ...
});

const configureApollo = ({ restLinkUri }: { restLinkUri: string }) => createApolloClient({
  restLink: createRestLink(restLinkUri)
});

export default configureApollo;
```

emmenko: Thanks for the example. However, this is up to the user to do it like this in their app, it's not something we need to change/expose from within appkit...right?

CarlosCortizasCT: Of course. 
I added this nitpick just as it seemed a best practice for me to have in our playground, but not needed at all.

emmenko:
@Rhotimee @ddouglasz Anything else to add?

Rhotimee:
> Anything else to add?

Currently reviewing.

packages/application-shell/src/apollo-links/header-link.ts
```diff
@@ -62,6 +66,30 @@ const getAppliedForwardToHeaders = (
...
 };
 
+const extractSessionTokenFromResponse = (
+  context: ApolloContextAfterResponse
+): string | null => {
+  const refreshedSessionToken = context.response?.headers?.get(
+    'x-refreshed-session-token'
+  );
+  if (refreshedSessionToken) {
+    return refreshedSessionToken ?? null;
```
Rhotimee, line 76: It is not clear to me why we are using the nullish coalescing operator since the if statement above already checked if `refreshedSessionToken` exist.

Rhotimee (approved):
Looks great.

packages/application-shell/src/apollo-links/header-link.ts
emmenko: Yes simply because the value could be `undefined`, but the function explicitly expects to return `null`.

PR #2734, opened 2022-07-29
feat(app-shell): new utility functions to configure custom HTTP clients

### LINKED ISSUE #2553: Document using a separate HTTP client
**Which part of the documentation does it affect and which topic does it refer to?**

Custom Applications are preconfigured to use Apollo (GraphQL client) and the JS SDK (HTTP client) for fetching data. This has the advantage of sending HTTP requests without having to worry about how to configure the requests.

However, developers can use any other HTTP client if they want to. There are no restrictions in place but it's up to the developers to configure the requests correctly.

**Describe what needs to be improved**

Document what things must be configured (e.g. HTTP headers) when using your own HTTP client.

We could also expose a utility function to preconfigure some of the HTTP headers.

emmenko:
Closes #2553 

- [x] Implement utility functions
- [x] Update Data fetching docs #2738

### DIFF
--- a/.changeset/tall-ties-jog.md
+++ b/.changeset/tall-ties-jog.md
@@ -28,19 +28,22 @@ const userAgent = createHttpUserAgent({
 });
 
 // Simple example using `fetch`.
-const data = await executeHttpClientRequest({ userAgent }, async (options) => {
-  const res = await fetch(buildApiUrl('/proxy/ctp/channels'), {
-    method: 'GET',
-    ...options,
-  });
-  const data = await res.json();
+const data = await executeHttpClientRequest(
+  async (options) => {
+    const res = await fetch(buildApiUrl('/proxy/ctp/channels'), {
+      method: 'GET',
+      ...options,
+    });
+    const data = await res.json();
 
-  return {
-    data,
-    statusCode: res.status,
-    getHeader: (key) => res.headers.get(key),
-  };
-});
+    return {
+      data,
+      statusCode: res.status,
+      getHeader: (key) => res.headers.get(key),
+    };
+  },
+  { userAgent }
+);
 ```
 
 The implementation example includes all the recommended functionalities such as:
--- a/.changeset/tall-ties-jog.md
+++ b/.changeset/tall-ties-jog.md
@@ -1,0 +1,54 @@
+---
+'@commercetools-frontend/application-shell': minor
+---
+
+New utility functions to configure HTTP requests when using custom HTTP clients.
+
+By default Custom Applications provide [preconfigured HTTP clients](https://docs.commercetools.com/custom-applications/development/data-fetching) for GraphQL and REST API requests.
+
+However, you could use any other HTTP client of your choice, for example `fetch`, `axios`, `swr`, etc.
+
+The main problem with that is the fact that you need to configure these clients on your own, in particular regarding the HTTP headers that should be sent with every request.
+
+This is now possible using some dedicated utility functions. See example usage below:
+
+```js
+import createHttpUserAgent from '@commercetools/http-user-agent';
+import {
+  buildApiUrl,
+  createHttpClientOptions,
+  executeHttpClientRequest,
+} from '@commercetools-frontend/application-shell';
+
+const userAgent = createHttpUserAgent({
+  name: 'fetch-client',
+  version: '2.6.0',
+  libraryName: window.app.applicationName,
+  contactEmail: 'support@my-company.com',
+});
+
+// Simple example using `fetch`.
+const data = await executeHttpClientRequest(async (headers) => {
+  const res = await fetch(buildApiUrl('/proxy/ctp/channels'), {
+    method: 'GET',
+    ...createHttpClientOptions({
+      headers,
+      userAgent,
+    }),
+  });
+  const data = await res.json();
+
+  return {
+    data,
+    statusCode: res.status,
+    getHeader: (key) => res.headers.get(key),
+  };
+});
+```
+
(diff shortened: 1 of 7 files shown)


.changeset/tall-ties-jog.md
```diff
@@ -0,0 +1,54 @@
...
+
+const userAgent = createHttpUserAgent({
+  name: 'fetch-client',
+  version: '2.6.0',
+  libraryName: window.app.applicationName,
+  contactEmail: 'support@my-company.com',
+});
+
+// Simple example using `fetch`.
+const data = await executeHttpClientRequest(async (headers) => {
```
CarlosCortizasCT, line 31: Personally, I found a little difficult to understand this `headers` object we receive here and it's needed to be passed when creating the http client options we provide to `fetch` function.

After reading the code, I understand is needed for the token renewal to work properly.
However, if we are going to provide a function to manage the whole renewal process, I think we could maybe hide that away from the consumers.
I'm thinking is something like this:
```
const data = await executeHttpClientRequest(userAgent, async (httpClientOptions) => {
  const res = await fetch(buildApiUrl('/proxy/ctp/channels'), {
    method: 'GET',
    ...httpClientOptions,
  });
  const data = await res.json();

  return {
    data,
    statusCode: res.status,
    getHeader: (key) => res.headers.get(key),
  };
});
```

This way, as a consumer I know the `executeHttpClientRequest` is providing me the options I should pass to `fetch` (or others) when sending the actual request.
If I need to add custom headers, I would expect something like this:
```
const data = await executeHttpClientRequest(userAgent, async (httpClientOptions) => {
  httpClientOptions.headers['X-Custom-Header'] = 'Custom-Value';

  const res = await fetch(buildApiUrl('/proxy/ctp/channels'), {
    method: 'GET',
    ...httpClientOptions,
  });
  const data = await res.json();

  return {
    data,
    statusCode: res.status,
    getHeader: (key) => res.headers.get(key),
  };
});
```

Perhaps we want to keep exporting the `createHttpClientOptions` for users if they want to use them in any other way but, if we do that, maybe we also would need to provide helpers to handle the session token renewal by exporting not only `oidc-storage.getSessionToken` but also a constant with the response header name `x-refreshed-session-token`.

The current proposal is also valid for me, but I wanted to share these thoughts.

emmenko: Yes the API of this function is not super simple. I'll take another look following your suggestions, thanks!

I think the important part is to "hide" the token renewal logic, as it would be weird to leave this up to the user. Keeping it an implementation details would allow us to manage changes to it internally without relying on users to migrate.

packages/application-shell/src/index.ts
```diff
@@ -21,6 +22,12 @@ export {
   useMcLazyQuery,
   useMcMutation,
 } from './hooks/apollo-hooks';
+export {
+  buildApiUrl,
+  createHttpClientOptions,
+  executeHttpClientRequest,
+} from './utils/http-client';
+export { getSessionToken } from './utils/oidc-storage';
```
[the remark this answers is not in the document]
emmenko: Ah this was a leftover, let me remove it again. Thanks for noticing.

.changeset/tall-ties-jog.md
emmenko: @CarlosCortizasCT How about this? https://github.com/commercetools/merchant-center-application-kit/pull/2734/commits/daacd10f7de4bcd973d1ddc8b45fb0b23509802c

CarlosCortizasCT: I find it easier to understand.
The only thing I think we perhaps could update is the order of `executeHttpClientRequest`: since the first parameter is now optional and the second is required, I would switch them:
```
// BEFORE
function executeHttpClientRequest<Data>(
  config: TConfig = {},
  fetcher: TFetcher<Data>
): Promise<Data>

// AFTER
function executeHttpClientRequest<Data>(
  fetcher: TFetcher<Data>,
  config: TConfig = {}
): Promise<Data> 
```

emmenko: Function overloading didn't help much. I changed now the order of the arguments as you suggested: https://github.com/commercetools/merchant-center-application-kit/pull/2734/commits/a8872e47e1819a489099c76f25d145e55448a491

Thanks!

packages/application-shell/src/utils/http-client.ts
```diff
@@ -0,0 +1,166 @@
...
+   */
+  data: Data;
+  /**
+   * The HTTP status code from the server response.
+   */
+  statusCode: number;
+  /**
+   * Implement a function to access the HTTP headers from the server response.
+   */
+  getHeader: (headerName: string) => string | null;
```
CarlosCortizasCT, line 76: nitpick: I was thinking maybe this property should be optional. The reason is that maybe users don't need to access response headers in their use cases and this way we would be asking them to always implement this function.

emmenko: This property should be required because is needed for this:

https://github.com/commercetools/merchant-center-application-kit/pull/2734/files#diff-2ea8fac16a771a4733de24f7b316e01175f44514f2baf4d4715f2443cfbef1c4R146-R149

The main point here is that HTTP clients might return different type of response objects and headers, so we can't reliably access the header.

Or we can try to "guess" but the user still needs to explicitly return the `headers` to the function.

Either way, something must be returned.

CarlosCortizasCT: Oh! I see.
Thanks for the explanation 👍

Rhotimee (approved):
🙌🏽

PR #2739, opened 2022-08-03
refactor: to add `includeSubDomains` directive to STS header

emmenko:
Just moving this from the proxy router config, to avoid duplication.

### DIFF
--- a/.changeset/many-balloons-appear.md
+++ b/.changeset/many-balloons-appear.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/constants': patch
+---
+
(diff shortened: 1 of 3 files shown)


tdeekens (approved):
Hm. Thought we had this before. Must have gotten lost.

emmenko:
> Hm. Thought we had this before. Must have gotten lost.

We have it internally and wanted to move it to appkit (but forgot to do so) 😅

image

CarlosCortizasCT (approved):

CarlosCortizasCT:
Just to double check: so we will have a follow-up PR in MC gateway to remove this header management?

emmenko:
Yes

Rhotimee (approved):

PR #2741, opened 2022-08-03
Remove japanese locale

CarlosCortizasCT:
## Summary

Remove Japanese locale as we're not supporting it anymore.

## Description

Since this locale is not being used, we would like to avoid confusion and reduce costs.


emmenko (approved):
💯

ddouglasz (commented):
👏

PR #2749, opened 2022-08-08
feat(l10n): add `applyTransformedLocalizedStrings` utility function

emmenko:
Similarly to the existing function `applyTransformedLocalizedFields`, we were missing the one for the opposite behavior.


CarlosCortizasCT (approved):
💯

kark (approved):

PR #2750, opened 2022-08-09
Add missing polyfill to `sdk` package

CarlosCortizasCT:
#### Summary

`sdk` package we use for *rest*-like requests is missing `buffer` polyfill needed by one of its dependencies.

#### Description

App-kit's `sdk` package relies on [`@commercetools/sdk-middleware-http`](https://github.com/commercetools/nodejs/tree/master/packages/sdk-middleware-http) which internally uses the `Buffer` object which is native to *nodejs* but does not exist in the browser so we need a polyfill for it.
This is a problem when sending requests with a body.

Also, there is an example use case in the _playground_ application which was not working due to:
* the problem with the `Buffer` polyfill
* no having the echo backend running when using the playground.

Having this example working is not super straightforward so I implemented a minimal proposal and document it in the *README*.


playground/README.md
```diff
@@ -8,3 +8,26 @@
 This is a Custom Application used to help developing features in this repository.
 
 Please read the [documentation](https://docs.commercetools.com/custom-applications) about Custom Applications before getting started.
+
+## Using the Echo Server view
```
emmenko, line 12: Can we maybe split the changes into two PRs? One for the fix with the polyfill, the other with the improvements to the playground.

playground/package.json
```diff
@@ -5,6 +5,7 @@
   "private": true,
   "scripts": {
     "build": "mc-scripts build",
+    "start:server": "NODE_ENV=development && ts-node ./src/server.ts",
```
emmenko, line 8: FYI: The playground uses the Vercel serverless functions (`./api/echo.ts`), so we can simply use [vercel dev](https://vercel.com/docs/cli#commands/dev) command.

Have you checked that to see if it works?

emmenko (commented):
Thanks for the fix with the polyfill! 👌

playground/package.json
CarlosCortizasCT: I checked but it seems we need to have a project defined in Vercel linked to the one you want to test locally.
When using that package it requires you to link or create a new project in Vercel.
If we can have a permanent project for this, we might use that package.

playground/README.md
CarlosCortizasCT: Ok. I used different commits to show both actions, but I will create a new PR for the playground and remove those changes from this one.

playground/package.json
emmenko: Ok, I double checked that and it seems the process now is a bit different than it was before.

Let's keep our custom script then, sorry about that.

emmenko:
Thanks for splitting. Remember to add a changeset.

CarlosCortizasCT:
> Thanks for splitting. Remember to add a changeset.

Added in this commit: 01fa06578e4544ab57bc616826f241670d452f67

playground/README.md
CarlosCortizasCT: Remove the playground part from this PR in this commit: 8ade2b4e51adba4bf2a0d4f207c19d46b5e3effa

emmenko (approved):
💯

PR #2752, opened 2022-08-09
Fix Playground echo server view

CarlosCortizasCT:
#### Summary

The Echo Server view of the playground application does not work properly when a user tries to send a request.

#### Description

The problem lies in that we don't have and echo backend running when using the playground.

Having this example working is not super straightforward so I implemented a minimal proposal and document it in the README.


playground/README.md
```diff
@@ -8,3 +8,26 @@
...
+
+You can use a [secure local tunnel](https://docs.commercetools.com/custom-applications/concepts/integrate-with-your-own-api#local-development-using-a-secure-tunnel) to help with this context. For instance, using **[ngrok](https://docs.commercetools.com/custom-applications/concepts/integrate-with-your-own-api#ngrok)**.
+The local server will listen in port `3003` by default (you can change it with the `PLAYGROUND_SERVER_PORT` environment variable). You will need to use it when starting the local tunnel.
+
+When using the tunnel, you will need to configure some environment variables with the domain assigned.
+Once you get the tunnel domain, head over the `.env.local` file (in the root directory) and make sure you use it in these variables:
+
+```
+PLAYGROUND_API_AUDIENCE="<tunnel_domain>"
+ECHO_SERVER_URL="<tunnel_domain>/api/echo"
```
emmenko, line 26: 💯

playground/src/server.ts
```diff
@@ -0,0 +1,19 @@
...
+// eslint-disable-next-line import/first
+import echoHandler from '../api/echo';
+
+const app = express();
+const port = process.env.PLAYGROUND_SERVER_PORT || 3003;
+
+// @ts-ignore
+app.get('/api/echo', echoHandler);
+// @ts-ignore
+app.post('/api/echo', echoHandler);
```
emmenko, line 13: Nit: simplify to use one handler?

```suggestion
app.use('/api/echo', echoHandler);
```

playground/src/server.ts
```diff
@@ -0,0 +1,19 @@
...
+
+app.listen(port, () => {
+  console.log(`Playground server listening on port ${port}`);
+});
+
+module.exports = {};
```
emmenko, line 19: Is this needed? 🤔

emmenko (commented):
Thanks for improving the usage of the playground and documenting the echo server 🤗

playground/src/server.ts
CarlosCortizasCT: Updated in this commit: a8f87d4a9dd0034e404e958535494d8c564cb997

playground/src/server.ts
CarlosCortizasCT: Sorry, this is a leftover.

Updated in this commit: a8f87d4a9dd0034e404e958535494d8c564cb997

emmenko (approved):
Looks good, thanks for fixing and improving the local development setup! 🤗

kark (approved):
💯

PR #2754, opened 2022-08-10
ci: try aggregate git release

emmenko:
There is an upcoming feature in the changesets/action related to aggregating github releases: https://github.com/changesets/action/pull/193

You can check [one](https://github.com/changesets/action/pull/193#issuecomment-1209176559) of the [last](https://github.com/changesets/action/pull/193#issuecomment-1209707621) comments to see some real world examples.

I think we can give it a try in our next release. For us it should work fine because we're already using fixed versions, so there is one single version for all packages.
The result is having GitHub releases per version as opposed to one GitHub release for each package (as it is now).


.changeset/config.json
```diff
@@ -9,9 +9,11 @@
   "commit": false,
   "fixed": [
     [
+      "@commercetools-applications/*",
       "@commercetools-backend/*",
       "@commercetools-frontend/*",
-      "!@commercetools-frontend/codemod"
+      "@commercetools-local/*",
+      "@commercetools-website/*"
```
emmenko, line 16: This is to keep ALL packages linked together, even the private apps (templates, playground, etc). The result is that these packages will get bumped to the same version as the others (even though they won't be released to NPM).

.changeset/config.json
```diff
@@ -9,9 +9,11 @@
   "commit": false,
   "fixed": [
     [
+      "@commercetools-applications/*",
       "@commercetools-backend/*",
       "@commercetools-frontend/*",
-      "!@commercetools-frontend/codemod"
```
emmenko, line 14: I thought about this again and I think it's better to keep it in sync with the other versions. It would be easier also to document new codemods for specific releases.

.github/workflows/release.yml
```diff
@@ -56,13 +56,24 @@ jobs:
       - name: Building packages
         run: yarn build
 
-      - name: Create Release Pull Request or Publish to npm
+      - name: Storing release version for changeset
+        id: release_version
+        run: echo "::set-output name=value::$(./scripts/print_release_version.sh)"
+        shell: bash
+        env:
+          GITHUB_TOKEN: ${{ secrets.RELEASE_GITHUB_TOKEN }}
```
emmenko, line 64: We need to "extract" the "next" release version before the `changesets/action` step, as the value is used for the `githubReleaseName` and `githubTagName` config values.

.github/workflows/release.yml
```diff
@@ -56,13 +56,24 @@ jobs:
...
-        uses: changesets/action@v1.3.0
+        # uses: changesets/action@v1.3.0
+        uses: dotansimha/changesets-action@1bd10e16a393db098c8f537c719621962c0d5571
         with:
           publish: yarn changeset publish
           version: yarn changeset:version-and-format
           commit: 'ci(changesets): version packages'
+          createGithubReleases: aggregate
+          githubReleaseName: v${{ steps.release_version.outputs.value }}
+          githubTagName: v${{ steps.release_version.outputs.value }}
```
emmenko, line 76: This is the new additional config

.github/workflows/release.yml
```diff
@@ -56,13 +56,24 @@ jobs:
...
+        run: echo "::set-output name=value::$(./scripts/print_release_version.sh)"
+        shell: bash
+        env:
+          GITHUB_TOKEN: ${{ secrets.RELEASE_GITHUB_TOKEN }}
+
+      - name: Creating release pull request or publishing release to npm registry
         id: changesets
-        uses: changesets/action@v1.3.0
+        # uses: changesets/action@v1.3.0
+        uses: dotansimha/changesets-action@1bd10e16a393db098c8f537c719621962c0d5571
```
emmenko, line 69: Using this version, as it hasn't been officially released yet. This is a way also to help testing this feature and give feedback.

.changeset/config.json
CarlosCortizasCT, line 16: Sorry but I don't the inclusion of this packages here. Will this make them visible in the release changesets?

CarlosCortizasCT (approved):
looks very promising 👍

tdeekens (approved):
This is really nice. Actually been one of my biggest struggles with vhangesets: navigating non aggregated releases in large mono-repos. Thanks for keeping an eye on recent work!

.changeset/config.json
emmenko: They all already have changesets, as they are part of the monorepo workspaces. E.g. https://github.com/commercetools/merchant-center-application-kit/blob/main/website/CHANGELOG.md

This change is just to align the bumped versions as I mentioned here: https://github.com/commercetools/merchant-center-application-kit/pull/2754#discussion_r942769379
Nothing else changes from the current status quo.

PR #2756, opened 2022-08-11
docs: add standalone-prices to list of reserved words

### LINKED ISSUE #2753: Add `standalone-prices` to reserved for internal usage warning block
**Which part of the documentation does it affect and which topic does it refer to?**

`standalone-prices` as an `entryPointUriPath` will be used by an internal merchant center application. We should add `standalone-prices` as a reserved value to the list of internal usage warning block.

https://github.com/commercetools/merchant-center-application-kit/blob/main/website/src/content/api-reference/application-config.mdx#entrypointuripath

kark:
#### Summary

Closes #2753

### DIFF
--- a/website/src/content/api-reference/application-config.mdx
+++ b/website/src/content/api-reference/application-config.mdx
@@ -160,8 +160,9 @@ Some values are reserved for internal usage and cannot be used:
 * `settings`
 * `welcome`
 * `disabled`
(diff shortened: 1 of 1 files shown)


dogayuksel (approved):
💯

CarlosCortizasCT (approved):

PR #2758, opened 2022-08-12
Human readable page title

### LINKED ISSUE #1830: Generate document.title based on location pathname

At the moment, each page of the Merchant Center shows the `Merchant Center` title. This is not very good for a number of reasons:

- With many tabs open, it's hard to find the correct one, as there is no page context information available.
- Using the browser history navigation is pretty much useless:

image

Each page should reflect its context in the `document.title` so that users can understand where they are. For that we can derive the information from the URL, as the default behavior. Users can still decide to provide a more precise value by passing a custom render function.

ddouglasz:
Closes #1830 

Page titles should display a human readable title with respect to the user navigation.

### DIFF
--- a/packages/application-shell/src/components/application-page-title/application-page-title.spec.tsx
+++ b/packages/application-shell/src/components/application-page-title/application-page-title.spec.tsx
@@ -8,50 +8,35 @@ const memoryHistory = (route: string) =>
   createMemoryHistory({ initialEntries: [route] });
 
 describe.each`
-  route              | title
-  ${`/login`}        | ${`Login - Merchant Center`}
-  ${`/login/new`}    | ${`Login - Merchant Center`}
-  ${`/login/sso`}    | ${`Login - Merchant Center`}
-  ${`/login/forgot`} | ${`Login - Merchant Center`}
-  ${`/login/choose`} | ${`Login - Merchant Center`}
+  route                                                                         | title
+  ${`/almond-40/dashboard`}                                                     | ${`Dashboard - almond-40 - Merchant Center`}
+  ${`/almond-40/products`}                                                      | ${`Products - almond-40 - Merchant Center`}
+  ${`/almond-40/products/new`}                                                  | ${`Products - almond-40 - Merchant Center`}
+  ${`/almond-40/products/<id>`}                                                 | ${`Products - almond-40 - Merchant Center`}
+  ${`/almond-40/products/<id>/variants`}                                        | ${`Products - almond-40 - Merchant Center`}
+  ${`/almond-40/products/<id>/variants/<id>`}                                   | ${`Products - almond-40 - Merchant Center`}
+  ${`/almond-40/products/<id>/variants/<id>/prices`}                            | ${`Products - almond-40 - Merchant Center`}
+  ${`/almond-40/orders/<id>`}                                                   | ${`Orders - almond-40 - Merchant Center`}
+  ${`/almond-40/settings/project/international`}                                | ${`Settings - almond-40 - Merchant Center`}
+  ${`/almond-40/settings/project/stores`}                                       | ${`Settings - almond-40 - Merchant Center`}
+  ${`/almond-40/settings/project/stores/<id>`}                                  | ${`Settings - almond-40 - Merchant Center`}
+  ${`/almond-40/settings/product-types`}                                        | ${`Settings - almond-40 - Merchant Center`}
+  ${`/almond-40/settings/product-types/<id>`}                                   | ${`Settings - almond-40 - Merchant Center`}
+  ${`/almond-40/settings/product-types/<id>/attribute-definitions/detail/<id>`} | ${`Settings - almond-40 - Merchant Center`}
+  ${`/account/projects`}                                                        | ${`Account - Merchant Center`}
+  ${`/account/organizations`}                                                   | ${`Account - Merchant Center`}
+  ${`/account/organizations/<id>/teams`}                                        | ${`Account - Merchant Center`}
+  ${`/account/organizations/<id>/teams/<id>/permissions/<id>`}                  | ${`Account - Merchant Center`}
+  ${`/account/organizations/<id>/custom-applications`}                          | ${`Account - Merchant Center`}
+  ${`/account/organizations/<id>/custom-applications/owned/<id>`}               | ${`Account - Merchant Center`}
+  ${`/account/organizations/<id>/custom-applications/installations/<id>`}       | ${`Account - Merchant Center`}
+  ${`/login`}                                                                   | ${`Login - Merchant Center`}
+  ${`/login/new`}                                                               | ${`Login - Merchant Center`}
+  ${`/login/sso`}                                                               | ${`Login - Merchant Center`}
+  ${`/login/forgot`}                                                            | ${`Login - Merchant Center`}
+  ${`/login/choose`}                                                            | ${`Login - Merchant Center`}
 `(`when location is "$route"`, ({ route, title }) => {
   it('should render default page title', async () => {
-    render(
-      <Router history={memoryHistory(route)}>
-        <ApplicationPageTitle />
-      </Router>
-    );
-    await waitFor(() => {
-      expect(document.title).toBe(title);
-    });
-  });
-});
-
-describe.each`
-  route                                                                         | title
-  ${`/almond-40/dashboard`}                                                     | ${`Dashboard - ... chant Center`}
-  ${`/almond-40/products`}                                                      | ${`Products - a ... chant Center`}
-  ${`/almond-40/products/new`}                                                  | ${`Products - a ... chant Center`}
-  ${`/almond-40/products/<id>`}                                                 | ${`Products - a ... chant Center`}
-  ${`/almond-40/products/<id>/variants`}                                        | ${`Products - a ... chant Center`}
-  ${`/almond-40/products/<id>/variants/<id>`}                                   | ${`Products - a ... chant Center`}
-  ${`/almond-40/products/<id>/variants/<id>/prices`}                            | ${`Products - a ... chant Center`}
-  ${`/almond-40/orders/<id>`}                                                   | ${`Orders - alm ... chant Center`}
-  ${`/almond-40/settings/project/international`}                                | ${`Settings - a ... chant Center`}
-  ${`/almond-40/settings/project/stores`}                                       | ${`Settings - a ... chant Center`}
-  ${`/almond-40/settings/project/stores/<id>`}                                  | ${`Settings - a ... chant Center`}
-  ${`/almond-40/settings/product-types`}                                        | ${`Settings - a ... chant Center`}
-  ${`/almond-40/settings/product-types/<id>`}                                   | ${`Settings - a ... chant Center`}
-  ${`/almond-40/settings/product-types/<id>/attribute-definitions/detail/<id>`} | ${`Settings - a ... chant Center`}
-  ${`/account/projects`}                                                        | ${`Account - M ... chant Center`}
-  ${`/account/organizations`}                                                   | ${`Account - M ... chant Center`}
-  ${`/account/organizations/<id>/teams`}                                        | ${`Account - M ... chant Center`}
-  ${`/account/organizations/<id>/teams/<id>/permissions/<id>`}                  | ${`Account - M ... chant Center`}
-  ${`/account/organizations/<id>/custom-applications`}                          | ${`Account - M ... chant Center`}
-  ${`/account/organizations/<id>/custom-applications/owned/<id>`}               | ${`Account - M ... chant Center`}
-  ${`/account/organizations/<id>/custom-applications/installations/<id>`}       | ${`Account - M ... chant Center`}
-`(`'when the page title is too long: "$route"`, ({ route, title }) => {
-  it('should render a truncated readable page title', async () => {
     render(
       <Router history={memoryHistory(route)}>
         <ApplicationPageTitle />
@@ -71,7 +56,30 @@ describe('when using the custom "renderPageTitle" function to override default m
       </Router>
     );
     await waitFor(() => {
-      expect(document.title).toBe('Create A New ... chant Center');
+      expect(document.title).toBe(
+        'Create A New Account - Login - Merchant Center'
+      );
+    });
+  });
+
+  describe('when the page title is too long', () => {
+    it('should render a truncated readable page title', async () => {
+      render(
+        <Router
+          history={memoryHistory(
+            '/almond-40/products/productid12345/variants/variantid12345'
+          )}
+        >
+          <ApplicationPageTitle
+            renderPageTitle={['Some product very long product name']}
+          />
+        </Router>
+      );
+      await waitFor(() => {
+        expect(document.title).toBe(
+          'Some Product ... Product Name - Products - almond-40 - Merchant Center'
+        );
+      });
     });
   });
 });
--- a/packages/application-shell/src/components/application-page-title/application-page-title.spec.tsx
+++ b/packages/application-shell/src/components/application-page-title/application-page-title.spec.tsx
@@ -49,11 +49,11 @@ describe.each`
 });
 
 describe.each`
-  content                                                                                                                  | title
-  ${[`Lemon shoes`]}                                                                                                       | ${`Lemon shoes - Products - my-project-key - Merchant Center`}
-  ${[`Some product, a very long product name`]}                                                                            | ${`Some product...product name - Products - my-project-key - Merchant Center`}
-  ${['short description1', 'short description2']}                                                                          | ${`short description1 - short description2 - Products - my-project-key - Merchant Center`}
-  ${['first description of product, a very long product name', 'second description of product, a very long product name']} | ${`first descri...product name - second descr...product name - Products - my-project-key - Merchant Center`}
+  content                                                                                                      | title
+  ${[`Short title`]}                                                                                           | ${`Short title - Products - my-project-key - Merchant Center`}
+  ${[`Short title, a very long product name`]}                                                                 | ${`Short title,...product name - Products - my-project-key - Merchant Center`}
+  ${['Short title1', 'Short title2']}                                                                          | ${`Short title1 - Short title2 - Products - my-project-key - Merchant Center`}
+  ${['First title of product, a very long product name', 'Second title of product, a very long product name']} | ${`First title ...product name - Second title...product name - Products - my-project-key - Merchant Center`}
 `(`'when the custom "content" is "$content"`, ({ content, title }) => {
   it('should render the custom value', async () => {
     render(
--- a/packages/application-shell/src/components/application-page-title/application-page-title.spec.tsx
+++ b/packages/application-shell/src/components/application-page-title/application-page-title.spec.tsx
@@ -1,75 +1,77 @@
-import { renderApp, waitFor } from '../../test-utils';
+import { waitFor } from '../../test-utils';
 import ApplicationPageTitle from './application-page-title';
+import { render } from '@testing-library/react';
+import { Router } from 'react-router';
+import { createMemoryHistory } from 'history';
+
+const memoryHistory = (route: string) =>
+  createMemoryHistory({ initialEntries: [route] });
 
 describe.each`
-  route                                                                         | title
-  ${`/almond-40/dashboard`}                                                     | ${`Dashboard - almond-40 - Merchant Center`}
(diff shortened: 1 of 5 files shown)


packages/application-shell/src/components/application-page-title/application-page-title.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+const ApplicationPageTitle = <Query extends {} = {}>(props: Props<Query>) => {
+  const location = useLocation();
+  const breadcrumb = compilePageTitle(location as TEnhancedLocation<Query>);
+  const pageTitle = props.renderPageTitle
+    ? props.renderPageTitle(breadcrumb)
+    : `${convertToReadablePrefix(breadcrumb.paths[0])} - ${breadcrumb.suffix}`;
+
+  useEffect(() => {
+    document.title = pageTitle;
```
ddouglasz, line 124: I am still going to look to see if there is an alternative to accessing the DOM directly, since that is not so advisable. 
We agreed to not use the helmet package as it seems not to be maintained for a while now.

packages/application-shell/src/components/application-page-title/application-page-title.tsx
```diff
@@ -0,0 +1,131 @@
...
+    case 'settings':
+      return 'Settings';
+    case 'projects':
+      return 'Projects';
+    case 'login':
+      return 'Login';
+    case 'account':
+      return 'Account';
+    default:
+      return startCase(prefix);
```
tdeekens, line 112: 👋🏼 Thanks for working on this. I am frankly new to the topic but am wondering if this is something we should allow configuration of via a Custom Application's config. Then we might not need this map here.

ddouglasz: Oh, thanks for pointing this out. I will look into it.

emmenko: No need to configure this, this mapping is not really necessary.

I would suggest to focus on this comment first: https://github.com/commercetools/merchant-center-application-kit/pull/2758#issuecomment-1216602555

Then things will fall nicely into place, following that path.

ddouglasz:
> @ddouglasz How can we ensure we're implementing this according to the requirements? 🤓

Ah, great question. As for the default mapping, I think I pretty much tested it to see that it matches and is consistent with all the possible requirements on  what we have in the documented requirements. From what is in the doc, we are either using the `page identifier`, `project key` and `Merchant Center`. 
The paths pattern may change sometimes so accessing some attributes of the paths might change for different locations eg `account`, and `login` as they do not seem to begin with a `project_key` for example. These are some of the cases I tried to cover too. Here default mapping uses  just `login` and `Merchant Center`, or `account` and `Merchant Center`.

emmenko:
> I think I pretty much tested it to see that it matches and is consistent with all the possible requirements on what we have in the documented requirements.

That's good yes, I just don't see that in the PR 😉

packages/application-shell/src/components/application-page-title/application-page-title.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+  it('should render a truncated readable page title', async () => {
+    render(
+      <Router history={memoryHistory(route)}>
+        <ApplicationPageTitle />
+      </Router>
+    );
+    await waitFor(() => {
+      expect(document.title).toBe(title);
+    });
+  });
```
CarlosCortizasCT, line 63: nit: perhaps we can extract this logic to a private helper function which can be used in all three tests ([here is an idea](https://gist.github.com/CarlosCortizasCT/ea4aab5614225e6becb19d45c1bb61b6))

packages/application-shell/src/components/application-page-title/application-page-title.tsx
```diff
@@ -0,0 +1,122 @@
...
+};
+
+const ApplicationPageTitle = <Query extends {} = {}>(props: Props) => {
+  const location = useLocation();
+  const breadcrumb = compilePageTitle(location as TEnhancedLocation<Query>);
+  const defaultMapping = `${startCase(breadcrumb.paths[0])} - ${
+    breadcrumb.suffix
+  }`;
+  const pageTitle = props.renderPageTitle
+    ? `${props.renderPageTitle
```
CarlosCortizasCT, line 100: i guess we would need to also test for an empty array, otherwise the page title will start with a slash:
```suggestion
    ? `${props.renderPageTitle && props.renderPageTitle.length > 0
```

packages/application-shell/src/components/application-page-title/application-page-title.tsx
```diff
@@ -0,0 +1,122 @@
...
+    : defaultMapping;
+
+  const truncatePageTitle = (pageTitle: string) => {
+    if (pageTitle.length > maxTitleCharLength) {
+      return `${pageTitle.slice(0, 12)} ... ${pageTitle.slice(
+        pageTitle.length - 12
+      )}`;
+    }
+    return pageTitle;
+  };
```
CarlosCortizasCT, line 112: This seems to be a pure function so we can move it outside the component so we don't create it on every render.

packages/application-shell/src/components/application-page-title/application-page-title.tsx
```diff
@@ -0,0 +1,122 @@
...
+};
+
+const ApplicationPageTitle = <Query extends {} = {}>(props: Props) => {
+  const location = useLocation();
+  const breadcrumb = compilePageTitle(location as TEnhancedLocation<Query>);
+  const defaultMapping = `${startCase(breadcrumb.paths[0])} - ${
+    breadcrumb.suffix
+  }`;
+  const pageTitle = props.renderPageTitle
```
CarlosCortizasCT, line 99: I'm wondering if we should move this code block inside the effect as it only needs to run when location or renderPageTitle changes:
```
  useLayoutEffect(() => {
    const breadcrumb = compilePageTitle(location as TEnhancedLocation<Query>);
    const defaultMapping = `${startCase(breadcrumb.paths[0])} - ${
      breadcrumb.suffix
    }`;
    const pageTitle = props.renderPageTitle && props.renderPageTitle.length > 0
      ? `${props.renderPageTitle
          .join(' - ')
          .replace(/\w+/g, capitalize)} - ${defaultMapping}`
      : defaultMapping;

    document.title = truncatePageTitle(pageTitle);
  }, [location, props.renderPageTitle]);
```

packages/application-shell/src/components/application-page-title/application-page-title.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+  const paths = isStaticPath(projectKeyOrStaticPath)
+    ? getLimitedPaths([projectKeyOrStaticPath])
+    : getLimitedPaths(
+        locationPaths.length === 0 ? [projectKeyOrStaticPath] : locationPaths
+      );
+  return {
+    suffix,
+    paths,
+    location,
```
CarlosCortizasCT, line 89: Here we are returning the same parameter we received so I guess we can remove it since the function caller would already have it.

CarlosCortizasCT (commented):
Great work over here! 💪 

I just left some comments with some suggestions.

packages/application-shell/src/components/application-page-title/application-page-title.tsx
ddouglasz: Thanks for all the feedback. 
updated here: https://github.com/commercetools/merchant-center-application-kit/pull/2758/commits/b422e3acf036da23e324d3dd1c701958ac9c7270

packages/application-shell/src/components/application-page-title/application-page-title.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+import { useLayoutEffect } from 'react';
+import { useLocation } from 'react-router-dom';
+import startCase from 'lodash/startCase';
+import capitalize from 'lodash/capitalize';
+
+type Breadcrumb<Query extends {} = {}> = {
+  suffix: string;
+  paths: string[];
+  location?: TEnhancedLocation<Query>;
```
CarlosCortizasCT, line 11: I'm sorry but I don't understand why the `Breadcrumb` type includes this optional attribute since it doesn't seem is used anymore.
Perhaps I'm missing something. Could you please clarify? 🙏

packages/application-shell/src/components/application-page-title/application-page-title.tsx
```diff
@@ -0,0 +1,127 @@
...
+};
+type Props = {
+  /**
+   * Allow to render a custom page title.
+   * Overrides default page title with the format:
+   * <?resource_name(manually inputed)> - <?page_location> - <application_identifier> - <project_key> - Merchant Center
```
CarlosCortizasCT, line 17: I find a little confusing this explanation.
Perhaps we could go for something like this:
```
Allows to extend the page title. 
When provided, the array items will be used as a prefix for the default title this component will generate.

It will follow a pattern like this:
<renderPageTitle.join(' - ')> - <?page_location> - <application_identifier> - <project_key> - Merchant Center
```

What do you think?

ddouglasz: Oh yes, this makes sense too. I just wanted to provide some clarity to how the titles would be formed from an additional content provided.

packages/application-shell/src/components/application-page-title/application-page-title.tsx
ddouglasz: Honestly, I think I missed it, it is not used anymore after removing it from the returned object 😅
Thank you

packages/application-shell/src/components/application-page-title/application-page-title.tsx
ddouglasz: updated here: https://github.com/commercetools/merchant-center-application-kit/pull/2758/commits/07686cec816448de631b4144c58f8c5cd3eddc28

packages/application-shell/src/components/application-page-title/application-page-title.tsx
```diff
@@ -0,0 +1,127 @@
...
+  suffix: string;
+  paths: string[];
+};
+type Props = {
+  /**
+   * Allow to render a custom page title.
+   * Overrides default page title with the format:
+   * <?resource_name(manually inputed)> - <?page_location> - <application_identifier> - <project_key> - Merchant Center
+   */
+  renderPageTitle?: string[];
```
emmenko, line 18: Nit: the name of this prop seems a bit confusing to me. In the RFC I proposed `content` but we can also find another name as long as it fits with what it does.

packages/application-shell/src/components/application-page-title/application-page-title.tsx
```diff
@@ -0,0 +1,127 @@
...
+        ...paths.slice(2),
+      ],
+      true
+    );
+  }
+  return compilePath(paths, hasTruncatedPaths);
+};
+
+const isStaticPath = (path: string) => {
+  const staticPaths = ['account', 'login'];
```
emmenko, line 58: Nit: Define this outside of the function, as it's a static variable.

packages/application-shell/src/components/application-page-title/application-page-title.tsx
```diff
@@ -0,0 +1,127 @@
...
+      return title;
+    })
+    .join(' - ');
+};
+
+const ApplicationPageTitle = <Query extends {} = {}>(props: Props) => {
+  const location = useLocation();
+
+  useLayoutEffect(() => {
+    const breadcrumb = compilePageTitle(location as TEnhancedLocation<Query>);
```
emmenko, line 110: I don't think we need `location` to be `TEnhancedLocation`, as we are only using the `pathname`. We also don't need the `Query` generic type.

packages/application-shell/src/components/application-page-title/application-page-title.tsx
```diff
@@ -0,0 +1,127 @@
...
+    const defaultMapping = `${startCase(breadcrumb.paths[0])} - ${
+      breadcrumb.suffix
+    }`;
+    const pageTitle =
+      props.renderPageTitle && props.renderPageTitle.length > 0
+        ? `${props.renderPageTitle
+            .join(' - ')
+            .replace(/\w+/g, capitalize)} - ${defaultMapping}`
+        : defaultMapping;
+    document.title = truncatePageTitle(pageTitle);
```
emmenko, line 120: This logic seems a bit confusing, also because we're constructing `pageTitle` to then deconstruct it again in `truncatePageTitle`.

I would suggest to work with an array of parts, map over it and then join the values once.

One other thing I noticed: we shouldn't capitalize every word, especially the ones that have been passed as prop. I think only the `entryPointUriPath` could be capitalized, but not the rest.

packages/application-shell/src/components/application-page-title/application-page-title.tsx
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * It will follow a pattern like this:
+ * <renderPageTitle.join(' - ')> - <?page_location> - <application_identifier> - <project_key> - Merchant Center
+ *
+ */
+
+const compilePageTitle = <Query extends {}>(
+  location: TEnhancedLocation<Query>
+): Breadcrumb => {
+  const [, projectKeyOrStaticPath, ...locationPaths] =
```
emmenko, line 74: Since we only look for the project key and the entry point, we only need to pick the first two:

```suggestion
  const [, projectKeyOrStaticPath, entryPointUriPath] =
```

image

packages/application-shell/src/components/application-page-title/application-page-title.tsx
```diff
@@ -0,0 +1,127 @@
...
+    const pageTitle =
+      props.renderPageTitle && props.renderPageTitle.length > 0
+        ? `${props.renderPageTitle
+            .join(' - ')
+            .replace(/\w+/g, capitalize)} - ${defaultMapping}`
+        : defaultMapping;
+    document.title = truncatePageTitle(pageTitle);
+  }, [location, props.renderPageTitle]);
+
+  return <></>;
```
emmenko, line 123:
```suggestion
  return null;
```

emmenko (commented):
Cool stuff, we're getting there 🙌

I did a quick review and left some feedback to some of the main points to focus on. Let me know if it's clear enough.

packages/application-shell/src/components/application-page-title/application-page-title.tsx
```diff
@@ -0,0 +1,127 @@
...
+    paths,
+  };
+};
+
+const truncatePageTitle = (pageTitle: string) => {
+  return pageTitle
+    .split(' - ')
+    .map((title) => {
+      if (title.length > maxTitleCharLength) {
+        return (title = `${title.slice(0, 12)} ... ${title.slice(
```
kark, line 97: nit: I guess according to our rfc it should be 24 + 3 chars,  without empty characters
`This is a very long title that exceeds 24 characters → This is a ve...s characters`
```suggestion
        return (title = `${title.slice(0, 12)}...${title.slice(
```

kark, line 97: nit: I guess according to our rfc it should be 24 + 3 chars,  without empty characters
`This is a very long title that exceeds 24 characters → This is a ve...s characters`

and we could also reference the `maxTitleCharLength` constant

```suggestion
        return (title = `${title.slice(0, maxTitleCharLength/2)}...${title.slice(
```

packages/application-shell/src/components/application-page-title/application-page-title.tsx
ddouglasz: Thank you, all feedback updated here: https://github.com/commercetools/merchant-center-application-kit/pull/2758/commits/e57096278223ab50a3baaffcb43433b5902c34b2

packages/application-shell/src/components/application-page-title/application-page-title.tsx
ddouglasz: Thank you Kacper 🙏:
Updated here: https://github.com/commercetools/merchant-center-application-kit/pull/2758/commits/e57096278223ab50a3baaffcb43433b5902c34b2

packages/application-shell/src/components/application-page-title/application-page-title.spec.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { render } from '@testing-library/react';
+import { Router } from 'react-router';
+import { createMemoryHistory } from 'history';
+
+const memoryHistory = (route: string) =>
+  createMemoryHistory({ initialEntries: [route] });
+
+describe.each`
+  route                                                                         | title
+  ${`/almond-40/dashboard`}                                                     | ${`Dashboard - almond-40 - Merchant Center`}
```
emmenko, line 12: ⚠️ Let's use a fake project key

packages/application-shell/src/components/application-page-title/application-page-title.spec.tsx
```diff
@@ -0,0 +1,85 @@
...
+          />
+        </Router>
+      );
+      await waitFor(() => {
+        expect(document.title).toBe(
+          'Some product...product name - Products - almond-40 - Merchant Center'
+        );
+      });
+    });
+  });
```
emmenko, line 84: Nit: We can also use the `describe.each` here and have the test data in the "table" format.

packages/application-shell/src/components/application-shell-provider/application-shell-provider.tsx
```diff
@@ -59,6 +60,13 @@ const ApplicationShellProvider = <AdditionalEnvironmentProperties extends {}>(
               <Suspense fallback={<ApplicationLoader />}>
                 <Router history={ApplicationShellProvider.history}>
                   <GtmBooter trackingEventList={props.trackingEventList || {}}>
+                    <ApplicationPageTitle
+                      content={[
+                        'A very long sentence that can be truncated',
+                        'Some red short text also kind of long',
+                      ]}
```
emmenko, line 67: Remember to remove this.

packages/application-shell/src/components/application-page-title/application-page-title.tsx
```diff
@@ -0,0 +1,134 @@
...
+            }
+            return title;
+          })
+          .join(' - ');
+      } else {
+        return `${startCase(breadcrumb.paths[0])} - ${breadcrumb.suffix}`;
+      }
+    };
+
+    document.title = truncatedPageTitle();
```
emmenko, line 127: I feel like we are overcomplicating things here.

I would expect something simple like:

```ts
const [, projectKeyOrStaticPath, entryPointUriPath] =
    location.pathname.split('/');

const customTitleParts = props.content.map((titlePart) => {
  if (titlePart.length <= 24) {
    return titlePart;
  }
  return [
    titlePart.slice(0, 12),
    titlePart.slice(titlePart.length - 12)
  ].join('...');
});

if (staticPaths.includes(projectKeyOrStaticPath)) {
  return [
    ...customTitleParts,
    upperFirst(projectKeyOrStaticPath),
    'Merchant Center'
  ].join('-');
}

return [
  ...customTitleParts,
  upperFirst(entryPointUriPath),
  projectKeyOrStaticPath,
  'Merchant Center'
].join('-');
```

emmenko (commented):
I left some suggestions to iterate on the implementation, as it's a bit overcomplicated at the moment.

Thanks

emmenko:
@ddouglasz I pushed a commit (https://github.com/commercetools/merchant-center-application-kit/pull/2758/commits/25406cf4ddcded3781d5cdf0d6a74de81a77b4f8) with some final improvements, mostly around the docs.

I also renamed the prop to `additionalParts`, as I felt `content` was not clear enough (sorry for suggesting that in the first place). I think this should make it more clear what it does now.

emmenko (approved):
This is good to go now from my side!

Thanks again for working on this 🤗

packages/application-shell/src/components/application-page-title/application-page-title.tsx
```diff
@@ -0,0 +1,76 @@
...
+   * <ApplicationPageTitle additionalParts={['T-Shirt red']} />
+   * // T-Shirt red - Products - my-shop - Merchant Center
+   */
+  additionalParts: string[];
+};
+
+const maxTitleCharLength = 24;
+const staticPaths = ['account', 'login'];
+
+const defaultProps: Pick<TApplicationPageTitleProps, 'additionalParts'> = {
```
CarlosCortizasCT, line 28: nit: perhaps we don't need to use `Pick` helper anymore since the type only has one prop.

CarlosCortizasCT (approved):
Nice job 💯

kark (approved):

PR #2766, opened 2022-08-17
Update Cypress to `v10`

CarlosCortizasCT:
#### Summary

Updates Cypress to [v10](https://docs.cypress.io/guides/references/migration-guide#Migrating-to-Cypress-version-10-0).

#### Description

Some changes are required in the test files location and naming. Also, some configuration files updates.
It's been conducted using the own Cypress migration tool.


cypress.config.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export default defineConfig({
+  retries: 1,
+  video: false,
+  e2e: {
+    // We've imported your old cypress plugins here.
+    // You may want to clean this up later by importing these.
+    setupNodeEvents(on, config) {
+      // eslint-disable-next-line import/extensions
+      return require('./cypress/plugins/index.js')(on, config);
```
emmenko, line 11: Nit: in my other PR I moved the logic inside this function, as it's also what's [recommended by Cypress](https://docs.cypress.io/guides/references/migration-guide#Plugins-File-Removed).

https://github.com/commercetools/merchant-center-application-kit/pull/2713/files#diff-3f62e885350a48939f240b9daf40f6b06d41da7751f2b613238fc7669ac0a1b2R8-R33

It's also easier to use `async/await`.

.changeset/cold-plums-attack.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-frontend/cypress': patch
```
emmenko, line 2: Nit: should this maybe be a minor bump? 🤔
I don't mind either way, just wondering...

```suggestion
'@commercetools-frontend/cypress': minor
```

emmenko (commented):
Ah thanks!! 🙌
What was that made it work now compared to #2713 ?

cypress.config.ts
CarlosCortizasCT: Thanks for pointing that out.

Updated in this commit: d0a488e8cc2ba865426d571c20b33fe0109a47ad

.changeset/cold-plums-attack.md
CarlosCortizasCT: Updated in this commit: 24f10f6f65da63aac5ae98daecd21707c6d02a73

CarlosCortizasCT:
> Ah thanks!! 🙌 What was that made it work now compared to #2713 ?

Sorry I didn't remember we had that PR already opened for updating Cypress.

The only difference I can tell is about updating `@percy/cli` dependency, which is something I'm not doing here. 
Updating that dependency is handled in [this another PR](https://github.com/commercetools/merchant-center-application-kit/pull/2764) since I thought it would be better to update both separately.
I was expecting that other PR to be approved and merged so I can then refresh this branch and test it altogether.

emmenko (approved):
Thanks, changes look good to me! 🙌

PR #2767, opened 2022-08-18
refactor(app-shell): avoid using deprecated fields

emmenko:
A leftover from the Org-level Custom Applications release.


packages/application-shell/src/components/navbar/fetch-project-extensions-navbar.settings.graphql
```diff
@@ -31,24 +31,24 @@ query FetchProjectExtensionsNavbar {
...
-            }
+        }
+        submenuLinks {
+          id
+          uriPath
+          permissions
+          defaultLabel
+          labelAllLocales {
+            locale
+            value
```
emmenko, line 51: This is the main change.

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -911,8 +912,8 @@ describe('navbar menu links interactions', () => {
                     }),
                   }),
                 }),
-                // TODO: also test installed applications!
-                installedApplications: [],
```
emmenko, line 915: A long overdue leftover 🙃

graphql-test-utils/graphql-models/custom-application-installation.js
```diff
@@ -0,0 +1,11 @@
+import { Factory } from 'rosie';
```
emmenko, line 1: FYI: as a follow up we should rewrite these test-data models to use our new setup.

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -198,8 +198,8 @@ const MenuGroup = (props: MenuGroupProps) => {
     props.level === 2 && props.isActive && !props.isExpanded;
   return (
     <ul
-      id={`${props.id}-group`}
```
emmenko, line 201: Apparently querying for this ID (in tests) randomly fails if the value starts with a number (when using UUIDs).

image

Therefore I changed the prefix to be `group-`

CarlosCortizasCT: 😮

CarlosCortizasCT (approved):
Looks very good to me 🚀

kark (approved):
💯

PR #2771, opened 2022-08-19
chore: downgrade `jsdom` to v16

emmenko:
To avoid using conflicting versions of `jsdom` (`v19` vs `v16`) we downgrade our explicit `jsdom` dependency to `v16`.

This is an old version of `jsdom` (latest is `v20`) which is used by `jest-environment-jsdom`.

image

Since we're still using Jest v27, we can't and probably shouldn't upgrade to latest versions of `jsdom`. Once we are able to upgrade to Jest v28+, we can then upgrade `jsdom` again.

### DIFF
--- a/packages/application-config/package.json
+++ b/packages/application-config/package.json
@@ -42,7 +42,7 @@
     "core-js": "^3.23.4",
     "cosmiconfig": "7.0.1",
     "dompurify": "^2.3.6",
-    "jsdom": "^19.0.0",
+    "jsdom": "^16.7.0",
     "lodash": "4.17.21",
     "omit-empty-es": "1.1.3"
   },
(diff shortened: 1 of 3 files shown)


packages/application-config/package.json
```diff
@@ -42,7 +42,7 @@
     "core-js": "^3.23.4",
     "cosmiconfig": "7.0.1",
     "dompurify": "^2.3.6",
-    "jsdom": "^19.0.0",
```
emmenko, line 45: Btw, this was also kind of "wrong" as we require the types for v16 `"@types/jsdom": "^16.2.14"`.

CarlosCortizasCT (approved):
LGTM

tdeekens (approved):

Issue #2732, opened 2022-07-28
Frontend SDK forwardTo actions - support PATCH and PUT requests


alochowski:
Hello guys,

I don't see option to send `PATCH` and `PUT` action using **forwardTo** request by your SDK. Is there any chance to send that kind of type requests?
(according to this [doc](https://docs.commercetools.com/custom-applications/concepts/integrate-with-your-own-api#usage-for-sdk-actions))

And second question in this topic: Is there any chance to put payload to `DELETE` action by **forwardTo** request?

Regards,
Arek

Rhotimee:
Hello @alochowski, 
We currently do not support PATCH or PUT requests using forwardTo. 
However, you can use the workaround below. 

```
const action = actions.forwardTo.post({
  uri: 'https://my-custom-app.com/graphql',
  payload: { say: 'Hello' },
});
action.payload.method = 'PUT';
```

As for the second question, you can pass a payload to the DELETE action if you wish to.

alochowski:
Hi @Rhotimee,

Thanks for your answer and an interesting workaround :+1: 

Unfortunately doesn't work for me :/
I've got message from https://mc-api.europe-west1.gcp.commercetools.com/proxy/forward-to that Method `PUT` and `PATCH` is not allowed by Access-Control-Allow-Methods in preflight response.
Do you plan to handle these methods as well?

Regarding `DELETE` method. I've tried to add payload to this request, but unfortunately is not passed to my server by `forwarTo` feature. I don't see body in request from MC. `POST` works correctly.

emmenko:
Hi @alochowski 👋

Sorry for the late reply.

> is not allowed by Access-Control-Allow-Methods in preflight response

Thanks for pointing it out. We will fix that, we'll let you know when it's available.

> I don't see option to send `PATCH` and `PUT` action using forwardTo request by your SDK.
> Is there any chance to put payload to `DELETE` action by forwardTo request?

Currently we don't have any plans on supporting these changes.

I would suggest maybe to use another HTTP client (instead of the `@commercetools-frontend/sdk`) to handle these requests. We actually just released some improvements in providing better support for custom HTTP clients.
You can read more about it here: https://docs.commercetools.com/custom-applications/development/data-fetching#custom-http-client

In your case it could be something like this (example using `fetch`):

```js
import createHttpUserAgent from '@commercetools/http-user-agent';
import {
  buildApiUrl,
  executeHttpClientRequest,
[...]

PR #2772, opened 2022-08-19
refactor(app-shell): throw if required `forwardToConfig.uri` option is missing

emmenko:
See https://github.com/commercetools/merchant-center-application-kit/issues/2732#issuecomment-1220331401


CarlosCortizasCT (approved):
💯

PR #2778, opened 2022-08-22
chore: update flopflip dependencies

tdeekens:
#### Summary

Sadly rolled out a bug in flopflip which prevented flags to update. Internally we pinned it in a resolution. I tested and verified this release containing a fix.


CarlosCortizasCT (approved):
Thanks! 👍

PR #2781, opened 2022-08-24
Allow to request custom claims in forward-to requests

CarlosCortizasCT:
#### Summary

Allow custom claims in exchange JWT send in forwarded requests to external services.
Allow to request custom claims to be included in the exchange JWT sent from MC Gateway when sending requests from a Custom Application to an external service by using the `/proxy/forward-to` endpoint ([docs](https://docs.commercetools.com/custom-applications/concepts/integrate-with-your-own-api)).

NOTE: _we will have another PR for updating the docs_.

#### Description

We agreed that requesting the Gateway to include custom claims will be _triggered_ by including a new header (*X-Forward-To-Claims*) with a space separated list of wanted claims (we will only support *permissions* in this first iteration).

In this PR we update our requests helpers to make it easy to include the new helper:
* GraphQL context
* Custom HTTP client
* SDK actions

### DIFF
--- a/.changeset/fifty-cows-reflect.md
+++ b/.changeset/fifty-cows-reflect.md
@@ -14,7 +14,7 @@ const useExternalApiFetcher = () => {
   const { loading, data, error } = useMcQuery(MyQuery, {
     context: createApolloContextForProxyForwardTo({
       uri: externalApiUrl,
-      exchangeTokenClaims: ['permissions'],
+      includeUserPermissions: true,
     }),
   });
 
@@ -46,7 +46,7 @@ const data = await executeHttpClientRequest(
     userAgent,
     forwardToConfig: {
       uri: 'https://my-api.com/my-endpoint',
-      exchangeTokenClaims: ['permissions'],
+      includeUserPermissions: true,
     },
   }
 );
--- a/.changeset/fifty-cows-reflect.md
+++ b/.changeset/fifty-cows-reflect.md
@@ -5,16 +5,16 @@
 Add support for requesting specific `claims` to be included in the exchange JWT sent from Merchant Center API to an external API.
 Currently we only support requesting a custom claim with logged in user's permissions.
 
-```
+```js
 // Apollo example
 const useExternalApiFetcher = () => {
   const externalApiUrl = useApplicationContext(
-    context => context.environment.externalApiUrl
+    (context) => context.environment.externalApiUrl
   );
   const { loading, data, error } = useMcQuery(MyQuery, {
     context: createApolloContextForProxyForwardTo({
       uri: externalApiUrl,
-      exchangeTokenClaims: ['permissions']
+      exchangeTokenClaims: ['permissions'],
     }),
   });
 
@@ -22,11 +22,11 @@ const useExternalApiFetcher = () => {
     loading,
     data,
     error,
-  }
+  };
 };
 ```
 
-```
+```js
 // Custom HTTP client example (using `fetch`)
 const data = await executeHttpClientRequest(
   async (options) => {
@@ -46,7 +46,7 @@ const data = await executeHttpClientRequest(
     userAgent,
     forwardToConfig: {
       uri: 'https://my-api.com/my-endpoint',
-      exchangeTokenClaims: ['permissions']
+      exchangeTokenClaims: ['permissions'],
     },
   }
 );
--- a/.changeset/fifty-cows-reflect.md
+++ b/.changeset/fifty-cows-reflect.md
@@ -1,0 +1,53 @@
+---
+'@commercetools-frontend/application-shell': patch
+---
+
+Add support for requesting specific `claims` to be included in the exchange JWT sent from Merchant Center API to an external API.
+Currently we only support requesting a custom claim with logged in user's permissions.
+
+```
+// Apollo example
+const useExternalApiFetcher = () => {
+  const externalApiUrl = useApplicationContext(
+    context => context.environment.externalApiUrl
+  );
+  const { loading, data, error } = useMcQuery(MyQuery, {
+    context: createApolloContextForProxyForwardTo({
+      uri: externalApiUrl,
+      exchangeTokenClaims: ['permissions']
+    }),
+  });
+
+  return {
+    loading,
+    data,
+    error,
+  }
+};
+```
+
+```
+// Custom HTTP client example (using `fetch`)
+const data = await executeHttpClientRequest(
+  async (options) => {
+    const res = await fetch(buildApiUrl('/proxy/forward-to'), {
+      method: 'GET',
+      ...options,
+    });
+    const data = await res.json();
+
+    return {
+      data,
+      statusCode: res.status,
+      getHeader: (key) => res.headers.get(key),
+    };
+  },
+  {
+    userAgent,
+    forwardToConfig: {
+      uri: 'https://my-api.com/my-endpoint',
+      exchangeTokenClaims: ['permissions']
+    },
+  }
+);
+```
--- a/.changeset/twenty-pears-admire.md
+++ b/.changeset/twenty-pears-admire.md
@@ -1,0 +1,15 @@
+---
+'@commercetools-frontend/sdk': patch
+---
+
+Add support for requesting specific `claims` to be included in the exchange JWT sent from Merchant Center API to an external API.
+Currently we only support requesting a custom claim with logged in user's permissions.
+
+```
+actions.forwardTo.get({
+  exchangeTokenClaims: ['permissions'],
+  // ...
+});
+```
+
+// "X-Forward-To-Claims": "permissions"
--- a/.changeset/twenty-pears-admire.md
+++ b/.changeset/twenty-pears-admire.md
@@ -7,7 +7,7 @@ Currently we only support requesting a custom claim with logged in user's permis
 
 ```js
 actions.forwardTo.get({
-  exchangeTokenClaims: ['permissions'],
+  includeUserPermissions: true,
   // ...
 });
 // "X-Forward-To-Claims": "permissions"
--- a/.changeset/twenty-pears-admire.md
+++ b/.changeset/twenty-pears-admire.md
@@ -5,7 +5,7 @@
 Add support for requesting specific `claims` to be included in the exchange JWT sent from Merchant Center API to an external API.
 Currently we only support requesting a custom claim with logged in user's permissions.
 
-```
+```js
 actions.forwardTo.get({
   exchangeTokenClaims: ['permissions'],
   // ...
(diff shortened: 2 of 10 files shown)


packages/application-shell/src/utils/http-client.ts
```diff
@@ -39,6 +41,10 @@ export type TForwardToConfig = {
    * The audience policy for verifying the incoming request from the Merchant Center API.
    */
   audiencePolicy?: TForwardToAudiencePolicy;
+  /**
+   * The claims to be included in the exchange JWT sent to the external API.
+   */
+  exchangeTokenClaims?: TForwardToExchangeTokenClaim[];
```
emmenko, line 47: I feel like the name might be a bit confusing and inconsistent with other options. For example the `audiencePolicy` is also technically a claim but we named it differently to better express its meaning.

How about something like `additionalFeatures`?

On the other hand I'm also wondering if we should expose the new claims as individual options. So for `permissions` it would be something like `includeUserPermissions`.

```js
{
  includeUserPermissions: true,
}
```

We can then follow the naming pattern `includeX` for other features. Internally we map those to the HTTP header `X-Forward-To-Claims`.

What do you think?

emmenko (commented):
Did a first round of review.

Looking very good thanks, just one main suggestion about naming the option🤗

packages/application-shell/src/utils/http-client.ts
CarlosCortizasCT: I tried to use a name very specific to what the prop will be used for, but maybe it's just meaningful to me because I know about the inners of this topic and it will not be as clear for our users.

So I guess your proposal is either using `additionalFeatures` or `includeX`, right?
Personally I find `additionalFeatures` a little abstract and I prefer the `includeX` pattern, although we need to take into account that, if we choose it, we will need to release a new `app-kit` version every time we want to use a new "claim". This might not be a big problem since we would need to do it anyway because of the express auth middleware (reading the new claim from the token and expose it in the request).

.changeset/fifty-cows-reflect.md
```diff
@@ -0,0 +1,53 @@
+---
+'@commercetools-frontend/application-shell': patch
+---
+
+Add support for requesting specific `claims` to be included in the exchange JWT sent from Merchant Center API to an external API.
+Currently we only support requesting a custom claim with logged in user's permissions.
+
+```
```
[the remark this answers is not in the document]
CarlosCortizasCT: Updated in this commit: 722895146356957f32571dbcddb8872f143da2c8

.changeset/fifty-cows-reflect.md
```diff
@@ -0,0 +1,53 @@
...
+
+  return {
+    loading,
+    data,
+    error,
+  }
+};
+```
+
+```
```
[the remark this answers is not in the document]
CarlosCortizasCT: Updated in this commit: 722895146356957f32571dbcddb8872f143da2c8

packages/application-shell/src/utils/http-client.ts
emmenko: Yes for every new feature claim we would need to release a new version anyway.
This is actually good in the sense that in the documentation we can clearly state from which version onwards that particular feature is available.

Then let's try with the `includeX` pattern?

CarlosCortizasCT: Updated in this commit: 31ecb19b93fb2e4c6985ea070ad24382d33b348b

packages/application-shell/src/utils/http-client.ts
```diff
@@ -42,9 +42,10 @@ export type TForwardToConfig = {
    */
   audiencePolicy?: TForwardToAudiencePolicy;
   /**
-   * The claims to be included in the exchange JWT sent to the external API.
+   * Request the Merchange Center API to include the user permission in the
+   * exchange token sent in the forwarded request to the external API.
```
emmenko, line 46: Nit (to follow the wording used for the other options)

```suggestion
   * A list of user permissions to be included in the request to the external API.
```

packages/application-shell/src/utils/http-client.ts
```diff
@@ -172,12 +179,10 @@ const getAppliedForwardToHeaders = (
     [SUPPORTED_HEADERS.X_FORWARD_TO]: forwardToConfig.uri,
     [SUPPORTED_HEADERS.X_FORWARD_TO_AUDIENCE_POLICY]:
       forwardToConfig.audiencePolicy ?? defaultForwardToAudiencePolicy,
-    ...(Boolean(forwardToConfig.exchangeTokenClaims?.length)
+    ...(Boolean(exchangeTokenClaims?.length)
```
emmenko, line 182: The condition shouldn't be necessary, because empty headers will be removed by the `omitEmpty`. So this can simply be:

```ts
[SUPPORTED_HEADERS.X_FORWARD_TO_CLAIMS]: exchangeTokenClaims.join(' '),
```

packages/sdk/src/actions/actions.ts
```diff
@@ -69,12 +76,10 @@ const enhancePayloadForForwardToProxy = (payload: TSdkActionPayloadForUri) => {
       'X-Forward-To': payload.uri,
       'X-Forward-To-Audience-Policy':
         payload.audiencePolicy || 'forward-url-full-path',
-      ...(Boolean(payload.exchangeTokenClaims?.length)
+      ...(Boolean(exchangeTokenClaims?.length)
```
emmenko, line 79: Same here.

PS: we should use the `omitEmpty` function actually here as well. Could you add that? Thanks

```diff
-headers: {
+headers: omitEmpty<THeaders>({
```

.changeset/fifty-cows-reflect.md
```diff
@@ -0,0 +1,53 @@
...
+```js
+// Apollo example
+const useExternalApiFetcher = () => {
+  const externalApiUrl = useApplicationContext(
+    (context) => context.environment.externalApiUrl
+  );
+  const { loading, data, error } = useMcQuery(MyQuery, {
+    context: createApolloContextForProxyForwardTo({
+      uri: externalApiUrl,
+      exchangeTokenClaims: ['permissions'],
```
emmenko, line 17:
```suggestion
      includeUserPermissions: true,
```

.changeset/fifty-cows-reflect.md
```diff
@@ -0,0 +1,53 @@
...
+      data,
+      statusCode: res.status,
+      getHeader: (key) => res.headers.get(key),
+    };
+  },
+  {
+    userAgent,
+    forwardToConfig: {
+      uri: 'https://my-api.com/my-endpoint',
+      exchangeTokenClaims: ['permissions'],
```
emmenko, line 49:
```suggestion
      includeUserPermissions: true,
```

.changeset/twenty-pears-admire.md
```diff
@@ -0,0 +1,15 @@
+---
+'@commercetools-frontend/sdk': patch
+---
+
+Add support for requesting specific `claims` to be included in the exchange JWT sent from Merchant Center API to an external API.
+Currently we only support requesting a custom claim with logged in user's permissions.
+
+```js
+actions.forwardTo.get({
+  exchangeTokenClaims: ['permissions'],
```
emmenko, line 10:
```suggestion
  includeUserPermissions: true,
```

.changeset/twenty-pears-admire.md
```diff
@@ -0,0 +1,15 @@
...
+  // ...
+});
+```
+
+// "X-Forward-To-Claims": "permissions"
```
emmenko, line 15:
````suggestion
// "X-Forward-To-Claims": "permissions"
```
````

**Review by emmenko — COMMENTED** [review]
Thanks, I think this is better now and more clear.

Just some minor adjustments left, otherwise 👌

File: .changeset/fifty-cows-reflect.md (hunk 7 above)
Comment (CarlosCortizasCT) [reply .changeset/fifty-cows-reflect.md]: Updated in this commit: 0f7be5b44a05e5012364c8ead78856d761445fd4

File: .changeset/fifty-cows-reflect.md (hunk 8 above)
Comment (CarlosCortizasCT) [reply .changeset/fifty-cows-reflect.md]: Updated in this commit: 0f7be5b44a05e5012364c8ead78856d761445fd4

File: .changeset/twenty-pears-admire.md (hunk 9 above)
Comment (CarlosCortizasCT) [reply .changeset/twenty-pears-admire.md:10]: Updated in this commit: 0f7be5b44a05e5012364c8ead78856d761445fd4

File: .changeset/twenty-pears-admire.md (hunk 10 above)
Comment (CarlosCortizasCT) [reply .changeset/twenty-pears-admire.md]: Updated in this commit: 0f7be5b44a05e5012364c8ead78856d761445fd4

File: packages/sdk/src/actions/actions.ts (hunk 6 above)
Comment (CarlosCortizasCT) [reply packages/sdk/src/actions/actions.ts]: Updated in this commit: 7cd2e9fb90a205a7b229b646c5166a3c718b37bc

File: packages/application-shell/src/utils/http-client.ts (hunk 5 above)
Comment (CarlosCortizasCT) [reply packages/application-shell/src/utils/http-client.ts]: Updated in this commit: de46f8f9da017cfa7be43230c0d08b17eaa67e0b

File: packages/application-shell/src/utils/http-client.ts (hunk 4 above)
Comment (CarlosCortizasCT) [reply packages/application-shell/src/utils/http-client.ts]: Updated in this commit: de46f8f9da017cfa7be43230c0d08b17eaa67e0b

**Review by emmenko — APPROVED** [review]
Looks good to me now! 💯

File: packages/application-shell/src/utils/http-client.ts [hunk 11]
```diff
@@ -150,6 +156,12 @@ const getAppliedForwardToHeaders = (
   if (!forwardToConfig.uri) {
     throw new Error(`Missing required "uri" option.`);
   }
+
+  const exchangeTokenClaims: TForwardToExchangeTokenClaim[] = [];
+  if (forwardToConfig.includeUserPermissions) {
+    exchangeTokenClaims.push('permissions');
```
ddouglasz, line 162: Good work @CarlosCortizasCT 🦾. 
Just curious, So after checking that the forwardToConfig has the `includeUserPermissions` property, we then add `permissions` into `exchangeTokenClaims`. I assume this is later used to build the exchangeTokenClaims right?

ddouglasz (approved):

packages/application-shell/src/utils/http-client.ts
CarlosCortizasCT: Yes.
So `exchangeTokenClaims` will include any custom claim we should include in the new `X-Forward-To-Claims` request header.
That header is added to the headers list that will be included in the request [in this same file](https://github.com/commercetools/merchant-center-application-kit/blob/7e19d7ba31c513e37eb3dc56bccf041c68104c99/packages/application-shell/src/utils/http-client.ts#L179).

ddouglasz: Thanks for clarifying 🙏

PR #2782, opened 2022-08-24
Read optional user permissions claim in backend auth middleware

CarlosCortizasCT:
#### Summary

Write (optional) user permissions in `request.session` object when validating an incoming requests in external services.

#### Description

When using our _auth_ middleware in an external service to validate incoming requests, it read information from the exchange JWT included in the incoming request and writes it into an attribute (called `session`) in the request for consumers to be able to access that info.

Now that the Merchant Center API can include a new public claim with logged in user permissions (when configured at the time of making the request from a Custom Application), the middleware will also read that value and put it into the `request.session` object under a property called *userPermissions*.
The value will be an array of strings with the permissions names.

### DIFF
--- a/.changeset/gorgeous-kings-cough.md
+++ b/.changeset/gorgeous-kings-cough.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-backend/express': patch
+---
+
+Now that users can request the Merchant Center API to include logged in user's permissions in the exchange JWT included in proxied requests, our _auth_ middleware will check for that information in the token and will make it available in the `request.session` object under a new property named **userPermissions**.
--- a/.changeset/gorgeous-kings-cough.md
+++ b/.changeset/gorgeous-kings-cough.md
@@ -2,4 +2,4 @@
 '@commercetools-backend/express': patch
 ---
 
-Now that users can request the Merchant Center API to include logged in user's permissions in the exchange JWT included in proxied requests, our _auth_ middleware will check for that information in the token and will make it available in the `request.session` object under a new property named **userPermissions**.
+Now that users can request the Merchant Center API to include logged in user's permissions in the exchange JWT included in proxied requests, our _auth_ middleware will check for that information in the token and will make it available in the `request.session` object under a new property named **userPermissions** (reference [documentation](https://docs.commercetools.com/custom-applications/concepts/integrate-with-your-own-api#validating-the-json-web-token)).
--- a/packages-backend/express/src/auth.spec.js
+++ b/packages-backend/express/src/auth.spec.js
@@ -59,37 +59,32 @@ describe('writeSessionContext', () => {
     expect(request).not.toHaveProperty('session');
   });
 
-  it.each`
-    publicClaims | sessionProperties
-    ${{
-  [`${mockIssuer}/claims/project_key`]: 'fake-project-key',
-}} | ${['projectKey']}
-    ${{
-  [`${mockIssuer}/claims/project_key`]: 'fake-project-key',
-  [`${mockIssuer}/claims/user_permissions`]: ['viewOrders', 'manageProducts'],
-}} | ${['projectKey', 'userPermissions']}
+  describe.each`
+    claimKey                                   | claimValue                          | expectedSessionPropertyKey
+    ${`sub`}                                   | ${'user-id-1'}                      | ${'userId'}
+    ${`${mockIssuer}/claims/project_key`}      | ${'fake-project-key'}               | ${'projectKey'}
+    ${`${mockIssuer}/claims/user_permissions`} | ${['viewOrders', 'manageProducts']} | ${'userPermissions'}
   `(
-    'should write session data in the request with new a property for every public claim',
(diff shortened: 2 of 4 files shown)


packages-backend/express/src/auth.ts
```diff
@@ -19,23 +19,29 @@ import { getFirstHeaderValueOrThrow } from './utils';
...
   sub: string;
   iss: string;
-  [property: string]: string;
+  [property: string]: string | string[];
 };
 
 const decodedTokenKey = 'decoded_token';
 // Assign a session object to the request object.
-const writeSessionContext = <Request extends TBaseRequest>(
+export const writeSessionContext = <Request extends TBaseRequest>(
```
emmenko, line 27: Nit: to keep the exports consistent and visible we define it at the bottom of the file

```js
export { createSessionAuthVerifier };
```

Can you add this function to be exported there? Thanks

.changeset/gorgeous-kings-cough.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-backend/express': patch
+---
+
+Now that users can request the Merchant Center API to include logged in user's permissions in the exchange JWT included in proxied requests, our _auth_ middleware will check for that information in the token and will make it available in the `request.session` object under a new property named **userPermissions**.
```
emmenko, line 5: Nit: we can also add a link to the docs (even though they are not live yet).

packages-backend/express/src/auth.spec.js
```diff
@@ -47,3 +47,55 @@ describe('getConfiguredAudience', () => {
...
+
+  it('should not include anything in request session if no "decoded_token" exists in the request', () => {
+    const request = {};
+
+    writeSessionContext(request);
+
+    expect(request).not.toHaveProperty('session');
+  });
+
+  it('should write "userId" and "projectKey" in request session property', () => {
```
emmenko, line 62: Nit: it look like this is basically the same as the other test. We can just have one with all possible properties.

packages-backend/express/src/auth.spec.js
```diff
@@ -47,3 +47,55 @@ describe('getConfiguredAudience', () => {
...
+    const mockDecodedToken = {
+      iss: mockIssuer,
+      sub: 'user-id-1',
+      [`${mockIssuer}/claims/project_key`]: 'almond-40',
```
emmenko, line 66: Let's use a fake project key 😉

emmenko (commented):
Nice, thanks!

.changeset/gorgeous-kings-cough.md
CarlosCortizasCT: Updated in this commit: b80cd139edcc473ceff46de82772a70329dcecc8

packages-backend/express/src/auth.spec.js
CarlosCortizasCT: Updated in this commit: 335c1119c938fd8d183875fd828803fa4afa42d0

packages-backend/express/src/auth.spec.js
CarlosCortizasCT: I'm sorry but I'm not sure I understand how to merge both tests: the first one would test that, if no `decoded_token` exists in the request, no error is raised and no `session` property is attached to the request. for the second one we want to test that the `session` property is attached to the request with the required info.
I thought having both tests would be more readable.

So what you would prefer is to have one test where we provided different tokens (all three cases currently tested) and the different expectations?

emmenko: I mean the test below, for the "userPermissions".

emmenko: Sorry for not being super clear. I pushed a commit with what I had in mind https://github.com/commercetools/merchant-center-application-kit/pull/2782/commits/12a73919ca7a7761b0a7cdbb52cdbc6116e302a4

image

Essentially all these properties can be tested individually, for the sake of the test they don't have any relation with each other.

CarlosCortizasCT: Ok. I wanted to test both of them together as that's what we will get at runtime, but I'm also ok with this simpler approach.

Thanks for contirbuting 🙇

emmenko (approved):
💯

Rhotimee (approved):
🙌🏽

packages-backend/express/src/auth.ts
```diff
@@ -31,11 +31,17 @@ const writeSessionContext = <Request extends TBaseRequest>(
...
 
     request.session = {
       userId: decodedToken.sub,
-      projectKey: decodedToken[publicClaimForProjectKey],
+      projectKey: decodedToken[publicClaimForProjectKey] as string,
     };
+
+    const userPermissions = decodedToken[publicClaimForUserPermissionsKey];
+    if (Boolean(userPermissions?.length)) {
+      request.session.userPermissions = userPermissions as string[];
```
ddouglasz, line 43: 💯

ddouglasz (approved):
Looks good! 👏

PR #2784, opened 2022-08-28
 Fix Renovate dependency updates to skip postinstall script

tdeekens:
#### Summary

This pull request skips the post install step/script whenever the PR is coming from Renovate.

#### Description

Renovate pull requests are failing across repositories. The Renovate pull requests fail as postinstall scripts run and perform tasks which aren't allowed in Renovate's sandbox.

Thinking about it conceptually Renovate has no interest in running post install steps. They're always unrelated to the dependency update itself which just bumps versions and updates a lockfile.

In other dependency managers: `yarn@2` or `npm` for that matter by using `--ignore-scripts`. Moreover, it has the notion of `postupgradetasks` as a config to run tasks before the commit is made but _after_ the dependencies are updated. Initially all a clear indicator that Renovate does not intend to run post install scripts. This assumption was proven right in my discussion with their team below.

After reading Renovate's configuration and trying to find alternative solutions I settled on this until there is hopefully something better.

There are better alternatives to this approach:

1. `yarn@3` and the `postinstall` plugin allows skipping with a e.g. `--skip-postinstall`
    - I intend to open a issue there. However Renovate can't assumpe that only one postinstall plugin is used as there are multiple for yarn
2. Renovate sets a `SKIP_POST_INSTALL` environment variable
    - It is a custom solution which is not understood by `yarn@3` natively and would require changes in user land

I tried out this change on flopflip and it works there on PRs.

Renovate discussion: https://github.com/renovatebot/renovate/discussions/17442
Slack Reference: https://commercetools.slack.com/archives/C020CPEGQLX/p1661519362801979

P.S.: Open to better solutions but after hours of debugging this is close to the only solution I can think of for now.

### DIFF
--- a/scripts/postinstall.sh
+++ b/scripts/postinstall.sh
@@ -4,14 +4,12 @@ set -e
 
 if [ -n "$SKIP_POSTINSTALL_DEV_SETUP" ]; then
   echo "Skipping development setup."
-elif [[ $NPM_CONFIG_CACHE =~ "renovate" ]]; then
-  echo "Renovate update detected, skipping post install."
+elif [[ $CI == "true" ]]; then
+  echo "Running on CI, skipping post install."
 else
   echo "Preparing development setup."
-  yarn husky install
-  yarn manypkg check
-  yarn preconstruct dev
 
+  yarn setup
 fi
 
 echo "Running prettier on package.json files"
--- a/scripts/postinstall.sh
+++ b/scripts/postinstall.sh
@@ -4,8 +4,8 @@ set -e
 
 if [ -n "$SKIP_POSTINSTALL_DEV_SETUP" ]; then
   echo "Skipping development setup."
-elif [[ $CI == "true" ]]; then
-  echo "Running on CI, skipping post install."
+elif [[ $YARN_ENABLE_SCRIPTS == "0" ]]; then
+  echo "Yarn scripts disabled, skipping post install."
 else
   echo "Preparing development setup."
 
--- a/scripts/postinstall.sh
+++ b/scripts/postinstall.sh
@@ -4,7 +4,8 @@ set -e
 
 if [ -n "$SKIP_POSTINSTALL_DEV_SETUP" ]; then
   echo "Skipping development setup."
-
+elif [[ $NPM_CONFIG_CACHE =~ "renovate" ]]; then
+  echo "Renovate update detected, skipping post install."
 else
   echo "Preparing development setup."
   yarn husky install
--- a/scripts/postinstall.sh
+++ b/scripts/postinstall.sh
@@ -4,8 +4,9 @@ set -e
 
 if [ -n "$SKIP_POSTINSTALL_DEV_SETUP" ]; then
   echo "Skipping development setup."
-elif [[ $YARN_ENABLE_SCRIPTS == "0" ]]; then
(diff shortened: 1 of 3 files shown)


scripts/postinstall.sh
```diff
@@ -4,7 +4,8 @@ set -e
 
 if [ -n "$SKIP_POSTINSTALL_DEV_SETUP" ]; then
   echo "Skipping development setup."
-
+elif [[ $NPM_CONFIG_CACHE =~ "renovate" ]]; then
```
tdeekens, line 7: The environment variable is set by Renovate and is "NPM_CONFIG_CACHE": "/tmp/renovate-cache/others/npm" anywhere I looked.

antoniolodias (approved):

emmenko:
Thanks @tdeekens for looking into it! Weird that it was working until a while ago and then it didn't.

Regarding the environment variable, I guess these are the ones available when renovate runs?

image

As an alternative, maybe we can look at the branch name? If it starts with `renovate`?

scripts/postinstall.sh
emmenko: Do we need to skip the entire `postinstall.sh` script to be executed or only certain parts?

For instance, the `yarn prettier` command is still executed here.

tdeekens:
1. Yes those are the variables Renovate sets. 
2. Yes, it was working but no consistently isn't across repositories. It surely is something in the postinstall step. One could also debug it one by one. 
3. I would have liked to use the branch name too but I am not sure we can use it unless I am mistaken. Keep in mind that the Renovate PRs first do not run on CircleCI or GitHub but Renovate's infrastructure. So it's not exposed as an env variable. I am not 100% sure git is installed there to get the branch name. Maybe worth a shot that way around if you mean that.

emmenko:
For the branch, I just saw that in [one of your logs screenshots](https://github.com/renovatebot/renovate/discussions/17442#discussioncomment-3492596) and was wondering if we can use that.

image

> what we would do is also skip the post install ok our CI which is not what we want. We only want to skip it in Renovate's infrastructure.

In theory we don't need the postinstall on CI, as it's more for development. Maybe only the `prettier` part, which we can also have as its own CI step.
So then we could simply check for `CI = 'true'`?

What do you think?

scripts/postinstall.sh
tdeekens: If you know and are willing to debug which parts can be included and excluded we can skip those. Still I think it's besides the point.

I didn't intend this to be a permanent solution. As noted in the description. Ideally Renovate should be able to skip the entire post install as it does with older yarn versions or NPM. It's not relevant for Renovate to run it.

tdeekens:
> In theory we don't need the postinstall on CI, as it's more for development. Maybe only the prettier part, which we can also have as its own CI step.

Theoretical and option but I didn't want to prescribe this as a solution for all repositories as a fix as I would still hope that Renovate could eventually skip post installs again as it did before yarn@3. 

What I did in flopflip is having a `setup` run script which I call from post install and CI. Then skip in the post install if we're on CI. Is that what you prefer too?

tdeekens:
It's not always set yet. So let's wait for their team's clarification.

emmenko:
FYI: I forked the postinstall plugin and applied the suggested fixes. See #2785 

We can close this I guess.

PR #2797, opened 2022-09-07
refactor: to remove jest/no-jest-import rule

tdeekens:
#### Summary

In `eslint-plugin-jest` v27 the rule `jest/no-jest-import` has been removed. So it can’t be used in the `eslint-config-mc-app`. Given that it’s not recommended to be used anymore already [here](https://github.com/jest-community/eslint-plugin-jest/releases/tag/v27.0.0) we could remove it already.

This will mean that consumers can update to v27 of `eslint-plugin-jest` already.


CarlosCortizasCT (approved):
👌

kark (approved):

PR #2801, opened 2022-09-12
refactor(application-shell): include additional OAuth scopes in OIDC flow for local development

kark:
https://jira.commercetools.com/browse/SHIELD-581

In preparation for the Custom Applications Multiple Permissions support.

### DIFF
--- a/packages/application-config/src/process-config.ts
+++ b/packages/application-config/src/process-config.ts
@@ -129,6 +129,7 @@ const processConfig = ({
                 : appConfig.env.development.initialProjectKey,
             teamId: appConfig.env.development?.teamId,
             oAuthScopes: appConfig.oAuthScopes,
+            additionalOAuthScopes: appConfig?.additionalOAuthScopes,
           }),
           menuLinks: {
             icon: customApplicationData.icon,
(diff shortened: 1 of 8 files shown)


packages/application-config/src/process-config.ts
```diff
@@ -129,6 +129,7 @@ const processConfig = ({
                 : appConfig.env.development.initialProjectKey,
             teamId: appConfig.env.development?.teamId,
             oAuthScopes: appConfig.oAuthScopes,
+            additionalOAuthScopes: appConfig?.additionalOAuthScopes,
```
kark, line 132: Let me pin my comment here.

This file calls `transformCustomApplicationConfigToData` function (defined [here](https://github.com/commercetools/merchant-center-application-kit/blob/main/packages/application-config/src/transformers.ts)) which returns an object containing [`permissions` array](https://github.com/commercetools/merchant-center-application-kit/blob/91c0da4f9afa92e44a22c8c3af38f545488d5e44/packages/application-config/src/transformers.ts#L38:L47).

It seems this property is not used anywhere in the codebase, and I'm wondering if that should be somehow updated (or perhaps even removed)? I guess I'm missing some context 🤔

packages/application-shell/src/utils/oidc.ts
```diff
@@ -27,6 +28,21 @@ const buildOidcScope = (options: BuilOidcScopeOptions): string => {
...
+          (viewOAuthScope) =>
+            `view/${additionalOAuthScope.name}:${viewOAuthScope}`
+        ),
+        ...(additionalOAuthScope.manage ?? []).map(
+          (manageOAuthScope) =>
+            `manage/${additionalOAuthScope.name}:${manageOAuthScope}`
+        )
+      );
+    });
+  }
```
emmenko, line 45: 💯

emmenko (commented):
Had a quick look. Nice, all good! 🙌

emmenko (approved):

PR #2802, opened 2022-09-12
Fix page title re-render issue with title still displaying previous location.

ddouglasz:
Fix page title re-render issue with title still displaying previous location instead of the current one.

Navigating to a new page actually persists the initial value, as a result, returning from a `products/<products_id>` page to say `/products` page, the title would still remain as though it was still in the `products/<products_id>` page instead of updating to the new location, in this case, `/products` .


emmenko (approved):
💯

tdeekens (approved):

PR #2807, opened 2022-09-15
fix: disable redux serializable check

emmenko:
Leftover of #2776


packages/application-shell/src/configure-store.ts
```diff
@@ -122,10 +122,7 @@ export const createReduxStore = (
       sdkMiddleware,
       ...getDefaultMiddleware({
         // https://redux-toolkit.js.org/usage/usage-guide#working-with-non-serializable-data
-        serializableCheck: {
-          // Ignore these field paths in all actions
-          ignoredActionPaths: ['dismiss'],
-        },
+        serializableCheck: false,
```
emmenko, line 125: This was the original behavior (no check) so let's just keep it like this for simplicity.

Rhotimee (approved):
👍🏽

PR #2813, opened 2022-09-19
refactor(pages): Move form control to the same line as title row.

Rhotimee:
#### Summary
Remove unnecessary space from form control, making the layout more compact

#### Screenshots
Before
image

After
image

### DIFF
--- a/packages/application-components/src/components/detail-pages/custom-form-detail-page/custom-form-detail-page.tsx
+++ b/packages/application-components/src/components/detail-pages/custom-form-detail-page/custom-form-detail-page.tsx
@@ -12,6 +12,7 @@ import Spacings from '@commercetools-uikit/spacings';
 import PageHeaderTitle from '../../internals/page-header-title';
 import PageTopBar from '../../internals/page-top-bar';
 import { ContentWrapper, PageWrapper } from '../../internals/page.styles';
+import { FormControlContainer } from '../../internals/form-control.styles';
 
 const DetailPageContainer = styled.div`
   background-color: ${customProperties.colorNeutral95};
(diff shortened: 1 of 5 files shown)


emmenko:
Could you provide some screenshots (before/after) to have a better idea of what exactly changed? Thanks

Rhotimee:
> Could you provide some screenshots (before/after) to have a better idea of what exactly changed? Thanks

Updated.

emmenko:
Wait a second, in the detail page we also have the form controls in a separate row

image

What's the reason for changing this for the main page? I'm a bit confused now 🤔

Rhotimee:
> Wait a second, in the detail page we also have the form controls in a separate row
> What's the reason for changing this for the main page? I'm a bit confused now 🤔

This was changed in the [design](https://commercetools.invisionapp.com/spec/cl5s2aq2u0zbx01zefyvi86p3/select). 
@FilPob and I also talked about this.

emmenko:
We should be consistent with the detail pages, one way or the other. @FilPob Can you provide some guidance which way it should be?
Either way, it needs to be consistent.

FilPob:
> 

I thought the form controls would never be in a separate row which occupies the full width. It's like that in Modal pages and also in the tabular detail page... So I guess then it's rather the non-tabular detail page which should be adjusted. Otherwise the form controls take an unnecessary amount of vertical space.

FilPob:
Also in the tabular detail page we have form controls which are in the same line as the tabs 
image

emmenko:
You mean to be like this?

image

FilPob:
Yes :)

emmenko:
Great, thanks.

@Rhotimee Can you take care of this? Thanks

Rhotimee:
> Great, thanks.
> 
> @Rhotimee Can you take care of this? Thanks

Sure 👍🏽

packages/application-components/src/components/detail-pages/custom-form-detail-page/custom-form-detail-page.tsx
```diff
@@ -87,18 +88,22 @@ const CustomFormDetailPage = (props: CustomFormDetailPageProps) => {
...
-              subtitle={props.subtitle}
-              titleSize="big"
-            />
-          )}
-          {!props.hideControls && props.formControls && (
-            <Spacings.Inline justifyContent="flex-end">
-              {props.formControls}
-            </Spacings.Inline>
-          )}
+          <Spacings.Inline scale="l" justifyContent="space-between">
```
emmenko, line 91: This is not correct. The `customTitleRow` should be in its own row.

image

emmenko (changes requested):
Hmm I don't think this is the right approach. Please also make sure the different layout combinations are correct. It might be that we also need to add one or two more VRTs.

I suppose all (form) pages (modal, detail, main) should have the same layout for the title, form controls, etc.
Maybe we should try to unify that logic.

PR #2814, opened 2022-09-19
feat: enables configuring of webpack dev server port via HTTP_PORT en…

RobertBorg:
…v var

#### Summary

By using process.env.HTTP_PORT for deciding developmentPort same as is done for start command at https://github.com/commercetools/merchant-center-application-kit/blob/main/packages/mc-scripts/src/commands/start.ts#L33
we can use a custom local port for dev server.
#### Description
as the initial compilation with mc-scripts start takes about half a year it would be really nice if i didn't have to restart it because something else needs port 3001

The issue of the callback after login is being called on localhost:3001 
remains but at least that is something that can be easily manually handled.

### DIFF
--- a/packages/application-config/src/process-config.ts
+++ b/packages/application-config/src/process-config.ts
@@ -25,8 +25,7 @@ type ProcessConfigOptions = Partial<LoadingConfigOptions> & {
   disableCache?: boolean;
 };
 
-// TODO: make it configurable.
-const developmentPort = 3001;
+const developmentPort = parseInt(String(process.env.HTTP_PORT), 10) || 3001;
 const developmentAppUrl = `http://localhost:${developmentPort}`;
 
 const omitDevConfigIfEmpty = (


emmenko:
Hi 👋,

Thanks for contributing! 

Unfortunately we can't make the port configurable yet, as there are other parts of the system that expect the port `3001` to be used.

The `start` command reads from `process.env.HTTP_PORT` yes. I can see how that might be confusing as we don't really support it (it was how `react-scripts` implemented it and we "just took it as-is").

> as the initial compilation with mc-scripts start takes about half a year it would be really nice if i didn't have to restart it because something else needs port 3001

It seems that the original issue comes from the fact the the dev server takes a lot of time to load? Is that the case also on subsequent starts? I believe Webpack uses a cache so I expect subsequent starts to be faster.
Is there any info you can share with us as to why it might take long? For example big codebase, do you use code splitting?, your computer has low CPU/RAM?

Thanks

RobertBorg:
> Hi 👋,
> 
> Thanks for contributing!
> 
> Unfortunately we can't make the port configurable yet, as there are other parts of the system that expect the port `3001` to be used.
> 
> The `start` command reads from `process.env.HTTP_PORT` yes. I can see how that might be confusing as we don't really support it (it was how `react-scripts` implemented it and we "just took it as-is").
> 
> > as the initial compilation with mc-scripts start takes about half a year it would be really nice if i didn't have to restart it because something else needs port 3001
> 
> It seems that the original issue comes from the fact the the dev server takes a lot of time to load? Is that the case also on subsequent starts? I believe Webpack uses a cache so I expect subsequent starts to be faster. Is there any info you can share with us as to why it might take long? For example big codebase, do you use code splitting?, your computer has low CPU/RAM?
> 
> Thanks

turning filesystem caching on for development builds solved the issue for me. Thanks!

emmenko:
Thanks for the feedback. Glad that you managed to improve your workflow.

Btw, we started working on switching the bundler (Webpack) to Vite which should improve the performance. It's not fully ready yet, so it's just an FYI for now.

ghandic:
Is it still not possible to set the port? I want to run multiple at once in a monorepo

PR #2817, opened 2022-09-20
fix(mc-scripts): enable sourcemaps in Vite build

emmenko:
Ref #2577


Rhotimee (approved):
💯

tdeekens (approved):

emmenko:
So it seems that when sourcemaps are enabled, Vite consumes a lot of memory.

https://github.com/vitejs/vite/issues/2433

Let's keep it disabled by default and only try using it on demand.

PR #2819, opened 2022-09-20
fix(sdk): return fixed reference for dispatch function

### LINKED ISSUE #2818: useAsyncDispatch returns a new reference every time it's called.
Consider the following line:

`const dispatch = useAsyncDispatch();`

When using dispatch in an effect it will add `dispatch` to the dependencies:

```
  useEffect(() => {
    dispatch(...)
  }, []);//dispatch is a dependency
```

Because useAsyncDispatch returns a new reference every time it's called you either cannot add it to the dependency list or will cause an infinite loop:

```
import { useEffect, useState } from 'react';
import { useAsyncDispatch, actions } from '@commercetools-frontend/sdk';

let lastValue = undefined;
const Component = () => {
  const [data, setData] = useState();
  const dispatch = useAsyncDispatch();
  if (!lastValue) {
    console.log('...setting last value');
    lastValue = dispatch;
  }
  useEffect(() => {
    async function execute() {
      try {
        const result = await dispatch(
          //docs: https://docs.commercetools.com/custom-applications/development/data-fetching#rest-http-client
          actions.forwardTo.get({
            uri: 'uri',
          })
        );
        setData(result);
        // Update state with `result`
      } catch (error) {
        // Update state with `error`
      }
    }
    execute();
  }, []); //cannot add dispatch (see log below)
  if (!(lastValue === dispatch)) {
    console.log('...setting dependency to dispatch will cause infinite loop');
  }
  return <pre>{JSON.stringify(data, undefined, 2)}</pre>;
};
```

I am unable to find the code on github (npm points to this repo but useAsyncDispatch is not defined in this repo). In node_modules/@commercetools-frontend/sdk/dist/commercetools-frontend-sdk.esm.js I can find the following:

```
function useAsyncDispatch() {
  var dispatch = useDispatch();
  return function (action) {
    return dispatch(action);
  };
}
```

This should be:

```
function useAsyncDispatch() {
  var dispatch = useDispatch();
  return useCallback(
    function (action) {
      return dispatch(action);
    },
    [dispatch]
  );
}
```

emmenko:
Fixes #2818

### DIFF
--- a/packages/sdk/src/hooks/use-async-dispatch.ts
+++ b/packages/sdk/src/hooks/use-async-dispatch.ts
@@ -1,13 +1,16 @@
 import type { Dispatch } from 'redux';
+import { useDispatch } from 'react-redux';
 import type { TSdkAction } from '../types';
 
-import { useDispatch } from 'react-redux';
+type TAsyncDispatch<Action extends TSdkAction, Data> = (
+  action: Action
+) => Promise<Data>;
 
 // Wraps `dispatch` and cast the return type to a Promise, as the middleware
 // returns a Promise.
 function useAsyncDispatch<Action extends TSdkAction, Data>() {
   const dispatch = useDispatch<Dispatch<Action>>();
-  return (action: Action) => dispatch(action) as unknown as Promise<Data>;
+  return dispatch as unknown as TAsyncDispatch<Action, Data>;
 }
 
 export default useAsyncDispatch;
--- a/.changeset/strong-kangaroos-itch.md
+++ b/.changeset/strong-kangaroos-itch.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/sdk': patch
+---
+
+Fix `useAsyncDispatch` to return a fixed reference


packages/sdk/src/hooks/use-async-dispatch.ts
```diff
@@ -1,13 +1,16 @@
...
-import { useDispatch } from 'react-redux';
+type TAsyncDispatch<Action extends TSdkAction, Data> = (
+  action: Action
+) => Promise<Data>;
 
 // Wraps `dispatch` and cast the return type to a Promise, as the middleware
 // returns a Promise.
 function useAsyncDispatch<Action extends TSdkAction, Data>() {
   const dispatch = useDispatch<Dispatch<Action>>();
-  return (action: Action) => dispatch(action) as unknown as Promise<Data>;
```
emmenko, line 10: I think this function was only used for defining the type casting, as it doesn't do anything specific.

We can remove it and simply return the `dispatch` function and to the type casting differently.

kark (approved):
Thanks, makes sense!

tdeekens (approved):

Rhotimee (approved):
💯

PR #2820, opened 2022-09-21
refactor(jest): improve output for console logs warnings

emmenko:
Some time ago I stumbled upon this library [jest-fail-on-console](https://www.npmjs.com/package/jest-fail-on-console) and I tried it out in our test-data repository (https://github.com/commercetools/test-data/pull/148).

Essentially it does more or less what we were already doing but I noticed the output is more readable. Therefore I tried to implement it in out jest-preset and (partially) replace our own custom implementation.

**Before**

Screenshot 2022-09-21 at 09 35 52

**After**

Screenshot 2022-09-21 at 09 36 27

### DIFF
--- a/packages/jest-preset-mc-app/fail-on-console.js
+++ b/packages/jest-preset-mc-app/fail-on-console.js
@@ -1,0 +1,123 @@
+// This module is based on the package `jest-fail-on-console` but with some custom adjustments.
+const util = require('util');
+const colors = require('colors/safe');
+
+const defaultErrorMessage = (methodName, bold) =>
+  `Expected test not to call ${bold(`console.${methodName}()`)}.\n\n` +
+  `If the ${methodName} is expected, test for it explicitly by mocking it out using ${bold(
+    'jest.spyOn'
+  )}(console, '${methodName}').mockImplementation() and test that the warning occurs.`;
+
+const failOnConsole = ({
+  errorMessage = defaultErrorMessage,
+  shouldFailOnAssert = false,
+  shouldFailOnDebug = false,
+  shouldFailOnLog = false,
+  shouldFailOnInfo = false,
+  shouldFailOnWarn = true,
+  shouldFailOnError = true,
+  skipTest,
+  silenceMessage,
+  logButNotThrowMessage,
+} = {}) => {
+  const flushUnexpectedConsoleCalls = (
+    methodName,
+    unexpectedConsoleCallStacks
+  ) => {
+    if (unexpectedConsoleCallStacks.length > 0) {
+      const messages = unexpectedConsoleCallStacks.map(([stack, message]) => {
+        const stackLines = stack.split('\n');
+        return (
+          `${colors.red(message)}\n` +
+          `${stackLines
+            .map((line, index) => {
+              if (index === stackLines.length - 1) {
+                return colors.white(line);
+              }
+              return colors.gray(line);
+            })
+            .join('\n')}`
+        );
+      });
+
+      const message = errorMessage(methodName, colors.bold);
+
+      throw new Error(`${message}\n\n${messages.join('\n\n')}`);
+    }
+  };
+
+  const patchConsoleMethod = (methodName) => {
+    let originalMethod = console[methodName];
+
+    const unexpectedConsoleCallStacks = [];
+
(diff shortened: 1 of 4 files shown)


packages/jest-preset-mc-app/fail-on-console.js
```diff
@@ -0,0 +1,123 @@
...
+      const message = util.format(format, ...args);
+
+      if (silenceMessage && silenceMessage(message, methodName)) {
+        return;
+      }
+
+      if (logButNotThrowMessage && logButNotThrowMessage(message, methodName)) {
+        originalMethod(format, ...args);
+        return;
+      }
```
emmenko, line 64: This is the only functionality (that we had in our implementation) "missing" from the original plugin. Therefore, I opted to simply copy over the implementation and adjust it accordingly.

packages/jest-preset-mc-app/setup-test-framework.js
```diff
@@ -23,7 +23,66 @@ require('@formatjs/intl-pluralrules/locale-data/fr');
...
+    messages,
+    jestConfig.notThrowWarnings
+  );
+  const additionallyNonThrowing = hasMatchingRegexForMessage(
+    messages,
+    additionalNonThrowingWarnings
+  );
+
+  return notThrowingByJestConfig || additionallyNonThrowing;
+}
```
emmenko, line 65: This is the same as before, just moved into the `setup-test-framework` file.

packages/jest-preset-mc-app/setup-test-framework.js
```diff
@@ -23,7 +23,66 @@ require('@formatjs/intl-pluralrules/locale-data/fr');
...
+  silenceMessage: (message) => {
+    if (!process.env.CI) {
+      return false;
+    }
+    return shouldSilenceWarnings(message);
+  },
+  logButNotThrowMessage: (message) => {
+    return shouldNotThrowWarnings(message);
+  },
+});
```
emmenko, line 81: This is how we configure the logger

packages/jest-preset-mc-app/setup-test-framework.js
```diff
@@ -23,7 +23,66 @@ require('@formatjs/intl-pluralrules/locale-data/fr');
...
+
+global.console.config = {};
+
+global.console.config.addSilencedWarning = (rexExp) =>
+  additionalSilencedWarnings.push(rexExp);
+global.console.config.addNonThrowingWarning = (rexExp) =>
+  additionalNonThrowingWarnings.push(rexExp);
```
emmenko, line 88: Also same as before.

packages/jest-preset-mc-app/fail-on-console.js
```diff
@@ -0,0 +1,123 @@
...
+      const testName = currentTestState.currentTestName;
+      const testPath = currentTestState.testPath;
+
+      if (skipTest && skipTest({ testName, testPath })) return true;
+
+      return false;
+    };
+    const shouldSkipTest = canSkipTest();
+
+    beforeEach(() => {
```
tdeekens, line 100: Can you share how this skip test functionality would work?

emmenko: We don't use that but I kept it because it's in the original library. https://github.com/ValentinH/jest-fail-on-console#skiptest

packages/jest-preset-mc-app/fail-on-console.js
ahmehri: How much is this custom functionality is important? Any chance we can drop it and simply use the original plugin as you did in `test-data`?

tdeekens (approved):
👍🏼

packages/jest-preset-mc-app/fail-on-console.js
emmenko: This replaces this functionality:

https://github.com/commercetools/merchant-center-application-kit/blob/237472109fbb27898b3b02e7b0c2a269a7ac95ca/packages/jest-preset-mc-app/setup-tests.js#L71-L73

If we want we can remove it, but it would probably be a breaking change. We could include the change once we support Jest v29 maybe.

ahmehri (approved):

kark (approved):

PR #2822, opened 2022-09-21
feat(cli): Use additional permissions with config-sync command

Rhotimee:
#### Summary
Update the CLI to use additional permissions

### DIFF
--- a/packages/application-config/src/formatters.ts
+++ b/packages/application-config/src/formatters.ts
@@ -120,4 +120,5 @@ function entryPointUriPathToPermissionKeys<PermissionName extends string>(
 export {
   entryPointUriPathToResourceAccesses,
   entryPointUriPathToPermissionKeys,
+  formatEntryPointUriPathToResourceAccessKey,
 };
--- a/packages/application-config/src/formatters.ts
+++ b/packages/application-config/src/formatters.ts
@@ -1,5 +1,4 @@
 import upperFirst from 'lodash/upperFirst';
-import type { JSONSchemaForCustomApplicationConfigurationFiles } from './schema';
 
 type TImplicitCustomApplicationResourceAccesses<
   PermissionName extends string = ''
@@ -118,58 +117,8 @@ function entryPointUriPathToPermissionKeys<PermissionName extends string>(
   );
 }
 
-type TResourceAccessPermissionName<PermissionName extends string = ''> =
-  | `view${Capitalize<PermissionName>}`
-  | `manage${Capitalize<PermissionName>}`;
-
-const getPermissions = (
-  appConfig: JSONSchemaForCustomApplicationConfigurationFiles
-) => {
-  const additionalResourceAccessKeyToOauthScopeMap = (
-    appConfig.additionalOAuthScopes || []
-  ).reduce((previousOauthScope, { name, view, manage }) => {
-    const formattedResourceKey =
-      formatEntryPointUriPathToResourceAccessKey(name);
-    return {
-      ...previousOauthScope,
-      [`view${formattedResourceKey}`]: view,
-      [`manage${formattedResourceKey}`]: manage,
-    };
-  }, {} as Record<TResourceAccessPermissionName<PermissionName>, string[]>);
-
-  const additionalPermissionNames =
-    appConfig.additionalOAuthScopes?.map(({ name }) => name) || [];
-
-  const permissionKeys = entryPointUriPathToResourceAccesses(
-    appConfig.entryPointUriPath,
-    additionalPermissionNames
-  );
-
-  const additionalPermissions = (
-    Object.keys(
-      additionalResourceAccessKeyToOauthScopeMap || []
-    ) as TResourceAccessPermissionName<PermissionName>[]
-  ).map((additionalResourceAccessKey) => ({
-    name: permissionKeys[additionalResourceAccessKey],
-    oAuthScopes:
-      additionalResourceAccessKeyToOauthScopeMap[additionalResourceAccessKey],
-  }));
-
-  return [
-    {
-      name: permissionKeys.view,
-      oAuthScopes: appConfig.oAuthScopes.view,
-    },
-    {
-      name: permissionKeys.manage,
-      oAuthScopes: appConfig.oAuthScopes.manage,
-    },
-    ...additionalPermissions,
-  ];
-};
-
 export {
   entryPointUriPathToResourceAccesses,
   entryPointUriPathToPermissionKeys,
-  getPermissions,
+  formatEntryPointUriPathToResourceAccessKey,
 };
--- a/packages/application-config/src/formatters.ts
+++ b/packages/application-config/src/formatters.ts
@@ -1,4 +1,5 @@
 import upperFirst from 'lodash/upperFirst';
+import type { JSONSchemaForCustomApplicationConfigurationFiles } from './schema';
 
 type TImplicitCustomApplicationResourceAccesses<
   PermissionName extends string = ''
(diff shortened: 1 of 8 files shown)


packages/application-config/src/transformers.ts
```diff
@@ -19,16 +22,49 @@ const computeUriPath = (uriPath: string, entryPointUriPath: string) => {
...
   validateEntryPointUriPath(appConfig);
   validateSubmenuLinks(appConfig);
 
+  const additionalResourceAccessKeyToOauthScopeMap = (
+    appConfig.additionalOAuthScopes || []
+  ).reduce(
+    (previousOauthScope, { name, view, manage }) => ({
+      ...previousOauthScope,
+      [`view${formatEntryPointUriPathToResourceAccessKey(name)}`]: view,
+      [`manage${formatEntryPointUriPathToResourceAccessKey(name)}`]: manage,
```
emmenko, line 45: Nit: instead of calling the function twice, extract the result into a variable.

packages/application-config/src/transformers.ts
```diff
@@ -19,16 +22,49 @@ const computeUriPath = (uriPath: string, entryPointUriPath: string) => {
...
+
+type TImplicitCustomApplicationResourceAccesses<
+  PermissionName extends string = ''
+> = Record<TResourceAccessPermissionName<PermissionName>, string[]>;
+
 function transformCustomApplicationConfigToData(
   appConfig: JSONSchemaForCustomApplicationConfigurationFiles
 ): CustomApplicationData {
   validateEntryPointUriPath(appConfig);
   validateSubmenuLinks(appConfig);
```
emmenko, line 37: We should validate a couple more things:
* The additional permission names should be validated the same as the entry point.
* There should be no duplicated permission names (similar to the sub-links).

packages/application-config/src/transformers.ts
```diff
@@ -19,16 +22,49 @@ const computeUriPath = (uriPath: string, entryPointUriPath: string) => {
...
 
+  const additionalPermissions = (
+    Object.keys(
+      additionalResourceAccessKeyToOauthScopeMap || []
+    ) as TResourceAccessPermissionName<PermissionName>[]
+  ).map((additionalResourceAccessKey) => ({
+    name: permissionKeys[additionalResourceAccessKey],
+    oAuthScopes:
+      additionalResourceAccessKeyToOauthScopeMap[additionalResourceAccessKey],
+  }));
```
emmenko, line 66: Nit: to not pollute much the `transformCustomApplicationConfigToData` function, maybe extract the transformation logic for the additional permissions into a separate function?

emmenko (commented):
Please also extend the tests for the CLI config diffs to include additional permissions and make sure the diffing logic still works.

Thanks, otherwise looks pretty straight forward 👌

packages/application-config/src/transformers.ts
Rhotimee: Updated [11ceeee](https://github.com/commercetools/merchant-center-application-kit/pull/2822/commits/11ceeee3115d57516d36fbd6193cacca1758ef99)

packages/application-config/src/transformers.ts
Rhotimee: Updated [11ceeee](https://github.com/commercetools/merchant-center-application-kit/pull/2822/commits/11ceeee3115d57516d36fbd6193cacca1758ef99)

packages/application-config/src/formatters.ts
```diff
@@ -117,7 +118,58 @@ function entryPointUriPathToPermissionKeys<PermissionName extends string>(
   );
 }
 
+type TResourceAccessPermissionName<PermissionName extends string = ''> =
+  | `view${Capitalize<PermissionName>}`
+  | `manage${Capitalize<PermissionName>}`;
+
+const getPermissions = (
```
emmenko, line 125: Nit: since we're only mapping some data, I would keep this function in the `transformers.ts` file.

packages/application-config/src/formatters.ts
```diff
@@ -117,7 +118,58 @@ function entryPointUriPathToPermissionKeys<PermissionName extends string>(
...
+    appConfig.additionalOAuthScopes || []
+  ).reduce((previousOauthScope, { name, view, manage }) => {
+    const formattedResourceKey =
+      formatEntryPointUriPathToResourceAccessKey(name);
+    return {
+      ...previousOauthScope,
+      [`view${formattedResourceKey}`]: view,
+      [`manage${formattedResourceKey}`]: manage,
+    };
+  }, {} as Record<TResourceAccessPermissionName<PermissionName>, string[]>);
```
emmenko, line 138: This `PermissionName` type doesn't seem to be defined anywhere. In fact, it seems it's coming from the (global) DOM types 🤔

image

Actually I also don't think we need the `TResourceAccessPermissionName` type and we can keep the record simple, otherwise we need the function overloading and all the related shenanigans.

```suggestion
  }, {} as Record<string, string[]>);
```

We can then cast the return type of `entryPointUriPathToResourceAccesses` to `Record<string, string>`.

packages/application-config/src/validations.ts
```diff
@@ -70,3 +70,17 @@ export const validateSubmenuLinks = (
     uriPathSet.add(uriPath);
   });
 };
+
+export const validateAdditionalOAuthScopes = (
+  config: JSONSchemaForCustomApplicationConfigurationFiles
+) => {
+  const additionalPermissionNames = new Set();
+  config.additionalOAuthScopes?.forEach(({ name }) => {
+    if (additionalPermissionNames.has(name)) {
```
emmenko, line 79: We should also validate that the name matches the entry point regex.

packages/application-config/src/formatters.ts
Rhotimee: Updated [fbd2537](https://github.com/commercetools/merchant-center-application-kit/pull/2822/commits/fbd25376f21bfc7d81da512270a7e3dded6ea308)

packages/application-config/src/formatters.ts
Rhotimee: Updated [fbd2537](https://github.com/commercetools/merchant-center-application-kit/pull/2822/commits/fbd25376f21bfc7d81da512270a7e3dded6ea308)

packages/application-config/src/validations.ts
Rhotimee: Updated [5a7680f](https://github.com/commercetools/merchant-center-application-kit/pull/2822/commits/5a7680f81bd8b3aa964ec4be22e6f7341f825ec9)

emmenko (approved):
💯 Awesome, thanks!

packages/application-config/src/validations.ts
```diff
@@ -70,3 +70,22 @@ export const validateSubmenuLinks = (
...
+      throw new Error(
+        'Duplicate additional permission. Every additional permission must have a unique name'
```
CarlosCortizasCT, line 81: nit: could we maybe include the permission name in the error message to make it easier to developers know about the problem?
```suggestion
        'Duplicate additional permission "${name}". Every additional permission must have a unique name'
```

CarlosCortizasCT (approved):
LGTM 👌

packages/application-config/src/validations.ts
Rhotimee: Updated [5f45a61](https://github.com/commercetools/merchant-center-application-kit/pull/2822/commits/5f45a61c753574fbd2509f67a18222edaf3b0670), Thanks.

PR #2823, opened 2022-09-22
feat(cypress): use `cy.session` and add new command `cy.loginToMerchantCenter` 🚀

emmenko:
See changeset for more info.

### DIFF
--- a/.changeset/bright-peaches-fetch.md
+++ b/.changeset/bright-peaches-fetch.md
@@ -2,17 +2,14 @@
 '@commercetools-frontend/cypress': minor
 ---
 
-Add new command `cy.loginByForm`.
+Implement [`cy.session`](https://www.cypress.io/blog/2021/08/04/authenticate-faster-in-tests-cy-session-command/) in the login commands, which caches and restores the user session between test runs.
+This ultimately results in subsequent tests to run much faster (by restoring the previous session) and makes the test behave as if the user is already authenticated.
 
-This command uses the [Cypress functionality to test across origins](https://cypress.io/blog/2022/04/25/cypress-9-6-0-easily-test-multi-domain-workflows-with-cy-origin/), available from Cypress `v9.6.0`.
+Make sure to have the option `experimentalSessionAndOrigin` turned on (in your Cypress config) to enable that.
 
-For Custom Applications it allows to authenticate the user via the login form (instead of authenticating the user in the background using the API), by following the redirect to the login page (different origin), filling out the form and redirecting back to `localhost`.
+Additionally, there is a new command `cy.loginToMerchantCenter` as a replacement of `cy.loginByOidc` (marked as _deprecated_).
 
-In addition to that, both commands `cy.loginByForm` and `cy.loginByOidc` implement [`cy.session`](https://www.cypress.io/blog/2021/08/04/authenticate-faster-in-tests-cy-session-command/), which caches and restores the user session between test runs.
-This ultimately results in subsequent tests to run much faster as the test starts as if the user is already authenticated.
+The `cy.loginToMerchantCenter` command detects whether the application is running on localhost or on production and choses the appropriate login mechanism.
 
-Make sure to have the option `experimentalSessionAndOrigin` turned on to benefit from using `cy.session` and `cy.origin`.
-If the option is turned off:
-
-- The `cy.loginByOidc` command will log a warning and fall back to not using `cy.session`.
-- The `cy.loginByForm` command will fail to run.
+- When the application runs locally, the same mechanism used in the `cy.loginByOidc` is used.
+- When the application runs on production, a normal login flow is used where the user credentials are typed into the login form.
--- a/.changeset/bright-peaches-fetch.md
+++ b/.changeset/bright-peaches-fetch.md
@@ -9,7 +9,7 @@ Make sure to have the option `experimentalSessionAndOrigin` turned on (in your C
 
 Additionally, there is a new command `cy.loginToMerchantCenter` as a replacement of `cy.loginByOidc` (marked as _deprecated_).
 
-The `cy.loginToMerchantCenter` command detects whether the application is running on localhost or on production and choses the appropriate login mechanism.
+The `cy.loginToMerchantCenter` command detects whether the application is running on localhost or on production and chooses the appropriate login mechanism.
 
 - When the application runs locally, the same mechanism used in the `cy.loginByOidc` is used.
 - When the application runs on production, a normal login flow is used where the user credentials are typed into the login form.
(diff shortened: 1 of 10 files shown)


ahmehri (approved):
💯

emmenko:
So I have good and bad news.

Let's start from the bad news:
* The e2e tests on CI were failing inconsistently between the Playground app and the Starter apps. I spent some time trying to debug the issues without much success only then to discover that the `cy.origin` logic was only working in the Playground app but not in the Starter apps. The only difference between them is that the Playground app uses Vite and the Starter apps use Webpack. For some unknown reason the `cy.origin` logic works correctly when the redirect is initiated by the Vite app but not for the Webpack apps.
* As a result, using `cy.origin` is out of question at the moment.

On the other hand, here's the good news:
* The improvements for caching/restoring the session via `cy.session` are still valid and work for both apps. I tested this already in our internal apps and it also works as expected. So that's already a big improvement.
* The `cy.loginByOidc` command was only used for applications running on localhost. However, for production apps the logic flow (via login form) can be used because there is no cross-origin redirect in-between. Therefore, we now have a new command `cy.loginToMerchantCenter` that automatically choses the better login flow based on the `baseUrl` value (localhost or not). This will simplify our setup for our internal apps but also allow anyone to easily test the login for applications running on production.

.changeset/bright-peaches-fetch.md
```diff
@@ -0,0 +1,15 @@
...
+---
+
+Implement [`cy.session`](https://www.cypress.io/blog/2021/08/04/authenticate-faster-in-tests-cy-session-command/) in the login commands, which caches and restores the user session between test runs.
+This ultimately results in subsequent tests to run much faster (by restoring the previous session) and makes the test behave as if the user is already authenticated.
+
+Make sure to have the option `experimentalSessionAndOrigin` turned on (in your Cypress config) to enable that.
+
+Additionally, there is a new command `cy.loginToMerchantCenter` as a replacement of `cy.loginByOidc` (marked as _deprecated_).
+
+The `cy.loginToMerchantCenter` command detects whether the application is running on localhost or on production and choses the appropriate login mechanism.
```
ahmehri, line 12:
```suggestion
The `cy.loginToMerchantCenter` command detects whether the application is running on localhost or on production and chooses the appropriate login mechanism.
```

packages/cypress/add-commands/index.d.ts
```diff
@@ -1,13 +1,29 @@
 declare namespace Cypress {
   interface Chainable<Subject> {
+    /**
+     * Log into the Custom Application.
+     * The command detects whether the application is running on localhost or on production
+     * and choses the appropriate login mechanism.
```
ahmehri, line 6:
```suggestion
     * and chooses the appropriate login mechanism.
```

packages/cypress/src/add-commands/login.ts
```diff
@@ -0,0 +1,285 @@
...
+        cy.url().should('include', commandOptions.initialRoute);
+      }
+    });
+  });
+}
+
+// Utilities
+
+function validateUserSession(mcApiUrl: string) {
+  // Validate that the session in valid by sending the `amILoggedIn` request.
```
ahmehri, line 260:
```suggestion
  // Validate that the session is valid by sending the `amILoggedIn` request.
```

ddouglasz (approved):
Looks good!

tdeekens (approved):

packages/cypress/src/add-commands/login.ts
```diff
@@ -0,0 +1,290 @@
...
+   */
+  login?: LoginCredentials;
+};
+// Alias for backwards compatibility
+export type CommandLoginByOidcOptions = CommandLoginOptions;
+
+function loginByForm(commandOptions: CommandLoginOptions) {
+  if (isLocalhost()) {
+    throw new Error(
+      `At the moment, the "loginByForm" command only works when testing a Merchant Center production URL. Testing an application running on localhost is not supported due to issues with "cy.origin".`
```
CarlosCortizasCT, line 62: **nit**: the message seems a little confusing to me as it says we can't test an application running locally.
```suggestion
      `At the moment, the "loginByForm" command only works when testing a Merchant Center production URL. Using form login in an application running on localhost is not supported due to issues with "cy.origin".`
```

packages/cypress/src/add-commands/login.ts
```diff
@@ -0,0 +1,290 @@
...
+
+    // Log loaded application config for debugging purposes.
+    Cypress.log({
+      displayName: 'task',
+      name: 'customApplicationConfig',
+      message: appConfig,
+    });
+
+    const userCredentials = commandOptions.login ?? {
+      email: Cypress.env('LOGIN_EMAIL') ?? Cypress.env('LOGIN_USER'),
```
CarlosCortizasCT, line 90: **nit**: If user has something like this in his env file:
```
LOGIN_EMAIL=""
LOGIN_USER="sam@nowhere.com"
```
I would expect `LOGIN_USER` to be used but `LOGIN_EMAIL`.

This might happen in there are several `env` files (local, global, ...) and user is trying to override values.

Perhaps we can update this line to be like:
```suggestion
      email: Cypress.env('LOGIN_EMAIL') || Cypress.env('LOGIN_USER'),
```

packages/cypress/src/add-commands/login.ts
```diff
@@ -0,0 +1,290 @@
...
+      cy.session(sessionKey, authCallback, {
+        validate() {
+          validateUserSession(appConfig.mcApiUrl);
+        },
+      });
+    } else {
+      cy.log(
+        `We recommend turning on the flag "experimentalSessionAndOrigin" to be able to use "cy.session" and thus reduce the time to log in between tests.`
+      );
+      authCallback();
```
CarlosCortizasCT, line 132: **nit**: I was wondering why we don't validate the user session in this code branch. Is it for performance reasons? (we don't want to validate the session on every login of every test when we're not leveraging session caching)

CarlosCortizasCT (approved):
LGTM 👌

packages/cypress/src/add-commands/login.ts
emmenko: Good point, thanks for noticing!

packages/cypress/src/add-commands/login.ts
emmenko: You mean to manually call `validateUserSession` part?

I haven't thought about it, good point. In theory validating the session is not strictly required. I assume the tests will fail anyway if there is no session.

Do you think it would make sense to do it here as well?

CarlosCortizasCT: If tests will fail anyway when login did not succeed, I guess I would remove that validation from the other code branch.
What was confusing for me is that we are only running the validating when using the cached session. I see this is done to ensure the reused session is still valid but it seems the test will fail in that case.

PR #2831, opened 2022-09-28
CON-1014: Add missing countries

bradd123:
#### Summary

There is a mismatch between the countries available on the backend and frontend. We are working on a feature where we use countries as a filter for pricing. For this, we need to add some missing countries.

#### Description

These countries are here for legacy reasons. Removing these countries on the backend is a breaking change. To continue forward with the feature, we decided to add these missing countries. There are no changes to the existing data. We are just adding extra countries to the list.


packages/l10n/scripts/missing-countries.js
```diff
@@ -0,0 +1,82 @@
+const missingCountries = {
```
emmenko, line 1: Let's add some comments here to explain a bit the context.

For instance, that these are legacy countries and do not exist anymore.

packages/l10n/scripts/generate-l10n-data.js
```diff
@@ -69,6 +89,7 @@ const extractCountryDataForLocale = (locale) => {
   const countryNames = {
     ...cldr.extractTerritoryDisplayNames(locale),
     ...getNorthernIrelandCountryCode(locale),
```
emmenko, line 91: Can you integrate this with the `missing-countries.js`?

emmenko (commented):
Wondering if it would make sense to expose (from the `l10n` package) a constant with a list of the "legacy" countries.

Consumers (like the MC) can use that list to have some special logic around these countries, for example:
* Preventing to select a country that is in that list.
* Showing the country as "legacy", so users could actively change that.

packages/l10n/scripts/missing-countries.js
```diff
@@ -0,0 +1,82 @@
...
+    FX: '法国，大都会',
+    NT: '中立区',
+    SF: '芬兰',
+    SU: '苏联',
+    TP: '东帝汶',
+    UK: '英国',
+    YU: '南斯拉夫',
+    ZR: '扎伊尔',
+  },
+  ja: {
```
ByronDWall, line 67: If I recall correctly, removed Japanese translation support, so this object is probably not necessary anymore

ByronDWall (commented):
Code-wise, aside from not needing the Japanese translations, I think this looks fine.

I have to ask, what is the use-case for displaying countries that have not existed since before the product was originally built (USSR, East Timor, Burma, Yugoslavia, etc)?

I would think that in at least some cases, the history behind why these countries do not currently exist would be a good reason not to display them?

bradd123:
We decided not add any countries to this repo and leave them as it is for now. So, I am closing this PR.

PR #2830, opened 2022-09-28
feat(cypress): enable caching session across spec by default

emmenko:
Follow up of #2823 now that [Cypress v10.9.0](https://docs.cypress.io/guides/references/changelog#&#8203;10-9-0) has added the option `cacheAcrossSpecs` 🥳

PR #2832, opened 2022-09-30
fix(cypress): backwards compat for version < 10.9.0

emmenko:
Leftover of #2830 

image

### DIFF
--- a/packages/cypress/src/add-commands/login.ts
+++ b/packages/cypress/src/add-commands/login.ts
@@ -1,4 +1,5 @@
 import { v4 as uuidv4 } from 'uuid';
+import semver from 'semver';
 import { buildOidcScope } from '@commercetools-frontend/application-shell/ssr';
 import type { ApplicationRuntimeConfig } from '@commercetools-frontend/application-config';
 import { STORAGE_KEYS, OIDC_RESPONSE_TYPES } from '../constants';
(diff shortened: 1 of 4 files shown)


packages/cypress/src/add-commands/login.ts
```diff
@@ -124,12 +130,18 @@ function loginByForm(commandOptions: CommandLoginOptions) {
     // For backwards compatibility.
     if (Cypress.config('experimentalSessionAndOrigin')) {
       // https://www.cypress.io/blog/2021/08/04/authenticate-faster-in-tests-cy-session-command/
-      cy.session(sessionKey, authCallback, {
-        cacheAcrossSpecs:
-          typeof commandOptions.disableCacheAcrossSpecs === 'boolean'
-            ? !commandOptions.disableCacheAcrossSpecs
-            : true,
-      });
+      cy.session(
```
CarlosCortizasCT, line 133: I was thinking if we should also check the version for using the `cy.session` command.
It seems our `cypress` package has support from Cypress [v8.0](https://github.com/commercetools/merchant-center-application-kit/blob/main/packages/cypress/package.json#L49) and this command was introduced in [v8.2](https://docs.cypress.io/guides/references/changelog#8-2-0).

CarlosCortizasCT (approved):
👌

packages/cypress/src/add-commands/login.ts
tdeekens: Good point. In an ideal case yes. In a slightly pragmatic case maybe not. All internal repos use Cypress v10. I am not aware of a customer actually using exactly 8.0.0 (or the package at all). So maybe close the eye over this (at least from my view).

tdeekens (approved):

packages/cypress/src/add-commands/login.ts
emmenko: That's already covered by the check `Cypress.config('experimentalSessionAndOrigin')`.

In v8 this flag was called differently.

PR #2834, opened 2022-09-30
fix(application-config): enable proper type inference in permission keys and resource accesses formatters

kark:
#### Summary

Currently the output type of  e.g. `entryPointUriPathToResourceAccesses` function does not work as expected.
Screenshot 2022-09-30 at 11 16 24

The goal of this PR is to fix that. 

I had to refactor a little the `Split` utility type taken from the `type-fest` library to work well in our case.

Refactoring [this test](https://github.com/commercetools/merchant-center-application-kit/blob/af9f6681dcd5d28ba036c5fd07140cbf904a3c52/packages/application-config/src/formatters.spec.ts#L22:L55) to enable static type checking (without any type assertion) turned out to be a bigger challenge than this fix 😕, so I left this test as is.

To play with this PR you might want to use [this playground](https://codesandbox.io/s/typescript-playground-export-forked-u297ue)

### DIFF
--- a/packages/application-config/src/formatters.spec.ts
+++ b/packages/application-config/src/formatters.spec.ts
@@ -5,7 +5,7 @@ describe.each`
   ${'avengers'}     | ${'Avengers'}
   ${'the-avengers'} | ${'TheAvengers'}
   ${'the_avengers'} | ${'The_Avengers'}
-  ${'avengers-01'}  | ${'Avengers01'}
+  ${'avengers-01'}  | ${'Avengers/01'}
   ${'avengers_01'}  | ${'Avengers_01'}
 `(
   'formatting the entryPointUriPath "$entryPointUriPath" to a resource access key "$formattedResourceAccessKey"',
@@ -26,7 +26,7 @@ describe.each`
   ${'avengers'}     | ${'Avengers'}              | ${['books']}              | ${'AvengersBooks'}
   ${'the-avengers'} | ${'TheAvengers'}           | ${['the-books']}          | ${'TheAvengersTheBooks'}
   ${'the_avengers'} | ${'The_Avengers'}          | ${['the_books']}          | ${'The_AvengersThe_Books'}
-  ${'avengers-01'}  | ${'Avengers01'}            | ${['books-01']}           | ${'Avengers01Books01'}
+  ${'avengers-01'}  | ${'Avengers/01'}           | ${['books-01']}           | ${'Avengers/01Books01'}
   ${'avengers_01'}  | ${'Avengers_01'}           | ${['books_01']}           | ${'Avengers_01Books_01'}
 `(
   'formatting the entryPointUriPath "$entryPointUriPath" with additional permission names "$additionalPermissionNames" to a resource access key "$formattedResourceAccessKey" and "$formattedResourceAccessKeyAdditionalNames"',
--- a/packages/application-config/src/formatters.spec.ts
+++ b/packages/application-config/src/formatters.spec.ts
@@ -22,18 +22,18 @@ describe.each`
 );
 
 describe.each`
-  entryPointUriPath | formattedResourceAccessKey | additionalPermissionNames | formattedResourceAccessKeyAdditionalNames
-  ${'avengers'}     | ${'Avengers'}              | ${['books']}              | ${'AvengersBooks'}
-  ${'the-avengers'} | ${'TheAvengers'}           | ${['the-books']}          | ${'TheAvengersTheBooks'}
-  ${'the_avengers'} | ${'The_Avengers'}          | ${['the-books']}          | ${'The_AvengersTheBooks'}
-  ${'avengers-01'}  | ${'Avengers/01'}           | ${['books-01']}           | ${'Avengers/01Books01'}
-  ${'avengers_01'}  | ${'Avengers_01'}           | ${['books-01']}           | ${'Avengers_01Books01'}
+  entryPointUriPath | formattedResourceAccessKey | permissionGroupNames | formattedResourceAccessKeyAdditionalNames
+  ${'avengers'}     | ${'Avengers'}              | ${['books']}         | ${'AvengersBooks'}
+  ${'the-avengers'} | ${'TheAvengers'}           | ${['the-books']}     | ${'TheAvengersTheBooks'}
+  ${'the_avengers'} | ${'The_Avengers'}          | ${['the-books']}     | ${'The_AvengersTheBooks'}
+  ${'avengers-01'}  | ${'Avengers/01'}           | ${['the-movies']}    | ${'Avengers/01TheMovies'}
+  ${'avengers_01'}  | ${'Avengers_01'}           | ${['the-movies']}    | ${'Avengers_01TheMovies'}
 `(
-  'formatting the entryPointUriPath "$entryPointUriPath" with additional permission names "$additionalPermissionNames" to a resource access key "$formattedResourceAccessKey" and "$formattedResourceAccessKeyAdditionalNames"',
+  'formatting the entryPointUriPath "$entryPointUriPath" with additional permission group names "$permissionGroupNames" to a resource access key "$formattedResourceAccessKey" and "$formattedResourceAccessKeyAdditionalNames"',
   ({
     entryPointUriPath,
     formattedResourceAccessKey,
-    additionalPermissionNames,
+    permissionGroupNames,
     formattedResourceAccessKeyAdditionalNames,
   }) => {
     it(`should format correctly`, () => {
@@ -44,7 +44,7 @@ describe.each`
       expect(
         entryPointUriPathToPermissionKeys(
           entryPointUriPath,
-          additionalPermissionNames
+          permissionGroupNames
         )
       ).toEqual({
         View: `View${formattedResourceAccessKey}`,
--- a/packages/application-config/src/formatters.spec.ts
+++ b/packages/application-config/src/formatters.spec.ts
@@ -25,7 +25,7 @@ describe.each`
   entryPointUriPath | formattedResourceAccessKey | additionalPermissionNames | formattedResourceAccessKeyAdditionalNames
   ${'avengers'}     | ${'Avengers'}              | ${['books']}              | ${'AvengersBooks'}
   ${'the-avengers'} | ${'TheAvengers'}           | ${['the-books']}          | ${'TheAvengersTheBooks'}
-  ${'the_avengers'} | ${'The_Avengers'}          | ${['the_books']}          | ${'The_AvengersThe_Books'}
+  ${'the_avengers'} | ${'The_Avengers'}          | ${['the_books']}          | ${'The_AvengersThe_books'}
   ${'avengers-01'}  | ${'Avengers/01'}           | ${['books-01']}           | ${'Avengers/01Books01'}
   ${'avengers_01'}  | ${'Avengers_01'}           | ${['books_01']}           | ${'Avengers_01Books_01'}
 `(
--- a/packages/application-config/src/formatters.spec.ts
+++ b/packages/application-config/src/formatters.spec.ts
@@ -25,9 +25,9 @@ describe.each`
   entryPointUriPath | formattedResourceAccessKey | additionalPermissionNames | formattedResourceAccessKeyAdditionalNames
   ${'avengers'}     | ${'Avengers'}              | ${['books']}              | ${'AvengersBooks'}
   ${'the-avengers'} | ${'TheAvengers'}           | ${['the-books']}          | ${'TheAvengersTheBooks'}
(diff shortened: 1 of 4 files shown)


packages/application-config/src/formatters.ts
```diff
@@ -44,15 +45,7 @@ const formatEntryPointUriPathToResourceAccessKey = (
...
     .split('-')
-    .map((word, i) => {
-      // Regex below checking if the character is numeric.
-      // If the word after the hyphen is numeric, replace the hyphen with a forward slash.
-      // If not, omit the hyphen and uppercase the first character
-      if (i > 0 && /^-?\d+$/.test(word[0])) {
-        return `/${word}`;
-      }
-      return upperFirst(word);
-    })
```
kark, line 55: This formatting of digits was removed (as discussed)

CarlosCortizasCT: Wouldn't this be a breaking change?
If we have an already installed application with the entry point `avengers-01`, the resources access keys used to create the permission resource name which would be stored in the database would be `viewAvengers/01` and `manageAvengers/01`.
After updating the `app-kit`, the permission resource name calculated in the application at runtime would be `viewAvengers01` and that would not match the one in the database.

Maybe I'm missing something, but I wanted to double check.

packages/application-config/src/types.ts
```diff
@@ -52,3 +52,86 @@ export type LoadingConfigOptions = {
...
+/**
+Convert a string literal to camel-case.
+This can be useful when, for example, converting some kebab-cased command-line flags or a snake-cased database result.
+
+source: https://github.com/sindresorhus/type-fest/blob/fedbc441a314c1f9f5f6225c993860d0886261da/source/camel-case.d.ts#L73
+
+@example
+```
+const someVariable: CamelCase<'foo-bar'> = 'fooBar';
+const anotherVariable: CamelCase<'foo_bar'> = 'foo_Bar';
```
CarlosCortizasCT, line 130: These would be types and not consts, right?
```suggestion
type someVariable = CamelCase<'foo-bar'> = 'fooBar';
type anotherVariable = CamelCase<'foo_bar'> = 'foo_Bar';
```

packages/application-config/src/types.ts
```diff
@@ -52,3 +52,86 @@ export type LoadingConfigOptions = {
...
+```
+*/
+export type CamelCase<K> = K extends string
```
CarlosCortizasCT, line 133: This is more advanced typescript than I can understand 😞 
Perhaps it would be great if you could share how this work in a presentation for us to learn about it. 🙏

packages/application-config/src/formatters.ts
kark: Hmm 🤔, you are right, thanks. It's possible I misunderstood something... 
I guess I can have another look if this digit formatting is somehow transferrable to types inference.

packages/application-config/src/types.ts
kark: Thanks, I think the example is ok. I other words it's:
```ts
type SomeVariable = CamelCase<'foo-bar'>;
const someVariable: SomeVariable = 'fooBar';

type AnotherVariable = CamelCase<'foo-bar'>;
const anotherVariable: AnotherVariable = 'foo_Bar';
```

CarlosCortizasCT: Ah, yes. Now I see.
Thanks.

emmenko (changes requested):

packages/application-config/src/types.ts
kark: Thanks Carlos, but you know that my input here was very limited, right? 😅 I guess we can think of some form of presentation

packages/application-config/src/formatters.ts
kark: Right, my bad, sorry for that

emmenko:
To clarify: at the moment we validate the additional permission names the same as the `entryPointUriPath`. However, due to the types inference issue we decided to restrict it a bit more, so that only hyphens `-` are allowed (no digits, no `_`).

I think this is a fair tradeoff for now, as I don't see how digits and `_` would be useful for permission names.

So the validation for these values should also be adjusted accordingly.

packages/application-config/src/types.ts
```diff
@@ -52,3 +52,107 @@ export type LoadingConfigOptions = {
...
+  ? [
+      Delimiter extends '_'
+        ? `${Head}_`
+        : Delimiter extends '-'
+        ? Tail extends ConvertibleToNumber<Tail>
+          ? `${Head}/`
+          : Head
+        : Head,
+      ...Split<Tail, Delimiter>
+    ]
```
kark, line 99: I gave it another shot and I managed to get it to a state where we are not forced to change the way we format permission names due to TS complexity. As far as I tested it, it seems that the corresponding transformation is done now in JS and in the type.
Screenshot 2022-10-04 at 08 46 07

emmenko: Thanks for looking into it! It does look promising even though we're stepping foot into "unknown/complex" TS territory here.

I would try to keep things as simple as possible as a first step, for instance only allowing to use `-` and for the value to be lowercase.

If we get more specific requirements from customers we can always update the validation as needed.

kark: Sorry, it took me a while to wrap up other things.
I'm not sure if I understand - so you wish to use the following function for formatting the permission name part?
```ts
/**
 * The function formats the `permissionName` to a resource access key.
 * It makes the first character of the string and the next character after a special character an uppercase.
 *
 * @example
 * - avengers --> Avengers
 * - the-avengers --> TheAvengers
 * - avengers-01 --> Avengers01 [ℹ️ mind the difference here compared to `formatEntryPointUriPathToResourceAccessKey`]
 */
const formatPermissionNameToResourceAccessKey = (permissionName: string) =>
  permissionName
    // Each word is split by a hyphen.
    .split('-')
    .map(upperFirst)
    .join('');
```
Or not even map with `upperFirst`?

kark: I refactored it a little to use this helper for the additional resource access part:
```ts
const formatPermissionNameToResourceAccessKey = (permissionName: string) =>
  permissionName
    // Each word is split by a hyphen.
    .split('-')
    .map(upperFirst)
    .join('');
```
I'm not really sure how to enforce
> for the value to be lowercase

as you suggested. Should each split word be mapped to lowercase before mapping with `upperFirst`?

packages/application-config/src/formatters.ts
```diff
@@ -55,6 +56,22 @@ const formatEntryPointUriPathToResourceAccessKey = (
...
+ *
+ * @example
+ * - avengers --> Avengers
+ * - the-avengers --> TheAvengers
+ * - avengers-01 --> Avengers01 [ℹ️ mind the difference here compared to `formatEntryPointUriPathToResourceAccessKey`] */
+const formatPermissionNameToResourceAccessKey = (permissionName: string) =>
+  permissionName
+    // Each word is split by a hyphen.
+    .split('-')
+    .map((word) => word.toLowerCase())
```
emmenko, line 71: I think we can omit this, this should be covered by the validations.

packages/application-config/src/formatters.ts
```diff
@@ -55,6 +56,22 @@ const formatEntryPointUriPathToResourceAccessKey = (
...
 
+/**
+ * The function formats the `permissionName` to a resource access key.
+ * It makes the first character of the string and the next character after a special character an uppercase.
+ *
+ * @example
+ * - avengers --> Avengers
+ * - the-avengers --> TheAvengers
+ * - avengers-01 --> Avengers01 [ℹ️ mind the difference here compared to `formatEntryPointUriPathToResourceAccessKey`] */
+const formatPermissionNameToResourceAccessKey = (permissionName: string) =>
```
emmenko, line 67:
```suggestion
const formatPermissionGroupNameToResourceAccessKey = (permissionName: string) =>
```

packages/application-config/src/formatters.spec.ts
```diff
@@ -23,8 +25,8 @@ describe.each`
   entryPointUriPath | formattedResourceAccessKey | additionalPermissionNames | formattedResourceAccessKeyAdditionalNames
   ${'avengers'}     | ${'Avengers'}              | ${['books']}              | ${'AvengersBooks'}
   ${'the-avengers'} | ${'TheAvengers'}           | ${['the-books']}          | ${'TheAvengersTheBooks'}
-  ${'the_avengers'} | ${'The_Avengers'}          | ${['the_books']}          | ${'The_AvengersThe_Books'}
-  ${'avengers-01'}  | ${'Avengers/01'}           | ${['books-01']}           | ${'Avengers/01Books/01'}
+  ${'the_avengers'} | ${'The_Avengers'}          | ${['the_books']}          | ${'The_AvengersThe_books'}
+  ${'avengers-01'}  | ${'Avengers/01'}           | ${['books-01']}           | ${'Avengers/01Books01'}
   ${'avengers_01'}  | ${'Avengers_01'}           | ${['books_01']}           | ${'Avengers_01Books_01'}
```
emmenko, line 30: I would only use examples with `-` for the additional groups, as we don't want to allow other chars.

packages/application-config/src/formatters.ts
```diff
@@ -77,7 +94,7 @@ function entryPointUriPathToResourceAccesses<PermissionName extends string>(
   const additionalResourceAccesses = (additionalPermissionNames ?? []).reduce(
     (resourceAccesses, permissionName) => {
       const additionalResourceAccessKey =
-        formatEntryPointUriPathToResourceAccessKey(permissionName);
+        formatPermissionNameToResourceAccessKey(permissionName);
```
emmenko, line 97:
```suggestion
        formatPermissionGroupNameToResourceAccessKey(permissionName);
```

packages/application-config/src/formatters.ts
```diff
@@ -55,6 +56,22 @@ const formatEntryPointUriPathToResourceAccessKey = (
...
     .join('');
 
+/**
+ * The function formats the `permissionName` to a resource access key.
+ * It makes the first character of the string and the next character after a special character an uppercase.
+ *
+ * @example
+ * - avengers --> Avengers
+ * - the-avengers --> TheAvengers
+ * - avengers-01 --> Avengers01 [ℹ️ mind the difference here compared to `formatEntryPointUriPathToResourceAccessKey`] */
```
emmenko, line 66:
```suggestion
 * The function formats the permission group name to a resource access key.
 * It makes the first character of the string and the next character after a special character (`-`) an uppercase.
 *
 * @example
 * - books --> Books
 * - the-books --> TheBooks
 */
```

emmenko (commented):
Looks good, thanks!

Just some minor adjustments based on the assumption that we only allow hypens `-` to be used.

packages/application-config/src/formatters.ts
kark: Thanks, I assume that if we want to change it here I need to refactor it everywhere to keep it consistent (`permissionName` -> `groupName`)

packages/application-config/src/formatters.ts
kark: Ah, got you :smile:

packages/application-config/src/formatters.ts
emmenko: Good point, maybe it's better yes so that we have less confusion to the already confused naming formats 😅

packages/application-config/src/formatters.spec.ts
```diff
@@ -23,9 +25,9 @@ describe.each`
   entryPointUriPath | formattedResourceAccessKey | additionalPermissionNames | formattedResourceAccessKeyAdditionalNames
   ${'avengers'}     | ${'Avengers'}              | ${['books']}              | ${'AvengersBooks'}
   ${'the-avengers'} | ${'TheAvengers'}           | ${['the-books']}          | ${'TheAvengersTheBooks'}
-  ${'the_avengers'} | ${'The_Avengers'}          | ${['the_books']}          | ${'The_AvengersThe_Books'}
-  ${'avengers-01'}  | ${'Avengers/01'}           | ${['books-01']}           | ${'Avengers/01Books/01'}
-  ${'avengers_01'}  | ${'Avengers_01'}           | ${['books_01']}           | ${'Avengers_01Books_01'}
+  ${'the_avengers'} | ${'The_Avengers'}          | ${['the-books']}          | ${'The_AvengersTheBooks'}
+  ${'avengers-01'}  | ${'Avengers/01'}           | ${['books-01']}           | ${'Avengers/01Books01'}
+  ${'avengers_01'}  | ${'Avengers_01'}           | ${['books-01']}           | ${'Avengers_01Books01'}
```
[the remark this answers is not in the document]
emmenko: Ok thanks. If users are requesting it, we can always consider allowing it as a follow up.

emmenko (approved):
💯

PR #2841, opened 2022-10-03
Expose user SSO info in application context

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=GZD_KB7)

#### Summary

For those users who logged in using SSO, expose some of their Identity Provider auth token claims.

#### Description

Currently we have the ability to get logged-in user information from the [application context](https://docs.commercetools.com/custom-applications/api-reference/commercetools-frontend-application-shell-connectors#user).
With this PR's changes, we augment that information for those users who use an Identity Provider to log in so we expose some of the data we get from the IdP.
The user object will have a new **idTokenUserInfo** _optional_ property with this signature:
```ts
{
    issuer: string;
    subject: string;
    audience: string;
    expirationTimestamp: number;
    creationTimestamp: number;
    email?: string | null | undefined;
    name?: string | null | undefined;
    additionalClaims: Record<string, unknown>;
  }
```


graphql-test-utils/graphql-models/user.js
```diff
@@ -21,6 +21,7 @@ const User = new Factory()
   .attr('launchdarklyTrackingTenant', 'gcp-eu')
   .attr('defaultProjectKey', () => null)
   .attr('businessRole', () => faker.name.jobDescriptor())
+  .attr('idTokenUserInfo', () => null)
```
CarlosCortizasCT, line 24: I was having this kind of error in the tests:
image

The info that I could find about this error suggest the `in-memory` apollo cache needs to write all the optional properties of the object it wants to store in the cache ([reference](https://stackoverflow.com/a/50167010))

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -51,6 +51,23 @@ type TApplicationContextDataFences = Partial<
   >
 >;
 type TApplicationContextEnvironment = ApplicationWindow['app'];
+type TApplicationContextUser = Pick<
+  TFetchedUser,
+  'id' | 'email' | 'firstName' | 'lastName' | 'businessRole' | 'projects'
```
CarlosCortizasCT, line 56: This `Pick` statement is not working but I wanted to force push to share it and ask for help since I don't understand the problem 🙏 .

kark: I haven't gone through the whole PR but I think the solution  here is:
```suggestion
type TApplicationContextUser = Pick<
  NonNullable<TFetchedUser>,
  'id' | 'email' | 'firstName' | 'lastName' | 'businessRole' | 'projects'
```

CarlosCortizasCT: Thanks @kark 🙇 

If I got it right, the problem was that `TFetchedUser` is coming from the `TFetchLoggedInUserQuery['user']` type and, over there, the `user` property can be null. So when using the `NonNullable` utility over the `TFetchedUser` we are stating that the type can't refer to a null value but only to an actual user object type.

Keep learning...

.changeset/chatty-flowers-poke.md
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools-frontend/application-shell': minor
+'@commercetools-frontend/application-shell-connectors': minor
+---
+
+Logged-in [user data](https://docs.commercetools.com/custom-applications/api-reference/commercetools-frontend-application-shell-connectors#user) that can be accessed from the [application context](https://docs.commercetools.com/custom-applications/api-reference/commercetools-frontend-application-shell-connectors#useapplicationcontext) can now optionally include new info for SSO logged-in users (Identity Provider OIDC token claims).
```
CarlosCortizasCT, line 6: Not sure if this is enough.
I would expect for us to have a follow-up task to update the application context's user [docs](https://docs.commercetools.com/custom-applications/api-reference/commercetools-frontend-application-shell-connectors#user). Maybe also a release note?

CarlosCortizasCT:
> e2e tests for starter templates run against production API so we need the changes in the backend to be deployed before continuing with this PR.

backend has already been deployed so we're good to go now in that regard.

packages/application-shell-connectors/src/components/application-context/application-context.spec.tsx
```diff
@@ -37,6 +37,19 @@ const createTestEnvironment = (
 });
 
 const testUser = UserMock.random().buildGraphql<TUserGraphql>();
+const testSSOUser = {
+  ...UserMock.random().buildGraphql<TUserGraphql>(),
+  idTokenUserInfo: {
+    iss: '',
+    sub: '',
+    aud: '',
```
emmenko, line 45: Nit: let's put some random values here for completeness (as these fields are always defined).

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -72,6 +89,33 @@ export const mapUserToApplicationContextUser = (user?: TFetchedUser) => {
...
+        '@commercetools-frontend/application-shell-connectors: Could not parse received user sso custom claims from server.',
+        (error as Error).message
+      );
+    }
+    applicationContextUser.idTokenUserInfo = {
+      issuer: user.idTokenUserInfo.iss,
+      subject: user.idTokenUserInfo.sub,
+      audience: user.idTokenUserInfo.aud,
+      expirationTimestamp: user.idTokenUserInfo.exp,
+      creationTimestamp: user.idTokenUserInfo.iat,
```
emmenko, line 111: Any particular reason for renaming these fields? I would expect to expose them as they are.

.changeset/chatty-flowers-poke.md
emmenko: Yes as a follow up we can document these new fields.

emmenko (commented):
Thanks, had a quick look. Overall looks good, just had one question.

packages/application-shell-connectors/src/components/application-context/application-context.tsx
CarlosCortizasCT: I thought that, since developers using this data would be frontend engineers, they (actually everyone) would be more comfortable with more meaningful names. I would use the abbreviations for internal usage but use something more clear in the user data model.

Perhaps I overthought it and it would be easier to just map the same attribute names we get from the IdP token.

Let me know if I should rename these props.

emmenko: I think for now I would keep the same names, as we're simply exposing the claims as-is.

Also, the idea with this new field `idTokenUserInfo` is to somehow align it with how you would expose these infos in the `/userinfo` endpoint.

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -51,6 +51,23 @@ type TApplicationContextDataFences = Partial<
   >
 >;
 type TApplicationContextEnvironment = ApplicationWindow['app'];
+type TApplicationContextUser = Pick<
+  NonNullable<TFetchedUser>,
+  'id' | 'email' | 'firstName' | 'lastName' | 'businessRole' | 'projects'
+> & {
+  locale: string;
+  timeZone: string;
+  idTokenUserInfo?: {
```
emmenko, line 60: Nit: we could also use the generated type `TIdTokenUserInfo` from `generated/mc.ts`. The only thing to change is the type of `additionalClaims` but I think it could be done using `Omit` and `&`.

packages/application-shell-connectors/src/components/application-context/application-context.spec.tsx
CarlosCortizasCT: Not sure what a good set of values would be, but I updated them in this commit: 0138b05933457a87daefeace91129884567af75a

packages/application-shell-connectors/src/components/application-context/application-context.tsx
CarlosCortizasCT: Updated in this commit: 0138b05933457a87daefeace91129884567af75a

packages/application-shell-connectors/src/components/application-context/application-context.tsx
CarlosCortizasCT: Updated in this commit: 0138b05933457a87daefeace91129884567af75a

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -51,6 +52,16 @@ type TApplicationContextDataFences = Partial<
...
 type TApplicationContextEnvironment = ApplicationWindow['app'];
+type TApplicationContextUser = Pick<
+  NonNullable<TFetchedUser>,
+  'id' | 'email' | 'firstName' | 'lastName' | 'businessRole' | 'projects'
+> & {
+  locale: string;
+  timeZone: string;
+  idTokenUserInfo?: Omit<TIdTokenUserInfo, 'additionalClaims'> & {
+    additionalClaims: Record<string, unknown>;
+  };
```
emmenko, line 63: 💯

emmenko (approved):
Top! Thanks 🤗

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -72,6 +83,27 @@ export const mapUserToApplicationContextUser = (user?: TFetchedUser) => {
...
+  // This property will only be populated when user has logged in using SSO
+  if (user.idTokenUserInfo) {
+    let additionalClaims: Record<string, unknown> = {};
+    try {
+      additionalClaims = JSON.parse(
+        user.idTokenUserInfo.additionalClaims || ''
+      );
+    } catch (error) {
+      console.warn(
+        '@commercetools-frontend/application-shell-connectors: Could not parse received user sso custom claims from server.',
```
tdeekens, line 96: Should/could we report this to Sentry? Just thinking out loud.

tdeekens (approved):
Nice work!

kark (approved):
Neat! 👏

packages/application-shell-connectors/src/components/application-context/application-context.tsx
CarlosCortizasCT: I guess we could use the [@commerctools-frontend/sentry](https://github.com/commercetools/merchant-center-application-kit/tree/main/packages/sentry) package to report the error.

My initial thought was to log it to the console so, if users experience this problem, they could see it and be aware of what's happening. Also, they could reach out to us and provide that error message.

Second thought: this error would only happen if we provide an invalid JSON from the MC backend, which I think is unlikely due to its implementation. But if it happens, it wouldn't be because of a user action but because of something wrong in our side.
Looking it that way, perhaps it may sense to report to Sentry.

@emmenko what do you think?

ddouglasz (approved):
Looks Good! 👏

Rhotimee (approved):
💯

packages/application-shell-connectors/src/components/application-context/application-context.tsx
emmenko: Sure why not. Thanks for suggesting!

CarlosCortizasCT: Updated in this commit: 0b705034fac39ae971d8db39d1e646f6151910c9

PR #2842, opened 2022-10-04
Fix date formatting in different locales

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=woibXHP)

#### Summary

`momentjs` locale used to format dates is not correctly managed.

#### Description

We had [this issue](https://github.com/commercetools/ui-kit/issues/2301) opened regarding a problem when formatting date using the `ui-kit`'s `DateInput` component.
We validated the component itself don't have any problem but the issue is regarding the locale internally used by component.

We could verify the problem by inspecting the _playground_ view we have for dates formats:

Screenshot 2022-10-04 at 09 40 34

We can see how all the different locales render the same date pattern.

The problem seems to be related in the `momentjs` versions we have in `appkit`. We are using two main dependencies:
* moment: version `^2.29.4`
* moment-timezone: version `^0.5.37`

When running the `yarn why moment` command, we could see we had two different versions of that dependency in the project. `moment-timezone` was using a different older one. So we were loading `moment` locales metadata files (using `AsyncLocaleData`, which is used by `ApplicationShell` component) in one `moment` dependency version, but consuming another version from `moment-timezone`, which didn't have access to locale metadata files loaded and it used always the american english locale by default.

After upgrading all `moment` dependencies (to version `2.29.4`) running the `yarn up -R moment` command, everything works again.

I took the opportunity to add a new test where we take a snapshot of the playground date formatter's view so we can notice if this issue happens again before reaching production.


playground/src/components/formatters-demo/formatters-demo.jsx
```diff
@@ -34,7 +34,7 @@ const DEMO_MONEY_PRICE = {
 
 function getSampleDate() {
   const now = new Date();
-  now.setDate(20);
+  now.setFullYear(2022, 8, 30);
```
CarlosCortizasCT, line 37: Since we will be taking snapshots, we need a fixed date so they always match.

CarlosCortizasCT:
Not sure about the changeset here because none of the public packages have been modified.

kark (approved):
Thanks for the fix and for going the extra mile with the vrt 💯 
Personally I also don't see a reason for adding a changeset.

Rhotimee (approved):
🙌🏽

emmenko (approved):
Thanks! 💯

PR #2844, opened 2022-10-07
feat(application-kit): implement `verificationStatus` field

jaikamat:
## Summary

This PR proposes a `verificationStatus` field following work [from this PR](https://github.com/commercetools/merchant-center-services/pull/2292) within `mc-services`.

## Description

One of the goals of the First Contact team is to improve email verification flow for users of the Merchant Center. We have identified that the `User` entity will require a field indicating verification status, which will be eventually implemented by the Distributed Systems team. 

However, we also wanted to get a head start on refactoring components in other Merchant Center applications that will require use of this field.

The goal of this PR is to:
1. Access this field in `ApplicationShell`'s `FetchLoggedInUser` query from the stubbed `mc-services` field, and
2. Expose this field in the `ApplicationShell` connectors for downstream use within each MC application

### DIFF
--- a/test-data/user/types.ts
+++ b/test-data/user/types.ts
@@ -32,6 +32,7 @@ type TBaseUser = {
     name?: string | null | undefined;
     additionalClaims?: string | null;
   };
+  isVerified: boolean;
 };
 
 export type TUser = TBaseUser & {
--- a/test-data/user/types.ts
+++ b/test-data/user/types.ts
@@ -1,7 +1,7 @@
 import type { TBuilder } from '@commercetools-test-data/core';
 import type { TProject, TProjectGraphql } from '../project';
 
-export type VerificationStatus = 'verified' | 'unverified';
+type VerificationStatus = 'Verified' | 'Unverified';
 
 export type TCreateUserBuilder = () => TBuilder<TUser>;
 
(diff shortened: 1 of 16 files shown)


jaikamat:
Yep, whole thing is broken and I can't introspect types via `yarn generate-types` because I've introduced that new field

jaikamat:
I've regenerated types based on work in [this PR](https://github.com/commercetools/merchant-center-services/pull/2292), and expect to keep this work in draft mode until we can:
1. Safely assume the `mc-services` PR will remain stable, and
2. Hear back from Distributed Systems about namespacing (`verificationStatus` has not been officially approved)

test-data/user/types.ts
```diff
@@ -1,4 +1,5 @@
 import type { TBuilder } from '@commercetools-test-data/core';
+import { TVerificationStatus } from '../../packages/application-shell/src/types/generated/mc';
```
jaikamat, line 2: It does seem like a good idea to import directly from autogenerated types, but I'm unsure about importing these types from distant directories, is there a better way to do this?

emmenko (approved):
Nice, much simpler now! 🤗
Thanks

tdeekens (approved):
This is nice!

CarlosCortizasCT (approved):
👍

stephsprinkle (approved):

PR #2846, opened 2022-10-14
fix(application-shell-connectors): prevent Sentry reports when `additionalClaims` value is falsy

kark:
Fixes https://jira.commercetools.com/browse/SHIELD-616

### DIFF
--- a/packages/application-shell-connectors/src/components/application-context/application-context.tsx
+++ b/packages/application-shell-connectors/src/components/application-context/application-context.tsx
@@ -90,7 +90,7 @@ export const mapUserToApplicationContextUser = (user?: TFetchedUser) => {
     let additionalClaims: Record<string, unknown> = {};
     try {
       additionalClaims = JSON.parse(
-        user.idTokenUserInfo.additionalClaims || ''
+        user.idTokenUserInfo.additionalClaims || '{}'
       );
     } catch (error) {
       reportErrorToSentry(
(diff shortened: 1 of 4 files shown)


packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -90,7 +90,7 @@ export const mapUserToApplicationContextUser = (user?: TFetchedUser) => {
     let additionalClaims: Record<string, unknown> = {};
     try {
       additionalClaims = JSON.parse(
-        user.idTokenUserInfo.additionalClaims || ''
+        user.idTokenUserInfo.additionalClaims || '{}'
```
kark, line 93: I’m not sure, should empty all falsy values fall back to parsable JSON here?

test-data/user/types.ts
```diff
@@ -30,7 +30,7 @@ type TBaseUser = {
     iat: number;
     email?: string | null | undefined;
     name?: string | null | undefined;
-    additionalClaims?: string;
+    additionalClaims?: string | null;
```
kark, line 33: Taken from real-life scenario
Screenshot 2022-10-14 at 15 32 42

CarlosCortizasCT (approved):
Looks good to me.

Thanks 👍

packages/application-shell-connectors/src/components/application-context/application-context.tsx
tdeekens: To me that's okay as it eases consumption a bit as you don't always have to guard against the additionalClaims being nil.

tdeekens (approved):
Danke :)

ddouglasz (approved):
Looks good 👏

emmenko (approved):
💯

PR #2849, opened 2022-10-18
Add flag to override failOnConsole

jaikumar-tj:
#### Summary

This PR provides an override mechanism for `failOnConsole` which makes the test to fail when a console is present during the execution. It was difficult to debug without logging and by adding this flag to the test cmd can make it easier for developers to test or debug locally.

| Before | After |
|---|---|
| Screenshot 2022-10-18 at 07 36 11  | Screenshot 2022-10-18 at 07 37 07  |

### DIFF
--- a/packages/jest-preset-mc-app/setup-test-framework.js
+++ b/packages/jest-preset-mc-app/setup-test-framework.js
@@ -65,10 +65,11 @@ function shouldNotThrowWarnings(...messages) {
 }
 
 failOnConsole({
-  shouldFailOnLog: true,
-  shouldFailOnInfo: true,
-  shouldFailOnWarn: true,
-  shouldFailOnError: true,
+  // By default all of them are 'true'
+  shouldFailOnLog: !process.env.allowLog,
+  shouldFailOnInfo: !process.env.allowInfo,
+  shouldFailOnWarn: !process.env.allowWarn,
+  shouldFailOnError: !process.env.allowError,
   silenceMessage: (message) => {
     if (!process.env.CI) {
       return false;


emmenko:
Thanks for looking into it.

Just for me to understand: the use case is that you are debugging a test, so then you want to put some console logs but then the test fails because of the usage of console?

Or is this addressing a different use case?

jaikumar-tj:
@emmenko Yes, exactly what you described.

jaikumar-tj:
> Have you considered adding a flag in the `console.config`?

I tried that and it seems the I had to add the config flag in every test. Most of the time, I run `yarn test -o` to test all the modified files and with env value approach I can hard code it to my custom zsh alias and don't have to add & remove the flag everytime.

emmenko:
> with env value approach I can hard code it to my custom zsh alias and don't have to add & remove the flag everytime

Isn't that going against what we are trying to prevent here, which is to avoid having logs? If you need the log for debugging, I would image you shouldn't hardcode it as you can end up including the console logs silently.

Or am I missing something? 🤔

jaikumar-tj:
I assume that if I add console logs accidentally then `failOnConsole` will still make the test fail on circleCi since I'm only changing it in my local alias. Or do you mean that the original purpose is to prevent dev from committing any `console log` itself?

jaikumar-tj:
Just for transparency. After the discussion, it was decided that it is better to avoid env since it might go unnoticed if we start hardcoding it in somewhere (eg: zshrc). So I will try to find a different solution which works similar to `global.console.config` and also we need to make sure this shouldn't affect in circleCI. Either we need an override mechanism in CI or we need addition ESLint rule to make sure developers don't commit this.

jaikumar-tj:
Closing this in favour of [this PR](https://github.com/commercetools/merchant-center-application-kit/pull/3086#pullrequestreview-1434467838)

PR #2852, opened 2022-10-20
chore: add user event to rtl

antoniolodias:
#### Summary

This PR aims to add `userEvent` method available on rtl tests. 
This was started actually more than a year ago and somehow forgotten...

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,6 +112,7 @@
     "@svgr/plugin-svgo": "6.5.0",
     "@testing-library/cypress": "8.0.3",
     "@testing-library/react": "12.1.5",
+    "@testing-library/user-event": "^14.4.3",
     "@tsconfig/cypress": "^1.0.1",
     "@types/jest-environment-puppeteer": "^5.0.2",
     "@types/node": "^16.11.58",
--- a/package.json
+++ b/package.json
@@ -112,7 +112,6 @@
     "@svgr/plugin-svgo": "6.5.0",
     "@testing-library/cypress": "8.0.3",
     "@testing-library/react": "12.1.5",
-    "@testing-library/user-event": "^14.4.3",
     "@tsconfig/cypress": "^1.0.1",
     "@types/jest-environment-puppeteer": "^5.0.2",
     "@types/node": "^16.11.58",
--- a/packages/application-shell/package.json
+++ b/packages/application-shell/package.json
@@ -122,7 +122,7 @@
     "@apollo/client": "3.x",
     "@testing-library/react": "12.x",
     "@testing-library/react-hooks": "7.x || 8.x",
-    "@testing-library/user-event": "^14.4.3",
+    "@testing-library/user-event": "14.x",
     "react": "17.x",
     "react-dom": "17.x",
     "react-intl": "5.x",
--- a/packages/application-shell/package.json
+++ b/packages/application-shell/package.json
@@ -109,6 +109,7 @@
     "@apollo/client": "3.7.0",
     "@testing-library/react": "12.1.5",
     "@testing-library/react-hooks": "8.0.1",
+    "@testing-library/user-event": "14.0.1",
     "msw": "0.47.4",
     "react": "17.0.2",
     "react-dom": "17.0.2",
--- a/packages/application-shell/package.json
+++ b/packages/application-shell/package.json
@@ -122,6 +122,7 @@
     "@apollo/client": "3.x",
     "@testing-library/react": "12.x",
     "@testing-library/react-hooks": "7.x || 8.x",
+    "@testing-library/user-event": "^14.4.3",
     "react": "17.x",
     "react-dom": "17.x",
     "react-intl": "5.x",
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -112,6 +112,7 @@
     "@svgr/plugin-svgo": "6.3.1",
     "@testing-library/cypress": "8.0.3",
     "@testing-library/react": "12.1.5",
+    "@testing-library/user-event": "^14.4.3",
```
emmenko, line 115: This shouldn't be needed here.

packages/application-shell/package.json
```diff
@@ -109,6 +109,7 @@
     "@apollo/client": "3.6.9",
     "@testing-library/react": "12.1.5",
     "@testing-library/react-hooks": "8.0.1",
+    "@testing-library/user-event": "^14.4.3",
```
emmenko, line 112: Nit: pin the version as we do for the other dependencies

packages/application-shell/package.json
```diff
@@ -122,6 +123,7 @@
     "@apollo/client": "3.x",
     "@testing-library/react": "12.x",
     "@testing-library/react-hooks": "7.x || 8.x",
+    "@testing-library/user-event": "^14.4.3",
```
emmenko, line 126: Please follow the same version range pattern

packages/application-shell/src/test-utils/test-utils.tsx
```diff
@@ -656,6 +656,9 @@ function renderHook<
 // re-export everything
 export * from '@testing-library/react';
 
+// re-export everything from @testing-library/user-event
+export * from '@testing-library/user-event';
```
emmenko, line 660: I would suggest to namespace the exports as `userEvents`, like for hooks to avoid possible conflicts.

packages/application-shell/package.json
emmenko: Also check that this is compatible with e.g. React v17, RTL v12, etc.

emmenko (commented):
Thanks for the contribution!

I have a question though: is there a particular reason for including this in AppKit and not using it directly?

antoniolodias:
> Thanks for the contribution!
> 
> I have a question though: is there a particular reason for including this in AppKit and not using it directly?

Sure 👍
Well, honestly I forgot a bit the entire context of this ticket, it was something I discussed with @ahmehri a year ago and yesterday @jocelyn-dunkley asked about it, so I decided to bring it up again. 
So, I don't have a personal opinion on where to set this up, it seems simpler to just install it directly on the frontend as you mentioned, I think the idea at the time was to make it the "correct way". Maybe @ahmehri remembers? or maybe do you suggest just installing it where we need it?

jocelyn-dunkley:
> > Thanks for the contribution!
> > I have a question though: is there a particular reason for including this in AppKit and not using it directly?
> 
> Sure 👍 Well, honestly I forgot a bit the entire context of this ticket, it was something I discussed with @ahmehri a year ago and yesterday @jocelyn-dunkley asked about it, so I decided to bring it up again. So, I don't have a personal opinion on where to set this up, it seems simpler to just install it directly on the frontend as you mentioned, I think the idea at the time was to make it the "correct way". Maybe @ahmehri remembers? or maybe do you suggest just installing it where we need it?

From the thread that I found Ahmed said that the other testing library import came directly from app-kit so he didn't want to diverge from the convention.
"I think using it directly is a divergence from the existing system (everything related to @testing-library is exposed from app-kit) and I wouldn’t recommend it. I suggest instead spinning a PR in app-kit to expose it."

emmenko:
It's a bit of a tradeoff. The other RTL dependencies for instance are also required to be installed by the consumers as they are required as `peerDependencies`. We do the re-exports mostly to group the imports and to ensure the test utils use the same package versions (and we also use these dependencies internally in appkit).

We can do the same with the `@testing-library/user-event` but the consumer would still need to install the package because of the peer dependency.

ahmehri:
> We can do the same with the @testing-library/user-event but the consumer would still need to install the package because of the peer dependency.

@emmenko then there is no objection to re-export the library from app-kit, as far as I understood. I think it's helpful to use the same system and re-export from here, and as with other re-exports, the consumer will still need to install the dependency, which shouldn't be a problem.

emmenko:
Yes, like I said the only "downside" here is that you have to install it even if you are not using it.

Also I don't know if this would be a breaking change technically 🤔

ahmehri:
So what remains is implementing the requested code review changes (@antoniolodias) and deciding the version.

packages/application-shell/src/test-utils/test-utils.tsx
```diff
@@ -656,6 +656,9 @@ function renderHook<
 // re-export everything
 export * from '@testing-library/react';
 
+// re-export everything from @testing-library/user-event
+export * as userEvent from '@testing-library/user-event';
```
antoniolodias, line 660: is this what you meant with - "I would suggest to namespace the exports as userEvents"?  @emmenko

emmenko (commented):
Thanks! Please also add a changeset.

jaikumar-tj (approved):

tdeekens (approved):
💯

PR #2857, opened 2022-10-22
fix: patch writing-style vale config to disable custom `BlockIgnores` regex

emmenko:
To unblock the updates to the `@commercetools-docs/writing-style` package, we're patching the `.vale.ini` config to disable the custom `BlockIgnores` regex, which doesn't work in this repository.

See https://github.com/commercetools/commercetools-docs-kit/issues/1341 for more info about the original issues.

### DIFF
--- a/.yarn/patches/@commercetools-docs-writing-style-npm-3.3.1-7d206a1314.patch
+++ b/.yarn/patches/@commercetools-docs-writing-style-npm-3.3.1-7d206a1314.patch
@@ -1,0 +1,13 @@
+diff --git a/.vale.ini b/.vale.ini
+index ca41229fb40f50879dab25dc0da8893abdd72f31..3a06aaa5e4225f8b36b7b6b251c2aa8d499c3e05 100644
+--- a/.vale.ini
++++ b/.vale.ini
+@@ -12,7 +12,7 @@ mdx = md
+ 
+ [*.{md,mdx}]
+ BasedOnStyles = Vale, write-good, Google, commercetools
+-BlockIgnores = (?s)(<[A-Z].*?\/>)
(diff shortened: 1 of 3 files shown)


.yarn/patches/@commercetools-docs-writing-style-npm-3.3.1-7d206a1314.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/.vale.ini b/.vale.ini
+index ca41229fb40f50879dab25dc0da8893abdd72f31..3a06aaa5e4225f8b36b7b6b251c2aa8d499c3e05 100644
+--- a/.vale.ini
++++ b/.vale.ini
+@@ -12,7 +12,7 @@ mdx = md
+ 
+ [*.{md,mdx}]
+ BasedOnStyles = Vale, write-good, Google, commercetools
+-BlockIgnores = (?s)(<[A-Z].*?\/>)
++; BlockIgnores = (?s)(<[A-Z].*?\/>)
```
emmenko, line 10: @commercetools/docs-tech FYI: I tried different things and regex combinations but I wasn't able to make it work.
The only way I could find to solve the error is to keep the default config and not use the custom `BlockIgnores` regex.

@nkuehn Do you remember what the original issue was?

emmenko: It seems that using `TokenIgnores` instead of `BlockIgnores` does the trick.

emmenko: Opened a PR in docs-kit: https://github.com/commercetools/commercetools-docs-kit/pull/1405

Issue #2851, opened 2022-10-19
Infinite reloads on deployed custom apps


Steven-Thelin:
I’ve migrated our three custom apps to v21. We’ve deployed these to our Azure staging environment, added the three applications to Merchant Center as well as mapped their respectful URL’s. However, opening any of the three custom apps just makes the page infinitely reload.

This doesn’t happen in the local development environment. An example of one of the custom apps that loops:
https://mc.europe-west1.gcp.commercetools.com/lakrids-by-blow-staging/order-management-org

**To Reproduce**
Steps to reproduce the behaviour:
Open any of the new custom apps in Merchant Center (I believe you have access to our MC?)

**Expected behaviour**

For it not to be stuck in a refresh loop

Please let me know what else you need to look into this. The code has been shared in the last issue where you guys politely helped me to get the custom apps to run locally.

CarlosCortizasCT:
Hello @Steven-Thelin 

The issue might be related with a misconfiguration around the entry point URI path of the Custom Applications.

If you access the configuration for your Custom Application in the Merchant Center, you could check which entry point is configured for it. Example:
![image](https://user-images.githubusercontent.com/97907068/196877509-5423e426-3689-4f88-b5e3-cf9995ba3eee.png)

You need to make sure the value for that property is the same as you have in your Custom Application config file. The name of the property in the config file is [entryPointUriPath](https://docs.commercetools.com/custom-applications/api-reference/application-config#entrypointuripath).

Could you please double check the values match for your Custom Applications?

Steven-Thelin:
Hi @CarlosCortizasCT 

The `entryUriPath` does already match the `entryUriPath` defined in the application 🙂

CarlosCortizasCT:
Hi @Steven-Thelin 

I guess we will need extra info to try to figure out what's happening.

These are the points that could help us:
* Could you share the contents of your `custom-application-config` file? (obfuscate private information if any)
  * If you are using any environment variable, could you make sure they are defined when building the application?
* Is the redirection happening at server response level (eg: status code 30X) or the page loads and then it reloads because of some javascript code running?
* If the reload is triggered by javascript client-side, you might need to stop the redirection loop in the browser and inspect the dev tools console. Is there any error message? Could you inspect the property `window.app` and share with us its contents?

Thanks.

Steven-Thelin:
Hi @CarlosCortizasCT ! Thanks for your response, and I'm sorry for the late reply.

[...]

PR #2860, opened 2022-10-24
Fix problem when calculating CSP headers from Custom Application config file

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=U74fN8l)

#### Summary

When users use an application URL without an ending backslash we add an invalid URL to CSP meta element in compiled HTML when building a Custom Application.

Related to #2851 (end of the thread)

#### Description

When compiling the main `index.html` file for a Custom Application we add a `meta` element with some CSP directives so we can control from which sources scripts/styles can be loaded.
In those directives we need to include the Custom Application hosting provider public URL since the application is going to be loaded from a commercetools domain.

We use [cdnUrl](https://docs.commercetools.com/custom-applications/api-reference/application-config#envproductioncdnurl) or [url](https://docs.commercetools.com/custom-applications/api-reference/application-config#envproductionurl) properties from the config file of the Custom Application.
Currently we just copy the value to the CSP directive, but that can be a problem when the URL does not contain an ending backslash.

Let's say the value we have in the config file is this:
(_for the purpose of this example, we don't host the application in the root directory of the domain_)
```
https://www.my-domain.com/admin
```
This is telling the browser only the file named `admin` from the domain `www.my-domain.com` is allowed to be loaded.

What we need is for that URL to have an ending backslash:
```
https://www.my-domain.com/admin/
```
Using this value we are telling the browser it can fetch anything `below` the directory `admin` from the domain `www.my-domain.com`.

In this PR we update the way we process the config file and calculate those CSP directives to make sure that, when using the `cdnUrl` or `url` properties, they have an ending backslash.

### DIFF
--- a/.changeset/stale-apes-watch.md
+++ b/.changeset/stale-apes-watch.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-config': patch
+---
+
+Fix CSP headers processing at Custom Application build time when using a deployment URL without an ending backslash
--- a/packages/application-config/src/process-config.ts
+++ b/packages/application-config/src/process-config.ts
@@ -167,7 +167,7 @@ const processConfig = ({
         // otherwise it might create an invalid value when application/CDN URL points to a
         // non-root directory (ex: https://www.my-domain.com/app). This is a valid URL but from
         // the CSP point of view, it will say only the file `app` can be used as a source, so
-        // any other file from that domain will be forbidden. Using the slash (ex: https://www.my-domain.com/app)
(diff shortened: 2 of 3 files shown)


emmenko (approved):
Good catch, thanks for fixing it! 🙌

tdeekens (approved):
Ouh, good find! Thanks for fixing it and cementing it with tests!

kark (approved):
💯

ddouglasz (approved):
👏

PR #2862, opened 2022-10-26
refactor(config): expose `sanitizeSvg` function, fix diffing logic

emmenko:
I noticed when using the CLI `config:sync` command that the SVG icon was always marked as changed.

The problem is in the fact that the API sanitizes the SVG before saving the data but we were also pre-sanitizing it when the config was parsed. This lead to updates to have run the sanitize 2 times.

Now we don't sanitize when parsing the icon but only during the diffing, so ensure the diffed data is the same.

### DIFF
--- a/packages/application-config/src/sanitize-svg.ts
+++ b/packages/application-config/src/sanitize-svg.ts
@@ -8,9 +8,10 @@ const DOMPurify = createDOMPurify(jsdom.window);
 export default function sanitizeSvg(data: string) {
   return DOMPurify.sanitize(data, {
     USE_PROFILES: { svg: true },
+    RETURN_DOM: true,
     FORBID_ATTR: [
       // To avoid injection by using `style="filter:url(\"data:image/svg+xml,<svg`
       'style',
     ],
-  });
+  }).innerHTML;
 }
--- a/packages/application-config/src/ssr.ts
+++ b/packages/application-config/src/ssr.ts
@@ -1,2 +1,1 @@
 export * from './formatters';
-export { default as sanitizeSvg } from './sanitize-svg';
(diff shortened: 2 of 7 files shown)


packages/application-config/src/ssr.ts
```diff
@@ -1 +1,2 @@
 export * from './formatters';
+export { default as sanitizeSvg } from './sanitize-svg';
```
emmenko, line 2: So that we can use the same logic in the backend.

packages/application-config/src/sanitize-svg.ts
```diff
@@ -8,9 +8,10 @@ const DOMPurify = createDOMPurify(jsdom.window);
 export default function sanitizeSvg(data: string) {
   return DOMPurify.sanitize(data, {
     USE_PROFILES: { svg: true },
+    RETURN_DOM: true,
     FORBID_ATTR: [
       // To avoid injection by using `style="filter:url(\"data:image/svg+xml,<svg`
       'style',
     ],
-  });
+  }).innerHTML;
```
CarlosCortizasCT, line 16: Could you perhaps explain what this changes are meant for?

emmenko: Yes sorry. We already changed this in the backend, due to some security findings.

image

https://research.securitum.com/mutation-xss-via-mathml-mutation-dompurify-2-0-17-bypass/

CarlosCortizasCT: Thanks for the explanation 👍

kark: Wowzers, thanks for explaining 🙏 I guess we should remember to change that then in the `<InlineSvg>` component from ui-kit too (probably as part of the upcoming svg changes)

emmenko: Yes, good point!

kark (approved):

PR #2863, opened 2022-10-26
Disable dev-only redux immutability check

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=Lszhi3H)

#### Summary

Disable dev-only redux immutability check

#### Description

We've seen some tests randomly failing in the frontend due to [this default check](https://redux-toolkit.js.org/api/immutabilityMiddleware) `redux-toolkit` middleware is adding to our store.
The check tries to make sure you don't mutate your store state directly by traversing all the state tree and it logs a warning to the console if it takes more that 32ms by default.
The message it logs is not very helpful to really understand what is causing the process to take so long.

In the context where we use the redux store (the `application-shell` package in this case) we don't see this check very valuable as we control state updates and we know for sure we're no mutating it so we decided to turn off the check.

### DIFF
--- a/.changeset/lemon-roses-camp.md
+++ b/.changeset/lemon-roses-camp.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': patch
+---
+
+Disable development-only Redux Toolkit default [immutability middleware](https://redux-toolkit.js.org/api/immutabilityMiddleware) check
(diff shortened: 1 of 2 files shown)


kark (approved):
Thanks for the thorough explanation of the issue 💯

emmenko (approved):
Don't forget a changeset. Otherwise 💯

PR #2864, opened 2022-10-28
feat(mc-scripts): attempt to automatically refresh the access token

emmenko:
While testing some things with the CLI command `config:sync`, I noticed that when the CoCo access token (not the session token) expires or is invalid the CLI fails (for obvious reasons).

Example reason:

```
Access token is inactive and therefore cannot be used anymore.
```

However, the CLI session token is technically still valid. This makes the CLI to be in a weird state, difficult to recover. For example if I try to `login` again it says that I already have a valid session.

In the Custom Application we have a built-in mechanism to retry an unauthorized request to force the MC API to get a new valid access token.
We need this same mechanism for the CLI requests.

See inline comments for more info.

Below you can how the CLI handles the refresh.

Screenshot 2022-10-28 at 09 55 30


packages/mc-scripts/bin/cli.js
```diff
@@ -1,3 +1,8 @@
 #!/usr/bin/env node
 
-require('@commercetools-frontend/mc-scripts/cli').run();
+const { run } = require('@commercetools-frontend/mc-scripts/cli');
+
+run().catch((error) => {
+  console.error(error.message || error.stack || error);
+  process.exit(1);
+});
```
emmenko, line 8: Changed this a bit to improve error handling.

packages/mc-scripts/src/cli.ts
```diff
@@ -199,8 +199,9 @@ const run = () => {
   cli.help();
   cli.version(pkgJson.version);
 
-  cli.parse();
-};
+  cli.parse(process.argv, { run: false });
+  await cli.runMatchedCommand();
```
emmenko, line 203: Allows to handle errors

See https://www.npmjs.com/package/cac#error-handling

packages/mc-scripts/src/commands/config-sync.ts
```diff
@@ -32,21 +32,23 @@ async function run(options: TCliCommandConfigSyncOptions) {
   const { data: localCustomAppData } = applicationConfig;
   const { mcApiUrl } = applicationConfig.env;
 
-  const token = credentialsStorage.getToken(mcApiUrl);
-  if (!token) {
+  console.log(`Using Merchant Center environment "${chalk.green(mcApiUrl)}".`);
+  console.log();
```
emmenko, line 36: Slightly unrelated, just improving the CLI output a bit.

packages/mc-scripts/src/utils/graphql-requests.ts
```diff
@@ -21,111 +20,189 @@ import FetchCustomApplicationFromCli from './fetch-custom-application.settings.g
 import UpdateCustomApplicationFromCli from './update-custom-application.settings.graphql';
 import CreateCustomApplicationFromCli from './create-custom-application.settings.graphql';
 import FetchMyOrganizationsFromCli from './fetch-user-organizations.core.graphql';
+import CredentialsStorage from './credentials-storage';
 
 type TFetchCustomApplicationOptions = {
   mcApiUrl: string;
-  token: string;
```
emmenko, line 27: Not needed anymore to pass explicitly. It's now read from within the requests.

packages/mc-scripts/src/utils/graphql-requests.ts
```diff
@@ -21,111 +20,189 @@ import FetchCustomApplicationFromCli from './fetch-custom-application.settings.g
...
 
-const graphQLClient = (
-  url: string,
-  token: string,
-  target: TGraphQLTargets = GRAPHQL_TARGETS.SETTINGS_SERVICE
-) =>
-  new GraphQLClient(`${url}/graphql`, {
+const credentialsStorage = new CredentialsStorage();
+
+const client = new GraphQLClient(
```
emmenko, line 46: Creating only one instance of the client now

packages/mc-scripts/src/utils/graphql-requests.ts
```diff
@@ -21,111 +20,189 @@ import FetchCustomApplicationFromCli from './fetch-custom-application.settings.g
...
+async function requestWithTokenRetry<Data, QueryVariables>(
+  document: DocumentNode,
+  requestOptions: {
+    variables?: QueryVariables;
+    mcApiUrl: string;
+    headers: HeadersInit;
+  },
+  retryCount: number = 0
+): Promise<Data> {
+  const token = credentialsStorage.getToken(requestOptions.mcApiUrl);
```
emmenko, line 66: Here we read the token now.

packages/mc-scripts/src/utils/graphql-requests.ts
```diff
@@ -21,111 +20,189 @@ import FetchCustomApplicationFromCli from './fetch-custom-application.settings.g
...
+
+  client.setEndpoint(`${requestOptions.mcApiUrl}/graphql`);
+  client.setHeaders(requestOptions.headers);
+  if (token) {
+    client.setHeader('x-mc-cli-access-token', token);
+  }
```
emmenko, line 72: Configuring the request

packages/mc-scripts/src/utils/graphql-requests.ts
```diff
@@ -21,111 +20,189 @@ import FetchCustomApplicationFromCli from './fetch-custom-application.settings.g
...
+
+  try {
+    const result = await client.rawRequest<Data, QueryVariables>(
```
emmenko, line 75: Using `rawRequest` to be able to access things like response headers.

packages/mc-scripts/src/utils/graphql-requests.ts
```diff
@@ -21,111 +20,189 @@ import FetchCustomApplicationFromCli from './fetch-custom-application.settings.g
...
+      console.log();
+      const refreshedSessionTokenExpiresAt = result.headers.get(
+        'x-refreshed-session-token-expires-at'
+      );
+      // Store the updated access token.
+      credentialsStorage.setToken(requestOptions.mcApiUrl, {
+        token: refreshedSessionToken,
+        expiresAt: Number(refreshedSessionTokenExpiresAt),
+      });
+    }
```
emmenko, line 95: This handles storing the refreshed session token from the response header.

packages/mc-scripts/src/utils/graphql-requests.ts
```diff
@@ -21,111 +20,189 @@ import FetchCustomApplicationFromCli from './fetch-custom-application.settings.g
...
+              'Expired or invalid session token, attempting to retry the request with a refreshed token...'
+            )
+          );
+          return requestWithTokenRetry(
+            document,
+            {
+              ...requestOptions,
+              headers: {
+                ...requestOptions.headers,
+                'X-Force-Token': 'true',
```
emmenko, line 121: This forces the refresh of the token in the MC API.

emmenko:
This can be reviewed but we should merge it once the MC API has been updated.

CarlosCortizasCT (approved):
👏

packages/mc-scripts/src/utils/graphql-requests.ts
Rhotimee: 💯

Rhotimee (approved):
Looks great. Thank you 🙌🏽

PR #2868, opened 2022-11-03
PCM-2794: fix quick access to product list

obulaworld:
#### Summary

This PR aims to fix the product list quick access. Currently the url to which it directs to is `/products/pim-search` which displays the text `We could not find what you are looking for`. In other to solve this, we need to use the real link which is `/products`.

### DIFF
--- a/.changeset/proud-ants-flow.md
+++ b/.changeset/proud-ants-flow.md
@@ -1,5 +1,5 @@
 ---
-"@commercetools-frontend/application-shell": patch
+'@commercetools-frontend/application-shell': patch
 ---
 
-PCM-2794: fix quick access to product list
+Fix QuickAccess component to make it use the right products list URL.
--- a/.changeset/proud-ants-flow.md
+++ b/.changeset/proud-ants-flow.md
@@ -1,0 +1,5 @@
+---
+"@commercetools-frontend/application-shell": patch
+---
+
+PCM-2794: fix quick access to product list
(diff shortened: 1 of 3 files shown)


CarlosCortizasCT (approved):
Looks good to me 👍

CarlosCortizasCT:
Please remember to add a changeset.

ahmehri (approved):

Rhotimee (approved):

.changeset/proud-ants-flow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@commercetools-frontend/application-shell": patch
+---
+
+PCM-2794: fix quick access to product list
```
CarlosCortizasCT, line 5: Bear in mind this description will go to our public changeset so we should not reference internal issues.

Perhaps we can update it to be something like this:
```
Fix QuickAccess component to make it use the right products list URL.
```

obulaworld: Hi @CarlosCortizasCT What is the next step for this change? Will it be deployed with the next deploy train? Just trying to be sure

CarlosCortizasCT: Hi @obulaworld 

There is no automatic release train for `app-kit`, we trigger a new release whenever we have a bunch of changes that worth it.

Currently the only interesting thing is this fix. If this is something important, we can publish a new version today.

obulaworld: Yeah it is important for the quick access links to work. Can we publish today?

PR #2870, opened 2022-11-07
Set up page layout components for the new theme

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=O2FsAVc)

#### Summary

Adding support for design tokens in AppKit and update page layout components to support new theme UI.

#### Description

The idea is to use the `themeOverrides` prop the `ui-kit` `ThemeProvider` expose to extend the `ui-kit` tokens with the ones needed in the `app-kit`.
This strategy works as long as there is a `ThemeProvider` in the tree where an `app-kit` component is rendered, which will be true for any application using the `ApplicationShell` component at the top of the tree but, otherwise, a `ThemeProvider` would must be included by the application developer.

We need to create some tokens similarly to what we do in `ui-kit`: since we need to support several themes, in our components we need to use a single token name but it must be linked to several values (one per theme).
(_take a look at inline comments for more details_).

In this PR we updated modal pages, detail pages, main pages and dialog components to match the new UI in the new theme.
We focused only in the changes that can be implemented by using design tokens which leaves out some requirements that will need more refactoring work to be accomplished.

This is the list of new theme UI requirements that have not been implemented (will be worked on in a follow-up PR):
* Page subtitles should have a different color
* Main pages should not have a divider line between header and content
* Dialogs main paddings should be adjusted
* Modals titles should be H1 (not they are H4)
* Modals close icon must be bigger
* Tabs controls labels should be H3 (now they are H4)
* Form control buttons should be in the same "row" as title/subtitle


packages/application-components/src/theming.ts
```diff
@@ -0,0 +1,25 @@
+import kebabCase from 'lodash/kebabCase';
```
CarlosCortizasCT, line 1: It would probably be better to have an independent package in `app-kit` to have the theming part.

packages/application-components/src/theming.ts
```diff
@@ -0,0 +1,25 @@
...
+import { designTokens } from '@commercetools-uikit/design-system';
+
+export const themesOverrides = {
```
CarlosCortizasCT, line 4: Here is where we define the new tokens and their values per theme.

packages/application-components/src/theming.ts
```diff
@@ -0,0 +1,25 @@
...
+    paddingVerticalForPage: designTokens.spacingM,
+    paddingHorizontalForPage: designTokens.spacingM,
+  },
+  test: {
+    paddingVerticalForPage: designTokens.spacingXl,
+    paddingHorizontalForPage: '40px',
+  },
+};
+
+export const appKitDesignTokens = Object.values(themesOverrides).reduce(
```
CarlosCortizasCT, line 15: Here is where we create the design tokens to be used in the components styling code.
It _knows_ how the `ThemeProvider` will convert the tokens names into CSS variables so it does the same.

packages/application-components/src/theming.ts
```diff
@@ -0,0 +1,25 @@
...
+  (appKitTokens, themeTokens) => {
+    Object.keys(themeTokens).forEach((tokenName) => {
+      if (!appKitTokens[tokenName]) {
+        appKitTokens[tokenName] = `var(--${kebabCase(tokenName)})`;
```
CarlosCortizasCT, line 19: Since this are new tokens I don't think we need default values for it, but it depends on the decision we need to make about if we will support using `app-kit` components without also using a `ThemeProvider`.

packages/application-shell/src/components/application-shell/application-shell.tsx
```diff
@@ -536,7 +539,7 @@ const ApplicationShell = <AdditionalEnvironmentProperties extends {}>(
 
   return (
     <>
-      <ThemeProvider theme="default" />
+      <ThemeProvider theme="default" themeOverrides={themesOverrides.default} />
```
CarlosCortizasCT, line 542: We will need to move the location of the `ThemeProvider` down the tree as it would need the `FlopFlipProvider` to be setup so we can read the feature flag we will use to know which theme to use.
No sure if we would also need to move the `GlobalStyles`.

packages/application-components/src/theming.ts
emmenko: We should probably expose the `toVar` function from the `design-system` package.

packages/application-components/src/components/detail-pages/custom-form-detail-page/custom-form-detail-page.tsx
```diff
@@ -12,11 +12,13 @@ import Spacings from '@commercetools-uikit/spacings';
...
 import PageTopBar from '../../internals/page-top-bar';
 import { ContentWrapper, PageWrapper } from '../../internals/page.styles';
+import { appKitDesignTokens } from '../../../theming';
 
 const DetailPageContainer = styled.div`
   background-color: ${customProperties.colorNeutral95};
-  padding: ${customProperties.spacingM};
   border-bottom: 1px ${customProperties.colorNeutral} solid;
+  padding: ${appKitDesignTokens.paddingVerticalForPage}
+    ${appKitDesignTokens.paddingHorizontalForPage};
```
emmenko, line 21: Can we have 1 token only?

packages/application-components/src/theming.ts
```diff
@@ -0,0 +1,25 @@
...
+export const themesOverrides = {
+  default: {
+    paddingVerticalForPage: designTokens.spacingM,
+    paddingHorizontalForPage: designTokens.spacingM,
+  },
+  test: {
+    paddingVerticalForPage: designTokens.spacingXl,
+    paddingHorizontalForPage: '40px',
+  },
+};
```
emmenko, line 13: Alternative approach, given that these tokens are mainly for the "migration" rather than the actual theme.

In the docs-kit (if you remember) I injected some inline css variables to the parent container, which can then be referenced in the styles.
https://github.com/commercetools/commercetools-docs-kit/blob/e873a23a6ea9647bb09d9e43d9c535880d616d48/packages/gatsby-theme-docs/src/layouts/internals/layout-page-with-hero.js#L89-L91

Maybe we can do the same here? So we're not really extending the theme but simply injecting custom css variables. The values can then be determined for example based on the feature flag (or something similar).

packages/application-components/src/components/detail-pages/custom-form-detail-page/custom-form-detail-page.tsx
CarlosCortizasCT: Updated in this commit: 1752af3561e6d0e5cb7d96d2ea8cb37dfb6bfa7d

CarlosCortizasCT:
Current Percy failing tests are due to the new icons in `ui-kit`.

packages/application-components/src/components/detail-pages/custom-form-detail-page/custom-form-detail-page.tsx
```diff
@@ -81,25 +85,25 @@ const CustomFormDetailPage = (props: CustomFormDetailPageProps) => {
   return (
     <PageWrapper>
       <DetailPageContainer>
-        <Spacings.Stack>
-          <PageTopBar
-            color="neutral"
-            previousPathLabel={props.previousPathLabel}
-            onClick={props.onPreviousPathClick}
```
CarlosCortizasCT, line 88: I removed the `Spacings.Stack` to be able to use different margins per theme.

packages/application-shell/src/components/theme-switcher/theme-switcher.tsx
```diff
@@ -1,10 +1,14 @@
...
 import { useFeatureToggle } from '@flopflip/react-broadcast';
+import { themesOverrides } from '@commercetools-frontend/application-components';
 import { UI_REDESIGN } from '../../feature-toggles';
 
 const ThemeSwitcher = () => {
   const isNewThemeEnabled = useFeatureToggle(UI_REDESIGN);
-  return <ThemeProvider theme={isNewThemeEnabled ? 'test' : 'default'} />;
+  const theme = isNewThemeEnabled ? 'test' : 'default';
+  return (
+    <ThemeProvider theme={theme} themeOverrides={themesOverrides[theme]} />
```
CarlosCortizasCT, line 10: Here is were we extend the `ui-kit` theme with `app-kit` design tokens.

visual-testing-app/src/components/theme-switcher/theme-switcher.tsx
```diff
@@ -0,0 +1,40 @@
+import { themes, designTokens } from '@commercetools-uikit/design-system';
+
+export type ThemeName = keyof typeof themes;
+
+type ThemeSwitcherProps = {
+  selectedTheme: string;
+  onThemeChange(theme: ThemeName): void;
+};
+
+const ThemeSwitcher = ({
```
CarlosCortizasCT, line 10: This is a basic utility I used during development to be able to compare default and new theme.
I guess it's helpful but we can remove it if we don't want it (or make it looks nicer if we want to keep it).

visual-testing-app/src/index.tsx
```diff
@@ -1,13 +1,35 @@
...
+      <ThemeProvider
+        theme={selectedTheme}
+        themeOverrides={themesOverrides[selectedTheme]}
+      />
+      {isLocalEnvironment ? (
+        <ThemeSwitcher
+          selectedTheme={selectedTheme}
+          onThemeChange={(newTheme) => updateTheme(newTheme)}
+        />
+      ) : null}
```
CarlosCortizasCT, line 28: We add the theme switcher only in development mode, where it's useful.
_(Otherwise, it would be included in Percy snapshots)_

emmenko (commented):
Thanks, looks good! I tested the changes in the VRT app.

One thing though that I wanted to ask (in particular @FilPob ): I'm wondering if we want to adjust the "sizes" of the dialog containers (widths), as with the new design (more spacing, bigger fonts) the content gets really narrow.

https://github.com/commercetools/merchant-center-application-kit/blob/66acf9b966c3eaaa410c2298dbfba45d959f528b/packages/application-components/src/components/dialogs/internals/dialog.styles.ts#L9-L26

tdeekens (approved):
Did some visual testing and followed along with the code. I like that the theming impact is centralized and hence also easy to remove again if needed.

FilPob:
@CarlosCortizasCT regarding the suggestion from Nicola above I would suggest to increase the widths of the Dialogs to:

Size M: From  horiz.constraint 7 to 9
Size L: From horiz. constraint from 10 to 13.

The new width is kinda the equivalent to the old ones. Taking the increased padding and bigger font size into consideration

emmenko:
FYI: The t-shirt sizes are deprecated, we use the other scale `7, 8, 9, 10`.

FilPob:
In the Appkit docs there is the "size"property for the info dialog with the values S and M. That's where I took it from

emmenko:
We should probably change that. Sorry for the confusion.

CarlosCortizasCT:
@emmenko, @FilPob sorry guys but I'm not sure. Are we finally changing something regarding the dialogs?

FilPob:
> @emmenko, @FilPob sorry guys but I'm not sure. Are we finally changing something regarding the dialogs?
> 
> These are the current values (`m` and `l` sizes are deprecated): image


Ah Ok, I didn't know that the S,M are deprecated. Therefore  I wanted to suggest to change the **M to hor.constraint 9** and **L to constraint 13**  By doing that, all dialogs in MC would get a bit wider right? That would be the preferred outcome.

But In case we dont have M and L anymore and the user is able to pick any horizontal constraint from the table you attached, It would be good to at least extend the options up to the constraint 13... But how would we ensure then that all dialogs in MC get wider? 

I am not sure though if I understand the whole thing 100%.

CarlosCortizasCT:
> > @emmenko, @FilPob sorry guys but I'm not sure. Are we finally changing something regarding the dialogs?
> > These are the current values (`m` and `l` sizes are deprecated): image
> 
> Ah Ok, I didn't know that the S,M are deprecated. Therefore I wanted to suggest to change the **M to hor.constraint 9** and **L to constraint 13** By doing that, all dialogs in MC would get a bit wider right? That would be the preferred outcome.
> 
> But In case we dont have M and L anymore and the user is able to pick any horizontal constraint from the table you attached, It would be good to at least extend the options up to the constraint 13... But how would we ensure then that all dialogs in MC get wider?
> 
> I am not sure though if I understand the whole thing 100%.

I've been gathering some information about how we use the dialogs in the MC frontend and came up with this data:
image

So, in the vast majority of usages, we don't specify a custom size for the dialogs meaning the default one is used. The default value is `l` size or `10` constraint.

We currently have both t-shirt and constraints scales but the former should be deprecated.
While we continue supporting both (until we introduce a breaking change removing the t-shirt sizes) we need to maintain both.

I think then what we have to do, based on latest @FilPob comment, is to update `m` and `l` t-shirt sizes and also expand the constraint available values with `11`, `12` and `13`.
Default value will change from `484px` to `642px` so all usages where the default value is used, they will automatically be updated.

emmenko:
> I think then what we have to do, based on latest @FilPob comment, is to update m and l t-shirt sizes and also expand the constraint available values with 11, 12 and 13.
Default value will change from 484px to 642px so all usages where the default value is used, they will automatically be updated.

Agree.

We can then deprecate values that we don't intend people to use, but for now we should keep all the options.

FilPob:
Agreed. Thanks :)

CarlosCortizasCT:
I've updated the dialog components sizes based on what we agreed.

This is the commit: a49a9fceefb65451038be05918b55bb3838471ef

packages/application-components/src/components/dialogs/internals/dialog-container.tsx
```diff
@@ -41,14 +41,14 @@ const getOverlayElement: ModalProps['overlayElement'] = (
...
-  size: 'm' | 'l' | 7 | 8 | 9 | 10 | 'scale';
+  size: 'm' | 'l' | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 'scale';
   zIndex?: number;
   title: string;
   children: ReactNode;
   getParentSelector: typeof getDefaultParentSelector;
 };
 const defaultProps: Pick<Props, 'size' | 'getParentSelector'> = {
-  size: 10,
+  size: 'l',
```
CarlosCortizasCT, line 51: We want to have a different default size per theme.
I didn't think changing the constraint 10 token to have a different values (it would have unwanted secondary effects), I used `l` here again and created a new token (`widthForDialogAsLarge`) for it which have a different value per theme.

This is a temporary measure: once the new theme styles are merged to the default theme, it could be reverted (and changed to `13`, which is the new constraint default).

emmenko:
@CarlosCortizasCT Is the padding of the dialog correct?

image

CarlosCortizasCT:
> @CarlosCortizasCT Is the padding of the dialog correct?
> 
> image

Dialogs paddings are not addressed in this PR since we can't adjust them using only design tokens.
Paddings are currently managed with the `Card` component so that's something to be worked on in the follow-up PR where we will address the list mentioned in the description of this PR.

emmenko:
Got it, thanks for double checking!

packages/application-components/src/theming.ts
```diff
@@ -0,0 +1,66 @@
...
+    backgroundColorForTabularMainPageContent: uiKitDesignTokens.colorSurface,
+    borderBottomForTabularPageHeader: `1px solid ${uiKitDesignTokens.colorNeutral90}`,
+    borderColorForDialogDivider: uiKitDesignTokens.colorNeutral90,
+    borderColorForModalPageHeaderDivider: uiKitDesignTokens.colorNeutral90,
+    borderColorForModalTopBarWhenSurface: uiKitDesignTokens.colorSurface,
+    heightForTab: '2px',
+    marginBottomForPageTopBar: uiKitDesignTokens.spacingL,
+    marginForModalPageHeader: `0 40px`,
+    marginLeftForModalPageHeaderControls: uiKitDesignTokens.spacingXl,
+    marginTopForPageHeaderControls: uiKitDesignTokens.spacingS,
```
kark, line 47: Just wanted to double check - the value for the default theme and the new one are the same for this token

emmenko:
Checked the VRTs, they look good to me and they show the expected changes ✅

packages/application-components/src/components/dialogs/internals/dialog-container.tsx
emmenko: Can you leave a comment here so that we remember what to do?

packages/application-components/src/theming.ts
```diff
@@ -0,0 +1,66 @@
...
+    backgroundColorForMainPageDivider: uiKitDesignTokens.colorNeutral90,
+    backgroundColorForPageHeader: uiKitDesignTokens.colorSurface,
+    backgroundColorForTabularMainPageContent: uiKitDesignTokens.colorSurface,
+    borderBottomForTabularPageHeader: `1px solid ${uiKitDesignTokens.colorNeutral90}`,
+    borderColorForDialogDivider: uiKitDesignTokens.colorNeutral90,
+    borderColorForModalPageHeaderDivider: uiKitDesignTokens.colorNeutral90,
+    borderColorForModalTopBarWhenSurface: uiKitDesignTokens.colorSurface,
+    heightForTab: '2px',
+    marginBottomForPageTopBar: uiKitDesignTokens.spacingL,
+    marginForModalPageHeader: `0 40px`,
```
emmenko, line 45: Should we move the value `40px` into a variable, as it's used a bunch of times here. Or using `calc`?

visual-testing-app/src/components/theme-switcher/theme-switcher.tsx
emmenko: FYI: we might want to add it in the AppShell as well, making it visible behind the feature flag. Then we can toggle the designs manually instead of restricting the usage to certain projects.

cc @tdeekens As he was asking about this.

emmenko (approved):
Great work, this is so helpful to unlock other changes in App Kit 🙌

Overall looks good, I think we already synced on the next steps anyway.

Thanks 🤗

packages/application-components/src/components/dialogs/internals/dialog-container.tsx
CarlosCortizasCT: Added in this commit: 1876e41835f0bec5584510de97b26e3add5bcf97

packages/application-components/src/theming.ts
CarlosCortizasCT: That's a leftover, Thanks!

Removed in this commit: bfad34b266ca207bde3008b3cb0a77647d2ef128

packages/application-components/src/theming.ts
CarlosCortizasCT: Thanks!

Extracted into a variable in this commit: bfad34b266ca207bde3008b3cb0a77647d2ef128

visual-testing-app/src/components/theme-switcher/theme-switcher.tsx
CarlosCortizasCT: Sure.
I'll do that in a follow-up PR.

Rhotimee (approved):
Nice work 🚀🚀

CarlosCortizasCT:
As a follow-up, I'll update the public docs for dialog components to update the information regarding the `size` property in a new PR.

PR #2876, opened 2022-11-11
refactor(app-shell): render theme based on feature flag

emmenko:
Prepare the theme to be switched based on a feature flag. This will be used initially for testing purposes only.

### DIFF
--- a/packages/application-shell/src/components/application-shell/application-shell.tsx
+++ b/packages/application-shell/src/components/application-shell/application-shell.tsx
@@ -29,7 +29,6 @@ import {
 import { PortalsContainer } from '@commercetools-frontend/application-components';
 import { NotificationsList } from '@commercetools-frontend/react-notifications';
 import { AsyncLocaleData } from '@commercetools-frontend/i18n';
-import { ThemeProvider } from '@commercetools-uikit/design-system';
 import version from '../../version';
 import internalReduxStore from '../../configure-store';
 import { selectProjectKeyFromUrl, getPreviousProjectKey } from '../../utils';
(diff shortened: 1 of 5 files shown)


packages/application-shell/src/components/application-shell/application-shell.tsx
```diff
@@ -243,6 +243,7 @@ export const RestrictedApplication = <
                     defaultFlags={props.defaultFeatureFlags}
                   >
                     <>
+                      <ThemeSwitcher />
```
emmenko, line 246: Moved here to be "within" the feature flags provider.

chloe0592:
Thank you for your work! I was wondering how we do feature flags. :)

chloe0592 (approved):

packages/application-shell/src/components/theme-switcher/theme-switcher.tsx
```diff
@@ -0,0 +1,10 @@
+import { ThemeProvider } from '@commercetools-uikit/design-system';
+import { useFeatureToggle } from '@flopflip/react-broadcast';
+import { UI_REDESIGN } from '../../feature-toggles';
+
+const ThemeSwitcher = () => {
```
CarlosCortizasCT, line 5: **nit**: I would expect a component with this name to actually implement a theme selector. Perhaps we will have it in the future when we implement dark theme.

emmenko: Yeah we will need to refactor things again once we get there.

Rhotimee (approved):
🙌🏽

PR #2877, opened 2022-11-14
fix: to allow disabling Sentry and GTM through environment variable

tdeekens:
#### Summary

Not all environments require Sentry and Google Tag Manager to run. This is usually the case when defining a Custom Application Config without the respective values then being parsed and placed on `window.app`. 

However, in some scenarios we intend to disable Sentry and GTM with a shared Custom Application Config.

#### Description

The values read in a Custom Application Config and defined as `env:FOO` have to be defined otherwise the script will rightfully throw an error. Whenever a value is set as e.g. `process.env.FOO` it will be parsed and replaced as `"null"`. This string is then coerced before putting it onto the application's context for usage.

The booting of Sentry and GTM however, happens outside of React. As such those values are read from `window.app` and are not covered. So `"null"` and `"false"` are possible values on `window.app`. To be able to disable Sentry and GTM via internal CLIs we would like to set their values to `null`. As a result we should handle `"null"` as a case in these few instances where we read from `window.app`.

### DIFF
--- a/.changeset/twenty-wombats-accept.md
+++ b/.changeset/twenty-wombats-accept.md
@@ -1,0 +1,6 @@
+---
+"@commercetools-frontend/application-shell": patch
+"@commercetools-frontend/sentry": patch
+---
+
+Fix to allow disabling Sentry and GTM through setting environment variable to `null`.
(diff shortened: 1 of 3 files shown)


emmenko (approved):
Thank!

As a follow up maybe we can think of introducing optional env placeholders `${env:?FOO}`?

tdeekens:
? As a follow up maybe we can think of introducing optional env placeholders ${env:?FOO}?

Yeah, however this would be the only usage. I think generally we should:

1. Parse/cast the substituted env placeholders
2. Maybe support optional values

However, I am not sure how much optionality is something others also need 🤔

PR #2878, opened 2022-11-14
refactor(sentry): enable performance tracing

emmenko:
This PR enables [Performance monitoring](https://docs.sentry.io/product/performance/) within Sentry, to help use monitoring important metrics within one tool.

As a follow up we can stop sending performance metrics to our own custom endpoint `/metrics`.

PR #2879, opened 2022-11-15
refactor(components-playground): to be built with Vite instead of Gatsby

emmenko:
In #2878 the components-playground website is failing to be built

image

This is likely due to Sentry using the `history` package.

The components-playground website is currently being built using Gatsby which builds using SSR. The error is likely related to (browser) `history` being incompatible with SSR.

However, the components-playground website does not need Gatsby at all, it can be built as a normal app. Therefore, we can refactor it to be built with Vite + TS.


packages/application-components/src/components/public-page-layout/index.ts
```diff
@@ -1 +1,8 @@
-export { default } from './public-page-layout';
+import { lazy } from 'react';
+
+const PublicPageLayout = lazy(
+  () =>
+    import('./public-page-layout' /* webpackChunkName: "public-page-layout" */)
+);
```
emmenko, line 6: This is now code splitted, mostly to avoid strange issues related to loading the backround png image (e.g. vrts, components-playground).

emmenko: Yes that's fine because the AppShell has Suspense already. Also, this specific component is only used in the authentication app so far.

Thanks for asking!

CarlosCortizasCT: Cool 🧊

CarlosCortizasCT (approved):
Awesome work 💪

PR #2881, opened 2022-11-17
refactor: use `workspace:` protocol for linking workspace packages

emmenko:
The `workspace:` protocol is supported by monorepo tools like Yarn and PNPM, as well as Changesets. It's the recommended way of defining dependencies within a workspaces to ensure that no versions from the remote registry are being accidentally used.

https://pnpm.io/workspaces#workspace-protocol-workspace
https://yarnpkg.com/features/protocols/#workspace


tdeekens (approved):

scripts/install-template.mjs
```diff
@@ -0,0 +1,69 @@
...
+);
+const createAppCmdResult = shelljs.exec(
+  [
+    binaryPath,
+    applicationName,
+    `--template=${templateName}`,
+    `--template-version=${branchName}`,
+    `--initial-project-key=${initialProjectKey}`,
+    `--yes`,
+    '--skip-install',
```
CarlosCortizasCT, line 46: I think I'm missing something here, sorry.
Why do we skip dependencies installation?
It seems binary will update the `package.json` of the template to replace the `workspace:*` values with the latest expected version.

CarlosCortizasCT: Ok. Understood.

Many thanks for the explanation 🙌

CarlosCortizasCT (approved):
💯

PR #2883, opened 2022-11-17
refactor: use global resets styles

emmenko:
In preparation to the upcoming UI Kit release.


CarlosCortizasCT:
We have two errors in Percy at the moment:
* Spacing issues when using custom titles in Main pages: this is due to our testing custom titles are using `h2` elements. The new `reset.css` file we're importing from `ui-kit` in the visual testing app resets the margin for all heading elements to 0. This is something we didn't have before in this application. It's an expected change.
* Link color in `PageNotFoud` component: it seems the previous visualization was not correct as now the link has the expected (primary) color.

So I think we can approve these changes.

emmenko:
Thanks for checking. I agree ✅

As for the TS errors, it might be related to dependency version conflicts. We might need to use a resolution for `react-select`. If you need help with that, let me know.

CarlosCortizasCT:
> Thanks for checking. I agree ✅
> 
> As for the TS errors, it might be related to dependency version conflicts. We might need to use a resolution for `react-select`. If you need help with that, let me know.

Yes, `ui-kit` version for `react-select` is newer than the one we have in `application-shell` package.
I updated the latter to be in sync with the former and TS errors are gone. Also, tests are still green and testing manually the playground does not show any issue.

emmenko:
It seems you need to run `yarn` to update the lockfile.

CarlosCortizasCT:
> It seems you need to run `yarn` to update the lockfile.

Yes, I did that but forgot to push the lockfile 🤦 

Pushed in this commit: f17adf3302b79400783164958ff3e9649ef28c34

CarlosCortizasCT (approved):
Awesome! 🎊

emmenko:
Nice! 🥳

PR #2886, opened 2022-11-18
Add missing changesets

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=EakYk5h)

### Summary

Add missing changesets from a previous PR: #2883 .

### DIFF
--- a/.changeset/friendly-pears-type.md
+++ b/.changeset/friendly-pears-type.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': minor
+---
+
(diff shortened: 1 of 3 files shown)


emmenko:
Thanks! Btw, I think one changeset is enough for all affected packages.

CarlosCortizasCT:
I wanted to separate the packages where we only update `ui-kit` dependencies from the ones where we also refactor the global styles.

Would it be ok to keep two changesets or is it better to have just one where the description says both things?

emmenko (approved):

PR #2894, opened 2022-11-25
refactor(navbar): soft redesign changes

emmenko:
FYI: I already reviewed the changes with @FilPob and they have been approved.

image image image


packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -96,6 +82,16 @@ type TProjectPermissions = {
...
 
+const HeartIcon = lazy(() => import('./legacy-icons/heart'));
+const PaperclipIcon = lazy(() => import('./legacy-icons/paperclip'));
+const PluginIcon = lazy(() => import('./legacy-icons/plugin'));
+const RocketIcon = lazy(() => import('./legacy-icons/rocket'));
+const StarIcon = lazy(() => import('./legacy-icons/star'));
+const ConnectedSquareIcon = lazy(
+  () => import('./legacy-icons/connected-square')
+);
+const WorldIcon = lazy(() => import('./legacy-icons/world'));
```
emmenko, line 93: Unrelated changes to the redesign but I thought to code split the legacy icons to not include them in the bundle by default.

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -628,7 +590,8 @@ const NavBar = <AdditionalEnvironmentProperties extends {}>(
     handleToggleItem,
     handleToggleMenu,
     shouldCloseMenuFly,
-    allApplicationNavbarMenu,
+    allInternalApplicationsNavbarMenu,
+    allCustomApplicationsNavbarMenu,
```
emmenko, line 594: I opted to split up the internal from external apps as there is a divider between the two groups.

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -665,7 +628,29 @@ const NavBar = <AdditionalEnvironmentProperties extends {}>(
...
+                projectPermissions={projectPermissions}
+                menuVisibilities={menuVisibilities}
+                applicationLocale={props.applicationLocale}
+                projectKey={props.projectKey}
+                useFullRedirectsForLinks={useFullRedirectsForLinks}
+                onMenuItemClick={props.onMenuItemClick}
+              />
+            );
+          })}
+          <MenuItemDivider />
```
emmenko, line 652: Here's the divider

packages/application-shell/src/hooks/use-applications-menu/fetch-applications-menu.proxy.graphql
```diff
@@ -1,7 +1,6 @@
 query FetchApplicationsMenu {
   applicationsMenu {
     navBar {
-      shouldRenderDivider
```
emmenko, line 4: Not needed anymore.

stylelint.config.js
```diff
@@ -33,7 +33,7 @@ module.exports = {
     'function-name-case': null,
     'comment-empty-line-before': null,
     'csstools/value-no-unknown-custom-properties': [
-      true,
+      null,
```
emmenko, line 36: Disabled, as the appkit tokens are not included here

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -108,23 +104,6 @@ const IconSwitcher = ({ icon, ...iconProps }: IconSwitcherProps) => {
...
-    case 'TagMultiIcon':
-      return <TagMultiIcon {...iconProps} />;
-    case 'CartIcon':
-      return <CartIcon {...iconProps} />;
-    case 'BoxIcon':
-      return <BoxIcon {...iconProps} />;
-    case 'GearIcon':
-      return <GearIcon {...iconProps} />;
-    case 'ListWithSearchIcon':
-      return <ListWithSearchIcon {...iconProps} />;
```
CarlosCortizasCT, line 127: Are these icons we don't use anymore?

packages/application-shell/src/constants.ts
```diff
@@ -1,7 +1,7 @@
 export const DIMENSIONS = {
   header: '43px',
   navMenu: '64px',
-  navMenuExpanded: '200px',
+  navMenuExpanded: '254px',
```
CarlosCortizasCT, line 4: It seems this value should be `245` since that's the one we use in the navbar:
image

Otherwise, the overlay does not cover all the expected space:
image

emmenko: Good catch!! I mixed up the numbers 🙈

packages/application-shell/src/components/navbar/navbar.tsx
emmenko: These are icons for our internal apps. Before the Org-level release we used to reference icons by their name (hence the logic here), however now they are embedded into the app configs / menus.

The logic here is just for backwards compatibility. I refactored it a bit to code split it so that it's only loaded when actually needed.

packages/application-shell/src/constants.ts
emmenko: Fixed here: https://github.com/commercetools/merchant-center-application-kit/pull/2894/commits/0df8c27bae683c743b197493e7a4f62119845519

packages/application-shell/src/components/navbar/navbar.tsx
CarlosCortizasCT: Yes, the code splitting is a great idea.
I was wondering about the icons which have been removed from the static import list but do not exists in the new lazy imports:
* TreeStructureIcon
* UserFilledIcon
* SpeedometerIcon
* TagMultiIcon
* CartIcon
* BoxIcon
* GearIcon
* ListWithSearchIcon

emmenko: Ah yes sorry. These can be removed as the internal apps not embed them (I made a PR about that last week).

CarlosCortizasCT (approved):
💯

tdeekens (approved):
Nice work. The CSS variables and math is what it is I guess.

PR #2897, opened 2022-11-29
refactor(navbar): define design tokens for colors

emmenko:
image image


packages/application-shell/src/components/navbar/navbar.mod.css
```diff
@@ -70,15 +73,24 @@
...
   height: var(--icon-size);
 }
+.list-item .icon > svg *:not([fill='none']) {
+  /* stylelint-disable-next-line csstools/value-no-unknown-custom-properties */
+  fill: var(--color-for-navbar-icon);
+}
+.list-item .icon__active > svg *:not([fill='none']) {
+  /* stylelint-disable-next-line csstools/value-no-unknown-custom-properties */
+  fill: var(--color-for-navbar-icon-when-active);
+}
```
emmenko, line 83: We're overwriting the fill colors directly here (similarly to what the icon itself does internally) to be able to use the design tokens, even though the icon component has a color prop.

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -702,6 +699,20 @@ const NavBar = <AdditionalEnvironmentProperties extends {}>(
...
+              height: 32px;
+              width: 100%;
+              background: linear-gradient(
+                0deg,
+                ${appkitDesignTokens.backgroundColorForNavbar} 0%,
+                rgba(0, 0, 0, 0) 100%
+              );
+              z-index: 1;
+            `}
+          />
```
CarlosCortizasCT, line 715: **nit**: perhaps we could have a local styled component for this. Perhaps something like `MenuItemContainer` or `MenuItemWraper`

packages/application-components/src/theming.ts
```diff
@@ -4,18 +4,33 @@ import {
 } from '@commercetools-uikit/design-system';
 
 const appKitSpacing55 = '40px';
+// TODO: move to uikit?
```
CarlosCortizasCT, line 7: I also agree 👍

packages/application-shell/src/components/navbar/navbar.mod.css
```diff
@@ -21,7 +21,8 @@
   flex-grow: 1;
   width: var(--width-left-navigation);
   z-index: 19999;
-  background: var(--color-accent);
+  /* stylelint-disable-next-line csstools/value-no-unknown-custom-properties */
```
CarlosCortizasCT, line 24: Perhaps we can update the [stylelint config](https://github.com/commercetools/merchant-center-application-kit/blob/main/stylelint.config.js#L39) to include the tokens from the `theming.ts` file? (although we would need to read a typescript file from a javascript one)

CarlosCortizasCT (approved):

packages/application-shell/src/components/navbar/navbar.mod.css
emmenko: I thought about it but I'm not sure it's worth the effort, as we either need to maintain the list manually or generate some kind of JSON file.

Keep in mind that when we refactor the navbar, we'll be using CSS-in-JS, so the stylelint config (for the appkit repo) is not needed anymore.

packages/application-shell/src/components/navbar/navbar.tsx
emmenko: See https://github.com/commercetools/merchant-center-application-kit/pull/2897/commits/996fb1f9c79809c5dd585bf4345c0032d2b47a8b

kark (approved):

PR #2898, opened 2022-11-29
Restore legacy app named icons

CarlosCortizasCT:
[workerB](https://workerb.linearb.io/v2/badge/collaboration-page?magicLinkId=Ih1JkWt)

#### Summary

Restore legacy app named icons

#### Description

This is a consequence of this PR: #2894.
Over there we removed some legacy icons used in legacy custom applications where you could configure the application icon (to be used in the navigation bar) using a code name.
The new policy is to provide the actual SVG at configuration level and we made an update to MC applications to set it up like this (github.com/commercetools/merchant-center-frontend/pull/13243/).

We catch an issue when deploying the new versions of the applications since their configuration is cached in the MC proxy for an hour so, after publishing them with the new version of `app-kit` where we didn't support legacy named icons anymore, as the MC proxy would have the previous applications configurations cached, the legacy icons names would be used but the new version of `app-kit` wouldn't know anything about theme.

To prevent such issue or making the deployment more complex (having to coordinated frontend applications deployments with MC proxy restart), we're adding those named icons back.
We can finally remove them once those deployments are done.

### DIFF
--- a/.changeset/tough-avocados-shop.md
+++ b/.changeset/tough-avocados-shop.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': patch
+---
+
+Restore legacy application named icons.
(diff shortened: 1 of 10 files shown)


emmenko (approved):
Thanks!

tdeekens (approved):

PR #2912, opened 2022-12-12
feat(application-shell): allow passing `themesOverrides` as `props`

tdeekens:
#### Summary

This adds a `themesOverrides` `prop` to the `ApplicationShell` allowing to override and pass down theming into an application. It's used now for the redesign but could enable other custom branding too.

#### Description

Assuming the requirement of an application wanting to override a theme while adding custom theme one could specify an entry point as 

```js
const EntryPoint = () => (
  <ApplicationShell
     themesOverrides={themesOverrides}
  >
    <Application>
      <AsyncRoutes />
    </Application>
  </ApplicationShell>
);
```

Where the `themesOverrides` would be 

```js
import {
  transformTokensToCssVarsReferences,
  designTokens as uiKitDesignTokens,

export const themesOverrides = {
  default: {
    backgroundColorForSaveToolbar: uiKitDesignTokens.colorGrey,
  },
  test: {
    backgroundColorForSaveToolbar: uiKitDesignTokens.colorAccent,
  },
};

export const designTokens = transformTokensToCssVarsReferences(
  themesOverrides.default,
  { includeDefaultValue: false }
);
```

These properties would then be accessible inside the application and can be used as 

```css
.container {
   background-color: var(--background-color-for-save-toolbar)
}
```

### DIFF
--- a/packages/application-shell/src/components/theme-switcher/theme-switcher.tsx
+++ b/packages/application-shell/src/components/theme-switcher/theme-switcher.tsx
@@ -5,10 +5,10 @@ import { themesOverrides as appKitThemesOverrides } from '@commercetools-fronten
 import { UI_REDESIGN } from '../../feature-toggles';
 
 type TThemeOverrides = Record<string, string>;
-export type TOverwridesPerTheme = Record<string, TThemeOverrides>;
+export type TOverridesPerTheme = Record<string, TThemeOverrides>;
 
 type ThemeSwitcherProps = {
-  themesOverrides?: TOverwridesPerTheme;
+  themesOverrides?: TOverridesPerTheme;
 };
 const ThemeSwitcher = (props: ThemeSwitcherProps) => {
   const isNewThemeEnabled = useFeatureToggle(UI_REDESIGN);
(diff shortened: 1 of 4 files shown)


packages/application-shell/src/components/theme-switcher/theme-switcher.tsx
```diff
@@ -1,14 +1,27 @@
+import { useMemo } from 'react';
 import { ThemeProvider } from '@commercetools-uikit/design-system';
 import { useFeatureToggle } from '@flopflip/react-broadcast';
-import { themesOverrides } from '@commercetools-frontend/application-components';
+import { themesOverrides as appKitThemesOverrides } from '@commercetools-frontend/application-components';
 import { UI_REDESIGN } from '../../feature-toggles';
 
-const ThemeSwitcher = () => {
+type TThemeOverrides = Record<string, string>;
+export type TOverwridesPerTheme = Record<string, TThemeOverrides>;
```
CarlosCortizasCT, line 8: Perhaps we can name this type `TOverridesPerTheme`?

CarlosCortizasCT:
Actually, this would only would be needed if you need to reference a design token from a JS/TS file.
When using CSS modules only, it would be useless.

image

tdeekens:
Take this as a suggestion. We can leave it as we have an alternative. Just so we have it also discussed.

PR #2914, opened 2022-12-13
fix: remove deprecated and experimental support for `Feature-Policy` header

emmenko:
The (experimental) `Feature-Policy` header is deprecated and only the `Permissions-Policy` header should be used instead.

image

image


CarlosCortizasCT (approved):
👍

tdeekens (approved):

kark (approved):

PR #2917, opened 2022-12-16
refactor: apply new design tokens for appbar and user menu

emmenko:
image

I already checked with @FilPob and is approved from his side.

### DIFF
--- a/packages/application-components/src/theming.ts
+++ b/packages/application-components/src/theming.ts
@@ -12,6 +12,7 @@ export const themesOverrides = {
     backgroundColorForMainPageDivider: uiKitDesignTokens.colorNeutral60,
     backgroundColorForPageHeader: uiKitDesignTokens.colorNeutral95,
     backgroundColorForTabularMainPageContent: uiKitDesignTokens.colorNeutral95,
+    backgroundColorForUserMenuItemWhenHovered: uiKitDesignTokens.colorNeutral90,
     borderBottomForTabularPageHeader: `1px solid ${uiKitDesignTokens.colorNeutral}`,
     borderColorForDialogDivider: uiKitDesignTokens.colorNeutral,
     borderColorForModalPageHeaderDivider: uiKitDesignTokens.colorNeutral60,
@@ -22,6 +23,7 @@ export const themesOverrides = {
     marginBottomForPageTopBar: uiKitDesignTokens.spacing20,
     marginForDialogContainerContents: `${uiKitDesignTokens.spacing30} 0`,
     marginForModalPageHeader: `0 ${uiKitDesignTokens.spacing30}`,
+    marginForUserMenuItem: '0',
     marginLeftForModalPageHeaderControls: uiKitDesignTokens.spacing30,
     marginRightForAppbar: uiKitDesignTokens.spacing30,
     marginTopForDialogFooter: uiKitDesignTokens.spacing30,
@@ -46,6 +48,7 @@ export const themesOverrides = {
     backgroundColorForMainPageDivider: uiKitDesignTokens.colorNeutral90,
     backgroundColorForPageHeader: uiKitDesignTokens.colorSurface,
     backgroundColorForTabularMainPageContent: uiKitDesignTokens.colorSurface,
+    backgroundColorForUserMenuItemWhenHovered: uiKitDesignTokens.colorNeutral95,
     borderBottomForTabularPageHeader: `1px solid ${uiKitDesignTokens.colorNeutral90}`,
     borderColorForDialogDivider: uiKitDesignTokens.colorNeutral90,
     borderColorForModalPageHeaderDivider: uiKitDesignTokens.colorNeutral90,
@@ -56,6 +59,7 @@ export const themesOverrides = {
     marginBottomForPageTopBar: uiKitDesignTokens.spacing40,
     marginForDialogContainerContents: `${uiKitDesignTokens.spacing30} 0 ${uiKitDesignTokens.spacing50} 0`,
     marginForModalPageHeader: `0 ${appKitSpacing55}`,
+    marginForUserMenuItem: `${uiKitDesignTokens.spacing10} 0`,
     marginLeftForModalPageHeaderControls: uiKitDesignTokens.spacing50,
     marginRightForAppbar: appKitSpacing55,
     marginTopForDialogFooter: uiKitDesignTokens.spacing50,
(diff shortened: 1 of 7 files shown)


packages/application-components/src/theming.ts
```diff
@@ -4,6 +4,7 @@ import {
 } from '@commercetools-uikit/design-system';
 
 const appKitSpacing55 = '40px';
+const newShadow1 = '0px 2px 5px 0px rgba(0, 0, 0, 0.15)';
```
emmenko, line 7: Would be moved to uikit, as it's one of the new shadows.

packages/application-shell/src/components/project-switcher/project-switcher.tsx
```diff
@@ -166,7 +164,7 @@ const ProjectSwitcher = (props: Props) => {
   return (
     <div
       css={css`
-        width: 225px;
+        width: ${designTokens.constraint6};
```
emmenko, line 167: It's fine to change it in general

packages/application-shell/src/constants.ts
```diff
@@ -1,5 +1,6 @@
 export const DIMENSIONS = {
-  header: '43px',
+  header: '56px',
```
emmenko, line 2: It's fine to change it in general

packages/application-shell/src/components/app-bar/app-bar.tsx
```diff
@@ -103,7 +104,7 @@ const AppBar = (props: Props) => {
           <div
             css={css`
               border-left: 1px ${customProperties.colorNeutral90} solid;
-              height: ${DIMENSIONS.header};
+              height: ${DIMENSIONS.headerItemDivider};
```
emmenko, line 107: It's fine to change it in general

packages/application-shell/src/components/project-switcher/project-switcher.tsx
```diff
@@ -14,7 +14,7 @@ import { reportErrorToSentry } from '@commercetools-frontend/sentry';
 import AccessibleHidden from '@commercetools-uikit/accessible-hidden';
 import SelectInput from '@commercetools-uikit/select-input';
 import { ErrorIcon } from '@commercetools-uikit/icons';
-import { customProperties } from '@commercetools-uikit/design-system';
+import { designTokens } from '@commercetools-uikit/design-system';
```
CarlosCortizasCT, line 17: **nit**: since we can use `ui-kit` design tokens and local ones in `app-kit`, we were aliasing them when importing to make it clear where the used tokens come from ([example](https://github.com/commercetools/merchant-center-application-kit/blob/fb4907897356c21e4b24b277db0df740609db870/packages/application-components/src/components/internals/tabular-page.tsx#L4)).

Perhaps we could use the same convention here.

packages/application-shell/src/components/user-settings-menu/user-settings-menu.tsx
```diff
@@ -1,19 +1,22 @@
-import type { ControllerStateAndHelpers, DownshiftProps } from 'downshift';
-import type { TUser } from '../../types/generated/mc';
-import type { TFetchApplicationsMenuQuery } from '../../types/generated/proxy';
-
 import { ReactNode, useCallback, useEffect, useRef, useState } from 'react';
```
CarlosCortizasCT, line 1: Perhaps we can add the redesign comment to this file:
```
// TODO: @redesign cleanup
```

packages/application-shell/src/components/user-settings-menu/user-settings-menu.tsx
```diff
@@ -115,38 +133,108 @@ const renderLabel = (
...
-  color: ${customProperties.colorSolid};
+  color: ${designTokens.colorSolid};
 
   :hover {
-    background-color: ${customProperties.colorNeutral90};
+    ${(props) =>
+      `background-color: ${
+        props.theme === 'test'
+          ? designTokens.colorNeutral95
+          : designTokens.colorNeutral90
```
CarlosCortizasCT, line 143: Maybe we could use a design token here. Something like this:
```suggestion
      `background-color: ${appKitDesignTokens.backgroundColorForUserSettingsMenuItem}`
```

packages/application-shell/src/components/user-settings-menu/user-settings-menu.tsx
```diff
@@ -115,38 +133,108 @@ const renderLabel = (
...
+      };`}
   }
 
   ${(props) =>
     props.hasDivider === true
       ? css`
-          border-bottom: 1px solid ${customProperties.colorNeutral};
+          border-bottom: 1px solid ${designTokens.colorNeutral};
+          margin: ${props.theme === 'test' ? designTokens.spacing10 : '0'} 0;
```
CarlosCortizasCT, line 151: perhaps we could use a design token here.

packages/application-shell/src/components/user-settings-menu/user-settings-menu.tsx
emmenko: Good point, thanks

packages/application-shell/src/components/user-settings-menu/user-settings-menu.tsx
emmenko: Since we need to use the hook anyway, I thought it's easier to use if-else logic instead of creating new tokens as it's only for the migration phase.

Do you think it's fine or do you prefer to use design tokens?

CarlosCortizasCT: I mean, that's the purpose of the tokens but I can see your point.
For sure we can use the current option or even the new `themedValue` helper from the `useTheme` hook to make it simpler, however I think I would have used a token for consistency.

packages/application-shell/src/components/project-switcher/project-switcher.tsx
CarlosCortizasCT: The design tokens import from `app-kit` was aliases in the `user-settings-menu.tsx` file. I think it would also be helpful to alias `ui-kit` design tokens import (something like `uiKitDesignTokens`) in that file and also in this one.

emmenko: Alright, I aliased the imports when we import tokens from both sources, otherwise if it's only uikit I'll prefer to keep the normal import. https://github.com/commercetools/merchant-center-application-kit/pull/2917/commits/4442548f4f89a010acf3e749c331464db2c42a01

CarlosCortizasCT (approved):
💯

kark (approved):

PR #2924, opened 2022-12-23
fix(mc-scripts): support prod sourcemaps when using Vite

emmenko:
Until now Vite (Rollup) had some issues when running production builds with the [sourcemaps option](https://vitejs.dev/config/build-options.html#build-sourcemap) activated.

You can read a bit the thread here: https://github.com/vitejs/vite/issues/2433

However, yesterday an attempt to mitigate this issue was released in Rollup: https://github.com/vitejs/vite/issues/2433#issuecomment-1363786257

The fix is opt-in via the `cache` option of Rollup by turning it off. To quote the outcome:

> The memory still scales proportionately to the number of files, but it should consume less memory per file now.

Let's try it out then!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -251,6 +251,7 @@ jobs:
       - name: Building Playground application
         run: yarn playground:build
         env:
+          NODE_OPTIONS: '--max-old-space-size=6144'
           APP_ID: ${{ secrets.APP_ID_PLAYGROUND }}
           MC_API_URL: ${{ secrets.MC_API_URL }}
           CTP_INITIAL_PROJECT_KEY: ${{ secrets.CYPRESS_PROJECT_KEY }}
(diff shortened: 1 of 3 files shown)


tdeekens (approved):

.github/workflows/main.yml
```diff
@@ -251,6 +251,7 @@ jobs:
       - name: Building Playground application
         run: yarn playground:build
         env:
+          NODE_OPTIONS: '--max-old-space-size=6144'
```
emmenko, line 254: Ok so it seems to work but we need to bump the memory.

Maybe we make it an opt-in option to generate the sourcemaps.

CarlosCortizasCT: I think this is a great idea 👍

CarlosCortizasCT (approved):
👍

PR #2927, opened 2023-01-10
fix(navbar): overlap with links

### LINKED ISSUE #2925: Not able to click on last link/sublink in the nav bar
**Describe the issue**

In Merchant Center sidebar, if there are more links, the user not able to select the last option since the presence of a div which is which is giving the shading.

**To Reproduce**

1. Go to Merchant Center
2. Expand the navbar if it is collapsed
3. Scroll down to the last link (Considering scenario that more links present in navbar)
4. Its difficult to click on the last link/sublink

**Expected behavior**

Should be able to click on last link/sublink without struggling.

**Screenshots**

<div>
    
    
</div

Left image refers to the issue and right image refers to one of the resolution that we can use

**Additional context**

We can resolve this issue by removing that extra div or reducing the height of that div

emmenko:
Fixes #2925

### DIFF
--- a/.changeset/tame-colts-guess.md
+++ b/.changeset/tame-colts-guess.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': patch
+---
+
+Fix bottom spacing to avoid faded container overlap with menu links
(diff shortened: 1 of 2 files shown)


CarlosCortizasCT (approved):
Thanks! 👍

tdeekens (approved):

PR #2933, opened 2023-01-11
Revert re-exporting `@testing-library/user-event` from `application-shell` package

CarlosCortizasCT:
#### Summary

Revert re-exporting `@testing-library/user-event` from `application-shell` package

#### Description

We are reverting the change introduced [here](https://github.com/commercetools/merchant-center-application-kit/pull/2852/) to re-export `@testing-library/user-event` functions from our `application-shell` package.

With that modification we are introducing a breaking change in `app-kit` consumers: anyone developing/maintaining a Custom Application and upgrading `app-kit` to the latest versions will see all their tests fails since we introduced new peer dependencies (`@testing-library/user-event` and `@testing-library/dom`) for our package.

We don't think this is a desirable side effect and also we are implying that all `app-kit` consumers are using or want to use RTL as their testing library. Otherwise, they will be needing to install these dependencies but they wouldn't be using them.

We think having to import the `user-event` functions directly from that package if consumers need it is not such a big deal and our current situation does not seem to benefit `app-kit` consumers but the opposite.


emmenko (approved):
Thanks

PR #2935, opened 2023-01-12
feat: add new no-image image for use with MC redesign

Sarah4VT:
#### Summary

This is a new image needed for the ImageContainer redesign for MC FE.

#### Description

For backwards compatibility, we need to keep the old image in addition to adding the new, so this is a strict addition rather than a replacement.

Old ImageContainer with no image:
image

New ImageContainer with no image:
image


ByronDWall (approved):
Thanks for adding the changeset, approved

emmenko (approved):

PR #2947, opened 2023-01-26
docs: clarify what happens when menu link permissions are empty

emmenko:
There was some confusion about this and I think we should clarify this in the docs as well.


CarlosCortizasCT (approved):
👏

kark (approved):

ddouglasz (approved):
🙌🏾

PR #2951, opened 2023-01-30
docs: improve code snippets for `TabularModalPage` and `InfoModalPage`

### LINKED ISSUE #2950: application-component documentation lists incorrect code example
Have a look here: https://docs.commercetools.com/custom-applications/api-reference/commercetools-frontend-application-components#infomodalpage
The example creates a pageModalState which is never used but refers to a function handleClose that does not exist.

kark:
#### Summary

closes #2950

### DIFF
--- a/website/src/content/api-reference/commercetools-frontend-application-components.mdx
+++ b/website/src/content/api-reference/commercetools-frontend-application-components.mdx
@@ -266,13 +266,13 @@ const Example = () => {
     <InfoModalPage
       title="Lorem ipsum"
       isOpen={true}
-      onClose={handleClose}
+      onClose={pageModalState.closeModal}
       subtitle={<Text.Body>{'Lorem ipsum ...'}</Text.Body>}
       topBarCurrentPathLabel="Lorem ipsum"
       topBarPreviousPathLabel="Back"
     >
       <Text.Body>{'Lorem ipsum ...'}</Text.Body>
-    </InfoModalPage>;
+    </InfoModalPage>
   )
 }
 ```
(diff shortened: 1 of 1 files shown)


tdeekens (approved):

website/src/content/api-reference/commercetools-frontend-application-components.mdx
```diff
@@ -266,13 +266,13 @@ const Example = () => {
     <InfoModalPage
       title="Lorem ipsum"
       isOpen={true}
```
CarlosCortizasCT, line 268: I was wondering whether we should also use the modal state for the `isOpen` property.
Perhaps we can initialize the hook with `true` value and use it here so when we call the `closeModal` function, the modal would actually close.

```jsx
const Example = () => {
  const pageModalState = useModalState(true); // modal open by default

  return (
    <InfoModalPage
      title="Lorem ipsum"
      isOpen'{pageModalState.isModalOpen}
      ...
  );
}
```


What do you think?

kark: I added `isModalOpen` but I'm not sure about `useModalState(true)`. I mean only in terms of consistency, as we don't use it in other code snippets 🤔

CarlosCortizasCT: Ok, if we don't use it in other examples, maybe we should aim for consistency, yes.

It just caught my eye as the default value for the state is _false_ so when using `useModalState` with no params the `InfoModalPage` would not be visible until some code calls `pageModalState.openModal()`

CarlosCortizasCT (approved):
👍

Rhotimee (approved):

PR #2956, opened 2023-02-08
Update `PageContentWide` to render all its children in single-column mode

CarlosCortizasCT:
#### Summary

Update `PageContentWide` to render all its children in single-column mode.

#### Description

The way this component was implemented counted on consumers to provide only one children when using it for rendering a single columns, so it only rendered its first children and skip any other one received.

Since this is causing some issues with its adoption (people having to provide a fragment wrapper over existing children to be used within the `PageContentWide`), we're changing its behaviour for it to render all received children when using a single column.

When using two columns, the component will still render just the first two received children (and skip any other it might receive).

### DIFF
--- a/packages/application-components/src/components/page-content-containers/page-content-wide/page-content-wide.tsx
+++ b/packages/application-components/src/components/page-content-containers/page-content-wide/page-content-wide.tsx
@@ -60,7 +60,7 @@ function PageContentWide(props: TPageContentWide) {
     <Container>
       <Content columns={props.columns} gapSize={props.gapSize}>
         {props.columns === '1' ? (
-          <>{leftChild}</>
+          <>{props.children}</>
         ) : (
           <>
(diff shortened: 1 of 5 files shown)


tdeekens (approved):

ahmehri (approved):

kark (approved):

packages/application-components/src/components/page-content-containers/page-content-wide/page-content-wide.tsx
```diff
@@ -60,7 +60,7 @@ function PageContentWide(props: TPageContentWide) {
     <Container>
       <Content columns={props.columns} gapSize={props.gapSize}>
         {props.columns === '1' ? (
-          <>{leftChild}</>
+          <>{props.children}</>
```
CarlosCortizasCT, line 63: This is the actual change

Issue #2953, opened 2023-02-02
High Severity Dependabot alerts for @commercetools-frontend/mc-scripts


nick-loyalty:
This issue is relevant to Dependabot alerts, we got two High Severity Alerts from Dependabot as it needs two components update 
![Screen Shot 2023-02-02 at 12 12 08 pm](https://user-images.githubusercontent.com/104817403/216206566-11e561f9-f158-487a-93c3-1d53a6429a1a.png)
![Screen Shot 2023-02-02 at 12 12 25 pm](https://user-images.githubusercontent.com/104817403/216206600-0ce932b2-3726-43b6-8c6a-d3c8263e9d16.png)

These two components are both relevant to @commercetools-frontend/mc-scripts when we do npm ls
<img width="582" alt="Screen Shot 2023-02-02 at 1 04 23 pm" src="https://user-images.githubusercontent.com/104817403/216213137-13cbaaae-104d-4615-9dc9-f6e43818aaa6.png">
<img width="508" alt="Screen Shot 2023-02-02 at 1 04 59 pm" src="https://user-images.githubusercontent.com/104817403/216213225-ee748515-5a14-4897-9ed3-181bde19835c.png">
As we can see in the above, with the latest @commercetools-frontend/mc-scripts, it is still using minimatch@3.0.4 which Dependabot needs minimatch@3.0.5 and json5@0.5.1 which Dependabot needs json5@1.0.2.

Can we upgrade the @commercetools-frontend/mc-scripts to support newer version minimatch and json5 ?

kark:
Hi @nick-loyalty,

thank you for creating the issue. 
We'll look into it.

PR #2958, opened 2023-02-09
chore: replace `rcfile` with Prettier API for reading Prettier config

kark:
#### Summary

Partial resolution of #2953.
We use `rcfile@1.0.3` in `create-mc-app` and `mc-scripts` for reading Prettier config. The package is used in the most recent version (but it hasn't been updated by the library authors for years). The package uses `json5@0.5.1` under the hood (and there is [a reported issue with the json5](https://github.com/advisories/GHSA-9c47-m6qq-7p4h) that causes the Dependabot alert).

We can use the native Prettier API for reading Prettier config.


packages/create-mc-app/src/tasks/update-application-constants.ts
```diff
@@ -34,12 +32,15 @@ function replaceEntryPointUriPathInConstants(
...
-      cwd: options.projectDirectoryPath,
-    });
-    const formattedData = prettier.format(result.code + os.EOL, prettierConfig);
+    const prettierConfig = prettier.resolveConfig.sync(
+      options.projectDirectoryPath
+    );
+
+    const formattedData = prettier.format(
+      result.code + os.EOL,
+      prettierConfig!
```
kark, line 42: the provided type is `Options | null` while the expected type is `Options | undefined` so I thought about using type assertion here.

emmenko (approved):
Nice, thanks!! 🥳

PR #2960, opened 2023-02-10
Update deps

emmenko:


packages/application-components/package.json
```diff
@@ -27,35 +27,35 @@
...
+    "@commercetools-uikit/icon-button": "^15.11.2",
+    "@commercetools-uikit/icons": "^15.11.2",
+    "@commercetools-uikit/messages": "^15.11.2",
+    "@commercetools-uikit/primary-button": "^15.11.2",
+    "@commercetools-uikit/secondary-button": "^15.11.2",
+    "@commercetools-uikit/secondary-icon-button": "^15.11.2",
+    "@commercetools-uikit/spacings": "^15.11.2",
+    "@commercetools-uikit/text": "^15.11.2",
+    "@commercetools-uikit/utils": "^15.11.2",
+    "@emotion/react": "11.10.5",
```
CarlosCortizasCT, line 50: Since we are using version `11.10.4` in `ui-kit` and you warned about potential issues derived from using mixed versions in the same project, I was wondering if we should consider updating the dependency here.

image

emmenko: That's fine, we have the version range so we can force the update the latest version here (see https://github.com/commercetools/merchant-center-application-kit/pull/2960/commits/f8a198c3ffface4c6c268d20c5192ecc3a03a2e5)

CarlosCortizasCT (approved):
👏

tdeekens (approved):

kark (approved):
🚀
That were odd type errors:
```bash
Property 'ids' does not exist on type 'Message'.
Property 'locale' does not exist on type 'IntlConfig'.
``` 
Great idea with patching 🎉

Rhotimee (approved):
💪🏽

PR #2961, opened 2023-02-10
chore: update docs-kit to v20

emmenko:
Uses MDX v2


website/src/releases/2021-04/release-v19.mdx
```diff
@@ -46,14 +46,14 @@ The equivalent mapping from `constraint` to `max` is as following:
 
 ## Text
 
-For the `<Text.Headline>` and `<Text.Subheadline>` components, the deprecated `elementType` prop has been removed. Now only the `as` prop is used instead.
+For the `<Headline>` and `<Subheadline>` components, the deprecated `elementType` prop has been removed. Now only the `as` prop is used instead.
```
CarlosCortizasCT, line 49: I was wondering why we are changing these components names since it seems to me the full qualified name is more clear.
Is it because the new MDX version does not support them outside code blocks?

website/package.json
```diff
@@ -6,21 +6,21 @@
   "scripts": {
     "clean": "gatsby clean",
     "check-links": "start-server-and-test 'yarn serve' http://localhost:9000/custom-applications '../.linkcheck/linkcheck.sh --skip-file ../.linkcheck/skip_list.txt http://localhost:9000/custom-applications'",
-    "build": "./scripts/prebuild.sh && yarn generate-icons && gatsby build --prefix-paths --verbose --log-pages && yarn check-links",
+    "build": "./scripts/prebuild.sh && yarn generate-icons && gatsby build --prefix-paths --verbose --log-pages",
```
CarlosCortizasCT, line 9: Is it intended to not check links at build time anymore?

website/src/releases/2021-04/release-v19.mdx
emmenko: It's related to the Vale style checker (the writing-style package). We managed to fix these issues in the code blocks but we still have this issue with the inline code.

https://github.com/commercetools/commercetools-docs-kit/issues/1341

It's a long standing issue with no clear fix unfortunately.

website/package.json
emmenko: Yes we do check it against the preview URL

https://github.com/commercetools/merchant-center-application-kit/blob/182f8f4f7be4936f236161815fc810c4d2f255b8/.github/workflows/docs.yml#L124-L146

Testing links against the local build is not necessary on CI.

kark (approved):

website/src/releases/2021-04/release-v19.mdx
CarlosCortizasCT: Ah! Thanks for the explanation.

I was thinking though, that perhaps we could use the meta-comments suggested in that issue to be able to use the qualified name since I believe our users would better understand where those components come from, but I'm not sure is those comments are only valid for code blocks.

image

emmenko: I managed to make it work (disabling vale for that specific line) by wrapping the vale comment into the mdx v2 comment 😅

https://github.com/commercetools/merchant-center-application-kit/pull/2961/commits/8caf15a9504a13371b27a8fbc9adc0e78836fe27

Thanks for double checking!

website/src/releases/2021-04/release-v19.mdx
```diff
@@ -46,14 +46,18 @@ The equivalent mapping from `constraint` to `max` is as following:
...
 
+{/* <!-- vale on --> */}
+
 ```diff
 - <Text.Headline elementType="h2" intlMessage={message} />
 + <Text.Headline as="h2" intlMessage={message} />
 ```
 
-For the `<Text.Body>` and `<Text.Detail>` components, the deprecated `isInline` prop has been removed. Now only the `as` prop can be used instead.
+For the `<Body>` and `<Detail>` components, the deprecated `isInline` prop has been removed. Now only the `as` prop can be used instead.
```
CarlosCortizasCT, line 60: Perhaps we can use the `vale` directive here as well:
```suggestion
{/* <!-- vale off --> */}

For the `<Text.Body>` and `<Text.Detail>` components, the deprecated `isInline` prop has been removed. Now only the `as` prop can be used instead.

{/* <!-- vale on --> */}
```

emmenko: Ah missed that, thanks

emmenko: https://github.com/commercetools/merchant-center-application-kit/pull/2961/commits/4284137818c99c9961549a82ecac0595fb4a8031

CarlosCortizasCT (approved):
💯

PR #2967, opened 2023-02-20
feat(pageContentWide gapsize): change page content wide gap size from…

ddouglasz:
#### Summary

- Change default gapSize value is from 10 to 20.

- Change VRT to reflect new changes.


tdeekens (approved):

visual-testing-app/src/components/page-content-container-wide/page-content-container-wide.visualroute.tsx
```diff
@@ -60,11 +60,11 @@ export const Component = () => (
           ),
         },
         {
-          name: 'two columns 1/1 with big gap',
-          path: 'two-columns-half-big-gap',
+          name: 'two columns 1/1 with small gap',
```
CarlosCortizasCT, line 63: There's another VRT called `two columns 2/1 with big gap` which we should update.
Also, I guess we should use the `l` size for the default two columns VRTs as now they have the big gap by default.

ddouglasz: Thank you. 

Updated here: https://github.com/commercetools/merchant-center-application-kit/pull/2967/commits/787277dd837da085601b9fc31c463bf44c9a3b84

visual-testing-app/src/components/page-content-container-wide/page-content-container-wide.visualroute.tsx
```diff
@@ -83,7 +83,7 @@ export const Component = () => (
           name: 'two columns 2/1 with big gap',
           path: 'two-columns-two-thirds-big-gap',
```
CarlosCortizasCT, line 84: The name of this test should also be changed
```suggestion
          name: 'two columns 2/1 with small gap',
          path: 'two-columns-two-thirds-small-gap',
```

ddouglasz: Updated here: https://github.com/commercetools/merchant-center-application-kit/pull/2967/commits/860a979efe49a26e94a9cfe903bc990e276dd91a

CarlosCortizasCT:
Checking the failed VRTs I was wondering about the gap when using a single columns or a 1/1 columns.
With the change we applied, all of them now use the big gap (`64px`) by default. That means for single column, that every item will have a `64px` vertical gap:
image.

Also, for `1/1` columns, the gap will also be bigger:
image

@FilPob is this something we expect or should we have different defaults per wide layout?

CarlosCortizasCT:
@ddouglasz, @FilPob I've been reviewing our previous decision to allow rendering all children in `PageContentWide` component when used in single column mode and it was because, when migrating views, the initial setup just swallowed extra children so developers where confused about what was happening, also because there was no warning about it.

I think it makes sense to only render the first children in this situation (`PageContentWide` in single mode) but also show a warning message (in browser console) in case more than one children is received so developers know about this behaviour.
Bear in mind when used in two columns mode, only the first two children are rendered.

I added a commit to implement this here: bdafc975dbc0a92eba574f14a26135b7fc062fc6

So, after all, we are changing the default gap size to be `64px` (instead of `32px`) for two columns layouts (in `PageContentWide` component); both `1/2 + 1/2` and `2/3 + 1/3`.
The gap size will only affect the horizontal spacing.

CarlosCortizasCT:
We need to check the failing VRTs 🧐

emmenko (approved):
💯

emmenko:
I left a comment in the VRTs regarding the button styles issue.

CarlosCortizasCT:
> I left a comment in the VRTs regarding the button styles issue.

Yes, I recursively updated `emotion` dependencies and that solved the issues.

Thanks.

CarlosCortizasCT (approved):

PR #2972, opened 2023-02-24
fix: use base path for playground app

emmenko:
Regression for serving the components playground pages.

image

I don't since when we had this regression to be honest. We switched the components playground to use Vite back in november #2879 but it seems weird to me that nobody noticed that in 3 months.

Anyway, this is a quick fix, we need to reconsider a bit the page structure and how we want to handle the playground components.

### DIFF
--- a/website-components-playground/vite.config.js
+++ b/website-components-playground/vite.config.js
@@ -3,6 +3,7 @@ import { defineConfig } from 'vite';
 
 // https://vitejs.dev/config/
 export default defineConfig({
+  base: '/custom-applications/playground/',
   define: {
     'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV),
     'process.env.CI': JSON.stringify(process.env.CI),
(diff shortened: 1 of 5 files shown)


website/src/components/playground.js
```diff
@@ -11,7 +11,7 @@ const getIframeUrl = (urlPath, isFullScreen = false) => {
 };
 
 const PlaygroundController = (props) => {
-  const [iframeHeight, setIframeHeight] = useState(0);
+  const [iframeHeight, setIframeHeight] = useState(400);
```
emmenko, line 14: This is a temporary fix to show the playground component, since the implementation expects the iframe to set the height.

Somehow the event sent via `postMessage` is not received by the parent. We'll need to find a better solution then as a follow up.

website-components-playground/vite.config.js
```diff
@@ -3,6 +3,7 @@ import { defineConfig } from 'vite';
 
 // https://vitejs.dev/config/
 export default defineConfig({
+  base: '/custom-applications/playground/',
```
emmenko, line 6: This is the location where the production assets are hosted, therefore we need to instruct Vite to use that as the base path.

kark (approved):
Thanks for the hotfix!

CarlosCortizasCT (approved):
👍

ddouglasz (approved):
💯

PR #2979, opened 2023-03-03
fix: refine engine version requirements

emmenko:
Most of our packages have the `engines` version requirements.

The `create-mc-app` package has the allowed versions listed as `14.x || 16.x || 17.x`, which effectively prevents installations from anyone using Node v18+.

image

Last year we fixed that by preventing v15 to be installed (see #2487 ), however we need to be forwards compatible.

Therefore, we can simply allow v14, skip v15 and allow all versions above that. ESLint has a similar version range (https://github.com/eslint/eslint/blob/b09a512107249a4eb19ef5a37b0bd672266eafdb/package.json#L174).


kark (approved):
💯
Worked great in node `v18.9.1`

PR #2980, opened 2023-03-03
refactor(templates): refine required dependencies

emmenko:
Testing out pnpm on a starter template yielded the following changes to ensure the dependencies resolve correctly.


application-templates/starter-typescript/src/components/channel-details/channel-details.spec.tsx
```diff
@@ -8,16 +8,13 @@ import {
   mapResourceAccessToAppliedPermissions,
   type TRenderAppWithReduxOptions,
 } from '@commercetools-frontend/application-shell/test-utils';
-import { reportErrorToSentry } from '@commercetools-frontend/sentry';
```
emmenko, line 11: We use sentry internally (not sure if a customer uses that), no need to have it included by default in the template then.

application-templates/starter-typescript/.gitignore
```diff
@@ -1,4 +1,5 @@
 node_modules
+public
```
kark, line 2: Perhaps we could do the same in the `@commercetools-applications/merchant-center-template-starter`

kark (approved):
Thanks 💯

ddouglasz (approved):
👏

CarlosCortizasCT (approved):

application-templates/starter-typescript/.gitignore
emmenko: Ah yes, thanks for noticing!

PR #2984, opened 2023-03-07
fix(sentry): make it compatible with SSR

emmenko:
Some of the packages of App Kit are used outside of the MC and they use a framework that does SSR.

TL;DR: whenever the Sentry package is loaded by some of the App Kit packages, it imports our custom `browser-history` object, which then fails as there is no DOM when doing SSR.

image

To fix that, we need to make sure that we only load the history object when the DOM is defined.


packages/sentry/src/sentry.ts
```diff
@@ -74,24 +79,36 @@ export const redactUnsafeEventFields = (event: Event) => {
...
+        onunhandledrejection: false,
+        onerror: false,
+      }),
+    ];
+
+    if (canUseDOM) {
+      // Require the browser history on runtime. This allows the package to be used on SSR.
+      const {
+        default: browserHistory,
+      } = require('@commercetools-frontend/browser-history');
```
kark, line 93: Just wondering is we shouldn't use the ESM `import` here 🤔

CarlosCortizasCT (approved):
👍

packages/sentry/src/sentry.ts
kark: I get it, but the code within this if statement block runs in the browser. Can CJS `required` be used there?

CarlosCortizasCT: Sure, but I think that `require` is transformed in compilation time the same as most imports when the build target is, for instance, ES5.

kark: I'm not sure, but I think the failing E2E tests are impacted exactly by this. The `playground` app crashes:
 
Screenshot 2023-03-07 at 14 16 03

emmenko: You're right, `require` doesn't work with Vite because it expects ESM.

I have to rethink the approach then 🤔

packages/actions-global/src/actions/handle-action-error.ts
```diff
@@ -49,11 +49,7 @@ export default function handleActionError(error: ActionError) {
 
     // When unauthorized, log the user out.
     if (error.statusCode === STATUS_CODES.UNAUTHORIZED) {
-      // Require the browser history on runtime. This allows the package to be used on SSR.
-      const {
-        default: browserHistory,
-      } = require('@commercetools-frontend/browser-history');
```
emmenko, line 55: This is now fixed as well.

packages/browser-history/src/enhanced-history.ts
```diff
@@ -1,14 +1,23 @@
...
   // Note: the "?" needs to be removed.
   return decode(search.substring(1)) as Q;
 }
 
 export const createEnhancedHistory = withQuery({ parse, stringify: encode });
 
-const browserHistory = createEnhancedHistory(createBrowserHistory());
+const browserHistory = createEnhancedHistory(
+  // For SSR compatibility
+  canUseDOM ? createBrowserHistory() : createMemoryHistory()
```
emmenko, line 20: This does the trick now ⬆️

kark (approved):
💯
Thank you

CarlosCortizasCT (approved):

PR #2985, opened 2023-03-08
refactor: change spacing on page-content-layout

tdeekens:
#### Summary

The `PublicPageLayout` is used internally mainly on the login pages. With the redesign the text alignment on the header and the spacing on the footer (legal message) should change.

There are different ways to achieve this:

1. Add `props` such as `welcomeMessageTextAlignment` and `legalMessageScale`
2. Using `themedValue` for old and new spacings

I wanted to suggest 2. first. I don't see any reason we should allow different text alignments in the header nor different paddings of the legal message to the top.

CleanShot 2023-03-08 at 14 31 51@2x

CleanShot 2023-03-08 at 15 07 45@2x

### DIFF
--- a/packages/application-components/src/components/public-page-layout/public-page-layout.tsx
+++ b/packages/application-components/src/components/public-page-layout/public-page-layout.tsx
@@ -2,7 +2,7 @@ import { FC, ReactNode } from 'react';
 import { css } from '@emotion/react';
 import styled from '@emotion/styled';
 import CommercetoolsLogoSvg from '@commercetools-frontend/assets/logos/commercetools_primary-logo_horizontal_white-text_RGB.svg';
-import { customProperties } from '@commercetools-uikit/design-system';
+import { customProperties, useTheme } from '@commercetools-uikit/design-system';
 import Spacings from '@commercetools-uikit/spacings';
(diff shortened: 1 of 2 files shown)


emmenko (approved):
👌

packages/application-components/src/components/public-page-layout/public-page-layout.tsx
```diff
@@ -78,16 +80,16 @@ const PublicPageLayout: FC<TProps> = (props) => {
...
+                  text-align: ${themedValue('left', 'center')};
                 `}
               >
                 {props.welcomeMessage}
               </div>
             </Text.Headline>
           </ContainerColumn>
         )}
-        <Spacings.Stack scale="s">
+        <Spacings.Stack scale={themedValue('xs', 'l')}>
```
CarlosCortizasCT, line 91: I wanted to double check if the changed value for the default theme is also intentional.

tdeekens: Nope. An 🦅 👁️ as usual. Thanks!

CarlosCortizasCT (approved):
👍

PR #2992, opened 2023-03-10
refactor: enable imports order rule (repo only)

emmenko:
We've been using the `import/order` plugin successfully in our internal repos and it's normally very useful.

Let's enable it here as well.


kark (approved):
4cokkt

Rhotimee (approved):
💯

PR #2995, opened 2023-03-10
refactor: clean up unnecessary props and types, refactor `<Route>` to always render as children

emmenko:
Extracted from #2821

* The generic type argument `AdditionalEnvironmentProperties` is not really used and needed by the internal components. Let's remove it.
* Renamed some types for component props.
* Refactored `<Route>` components to render as children ([as recommended when upgrading to v6](https://reactrouter.com/en/main/upgrading/v5)).

PR #2821, opened 2022-09-21
Bundles optimizations

emmenko:
Trying to optimize the bundles, for instance by splitting up more chunks and thus reducing bundle sizes.

Still working on it...

PR #2997, opened 2023-03-10
refactor(app-shell): add more code split entry points

emmenko:
Branched off #2995 

Extracted from #2821 

Main goal is to reduce the size of the main bundle by code splitting more components within the App Shell.

**Webpack** (Starter template with code splitted routes)
image

**Vite** (Playground with code splitted routes)
image


.changeset/spotty-points-cross.md
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools-frontend/application-shell': patch
+'@commercetools-local/playground': patch
```
kark, line 3: I think our template packages are missing

kark (approved):
🙌

Rhotimee (approved):
👏🏽👏🏽

packages/application-shell/src/components/application-shell/application-shell.tsx
```diff
@@ -535,11 +71,11 @@ const ApplicationShell = (props: TApplicationShellProps) => {
...
               <Switch>
-                <Route path="/logout">
+                <SuspendedRoute path="/logout">
                   <RedirectToLogout />
-                </Route>
-                <Route>
-                  <RestrictedApplication
+                </SuspendedRoute>
+                <SuspendedRoute>
+                  <ApplicationShellAuthenticated
```
CarlosCortizasCT, line 78: In general I think this is a great idea.

However, I would also share some thoughts about lazy loading component with a very high chance of been rendered as this `ApplicationShellAuthenticated`, or project and locale switchers. By moving them out from the initial bundle, we are introducing a delay in the user experience since now they will need to be fetched when React detects it needs them.

Maybe the latency for a backoffice application is not such a high requirement because most users will have a wired internet connection, but just wanted to share some concerns I had in the past when using lazy loading too much.

ddouglasz (approved):
👏

packages/application-shell/src/components/application-shell/application-shell.tsx
emmenko: Thanks for the feedback. Yes some parts might have a slight delay now. The `<ApplicationShellAuthenticated>` is only rendered if the user is authenticated so I think we can try code splitting it.

In general we can observe the loading behavior / performance and make further adjustments along the way.

Regarding the switcher components, you're right they probably don't need to be code splitted. Let me revert that part.

PR #3008, opened 2023-03-17
chore: upgrade docs-kit to v21

emmenko:
* Reverts the MDX v2 upgrade (from docs-kit v20)
* Upgrade Gatsby to v5
* Set Node.js version to v18


nkuehn (approved):
Thanks @emmenko  

I took the update the actual config directly here so we don't have to do another iteration 
 - GA and hubspot IDs
 - updated beta link
 - updated flags to a combination that makes sense, updated link to flags docs

Rhotimee (approved):

emmenko:
@nkuehn Lovely, thanks!

.github/workflows/main.yml
```diff
@@ -88,7 +88,7 @@ jobs:
         uses: ./.github/actions/ci
 
       - name: Testing Starter application development server
-        run: yarn start-server-and-test 'yarn template-starter:start' 3001 'exit 0'
+        run: yarn start-server-and-test 'yarn template-starter:start' http-get://127.0.0.1:3001 'exit 0'
```
CarlosCortizasCT, line 91: Just curious about where this `http-get` syntax comes from.

CarlosCortizasCT (approved):

.github/workflows/main.yml
nkuehn: It's a special notation of the "start-server-and-test" library https://github.com/bahmutov/start-server-and-test#use-http-get-requests

CarlosCortizasCT: Thanks for the explanation 👍

PR #3012, opened 2023-03-21
feat(config): support loading intl messages that use TX Structured JSON format

emmenko:
As we're looking into adopting [Transifex's Structured JSON format](https://help.transifex.com/en/articles/6220899-structured-json), we can provide better support for it in our tools.

### DIFF
--- a/packages/application-config/src/substitute-variable-placeholders.ts
+++ b/packages/application-config/src/substitute-variable-placeholders.ts
@@ -1,6 +1,15 @@
 import fs from 'fs';
 import type { LoadingConfigOptions } from './types';
 
+type TMessageKeyValue = string;
+// Transifex's Structured JSON format.
+// https://help.transifex.com/en/articles/6220899-structured-json
+type TMessageStructuredJson = {
+  string: string;
+  // ...
+};
+type TMessage = Record<string, TMessageKeyValue | TMessageStructuredJson>;
+
 /**
  * NOTE:
  * Allows variable placeholders. Supported types are:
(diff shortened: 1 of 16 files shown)


dogayuksel (approved):
Hey, thank you for looking into this. Menu substitution should work now. Only concern is Transifex's structured_json implementation supports nesting, though not specified how deep. From [their docs](https://help.transifex.com/en/articles/6220899-structured-json) slightly modified:
```
...
   "key_1":{
      "key_nested_1a":{
         "string":"localize"
      },
      "key_nested_1b":{
         "key_nested_1ba":{
            "string":"translation tools"
         },
         "key_nested_1bb":{
            "string":"some message"
         }
...
```

I don't think it is necessary for us to adopt this, even their example looks a bit confusing. However we should mention it on the docs that our `structured_json` approach expects a flat list, so that custom application customers know what to expect.

packages/application-config/src/substitute-variable-placeholders.ts
```diff
@@ -31,6 +40,11 @@ const isIntlVariablePlaceholder = (valueOfPlaceholder: string) =>
 const isFilePathVariablePlaceholder = (valueOfPlaceholder: string) =>
   Boolean(valueOfPlaceholder.match(filePathRefSyntax));
 
+const isStructuredJson = (
```
CarlosCortizasCT, line 43: Could we maybe use the new exported function from the `i18n` package instead?
(_maybe we don't want to add that dependency to this package_)

CarlosCortizasCT (approved):
Cool 😎

tdeekens (approved):
Thanks. This should do the trick.

packages/application-config/src/substitute-variable-placeholders.ts
emmenko: Yeah I didn't want to include an extra dependency. I think it's fine to have it duplicated for now.

emmenko:
> However we should mention it on the docs that our structured_json approach expects a flat list, so that custom application customers know what to expect.

Thanks @dogayuksel , good point.

We'll update the docs as a follow up, first I wanted us to adopt it internally and make sure it works properly.

PR #3013, opened 2023-03-22
Update AWS link in S3 deployment examples

melindafekete:
#### Summary

Updating an AWS link that points to the billing/signup page because we are getting 401 errors from the link checker.


emmenko (approved):
Thanks

PR #3019, opened 2023-03-30
feat(app-shell): add hook `useRoutesCreator`

emmenko:
Making the routes abstraction implementation available to everyone.

### DIFF
--- a/.changeset/small-walls-live.md
+++ b/.changeset/small-walls-live.md
@@ -6,7 +6,7 @@ New React hook `useRoutesCreator`.
 
 Managing routes within an application can be tricky and error prone, for example when navigating back and forth within nested routes.
 
-To make is easier to manage and consume routes throughout the application, we implemented an abstraction on top of the routing functionality.
+To make it easier to manage and consume routes throughout the application, we implemented an abstraction on top of the routing functionality.
 
 The hook `useRoutesCreator` can be used to configure routes for your application. We recommend defining the routes in a hook `useRoutes` as following:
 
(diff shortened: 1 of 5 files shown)


kark (approved):
Awesome! Thank you!

Rhotimee (approved):
Nice 🙌🏽

.changeset/small-walls-live.md
```diff
@@ -0,0 +1,59 @@
+---
+'@commercetools-frontend/application-shell': minor
+---
+
+New React hook `useRoutesCreator`.
+
+Managing routes within an application can be tricky and error prone, for example when navigating back and forth within nested routes.
+
+To make is easier to manage and consume routes throughout the application, we implemented an abstraction on top of the routing functionality.
```
CarlosCortizasCT, line 9:
```suggestion
To make it easier to manage and consume routes throughout the application, we implemented an abstraction on top of the routing functionality.
```

.changeset/small-walls-live.md
```diff
@@ -0,0 +1,59 @@
...
+  - `getUrl`: returns the computed URL based on the route parameters. You should use this for links or when you need the real location path.
+  - `go`: navigates to the given path.
+
+If a route has parameters you should pass the parameter value when calling the `getUrl` or `go` methods.
+
+```ts
+routes.heroDetail.go({ id: '123' });
+```
+
+In a Custom Application the `projectKey` value is always implied and does not need to be provided when calling `getUrl` or `go`. However, it must be specified when configuring the route path.
```
CarlosCortizasCT, line 45: > However, it must be specified when configuring the route path.

What we mean here is that we must use a placeholder in the route configuration, right?
```js
`/:projectKey/${entryPointUriPath}`
```

If that's the case, I personally found it a little confusing if I should use a specific project key.

CarlosCortizasCT (approved):
💯

tdeekens (approved):
Thanks. Good idea. As far as I recall there are at times 1-2 very slight deviations internally on this module. Let's see how we can tackle them once we adopt this.

.changeset/small-walls-live.md
emmenko: This should do it: https://github.com/commercetools/merchant-center-application-kit/pull/3019/commits/2581e8a6e48e79faaa40ae3f9f2ce070426a170a

CarlosCortizasCT: It's now clear to me.

Thanks a lot 👍

PR #3024, opened 2023-04-03
fix(app-shell): spacing for long label in menu items

emmenko:
For long labels, when the menu is collapsed and hovered, the main menu label overflows with the submenu items.

Before

Screenshot 2023-04-03 at 12 49 40

Due to limitations in how the navbar menu is implemented, we need a workaround to fix this edge case. We can set a max height for the main menu label to approx 2 lines, after which it gets truncated. This way we control the spacing with the submenu items.

The truncation is done using the [`-webkit-line-clamp`](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp) CSS property, which is [supported by all browsers](https://caniuse.com/css-line-clamp).

After

Screenshot 2023-04-03 at 12 46 32
Screenshot 2023-04-03 at 12 46 40

### DIFF
--- a/.changeset/weak-poets-drive.md
+++ b/.changeset/weak-poets-drive.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': patch
+---
+
+Fix navbar menu spacing for long labels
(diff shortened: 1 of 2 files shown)


Rhotimee (approved):
🙌🏽

ddouglasz (approved):
👏

chloe0592 (approved):

emmenko:
@kark This should fix the issue you noticed: https://github.com/commercetools/merchant-center-application-kit/pull/3024/commits/b25447cc0510fc6b5d8b7117cf47b766fa05d0f6

Can you double check? Thanks 🤗

PR #3026, opened 2023-04-04
[Breaking] Upgrade to Jest `v29`

emmenko:
Time to start upgrading some of our dependencies to major versions.


packages-backend/express/src/middlewares/session-middleware.spec.ts
```diff
@@ -195,7 +195,7 @@ describe.each`
 describe('when audience is missing', () => {
   it('should throw a validation error', () => {
     // @ts-ignore
-    expect(() => createSessionMiddleware({})).toThrowError(
+    expect(() => createSessionMiddleware({})).toThrow(
```
CarlosCortizasCT, line 198: Is this a mandatory change? 
I'm wondering if users updating Jest would also need to update their tests like this (and the other updates we needed to apply).
If that's so, maybe we could also mention it in the changelog.

kark (approved):
Thank you 🎉

packages-backend/express/src/middlewares/session-middleware.spec.ts
emmenko: For reference, these changes are due to ESLint errors

image

https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-alias-methods.md

PR #3044, opened 2023-04-19
chore: update ui-kit dependencies

kark:


tdeekens (approved):

kark:
The difference in VRTs is the effect of [this change](https://github.com/commercetools/ui-kit/pull/2487). The change was made for the old and the new theme.

PR #3045, opened 2023-04-19
docs: Fix formatting to avoid table content overflow

Anshuman71:
#### Summary

As pointed out by peers, the table on [@/application-shell](https://docs.commercetools.com/custom-applications/api-reference/commercetools-frontend-application-shell#options) page overflows it's parent container. Moreover, the content makes the table look sparse as the `description` column is densely packed.

#### Description

As a solution, we'll reformat the tables on this page for better readability and fix the overflow issues as a result.

**Preview**: https://appkit-pr-3045.commercetools.vercel.app/custom-applications/api-reference/commercetools-frontend-application-shell

### DIFF
--- a/website/src/content/api-reference/commercetools-frontend-application-shell.mdx
+++ b/website/src/content/api-reference/commercetools-frontend-application-shell.mdx
@@ -63,81 +63,21 @@ const EntryPoint = () => (
 export default EntryPoint;
 ```
 
-### Properties original
+### Properties
 
-| Props                            | Type               |        Required        | Description                                                                                                                                                                                                                                                                                                                                                                      |
-| -------------------------------- | ------------------ | :---------------------: | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


Anshuman71:
Hi @emmenko,
I added two alternatives to the table. You’ll see one original, 1 with ChildSectionNav, and 1 with Table+description. 
I’ve only done it for one table once we finalize the variation. I’ll replicate the changes to all other tables on this page.

emmenko:
Thanks!

The one with the ChildNav I don't like it much to be honest. Same for the "Table + description". It feels too detached and not how you would expect component props to be documented.

Could you try using the table but with only two columns? The left column contains the prop name, the type and if it's required. The right column contains the description.

In any case, this is a temporary solution so we don't need to go crazy about it.

Anshuman71:
@emmenko, the two-column layout looks good to me for an interim solution. Can you take a look today, if it's good?

emmenko:
I pushed a commit to try the single column approach (https://github.com/commercetools/merchant-center-application-kit/pull/3045/commits/c6e5e6de5654883a9d13e88be41072fea004f5ca) as I feel like the two columns is still too cluttered.

| Single column | Two columns |
| --- | --- |
| image | image |

emmenko (approved):
I think this is an acceptable solution for the time being, until we have a more appropriate component.

Thanks a lot for contributing and trying out different variations 🤗

PR #3052, opened 2023-04-21
test: prevent regressions with scrollable menu

kark:
#### Summary

Having in mind [the regression we had yesterday](https://github.com/commercetools/merchant-center-application-kit/pull/3046) with the pop up submenu when the navbar is collapsed I thought about adding a test to prevent this in the future.

### DIFF
--- a/packages/application-shell/src/components/navbar/menu-items.tsx
+++ b/packages/application-shell/src/components/navbar/menu-items.tsx
@@ -246,7 +246,7 @@ type MenuItemProps = {
 };
 const MenuItem = (props: MenuItemProps) => (
   <li
-    role="menu-item"
+    role="menuitem"
     className={classnames(styles['list-item'], {
       [styles.item__active]: props.isActive,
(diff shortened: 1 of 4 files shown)


packages/application-shell/src/components/navbar/menu-items.tsx
```diff
@@ -246,7 +246,7 @@ type MenuItemProps = {
 };
 const MenuItem = (props: MenuItemProps) => (
   <li
-    role="menu-item"
+    role="menuitem"
```
kark, line 249: RTL `findByRole` wouldn't work otherwise. 
([MDN ref](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menuitem_role))
Screenshot 2023-04-24 at 15 39 17

emmenko (approved):
Thanks!

CarlosCortizasCT (commented):
Thanks! 👏

PR #3066, opened 2023-05-05
Add placeholder container above the top navigation

jaikumar-tj:
This PR adds a new placeholder container for rendering something above the top navigation (for internal usage only).

### DIFF
--- a/.changeset/large-bikes-boil.md
+++ b/.changeset/large-bikes-boil.md
@@ -2,4 +2,4 @@
 '@commercetools-frontend/application-shell': minor
 ---
 
-Adds a portal above `top-navigation` to inject `staffbar`
+Adds an empty element above the `top-navigation`. 
--- a/.changeset/large-bikes-boil.md
+++ b/.changeset/large-bikes-boil.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': minor
+---
+
+Adds a portal above `top-navigation` to inject `staffbar`
--- a/packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
+++ b/packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
@@ -202,6 +202,13 @@ export const ApplicationShellAuthenticated = (
                     defaultFlags={props.defaultFeatureFlags}
                   >
                     <>
+                      <div
+                        id={'staff-bar'}
+                        css={css`
+                          position: relative;
+                          z-index: 30000;
+                        `}
+                      />
                       <ThemeSwitcher />
                       <VersionTracker />
                       {/* NOTE: the requests in flight loader will render a loading
@@ -275,7 +282,6 @@ export const ApplicationShellAuthenticated = (
                             );
                           }}
                         </Route>
-
                         <header
                           css={css`
                             grid-row: 2;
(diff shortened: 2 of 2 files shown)


packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
```diff
@@ -202,6 +202,13 @@ export const ApplicationShellAuthenticated = (
                     defaultFlags={props.defaultFeatureFlags}
                   >
                     <>
+                      <div
+                        id={'staff-bar'}
```
tdeekens, line 206:
```suggestion
                        id="above-top-navigation"
```

packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
```diff
@@ -202,6 +202,13 @@ export const ApplicationShellAuthenticated = (
...
+                        css={css`
+                          position: relative;
+                          z-index: 30000;
```
tdeekens, line 209: Why this value?

jaikumar-tj: You are right. Not needed. I added it because dropdown was going behind the top navigation container. I could fix it using ui-kit props.

tdeekens (approved):
Don't forget a changeset.

.changeset/large-bikes-boil.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': minor
+---
+
+Adds a portal above `top-navigation` to inject `staffbar`
```
tdeekens, line 5:
```suggestion
Adds an empty element above the `top-navigation`. 
```

It's not a portal. It's used by one. I'd be more vague also.

tdeekens (approved):

packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
```diff
@@ -202,6 +202,7 @@ export const ApplicationShellAuthenticated = (
                     defaultFlags={props.defaultFeatureFlags}
                   >
                     <>
+                      <div id="above-top-navigation" />
```
emmenko, line 205: Nit: putting it here, will "push the page down" thus creating a scrollbar, which might not be necessary.

I think we can try including it into the global notifications container, as a sibling of the global notifications list.

```diff
+<div id="above-top-navigation" />
<NotificationsList domain={DOMAINS.GLOBAL} />
```

This has the benefit of using the existing grid layout so there should be no potential "issues" with the layout

image

What do you think?

emmenko (commented):
Thanks.

PS: since the staffbar project is only an internal thing, we don't need to share too many info here, as it's a public repo.

packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
jaikumar-tj: This is much better. No unnecessary scrolling. Just a few minor changes required which can be taken care in MC later. Thanks.

tdeekens (approved):

emmenko (approved):
💯

PR #3075, opened 2023-05-09
docs: add some info about the submenu links matching guidelines

emmenko:


website/src/content/api-reference/application-config.mdx
```diff
@@ -574,6 +574,19 @@ For example, given the `entryPointUriPath: avengers`, and the (sub)route `uriPat
 }
 ```
 
+<Info>
+
+About the route matching behavior: links in the Merchant Center main navigation are marked as active based on an **exact matching strategy** with the current page route.
```
mmaltsev-ct, line 579:
```suggestion
Concerning the route matching behavior: links in the Merchant Center main navigation are marked as active based on an **exact matching strategy** with the current page route.
```

website/src/content/api-reference/application-config.mdx
```diff
@@ -574,6 +574,19 @@ For example, given the `entryPointUriPath: avengers`, and the (sub)route `uriPat
...
+
+For example, if a sub-link is configured as `new` and the page route is at `/avengers/new`, the link is marked as active. However, if the page route is at `/avengers` or `/avengers/123` the link does not match and is not marked as active.
```
mmaltsev-ct, line 581:
```suggestion
For example, if a sub-link is configured as `new` and the page route is at `/avengers/new`, the link is marked as active. However, if the page route is at `/avengers` or `/avengers/123`, the link does not match and is not marked as active.
```

website/src/content/api-reference/application-config.mdx
```diff
@@ -574,6 +574,19 @@ For example, given the `entryPointUriPath: avengers`, and the (sub)route `uriPat
...
+
+Be aware of how you configure the sub-links and how they are supposed to match with the application routes:
+
+* Avoid using `/` as the sub-link. While this is technically allowed, you are adding a duplicated menu link as it is the same as the `mainMenuLink`.
+* Avoid links to floating routes. If you have a main page with sub-routes such as `/avengers/config/general` and `/avengers/config/something`, and the routes are configured to redirect from `/avengers/config` to `/avengers/config/general`, defining the sub-link as `/avengers/config` will never match any route because of the redirect.
```
Anshuman71, line 586:
```suggestion
The main navigation links in the Merchant Center are marked as active based on an **exact match strategy** with the current page route.

For example, if a sub-link is set as `new` and the current page route is `/avengers/new`, the link is marked as active. However, if the page route is at `/avengers` or `/avengers/123` instead of `/avengers/new`, the link doesn't match and isn't marked as active.

Be aware of how you configure the sub-links and how they're supposed to match with the application routes:

* Avoid using `/` as the sub-link. While this is technically allowed, you're adding a duplicated menu link because it's the same as the `mainMenuLink`.
* Avoid links to floating routes. If you have a main page with sub-routes such as `/avengers/config/general` and `/avengers/config/something`, and the routes are configured to redirect from `/avengers/config` to `/avengers/config/general`, defining the sub-link as `/avengers/config` will never match any route because of the redirect.
```

Anshuman71 (commented):
@emmenko, this looks good. A few questions:

- What are floating routes?
- Where will the user configure redirects? Are we discussing Vercel or Netlify redirects here or something specific to custom apps?

website/src/content/api-reference/application-config.mdx
mmaltsev-ct, line 586:
```suggestion
* Avoid links to floating routes. Let's imagine you have a page with the route `/avengers/config` configured to redirect to `/avengers/config/general`. In this case, defining the sub-link as `/avengers/config` will never match any route because of the redirect.
```

Anshuman71:
@emmenko,

1. let’s remove the phrase: "Avoid links to floating routes." and start with "If you...."
2. I'm still a bit confused about how I make the `avengers/config` redirect to `/avengers/config/general` in an MC Custom app.

emmenko:
> let’s remove the phrase: "Avoid links to floating routes." and start with "If you...."

The intention was to phrase it as a guideline, so `Avoid this`, `Do that`, `Don't do that`, etc.

> I'm still a bit confused about how I make the `avengers/config` redirect to `/avengers/config/general` in an MC Custom app.

This is client side routing (react router). Nothing special.

Anshuman71:
@emmenko, thanks. Now it’s all good. I've updated my suggestion.

website/src/content/api-reference/application-config.mdx
Anshuman71: We try to keep away from words such as "imagine" or "consider". Instead, we use "For example".

PR #3079, opened 2023-05-10
Accept also React Element as a title type inside tabular detail page

LucasGrahn:
#### Summary

Accept not only `string` as a title for `Tabular Detail Page` but also `ReactElement`

#### Description

I need for a [PR](https://github.com/commercetools/merchant-center-frontend/pull/14587) the possibility to use a ReactElement to show a label component.

### DIFF
--- a/packages/application-components/src/components/detail-pages/tabular-detail-page/tabular-detail-page.tsx
+++ b/packages/application-components/src/components/detail-pages/tabular-detail-page/tabular-detail-page.tsx
@@ -30,7 +30,7 @@ type TTabularDetailPageProps = {
   /**
    * The title of the page.
    */
-  title?: string;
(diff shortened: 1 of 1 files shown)


CarlosCortizasCT:
So currently none of the `app-kit` page components allows for anything but a `string` for the `title` property. I think it is so we are sure the right text element is used (`H1`). If we allow for anything to be rendered over there, we lose control.

If we do, I guess we would need to update the docs to say we expect consumers to use `H1` elements if the don't provide a `string` as the page title. Also, I think we would need to do the same for all the page components to be consistent.

Maybe another option is to have some kind of "slot" as a prop to be rendered right after the title.

@emmenko, @FilPob what's your take on this?

PR #3084, opened 2023-05-16
Improve data fetching code example

CarlosCortizasCT:
#### Summary

Improve data fetching code example

#### Description

The code example we're using in the "[Calling REST APIs inside your GraphQL queries](https://docs.commercetools.com/custom-applications/development/data-fetching#calling-rest-apis-inside-your-graphql-queries)" section of the docs is using a constant from a package which is not imported in the example so we add that missing import in this PR to avoid confusion.

### DIFF
--- a/website/src/content/development/data-fetching.mdx
+++ b/website/src/content/development/data-fetching.mdx
@@ -552,7 +552,7 @@ To use the Apollo REST link with the built-in [GraphQL HTTP client](#graphql-htt
 
 Here is an example of fetching Channels using the `@rest` directive.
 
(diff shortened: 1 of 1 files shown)


website/src/content/development/data-fetching.mdx
```diff
@@ -555,6 +555,7 @@ Here is an example of fetching Channels using the `@rest` directive.
 ```js title="channels.js" highlightLines="18,28"
```
emmenko, line 555: The highlight lines should be adjusted as well

emmenko (approved):

PR #3085, opened 2023-05-17
Fix broken link to Azure account.

stmeissner:
#### Description

Our link checker is complaining about this broken link, so I've updated it with the new link target.


kark (commented):
💯
Would you be so kind to change it also here:
https://github.com/commercetools/merchant-center-application-kit/blob/4afb4f69f1d200cd7598231076ba027e92b2bf78/website/src/content/deployment-examples/azure-static-web-apps.mdx?plain=1#L11
Many thanks 🙏

emmenko (approved):
Thanks!

PR #3086, opened 2023-05-17
test(setupFilesAfterEnv): do not fail on console when not on CI

ahmehri:
#### Summary

Fix the regression introduced in https://github.com/commercetools/merchant-center-application-kit/pull/2820.

#### Description

In https://github.com/commercetools/merchant-center-application-kit/pull/2820, a developer experience regression has been introduced.

Before, debugging tests locally by adding `console.log` won’t fail the test, the output will remain clean and compact and include only the expected log output, in addition to other positive error logs.

But now, the output is cluttered with these false-positive “Expected test not to call console.log()” errors, making the debug experience worse.

The solution is to restore the old state, i.e. to not fail on console when tests are not running on CI.

<details>
<summary>Screenshots</summary>

CleanShot 2023-05-17 at 16 44 55@2x
CleanShot 2023-05-17 at 16 44 03@2x
</details>

#### Future Improvements

Allow this config, `shouldFailOnLog` etc, to be overridden at the consumer level.

### DIFF
--- a/packages/jest-preset-mc-app/setup-test-framework.js
+++ b/packages/jest-preset-mc-app/setup-test-framework.js
@@ -65,10 +65,10 @@ function shouldNotThrowWarnings(...messages) {
 }
 
 failOnConsole({
-  shouldFailOnLog: true,
-  shouldFailOnInfo: true,
-  shouldFailOnWarn: true,
-  shouldFailOnError: true,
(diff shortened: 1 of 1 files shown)


emmenko (approved):
Thanks for contributing!

I think this is a fair approach, thanks for proposing it. Let's give it a try and see if it improves the experience 🤗

PR #2976, opened 2023-02-28
WIP migrate to `pnpm`

emmenko:
- [ ] Installation works
- [ ] Linters works
- [ ] Tests works
- [ ] Types works
- [ ] Build works
- [ ] Local dev works
- [ ] CI works

PR #3087, opened 2023-05-17
feat(create-mc-app): support new cli options to specify preferred package manager

emmenko:
Follow up of #2976

### DIFF
--- a/packages/create-mc-app/src/validations.ts
+++ b/packages/create-mc-app/src/validations.ts
@@ -1,7 +1,7 @@
 import fs from 'fs';
 import path from 'path';
 import semver from 'semver';
-import type { TCliCommandOptions } from './types';
+import type { TPackageManager, TTemplate } from './types';
 import { isSemVer } from './utils';
 
 const availableTemplates = {
@@ -9,9 +9,7 @@ const availableTemplates = {
   'starter-typescript': 'starter-typescript',
 } as const;
 
-const throwIfTemplateIsNotSupported = (
-  templateName: TCliCommandOptions['template']
-) => {
+const throwIfTemplateIsNotSupported = (templateName: TTemplate) => {
   switch (templateName) {
     case availableTemplates.starter:
     case availableTemplates['starter-typescript']:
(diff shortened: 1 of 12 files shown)


packages/create-mc-app/src/validations.ts
```diff
@@ -81,10 +79,22 @@ const throwIfNodeVersionIsNotSupported = (
   }
 };
 
+const throwIfPackageManagerVersionIsMissing = (
+  packageManager?: TPackageManager,
+  packageManagerVersion?: string
+) => {
+  if (packageManager && !packageManagerVersion) {
```
CarlosCortizasCT, line 86: **nit**: I was wondering why users need to specify the package manager version.
As a developer using the CLI, I can see it asking me about the package manager I want to use but not sure why it would ask me about the version. Most of the time, I wouldn't even know and need to search for it.

CarlosCortizasCT (approved):
Thanks! 💯

packages/create-mc-app/src/validations.ts
emmenko: The version is needed for the `packageManager` field in `package.json`.
Unfortunately it doesn't work without it.

One option I was thinking about was to automatically fetch the latest version.

I was also thinking of automatically detect the package manager but that's tricky because yarn or pnpm don't need to be installed globally.

As a follow up I also wanted to look into enabling yarn or pnpm (if selected) via Corepack. That requires though to have certain nodejs versions installed.

So there are some tricky scenarios to consider and for the sake of simplicity I opted to require the package manager version to be explicitly requested.

Let me take another look at this though, because for pnpm (due to how it should be installed via Corepack) I opted to skip the installation step. So the version is not strictly needed anymore even though the user will need to specify it anyway in the `packageManager` field.

emmenko: Decided to remove it for now: https://github.com/commercetools/merchant-center-application-kit/pull/3087/commits/e2dc2b24c5f314ed79767166cfeac71f0c95c004

PR #3088, opened 2023-05-18
Old look and field removal

CarlosCortizasCT:
#### Summary

Old look and field removal

#### Description

During the last months we've been working on implementing a new look and feel for `app-kit` components. The process involved including some code to support both the old and the new styles during the transition.

Now that the new look and feel has finally been released, we are removing all that support code and keep just the new styles.


packages/application-components/src/components/page-content-containers/page-content-wide/page-content-wide.tsx
```diff
@@ -1,13 +1,13 @@
...
-import { designTokens, useTheme } from '@commercetools-uikit/design-system';
+import { designTokens } from '@commercetools-uikit/design-system';
 import { useWarning } from '@commercetools-uikit/utils';
 
 export type TPageContentWide = {
   columns: '1' | '1/1' | '2/1';
   gapSize: '10' | '20';
   children: ReactNode;
+  // @deprecated
   themeParentSelector?: () => HTMLElement | null;
```
CarlosCortizasCT, line 11: This is mostly an internal prop we used in the MC frontend so I think it would be safe to remove it but I thought it would help to just deprecate it first.

packages/application-components/src/theming.ts
```diff
@@ -12,54 +12,6 @@ const newShadow1 = '0px 2px 5px 0px rgba(0, 0, 0, 0.15)';
 
 export const themesOverrides = {
   default: {
-    colorForPageHeaderBottomBorder: uiKitDesignTokens.colorNeutral,
```
CarlosCortizasCT, line 15: `test` theme has been removed and all `default` theme values has been overwritten with the ones we had in the `test` one.

packages/application-shell/src/components/theme-switcher/theme-switcher.tsx
```diff
@@ -1,55 +1,23 @@
...
-    hasUserSeenNewDesignReleaseNotificationDialog,
-    setHasUserSeenNewDesignReleaseNotificationDialog,
-  ] = useState(
-    window.localStorage.getItem(
-      STORAGE_KEYS.IS_NEW_DESIGN_RELEASE_NOTIFICATION_CLOSED
-    ) === 'true'
-  );
+  // This helps decoupling app-kit update from ui-kit update
+  // TODO: Remove after ui-kit redesign cleanup has been done
+  const theme = 'test';
```
CarlosCortizasCT, line 9: If someone updates `app-kit` before updating `ui-kit`, we will be using the new theme styles by default for `app-kit` but we need to make sure we still tell `ui-kit` we want the new theme.
Once they update `ui-kit`, passing it `test` as the theme will just show a warning in the console but it will be ignored and the `default` one will be used.

We can also wait for the new version of `ui-kit` to be released and update `app-kit` to use it as the minimum dependant versions.

jaikamat:
@CarlosCortizasCT Thanks for bringing this up, the `welcomeMessage` prop is [indeed unused](https://github.com/commercetools/merchant-center-frontend/blob/071b5a193f0776bea7491d2dffa714169a04ab32/packages-application/application-authentication/src/components/login-layout-card/login-layout-card.jsx#L18) as you cited in the current login page post-redesign. I believe it's safe to deprecate 😄

packages/application-shell/src/components/theme-switcher/theme-switcher.tsx
```diff
@@ -1,55 +1,23 @@
...
+  }, []);
 
   return (
     <>
-      <ThemeProvider theme={theme} themeOverrides={themesOverrides[theme]} />
-      {!isRedesignAnnouncementDisabled &&
-      !hasUserSeenNewDesignReleaseNotificationDialog &&
-      isNewThemeEnabled ? (
-        <NewDesignReleaseInfoDialog onClose={handleCloseDialog} />
-      ) : null}
```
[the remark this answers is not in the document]
emmenko: Yes, this needs to stay for a little longer.

visual-testing-app/src/index.tsx
```diff
@@ -1,31 +1,15 @@
...
-        theme={selectedTheme}
-        themeOverrides={themesOverrides[selectedTheme]}
-      />
-      {isLocalEnvironment ? (
-        <ThemeSwitcher
-          selectedTheme={selectedTheme}
-          onThemeChange={(newTheme) => updateTheme(newTheme)}
-        />
-      ) : null}
+      <ThemeProvider theme="test" themeOverrides={themesOverrides.default} />
```
emmenko, line 12: Shouldn't we remove the `theme=test` (assuming that we use the latest UI Kit version with the old theme being removed)?

emmenko (commented):
🥳

visual-testing-app/src/index.tsx
CarlosCortizasCT: I decided to keep it like this so we don't have to depend on the new version of `ui-kit`.

packages/application-shell/src/components/theme-switcher/theme-switcher.tsx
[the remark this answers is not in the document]
CarlosCortizasCT: Restored in this commit: ccb8a9144079058df6dd0365d6802a5d5fc172de

packages/application-components/src/components/public-page-layout/public-page-layout.tsx
```diff
@@ -65,55 +62,41 @@ const PublicPageLayoutContent: FC<TProps> = (props) => {
 };
 
 const PublicPageLayout: FC<TProps> = (props) => {
-  const { themedValue } = useTheme();
-
   return (
-    <Container showBackground={themedValue(true, false)}>
+    <Container showBackground={false}>
```
emmenko, line 66: Nit: let's also remove the background completely, which includes the png image.

emmenko (approved):
Awesome job, thanks! 🥳

The background image clean up can also be a follow up. Up to you.

emmenko (approved):
Awesome job, thanks! 🥳

The background image clean up can also be a follow up. Up to you.

kark (approved):
Thank you 🙌

PR #3089, opened 2023-05-18
fix: wrong ApiErrorMessage for DuplicateField GraphQL error

ragafus:


packages/react-notifications/src/components/notification-kinds/api-error-message/api-error-message.spec.tsx
```diff
@@ -96,10 +96,12 @@ describe('render', () => {
   });
   it('should show message for DuplicateField', () => {
```
emmenko, line 97: What about testing for both cases?

emmenko (commented):
Thanks for contributing. Overall looks good, just have one question about testing.

packages/react-notifications/src/components/notification-kinds/api-error-message/api-error-message.spec.tsx
ragafus: > What about testing for both cases?

I've added cases for REST and GraphQL errors 👍

CarlosCortizasCT (approved):
Thanks! 🙇

PR #3094, opened 2023-05-22
fix: wrong ApiErrorMessage for InvalidField GraphQL error

ragafus:
#### Summary
This PR adds support for the GraphQL variant of the `InvalidField` error.

#### Description
The current message is not properly formatted. It doesn't replace the `field` placeholder : `The value entered is not valid for the field "{field}".`


emmenko (approved):
Thanks!

CarlosCortizasCT (approved):
Many thanks 🙇

PR #3114, opened 2023-06-20
fix(website): resolving transitive dependencies in dev mode

kark:
#### Summary

Starting `custom-applications` docs in dev mode results in problems with resolving some transitive dependencies after migration to `pnpm`:
```bash
Module not found: Error: Can't resolve 'query-string' in
‘…/merchant-center-application-kit/website/.cache'
Module not found: Error: Can't resolve 'lodash/merge' in
‘…/merchant-center-application-kit/website/.cache'
``` 

From what I found out [the accepted solution](https://github.com/pnpm/pnpm/issues/991#issuecomment-1120101531) is using a special gatsby plugin for that. It's just a proposal, perhaps you have a better idea how to solve that.

Side note:
Adding as a dev dependency as advised here:
https://github.com/Js-Brecht/gatsby-plugin-pnpm/#how-to-install

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -2951,6 +2951,9 @@ importers:
       '@svgr/cli':
         specifier: 6.5.1
         version: 6.5.1
+      gatsby-plugin-pnpm:
+        specifier: 1.2.10
+        version: 1.2.10(gatsby@5.9.1)
       linkinator:
         specifier: 4.1.2
         version: 4.1.2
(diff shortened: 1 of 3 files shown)


emmenko:
Ah interesting, thanks for looking into it!

So with this plugin everything works fine?

kark:
> Ah interesting, thanks for looking into it!
> 
> So with this plugin everything works fine?

Yes, it seems to work fine 🙂

emmenko (approved):
💯

CarlosCortizasCT (approved):
🙇

tdeekens (approved):

Issue #3126, opened 2023-07-03
<PageContentWide> silently ignores 3rd child component and doesn't throw any error


jaikumar-tj:
**Is your feature request related to a problem? Please describe.**

The `PageContentWide` doesn't throw any error or warning when more than 2 childrens are passed as props. It only renders first 2 child since the column prop determines how many columns to render. It goes unnoticed and developers don't realise that 3rd component (eg: info dialog or warn on leave or save bar) is not rendered on the page after a while.

**Describe the solution you'd like**

It would be helpful if the components throw an error so developers could avoid the mistake of passing 3rd component and wrap everything under fragment.

**Describe alternatives you've considered**

Atleast a console warning message would also be helpful to shorten the debug time. But the above solution stops the mistake from happening at very early stage.

CarlosCortizasCT:
Hi @jaikumar-tj,

Thanks for opening this feature request 👍 

We will look into it and try to improve the DX around that component.

Regarding the use cases you mention, I would say those extra components (info dialog, save bar,...) should not be wrapped in a `PageContentWide` though. This component is about laying out main content in a page and those examples are more about floating or fixed elements which might not be good candidates for this component.

Anyway, we'll try to improve this component and let you know.

PR #3127, opened 2023-07-04
Added warnings to the `PageContentWide` component

chloe0592:
The `PageContentWide` doesn't throw a warning when more children are passed as props. 
For example if 3 are passed, it only renders first 2 children and it goes unnoticed that 3rd component is not rendered and why is that the case.

In this PR we cover 3 possible scenarios when the warning is shown:
1. There's 1 column and more than 1 child.
2. There's 1/1 or 2/1 columns and only 1 child.
3. There's 1/1 or 2/1 columns and more than 2 children.

This PR closes - https://github.com/commercetools/merchant-center-application-kit/issues/3126.

### DIFF
--- a/.changeset/great-eggs-try.md
+++ b/.changeset/great-eggs-try.md
@@ -1,0 +1,6 @@
+---
+'@commercetools-frontend/application-components': patch
+'@commercetools-local/visual-testing-app': patch
+---
+
+Added warnings to the `page-content-wide' and tests update
--- a/packages/application-components/src/components/page-content-containers/page-content-wide/page-content-wide.tsx
+++ b/packages/application-components/src/components/page-content-containers/page-content-wide/page-content-wide.tsx
@@ -55,10 +55,33 @@ const Container = styled.div`
 
 function PageContentWide(props: TPageContentWide) {
   const [leftChild, rightChild] = Children.toArray(props.children);
+  const countChildren = Children.count(props.children);
+
+  // if there's 1 column and more than 1 child
+  const isColumnsOneAndMoreThanOneChild =
+    props.columns === '1' && countChildren > 1;
 
   useWarning(
-    props.columns !== '1' || !Boolean(rightChild),
-    'PageContentWide: This component only renders its first children when using a single column but you provided more that one.'
+    !isColumnsOneAndMoreThanOneChild,
+    'PageContentWide: This component has more than 1 child. Only the first child will be rendered.'
+  );
+
+  // if there's 1/1 or 2/1 columns and only 1 child
+  const isColumnsTwoAndOneChild =
+    (props.columns === '1/1' || props.columns === '2/1') && countChildren === 1;
+
+  useWarning(
+    !isColumnsTwoAndOneChild,
+    'PageContentWide: This component has only 1 child. The second child will be ignored.'
+  );
+
+  // if there's 1/1 or 2/1 columns and more than 2 children
+  const isColumnsTwoAndMoreThanTwoChildren =
+    (props.columns === '1/1' || props.columns === '2/1') && countChildren > 2;
+
+  useWarning(
+    !isColumnsTwoAndMoreThanTwoChildren,
+    'PageContentWide: This component has more than 2 children. Only the first 2 children will be rendered.'
   );
 
   return (
--- a/visual-testing-app/src/components/page-content-container-wide/page-content-container-wide.visualroute.tsx
+++ b/visual-testing-app/src/components/page-content-container-wide/page-content-container-wide.visualroute.tsx
@@ -38,18 +38,6 @@ export const Component = () => (
           ),
         },
         {
-          name: 'single column with several children',
-          path: 'single-column-with-several-children',
-          spec: (
-            <PageContentWide>
-              <Box />
-              <Box />
-              <Box />
-            </PageContentWide>
-          ),
-        },
-
-        {
           name: 'two columns 1/1',
           path: 'two-columns-half',
           spec: (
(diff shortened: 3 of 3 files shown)


packages/application-components/src/components/page-content-containers/page-content-wide/page-content-wide.tsx
```diff
@@ -55,10 +55,33 @@ const Container = styled.div`
...
+    props.columns === '1' && countChildren > 1;
+
+  useWarning(
+    !isColumnsOneAndMoreThanOneChild,
+    'PageContentWide: This component has more than 1 child. Only the first child will be rendered.'
+  );
+
+  // if there's 1/1 or 2/1 columns and only 1 child
+  const isColumnsTwoAndOneChild =
+    (props.columns === '1/1' || props.columns === '2/1') && countChildren === 1;
```
chloe0592, line 71: Wasn't sure if we want to show a warning in this case, but it made sense to me that if only one child is passed it would be usually by a mistake. But maybe I'm wrong?

jaikumar-tj (approved):
Thanks for the quick fixup. Just curious, does `useWarning` simply show a console warning or does it throw an error and prevent the developer from making mistake?

tdeekens (approved):

CarlosCortizasCT:
> > Thanks for the quick fixup. Just curious, does `useWarning` simply show a console warning or does it throw an error and prevent the developer from making mistake?
> 
> @jaikumar-tj yes, exactly it'll show a warning in the console, it doesn’t throw an error. It will be looking like this. Screenshot 2023-07-04 at 14 00 50

@jaikumar-tj we decided to follow @tdeekens advice of using a warning.

Bear in mind this will use a `console.warn` which should make the MC frontend consuming component tests to fail.

packages/application-components/src/components/page-content-containers/page-content-wide/page-content-wide.tsx
CarlosCortizasCT: Technically, the second column is not required when using two columns.
I've found some use cases in MC frontend repository where we're providing just one column in a two columns layout.
I'm not very sure if that was intentional, but I think it's safer for us to now apply that restriction.

I believe it would be better to keep the previous one and just add another for the use case of receiving more that two children on a two column layout:
```js
  const [leftChild, rightChild] = Children.toArray(props.children);
  const childrenCount = Children.count(props.children);

  useWarning(
    props.columns === '1' && childrenCount > 1,
    'PageContentWide: This component only renders its first children when using a single column but you provided more that one.'
  );

  useWarning(
    props.columns !== '1' && childrenCount > 2,
    'PageContentWide: This component only renders its first two children when using a two columns layout but you provided more that two.'
  );
```

chloe0592: Thanks for the feedback! 💡  Here's the change - [aaa1609](https://github.com/commercetools/merchant-center-application-kit/pull/3127/commits/aaa16098ceb3af35f12d773f93f2a4e22910a6ce).

packages/application-components/src/components/page-content-containers/page-content-wide/page-content-wide.tsx
```diff
@@ -55,12 +55,18 @@ const Container = styled.div`
...
   const [leftChild, rightChild] = Children.toArray(props.children);
+  const childrenCount = Children.count(props.children);
 
   useWarning(
-    props.columns !== '1' || !Boolean(rightChild),
+    props.columns === '1' && childrenCount > 1,
     'PageContentWide: This component only renders its first children when using a single column but you provided more that one.'
   );
 
+  useWarning(
```
CarlosCortizasCT, line 65: Maybe it would be a good idea to add a test to check this warning the same way we do for the other one.

.changeset/great-eggs-try.md
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools-frontend/application-components': patch
+'@commercetools-local/visual-testing-app': patch
+---
+
+Added warnings to the `page-content-wide' and tests update
```
CarlosCortizasCT, line 6: Since the changeset is targeted for `app-kit` users, perhaps we can explain a little further what we're changing here.

Here's an idea:
```
We included a new warning in the `PageContentWide` component to let consumers know the component will only render its first two children when configured to use two columns but consumers provided more than two children.
```

What do you think? 🤔

CarlosCortizasCT (approved):

packages/application-components/src/components/page-content-containers/page-content-wide/page-content-wide.tsx
```diff
@@ -55,12 +55,18 @@ const Container = styled.div`
 
 function PageContentWide(props: TPageContentWide) {
   const [leftChild, rightChild] = Children.toArray(props.children);
+  const childrenCount = Children.count(props.children);
 
   useWarning(
-    props.columns !== '1' || !Boolean(rightChild),
+    props.columns === '1' && childrenCount > 1,
     'PageContentWide: This component only renders its first children when using a single column but you provided more that one.'
```
ddouglasz, line 62:
```suggestion
    'PageContentWide: This component only renders its first children when using a single column but you provided more than one.'
```

packages/application-components/src/components/page-content-containers/page-content-wide/page-content-wide.tsx
```diff
@@ -55,12 +55,18 @@ const Container = styled.div`
...
   );
 
+  useWarning(
+    props.columns !== '1' && childrenCount > 2,
+    'PageContentWide: This component only renders its first two children when using a two columns layout but you provided more that two.'
```
ddouglasz, line 67:
```suggestion
    'PageContentWide: This component only renders its first two children when using a two columns layout but you provided more than two.'
```

ddouglasz (approved):
Nice! 🫶🏻
Just tiny observations

packages/application-components/src/components/page-content-containers/page-content-wide/page-content-wide.tsx
chloe0592: Great catch! 🦅 Thanks - [c7557ec](https://github.com/commercetools/merchant-center-application-kit/pull/3127/commits/c7557ecb5477db0f99acf7e076887da7249ff3e8).

.changeset/great-eggs-try.md
chloe0592: Sure, make sense. :) Updated here - [c7557ec](https://github.com/commercetools/merchant-center-application-kit/pull/3127/commits/c7557ecb5477db0f99acf7e076887da7249ff3e8)

packages/application-components/src/components/page-content-containers/page-content-wide/page-content-wide.tsx
chloe0592: Added here - [c7557ec](https://github.com/commercetools/merchant-center-application-kit/pull/3127/commits/c7557ecb5477db0f99acf7e076887da7249ff3e8)

CarlosCortizasCT (approved):
💯

PR #3128, opened 2023-07-05
`PageContentWide` warning fix

chloe0592:
FIX: The warning will be shown if the first param is `false` so we needed to adjust the condition.

### DIFF
--- a/.changeset/old-sheep-thank.md
+++ b/.changeset/old-sheep-thank.md
@@ -2,4 +2,6 @@
 '@commercetools-frontend/application-components': patch
 ---
 
-Fix for the `PageContentWide` warning
+In version `22.3.3` we introduced a bug in the `PageContentWide` component leading to it sending warning messages to browser console log when correctly using it.
+
+We're now fixing that incorrect behaviour.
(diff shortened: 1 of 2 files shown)


.changeset/old-sheep-thank.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-frontend/application-components': patch
+---
+
+Fix for the `PageContentWide` warning
```
CarlosCortizasCT, line 5: Maybe we can provide some more details here:
```suggestion
In version `22.3.3` we introduced a bug in the `PageContentWide` component leading to it sending warning messages to browser console log when correctly using it.

We're now fixing that incorrect behaviour.
```
```

chloe0592: Thanks! 🙏 Updated - [c4ee7f2](https://github.com/commercetools/merchant-center-application-kit/pull/3128/commits/c4ee7f252c01086eced77694f505aeba709d6ca2)

CarlosCortizasCT (approved):
Thanks 💯

Issue #3110, opened 2023-06-09
create-mc-app fails during npm install


quaker-schneider-commercetools:
**Describe the issue**

Installing a merchant center application template via `npx @commercetools-frontend/create-mc-app@latest my-new-custom-application-project --template starter` fails with the following error message during the `npm install` step:
```
...
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! While resolving: my-new-custom-application-project@1.0.0
npm ERR! Found: graphql@16.6.0
npm ERR! node_modules/graphql
npm ERR!   graphql@"^16.6.0" from the root project
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer graphql@"^0.12.0 || ^0.13.0 || ^14.0.0 || ^15.0.0" from eslint-plugin-graphql@4.0.0
npm ERR! node_modules/eslint-plugin-graphql
npm ERR!   eslint-plugin-graphql@"^4.0.0" from the root project
...
```
After running `create-mc-app`, the template project is created in the active directory. 
However, Running `npm start` the template directory results in an error similar to the one above.

**To Reproduce**

1. Open terminal in empty directory
2. Enter the following: 
```
npx @commercetools-frontend/create-mc-app@latest \
  my-new-custom-application-project \
  --template starter
```
3. Observe error in console output

**Expected behavior**

`create-mc-app` downloads and builds the template application with no error. 

**Screenshots**

<img width="567" alt="Screen Shot 2023-06-09 at 10 48 30 AM" src="https://github.com/commercetools/merchant-center-application-kit/assets/135636519/5a62e78c-aab4-4033-8b48-7188de763fe4">

**Additional context**
Logs from a failed run of `create-mc-app`:
[2023-06-09T14_46_52_505Z-debug-0.log](https://github.com/commercetools/merchant-center-application-kit/files/11707423/2023-06-09T14_46_52_505Z-debug-0.log)
[2023-06-09T14_46_52_505Z-eresolve-report.txt](https://github.com/commercetools/merchant-center-application-kit/files/11707424/2023-06-09T14_46_52_505Z-eresolve-report.txt)

Fails for both `--template=starter` and `--template=starter-typescript`
System info:
- macOS Monterey `12.6.6`
- Node `20.2.0` and `18.13.0` (tried on both)
- npm `9.6.6` and `8.19.3` (tried on both)

[...]

Issue #1978, opened 2021-01-18
Investigate using @graphql-eslint/eslint-plugin instead of our current plugin


emmenko:
Investigate using https://github.com/dotansimha/graphql-eslint/ instead of our current plugin.

mmaltsev-ct:
Background: GraphQL Schema introspection is used to fetch GraphQL Schema (and for instance, populate Docs in GraphiQL UI). The ESLint plugin uses this Schema to validate GraphQL Documents. The plugin we currently use seems to be less maintained and less active than the alternative @graphql-eslint/eslint-plugin.

PR #3134, opened 2023-07-14
Downgrade `graphql` dependency version

CarlosCortizasCT:
#### Summary

Downgrade `graphql` dependency version

#### Description

Some users using our `@commercetools-frontend/create-mc-app` to create a new Custom Application are experiencing problems when running the command because of a peer dependency miss match with the `graphql` package.

More info here: #3110 

Our starter templates rely on the [eslint-plugin-graphql](https://github.com/apollographql/eslint-plugin-graphql/tree/master) package to add static validation to graphql mutations and queries.
This package uses the `graphql` one as a peer dependency and the one we are currently using is `16.x` which [is not supported](https://github.com/apollographql/eslint-plugin-graphql/blob/master/package.json#L50) by the former.

We're downgrading our `graphql` version to the latest supported by the plugin (`v15`) as we don't want to loose that static check.

We also created [an issue](https://github.com/apollographql/eslint-plugin-graphql/issues/333) in the `eslint-plugin-graphql` repository asking for `graphql` `v16.x` version support.

### DIFF
--- a/.changeset/fair-peas-pull.md
+++ b/.changeset/fair-peas-pull.md
@@ -1,0 +1,10 @@
+---
+'@commercetools-applications/merchant-center-template-starter-typescript': minor
+'@commercetools-frontend/application-shell-connectors': minor
+'@commercetools-applications/merchant-center-template-starter': minor
+'@commercetools-frontend/jest-preset-mc-app': minor
+'@commercetools-frontend/application-shell': minor
+'@commercetools-frontend/mc-scripts': minor
+---
+
+Downgrade `graphql` package to version `15.x`.
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -53,6 +53,10 @@
     {
       "matchPackageNames": ["@types/react"],
       "allowedVersions": "<17.0.57"
+    },
+    {
+      "matchPackageNames": ["graphql"],
+      "allowedVersions": "<16"
     }
   ],
   "lockFileMaintenance": {
--- a/application-templates/starter-typescript/package.json
+++ b/application-templates/starter-typescript/package.json
@@ -75,7 +75,7 @@
     "eslint-formatter-pretty": "4.1.0",
     "eslint-plugin-graphql": "^4.0.0",
     "formik": "2.2.9",
-    "graphql": "^16.6.0",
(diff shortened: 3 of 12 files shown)


tdeekens (approved):

ddouglasz (approved):

kark (approved):

emmenko:
Wouldn't this cause issues in our internal apps that use v16? 🤔

CarlosCortizasCT:
> Wouldn't this cause issues in our internal apps that use v16? 🤔

Yes, that's a good point.
I've just checked and they fail indeed.

I wanted to downgrade the version in all packages for consistency, but maybe it would be better to just do it in the templates.

What do you think?

emmenko:
It seems to me that this is mostly an issue when using `npm` as it's strict by default regarding peer dependencies.

In the issue in mentioned that by using `npm install --legacy-peer-deps` it does work.

I would keep using graphql `v16`. For the issue with `npm` we could update the `create-mc-app` CLI to include the `--legacy-peer-deps` option when using `npm` as the package manager
https://github.com/commercetools/merchant-center-application-kit/blob/0aa9c188807988380c050a7382abf9eeff9b0d6c/packages/create-mc-app/src/tasks/install-dependencies.ts#L14

As for the `eslint-plugin-graphql`, it seems it's not maintained much. I already opened an issue some time ago to investigate using another library #1978.

What do you think?

CarlosCortizasCT:
> It seems to me that this is mostly an issue when using `npm` as it's strict by default regarding peer dependencies.
> 
> In the issue it mentioned that by using `npm install --legacy-peer-deps` it does work.
> 
> I would keep using graphql `v16`. For the issue with `npm` we could update the `create-mc-app` CLI to include the `--legacy-peer-deps` option when using `npm` as the package manager
> 
> https://github.com/commercetools/merchant-center-application-kit/blob/0aa9c188807988380c050a7382abf9eeff9b0d6c/packages/create-mc-app/src/tasks/install-dependencies.ts#L14
> 
> As for the `eslint-plugin-graphql`, it seems it's not maintained much. I already opened an issue some time ago to investigate using another library #1978.
> 
> What do you think?

Ok.
Makes sense to me.

PR #3138, opened 2023-07-17
fix(create-mc-app): run `npm install` cmd with `--legacy-peer-deps` option

emmenko:
Fixes #3110 

Refers to comment https://github.com/commercetools/merchant-center-application-kit/pull/3134#issuecomment-1637648922

PR #3139, opened 2023-07-17
Revert "Downgrade `graphql` dependency version"

emmenko:
Reverts commercetools/merchant-center-application-kit#3134

Proposed alternative solution to avoid downgrading graphql: https://github.com/commercetools/merchant-center-application-kit/pull/3138


CarlosCortizasCT (approved):
Thanks.

PR #3141, opened 2023-07-18
feat(packages-backend/loggers): support to not unset from field

tdeekens:
#### Summary

In 90% of the cases you intend to unset the from field when rewriting log field. In the other 10% however you may want to keep the from field.

This use case particularly holds when you want to gracefully migrate fields logged from an old to a new field without breaking dashboards in logging tools.

The process there would be to:

1. Rewrite a field to a new field
2. Keep the old field (feature here)
3. Keep logs ingesting for a while
4. Update all dashboards and saved queries in the logging tool
5. Unset the old from field

Internally this use case is needed when migrating to standardized logging field names

### DIFF
--- a/.changeset/mighty-donkeys-sip.md
+++ b/.changeset/mighty-donkeys-sip.md
@@ -2,15 +2,15 @@
 "@commercetools-backend/loggers": minor
 ---
 
-Support to not unset the original field when rewriting logs.
+Add support to optionally preserve the original field when rewriting log fields with `rewriteFieldsFormatter`.
 
-When rewriting a field you can use the `unsetFromField` option and set it to `false`. For example:
+When rewriting a field you can use the `preserveFromField` option and set it to `true`. For example:
 
 ```js
 {
   from: 'meta.req.statusCode',
   to: 'meta.req.status_code',
-  unsetFromField: false,
+  preserveFromField: true,
 }
 ```
 
--- a/.changeset/mighty-donkeys-sip.md
+++ b/.changeset/mighty-donkeys-sip.md
@@ -1,0 +1,17 @@
+---
+"@commercetools-backend/loggers": minor
+---
+
+Support to not unset the original field when rewriting logs.
+
+When rewriting a field you can use the `unsetFromField` option and set it to `false`. For example:
+
+```js
+{
+  from: 'meta.req.statusCode',
+  to: 'meta.req.status_code',
+  unsetFromField: false,
+}
+```
+
+Would add a field `meta.req.status_code` while keeping the original `meta.req.statusCode`. This can be helpful when gracefully migrating fields.
--- a/packages-backend/loggers/src/formatters/rewrite-fields.ts
+++ b/packages-backend/loggers/src/formatters/rewrite-fields.ts
@@ -9,16 +9,31 @@ type TRewriteField = {
   from: string;
   to: string;
   replaceValue?: (value: unknown) => unknown;
+  unsetFromField?: boolean;
 };
 type TRewriteFieldsFormatterOption = {
   fields: TRewriteField[];
 };
 
+const defaultRewriteField: TRewriteField = {
+  unsetFromField: true,
+};
 function rewriteField(info: TransformableInfo, field: TRewriteField) {
-  const val = getIn(info, field.from);
-  if (val) {
-    unsetIn(info, field.from);
-    setIn(info, field.to, field.replaceValue ? field.replaceValue(val) : val);
+  const mergedRewriteField = { ...defaultRewriteField, ...field };
+  const fromFieldValue = getIn(info, mergedRewriteField.from);
+
+  if (fromFieldValue) {
+    if (mergedRewriteField.unsetFromField) {
+      unsetIn(info, mergedRewriteField.from);
+    }
+
+    setIn(
+      info,
+      mergedRewriteField.to,
+      mergedRewriteField.replaceValue
+        ? mergedRewriteField.replaceValue(fromFieldValue)
+        : fromFieldValue
+    );
   }
 }
 
--- a/packages-backend/loggers/src/formatters/rewrite-fields.ts
+++ b/packages-backend/loggers/src/formatters/rewrite-fields.ts
@@ -15,7 +15,7 @@ type TRewriteFieldsFormatterOption = {
   fields: TRewriteField[];
 };
 
-const defaultRewriteField: TRewriteField = {
+const defaultRewriteField: Partial<TRewriteField> = {
   unsetFromField: true,
 };
 function rewriteField(info: TransformableInfo, field: TRewriteField) {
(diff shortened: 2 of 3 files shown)


packages-backend/loggers/src/formatters/rewrite-fields.ts
```diff
@@ -9,16 +9,31 @@ type TRewriteField = {
   from: string;
   to: string;
   replaceValue?: (value: unknown) => unknown;
+  unsetFromField?: boolean;
```
tdeekens, line 12: Argued with myself for a name. This is the best I could come up with.

packages-backend/loggers/src/formatters/rewrite-fields.ts
```diff
@@ -9,16 +9,31 @@ type TRewriteField = {
...
 
+const defaultRewriteField: TRewriteField = {
+  unsetFromField: true,
+};
 function rewriteField(info: TransformableInfo, field: TRewriteField) {
-  const val = getIn(info, field.from);
-  if (val) {
-    unsetIn(info, field.from);
-    setIn(info, field.to, field.replaceValue ? field.replaceValue(val) : val);
+  const mergedRewriteField = { ...defaultRewriteField, ...field };
```
tdeekens, line 22: This is needed as as `field: TRewriteField = defaultRewriteField` would not work as defaulted parameters are not merged when they are objects.

packages-backend/loggers/src/formatters/rewrite-fields.ts
```diff
@@ -9,16 +9,31 @@ type TRewriteField = {
...
+  const fromFieldValue = getIn(info, mergedRewriteField.from);
```
tdeekens, line 23: Just renamed.

packages-backend/loggers/src/formatters/rewrite-fields.ts
```diff
@@ -9,16 +9,31 @@ type TRewriteField = {
...
+
+  if (fromFieldValue) {
+    if (mergedRewriteField.unsetFromField) {
+      unsetIn(info, mergedRewriteField.from);
+    }
```
tdeekens, line 28: The condition.

packages-backend/loggers/src/loggers.spec.ts
```diff
@@ -2,47 +2,73 @@ import createApplicationLogger from './create-application-logger';
...
-      formatters: [
-        rewriteFieldsFormatter({
-          fields: [
-            {
-              from: 'meta.req.headers',
-              to: 'meta.req.headersJsonString',
-              replaceValue: (value) => {
-                return JSON.stringify(value);
+  describe('when replacing field', () => {
+    it('should log and replace fields', () => {
```
tdeekens, line 6: This is the old case just nested in.

packages-backend/loggers/src/loggers.spec.ts
```diff
@@ -2,47 +2,73 @@ import createApplicationLogger from './create-application-logger';
...
-      ],
+            ],
+          }),
+        ],
+      });
+      logger.info('Test log', {
+        meta: { req: { headers: { Accept: 'application/json' } } },
+      });
+      // Note: that both only headersJsonString is present
+      expect(console._stdout.write.mock.lastCall).toMatchInlineSnapshot(`
```
tdeekens, line 29: I suggest to move to inline snapshots. While the string matchers are not hardened to change they're also troublesome to debug. As a tradeoff a added a comment.

packages-backend/loggers/src/loggers.spec.ts
```diff
@@ -2,47 +2,73 @@ import createApplicationLogger from './create-application-logger';
...
+          "{"level":"info","message":"Test log","meta":{"req":{"headersJsonString":"{\\"Accept\\":\\"application/json\\"}"}}}
+        ",
+        ]
+      `);
     });
-    logger.info('Test log', {
-      meta: { req: { headers: { Accept: 'application/json' } } },
+  });
+  describe('when not replacing field', () => {
+    it('should log and add but not replace fields', () => {
```
tdeekens, line 38: This is the new functionality.

packages-backend/loggers/src/loggers.spec.ts
```diff
@@ -85,18 +111,12 @@ describe('application logger', () => {
...
-            headers: {
-              authorization: '[REDACTED]',
-            },
-            headersJsonString: JSON.stringify({ authorization: '[REDACTED]' }),
-          },
-        })}`
-      )
-    );
+    // Note that authorization is REDACTED in both headers and headersJsonString
+    expect(console._stdout.write.mock.lastCall).toMatchInlineSnapshot(`
```
tdeekens, line 115: Just migrated the expectation too after making sure the old one worked with my changes made.

packages-backend/loggers/src/formatters/rewrite-fields.ts
emmenko: Would it be better to have it `false` by default and enable the functionality with `true`? That's what we usually do with component props, right?

```suggestion
  keepOriginalField?: boolean;
```

packages-backend/loggers/src/formatters/rewrite-fields.ts
emmenko: Do we need to merge here? Since it's only one option maybe we can simply inline the condition `field.keepOriginalField ?? false`?

emmenko (commented):
👌 Just a few suggestions, otherwise looks good!

packages-backend/loggers/src/formatters/rewrite-fields.ts
tdeekens: Works for me too. Was an alternative I pondered with. I liked the term unset as it follows the implementation nicely. 

keep* or preserve* are the inverse options. I can change it.

packages-backend/loggers/src/formatters/rewrite-fields.ts
tdeekens: Would work too. Usually I find merging more scalable. Will adjust here.

.changeset/mighty-donkeys-sip.md
```diff
@@ -0,0 +1,17 @@
...
+
+Support to not unset the original field when rewriting logs.
+
+When rewriting a field you can use the `unsetFromField` option and set it to `false`. For example:
+
+```js
+{
+  from: 'meta.req.statusCode',
+  to: 'meta.req.status_code',
+  unsetFromField: false,
```
emmenko, line 13:
```suggestion
Add support to optionally preserve the original field when rewriting log fields with `rewriteFieldsFormatter`.

When rewriting a field you can use the `preserveFromField` option and set it to `true`. For example:

```js
{
  from: 'meta.req.statusCode',
  to: 'meta.req.status_code',
  preserveFromField: true,
```

emmenko (commented):
Thanks!

emmenko (approved):

PR #3142, opened 2023-07-19
feat(context): add project extension import data

tylermorrisford:
#### Summary

This PR adds `sampleDataImportDataset` to application-context.

### DIFF
--- a/.changeset/kind-buckets-burn.md
+++ b/.changeset/kind-buckets-burn.md
@@ -1,0 +1,6 @@
+---
+'@commercetools-frontend/application-shell-connectors': minor
+'@commercetools-frontend/application-shell': minor
+---
+
+Add `sampleDataImportDataset` to project query, map value to project context.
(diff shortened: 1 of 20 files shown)


emmenko:
Hi @tylermorrisford ,

When would you need this information? The thing is, adding it here will result in always fetching the data every time the project loads. Is that intended?
Or do you need this data only in specific places?

tdeekens:
For context: this data will always be needed upon boot. It will guide different things:

1. Be added to FullStory as additional context to be able to split behaviour of users who have imported sample data and those who have not and how the use the MC. This is crucial for metrics
2. Be added to Intercom to improve the guided tours based on knowing if a user has imported sample data or not. In fact this already is a boot query executed a lot as a result now. This in turn is a follow-up to that
3. Used on the welcome page to alternate the widgets shown

As a result it's needed in so many cases (all apps) that I recommended making adding it here and resolving it in our API to not slow down the start of the MC.

stephsprinkle (approved):

jaikamat (approved):

Issue #3122, opened 2023-06-28
Improvement to how access is granted for local custom app development


mp-marc:
**Is your feature request related to a problem? Please describe.**

The current method for allowing access to local custom app development does not seem very practical, in the way it requires admin accounts to be used
 
**Describe the solution you'd like**

It would be great if instead of requiring admin rights for local development/login, the permission was instead at a team level

**Describe alternatives you've considered**

Given that admin access is required, there aren't many, if any alternatives that we are aware of - though any help here would be appreciated! 

**Additional context**

Hi!

Am hoping this is an issue you can help with - was advised to create an issue here by commercetools support (ticket ref https://commercetools.atlassian.net/servicedesk/customer/portal/22/SUPPORT-21751?created=true) 

For context, we’ve recently been asked to give the engineers in one of our teams the ability to develop and run the custom apps locally - however in doing so we have discovered that this is only possible if the users in question are part of the Administrators group 

This seems to us a quite strange requirement, as the normal approach with an admin team is to restrict membership to as few people as possible (who are responsible for granting access to the rest of the organisation) - whereas with the method used by CT it would mean (depending on requirements) that potentially everyone in an organisation would be an administrator. This is not an ideal situation, given the functions and options that are available (i.e. ability to do pretty much anything). 

We were wondering therefore if there are any alternatives available - for example a permission which can be set at a team level?

Also I would be grateful if could help us understand to the motivation behind this restriction on developer access, as would seem more logical and intuitive (e.g.) for this access to controlled at a team level? 

Many thanks!

Marc

kark:
Hi @mp-marc 👋

Thank you for reaching out to us and providing feedback.

As mentioned, local development and login are restricted to users who are part of the Administrators team within an organization. This restriction is in place because, during the development of a Custom Application, there is potential for accessing any type of data within a project. Even if granular permissions are configured for the Custom Application, they are not applied in dev mode.

We will internally discuss your situation and attempt to find a solution for you.

mp-marc:
Hi @kark 👋, hope all is good! 
Many thanks for your quick reply, and providing that additional context! 

I did wonder if this restriction is related to the type of data available - as such I do understand the desire to have some protection in place. However it does feel awkward having to grant administrator rights to potentially so many people! Also the concept of needing admin rights to perform local development does seem a strange one, and not something we have encountered before. 

If a solution could be found that did not require such access, I'm sure it would benefit all customers of Commercetools, and not just ourselves!

[...]

PR #3143, opened 2023-07-20
feat(app-shell, app-config): enable sending `application_id` claim

kark:
#### Summary

Further changes needed to enable Custom App development for non-admins ([ref](https://github.com/commercetools/merchant-center-application-kit/issues/3122))

### DIFF
--- a/packages/application-config/schema.json
+++ b/packages/application-config/schema.json
@@ -111,6 +111,9 @@
             },
             "teamId": {
               "type": "string"
+            },
+            "applicationId": {
+              "type": "string"
             }
           },
           "additionalProperties": false,
--- a/packages/application-config/src/process-config.ts
+++ b/packages/application-config/src/process-config.ts
@@ -129,6 +129,7 @@ const processConfig = ({
                 ? undefined
                 : appConfig.env.development.initialProjectKey,
             teamId: appConfig.env.development?.teamId,
+            applicationId: appConfig.env.development?.applicationId,
             oAuthScopes: appConfig.oAuthScopes,
             additionalOAuthScopes: appConfig?.additionalOAuthScopes,
           }),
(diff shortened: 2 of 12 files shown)


CarlosCortizasCT (approved):
Looks good to me 💯 
Thanks!

Rhotimee (approved):
🙌🏽

ddouglasz (approved):
Thank you 🙏🏾

tdeekens (approved):

packages/application-config/schema.json
```diff
@@ -111,6 +111,9 @@
             },
             "teamId": {
               "type": "string"
+            },
+            "applicationId": {
+              "type": "string"
```
emmenko, line 116: We already have it in the `production` section. Do you think we need to have it duplicated here? It would probably be the same ID so I think we can rely on the `env.production.applicationId`.

packages/application-config/test/fixtures/config-oidc.json
emmenko, line 1: Let's maybe have a separate fixture scenario where the `teamId` is used.

packages/application-config/src/process-config.ts
```diff
@@ -129,6 +129,7 @@ const processConfig = ({
                 ? undefined
                 : appConfig.env.development.initialProjectKey,
             teamId: appConfig.env.development?.teamId,
+            applicationId: appConfig.env.development?.applicationId,
```
emmenko, line 132: Should we validate that the `applicationId` is provided if the `teamId` is used?

emmenko (commented):
@kark Sorry for the late review. Can you check my comments and maybe address them as a follow up?

Thanks

kark:
> @kark Sorry for the late review. Can you check my comments and maybe address them as a follow up?
> 
> Thanks

Thanks, sure. Sorry for not checking with you before merging

PR #3148, opened 2023-07-25
fix: added missing cloud identifier

mmaltsev-ct:
#### Summary

Added missing cloud identifier for China environment (`aws-cn`) in the [application config page](https://docs.commercetools.com/custom-applications/api-reference/application-config#cloudidentifier) of docs


emmenko (approved):
Thanks!

PR #3149, opened 2023-07-26
`CustomPanel` new component

CarlosCortizasCT:
#### Summary

This PR introduces a new component (`CustomPanel`) which will be used in the context of the **Custom Views** project.

#### Description

I thought about two strategies to implement this new component which will contain the customer's custom views when they select to use a sidebar panel.

image

The first one is try to reuse the same component we use for our modal page components.
This is easier as we don't need to implement everything from the scratch but maybe the transition is not the exact one expected (need to confirm this point). Also, we're extending our current shared `ModalPage` component to include new use cases, making it a little more complex.

The second one is to implement a brand new component.
This one takes a little more time since we need to build everything (maybe we can search for a package that implements the behaviour of this component) but allows us to control every aspect.

In this PR I'm first going for the first option but I'm not 100% satisfied with this approach.
Please let me know your thoughts about it.

Anyway, I would like to have at least a basic implementation available so we can focus in the rest components which are more relevant.
We can come back to this one as a follow-up later so we don't block the rest of the work.

### DIFF
--- a/packages/application-components/src/components/custom-views/custom-panel/custom-panel.tsx
+++ b/packages/application-components/src/components/custom-views/custom-panel/custom-panel.tsx
@@ -3,7 +3,7 @@ import styled from '@emotion/styled';
 import { designTokens } from '@commercetools-uikit/design-system';
 import ModalPage from '../../modal-pages/internals/modal-page';
 
-type TCustoPanelContainerProps = {
+type TCustomPanelProps = {
   title: string;
   size: 'small' | 'large';
   onClose: () => void;
@@ -14,7 +14,7 @@ const ContentWrapper = styled.div`
   padding: ${designTokens.spacing40} 40px;
 `;
 
-function CustomPanelContainer(props: TCustoPanelContainerProps) {
+function CustomPanel(props: TCustomPanelProps) {
   return (
     <ModalPage
       hidePathLabel
@@ -28,4 +28,4 @@ function CustomPanelContainer(props: TCustoPanelContainerProps) {
   );
 }
 
-export default CustomPanelContainer;
+export default CustomPanel;
--- a/packages/application-components/src/components/modal-pages/internals/modal-page-top-bar.tsx
+++ b/packages/application-components/src/components/modal-pages/internals/modal-page-top-bar.tsx
@@ -50,12 +50,17 @@ type Props = {
   color: 'surface' | 'neutral';
   currentPathLabel?: string;
   previousPathLabel: Label;
+  hidePathLabel?: boolean;
   onClose: (event: SyntheticEvent) => void;
   children?: never;
 };
-const defaultProps: Pick<Props, 'color' | 'previousPathLabel'> = {
+const defaultProps: Pick<
+  Props,
+  'color' | 'previousPathLabel' | 'hidePathLabel'
+> = {
   color: 'surface',
   previousPathLabel: messages.back,
+  hidePathLabel: false,
 };
 
 const ModalPageTopBar = (props: Props) => {
@@ -97,16 +102,18 @@ const ModalPageTopBar = (props: Props) => {
           }
         `}
       >
-        <FlatButton
-          tone="primary"
-          label={
-            typeof props.previousPathLabel === 'string'
-              ? props.previousPathLabel
-              : intl.formatMessage(props.previousPathLabel)
-          }
-          icon={<AngleLeftIcon size="medium" color="primary" />}
-          onClick={props.onClose}
-        />
+        {!props.hidePathLabel && (
+          <FlatButton
+            tone="primary"
+            label={
+              typeof props.previousPathLabel === 'string'
+                ? props.previousPathLabel
+                : intl.formatMessage(props.previousPathLabel)
+            }
+            icon={<AngleLeftIcon size="medium" color="primary" />}
+            onClick={props.onClose}
+          />
+        )}
         {props.currentPathLabel && (
           <>
             <Text.Detail as="span">/</Text.Detail>
--- a/packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
+++ b/packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
@@ -5,6 +5,7 @@ export const TRANSITION_DURATION = 200;
 
 type StyleProps = {
   zIndex?: number;
+  size?: 'small' | 'large';
 };
 
 export const getContainerStyles = (_props: StyleProps): SerializedStyles => css`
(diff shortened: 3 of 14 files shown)


CarlosCortizasCT:
This is how it looks the **large** size (it looks like the `InfoModalPage`):
image


This is how it looks the **small** size:
image

packages/application-components/src/components/modal-pages/internals/modal-page-top-bar.tsx
```diff
@@ -97,16 +102,18 @@ const ModalPageTopBar = (props: Props) => {
...
-          tone="primary"
-          label={
-            typeof props.previousPathLabel === 'string'
-              ? props.previousPathLabel
-              : intl.formatMessage(props.previousPathLabel)
-          }
-          icon={<AngleLeftIcon size="medium" color="primary" />}
-          onClick={props.onClose}
-        />
+        {!props.hidePathLabel && (
```
CarlosCortizasCT, line 105: We don't want the "Go back" link in the new component.

packages/application-components/src/components/custom-views/custom-panel/custom-panel.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+type TCustomPanelProps = {
+  title: string;
+  size: 'small' | 'large';
+  onClose: () => void;
+  children: ReactNode;
+};
+
+const ContentWrapper = styled.div`
+  padding: ${designTokens.spacing40} 40px;
```
kark, line 14: Just wanted to double check if the horizontal padding should be out of our spacings scale.

kark (approved):
Thank you Carlos  🙌

> In this PR I'm first going for the first option but I'm not 100% satisfied with this approach.

I'm only wondering if there are other drawbacks that result from this PR or just changes in the ModalPage api?

I think it looks OK if the intention is to have some follow-up iterations.

> **_NOTE_** change set

packages/application-components/src/components/custom-views/custom-panel/custom-panel.tsx
CarlosCortizasCT: So this is coming from an `app-kit` [design token](https://github.com/commercetools/merchant-center-application-kit/blob/main/packages/application-components/src/theming.ts#L6) we will check if we want to keep or move to `ui-kit` in a follow-up task we have in our backlog.

CarlosCortizasCT:
> I'm only wondering if there are other drawbacks that result from this PR or just changes in the ModalPage api?

I'm not aware of other drawbacks.
I'm not 100% satisfied because we're using internally  a `page` component for something that's not a page.

> I think it looks OK if the intention is to have some follow-up iterations.

I discussed with Filip we might want to have a Drawer ([example](https://mui.com/material-ui/react-drawer/)) component in `ui-kit` we could use here and we agree that it could be a good idea.

packages/application-components/src/components/custom-views/custom-panel/custom-panel.tsx
```diff
@@ -0,0 +1,31 @@
...
+`;
+
+function CustomPanel(props: TCustomPanelProps) {
+  return (
+    <ModalPage
+      hidePathLabel
+      isOpen
```
emmenko, line 21: Just checking if it's expected the modal to always being open on render

packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
```diff
@@ -5,14 +5,15 @@ export const TRANSITION_DURATION = 200;
...
+  size?: 'small' | 'large';
 };
 
 export const getContainerStyles = (_props: StyleProps): SerializedStyles => css`
   position: absolute;
   top: 0;
   right: 0;
   height: 100%;
-  width: 100%;
+  width: ${_props.size === 'small' ? '600px !important' : '100%'};
```
emmenko, line 16: Wondering if we should use `maxWidth` instead of a fixed width? What happens when you resize your viewport?

emmenko (commented):
🤗

packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
CarlosCortizasCT: I think we do want to use a fixed size. If we use `max-width`, the panel can be narrower than we want it to be.

Anyway, it can be addressed as a follow-up when Natalia is back so I can double check with her.

[App-kit-playground - e2e-testaccount-39 - Merchant Center.webm](https://github.com/commercetools/merchant-center-application-kit/assets/97907068/0a91b3f3-23bb-40a4-b7aa-158e5e2b4cb1)

playground/src/components/custom-views/custom-panel-demo.jsx
```diff
@@ -0,0 +1,51 @@
+import { useState } from 'react';
+import { CustomPanel } from '@commercetools-frontend/application-components';
+import SecondaryButton from '@commercetools-uikit/secondary-button';
+import Spacings from '@commercetools-uikit/spacings';
+import Text from '@commercetools-uikit/text';
+
+function CustomPanelDemo() {
+  const [isOpen, setIsOpen] = useState(false);
```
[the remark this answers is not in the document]
CarlosCortizasCT: Updated here: fcde43f3914b28595639e695828e31bc3283c914

packages/application-components/src/components/custom-views/custom-panel/custom-panel.tsx
CarlosCortizasCT: Yes it is.

packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
emmenko: Ok, no problem. Thanks for double checking.

I think it's fine then, 600px is small enough to be within the "minimal supported" viewport.

chloe0592 (approved):
Looks all good to me! Thanks 🙌

playground/src/components/custom-views/custom-panel-demo.jsx
```diff
@@ -0,0 +1,55 @@
...
+import {
+  CustomPanel,
+  useModalState,
+} from '@commercetools-frontend/application-components';
+import SecondaryButton from '@commercetools-uikit/secondary-button';
+import Spacings from '@commercetools-uikit/spacings';
+import Text from '@commercetools-uikit/text';
+
+function CustomPanelDemo() {
+  // const [isOpen, setIsOpen] = useState(false);
```
ddouglasz, line 11: Looks like we won't be needing it anymore
```suggestion
```

ddouglasz (approved):
Thank you 🙌🏽

emmenko (approved):
💯

emmenko:
@CarlosCortizasCT ℹ️ A changeset is missing

PR #3151, opened 2023-08-01
New component to load Custom Views

CarlosCortizasCT:
#### Summary

New component to load Custom Views.

#### Description

This new component will be used to actually load in native application the external Custom Views.

It will use an iFrame to isolate contexts although the Custom View code will be loaded by using our `http-proxy`.

Internally the host application and the Custom View will have a communication channel using the [Channel Messaging API](https://developer.mozilla.org/en-US/docs/Web/API/Channel_Messaging_API). It will initially be use to send the host application context to the Custom View so it knows about the project and language the user is using.


packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.spec.tsx
```diff
@@ -0,0 +1,128 @@
...
+      <CustomViewLoader customView={TEST_CUSTOM_VIEW} onClose={jest.fn()} />
+    );
+
+    const iFrame = screen.getByTitle(
+      `Custom View: ${TEST_CUSTOM_VIEW.defaultLabel}`
+    ) as HTMLIFrameElement;
+    expect(iFrame.getAttribute('id')).toBe(
+      `custom-view-${TEST_CUSTOM_VIEW.id}`
+    );
+    expect(iFrame.getAttribute('src')).toBe(TEST_CUSTOM_VIEW.url);
```
CarlosCortizasCT, line 70: I wanted to also check the initial `postMessage` call but I don't find a way to replicate in the test a successful iFrame load.
If anyone has any idea, please share it 🙏

packages/application-components/src/components/custom-views/custom-view-loader/messages.ts
```diff
@@ -0,0 +1,11 @@
+import { defineMessages } from 'react-intl';
+
+const messages = defineMessages({
+  loadError: {
```
CarlosCortizasCT, line 4: Not sure how to extract this message to the `core` json files in the `i18n` packages.
In that package README [it says](https://github.com/commercetools/merchant-center-application-kit/tree/main/packages/i18n#generating-translation-files) to run the `mc-scripts extract-intl [options]` command, buit it seems `extract-intl` does not exist anymore.
Any help would be much appreciated..

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -0,0 +1,142 @@
...
+      showNotification({
+        domain: DOMAINS.PAGE,
+        kind: NOTIFICATION_KINDS_PAGE.error,
+        text: intl.formatMessage(messages.loadError),
+      });
+      return;
+    }
+
+    // Listen for messages from the iFrame
+    intercomChannel.current.port1.onmessage = messageFromIFrameHandler;
```
CarlosCortizasCT, line 92: We're not expecting the iFrame to send any message in the MVP.
I left this part here mostly for reference purposes but maybe we should remove it.

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -0,0 +1,142 @@
...
+      console.log('message received from iframe port: ', event);
+    }
+  }, []);
+
+  // onLoad handler is called from the iFrame even where the URL is not valid
+  // (blocked by CORS, 404, etc.) so we need to make sure the iFrame is ready
+  const onLoadSuccessHandler = useCallback(() => {
+    // Show error and block if the iFrame is not ready
+    // (error loading it)
+    if (!isIframeReady(iFrameElementRef.current!)) {
```
CarlosCortizasCT, line 82: Based on my tests, I found out the `onLoad` callback is called even when the iFrame fails to load so we need to make sure here if it was loaded or not.

I also found [this reference](https://javascript.info/onload-onerror#other-resources):
image

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -0,0 +1,142 @@
...
+      source: 'mc-host-application',
+      destination: `custom-view-${props.customView.id}`,
+      eventName: 'hostAcknowledgesIframeLoaded',
+      eventData: {
+        appContext,
+      },
+    } as TCustomViewIframeMessage);
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
CarlosCortizasCT, line 112: We want the effect to run only once so I think we don't need the dependencies array.

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -0,0 +1,142 @@
...
+      [intercomChannel.current.port2]
+    );
+
+    // Send the initialization message to the iFrame
+    intercomChannel.current.port1.postMessage({
+      source: 'mc-host-application',
+      destination: `custom-view-${props.customView.id}`,
+      eventName: 'hostAcknowledgesIframeLoaded',
+      eventData: {
+        appContext,
```
CarlosCortizasCT, line 107: Currently we will only need the projectKey and the data language in the iFrame, but I thought passing all the application context just in case.
Let me know if you think we should be more specific.

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -0,0 +1,142 @@
...
+      new Error(
+        `CustomViewLoader: Provided Custom View has an unsupported type: ${props.customView.type}`
+      )
+    );
+    return null;
+  }
+
+  return (
+    <CustomPanel
+      title={`Custom View: ${props.customView.defaultLabel}`}
```
emmenko, line 126: Is the prefix needed?

emmenko (commented):
Wow, cool stuff! Looking forward to see it in action.

I left some minor comments and suggestions, otherwise 👌

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -0,0 +1,142 @@
...
+  onClose: () => void;
+};
+
+const isIframeReady = (iFrameElementRef: HTMLIFrameElement) => {
+  try {
+    return iFrameElementRef?.contentWindow?.document.readyState === 'complete';
+  } catch {
+    // Trying to access the contentWindow of a cross-origin iFrame will throw an error.
+    // We are not supposed to even get here because the iFrame must use
+    // a URL from our very same domain (the custom view is proxied through our http-proxy service).
```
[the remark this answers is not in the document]
CarlosCortizasCT: I was hesitant about it.

This should not happen because the type of the Custom View should come from a select input in the configuration view so, if it happens, I think it would be because of a bug in our side, that's why I decided to report it.

But we can also just log it into the console if we prefer to not have this kind of errors in Sentry.

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -0,0 +1,142 @@
...
+function CustomViewLoader(props: TCustomViewLoaderProps) {
+  const iFrameElementRef = useRef<HTMLIFrameElement>(null);
+  const appContext = useApplicationContext();
+  const intercomChannel = useRef(new MessageChannel());
+  const showNotification = useShowNotification();
+  const intl = useIntl();
+
+  const messageFromIFrameHandler = useCallback((event: MessageEvent) => {
+    if (event.data.origin === window.location.origin) {
+      console.log('message received from iframe port: ', event);
```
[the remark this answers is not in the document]
CarlosCortizasCT: It is related to [this comment](https://github.com/commercetools/merchant-center-application-kit/pull/3151#discussion_r1281474433).

packages/application-components/src/components/custom-views/custom-view-loader/messages.ts
CarlosCortizasCT: Oh! thanks.

Message added here: 25db0c13a6356c58065cb5bb8e3aa937b949b343

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
CarlosCortizasCT: Ok. 

Updated here: 041828a83cab9f34b7a92a11a099dcd6546c3369

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -0,0 +1,142 @@
...
+      eventData: {
+        appContext,
+      },
+    } as TCustomViewIframeMessage);
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  // Currently we only support custom panels
+  if (props.customView.type !== 'custom-panel') {
```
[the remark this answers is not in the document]
CarlosCortizasCT: Yes, you're right. Thanks.

Updated here: d51ec978be840b860f3e07a008aaf5bdd9e58260

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
CarlosCortizasCT: Since titles should be user friendly, I think it's worth it to have the prefix so readers can know what the iFrame is about.

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.spec.tsx
```diff
@@ -0,0 +1,128 @@
...
+  fireEvent,
+} from '../../../test-utils';
+import CustomViewLoader, { type TCustomView } from './custom-view-loader';
+
+const mockShowNotification = jest.fn();
+
+jest.mock('@commercetools-frontend/sentry');
+jest.mock('@commercetools-frontend/actions-global', () => ({
+  useShowNotification: () => mockShowNotification,
+}));
```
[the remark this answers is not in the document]
CarlosCortizasCT: If we want the notification to work, we need to setup Redux for the test and I thought it would be easier if we mock it.

Do you prefer us to actually render the notification?

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.spec.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+const TEST_CUSTOM_VIEW: TCustomView = {
```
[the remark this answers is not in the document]
CarlosCortizasCT: Added here: b03ee02b9af76ac30f57590821276ff81577bfde

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.spec.tsx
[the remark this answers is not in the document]
emmenko: We have the `renderAppWithRedux`. Is it possible to use that?

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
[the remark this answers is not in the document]
emmenko: Ok, I think it's fine then to report it to Sentry to help us catching such (unlikely) cases.

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
emmenko: Makes sense. If you think it helps, you can also leave a short comment about it.

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -0,0 +1,149 @@
...
+      'mc_init',
+      window.location.href,
+      [iFrameCommunicationChannel.current.port2]
+    );
+
+    // Send the initialization message to the iFrame
+    iFrameCommunicationChannel.current.port1.postMessage({
+      source: 'mc-host-application',
+      destination: `custom-view-${props.customView.id}`,
+      eventName: 'hostAcknowledgesIframeLoaded',
```
emmenko, line 108: Is this a standard event name or something we need to define?

If the latter, maybe we can put these constants into a constants file/object, as I see a bunch of other cases here like `mc_init`, `mc-host-application`, etc.
Also the casing seems to be a bit inconsistent. Is this on purpose?

emmenko (commented):
Looks good! 🙌

Some final minor comments but otherwise this is good to go from my side. Thanks 🤗

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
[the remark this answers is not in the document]
CarlosCortizasCT: Added here: d790a17fb8197e64f1c44a54c64eec708e911e7e

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.spec.tsx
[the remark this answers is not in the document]
emmenko: Ah right I forgot about that. Ok, then it's fine for now.

emmenko (approved):
🥳 Thanks, awesome work!

chloe0592:
Nice! Awesome work @CarlosCortizasCT! 👏 👏 👏

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+    // We want the effect to run only once so we don't need the dependencies array.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  // Currently we only support custom panels
+  if (props.customView.type !== 'CustomPanel') {
+    reportErrorToSentry(
+      new Error(
+        `CustomViewLoader: Provided Custom View has an unsupported type: ${props.customView.type}`
```
chloe0592, line 130: Nit: I'm thinking maybe it might be useful to add in the error message what is the only supported type at the moment. 
Something like: `CustomViewLoader: The Custom View has an unsupported type: ${props.customView.type. Presently, the sole supported type is 'CustomPanel'.`

chloe0592 (approved):

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
CarlosCortizasCT: Thanks for the suggestion 👍 
I updated the message here: 744c59951b328d9dac65dd2b09fcd11cc814e981

Bear in mind this is a very unlikely error to happen and it's more for internal debug that customers facing ([explanation](https://github.com/commercetools/merchant-center-application-kit/pull/3151#discussion_r1281688722)).

PR #3152, opened 2023-08-02
fix: vite config for playground app

kark:
#### Summary

The issue:
Screenshot 2023-08-02 at 13 43 29

This worked for me.
As an alternative we could downgrade `vite` in the monorepo to `4.1.4`.

### DIFF
--- a/packages/mc-scripts/src/commands/start-vite.ts
+++ b/packages/mc-scripts/src/commands/start-vite.ts
@@ -46,6 +46,9 @@ async function run() {
       pluginSvgr(),
       pluginCustomApplication(applicationConfig),
     ],
+    optimizeDeps: {
+      force: true,
+    },
   });
   await server.listen();
 
--- a/packages/mc-scripts/src/commands/start-vite.ts
+++ b/packages/mc-scripts/src/commands/start-vite.ts
@@ -46,9 +46,6 @@ async function run() {
       pluginSvgr(),
       pluginCustomApplication(applicationConfig),
     ],
-    optimizeDeps: {
-      force: true, // based on https://github.com/storybookjs/storybook/issues/22253#issuecomment-1661092137 TODO: remove when the issue with vite is resolved
-    },
   });
   await server.listen();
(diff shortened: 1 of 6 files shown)


chloe0592 (approved):
Thanks a lot for this fix! 🙇‍♀️

ddouglasz (approved):
👏

packages/mc-scripts/src/commands/start-vite.ts
```diff
@@ -46,6 +46,9 @@ async function run() {
       pluginSvgr(),
       pluginCustomApplication(applicationConfig),
     ],
+    optimizeDeps: {
```
CarlosCortizasCT, line 49: I'm having problems checking this locally, but I'd like to share my opinion anyway: if we configure it like this, I believe it will be harder for us to keep track of the workaround. I mean, it is a change in a file that we don't change quite often. On the other hand, if we downgrade `vite` version, even `renovate` bot will warn us about it so we can have a PR where we can add a label "blocking" it with a comment describing why.

kark: Thanks. Np.
If this doesn't work then we can downgrade vite. I felt that downgrading would be a last resort.

CarlosCortizasCT (approved):
Thanks 👍

CarlosCortizasCT:
When installing the dependencies, I see this warning in the terminal:
image

The playground seems to work anyway, but maybe we would need to consider downgrading that package as well.

kark:
> When installing the dependencies, I see this warning in the terminal: image
> 
> The playground seems to work anyway, but maybe we would need to consider downgrading that package as well.

Thanks, that's a good catch.
We need to go down one major version.
I think we weren't using any new features:
https://github.com/vitejs/vite-plugin-react/blob/main/packages/plugin-react/CHANGELOG.md#400-2023-04-20

CarlosCortizasCT (approved):
Thanks! 🙇

PR #3153, opened 2023-08-02
Extend default requests headers

CarlosCortizasCT:
#### Summary

Extend default requests headers

#### Description

We're extending the list of default headers we include in all requests to add the (upcoming) custom views identifier when it is defined in the application config.


packages/application-shell/src/configure-apollo.ts
```diff
@@ -40,6 +41,8 @@ const createApolloLink = (options: TApolloClientOptions = {}) =>
     // See https://www.apollographql.com/docs/react/api/link/apollo-link-rest/#link-order
     // State & context links should happen before (to the left of) `restLink`.
     ...(options.restLink ? [options.restLink] : []),
+    // Consumer provided link should happen after we set the base headers
+    ...(options.customLinks ?? []),
```
emmenko, line 45: The order of the links is important. This does not guarantee that any provided link will work properly.

That's why for example we allowed the `restLink` as a specific option, so that we can position it correctly in the links chain.

Can you share what would be the use cases to support custom links?

CarlosCortizasCT: Ok. So the idea here is that we might want to allow consumers to provide their custom links and you're right we don't know if the position we will set them in the chain would be the right one because we don't know what the custom link intends to do.
My goal was to set them in a sensible position in the current chain and state in the (future) docs how and when they will be executed.

If I look at it with a narrower scope, our first use case is to add a custom header so, instead of accepting an array of custom links, we can just expose another optional link or even a simple function where we provide the base header and consumers can extend them.

do you think that solution would fit better?

emmenko: > our first use case is to add a custom header

I might be missing some background here. What custom header are we talking about? 🤔

emmenko: I don't think we need a custom link for this, we should be able to reuse the normal header link and include the header functionality there.

I might be missing something though. It would be good to have a sync maybe, thanks.

packages/application-shell/src/utils/http-client.ts
```diff
@@ -197,6 +197,7 @@ function createHttpClientOptions(config: TConfig = {}): TOptions {
         ? `Bearer ${sessionToken}`
         : undefined,
       [SUPPORTED_HEADERS.X_APPLICATION_ID]: window.app.applicationId,
+      [SUPPORTED_HEADERS.X_CUSTOM_VIEW_ID]: window.app.customViewId,
```
kark, line 200: I'm only wondering if we should send the header if `customViewId` is undefined?
```suggestion
      ...(window.app.customViewId
        ? { [SUPPORTED_HEADERS.X_CUSTOM_VIEW_ID]: window.app.customViewId }
        : {}),
```

kark (approved):
Thanks!

packages/application-shell/src/utils/http-client.ts
CarlosCortizasCT: Thanks Kacper 👍 

I had the same thought at first, but notice this is inside a call to the `omitEmpty` helper function which already would remove it in such case.

VictorAvelar (approved):

chloe0592 (approved):

PR #3154, opened 2023-08-03
fix(mc-scripts): sends the app identifier header when using the MC API

emmenko:
To ensure that the logic around permissions in the API works, as it relies on the application identifier.


CarlosCortizasCT (approved):
LGTM 👍 

Thanks!

kark (approved):
🙌

PR #3162, opened 2023-08-07
New component to wrap Custom Views

CarlosCortizasCT:
#### Summary

New component to wrap Custom Views

#### Description

The main feature of this PR is the implementation of the component `CustomViewShell`. 
The purpose of this component is to wrap the upcoming Custom Views in order to integrate them in the Merchant Center by initializing it's context so they can properly work inside a Merchant Center application.
This should be the root component of a Custom View implementation so developers can focus on their use case and not in the integration details.

The key components introduced in this PR are:
* [CustomViewShell](https://github.com/commercetools/merchant-center-application-kit/pull/3162/files#diff-15aa67882ed3820b8ed538e323d2d9df2714e3ffd7140e9d7b55e3d15bf73783)
* [CustomViewAuthenticatedShell](https://github.com/commercetools/merchant-center-application-kit/pull/3162/files#diff-856fbab4e083bf369a9123003a2e9b5e1f3fbb6f02b9ddd6ddcb08c5786e2375)
* [CustomViewContext](https://github.com/commercetools/merchant-center-application-kit/pull/3162/files#diff-46c9a960e43894a9119bec2f6b51b8888e9738701697f1a658550b1f2df35073)

We are also including a new section in the playground application where we emulate loading a demo Custom View so we can get a grasp of how working with them will look like and to ensure the new components work as expected together.

There are another bunch of files updated in this PR that are not key to review:
* **graphql-test-utils**: Files in this folder just were renamed to `.ts` since we now use it from a Typescript context
* **backend services types**: updated these types to get the new shapes for Custom Views. Also, we included `mc-settings` types in the `application-component` package as we need them in some components over there.


package.json
```diff
@@ -117,6 +117,7 @@
     "@types/eslint": "^8.2.2",
     "@types/jest": "^29.5.1",
     "@types/node": "^18.15.11",
+    "@types/rosie": "^0.0.42",
```
CarlosCortizasCT, line 120: All the changes in the `graphql-test-utils` directory is just renaming files from `.js` to `.ts` so we don't have typescript errors when using them in the new typescript test files.

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.spec.tsx
```diff
@@ -68,7 +68,9 @@ describe('CustomViewLoader', () => {
     expect(iFrame.getAttribute('id')).toBe(
       `custom-view-${TEST_CUSTOM_VIEW.id}`
     );
-    expect(iFrame.getAttribute('src')).toBe(TEST_CUSTOM_VIEW.url);
+    expect(iFrame.getAttribute('src')).toContain(
+      `/no-project/custom-view/${TEST_CUSTOM_VIEW.id}`
```
CarlosCortizasCT, line 72: The expected URL for the iFrame is:
`<MC_HOST>/<project_key>/custom-view/<custom_view_id>`

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -144,7 +146,9 @@ function CustomViewLoader(props: TCustomViewLoaderProps) {
         key={`custom-view-${props.customView.id}`}
         ref={iFrameElementRef}
         title={`Custom View: ${props.customView.defaultLabel}`}
-        src={props.customView.url}
+        src={`${window.location.origin}/${
```
CarlosCortizasCT, line 149: iFrame URL will be a MC one since we want to proxy the request, the same way we do for Custom Applications:
```
https://<MC_HOST>/<project_key>/custom-view/<custom_view_id>
```

This URL will not be visible for users.
It has a fixed path so we can treat it specially in the `http-proxy`.

packages/application-shell/src/components/custom-view-authenticated-shell/custom-view-authenticated-shell.tsx
```diff
@@ -0,0 +1,100 @@
...
+                  <ApplicationContextProvider
+                    user={user}
+                    environment={props.environment}
+                  >
+                    <SetupFlopFlipProvider
+                      user={user}
+                      projectKey={props.projectKey}
+                      ldClientSideId={props.environment.ldClientSideId}
+                      // flags={props.featureFlags}
+                      // defaultFlags={props.defaultFeatureFlags}
```
CarlosCortizasCT, line 58: Not sure if we want to allow feature flags in Custom Views. If so, also not sure which ones should we pass (defaultFeatureFlags?).

packages/application-shell/src/components/custom-view-context/custom-view-context.tsx
```diff
@@ -0,0 +1,68 @@
...
+type TCustomViewContext = {
+  config: TCustomView;
+};
+
+type TCustomViewContextProviderProps = {
+  config: TCustomView;
+  children: ReactNode;
+};
+
+const CustomViewContext = createContext<TCustomViewContext>({
```
CarlosCortizasCT, line 34: This is useful to have the CustomView configuration easily available in its components tree.

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
```diff
@@ -0,0 +1,132 @@
...
+        window.removeEventListener('message', initializationMessageHandler);
+      } else {
+        console.warn(
+          `CustomViewShell: Received an event that is not allowed: ${event.data}`,
+          { event }
+        );
+      }
+    };
+
+    window.addEventListener('message', initializationMessageHandler);
```
CarlosCortizasCT, line 83: Once the Custom View has been loaded, we will listen for a bootstrap event from the host MC native application.

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
```diff
@@ -0,0 +1,132 @@
...
+    []
+  );
+
+  useEffect(() => {
+    const initializationMessageHandler = (event: MessageEvent) => {
+      console.log('Custom view shell received message', { event });
+
+      if (
+        event.origin === window.location.origin &&
+        event.data === CUSTOM_VIEWS_EVENTS_NAMES.CUSTOM_VIEW_BOOTSTRAP
```
CarlosCortizasCT, line 68: The only message we should listen in the global channel in the bootstrap one.
After that, we will listen for messages only in the private channel (the first one received in the `ports` array of the event)

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
```diff
@@ -0,0 +1,132 @@
...
+  children: ReactNode;
+};
+
+function CustomViewShell(props: TCustomViewShellProps) {
+  const [hostContext, setHostContext] = useState<THostContext>();
+  const iFrameCommunicationPort = useRef<MessagePort>();
+
+  const hostMessageHandler = useCallback(
+    (event: MessageEvent<THostEventData>) => {
+      if (
```
CarlosCortizasCT, line 47: Here is where we process private messages received from the host.
Messages here can only come from the host MC native application.

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
```diff
@@ -0,0 +1,132 @@
...
+                messages={props.messages}
+                projectKey={hostContext.projectKey}
+              >
+                {props.children}
+              </CustomViewAuthenticatedShell>
+            </CustomViewContextProvider>
+          );
+        }
+
+        // TODO: Render a proper error view
```
CarlosCortizasCT, line 125: Not sure if we need to do anything here since I can't think of a case where the user is not authenticated. He needs to be it in order to open the Custom View from the MC native application.

playground/custom-application-config.mjs
```diff
@@ -31,6 +31,7 @@ const config = {
   headers: {
     csp: {
       'connect-src': ['${env:HOST_GCP_STAGING}'],
+      'frame-src': ['${env:LOCAL_HOST}'],
```
CarlosCortizasCT, line 34: This is to allow rendering a Custom View in the Playground application.

playground/src/components/entry-point/entry-point.jsx
```diff
@@ -20,13 +23,23 @@ setupGlobalErrorListener();
...
 
 const EntryPoint = () => (
-  <ApplicationShell
-    environment={window.app}
-    applicationMessages={loadMessages}
-    apolloClient={apolloClient}
-  >
-    <AsyncPlaygroundRoutes />
-  </ApplicationShell>
+  <Router history={history}>
```
CarlosCortizasCT, line 26: A change to allow loading the Custom View iFrame from the same application.

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
```diff
@@ -0,0 +1,132 @@
...
+import type { TAsyncLocaleDataProps } from '@commercetools-frontend/i18n';
+import LoadingSpinner from '@commercetools-uikit/loading-spinner';
+import ApplicationShellProvider from '../application-shell-provider';
+import CustomViewAuthenticatedShell from '../custom-view-authenticated-shell';
+import { CustomViewContextProvider, TCustomView } from '../custom-view-context';
+import { CUSTOM_VIEWS_EVENTS_NAMES } from './constants';
+
+declare let window: ApplicationWindow;
+
+type THostContext = {
```
CarlosCortizasCT, line 18: These are the properties we will receive from the host application.

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
```diff
@@ -0,0 +1,132 @@
...
+  hostUrl: string;
+  userLocale: string;
+  dataLocale: string;
+  customViewConfig: TCustomView;
+  projectKey?: string;
+};
+
+type THostEventData = {
```
CarlosCortizasCT, line 26: This is the structure of the private messages received from the host application.

playground/src/components/custom-views/demo-custom-view.jsx
```diff
@@ -0,0 +1,87 @@
...
+}
+
+function DemoCustomView() {
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    window.app.customViewId = `__local:${CUSTOM_VIEW_ID}`;
+    document.querySelector('.loading-screen').remove();
+    setIsLoading(false);
+  }, []);
```
CarlosCortizasCT, line 74: This is temporary until we update `mc-scripts` to inject the CustomView ID in the document the same way it does for Custom Applications.

playground/src/components/custom-views/demo-custom-view.jsx
```diff
@@ -0,0 +1,87 @@
...
+];
+
+function ChannelsCustomView() {
+  const { projectName, locale } = useApplicationContext((context) => ({
+    projectKey: context.project.key,
+    projectName: context.project.name,
+    locale: context.user.locale,
+  }));
+
+  const { data, error, loading } = useMcQuery(FetchChannelsQuery, {
```
CarlosCortizasCT, line 34: I don't understand why but this query is failing with this error:
```json
{
  "errors": [
    {
      "message": "Cannot query field \"channels\" on type \"Query\".",
      "extensions": {
        "code": "GRAPHQL_VALIDATION_FAILED"
      }
    }
  ]
}
```

I doubled checked the headers we're sending and they are the same as the ones sent from the Merchant Center channels view.

codegen.settings.yml
```diff
@@ -37,6 +37,16 @@ extensions:
           scalars:
             DateTime: string
             Json: '{ [key: string]: unknown }'
+      packages/application-components/src/types/generated/settings.ts:
```
CarlosCortizasCT, line 40: We need to use Custom Views types in the `application-components` package.

packages/application-components/src/components/custom-views/custom-view-loader/constants.ts
```diff
@@ -1,9 +0,0 @@
-export const CUSTOM_VIEWS_EVENTS_NAMES = {
```
CarlosCortizasCT, line 1: These have been moved to the shared constants package.

playground/src/components/custom-views/custom-panel-demo.jsx
```diff
@@ -1,15 +1,45 @@
...
+  locators: ['/almond-40/app-kit-playground/custom-panel'],
+  permissions: [
+    {
+      name: 'ViewProducts',
+      oAuthScopes: ['view_channels'],
+    },
+  ],
+};
+
 function CustomPanelDemo() {
```
CarlosCortizasCT, line 30: This component just renders a couple of buttons to open a Custom View (with different sizes).

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.spec.tsx
```diff
@@ -20,11 +26,23 @@ const TEST_CUSTOM_VIEW: TCustomView = {
   defaultLabel: 'Test Custom View',
   labelAllLocales: {},
   url: '/',
-  type: 'CustomPanel',
-  typeConfig: {
-    size: 'SMALL',
+  type: TCustomViewType.CustomPanel,
+  typeSettings: {
+    size: TCustomViewSize.Small,
```
emmenko, line 31: 💯

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
emmenko: If there is a Custom Application with the entry point `custom-view`, we might have a conflict / issue here.

Since this will be a route served through the MC Proxy, maybe we can use another (more unique) path?

Some examples:

```
<MC_HOST>/custom-views/<project_key>/<custom_view_id>
<MC_HOST>/frames/<project_key>/custom-views/<custom_view_id>
<MC_HOST>/custom-views/<custom_view_id>?project-key=<project_key>
```

What do you think? Essentially an endpoint that does not start with `<project_key>`.

packages/application-shell/src/components/custom-view-authenticated-shell/custom-view-authenticated-shell.tsx
emmenko: Yeah good point, we can omit the provider for now. Also (external) Custom Applications can't use the feature flags, it's only for internal usage.

playground/src/components/entry-point/entry-point.jsx
emmenko: Not sure I follow 🤔

playground/custom-application-config.mjs
emmenko: Shouldn't we include the `frame-src` directive in the default CSP config?

https://github.com/commercetools/merchant-center-application-kit/blob/51deaa6fc32b3be2439853d952ce8a5ae6db93e3/packages/mc-html-template/src/process-headers.ts#L72-L118

emmenko (commented):
Exciting stuff 🙌

playground/custom-application-config.mjs
CarlosCortizasCT: Yes, you're right.

Since frames are expected to be loaded from the same origin as the host application, we shouldn't need to even add the policy but since we have a [very restrictive default](https://github.com/commercetools/merchant-center-application-kit/blob/main/packages/mc-html-template/src/process-headers.ts#L74), I just added this directive to the default ones to allow frames from the same origin here: c7ddc2df4bd2f6abee3ecf24e31037116e7ae695

playground/src/components/entry-point/entry-point.jsx
CarlosCortizasCT: We are loading both the Playground Custom Application and the Custom View from the same server which renders this same entry point.
When loading the Custom View, we don't wan't to render the `ApplicationShell` (Custom Application) but something totally different (the Custom View demo).

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
```diff
@@ -0,0 +1,142 @@
...
+  const hostMessageHandler = useCallback(
+    (event: MessageEvent<THostEventData>) => {
+      if (
+        event.data.eventName ===
+        CUSTOM_VIEWS_EVENTS_NAMES.CUSTOM_VIEW_INITIALIZATION
+      ) {
+        setHostContext(event.data.eventData.context);
+      } else {
+        console.warn(
+          `CustomViewShell: Unkown received event with name: ${event.data.eventName}`,
```
ddouglasz, line 65:
```suggestion
          `CustomViewShell: Unknown, received event with name: ${event.data.eventName}`,
```

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
```diff
@@ -0,0 +1,142 @@
...
+    const bootstrapMessageHandler = (event: MessageEvent) => {
+      console.log('Custom view shell received message', { event });
+
+      if (
+        event.origin === window.location.origin &&
+        event.data === CUSTOM_VIEWS_EVENTS_NAMES.CUSTOM_VIEW_BOOTSTRAP
+      ) {
+        iFrameCommunicationPort.current = event.ports[0];
+        iFrameCommunicationPort.current.onmessage = hostMessageHandler;
+        // Once bootstraped, we don't want to listen for global messages anymore.
```
ddouglasz, line 83: Thanks for the comment 🙏🏾

ddouglasz (commented):
Thank you @CarlosCortizasCT , really great work 🚀
Have a few feedback 🙏🏾.

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -4,15 +4,14 @@ import { useIntl } from 'react-intl';
 import { useShowNotification } from '@commercetools-frontend/actions-global';
 import { useApplicationContext } from '@commercetools-frontend/application-shell-connectors';
 import {
+  CUSTOM_VIEWS_EVENTS_NAMES,
+  CUSTOM_VIEWS_EVENTS_META,
   DOMAINS,
   NOTIFICATION_KINDS_PAGE,
 } from '@commercetools-frontend/constants';
 import { reportErrorToSentry } from '@commercetools-frontend/sentry';
+import { TCustomView } from '../../../types/generated/settings';
```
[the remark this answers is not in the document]
CarlosCortizasCT: Updated here: 2056097cf6343cbb22abd44c2efc5a88c1d8e893

packages/application-shell/src/components/custom-view-authenticated-shell/custom-view-authenticated-shell.tsx
```diff
@@ -0,0 +1,91 @@
+import { type ReactNode } from 'react';
+import { PageUnauthorized } from '@commercetools-frontend/application-components';
+import { ApplicationContextProvider } from '@commercetools-frontend/application-shell-connectors';
+import { ApplicationWindow } from '@commercetools-frontend/constants';
```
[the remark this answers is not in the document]
CarlosCortizasCT: Updated here: 2056097cf6343cbb22abd44c2efc5a88c1d8e893

packages/application-shell/src/components/custom-view-context/custom-view-context.tsx
```diff
@@ -0,0 +1,62 @@
+import { createContext, type ReactNode, useContext, useMemo } from 'react';
+import {
+  TCustomView,
```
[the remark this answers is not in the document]
CarlosCortizasCT: Updated here: 2056097cf6343cbb22abd44c2efc5a88c1d8e893

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.spec.tsx
```diff
@@ -5,7 +5,13 @@ import {
   waitFor,
   fireEvent,
 } from '../../../test-utils';
-import CustomViewLoader, { type TCustomView } from './custom-view-loader';
+import {
+  TCustomView,
```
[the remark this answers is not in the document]
CarlosCortizasCT: Updated here: 2056097cf6343cbb22abd44c2efc5a88c1d8e893

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
CarlosCortizasCT: Updated here: 2056097cf6343cbb22abd44c2efc5a88c1d8e893

playground/src/components/custom-views/demo-custom-view.jsx
CarlosCortizasCT: This issue has been fixed here: dbdbf6d6a3a3a00353cfaf5b338ad2950a8029f4

Thanks @emmenko for helping in finding the root cause 🙇

playground/.env
```diff
@@ -6,4 +6,5 @@ HOST_GCP_STAGING=""
 ENABLE_NEW_JSX_TRANSFORM="true"
 ENABLE_EXPERIMENTAL_VITE_BUNDLER="true"
 FAST_REFRESH="true"
-LD_CLIENT_ID_STAGING="5979d95f6040390cd07b5e00"
\ No newline at end of file
+LD_CLIENT_ID_STAGING="5979d95f6040390cd07b5e00"
+LOCAL_HOST="http://localhost:3001"
```
emmenko, line 10: Do we need this?

packages/application-shell/src/components/custom-view-context/custom-view-context.tsx
```diff
@@ -0,0 +1,62 @@
...
+type TCustomViewContext = {
+  config: TCustomView;
+};
+
+type TCustomViewContextProviderProps = {
+  config: TCustomView;
+  children: ReactNode;
+};
+
+const CustomViewContext = createContext<TCustomViewContext>({
```
emmenko, line 17: Is this something supposed to be consumed by a Custom View?

Asking because the Custom View will likely have this data anyway exposed via a global variable like `window.app`, which is loaded from the config file.

Do you see other use cases that would benefit from using this context?

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
```diff
@@ -0,0 +1,142 @@
...
+import {
+  AsyncLocaleData,
+  type TAsyncLocaleDataProps,
+} from '@commercetools-frontend/i18n';
+import type { TCustomView } from '../../types/generated/settings';
+import ApplicationLoader from '../application-loader/application-loader';
+import ApplicationShellProvider from '../application-shell-provider';
+import { getBrowserLocale } from '../application-shell-provider/utils';
+import ConfigureIntlProvider from '../configure-intl-provider';
+import CustomViewAuthenticatedShell from '../custom-view-authenticated-shell';
```
emmenko, line 22: Nit: for naming consistency, it should be `custom-view-shell-authenticated`

packages/application-shell/src/utils/select-project-key-from-url/select-project-key-from-url.ts
```diff
@@ -6,7 +6,16 @@ const staticUrlPathsInPositionOfProjectKey = ['login', 'logout', 'account'];
...
+  let possibleProjectKey = '';
+  const pathParts = locationPath.split('/');
+
+  if (pathParts[1] === 'custom-views') {
+    // Custom Views paths: /custom-views/:customViewId/projects/:projectKey
+    possibleProjectKey = pathParts[4];
+  } else {
+    // Application paths: /:projectKey/:applicationId
+    possibleProjectKey = pathParts[1];
+  }
```
emmenko, line 18: 💯

packages/mc-html-template/src/process-headers.ts
```diff
@@ -99,6 +99,8 @@ const processHeaders = (
         // htmlStylesHashes.map(assetHash => `'${assetHash}'`)
       ),
       'font-src': ["'self'", 'fonts.gstatic.com', 'data:'],
+      // Required for Custom Views
+      'frame-src': ["'self'"],
```
emmenko, line 103: 💯

emmenko (approved):
Cool, nice work!

playground/.env
CarlosCortizasCT: Nope. It's not needed anymore. Thanks.

Removed here: 9446cae41fa966890bae8d2f49c790bf0fda335d

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
CarlosCortizasCT: Renamed here: 9446cae41fa966890bae8d2f49c790bf0fda335d

packages/application-shell/src/components/custom-view-context/custom-view-context.tsx
CarlosCortizasCT: The way I see it, `window.app` will not contain all the Custom View configuration and I'm also using this context in a follow-up PR to store another piece of information.
Perhaps we can discuss about it later in that PR.

PR #3165, opened 2023-08-09
refactor(app-shell): new NavBar skeleton

kark:
#### Summary

|  Designs  |  Screencast |
|---|---|
| Screenshot 2023-08-09 at 15 37 03 |   |

[A direct link to deployment in the MC](https://mc.europe-west1.gcp.escemo.com/main-navigation-test/navbar-skeleton)

### DIFF
--- a/.changeset/little-keys-retire.md
+++ b/.changeset/little-keys-retire.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': minor
+---
+
+Use new NavBar skeleton
--- a/.changeset/little-keys-retire.md
+++ b/.changeset/little-keys-retire.md
@@ -2,4 +2,4 @@
 '@commercetools-frontend/application-shell': minor
 ---
 
-Use new NavBar skeleton
+Support a new `NavBar` skeleton ui based on a feature flag.
--- a/packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
+++ b/packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
@@ -367,7 +367,7 @@ export const ApplicationShellAuthenticated = (
                                             // a different React context.
                                           >
                                             <NavBar
-                                              applicationLocale={locale ?? ''}
+                                              applicationLocale={locale}
                                               projectKey={projectKeyFromUrl}
                                               project={project}
                                               environment={
--- a/packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
+++ b/packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
@@ -390,7 +390,6 @@ export const ApplicationShellAuthenticated = (
                                               onMenuItemClick={
                                                 props.onMenuItemClick
                                               }
-                                              // isLoading={isLoading}
                                               isLoading={
                                                 // isLoading && // TODO: uncomment before merging
                                                 isNewNavigationEnabled
--- a/packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
+++ b/packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
@@ -217,7 +217,7 @@ export const ApplicationShellAuthenticated = (
                     <NewMainNavigationFlagWrapper>
                       {({
                         isNewNavigationEnabled,
-                        isNewNavigationEnabledEvaluationPending,
+                        isNewNavigationEnabledEvaluationReady,
                       }) => {
                         return (
                           <>
@@ -383,8 +383,8 @@ export const ApplicationShellAuthenticated = (
                                               isNewNavigationEnabled={
                                                 isNewNavigationEnabled
                                               }
-                                              isNewNavigationEnabledEvaluationPending={
-                                                isNewNavigationEnabledEvaluationPending
+                                              isNewNavigationEnabledEvaluationReady={
+                                                isNewNavigationEnabledEvaluationReady
                                               }
                                             />
                                           </ApplicationContextProvider>
--- a/packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
+++ b/packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
@@ -215,7 +215,10 @@ export const ApplicationShellAuthenticated = (
                   >
                     {/* This is a temporary container. It will be removed once the new navigation is enabled by default. */}
                     <NewMainNavigationFlagWrapper>
-                      {(isNewNavigationEnabled) => {
+                      {({
+                        isNewNavigationEnabled,
+                        isNewNavigationEnabledEvaluationPending,
+                      }) => {
                         return (
                           <>
                             <ThemeSwitcher />
@@ -379,6 +382,9 @@ export const ApplicationShellAuthenticated = (
                                               isLoading={isLoading}
                                               isNewNavigationEnabled={
                                                 isNewNavigationEnabled
+                                              }
+                                              isNewNavigationEnabledEvaluationPending={
+                                                isNewNavigationEnabledEvaluationPending
                                               }
                                             />
                                           </ApplicationContextProvider>
--- a/packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
+++ b/packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
@@ -216,7 +216,7 @@ export const ApplicationShellAuthenticated = (
                   >
                     {/* This is a temporary container. It will be removed once the new navigation is enabled by default. */}
                     <NewMainNavigationFlagWrapper>
-                      {(isNewNavigationEnabled: boolean) => {
+                      {(isNewNavigationEnabled) => {
                         return (
                           <>
                             <ThemeSwitcher />
@@ -352,8 +352,6 @@ export const ApplicationShellAuthenticated = (
                                         isLoading: isLoadingProject,
                                         project,
                                       }) => {
-                                        // Render the loading navbar as long as all the data
-                                        // hasn't been loaded, or if the project does not exist.
                                         const isLoading =
                                           isLoadingUser ||
                                           isLoadingLocaleData ||
@@ -361,13 +359,13 @@ export const ApplicationShellAuthenticated = (
                                           !locale ||
                                           !project;
 
-                                        if (
-                                          isLoading &&
-                                          !isNewNavigationEnabled
-                                        ) {
-                                          return <LoadingNavBar />;
+                                        if (!isNewNavigationEnabled) {
+                                          // Render the loading navbar as long as all the data
+                                          // hasn't been loaded, or if the project does not exist.
+                                          if (isLoading) {
+                                            return <LoadingNavBar />;
+                                          }
                                         }
-
                                         return (
                                           <ApplicationContextProvider
                                             user={user}
@@ -386,8 +384,8 @@ export const ApplicationShellAuthenticated = (
                                               onMenuItemClick={
                                                 props.onMenuItemClick
                                               }
-                                              isLoading={
-                                                // isLoading && // TODO: uncomment before merging
+                                              isLoading={isLoading}
+                                              isNewNavigationEnabled={
                                                 isNewNavigationEnabled
                                               }
                                             />
--- a/packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
+++ b/packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
@@ -358,6 +358,7 @@ export const ApplicationShellAuthenticated = (
                                       }) => {
                                         // Render the loading navbar as long as all the data
                                         // hasn't been loaded, or if the project does not exist.
+                                        // const isLoading = isLoadingUser || isLoadingLocaleData || isLoadingProject || !locale || !project
                                         if (
                                           isLoadingUser ||
                                           isLoadingLocaleData ||
@@ -378,7 +379,7 @@ export const ApplicationShellAuthenticated = (
                                             {isNewNavigationEnabled ===
                                               false && (
                                               <NavBar
-                                                applicationLocale={locale}
+                                                applicationLocale={locale ?? ''}
                                                 projectKey={projectKeyFromUrl}
                                                 project={project}
                                                 environment={
@@ -387,6 +388,9 @@ export const ApplicationShellAuthenticated = (
                                                 onMenuItemClick={
                                                   props.onMenuItemClick
                                                 }
+                                                // isLoading={isLoading}
+                                                // aria-busy={isLoading}
+                                                isLoading
                                               />
                                             )}
                                           </ApplicationContextProvider>
--- a/packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
+++ b/packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
@@ -358,15 +358,19 @@ export const ApplicationShellAuthenticated = (
                                       }) => {
                                         // Render the loading navbar as long as all the data
                                         // hasn't been loaded, or if the project does not exist.
-                                        // const isLoading = isLoadingUser || isLoadingLocaleData || isLoadingProject || !locale || !project
-                                        if (
+                                        const isLoading =
                                           isLoadingUser ||
                                           isLoadingLocaleData ||
                                           isLoadingProject ||
                                           !locale ||
-                                          !project
-                                        )
+                                          !project;
+
+                                        if (
+                                          isLoading &&
+                                          !isNewNavigationEnabled
+                                        ) {
                                           return <LoadingNavBar />;
+                                        }
 
                                         return (
                                           <ApplicationContextProvider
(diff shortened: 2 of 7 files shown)


packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
```diff
@@ -229,12 +229,8 @@ export const ApplicationShellAuthenticated = (
...
                                 height: 100vh;
                                 display: grid;
-                                grid-template-rows: ${isNewNavigationEnabled
-                                  ? `auto ${DIMENSIONS.header} 1fr`
-                                  : `auto ${DIMENSIONS.header} 1fr`};
-                                grid-template-columns: ${isNewNavigationEnabled
-                                  ? 'minmax(80px, 256px) 1fr'
-                                  : 'auto 1fr'};
+                                grid-template-rows: auto ${DIMENSIONS.header} 1fr;
+                                grid-template-columns: auto 1fr;
```
kark, line 233: I reverted those changes as using `minmax()` here was causing buggy behaviour (in my opinion) which can be observed in the screencast in the description👆(beige background was visible when changing size of the viewport in between 80px and 256px wide side panel). 

But it's probable are I don't have the full understanding why it was introduced.

packages/application-shell/src/components/navbar/navbar-skeleton.tsx
```diff
@@ -0,0 +1,163 @@
...
+import { css } from '@emotion/react';
+import styled from '@emotion/styled';
+import { designTokens } from '@commercetools-uikit/design-system';
+
+const NavBarLayout = (
+  props: TNavBarSkeletonProps & { children: ReactNode }
+) => {
+  return (
+    <nav
+      aria-busy={true}
```
kark, line 11: Adding `aria-busy` here will prevent the screen-readers from reading this part until the skeleton is not rendered anymore. That is why I'm also using styled divs instead of more descriptive markup.

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -263,6 +265,10 @@ const NavBar = (props: TNavbarProps) => {
     [props.project]
   );
 
+  if (props.isLoading) {
+    return <NavBarSkeleton isExpanded={isMenuOpen} />;
+  }
```
kark, line 270: Unlike previously, as we have 2 states of the skeleton (expanded and non-expanded), and the new navbar skeleton is rendered inside the navbar

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -263,6 +267,17 @@ const NavBar = (props: TNavbarProps) => {
...
 
+  // TODO: uncomment before merging
+  // if (props.isLoading) {
+  // Render the loading navbar as long as all the data
+  // hasn't been loaded, or if the project does not exist.
+  return props.isNewNavigationEnabled ? (
+    <NavBarSkeleton isExpanded={isMenuOpen} />
+  ) : (
+    <LoadingNavBar />
+  );
```
kark, line 278: I noticed there is some flickering due to the feature flag evaluation. I'm not sure if I'm doing something wrong and what I can do about it, but before evaluating to `true` the flag always defaults to `false`

FilPob:
Thanks @kark  looks super nice. Got some things that can be maybe adjusted straight away (if you feel thats its not related to this PR you can leave it out):

- change the screen width when the navigation jumps from expanded to collapsed to 1200px
- The padding left&right of the expanded menu item should be 12px instead of 16px.
image

- The height of the black header in the navigation should be fixed at 56px (in order to match the height of the App-bar)
image

kark:
> Thanks @kark looks super nice. Got some things that can be maybe adjusted straight away (if you feel thats its not related to this PR you can leave it out):
> 
> * change the screen width when the navigation jumps from expanded to collapsed to 1200px
> * The padding left&right of the expanded menu item should be 12px instead of 16px.
> 
> image
> * The height of the black header in the navigation should be fixed at 56px (in order to match the height of the App-bar)
> 
> image

Hey @FilPob,
Big thanks for the review. I addressed the points and re-deployed the Custom App.
One side note, I assumed that the padding of the menu item both expanded and non-expanded should always be 12px. Hope it's fine.
Screenshot 2023-08-11 at 09 19 41

FilPob:
@kark now the navbar is collapsed by default and I cant expand it :D

kark:
> @kark now the navbar is collapsed by default and I cant expand it :D

Oh, yes that's possible in some cases 😂
You can "unblock" it by going to another app and expanding the menu with the expand button and going back.
Another solution is clearing the local storage.
Unfortunately, we have to leave this change out for now. In order for this logic to work we need to have a working expand button.
And for the sake of testing I am always rendering the skeleton. 
I will re-deploy it in a minute.

.changeset/little-keys-retire.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': minor
+---
+
+Use new NavBar skeleton
```
CarlosCortizasCT, line 5:
```suggestion
Support a new `NavBar` skeleton ui based on a feature flag.
```

packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
CarlosCortizasCT: Maybe @chloe0592 can share something here.

However I don't really get what's the problem. Is it that the change from collapsed to expanded should happen before?
The background color should not be a problem because it will be white. I guess is now beige only for debugging purposes.

packages/application-shell/src/components/navbar/navbar.tsx
CarlosCortizasCT: I think we can use the `useAdapterStatus` [hook](https://github.com/tdeekens/flopflip#useadapterstatus-adapterstatus) from `flopflip` to know about the readiness of the feature flags so we don't render anything until we have the remove value locally.

Probably something like this:
```js
import { useAdapterStatus } from '@flopflip/react-broadcast';

const ComponentWithFeatureToggle = () => {
  const isFeatureEnabled = useFeatureToggle('myFeatureToggle');
  const { isReady } = useAdapterStatus();

  if (!isConfigured) return null;
  else if (!isFeatureEnabled) <NewSkeleton />;
  else return <LegacySkeleton />;
};
```

Credits to @tdeekens for letting me know about this `flopflip` hook 🙇 .

packages/application-shell/src/components/navbar/navbar-skeleton.tsx
```diff
@@ -0,0 +1,184 @@
...
+        height: 100%;
+        width: ${props.isExpanded ? '256px' : '80px'};
+      `}
+    >
+      {props.children}
+    </nav>
+  );
+};
+
+const NavBarHeader = styled.div`
```
CarlosCortizasCT, line 25: **nit**: I would move all the styled components to a `navbar-skeleton.styles.tsx` file so we separate styling from structure and keep files smalls but it's up to you.

packages/application-shell/src/components/navbar/navbar-skeleton.tsx
CarlosCortizasCT: Awesome! 💯

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -213,13 +215,15 @@ const ApplicationMenu = (props: ApplicationMenuProps) => {
 ApplicationMenu.displayName = 'ApplicationMenu';
 
 type TNavbarProps = {
-  applicationLocale: string;
+  applicationLocale?: string;
```
CarlosCortizasCT, line 218: Just wondering why is not this property optional when it was required 🤔

packages/application-shell/src/components/application-shell-authenticated/new-main-navigation-flag-wrapper.tsx
```diff
@@ -2,7 +2,9 @@
 import { useFeatureToggle } from '@flopflip/react-broadcast';
 import { MAIN_NAVIGATION } from '../../feature-toggles';
 
-const NewMainNavigationFlagWrapper = (props: { children: Function }) => {
+const NewMainNavigationFlagWrapper = (props: {
```
CarlosCortizasCT, line 5: Maybe we can add a comment here that this component is intended to be temporary and we will remove it once the new navigation has been rolled out.

packages/application-shell/src/components/navbar/navbar-skeleton.tsx
```diff
@@ -0,0 +1,184 @@
...
+  return (
+    <MenuItemContainer
+      placement={props.placement}
+      isExpanded={props.isExpanded}
+    >
+      <MenuItemIcon />
+      {props.isExpanded && <MenuItemTitle placement={props.placement} />}
+    </MenuItemContainer>
+  );
+};
```
CarlosCortizasCT, line 146: I'm sorry Kacper but I'm struggling to understand the `placement` attribute purpose. 

Could you please explain a little what it is used for? 🙏

FilPob:
@kark cool! looks all good from my side 👏

FilPob (approved):

packages/application-shell/src/components/navbar/navbar.tsx
tdeekens: You're welcome. It's just

```diff
-if (!isConfigured) return null;
+if (!isReady) return null;
```

kark: Wowzers, TIL, thank you both for sharing 🎉

packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
kark: > However I don't really get what's the problem.

My point is that right now we are using 2 widths for the navbar (either 80px or 256px) and the side panel's width is `auto`. And I think that works fine. 

Using minmax() makes the side panel flexible in that range, but navbar doesn't fill up that space. So the white space seems to be redundant.

chloe0592: @kark feel free to iterate these stylings as you need. The changes were indeed made only for the purpose of having a 'visible' container where the new navigation will be placed.

packages/application-shell/src/components/application-shell-authenticated/new-main-navigation-flag-wrapper.tsx
kark: It seems this comment has already been added by @chloe0592 https://github.com/commercetools/merchant-center-application-kit/blob/2cb7e8227d7406a2c78addd78181f0f3dac66e30/packages/application-shell/src/components/application-shell-authenticated/new-main-navigation-flag-wrapper.tsx#L1

packages/application-shell/src/components/navbar/navbar.tsx
kark: Thanks, that's a great point. Now the possibility of `undefined` is evaluated inside `NavBar`.
Type checking is extremely verbose 🙁:
https://github.com/commercetools/merchant-center-application-kit/blob/3004ea9b3d5441544f7f4337fbb5aadb32fb4c1d/packages/application-shell/src/components/navbar/navbar.tsx#L271-L284
but that's the way I could think of to narrow down the type without going for `applicationLocale={locale ?? ""}` in here:
https://github.com/commercetools/merchant-center-application-kit/blob/3004ea9b3d5441544f7f4337fbb5aadb32fb4c1d/packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx#L373
which could possibly change the behaviour (I think).

packages/application-shell/src/components/navbar/navbar-skeleton.tsx
kark: Thanks, I had hard time with this one TBH 😅
I don't think `<MenuItemContainer>` and `<MenuItemTitle>` components should be aware of the context where they're rendered.
But it also made sense to me to reuse them as part of `MenuItem` component since they're almost identical in header and body.
|  header |  body |  
|---|---|
|  Screenshot 2023-08-11 at 15 18 00 |  Screenshot 2023-08-11 at 15 18 15 |

I couldn't think of any better name for the prop (`theme` doesn't really work I think) so I went for the naming borrowed from our [`<Tooltip>`](https://github.com/commercetools/ui-kit/blob/5f2886d1f29234c28c6201922d89ed7d14d5d17e/packages/components/tooltip/src/tooltip.tsx#L81-L93).

As an alternative we could have just separate styled components for both cases if that works better for you.

packages/application-shell/src/components/application-shell-authenticated/new-main-navigation-flag-wrapper.tsx
```diff
@@ -1,10 +1,25 @@
...
+
+const NewMainNavigationFlagWrapper = (props: {
+  children: ({
+    isNewNavigationEnabled,
+    isNewNavigationEnabledEvaluationPending,
+  }: TNewMainNavigationFlagWrapperChildrenProps) => JSX.Element;
+}) => {
   const isNewNavigationEnabled = useFeatureToggle(MAIN_NAVIGATION);
-  return props.children(isNewNavigationEnabled);
+  const { isConfiguring } = useAdapterStatus();
```
tdeekens, line 17: I think you want `isReady` here. Ready state means the flags have been fetched. Configuration happens before.

kark: Thanks, I am 100% sure you are right, I need to have a proper look into it. I won't bother you before doing that.
Although it seems straightforward I simply couldn't make it work this way for now.

Just for some context:
[Here](https://github.com/commercetools/merchant-center-application-kit/compare/kk-navbar-skeleton..kk-navbar-skeleton-ff-evaluation) I pushed what supposedly should work.
But even thought the value of the flag is fetched, the hook's value of [adapterStatus](https://github.com/tdeekens/flopflip/blob/fd66f538640da02bdb6a3479a7a73135106cf142/packages/react-broadcast/src/hooks/use-adapter-status/use-adapter-status.ts#L14) strangely doesn't change.
The request:
Screenshot 2023-08-11 at 16 18 51
The adapter's state is always `false`:
Screenshot 2023-08-11 at 16 27 29

tdeekens: Ok that's odd. I can see the same in the MC FE. Not sure if that's a bug in flopflip. Open to PRs there!

CleanShot 2023-08-11 at 16 57 55@2x

Whatever works use it ;)

kark: > **Note:** a temporary workaround is used for computing the combined adapter status

https://github.com/commercetools/merchant-center-application-kit/blob/5eea6d29db452ecfa89994bcc47d8130e3c52515/packages/application-shell/src/components/application-shell-authenticated/new-main-navigation-flag-wrapper.tsx#L17-L20

tdeekens: Let's see if this works. I can confirm that the combine-adapters doesn't combine the adapter status of the underlying adapters. We combine the launch-darkly and http-adapter to load flags from LD and the MC API. 

I can think of a way how to combine the adapter status without being ambiguous. Back then I opted out to wait until it would be needed.

tdeekens (approved):

packages/application-shell/src/components/application-shell-authenticated/new-main-navigation-flag-wrapper.tsx
CarlosCortizasCT: Oops! Didn't see the comment in the PR.

Thanks!

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -263,6 +268,21 @@ const NavBar = (props: TNavbarProps) => {
...
+    // hasn't been loaded, or if the project does not exist.
+    if (!props.isNewNavigationEnabledEvaluationReady) {
+      return null;
+    }
+    return props.isNewNavigationEnabled ? (
+      <NavBarSkeleton isExpanded={isMenuOpen} />
+    ) : (
+      <LoadingNavBar />
+    );
+  }
```
CarlosCortizasCT, line 284:
```suggestion
  // Render the loading navbar as long as all the data
  // hasn't been loaded, or if the project does not exist.
  if (props.isLoading || typeof applicationLocale === 'undefined') {
    // Do not render anything until we have the feature flag remote value fetched so 
    // we avoid a flick effect in case it is 'true' (different value from the default one)
    if (!props.isNewNavigationEnabledEvaluationReady) {
      return null;
    }
    return props.isNewNavigationEnabled ? (
      <NavBarSkeleton isExpanded={isMenuOpen} />
    ) : (
      <LoadingNavBar />
    );
  }
```

packages/application-shell/src/components/navbar/navbar-skeleton.tsx
CarlosCortizasCT: I feel both visualizations are quite similar. I think the only main difference is the width of the placeholder.

I created [this PR](https://github.com/commercetools/merchant-center-application-kit/pull/3166/) to suggest what I believe is a little more simpler approach.

Let me know what you think.

packages/application-shell/src/components/navbar/navbar.tsx
chloe0592: Thanks @CarlosCortizasCT! The change is here - [4f94eaa](https://github.com/commercetools/merchant-center-application-kit/pull/3165/commits/4f94eaac64d6f5fc4374bbc675e4d60e02316cc6).

packages/application-shell/src/components/navbar/navbar-skeleton.tsx
chloe0592: Thanks a lot for the suggestion! Merged here - [0c844b2](https://github.com/commercetools/merchant-center-application-kit/pull/3165/commits/0c844b2560208d7606f8fdd93fe177869404b6e7).

chloe0592 (approved):
Thanks @kark for all the work! 🙏

packages/application-shell/src/components/application-shell-authenticated/new-main-navigation-flag-wrapper.tsx
```diff
@@ -2,9 +2,27 @@
...
+
+const NewMainNavigationFlagWrapper = (props: {
+  children: ({
+    isNewNavigationEnabled,
+    isNewNavigationEnabledEvaluationReady,
+  }: TNewMainNavigationFlagWrapperChildrenProps) => JSX.Element;
+}) => {
   const isNewNavigationEnabled = useFeatureToggle(MAIN_NAVIGATION);
-  return props.children(isNewNavigationEnabled);
+  // A temporary workaround for a glitch in computing the aggregated adapter status within useAdapterStatus
```
CarlosCortizasCT, line 17: **nit**: So it will easier for us to find pending things to do in the future
```suggestion
  // TODO: A temporary workaround for a glitch in computing the aggregated adapter status within useAdapterStatus
```

CarlosCortizasCT (approved):
Looks good to me 👍 

Great work!!! 💯

PR #3167, opened 2023-08-16
Adjust Google Fonts loader url

CarlosCortizasCT:
#### Summary

Adjust Google Fonts loader url.

#### Description

We've noticed some issues in the text styles of all Merchant Center applications that seems related to how fonts are loaded in the document.
We load fonts using [Google Fonts API v2](https://developers.google.com/fonts/docs/css2?hl=en).

It seems using the `slant` attribute somehow it's making fonts to fail loading the right ones based on the text styles so we're removing it from the loader URL. We didn't notice any downsides from removing this parameter.

Also, we are taking the opportunity to remove `Open Sans` from the loader URL since we're not using it anymore.

### DIFF
--- a/.changeset/lazy-jokes-train.md
+++ b/.changeset/lazy-jokes-train.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/mc-html-template': patch
+---
+
+Adjust Google Fonts loader URL. Removed "Open Sans" since we're not using it anymore.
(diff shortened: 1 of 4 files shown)


Rhotimee (approved):
Thanks @CarlosCortizasCT 🙌🏽

ddouglasz (approved):
Thank you 🙌🏾

emmenko (approved):

chloe0592 (approved):
Thank you!

PR #3169, opened 2023-08-18
Scrollable navigation

chloe0592:
Adding the ability to scroll through the navigation, both in expanded and collapsed states, while submenu is showing on the side. The changes are behind the feature flag and in the project `Main navigation test`.

There will be more changes necessary for each of those states and also adapting the styling according to new design, but the idea of this PR is creating a base to move forward.

Short visual:
(first part is current state, the second part is with changes)

https://github.com/commercetools/merchant-center-application-kit/assets/52276952/ceb4eb0d-22fe-4bdf-9d99-189da681963d


CarlosCortizasCT:
Thanks for working on this issue Klaudija 👍 

I'm not sure about the scope of this PR so maybe some of my findings are supposed to be addressed later, but let me list them here so we all are aware of them:
* In expanded mode, the navbar footer is sticky but this does not happen in collapsed mode
* Submenu links are only shown is expanded mode upon clicking (not when hovering)
* When clicking different navbar items, we can see two of them selected (are we actually going to require click to open submenu links or we will only support hover)
* In collapsed mode, there's a weird white space above the support link
* In expanded mode, when clicking on a navbar item that requires scrolling down, its submenu links panel is not positioned correctly

[App-kit-playground3 - main-navigation-test - Merchant Center.webm](https://github.com/commercetools/merchant-center-application-kit/assets/97907068/afa4fbc7-580f-4347-b0aa-37d32db8ad5a)

chloe0592:
Note: to mock multiple applications for testing purpuses you can add this block of code [here](https://github.com/commercetools/merchant-center-application-kit/blob/74afe38d1662a7208815c7a3d592ddf8f0f3eba5/packages/application-shell/src/components/navbar/use-navbar-state-manager.ts#L233).

But bare in mind that links won't work in that case.

```
 const allInternalApplicationsNavbarMenu = [];

  if (applicationsNavBarMenu) {
    for (let i = 10; i > 0; i--) {
      allInternalApplicationsNavbarMenu.push(
        {
          ...applicationsNavBarMenu[0],
          key: applicationsNavBarMenu[0].key + i,
          uriPath: applicationsNavBarMenu[0].uriPath + i,
          submenu: applicationsNavBarMenu[0].submenu.map((submenu) => {
            return {...submenu,
              key: submenu.key + i,
              uriPath: submenu.uriPath + i,
            }
          })
        }
      )
    }
  }
```

chloe0592:
Thank you so much for this valuable feedback @CarlosCortizasCT! 🍫 🍫 

So I'll try explain my idea one by one:

1. In expanded mode, the navbar footer is sticky but this does not happen in collapsed mode
 - This will be resolved when we'll work on the footer task.

2. Submenu links are only shown is expanded mode upon clicking (not when hovering).
 - This might be looked into when we'll work on the submenu and menu redesign task.

3. When clicking different navbar items, we can see two of them selected (are we actually going to require click to open submenu links or we will only support hover).
 - I believe we will only support hover in both states, but @FilPob could confirm this? 🙏 
 - In any case we could fix that when working on the submenu and menu redesign task.

4. In collapsed mode, there's a weird white space above the support link.
 - Will be addressed when we'll work on the redesign of menu task (when all other spacings will be adjusted).

5. In expanded mode, when clicking on a navbar item that requires scrolling down, its submenu links panel is not positioned correctly
- Great catch! That's something I'll be looking into now.

chloe0592:
@CarlosCortizasCT here's the fix for the submenu misplaced when hovered - [19e14e6](https://github.com/commercetools/merchant-center-application-kit/pull/3169/commits/19e14e6b4f6965f6e06a821536024df870ea93a8)

https://github.com/commercetools/merchant-center-application-kit/assets/52276952/b07dbd8b-4043-4766-952f-cc5b2e89f516

CarlosCortizasCT:
After the latest updates, I noticed a couple of things I'd like to mention:
* We don't have an independent scroll in the navbar when using the collapsed state
* If we want to se the submenu links for a navbar item at the bottom of the screen, the popup panel will not be fully visible. In this situation, I think we should position the panel upwards instead of downwards the navbar item position.
* 
[App-kit-playground - almond-40 - Merchant Center.webm](https://github.com/commercetools/merchant-center-application-kit/assets/97907068/e11d2640-891c-4294-b9a7-7d33b35e55fe)

chloe0592:
> Thanks for your explanations @chloe0592 🙇
> 
> Now the only other thing I could spot is the clickable area for the last item in the navbar: it is smaller than the rest, maybe because of the gradient we're using on the footer.
> 
> Is this something to be addressed in the upcoming footer task?
> 
>  App-kit-playground.-.main-navigation-test.-.Merchant.Center.webm

Eagle eyes! 🦅 Great catch and yes exactly, I'll be fixing that in the footer task.

packages/application-shell/src/components/navbar/menu-items.tsx
```diff
@@ -191,6 +194,12 @@ const MenuGroup = (props: MenuGroupProps) => {
     props.level === 2 && props.isActive && !props.isExpanded;
   return (
     <ul
+      style={{
+        top: props.isNewNavigationEnabled
+          ? // calculating the submenu position based on the parent's top position and the scroll position
+            `calc(${props.topPosition}px - ${props.scrollTop}px)`
+          : undefined,
+      }}
```
kark, line 202: nit: perhaps we could use emotion's css here
```suggestion
css={
        props.isNewNavigationEnabled &&
        css`
          // calculating the submenu position based on the parent's top position and the scroll position
          top: calc(${props.topPosition}px - ${props.scrollTop}px);
        `
      }
```

ddouglasz (approved):
Tested this  🙌🏾 
 Code looks good! one of my observations was around the submenu display when `clicked` on expand and when menubar is collapsed its display when `hovered` . I see you already talked about handling it in follow-up work. 
Thank you  💯

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -79,6 +88,15 @@ const getMenuVisibilityOfMainmenu = (menu: TNavbarMenu) =>
...
 
 const ApplicationMenu = (props: ApplicationMenuProps) => {
+  const [topPosition, setTopPosition] = useState(0);
+  const elementRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    if (elementRef.current != null) {
+      setTopPosition(elementRef.current.offsetTop);
+    }
+  }, [elementRef.current]);
```
kark, line 98: I don't think this ref ever changes so I'm not sure about adding that to the deps array.

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -79,6 +88,15 @@ const getMenuVisibilityOfMainmenu = (menu: TNavbarMenu) =>
   menu.menuVisibility ? [menu.menuVisibility] : [];
 
 const ApplicationMenu = (props: ApplicationMenuProps) => {
+  const [topPosition, setTopPosition] = useState(0);
+  const elementRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    if (elementRef.current != null) {
+      setTopPosition(elementRef.current.offsetTop);
```
kark, line 96: I guess it's ok for the first iteration, I've seen some corner cases
Screenshot 2023-08-23 at 12 12 22

kark (commented):
Thank you for working on this. I think it looks good 🙌 I don't want to block any progress.
However, I noticed some behaviour with scrollbar being visible when navbar is first rendered in expanded mode and I'm not sure if that's expected

chloe0592:
> Thank you for working on this. I think it looks good 🙌 I don't want to block any progress. However, I noticed some behaviour with scrollbar being visible when navbar is first rendered in expanded mode and I'm not sure if that's expected 

@kark thanks so much for all the valuable feedback! 🙇‍♀️ 
The visible scrollbar is a great catch. I think we shouldn’t be seeing it at all. Here's the fix - [7749c6a].(https://github.com/commercetools/merchant-center-application-kit/pull/3169/commits/7749c6a86d0b66a803f8a7981aaa2daa5a19d11c)
And hopefully you won't see it anymore.


https://github.com/commercetools/merchant-center-application-kit/assets/52276952/fa822c0a-625a-4909-9d79-1a0e93c00d7d

packages/application-shell/src/components/navbar/navbar.tsx
chloe0592: Yes, good point. It's always the same so no need for re-runs. Thanks! [c02833c](https://github.com/commercetools/merchant-center-application-kit/pull/3169/commits/c02833c4529801c0282dd2366d22e5dd99f2840e)

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -166,6 +185,9 @@ const ApplicationMenu = (props: ApplicationMenuProps) => {
           isActive={props.isActive}
           isExpanded={props.isMenuOpen}
           hasSubmenu={hasSubmenu}
+          topPosition={topPosition}
+          scrollTop={props.scrollTop}
```
CarlosCortizasCT, line 189: **nit**: I personally feel this a little weird. I would expect just a `topPosition` prop where we indicate to the `MenuItem` the vertical position it should use: we can then pass the calculation (`topPosition + props.scrollTop`) as the value.
I don't feel the `MenuItem` component needs to know both parameters as it only cares about what vertical position it should use.

What do you think?

CarlosCortizasCT:
@chloe0592 please bear in mind we now support keyboard navigation. You can use the `tab` key to navigate to nav bar main items and use `enter` key to display the submenu links.

packages/application-shell/src/components/navbar/navbar.tsx
kark: Sure,

It's when the navbar is being scrolled and menu item is half-visible.
However, I believe this isn't a matter of concern for this PR.

kark:
One more issue I noticed is that navbar width goes to 0 when viewport is narrow

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -245,6 +267,13 @@ const NavBar = (props: TNavbarProps) => {
   );
   const location = useLocation();
 
+  const [scrollTop, setScrollTop] = useState(0);
+
+  // we need this scroll position to set the correct height of the submenu
+  const handleScroll = (event: SyntheticEvent) => {
```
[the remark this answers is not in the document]
chloe0592: @CarlosCortizasCT thanks for pointing that out! 
Done here - [727b1f1](https://github.com/commercetools/merchant-center-application-kit/pull/3169/commits/727b1f18130464efa1edc4bab3623c96f7bd4a43).

emmenko:
The scrolling functionality looks great from what I could see 🙌

I noticed one other thing (maybe not related to this PR). If menu items don't have submenus, we show an empty submenu container...not sure what the designs for it would be.

image

packages/application-shell/src/components/navbar/navbar.tsx
chloe0592: @CarlosCortizasCT yes, now that you've mentioned, I don’t know why I didn’t pass it this way in the first place. 😄 
[5b8e33d](https://github.com/commercetools/merchant-center-application-kit/pull/3169/commits/5b8e33dffa048ac4de7bfa9c1f0dac334d25223f)

chloe0592:
> The scrolling functionality looks great from what I could see 🙌
> 
> I noticed one other thing (maybe not related to this PR). If menu items don't have submenus, we show an empty submenu container...not sure what the designs for it would be.
> 
> image

Oh yes, that's a very good point... Thank you @emmenko!
I have no idea what we want to show or how it should behave in case there's no submenu. 

@FilPob, could you help us clarify that please? 🙏

emmenko:
Ok thanks for checking. Let's look at it separately, as it's not related to the scrolling functionality.

packages/application-shell/src/components/navbar/navbar.mod.css
```diff
@@ -381,6 +403,22 @@
...
 
+.item-link:hover .sublist-collapsed-new__active,
+.item-link:hover .sublist-expanded-new__active {
+  display: block;
+  text-align: left;
+  /* stylelint-disable-next-line csstools/value-no-unknown-custom-properties */
+  background-color: var(--background-color-for-navbar-when-active);
+  min-height: 50px;
+  width: 253px;
+  border-radius: 8px;
```
emmenko, line 414: Would be nice to not hardcode these values, maybe using design tokens variables. You can also assign some variables in the `:root` object at the top of this file.

emmenko (commented):
Code changes look good to me 👌

chloe0592:
> One more issue I noticed is that navbar's width goes to 0 when viewport is narrow

@kark thanks for noticing! I updated it here [975b20f](https://github.com/commercetools/merchant-center-application-kit/pull/3169/commits/975b20fe520185fb746ce9cd84d149dc480d7133). 
When we'll be working on the navigation styling the new constants will be applied.

packages/application-shell/src/components/navbar/navbar.mod.css
chloe0592: Yes, true. Thanks! That will be adjusted in the following tasks when styling the submenu and navigation.
But I added the comment just in case. [0a4b394](https://github.com/commercetools/merchant-center-application-kit/pull/3169/commits/0a4b39490b33631f00fe74eacef252806f040bca)

FilPob:
> > The scrolling functionality looks great from what I could see 🙌
> > I noticed one other thing (maybe not related to this PR). If menu items don't have submenus, we show an empty submenu container...not sure what the designs for it would be.
> > image
> 
> Oh yes, that's a very good point... Thank you @emmenko! I have no idea what we want to show or how it should behave in case there's no submenu.
> 
> @FilPob, could you help us clarify that please? 🙏

When no sublinks are present, a tooltip is supposed to show with the application name. See design [HERE](https://www.figma.com/file/gC6Yz03ThScI6Mk23cjKDt/Main-Navi2023?type=design&node-id=755%3A21690&mode=design&t=0WtDbdeo5etqcaPn-1)

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -1,5 +1,13 @@
...
+import {
+  MouseEventHandler,
+  useEffect,
+  useMemo,
+  useState,
+  useRef,
+  SyntheticEvent,
+} from 'react';
 import classnames from 'classnames';
+import { debounce } from 'lodash';
```
Rhotimee, line 10:
```suggestion
import debounce from 'lodash/debounce';
```

Rhotimee, line 10:
```suggestion
import debounce from 'lodash/debounce';
```

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -1,5 +1,13 @@
-import { MouseEventHandler, useEffect, useMemo } from 'react';
+import {
+  MouseEventHandler,
+  useEffect,
+  useMemo,
+  useState,
+  useRef,
+  SyntheticEvent,
+} from 'react';
```
Rhotimee, line 8:
```suggestion
import {
  type MouseEventHandler,
  useEffect,
  useMemo,
  useState,
  useRef,
  type SyntheticEvent,
} from 'react';
```

Rhotimee (commented):
Nice work 💪🏽

FilPob:
@chloe0592 I got one more thing that I noticed in the gif that is missing, the collapse button + support link) are supposed to be sticky at the bottom. specs **[here](https://www.figma.com/file/gC6Yz03ThScI6Mk23cjKDt/Main-Navi2023?type=design&node-id=322%3A17612&mode=design&t=G3AoyYeTjv4YTKnr-1)**

chloe0592:
> here

Thanks @FilPob! Yes, that's done in the Footer ticket. I've already fixed it and it'll be another PR.

emmenko (approved):
Nice work! 🤗

emmenko:
PS: don't forget a changeset

packages/application-shell/src/components/navbar/navbar.tsx
chloe0592: Thank you @Rhotimee for the suggestions! Done here - 63e83369db138986fa5abe2e7b361dbc5bf3018d.

kark (approved):
🎉

Rhotimee (approved):
🙌🏽

PR #3172, opened 2023-08-21
refactor(application-shell): navigation to use useAdapterStatus

tdeekens:
Follow up on https://github.com/commercetools/merchant-center-application-kit/pull/3165/files

#### Summary

This refactors the navigation to use the `useAdapterStatus` hook to evaluate the readiness of flopflip and the underlying LaunchDarkly adapter.

#### Description

`flopflip` and the underlying state of combined adapters (if e.g. one uses `http` and the `launchdarkly` adapter) wasn't kept separately. This means `flopflip` just kept the state of the combined adapters which resolved too early to `Configured`. Instead it should keep the state of each adapter used even when combined in a _backwards compatible way_. 

This means the state inside `flopflip` now is:

```js
status: {
   http: {
       configurationStatus: 2
   },
   launchdarkly: {
       configurationStatus: 2
   },
   combined: {
       configurationStatus: 2
   }
}
```

We can now use this split out state with the `useAdatperStatus` hook as expected. When using it without an argument like `const { isReady } = useAdapterStatus()` then `isReady` is only `true` if all adapters are ready (or once they become).

We can also use this to pluck the state of an individual adapter as in `const { isReady } = useAdapterStatus({ adapterIdentifiers: ['http'] })` which resolves to `{ isReady: true }` already when the `http-adapter` indicates it. Note that we pass in array so we can also pluck down the state of multiple adapters. 

This all needed quite a bit of work inside of `flopflip` which you can ready up on:

1. https://github.com/tdeekens/flopflip/pull/1790/files
2. https://github.com/tdeekens/flopflip/pull/1791/files
3. https://github.com/tdeekens/flopflip/pull/1797/files


kark (approved):
Thank you, works great 🎉

packages/application-shell/src/components/application-shell-authenticated/new-main-navigation-flag-wrapper.tsx
```diff
@@ -14,10 +15,9 @@ const NewMainNavigationFlagWrapper = (props: {
   }: TNewMainNavigationFlagWrapperChildrenProps) => JSX.Element;
 }) => {
   const isNewNavigationEnabled = useFeatureToggle(MAIN_NAVIGATION);
```
chloe0592, line 17: @tdeekens I have one question. Now the value of `isNewNavigationEnabled` flag is always false. And also there's no network call to Darkly like before the changes.

But maybe I'm missing something perhaps? 
The same is currently on main before since [your last](https://github.com/commercetools/merchant-center-application-kit/pull/3001) changes(I think).

chloe0592: The network call before the changes ⬇️ 
Screenshot 2023-08-21 at 11 09 24

chloe0592 (approved):
Thank you! 🍫

packages/application-shell/src/components/application-shell-authenticated/new-main-navigation-flag-wrapper.tsx
```diff
@@ -14,10 +15,9 @@ const NewMainNavigationFlagWrapper = (props: {
...
-  // TODO: A temporary workaround for a glitch in computing the aggregated adapter status within useAdapterStatus
-  const isReady =
-    // @ts-ignore
-    window?.__flopflip__?.launchdarkly?.getIsConfigurationStatus?.(2);
+  const { isReady } = useAdapterStatus({
+    adapterIdentifiers: [adapterIdentifiers.launchdarkly],
```
CarlosCortizasCT, line 19: Out of curiosity, what's the difference between `adapterIdentifiers.launchdarkly` and `adapterIdentifiers.http` ([reference](https://github.com/tdeekens/flopflip/blob/cb0928d237503cfc95d4954cb96e996c8cace286/packages/types/src/index.ts#L132))?

I would expect `http` to represent any adapter that requires HTTP calls to fetch its state. 🤔

PR #3175, opened 2023-08-22
Support Custom Views local development

CarlosCortizasCT:
#### Summary

This PR focuses on creating a new starter template to bootstrap a Custom View and to allow running it locally.

#### Description

Most of the change files are actually the new files for the new template, which was created based on the Custom Applications typescript starter template.
The main two differences are:
* It has a different configuration file (named `custom-view-config`)
* It does not use the `ApplicationShell` component in its `entry-point` but new components tailored to run the Custom View locally:
  * `CustomViewShell`: new main wrapper for Custom Views
  * `CustomViewDevHost`: new component to emulate a local host application where we can run the Custom View from

Other relevant changes are:
* Custom Views have their config file with its validation, so the process for loading, processing and validating the configuration has been updated. We just updated the current `application-config` package instead of creating a new one for Custom Views.
* There's a new component which abstracts away the logic to run a Custom View locally: `CustomDevHost`.
* We also have new helpers dedicated to Custom View tests.


packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -92,7 +93,7 @@ function CustomViewLoader(props: TCustomViewLoaderProps) {
         context: {
           dataLocale: appContext.dataLocale,
           customViewConfig: props.customView,
-          hostUrl: window.location.href,
+          hostUrl: props.hostUrl || window.location.href,
```
CarlosCortizasCT, line 96: This allows for mocking the URL in the local custom-view development template

packages/application-config/custom-view.schema.json
```diff
@@ -0,0 +1,206 @@
+{
```
CarlosCortizasCT, line 1: Custom Views have a slightly different configuration requirements than Custom Apps, so that's why we renamed the previous schema and created this one.

packages/application-config/package.json
```diff
@@ -31,7 +31,9 @@
     "README.md"
   ],
   "scripts": {
-    "build:schema": "json2ts schema.json src/schema.ts --style.singleQuote --bannerComment '/* eslint-disable prettier/prettier */\n// This file was automatically generated by json-schema-to-typescript.\n// DO NOT MODIFY IT BY HAND. Instead, modify the source schema.json file.'"
+    "build:customApplicationSchema": "json2ts custom-application.schema.json src/custom-application.schema.ts --style.singleQuote --bannerComment '/* eslint-disable prettier/prettier */\n// This file was automatically generated by json-schema-to-typescript.\n// DO NOT MODIFY IT BY HAND. Instead, modify the source custom-application.schema.json file and run the build:schemas npm script.'",
+    "build:customViewSchema": "json2ts custom-view.schema.json src/custom-view.schema.ts --style.singleQuote --bannerComment '/* eslint-disable prettier/prettier */\n// This file was automatically generated by json-schema-to-typescript.\n// DO NOT MODIFY IT BY HAND. Instead, modify the source custom-view.schema.json file and run the build:schemas npm script.'",
+    "build:schemas": "pnpm build:customApplicationSchema && pnpm build:customViewSchema"
```
CarlosCortizasCT, line 36: We now have to manage two schemas for configuration files (Custom Applications and Custom Views).

packages/application-config/src/load-config.ts
```diff
@@ -37,52 +37,97 @@ const loadJsModule: LoaderSync = (filePath) => {
...
+      '.js': loadJsModule,
+      '.cjs': loadJsModule,
+      '.mjs': loadJsModule,
+      '.ts': loadJsModule,
+    },
+  });
+};
+
+const customApplicationExplorer = getExplorer('custom-application-config');
+const customViewExplorer = getExplorer('custom-view-config');
```
CarlosCortizasCT, line 67: We need to be able to load config files for both Custom Applications and Custom Views

packages/application-config/src/load-config.ts
```diff
@@ -37,52 +37,97 @@ const loadJsModule: LoaderSync = (filePath) => {
...
   applicationPath: string
-): JSONSchemaForCustomApplicationConfigurationFiles => {
-  const configFile = explorer.search(applicationPath);
+): {
+  filepath: string;
+  config: JSONSchemaForCustomApplicationConfigurationFiles;
+} => {
+  const customApplicationConfigFile =
+    customApplicationExplorer.search(applicationPath);
+  const customViewConfigFile = customViewExplorer.search(applicationPath);
```
CarlosCortizasCT, line 120: We try to load either config file for Custom Apps and Custom Views as only one of them will be found.
You won't have both in your project.

packages/application-config/src/process-config.ts
```diff
@@ -28,17 +32,23 @@ type ProcessConfigOptions = Partial<LoadingConfigOptions> & {
 const developmentPort = 3001;
 const developmentAppUrl = `http://localhost:${developmentPort}`;
 
+const getConfigurationType = (configFileName: string): ConfigType => {
```
CarlosCortizasCT, line 35: As we will need to process some things differently, we need to distinguish between Custom Apps and Custom Views configuration files. We do it by inspecting the config file name as they are predefined.

packages/application-config/src/load-config.ts
```diff
@@ -37,52 +37,97 @@ const loadJsModule: LoaderSync = (filePath) => {
...
+    (!customApplicationConfigFile || !customApplicationConfigFile.config) &&
+    (!customViewConfigFile || !customViewConfigFile.config)
+  ) {
     throw new MissingOrInvalidConfigError(
-      `Missing or invalid Custom Application configuration file.`
+      `Missing or invalid configuration file.`
     );
   }
-  return configFile.config;
+  return (customApplicationConfigFile || customViewConfigFile)!;
```
CarlosCortizasCT, line 130: We now return not only the configuration but also the filename so later we can know which one we processed.

packages/application-config/src/process-config.ts
```diff
@@ -53,16 +63,19 @@ const processConfig = ({
...
+  validateConfig(configType, rawConfig);
   const appConfig =
     substituteVariablePlaceholders<JSONSchemaForCustomApplicationConfigurationFiles>(
       rawConfig,
       { applicationPath, processEnv }
     );
-  const customApplicationData =
-    transformCustomApplicationConfigToData(appConfig);
+
+  const configurationData = transformConfigurationToData(configType, appConfig);
```
CarlosCortizasCT, line 78: We provide the config type to the transform function so it know which output we expect (Custom Apps or Custom Views)

packages/application-config/src/process-config.ts
```diff
@@ -104,8 +117,14 @@ const processConfig = ({
   // This is important to determine to which URL the MC should redirect to
   // after successful login.
   const applicationId = isProd
-    ? `${customApplicationData.id}:${customApplicationData.entryPointUriPath}`
-    : `__local:${customApplicationData.entryPointUriPath}`;
+    ? `${configurationData.id}:${
+        (configurationData as CustomApplicationData).entryPointUriPath ||
+        'custom-view'
```
CarlosCortizasCT, line 122: Custom Views does not have an entry point, so we use a placeholder to keep using the same application identifier value pattern.

packages/application-config/src/process-config.ts
```diff
@@ -104,8 +117,14 @@ const processConfig = ({
...
+      }`
+    : `__local:${
+        (configurationData as CustomApplicationData).entryPointUriPath ||
+        configurationData.id
```
CarlosCortizasCT, line 126: For Custom Views, we use their ID since they don't have an entry point.

packages/application-config/src/transformers.ts
```diff
@@ -44,7 +51,10 @@ const getPermissions = (
     appConfig.additionalOAuthScopes?.map(({ name }) => name) || [];
 
   const permissionKeys = entryPointUriPathToResourceAccesses(
-    appConfig.entryPointUriPath,
+    (appConfig as JSONSchemaForCustomApplicationConfigurationFiles)
+      .entryPointUriPath ||
+      (appConfig as JSONSchemaForCustomViewConfigurationFiles).env.production
```
CarlosCortizasCT, line 56: Permissions are calculated based on ID for Custom Views, as they don't have an entry point.
For a Custom View with the ID `123456`, the permissions keys would be `view123456` and `manage123456`.

packages/application-config/src/validations.ts
```diff
@@ -42,11 +51,22 @@ const printErrors = (errors?: ErrorObject[] | null) => {
...
 ): void => {
-  const valid = validate(config);
-  if (!valid) {
-    throw new Error(printErrors(validate.errors));
+  let validation;
+
+  if (configType === ConfigType.CUSTOM_APPLICATION) {
+    validation = validateCustomApplicationConfig;
+  } else if (configType === ConfigType.CUSTOM_VIEW) {
+    validation = validateCustomViewConfig;
```
CarlosCortizasCT, line 62: We need to validate differently the config file depending if it's for a Custom Application or a Custom View.

packages/application-config/test/process-config.spec.js
```diff
@@ -26,7 +26,10 @@ beforeEach(() => {
 
 describe('processing a simple config', () => {
   beforeEach(() => {
-    loadConfig.mockReturnValue(fixtureConfigSimple);
+    loadConfig.mockReturnValue({
```
CarlosCortizasCT, line 29: This function now returns not only the configuration but also the filename, so we need to update all these mocks.

packages/application-shell/src/components/authenticated/authenticated.tsx
```diff
@@ -52,6 +52,14 @@ const AuthenticationRoutes = (props: TAuthenticatedProps) => (
         applicationMessages={props.applicationMessages}
       />
     </SuspendedRoute>
+    <SuspendedRoute
```
CarlosCortizasCT, line 55: We use this new route in Custom Views local auth flow.

packages/application-shell/src/components/custom-view-dev-host/custom-view-dev-host.tsx
```diff
@@ -0,0 +1,148 @@
...
+  );
+};
+
+type TCustomViewDevHost = {
+  environment: TApplicationContext<{}>['environment'];
+  applicationMessages: TAsyncLocaleDataProps['applicationMessages'];
+  children: ReactNode;
+};
+
+const CustomViewDevHost = (props: TCustomViewDevHost) => {
```
CarlosCortizasCT, line 125: This component is used as the landing page for Custom Views local development so users can use a Custom Application like host to trigger rendering the Custom View as it will be done in the Merchant Center.

image

image

packages/jest-preset-mc-app/setup-tests.js
```diff
@@ -1,4 +1,5 @@
 const util = require('util');
+const { MessageChannel } = require('worker_threads');
```
CarlosCortizasCT, line 2: This is needed to emulate how we create a private messages channel between host and Custom View in the test environment.

patches/jsdom@20.0.3.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/lib/jsdom/living/post-message.js b/lib/jsdom/living/post-message.js
+index db16576db84d3aa45fe8469b4325447c945097d7..1fe3890dcf7cf22f48a62bd8e2bafa41f85cb1e1 100644
+--- a/lib/jsdom/living/post-message.js
++++ b/lib/jsdom/living/post-message.js
+@@ -6,7 +6,7 @@ const { isValidTargetOrigin } = require("../utils");
+ const { fireAnEvent } = require("./helpers/events");
+ 
+ module.exports = function (globalObject) {
+-  return function (message, targetOrigin) {
++  return function (message, targetOrigin, ports) {
```
CarlosCortizasCT, line 10: We use the third parameter of `postMessage` in our code to create the private channel between the host application and the Custom View but `jsdom` does not support it yet:
* [Here](https://github.com/jsdom/jsdom/blob/master/lib/jsdom/living/post-message.js#L33) is where they have a TODO about it
* There's [this issue](https://github.com/jsdom/jsdom/issues/2448) asking for its support
* [Here's a PR](https://github.com/jsdom/jsdom/pull/3521) trying to implement this missing functionality (already 4 months old)

Since our case is very basic, I implemented this simplistic patch so it works in our tests.

packages/application-shell/src/test-utils/custom-views-utils.tsx
```diff
@@ -0,0 +1,107 @@
+import { type ReactNode, useEffect, useRef } from 'react';
```
CarlosCortizasCT, line 1: This new file contains helpers specific for Custom Views tests so we don't have to include them in the template so we keep control over them.

custom-views-templates/starter-typescript/src/components/channels/channels.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+  return (
+    <Spacings.Stack scale="xl">
+      <Spacings.Stack scale="s">
+        <Text.Headline as="h2" intlMessage={messages.title} />
+        <Text.Subheadline as="h4">
+          {intl.formatMessage(messages.subtitle, {
+            firstName: customViewContext.user?.firstName,
+            lastName: customViewContext.user?.lastName,
+          })}
```
CarlosCortizasCT, line 64: This is just to showcase how to access information from the context.

custom-views-templates/starter-typescript/src/components/entry-point/entry-point.tsx
```diff
@@ -0,0 +1,66 @@
...
+        environment={environment}
+        applicationMessages={loadMessages}
+      >
+        <CustomViewMain
+          customViewId={environment.customViewId}
+          messages={loadMessages}
+        />
+      </CustomViewDevHost>
+    );
+  }
```
CarlosCortizasCT, line 55: Locally we use the new component to emulate a MC application where we can trigger the Custom View from.

custom-views-templates/starter-typescript/src/test-utils/index.tsx
```diff
@@ -0,0 +1,103 @@
+import { graphql } from 'msw';
+import { mapResourceAccessToAppliedPermissions } from '@commercetools-frontend/application-shell/test-utils';
+
+export const getDefaultCustomViewServerMocks = () => [
```
CarlosCortizasCT, line 4: I would like to move this to the new Custom views test utils file (in the `application-shell` package) but that would imply to include `msw` as a dependency of that package and we don't want that (would be bloating the bundle size).

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.ts
```diff
@@ -85,23 +114,11 @@ const useIsAuthorized = ({
...
-  const actualActionRights =
-    useApplicationContext<TNormalizedActionRights | null>(
-      (applicationContext) =>
-        projectPermissions?.actionRights ?? applicationContext.actionRights
-    );
-  const actualDataFences = useApplicationContext<TNormalizedDataFences | null>(
-    (applicationContext) =>
-      projectPermissions?.dataFences ?? applicationContext.dataFences
-  );
+  const { resolvedPermissions, resolvedActionRights, resolvedDataFences } =
```
CarlosCortizasCT, line 117: Up until now we were reading the permissions from the application context but this is not enough when using this hook from a Custom View, since we won't have them over there but in the custom view context so I decided to create a private utility function to try to fetch permissions from both so we don't need to change the signature of this hook or create a new one.

packages/application-shell/src/components/custom-view-shell-authenticated/custom-view-shell-authenticated.tsx
```diff
@@ -63,14 +88,19 @@ function CustomViewShellAuthenticated(
...
+
                             return (
                               <ApplicationContextProvider
                                 user={user}
                                 project={project}
                                 projectDataLocale={props.dataLocale}
                                 environment={props.environment}
                               >
-                                {props.children}
+                                <CustomViewWithPermissionCheck
```
CarlosCortizasCT, line 99: We make sure the user has, at least, view permission for the current custom view.

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.ts
emmenko: I'm a bit confused. Isn't the `ApplicationContext` also used in Custom Views? The permissions are usually exposed from the `FetchProject` data and mapped into the application context.

I'm also wondering if we really need a separate context for Custom Views but (re)use the Application Context and inject some `customView` specific data.

Let's have a chat about this maybe.

packages/application-shell/src/components/custom-view-dev-host/custom-view-dev-host.tsx
```diff
@@ -0,0 +1,146 @@
...
+  ownerId: '',
+  permissions: [],
+  status: TCustomViewStatus.Draft,
+  type: TCustomViewType.CustomPanel,
+  typeSettings: {
+    size: TCustomViewSize.Small,
+  },
+  updatedAt: '',
+  url: '',
+};
```
emmenko, line 41: Does this need to be hardcoded here? Isn't this parsed and loaded from the config file? 🤔

emmenko (commented):
Awesome work so far, definitely lots of places that depend on each other. Hopefully that helps you to better understand the setup 💪

I went through the changes and I ended up having several questions and clarification points.

Let's have a sync to discuss them, thanks 🤗

website/scripts/prebuild.sh
```diff
@@ -4,4 +4,4 @@ set -e
 
 # Copy the custom application config `schema.json` to the website static assets.
 echo "Copying JSON schema for Custom Applications"
-cp ../packages/application-config/schema.json static/
+cp ../packages/application-config/custom-application.schema.json static/
```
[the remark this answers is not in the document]
CarlosCortizasCT: Ok, I guess I didn't realized this is how we host the schema publicly, right?
Perhaps we can have this kind of configuration instead:
```
cp ../packages/application-config/custom-application.schema.json static/schema.json
cp ../packages/application-config/custom-application.schema.json static/
cp ../packages/application-config/custom-view.schema.json static/
```

packages/application-config/src/load-config.ts
CarlosCortizasCT: I didn't expect a scenario where a user have both configuration files in the same application.
Maybe you can share your concern about how that could be possible?

In any case, given such a use case, why we would be preferring one over the other? We would not know what the users' application is supposed to implement.

packages/application-config/src/process-config.ts
CarlosCortizasCT: Yes, I thought about it but didn't want to deviate too much and adjusting things here and there seemed fine to me, but I agree it would be cleaner.
Let me try to refactor this part.

packages/application-config/src/process-config.ts
CarlosCortizasCT: We would be creating a fake ID if it does not exist in the config file.
Currently, it will be just a timestamp.

Please take a look at line 25 in the template [entry point file](https://github.com/commercetools/merchant-center-application-kit/pull/3175/files/42d36ac7b1891f6366cfc6ff946ab1f389b2619a#diff-b5adf3d372fa6efa1c98b999fea58ccd2de7b26112c36f095e89f495ace3df8d).

packages/application-config/src/transformers.ts
CarlosCortizasCT: I think I don't follow the change you would like to introduce. 
Perhaps we can sync about this one.

patches/jsdom@20.0.3.patch
CarlosCortizasCT: Not needed anymore as we are now mocking this part in a different way (mock the providers instead of relying on the communication to happen in the tests environment).

Just leave the comment here for reference purposes.

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.ts
```diff
@@ -49,6 +52,32 @@ const useWarning = (condition: boolean, message: string) => {
...
 
+const useResolvePermissionsData = (
+  projectPermissions?: TProjectPermissions
+): {
+  resolvedPermissions: TNormalizedPermissions | null;
+  resolvedActionRights: TNormalizedActionRights | null;
+  resolvedDataFences: TNormalizedDataFences | null;
+} => {
+  const customApplicationContext = useApplicationContext();
+  const customViewContext = useCustomViewContext();
```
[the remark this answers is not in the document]
CarlosCortizasCT: Currently the context always have an empty config by default so it won't throw an error.

packages/application-shell/src/components/custom-view-dev-host/custom-view-dev-host.tsx
CarlosCortizasCT: We don't have all the information in the config file.
This will come from the host application in runtime but I didn't want to have a null value in the context for starters.

CarlosCortizasCT:
> I'm a bit confused. Isn't the ApplicationContext also used in Custom Views? The permissions are usually exposed from the FetchProject data and mapped into the application context.

> I'm also wondering if we really need a separate context for Custom Views but (re)use the Application Context and inject some customView specific data.

> Let's have a chat about this maybe.

@emmenko 

Yes, that change wasn't actually needed. I implemented it at some point where things were not working but I think I then solved it the right way.
I rolled back that change in this commit: f4caaf5d75231242b6f6a73764f2317c43f1ceeb

packages/application-config/src/transformers.ts
```diff
@@ -92,4 +102,38 @@ function transformCustomApplicationConfigToData(
...
+
+  return {
+    id: customViewConfig.env.production.customViewId,
+    name: customViewConfig.name,
+    description: customViewConfig.description,
+    url: customViewConfig.env.production.url,
+    permissions: getPermissions(customViewConfig),
+    type: customViewConfig.type,
+    typeSettings: customViewConfig.typeSettings,
+    hostUrl: customViewConfig.env.development.hostUrl,
```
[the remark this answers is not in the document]
CarlosCortizasCT: Yes this is used only in development so we can load the emulated host URL from the configuration file (you can see it in the line of the [entry point file](https://github.com/commercetools/merchant-center-application-kit/pull/3175/files#diff-b5adf3d372fa6efa1c98b999fea58ccd2de7b26112c36f095e89f495ace3df8d)).

Maybe there is another way of doing it so I'm open for suggestions.

packages/application-config/src/process-config.ts
```diff
@@ -28,17 +32,23 @@ type ProcessConfigOptions = Partial<LoadingConfigOptions> & {
...
 const developmentAppUrl = `http://localhost:${developmentPort}`;
 
+const getConfigurationType = (configFileName: string): ConfigType => {
+  if (configFileName.includes('custom-application-config')) {
+    return ConfigType.CUSTOM_APPLICATION;
+  } else if (configFileName.includes('custom-view-config')) {
+    return ConfigType.CUSTOM_VIEW;
+  } else {
+    throw new Error(`Invalid config filename: ${configFileName}`);
+  }
```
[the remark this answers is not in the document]
CarlosCortizasCT: Thanks.

Update here: 698d360623536be506ec367233c4a1fb221179e5

packages/application-config/src/process-config.ts
```diff
@@ -28,17 +32,23 @@ type ProcessConfigOptions = Partial<LoadingConfigOptions> & {
...
 const omitDevConfigIfEmpty = (
   devConfig: ApplicationRuntimeConfig['env']['__DEVELOPMENT__']
 ) => {
-  if (
-    // @ts-expect-error: the `accountLinks` is not explicitly typed as it's only used by the account app.
-    devConfig?.accountLinks ||
-    devConfig?.menuLinks ||
-    devConfig?.oidc
-  )
+  // @ts-expect-error: the `accountLinks` is not explicitly typed as it's only used by the account app.
```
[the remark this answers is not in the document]
CarlosCortizasCT: This was an unintended change, sorry.

Reverted here: 698d360623536be506ec367233c4a1fb221179e5

packages/permissions/src/hooks/use-is-authorized/use-is-authorized.ts
CarlosCortizasCT: Yes, that change wasn't actually needed. I implemented it at some point where things were not working but I think I then solved it the right way.
I rolled back that change in this commit: https://github.com/commercetools/merchant-center-application-kit/commit/f4caaf5d75231242b6f6a73764f2317c43f1ceeb

Regarding unifying contexts, I still believe is better for organization to have different ones serving different purposes.
Also, the new one would not change at runtime because it's for improve context data to be easily available throughout the Custom View code and it will be populated with the (static) data received from the host application.

custom-views-templates/starter-typescript/src/components/entry-point/entry-point.tsx
```diff
@@ -0,0 +1,27 @@
...
+);
+
+// Ensure to setup the global error listener before any React component renders
+// in order to catch possible errors on rendering/mounting.
+setupGlobalErrorListener();
+
+const EntryPoint = () => (
+  <CustomViewShell applicationMessages={loadMessages}>
+    <AsyncApplicationRoutes />
+  </CustomViewShell>
```
emmenko, line 22: From a user perspective this is now essentially the same as rendering `<ApplicationShell>` (simple and familiar).

packages/application-components/src/index.ts
```diff
@@ -44,7 +44,6 @@ export type { TPageContentFull } from './components/page-content-containers/page
 
 // Custom views
 export { default as CustomPanel } from './components/custom-views/custom-panel';
-export { default as CustomViewLoader } from './components/custom-views/custom-view-loader';
```
emmenko, line 47: I decided to move it to `application-shell` package, as it does have some logic depending on the application context and it's easier to test it from within the application-shell package.

Users won't use this component directly anyway, so it's only for us internally.

packages/application-config/custom-view.schema.json
```diff
@@ -0,0 +1,245 @@
...
+              "description": "See https://docs.commercetools.com/TODO",
+              "type": "string"
+            },
+            "teamId": {
+              "type": "string"
+            },
+            "hostUriPath": {
+              "description": "See https://docs.commercetools.com/TODO",
+              "type": "string"
+            }
```
emmenko, line 114: I've been thinking about this (originally it was `hostUrl`) and opted to change it to `hostUriPath` as the URL origin is actually irrelevant and users don't have to worry about that.

I also don't expect users to use this much, as we now also provide a computed value for development

```ts
const defaultHostUriPath = oidcConfig.initialProjectKey
  ? `/${oidcConfig.initialProjectKey}/${entryPointUriPath}`
  : `/${entryPointUriPath}`;
```

packages/application-config/src/process-config.ts
```diff
@@ -28,21 +40,182 @@ type ProcessConfigOptions = Partial<LoadingConfigOptions> & {
...
+  ) {
     return devConfig;
+  }
   return undefined;
 };
 
+const isCustomViewData = (
+  data: CustomApplicationData | CustomViewData
+): data is CustomViewData =>
+  (data as CustomApplicationData).entryPointUriPath === undefined;
```
emmenko, line 70: Helper function for type guards

packages/application-config/src/process-config.ts
```diff
@@ -28,21 +40,182 @@ type ProcessConfigOptions = Partial<LoadingConfigOptions> & {
...
+  return omitDevConfigIfEmpty<ApplicationRuntimeEnvironmentForDevelopment>({
+    oidc: oidcConfig,
+    menuLinks: {
+      icon: configurationData.icon,
+      ...configurationData.mainMenuLink,
+      submenuLinks: configurationData.submenuLinks,
+    },
+    // @ts-expect-error: the `accountLinks` is not explicitly typed as it's only used by the account app.
+    accountLinks: configurationData.accountLinks,
+  });
```
emmenko, line 142: Keeping the two versions of the dev config separate

packages/application-config/src/process-config.ts
```diff
@@ -28,21 +40,182 @@ type ProcessConfigOptions = Partial<LoadingConfigOptions> & {
...
+  if (isCustomViewData(configurationData)) {
+    const hostUriPath = (appConfig as JSONSchemaForCustomViewConfigurationFiles)
+      .env.development.hostUriPath;
+    const defaultHostUriPath = oidcConfig.initialProjectKey
+      ? `/${oidcConfig.initialProjectKey}/${entryPointUriPath}`
+      : `/${entryPointUriPath}`;
+    const hostUrl = new URL(
+      hostUriPath || defaultHostUriPath,
+      'http://localhost:3001'
+    );
```
emmenko, line 125: For development we construct the host URL programmatically, unless the URI path is explicitly provided in the config file

packages/application-config/src/process-config.ts
```diff
@@ -28,21 +40,182 @@ type ProcessConfigOptions = Partial<LoadingConfigOptions> & {
...
+  revision: string;
+  appConfig:
+    | JSONSchemaForCustomApplicationConfigurationFiles
+    | JSONSchemaForCustomViewConfigurationFiles;
+}): ApplicationRuntimeEnvironment => {
+  const entryPointUriPath = isCustomViewData(configurationData)
+    ? // When the application acts as the host for Custom Views, there is no real
+      // entry point to be used, therefore we use a special identifier.
+      CUSTOM_VIEW_HOST_ENTRY_POINT_URI_PATH
+    : configurationData.entryPointUriPath;
```
emmenko, line 172: There are many places in App Kit that rely in the `entryPointUriPath` field. I tried making it optional but there are too many tricky places to refactor.

Actually if we think about it, the Custom View is rendered as if it was a Custom Application. In particular, in development it does render as a Custom Application and therefore making the `entryPointUriPath` required.

Therefore, we are now using a "special" value for the `entryPointUriPath`: `@@custom-view-host@@`.
We can then leverage this value in the MC API to run some logic specific to Custom Views, together with the `X-Custom-View-Id` header.

packages/application-config/src/process-config.ts
```diff
@@ -28,21 +40,182 @@ type ProcessConfigOptions = Partial<LoadingConfigOptions> & {
...
+
+  // The real application ID is only used in production.
+  // In development, we prefix the entry point with the "__local" prefix.
+  // This is important to determine to which URL the MC should redirect to
+  // after successful login.
+  const applicationIdentifier = isProd
+    ? `${configurationData.id}:${entryPointUriPath}`
+    : `__local:${entryPointUriPath}`;
```
emmenko, line 180: Then we have the `applicationIdentifier`, which is what we send as `X-Application-Id` header.

After some back and forth, I realized this is a required field that can be use for both scenarios: Custom Applications and Custom Views.

As mentioned above, even for Custom Views we still need a `entryPointUriPath` even though it's a special value. Therefore, we can keep the same format for the `applicationIdentifier` of Custom Applications.

Examples:

```
# Custom Applications
123:avengers
__local:avengers

# Custom Views
123:@@custom-view-host@@
__local:@@custom-view-host@@
```

packages/application-config/src/process-config.ts
```diff
@@ -28,21 +40,182 @@ type ProcessConfigOptions = Partial<LoadingConfigOptions> & {
...
+    env: appEnvKey,
+    frontendHost: appUrl.host,
+    location: appConfig.cloudIdentifier,
+    mcApiUrl: mcApiUrl.origin,
+    revision,
+    servedByProxy: isProd,
+
+    // Application config
+    applicationId: applicationIdentifier,
+    applicationIdentifier,
```
emmenko, line 203: Previously this field was named `applicationId` but I prefer now to name it `applicationIdentifier` to make it less confusing with the actual application ID.

For backwards compatibility we'll keep it around for now.

packages/application-config/src/process-config.ts
```diff
@@ -28,21 +40,182 @@ type ProcessConfigOptions = Partial<LoadingConfigOptions> & {
...
+    applicationName: isCustomViewData(configurationData)
+      ? configurationData.defaultLabel
+      : configurationData.name,
```
emmenko, line 206: The `applicationName` is also used in different places (e.g. user agent) and therefore we have to keep it around.

packages/application-config/src/process-config.ts
```diff
@@ -28,21 +40,182 @@ type ProcessConfigOptions = Partial<LoadingConfigOptions> & {
...
+    entryPointUriPath,
+
+    // Custom view config
+    ...(isCustomViewData(configurationData)
+      ? { customViewId: configurationData.id }
+      : {}),
```
emmenko, line 212: For now we'll also keep this around, depending on how we'll adjust the permissions logic in the MC API.

packages/application-shell-connectors/src/components/custom-view-context/custom-view-context.tsx
```diff
@@ -0,0 +1,56 @@
...
+  // context object and not the initial value.
+  // Therefore, we can safely cast the value to be out `TCustomViewContext` type.
+  const customViewContext = context as TCustomViewContext;
+  return selector ? selector(customViewContext) : customViewContext;
+}
+
+// This is a workaround to trick babel/rollup to correctly export the function.
+// Most likely the problem arises with the use of overloading.
+// See related issue: https://github.com/babel/babel/issues/8361
+const useCustomViewContext = useCustomViewContextHook;
```
emmenko, line 54: Same as for `useApplicationContext`, we provide the ability to pass a selector function.

packages/application-shell/src/components/custom-view-loader/custom-view-loader.tsx
```diff
@@ -112,6 +118,15 @@ function CustomViewLoader(props: TCustomViewLoaderProps) {
...
   }
+  const panelSize = (props.customView.typeSettings?.size?.toLocaleLowerCase() ||
+    'large') as Lowercase<'SMALL' | 'LARGE'>;
+  const iFrameUrl = [
+    window.location.origin,
+    'custom-views',
+    props.customView.id,
+    'projects',
+    projectKey,
+  ].join('/');
```
emmenko, line 129: This would be the new route in the MC Proxy to serve the Custom Views

packages/application-shell/src/components/custom-view-shell-authenticated/custom-view-shell-authenticated.tsx
```diff
@@ -1,23 +1,50 @@
...
+
+const CustomViewWithPermissionCheck = (
+  props: TCustomViewWithPermissionCheckProps
+) => {
+  const permissionKeys = entryPointUriPathToPermissionKeys(props.customViewId);
+
+  // Require View permission to render the application.
+  const canView = useIsAuthorized({
+    demandedPermissions: [permissionKeys.View],
+  });
```
emmenko, line 34: Currently this always yields `false`, until we adjust the applied permissions logic in the MC API to expose the computed permission keys for Custom Views (for local development).

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
```diff
@@ -139,4 +157,23 @@ function CustomViewShell(props: TCustomViewShellProps) {
...
+    return (
+      <Suspense fallback={<ApplicationLoader />}>
+        <CustomViewDevHost applicationMessages={props.applicationMessages}>
+          <CustomViewShell applicationMessages={props.applicationMessages}>
+            {props.children}
+          </CustomViewShell>
+        </CustomViewDevHost>
+      </Suspense>
+    );
+  }
```
emmenko, line 171: For local development we're wrapping the `<CustomViewShell>` with `<CustomViewDevHost>`

playground/src/components/custom-views/demo-custom-view.jsx
```diff
@@ -68,7 +75,10 @@ function DemoCustomView() {
   const [isLoading, setIsLoading] = useState(true);
 
   useEffect(() => {
-    window.app.customViewId = `__local:${CUSTOM_VIEW_ID}`;
+    window.app.applicationIdentifier = `__local:${CUSTOM_VIEW_HOST_ENTRY_POINT_URI_PATH}`;
+    window.app.customViewId = CUSTOM_VIEW_ID;
+    window.app.__DEVELOPMENT__.customViewConfig = DEMO_CUSTOM_VIEW;
+    window.app.__DEVELOPMENT__.customViewHostUrl = window.location.href;
```
emmenko, line 81: For now it's fine but it's not very maintainable like this.

playground/src/components/custom-views/demo-custom-view.jsx
```diff
@@ -78,7 +88,7 @@ function DemoCustomView() {
   }
 
   return (
-    <CustomViewShell customViewId={CUSTOM_VIEW_ID} messages={{}}>
+    <CustomViewShell disableDevHost applicationMessages={{}}>
```
emmenko, line 91: Since we are rendering the `<CustomViewDevHost>` from within the `<CustomViewShell>`, we need to explicitly opt out of it even in development, as the setup we have in playground does not need that component.

emmenko (approved):
Alright, there are some follow ups to look into but overall it's good to go from my side! 🚀

Amazing job Carlos for the core and demo setups. Looking forward to see this in action in the Merchant Center 🙌

.changeset/nervous-bags-clap.md
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools-frontend/application-components': patch
+'@commercetools-frontend/application-shell': patch
+---
+
+Move component to application-shell package
```
Rhotimee, line 6: NIT: It will be nice to know the component(s) that were moved. 

```suggestion
Move CustomViewLoader to application-shell package
```

packages/application-config/src/process-config.ts
```diff
@@ -28,21 +40,182 @@ type ProcessConfigOptions = Partial<LoadingConfigOptions> & {
...
+
+  if (isCustomViewData(configurationData)) {
+    const hostUriPath = (appConfig as JSONSchemaForCustomViewConfigurationFiles)
+      .env.development.hostUriPath;
+    const defaultHostUriPath = oidcConfig.initialProjectKey
+      ? `/${oidcConfig.initialProjectKey}/${entryPointUriPath}`
+      : `/${entryPointUriPath}`;
+    const hostUrl = new URL(
+      hostUriPath || defaultHostUriPath,
+      'http://localhost:3001'
```
Rhotimee, line 124:
```suggestion
      developmentAppUrl
```

packages/application-config/src/process-config.ts
Rhotimee: That is a very good idea 🙌🏽

playground/src/components/custom-views/constants.js
```diff
@@ -0,0 +1,27 @@
+import {
+  TCustomViewSize,
+  TCustomViewType,
+} from '@commercetools-frontend/application-shell';
```
Rhotimee, line 4:
```suggestion
import type {
  TCustomViewSize,
  TCustomViewType,
} from '@commercetools-frontend/application-shell';
```

Rhotimee (approved):
This looks great! Thanks @CarlosCortizasCT and @emmenko. 

I left some minor suggestions but, generally, it's solid work. 

🚀

playground/src/components/custom-views/constants.js
emmenko: These are enums...actually let me remove these exports, we don't really need them. https://github.com/commercetools/merchant-center-application-kit/pull/3175/commits/4894226190ea078a6740d2adf46e5ecc02ab6ea1

.changeset/nervous-bags-clap.md
emmenko: Thanks https://github.com/commercetools/merchant-center-application-kit/pull/3175/commits/0f94a4aff80ab993378a8e76bc285060aba0cbe9

packages/application-config/src/process-config.ts
emmenko: Good catch! https://github.com/commercetools/merchant-center-application-kit/pull/3175/commits/88ef166358fab01a61de1d4e458f7878e22b783d

custom-views-templates/starter-typescript/custom-view-config.mjs
```diff
@@ -0,0 +1,28 @@
+/**
+ * @type {import('@commercetools-frontend/application-config').ConfigOptionsForCustomView}
+ */
+const config = {
+  name: 'Custom View Template Starter',
+  cloudIdentifier: 'gcp-eu',
+  env: {
+    development: {
+      // TODO: This should be populated in the template installation process
+      initialProjectKey: 'almond-40',
```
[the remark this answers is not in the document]
emmenko: Thanks https://github.com/commercetools/merchant-center-application-kit/pull/3175/commits/8700882503c196eba76deba810aabd509dcc571a

PR #3177, opened 2023-08-22
fix(app-shell): Navbar's z-index in Safari

kark:
#### Summary

|  Before |  After | 
|---|---|
|   |   | 

#### Description

This PR aims at solving an issue with `z-index` of navbar in Safari.

I came a long way to arrive at this solution.
Unfortunately, due to the way we compile navbar CSS styles we can't easily add Safari-targeted styles to https://github.com/commercetools/merchant-center-application-kit/blob/main/packages/application-shell/src/components/navbar/navbar.mod.css
because the respective [Safari-specific class names](https://stackoverflow.com/questions/16348489/is-there-a-way-to-apply-styles-to-safari-only) are edited by our post-css script.

I even attempted to add style rules to the compiled CSS file, but encountered challenges since we don't have access to the CSS module class names when appending the styles.

Hence, this is the approach I've come up with.

[Demo](https://mc-15223.mc-preview.europe-west1.gcp.escemo.com/adnan-nodejs-test-project/products)

### DIFF
--- a/packages/application-shell/src/components/navbar/navbar.mod.css
+++ b/packages/application-shell/src/components/navbar/navbar.mod.css
@@ -13,6 +13,10 @@
   --width-left-navigation: 64px;
   --width-left-navigation-when-expanded: 245px;
   --width-left-navigation-sublist: 205px;
+
+  /* Active list item with icon text */
+  --left-list-item-icon-text-active: 46px;
+  --width-list-item-icon-text-active: 166px;
 }
 
 /* Left Nav */
@@ -349,8 +353,9 @@
   -webkit-line-clamp: 2;
   -webkit-box-orient: vertical;
   position: absolute;
-  left: 46px;
-  width: 166px;
+  /* `left` and `width` required for displaying focus indicator properly */
+  left: var(--left-list-item-icon-text-active);
+  width: var(--width-list-item-icon-text-active);
   z-index: 1;
 }
 
--- a/packages/application-shell/src/components/navbar/navbar.mod.css
+++ b/packages/application-shell/src/components/navbar/navbar.mod.css
@@ -262,7 +262,6 @@
   background-color: var(--background-color-for-navbar);
   top: 0;
   left: 64px;
-  z-index: -1;
   list-style: none;
   opacity: 0.01;
   visibility: hidden;
@@ -317,10 +316,6 @@
 }
 
 .list-item.item__active .item-icon-text {
-  width: calc(
-    var(--width-left-navigation-sublist) + var(--width-left-navigation) -
-      var(--spacing-m) * 2
-  );
   justify-content: flex-start;
 }
 .list-item .icon-container {
@@ -353,6 +348,10 @@
   display: -webkit-box;
   -webkit-line-clamp: 2;
   -webkit-box-orient: vertical;
+  position: absolute;
+  left: 46px;
+  width: 166px;
+  z-index: 1;
 }
 
 :global(.body__menu-open) .list-item.item__active .item-icon-text .title {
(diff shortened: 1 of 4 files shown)


emmenko:
Thanks for looking into this! ❤️

One question first: do you know if this issue is also happening with the new navbar? cc @chloe0592

kark:
> Thanks for looking into this! ❤️
> 
> One question first: do you know if this issue is also happening with the new navbar? cc @chloe0592

Thanks. Actually, the new submenu hasn't been implemented yet, but since I will be doing this I will make sure it also works in Safari.

CarlosCortizasCT:
I'm sorry Kacper but I don't fully understand the problem and also not sure about the solution.
Maybe you could explain a little further why we have the problem and the z-index management needed for Safari 🙏 

Are those styles usable in other browsers? I mean, what happen if we do not conditionally apply them but use them for every browser?

If we need specific styles and PostCSS is messing with them, perhaps we can ignore them ([example](https://www.npmjs.com/package/postcss-plugin-ignore-file))

kark:
> I'm sorry Kacper but I don't fully understand the problem and also not sure about the solution. Maybe you could explain a little further why we have the problem and the z-index management needed for Safari 🙏
> 
> Are those styles usable in other browsers? I mean, what happen if we do not conditionally apply them but use them for every browser?
> 
> If we need specific styles and PostCSS is messing with them, perhaps we can ignore them ([example](https://www.npmjs.com/package/postcss-plugin-ignore-file))

Thanks Carlos,
That's a great question. I suppose it's due to WebKit's specific way of managing stacking contexts. I don't know the root cause of the problem and if this is a regression due to some recent changes or has existed for a longer time. I don't think that bisecting here would lead anywhere since we're working on restyling Navbar anyway so this is merely a hot-fix. 

Those styles work fine in Chrome and FF so the idea behind this PR is not to introduce any changes for those browsers.
If instead of targeting Safari specifically we add those styles for all browsers we'll end up with another bug in Chrome:
Screenshot 2023-08-21 at 17 25 49
which is probably even more tricky to solve and has bigger impact for our users.

The problem I see with using the ignore pattern for the second css file that you proposed is that we need to first pre-compile:
https://github.com/commercetools/merchant-center-application-kit/blob/main/packages/application-shell/src/components/navbar/navbar.mod.css
and then use the class names from `compiled/navbar.css.json` in the second iteration of compilation. Or perhaps we can do that easier in some other way I'm not aware of.

Apart from that, I tried to solve this issue with css with in-browser styles and I wasn't able to make it work 🤷‍♂️ 
[None of the solutions with browser-specific media-queries and nested styles](https://stackoverflow.com/a/25975282) worked well, for example:
Screenshot 2023-08-22 at 17 14 01

CarlosCortizasCT:
Thanks @kark for the detailed explanation.

So it seems the new styles would be a problem for users navigating the document using the tab key as, when they focus the icon in the collapsed state, the visual border indicator is not correct.

I was still trying to find out if we could have a solution not involving User-Agent sniffing since it uses to be a risky solution from the maintainability perspective.

I pushed [this PR](https://github.com/commercetools/merchant-center-application-kit/pull/3182) with an alternative where I just modified some styles which seems to work in Safari, Chrome and Firefox.
Could you please take a look and let me know if you see any issue with it? 🙏 

Anyway, I agree this issue will be gone once we implement @chloe0592 work on NavBar scrolling (reference).

kark:
Hey @commercetools/shield-team-fe,
a solution proposed by Carlos was integrated to this PR as it makes much more sense.
I wanted to make everyone aware of one change introduced with this solution compared to the current state which is the focus indicator (visible when e.g. using keyboard navigation).
| Before  | After  | 
|---|---|
| Screenshot 2023-08-24 at 13 22 32 |  Screenshot 2023-08-24 at 13 20 36 |

CarlosCortizasCT (approved):

chloe0592 (approved):
Thank you! 🙌

packages/application-shell/src/components/navbar/navbar.mod.css
```diff
@@ -353,6 +348,10 @@
   display: -webkit-box;
   -webkit-line-clamp: 2;
   -webkit-box-orient: vertical;
+  position: absolute;
+  left: 46px;
+  width: 166px;
```
emmenko, line 353: Can we do something about these magic numbers?

emmenko (approved):
Love it, thanks! ❤️

kark:
I will merge it now to test out the canary release in MC FE. If css vars names have to be changed please let me know.

PR #3178, opened 2023-08-22
fix(dev-auth): request / response compatibility

emmenko:
Fixes the issue when developing locally against local API that `response.send` is not a function.

I also refactored the types to use the ones from `connect`, as this is also what [Vite uses for its middlewares](https://vitejs.dev/guide/ssr.html#setting-up-the-dev-server)
image

### DIFF
--- a/packages/mc-dev-authentication/package.json
+++ b/packages/mc-dev-authentication/package.json
@@ -28,10 +28,10 @@
     "@babel/runtime-corejs3": "^7.20.13"
   },
   "devDependencies": {
-    "@commercetools-frontend/application-config": "workspace:*",
-    "@types/express": "^4.17.17",
+    "@commercetools-frontend/application-config": "22.6.0",
+    "@types/connect": "^3.4.35",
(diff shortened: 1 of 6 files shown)


packages/mc-dev-authentication/package.json
```diff
@@ -28,10 +28,10 @@
     "@babel/runtime-corejs3": "^7.20.13"
   },
   "devDependencies": {
-    "@commercetools-frontend/application-config": "workspace:*",
-    "@types/express": "^4.17.17",
+    "@commercetools-frontend/application-config": "22.6.0",
```
CarlosCortizasCT, line 31: Just double checking if changing this dependency from workspace to an specific one intended

emmenko: Ah this is a dev dep...yeah sorry we can keep using `workspace`, I'll revert

CarlosCortizasCT (approved):
Thanks! 👍

PR #3180, opened 2023-08-23
Change <MainPageContent> style from margin to padding

ddouglasz:
#### Summary

Change `MainPageContent` style from margin to padding to avoid cutting off some stylings from it's component's children

#### Description

The current styling from the `MainPageContent` used in the `CustomFormMainPage` has a margin property to space the container children from external contents but this seems like the property should be padding instead. The current margin property seems to cut of some of the features of the pagination dropdown box shadow. 

Before: 
Screenshot 2023-08-23 at 10 12 25

After: 
Screenshot 2023-08-23 at 10 15 16

### DIFF
--- a/.changeset/hungry-trees-vanish.md
+++ b/.changeset/hungry-trees-vanish.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-components': patch
+---
+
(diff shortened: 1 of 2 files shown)


emmenko (approved):
Good catch, thanks!

chloe0592 (approved):
Nice! Thank you 😊

PR #3181, opened 2023-08-23
feat: support new application locale `pt-BR`

emmenko:


packages/i18n/src/load-i18n.ts
```diff
@@ -21,6 +21,10 @@ const getUiKitChunkImport = (locale: string): Promise<TI18NImportData> => {
       return import(
         /* webpackChunkName: "i18n-ui-kit-locale-fr-FR" */ '@commercetools-uikit/i18n/compiled-data/fr-FR.json'
       );
+    // case 'pt-BR':
+    //   return import(
+    //     /* webpackChunkName: "i18n-ui-kit-locale-pt-BR" */ '@commercetools-uikit/i18n/compiled-data/pt-BR.json'
+    //   );
```
emmenko, line 27: Requires https://github.com/commercetools/ui-kit/pull/2583 to be released first

packages/application-shell/src/hooks/use-applications-menu/use-applications-menu.ts
```diff
@@ -16,8 +17,6 @@ import type {
 
 import FetchApplicationsMenu from './fetch-applications-menu.proxy.graphql';
 
-const supportedLocales = ['en', 'de', 'es', 'fr-FR', 'zh-CN', 'ja'];
```
emmenko, line 19: Moved to the constants package

packages/i18n/scripts/compile_data.sh
```diff
@@ -2,6 +2,8 @@
 
 set -e
 
+export NODE_ENV="production"
```
emmenko, line 5: Since the scripts rely on the constants package, babel needs to know the environment, so we're setting it explicitly.

packages/application-shell/src/hooks/use-applications-menu/use-applications-menu.ts
```diff
@@ -8,16 +8,14 @@ import type {
...
+import { getSupportedLocales } from '@commercetools-frontend/l10n';
 import { useMcQuery } from '../../hooks/apollo-hooks';
 import type {
   TFetchApplicationsMenuQuery,
   TFetchApplicationsMenuQueryVariables,
 } from '../../types/generated/proxy';
-
 import FetchApplicationsMenu from './fetch-applications-menu.proxy.graphql';
 
-const supportedLocales = ['en', 'de', 'es', 'fr-FR', 'zh-CN', 'ja'];
```
emmenko, line 19: Moved to the `l10n` package, as it contains most of the logic around locales

packages/l10n/src/utils.ts
```diff
@@ -1,23 +1,30 @@
+import type { TSupportedLocale } from '../supported-locales';
+// https://babeljs.io/blog/2017/09/11/zero-config-with-babel-macros
+import supportedLocales from /* preval */ '../supported-locales';
```
emmenko, line 3: Since the `supported-locales` is a CJS module, it's very tricky to make it so that it can be loaded from both CJS and TS/ESM.

As we do in other places of the codebase, we can use preval to load the module as CJS and inject it the variable here.

transifex.yml
emmenko, line 1: I'll update the TX integration to use this config file. Makes it easier to manage and update.

packages/l10n/src/utils.ts
chloe0592: @emmenko I've seen that we used `/* preval */` couple of times in our code, but wasn’t exactly sure when we use it. 
If I understood correctly, we import `supportedLocales` here and with `prevail` it will be evaluated at build time, right?
Is it used here for performances reason or to make sure imported data remains consistent? Or something else entirely?

Thank you!

kark (approved):
💯

ddouglasz (approved):
Looks good🙌🏾

Rhotimee (approved):
Nice work 🙌🏽

packages/l10n/src/utils.ts
emmenko: Thanks for asking.

Most of the time we use preval to perform some operations with the filesystem that would be otherwise impossible within a browser.
The preval macro will evaluate that code and inject it into the bundle.

Here we're leveraging that just for the simplicity of loading a CJS module that would be otherwise tricky to load, as we also need it for nodejs scripts.

I tried different ways of loading the module (as we also need to assign the type declarations to it) but I would get errors like

```
Error: 'default' is not exported by packages/l10n/supported-locales.js, imported by packages/l10n/src/utils.ts
```

tdeekens (approved):

PR #3182, opened 2023-08-23
Alternative to fix `NavBar` unexpected behaviour in Safari

CarlosCortizasCT:
…ri behaviour

#### Summary

Alternative to fix `NavBar` unexpected behaviour in Safari.

#### Description

Here I'm trying to refactor the styles to avoid User-Agent sniffing.

Maybe this is not the most elegant solution but it seems to work in Safari, Chrome an Firefox. 
Also, this would be a temporary fix anyway until we implement the new scrolling behaviour for the navbar.

### DIFF
--- a/packages/application-shell/src/components/navbar/menu-items.tsx
+++ b/packages/application-shell/src/components/navbar/menu-items.tsx
@@ -4,8 +4,6 @@ import {
   MouseEventHandler,
   ReactNode,
   SyntheticEvent,
-  useRef,
-  useLayoutEffect,
 } from 'react';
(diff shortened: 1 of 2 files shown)


kark (approved):
Thanks Carlos, 
That looks good to me and definitely makes more sense than my proposal 👍
I'll close my PR now

CarlosCortizasCT:
> Thanks Carlos, That looks good to me and definitely makes more sense than my proposal 👍 I'll close my PR now

@kark Maybe we can keep your PR and just merge this over there so we can keep track of the different approaches we went through

PR #3183, opened 2023-08-23
fix: hidden submenu overflow

kark:
#### Summary

The issue was identified by and solution was proposed by @jaikumar-tj.
Many thanks for that 🙇 

|  Before |  After | 
|---|---|
| Screenshot 2023-08-23 at 18 21 27 |  Screenshot 2023-08-23 at 18 22 30 |

The root cause is hidden submenu not being removed from the visual flow due to `visibility: hidden`
Screenshot 2023-08-23 at 18 25 19

[Link to demo](https://mc-15223.mc-preview.europe-west1.gcp.escemo.com/test-with-big-data-44/welcome)

### DIFF
--- a/.changeset/stupid-dolphins-decide.md
+++ b/.changeset/stupid-dolphins-decide.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': patch
+---
+
+Resolve the problem of a hidden menu with multiple items overflowing the layout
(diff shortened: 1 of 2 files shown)


jaikumar-tj (approved):
@kark Thanks for fixing it quickly. Just want to confirm, there is no animation applied for submenu visibility (like fade-in, swipe-in), right? because `display:none` might disrupt the animation and we may have to think different way to fix it.

CarlosCortizasCT (approved):
Thanks! 👍

kark:
> @kark Thanks for fixing it quickly. Just want to confirm, there is no animation applied for submenu visibility (like fade-in, swipe-in), right? because `display:none` might disrupt the animation and we may have to think different way to fix it.

@jaikumar-tj Thanks, that's a great point. I don't think there are any animations / transitions applied to the submenu itself.

PR #3185, opened 2023-08-28
refactor(Navbar): new Submenu - part 1

kark:
#### Summary

This PR adds first part of the new Submenu styles. 

**It does not cover the "Tooltip" and the correct positioning of the submenu**.

Screenshot 2023-08-28 at 14 14 32
Screenshot 2023-08-28 at 14 14 43

### DIFF
--- a/.changeset/serious-hats-prove.md
+++ b/.changeset/serious-hats-prove.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': patch
+---
+
+Add new submenu styles
--- a/.changeset/serious-hats-prove.md
+++ b/.changeset/serious-hats-prove.md
@@ -2,4 +2,4 @@
 '@commercetools-frontend/application-shell': patch
 ---
 
-Add new submenu styles
+Add new navigation submenus styles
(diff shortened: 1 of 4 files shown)


kark:
> Nice work @kark 💯, I see the PR already stated not covering the correct position of the submenu but I wanted to point if that also meant for the z-index in cases like these where the submenu is layered together with some modal components?
> 

Thank you Douglas, good point 👍

chloe0592:
@kark thank you for working on it! Nicely done 💪

chloe0592:
Just wanted to mention something when we are already on the topic of `z-index `. But I believe it can be done in a separate PR, not to block this one, if it's not a quick fix.
There's a misplacement of submenus when we have notifications popping out. So that's something we'll need to look into.

Screenshot 2023-08-29 at 08 33 05
Screenshot 2023-08-29 at 08 33 21

kark:
> Just wanted to mention something when we are already on the topic of `z-index `. But I believe it can be done in a separate PR, not to block this one, if it's not a quick fix. There's a misplacement of submenus when we have notifications popping out. So that's something we'll need to look into.
> 
> Screenshot 2023-08-29 at 08 33 05 Screenshot 2023-08-29 at 08 33 21

Thanks, that's a very good point.
I believe it would be more suitable to handle the correct submenu placement in a separate PR since there are many more factors. This PR only applies the submenu styles.

chloe0592 (approved):

ddouglasz (approved):
🙏🏾

PR #3188, opened 2023-08-29
fix: Navbar submenu causing issues in e2e tests

kark:
#### Summary

This PR reverts the change made in #3183.
The change led to failures of the e2e tests due to the submenu consistently appearing during testing.

~~While I'm not entirely certain that this pull request will definitively resolve the test issue, since I can't replicate the bug locally (it appears only in the CI), I've been thinking of using of a branch deployment in the MC FE with a canary release of the `@commercetools-frontend/*` packages to verify if it works.~~

I have a reason to believe this should work. In the last pipeline [here](https://github.com/commercetools/merchant-center-frontend/pull/15223) most cases were resolved (although it's worth noting that some issues in the `discounts` app remain unresolved)

### DIFF
--- a/.changeset/sharp-donkeys-move.md
+++ b/.changeset/sharp-donkeys-move.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': patch
+---
+
+Enable unhiding Navbar's submenu by unsetting `display` value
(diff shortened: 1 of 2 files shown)


emmenko (approved):
Thanks

PR #3191, opened 2023-08-31
Navigation footer part 1

chloe0592:
**Footer requirements done with this PR:**
- Footer should be sticky on the bottom
- Imported new icons for expander
- New color tokens
- New divider
- Hover effect on the expander
- Change breakpoint when the expander is hidden (less than 1200 px)

Will be done later:
- Adjust width and spacings of the footer
- Add correct styling to 'Support' (will be done when changing `MenuItem` style)
- Space added so that the Faded component doesn’t hide the last `MenuItem` of the side nav

**Screenshots:**
Screenshot 2023-08-31 at 09 31 10
Screenshot 2023-08-31 at 09 31 03

### DIFF
--- a/packages/application-shell/src/components/navbar/menu-items.tsx
+++ b/packages/application-shell/src/components/navbar/menu-items.tsx
@@ -124,6 +124,16 @@ type MenuExpanderProps = {
   isMenuOpen: boolean;
   isNewNavigationEnabled: boolean;
 };
+
+const getIcon = ({ isNewNavigationEnabled, isMenuOpen }: MenuExpanderProps) => {
+  const Icon = isNewNavigationEnabled
+    ? isMenuOpen
+      ? SidebarCollapseIcon
+      : SidebarExpandIcon
+    : ArrowRightIcon;
+  return <Icon color="surface" size="big" />;
+};
+
 const MenuExpander = (props: MenuExpanderProps) => {
   return (
     <li
@@ -142,19 +152,7 @@ const MenuExpander = (props: MenuExpanderProps) => {
         })}
         data-testid="menu-expander"
       >
-        {/*
-          FIXME: define hover effect.
-          https://github.com/commercetools/merchant-center-frontend/issues/2216
-        */}
-        {props.isNewNavigationEnabled ? (
-          props.isMenuOpen ? (
-            <SidebarCollapseIcon color="surface" size="big" />
-          ) : (
-            <SidebarExpandIcon color="surface" size="big" />
-          )
-        ) : (
-          <ArrowRightIcon color="surface" size="big" />
-        )}
+        {getIcon(props)}
       </div>
     </li>
   );
--- a/packages/application-shell/src/components/navbar/menu-items.tsx
+++ b/packages/application-shell/src/components/navbar/menu-items.tsx
@@ -21,7 +21,12 @@ import type {
 import MissingImageSvg from '@commercetools-frontend/assets/images/diagonal-line.svg';
 import { NO_VALUE_FALLBACK } from '@commercetools-frontend/constants';
 import { RestrictedByPermissions } from '@commercetools-frontend/permissions';
-import { ArrowRightIcon, BackIcon } from '@commercetools-uikit/icons';
+import {
+  ArrowRightIcon,
+  BackIcon,
+  SidebarExpandIcon,
+  SidebarCollapseIcon,
+} from '@commercetools-uikit/icons';
 import InlineSvg from '@commercetools-uikit/icons/inline-svg';
 import type {
   TDataFence,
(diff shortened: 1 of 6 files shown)


packages/application-shell/src/components/navbar/menu-items.tsx
```diff
@@ -131,42 +122,60 @@ type MenuExpanderProps = {
...
-        <ArrowRightIcon color="surface" size="big" />
+        {props.isNewNavigationEnabled ? (
+          props.isMenuOpen ? (
+            <SidebarCollapseIcon color="surface" size="big" />
+          ) : (
+            <SidebarExpandIcon color="surface" size="big" />
+          )
+        ) : (
+          <ArrowRightIcon color="surface" size="big" />
+        )}
```
kark, line 157: nit: I'm personally fine with that, but I think we should aim to avoid double-nested conditions. Perhaps you could consider using a `do {}` syntax or refactor to use an external function. But it's totally up to you.

packages/application-shell/src/components/navbar/navbar.mod.css
```diff
@@ -294,6 +341,21 @@
...
+  font-size: var(--font-size-for-navbar-link);
+  /* stylelint-disable-next-line csstools/value-no-unknown-custom-properties */
+  background-color: var(--background-color-for-navbar);
+  left: 58px;
+  z-index: -1;
+  list-style: none;
+  position: fixed;
+  display: none;
+}
+
```
kark, line 358: I think it's redundant (added 2 times in this file)

kark (commented):
Thank you, looks great 🎉
I only wanted to make sure that this behaviour with the expand button hiding is now expected:


With the menu collapsed we can only see the expand button when viewport is exactly 1201 px wide (most of the time it's hidden). 
Screenshot 2023-09-04 at 09 53 12

packages/application-shell/src/components/navbar/menu-items.tsx
chloe0592: @kark thanks for pointing that out. I agree it can be written more concisely. 😊 
Let me know if this looks better now with this change - [7198728](https://github.com/commercetools/merchant-center-application-kit/pull/3191/commits/71987284dc9a3172c353905aaab8fd9cda0e28f8)

packages/application-shell/src/components/navbar/navbar.mod.css
chloe0592: Good catch! [33821ab](https://github.com/commercetools/merchant-center-application-kit/pull/3191/commits/33821ab1f3e635251915717f6846bf21137138fa)

chloe0592:
> Thank you, looks great 🎉 I only wanted to make sure that this behaviour with the expand button hiding is now expected: 
> 

Thank you @kark for your feedback! 🙏 

I see what you mean. It's a good point!

Maybe this could fix it - [5e96f1e](https://github.com/commercetools/merchant-center-application-kit/pull/3191/commits/5e96f1e688a63531cd899e876561d3c0864ec17c).

https://github.com/commercetools/merchant-center-application-kit/assets/52276952/cdead007-de55-4c25-a9f2-b52745b2f5bc

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
```diff
@@ -135,7 +136,8 @@ const useNavbarStateManager = (props: HookProps) => {
   const checkSize = useCallback(
     throttle(() => {
       const shouldOpen = window.innerWidth > 1024;
-      const canExpandMenu = window.innerWidth > 918;
+      const canExpandMenu =
+        window.innerWidth > (props.isNewNavigationEnabled ? 1200 : 918);
```
CarlosCortizasCT, line 140: I was wondering if we should move these values to the package [constants file](https://github.com/commercetools/merchant-center-application-kit/blob/main/packages/application-shell/src/constants.ts#L1).

CarlosCortizasCT:
Just wanted to double check something to know whether it will be addressed later:

Will the gradient size will be adjusted so the last menu item can be fully visible/clickable?
Screenshot 2023-09-04 at 17 15 38


Also, I was wondering about the hover effect for the support menu item as we're changing the background color for the whole element but the horizontal separator has not the full width so it looks odd to me:
Screenshot 2023-09-04 at 17 09 16

chloe0592:
> Also, I was wondering about the hover effect for the support menu item as we're changing the background color for the whole element but the horizontal separator has not the full width so it looks odd to me: Screenshot 2023-09-04 at 17 09 16

Yes, since the `list-item` styles will be adjusted in the side nav PR so will the support menu as well. 
And it'll behave like the rest of the `list-items` with a small spacing underneath to divide it from the separator.

**Figma prototype** 

https://github.com/commercetools/merchant-center-application-kit/assets/52276952/8b5ae1c8-c1de-4fd2-9c72-eb7fc4c22286

kark:
> Thank you @kark for your feedback! 🙏
> 
> I see what you mean. It's a good point!
> 
> Maybe this could fix it - [5e96f1e](https://github.com/commercetools/merchant-center-application-kit/pull/3191/commits/5e96f1e688a63531cd899e876561d3c0864ec17c).
> 
Thank you, that is much better (although please note that now the Navbar doesn't auto-expand when the viewport is above 1200 px wide). I think we should ask @FilPob if the expand button should be hidden for small viewports. I don't see that in the designs, but maybe it was omitted on purpose.

FilPob:
Thank you all, looks all good so far :) 

For the expanding and collapsing, I wanted to reuse the same logic as we have right now, just change the breakpoint to 1200px. 

- If screen is < than 1200px, the nav is collapsed and the expand button is hidden. 
- If the screen is > 1200px the navigation expands and the collapse button appears. 
- In case the user explicitly sets the nav as collapsed, it stays collapsed regardless of the screen size. Only the expand button should be hidden when < than 1200px

chloe0592:
Closed since this PR is opened - https://github.com/commercetools/merchant-center-application-kit/pull/3200

PR #3192, opened 2023-08-31
refactor(Navbar): Show submenu and a tooltip when collapsed 

kark:
#### Summary

This PR adds a tooltip with the app name in the collapsed mode of the navbar.

It is deployed [here](https://mc-15303.mc-preview.europe-west1.gcp.escemo.com/main-navigation-test/welcome).

Please note that for now the submenu pops up on click when the navbar is expanded. I assumed the correct behaviour (submenu showing up on hover) will be part of [this](https://commercetools.atlassian.net/browse/SHIELD-884) ticket.

### DIFF
--- a/packages/application-shell/src/components/navbar/navbar-new/navbar.mod.css
+++ b/packages/application-shell/src/components/navbar/navbar-new/navbar.mod.css
@@ -6,6 +6,8 @@
   --icon-container-offset: 4px;
   --item-height: 56px;
   --item-width: 64px;
+  --sublist-new-indentantion: 58px;
+  --sublist-width: 253px;
   --submenu-margin: calc(
     var(--icon-size) + var(--spacing-m) + var(--icon-container-offset) + 4px
   );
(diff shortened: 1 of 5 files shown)


packages/application-shell/src/components/navbar/navbar-new/navbar.mod.css
```diff
@@ -242,13 +226,24 @@
...
   background-color: var(--background-color-for-navbar);
-  left: 58px;
+  left: var(--sublist-indentantion);
   z-index: -1;
   list-style: none;
   position: fixed;
   display: none;
 }
 
+/* This pseudo-element is required to enable smooth coursor movement from the main menu item to submenu items when tooltip is displayed */
```
CarlosCortizasCT, line 236: 👏

packages/application-shell/src/components/navbar/navbar-new/navbar.tsx
```diff
@@ -50,6 +50,8 @@ import compiledStyles from /* preval */ './navbar.styles';
 
 const styles = compiledStyles.jsonMap;
 
+const menuItemHeightInPx = 56;
```
CarlosCortizasCT, line 53: **nit**: do you think it would make sense to move this value to the [constants file](https://github.com/commercetools/merchant-center-application-kit/blob/main/packages/application-shell/src/constants.ts#L1)?

CarlosCortizasCT (approved):
Thanks! 👏

packages/application-shell/src/components/navbar/navbar-new/navbar.tsx
kark: Thank you, that's a great idea 🙌

packages/application-shell/src/components/navbar/navbar-new/navbar.mod.css
chloe0592: Great solution! 🚀

PR #3193, opened 2023-09-04
fix: use default value for `isForcedMenuOpen`

kark:
#### Summary

Fixing inconsistencies in the Navbar's behavior. Check E2E tests now passing [here](https://github.com/commercetools/merchant-center-frontend/pull/15223)

For more context, currently the Navbar's behavior is somewhat inconsistent. For viewport widths below 1024 px until  `isForcedMenuOpen` is set by clicking the expand button, it sometimes renders with menu expanded and sometimes collapsed (some kind of race condition takes place).

### DIFF
--- a/packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
+++ b/packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
@@ -130,10 +130,12 @@ const useNavbarStateManager = (props: HookProps) => {
     ? isForcedMenuOpenDefaultValue
     : (JSON.parse(cachedIsForcedMenuOpen) as boolean);
 
-  window.localStorage.setItem(
-    STORAGE_KEYS.IS_FORCED_MENU_OPEN,
-    String(isForcedMenuOpen)
-  );
+  if (!cachedIsForcedMenuOpen) {
+    window.localStorage.setItem(
+      STORAGE_KEYS.IS_FORCED_MENU_OPEN,
+      String(isForcedMenuOpen)
+    );
+  }
 
   const [state, dispatch] = useReducer<
     (prevState: State, action: Action) => State
--- a/packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
+++ b/packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
@@ -130,7 +130,7 @@ const useNavbarStateManager = (props: HookProps) => {
     ? isForcedMenuOpenDefaultValue
     : (JSON.parse(cachedIsForcedMenuOpen) as boolean);
 
-  if (!cachedIsForcedMenuOpen) {
+  if (isNil(cachedIsForcedMenuOpen)) {
     window.localStorage.setItem(
       STORAGE_KEYS.IS_FORCED_MENU_OPEN,
       String(isForcedMenuOpen)
--- a/packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
+++ b/packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
@@ -132,7 +132,7 @@ const useNavbarStateManager = (props: HookProps) => {
 
   window.localStorage.setItem(
     STORAGE_KEYS.IS_FORCED_MENU_OPEN,
-    String(isForcedMenuOpenDefaultValue)
+    String(isForcedMenuOpen)
   );
 
   const [state, dispatch] = useReducer<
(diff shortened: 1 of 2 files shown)


packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
```diff
@@ -125,9 +127,14 @@ const useNavbarStateManager = (props: HookProps) => {
     STORAGE_KEYS.IS_FORCED_MENU_OPEN
   );
   const isForcedMenuOpen = isNil(cachedIsForcedMenuOpen)
-    ? null
+    ? isForcedMenuOpenDefaultValue
     : (JSON.parse(cachedIsForcedMenuOpen) as boolean);
 
+  window.localStorage.setItem(
+    STORAGE_KEYS.IS_FORCED_MENU_OPEN,
+    String(isForcedMenuOpenDefaultValue)
```
CarlosCortizasCT, line 135: Shouldn't we be using here `isForcedMenuOpen`?

If we don't and the cached value is `true`, the we would be using `true` in next code flow but we would be storing `false` in the local storage.

kark: Oh, right, good point 👍

CarlosCortizasCT (approved):
Thanks 👍

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
```diff
@@ -125,9 +127,14 @@ const useNavbarStateManager = (props: HookProps) => {
...
   );
   const isForcedMenuOpen = isNil(cachedIsForcedMenuOpen)
-    ? null
+    ? isForcedMenuOpenDefaultValue
     : (JSON.parse(cachedIsForcedMenuOpen) as boolean);
 
+  window.localStorage.setItem(
+    STORAGE_KEYS.IS_FORCED_MENU_OPEN,
+    String(isForcedMenuOpen)
+  );
```
emmenko, line 136: Should we write it only if it wasn't set before?

```suggestion
  if (!cachedIsForcedMenuOpen) {
    window.localStorage.setItem(
      STORAGE_KEYS.IS_FORCED_MENU_OPEN,
      String(isForcedMenuOpen)
    );
  }
```

emmenko (commented):
Thanks!

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
emmenko: Yeah I think we need to set it only if it wasn't set, otherwise if the user expands the menu and reloads the page, the value will always be overwritten with `false`.

kark: Yep, thank you

CarlosCortizasCT: Sorry, maybe I'm missing something...

We first read the value from the localeStorage.
Then we calculate `isForcedMenuOpen` based on that value.
Afterwards, we store that calculated value in the localStorage, only if it's truthy.

So, the value won't be stored in the the localStorage if we didn't have any value in localStorage before (`cachedIsForcedMenuOpen` would be `null`) or if the value over there was already `false`.

Perhaps what would would like to do is to store the calculated value in the localStorage if it's different to what we didn't have any value over there before.

```
  if (cachedIsForcedMenuOpen === null) {
    window.localStorage.setItem(
      STORAGE_KEYS.IS_FORCED_MENU_OPEN,
      String(isForcedMenuOpen)
    );
  }
```

What do you think?

emmenko: Yes good point, it should be `cachedIsForcedMenuOpen === null` 💯

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
```diff
@@ -125,9 +127,16 @@ const useNavbarStateManager = (props: HookProps) => {
     STORAGE_KEYS.IS_FORCED_MENU_OPEN
   );
   const isForcedMenuOpen = isNil(cachedIsForcedMenuOpen)
-    ? null
+    ? isForcedMenuOpenDefaultValue
     : (JSON.parse(cachedIsForcedMenuOpen) as boolean);
 
+  if (!cachedIsForcedMenuOpen) {
```
emmenko, line 133:
```suggestion
  if (isNil(cachedIsForcedMenuOpen)) {
```

emmenko (approved):

PR #3194, opened 2023-09-04
Grid columns fix

chloe0592:
There's an issue with grid columns in new navigation layout in the MC account application.

Setting `grid-template-columns: min-content 1fr` specifies that the first column should be sized based on the minimum content width of its grid items instead of giving it a fixed value. 

This should fix the issue both for the MC account app and for navigation in the App-Kit.

Before
Screenshot 2023-09-04 at 11 52 44

After
Screenshot 2023-09-04 at 11 52 28

### DIFF
--- a/.changeset/twelve-windows-notice.md
+++ b/.changeset/twelve-windows-notice.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': patch
+---
+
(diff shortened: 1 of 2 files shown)


ddouglasz (approved):
Fixes the issue. Works fine from my end. 🙌🏾

emmenko (approved):

PR #3198, opened 2023-09-05
refactor(navbar): consolidate feature flag evaluation

kark:
#### Summary

This PR focuses on refactoring the Navbar component implementation to centralize the feature flag evaluation for rendering either the old or new navbar. It does introduce some code duplication, but the primary goal is to prevent any regressions in the old navbar rendering.


packages/application-shell/src/components/navbar/navbar.styles.js
emmenko, line 1: Should we duplicate the styles as well?

emmenko (commented):
Thanks for working on this! 🙌

packages/application-shell/src/components/navbar/navbar.styles.js
kark: Thank you! That sounds great. Let me make that update.

packages/application-shell/src/components/navbar/navbar-old/navbar.styles.js
```diff
@@ -0,0 +1,11 @@
+const fs = require('fs');
```
CarlosCortizasCT, line 1: I think this would be a good time to learn why we have this setup here.

I don't see any other setup like this one for styles in the repository so it would be great to understand why we need it here.

@emmenko do you have more context about this? 🙏

CarlosCortizasCT (approved):
LGTM 👍

packages/application-shell/src/components/navbar/navbar-old/navbar.styles.js
emmenko: This is due to the fact that the navbar is the only component that uses CSS modules (between App Kit and UI Kit) instead of Emotion, and this poses a problem with bundling.

Therefore, we use a Babel macro to inject the styles into the JS bundle so that we don't need to manage and load external CSS files.

That's why I wanted us to refactor the navbar implementation, so that we can use Emotion and remove this hack.

CarlosCortizasCT: Thanks for the explanation 👍

chloe0592 (approved):
@kark thanks for doing this! 🙏  When this is merged I'll update it the footer PR as well.

emmenko (approved):
Thanks ❤️

PR #3200, opened 2023-09-06
Updating style of navigation footer (part 1)

chloe0592:
Since now we have a [separate folders](https://github.com/commercetools/merchant-center-application-kit/pull/3198) for new and old navigation, I updated the changes from [previous footer PR](https://github.com/commercetools/merchant-center-application-kit/pull/3191) accordingly.

**The requirements done with this PR stay the same as the previous one:**

- Footer should be sticky on the bottom
- Imported new icons for expander
- New color tokens
- New divider
- Hover effect on the expander
- Change breakpoint when the expander is hidden (less than 1200 px)

**Will be done later:**

- Adjust width and spacings of the footer
- Add correct styling to 'Support' (will be done when changing MenuItem style)
- Space added so that the Faded component doesn’t hide the last MenuItem of the side nav

### DIFF
--- a/.changeset/dirty-tomatoes-jump.md
+++ b/.changeset/dirty-tomatoes-jump.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': patch
+---
+
+Updating style of the navigation footer
--- a/packages/application-shell/src/components/navbar/navbar-new/menu-items.tsx
+++ b/packages/application-shell/src/components/navbar/navbar-new/menu-items.tsx
@@ -11,7 +11,6 @@ import { useFlagVariation } from '@flopflip/react-broadcast';
 import type { TFlagVariation } from '@flopflip/types';
 import classnames from 'classnames';
 import { NavLink } from 'react-router-dom';
-import { designTokens as appkitDesignTokens } from '@commercetools-frontend/application-components';
 import type {
   TNormalizedMenuVisibilities,
   TNormalizedPermissions,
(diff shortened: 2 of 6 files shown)


kark (approved):
Cool stuff, LGTM 🙌
Thanks for listing all the upcoming changes!

CarlosCortizasCT:
Sorry for being late but I see what I think it's a wrong background color for the support footer link when running the playground locally and I wanted to double check is it's only me:

image

kark:
> Sorry for being late but I see what I think it's a wrong background color for the support footer link when running the playground locally and I wanted to double check is it's only me

I believe you're correct. I noticed the following item in the description

> * Add correct styling to 'Support' (will be done when changing MenuItem style)

This seems to align with the issue and reusing MenuItem style later on makes sense to me

CarlosCortizasCT:
> > Sorry for being late but I see what I think it's a wrong background color for the support footer link when running the playground locally and I wanted to double check is it's only me
> 
> I believe you're correct. I noticed the following item in the description
> 
> > * Add correct styling to 'Support' (will be done when changing MenuItem style)
> 
> This seems to align with the issue and reusing MenuItem style later on makes sense to me

Thanks for the clarification @kark 🙇 

I mentioned because I think that background color was already green in the previous PR but as far as we know about it and have the intention to fixed in a later PR, that's fine for me. 👍

chloe0592:
@CarlosCortizasCT thanks for pointing that out! 🙏 And as @kark mentioned, I'm working on the styling of MenuItems (including 'Support') in another branch. This will be a PR soon as well. 😄 
Screenshot 2023-09-07 at 09 46 47

PR #3201, opened 2023-09-06
Update `create-mc-app` CLI to support creating custom views

CarlosCortizasCT:
#### Summary

Update `create-mc-app` CLI to support creating custom views.

This is a follow-up of #3175, where we're creating the new custom views template.

#### Description

We're adding a new parameter to `create-mc-app` CLI so users can decide which kind of application they want to create (defaulting to Custom Applications).

The usage for creating a Custom Application does not change, but if you want to create a Custom View, you just need to use the new parameter like this:
```
npx @commercetools-frontend/create-mc-app@latest \
  my-new-custom-view-project \
  --application-type custom-view \
  --template starter-typescript
```

_(This PR will be merged after the one referenced in the summary: #3175)_


packages/create-mc-app/src/cli.ts
```diff
@@ -25,7 +25,12 @@ const run = () => {
   cli
     .command('[project-directory]')
     .usage(
-      '[project-directory]\n\n  Bootstraps a new Custom Application project using one of the predefined templates.'
+      '[project-directory]\n\n  Bootstraps a new Application project using one of the predefined templates.'
+    )
+    .option(
```
CarlosCortizasCT, line 30: We add a new parameter so developers can specify which type of application wants to create.

packages/create-mc-app/src/cli.ts
```diff
@@ -87,7 +93,9 @@ const run = () => {
           tasks.downloadTemplate(taskOptions),
           tasks.updatePackageJson(taskOptions, releaseVersion),
           tasks.updateCustomApplicationConfig(taskOptions),
-          tasks.updateApplicationConstants(taskOptions),
+          options.applicationType === ApplicationTypes.CustomApplication
```
CarlosCortizasCT, line 96: There are no constants to update for Custom Views.

packages/create-mc-app/src/process-options.ts
```diff
@@ -72,6 +74,7 @@ async function processOptions(
   const templateName = options.template;
 
   // Validate options
+  throwIfApplicationTypeIsNotSupported(applicationType);
```
CarlosCortizasCT, line 77: Check that the received value for the new parameter is valid: `custom-application`, `custom-view`.

packages/create-mc-app/src/tasks/download-template.ts
```diff
@@ -26,7 +26,9 @@ function downloadTemplate(options: TCliTaskOptions): ListrTask {
       );
       const templateFolderPath = path.join(
         clonedRepositoryPath,
-        'application-templates',
+        options.applicationType === ApplicationTypes.CustomApplication
```
CarlosCortizasCT, line 29: We update values in a different config file depending on the type of application we want to create.

packages/create-mc-app/src/tasks/update-custom-application-config.ts
```diff
@@ -72,8 +72,12 @@ function updateCustomApplicationConfig(options: TCliTaskOptions): ListrTask {
   return {
     title: 'Updating Custom Applications config',
     task: () => {
+      const configFileName =
```
CarlosCortizasCT, line 75: We update values in a different config file depending on the type of application we want to create.

.github/workflows/main.yml
```diff
@@ -228,6 +229,28 @@ jobs:
       - name: Installing and building template starter
         run: node ./scripts/install-template.mjs
         env:
+          APPLICATION_NAME: custom-application
```
CarlosCortizasCT, line 232: Now we need to use the new script parameter.

emmenko:
@CarlosCortizasCT Have you had a chance to test it out? Is everything fine?

emmenko:
We also discussed to maybe enable this via an environment variable. How do you want to proceed?

CarlosCortizasCT:
> We also discussed to maybe enable this via an environment variable. How do you want to proceed?

I had forgot about it.

So your requirement is that custom views will not be "very" accessible so far, right?
I think I remember we talked about to only allow the cli to create such applications if it's run from an environment with a custom variable, didn't we?

emmenko:
Yeah, so basically just to be a bit more on the safe side that someone installs the template for Custom Views but they won't be usable yet.

CarlosCortizasCT:
> Yeah, so basically just to be a bit more on the safe side that someone installs the template for Custom Views but they won't be usable yet.

Ok, let's see if [this](https://github.com/commercetools/merchant-center-application-kit/pull/3201/commits/d89c9f4e76da14aef70e4ec07aa328c1c31b691e) makes sense for you

emmenko:
Thanks, looks good!

PS: I pushed a commit to fix the fallthrough case (the `break` was missing).

CarlosCortizasCT:
> Thanks, looks good!
> 
> PS: I pushed a commit to fix the fallthrough case (the `break` was missing). [0ebe084](https://github.com/commercetools/merchant-center-application-kit/commit/0ebe0848026b5478538e83f47deb9e168ff21317)

It was intentional as I didn't think the break was necessary, but thanks anyway 🙇

emmenko:
@CarlosCortizasCT Do you want to provide a changeset as well?

kark (approved):
Thank you, I think it works great.
As a side note, it took me a while to figure out we're only supporting `--template starter-typescript` unlike it says in the description 👆🙈

CarlosCortizasCT:
> Thank you, I think it works great. As a side note, it took me a while to figure out we're only supporting `--template starter-typescript` unlike it says in the description 👆🙈

I'm so sorry about it, @kark 😞 

I'll buy you a beer in our next off-site 😃

kark:
> > Thank you, I think it works great. As a side note, it took me a while to figure out we're only supporting `--template starter-typescript` unlike it says in the description 👆🙈
> 
> I'm so sorry about it, @kark 😞
> 
> I'll buy you a beer in our next off-site 😃

Oh, not at all. It's my bad 😅

emmenko (approved):
All good, thanks! 🚀

PR #3204, opened 2023-09-07
refactor(Navbar): submenu position

kark:


kark:
@commercetools/shield-team-fe 
Kindly disregard my review request, I will try to quickly explore using Intersection Observer API to simplify things

kark:
I used some time trying to refactor any static submenu height computations and explored the use of the `IntersectionObserver` but the effects are worse I think (submenu bahavior is bouncy). Perhaps investing more time in this could bring improvements, so personally I'd add that to the tech-debt list.

packages/application-shell/src/components/navbar/navbar-new/navbar.tsx
```diff
@@ -99,21 +98,54 @@ const getIsSubmenuRouteActive = (
...
+      setIsSubmenuAboveMenuItem(false);
+      setSubmenuVerticalPosition(
+        props.isMenuOpen ? menuItemTop : menuItemBottom
+      );
     }
-  }, []);
+  };
+  useLayoutEffect(() => {
+    const observer = new IntersectionObserver(callbackFn, {
+      rootMargin: '-100% 0px 0px 0px',
```
emmenko, line 136: Could you leave a comment explaining the need of this?

packages/application-shell/src/components/navbar/navbar-new/navbar.tsx
```diff
@@ -99,21 +98,54 @@ const getIsSubmenuRouteActive = (
...
-    if (elementRef.current != null) {
-      setTopPosition(elementRef.current.offsetTop);
+  const menuItemBoundingClientRect =
+    menuItemRef.current?.getBoundingClientRect();
+  const menuItemTop = menuItemBoundingClientRect?.top || 0;
+  const menuItemBottom = menuItemBoundingClientRect?.bottom || 0;
+
+  const callbackFn: IntersectionObserverCallback = (entries) => {
+    const [entry] = entries;
+    const doesSubmenuFitViewportBelowMenuItem =
```
emmenko, line 116: Thanks for giving meaningful variable names

emmenko (commented):
Looks pretty cool, thanks! 🙌

packages/application-shell/src/components/navbar/navbar-new/navbar.tsx
```diff
@@ -188,10 +218,19 @@ export const ApplicationMenu = (props: ApplicationMenuProps) => {
...
+          ref={submenuRef}
         >
           {!props.isMenuOpen && (
-            <div className={styles['tooltip-container']}>
+            <div
+              className={styles['tooltip-container']}
+              css={css`
+                ${isSubmenuAboveMenuItem
+                  ? 'bottom'
+                  : 'top'}: -${DIMENSIONS.navMenuItemHeight};
```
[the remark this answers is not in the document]
kark: Thanks, great question, I had to double check myself 😅
I think that's correct I think. This bit styles the `::before` that is needed for the cursor movement from the menu item to subemenu item.
|  When submenu is below | when submenu is above  |
|---|---|
| Screenshot 2023-09-15 at 10 00 00 |  Screenshot 2023-09-15 at 10 00 16 | 

Either way it needs to be moved upwards / downwards respectively with the negative value.

packages/application-shell/src/components/navbar/navbar-new/navbar.tsx
```diff
@@ -99,21 +99,63 @@ const getIsSubmenuRouteActive = (
...
+        setSubmenuVerticalPosition(
+          props.isMenuOpen ? menuItemTop : menuItemBottom
+        );
+      }
+    },
+    [menuItemBottom, menuItemTop, props.isMenuOpen]
+  );
+  useLayoutEffect(() => {
+    const observer = new IntersectionObserver(callbackFn, {
+      rootMargin: '-100% 0px 0px 0px', // we want to observe if the submenu crosses the bottom line of the viewport - therefore we set the root element top margin to -100% of the viewport height
```
ddouglasz, line 140: Makes sense 💯 
Thanks for explaining 🙏🏾

ddouglasz (approved):
Looks good to me. Thank you  🙌🏾

chloe0592 (approved):
This was a really tricky one! I don't see any setbacks, the solution works perfectly.
Great work! 👏 🙇‍♀️

packages/application-shell/src/components/navbar/navbar-new/navbar.tsx
```diff
@@ -111,29 +112,32 @@ export const ApplicationMenu = (props: ApplicationMenuProps) => {
...
+        setIsSubmenuAboveMenuItem(false);
+        setSubmenuVerticalPosition(
+          props.isMenuOpen ? menuItemTop : menuItemBottom
+        );
+      }
+    },
+    [menuItemBottom, menuItemTop, props.isMenuOpen]
+  );
   useLayoutEffect(() => {
     const observer = new IntersectionObserver(callbackFn, {
```
emmenko, line 139: Is it expected to create a new observer instance every time the layout effect runs? 🤔

Looking at our implementation of `useMutationObserver`, we cache the instance to not create it every time: https://github.com/commercetools/ui-kit/blob/73bc8841022ea3b46dc9e2b75d3f9fd3e1d4fff6/packages/hooks/src/use-mutation-observer/use-mutation-observer.ts#L68-L71

kark: I don't think so. I guess that would be a nice enhancement 🙂
I just wanted to double check if your expectation here is to:
- create our own `useIntersectionObserver` hook implementation (based on the implementation of our `useMutationObserver`)
- use [this](https://github.com/jaredLunde/react-hook/tree/master/packages/intersection-observer#readme) polyfill
- simply create `IntersectionObserver` cache without any additional hooks?

emmenko: Just the caching part. You can also use a `ref` to cache it.

emmenko: It also depends on what are you observing. Is it a static element (you then need only 1 observer) or multiple elements, in which case you might need multiple instantiations.

emmenko: Looking at the implementation, it seems that the intersection observer is needed for the `<MenuGroup>` component. Would it make sense to define the observer in that component instead of the parent `<ApplicationMenu>`? Then you have one observer for each rendered `<MenuGroup>`.

kark: Thanks.

> Would it make sense to define the observer in that component instead of the parent `<ApplicationMenu>`

Yes, it's definitely [a potential improvement](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API#:~:text=We%20could%20opt%20to%20monitor%20multiple%20elements%20for%20visibility%20intersection%20changes%20with%20respect%20to%20the%20viewport%20by%20calling%20observer.observe()%20for%20each%20of%20those%20elements%2C%20if%20we%20wanted%20to%20do%20so.), but for now I wanted to avoid dynamically creating all the refs and prop-drilling them. I have some experience with something similar in the LocalizedRichTextEditor, but I'm a bit concerned about our time-frame for this feature. But I can look into that.

emmenko: Thanks. It's fine for me then as long as we are aware of potential implications.

chloe0592: Perhaps we could add that to our technical depth list for after completing phase 1? As long as it's not a critical issue.

chloe0592: We could also just merge this branch for now since it includes a lot of other changes and then take a closer look at this separately.

emmenko: Yes please go ahead, don't feel blocked by this by any means. I was mostly curious about understanding the background a bit more.

PR #3209, opened 2023-09-11
Menu-item clickable area

chloe0592:
In the previous navigation menu, `menu-items` without submenus were only clickable on their corresponding links and not on their containers. 
Also, `menu-items` with submenus were treated as clickable links. 
Now, we're planning to change how this works for both collapsed and expanded state.

**No submenu:**
The `menu-item` should be clickable whole not just the link.

**With submenu:**
The `menu-item` should not be clickable at all. Only the links inside the submenu.

https://github.com/commercetools/merchant-center-application-kit/assets/52276952/ae4cc19f-28b2-4fb5-98a9-73a99b01b5c6


packages/application-shell/src/components/navbar/navbar-new/navbar.mod.css
```diff
@@ -213,11 +211,6 @@
   fill: var(--color-for-navbar-icon-when-active);
 }
 
-.list-item:not(.item__active):hover .icon {
```
chloe0592, line 216: This change will be re-done when we add transitions

packages/application-shell/src/components/navbar/navbar-new/navbar.mod.css
```diff
@@ -343,7 +336,8 @@
 }
 
 .support-menu {
-  padding: var(--spacing-20) var(--spacing-30);
+  padding: 0 var(--spacing-30) var(--spacing-20) var(--spacing-30);
+  height: calc(var(--item-size) + var(--spacing-20));
```
chloe0592, line 340: Fixed the bouncy 'support menu' behaviour

kark (commented):
Please mind that this PR will make the navbar non-accessible / non-navigable by keyboard. If the element is rendered as `<a>` it's within keyboard tabbing sequence and has `tabindex="0"`. 

In this PR only menu items without submenu will be navigable as only those elements are now rendered as links. Not sure if this was intended, so I just wanted to check.

We could make sure that all menu items are navigable by adding `tabindex` attribute.

FilPob:
I think we discussed also with Maxim  that we can postpone the accessibility related stuff for later right? @mmaltsev-ct see the comment from Kacper above

chloe0592:
> Please mind that this PR will make the navbar non-accessible / non-navigable by keyboard. If the element is rendered as `<a>` it's within keyboard tabbing sequence and has `tabindex="0"`.
> 
> In this PR only menu items without submenu will be navigable as only those elements are now rendered as links. Not sure if this was intended, so I just wanted to check.
> 
> We could make sure that all menu items are navigable by adding `tabindex` attribute.
> 
> EDIT: Actually, if you believe that we have more pressing priorities at the moment, and our stakeholders sign off on that, we can delay addressing a11y concerns for now. That would be fine to me. Apart from that, the PR looks good 👍

@kark thanks for the raised point! We agreed we'll be addressing accessibility in the next phase of navigation, but it's still good to let Filip knows. Thank you!

emmenko:
Do we have a list of all the tech debts that we are accumulating? Just so we don't forget and can plan for them.

Thanks 😊

packages/application-shell/src/components/navbar/navbar-new/navbar.mod.css
```diff
@@ -81,12 +85,7 @@
   height: var(--icon-size);
 }
 .list-item .icon > svg *:not([fill='none']) {
-  /* stylelint-disable-next-line csstools/value-no-unknown-custom-properties */
-  fill: var(--color-for-navbar-icon);
-}
-.list-item .icon__active > svg *:not([fill='none']) {
-  /* stylelint-disable-next-line csstools/value-no-unknown-custom-properties */
-  fill: var(--color-for-navbar-icon-when-active);
+  fill: var(--color-surface);
```
chloe0592, line 88: Changed the icon color. The icon also doesn't change color anymore if active state.

chloe0592:
> Do we have a list of all the tech debts that we are accumulating? Just so we don't forget and can plan for them.
> 
> Thanks 😊

@emmenko yes, thank you. I'm keeping track of what we'll get back to after releasing. But we'll have a meeting as well after, just so I can gather more ideas and thoughts.

ddouglasz (approved):
Other than a tiny nit, this looks good to me ✅

packages/application-shell/src/components/navbar/navbar-new/navbar.tsx
```diff
@@ -98,6 +102,32 @@ const getIsSubmenuRouteActive = (
     })
   );
 
+type ItemContainerProps = {
+  labelAllLocales: TLocalizedField[];
+  defaultLabel: string | undefined;
```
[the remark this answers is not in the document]
chloe0592: @ddouglasz thank you, nice catch! 🙌  [4e780ac](https://github.com/commercetools/merchant-center-application-kit/pull/3209/commits/4e780aceadc1d539ed8fe91c80d8136a0ad9e01d)

kark (approved):
Thanks, I think it looks good!
I noticed a little regression - a collapsed menu item doesn't now have a label (I can take care of it later)
|  Before | After  | 
|---|---|
|   |  |

packages/application-shell/src/components/navbar/navbar-new/navbar.tsx
```diff
@@ -98,6 +102,32 @@ const getIsSubmenuRouteActive = (
...
+        <MenuLabel
+          labelAllLocales={props.labelAllLocales}
+          defaultLabel={props.defaultLabel}
+          applicationLocale={props.applicationLocale}
+        />
+      </div>
+    </div>
+  );
+};
+
```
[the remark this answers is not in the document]
chloe0592: @kark yeah, that actually makes more sense, thanks! [de56858](https://github.com/commercetools/merchant-center-application-kit/pull/3209/commits/de56858c605034277098532558d83be713f3d303)

PR #3212, opened 2023-09-13
Smaller navigation UI improvements

chloe0592:
There were few issues:
- Bouncy Support menu
- The menu items moves slightly to the right when expanded
- An active menu item should not have a hover state.
- Added correct icon color
- Added correct title font-weight

https://github.com/commercetools/merchant-center-application-kit/assets/52276952/a67dcd0d-2534-4800-aa35-330976a72cfa


kark (approved):
Thanks 👍

ddouglasz (approved):
🙌🏾

PR #3213, opened 2023-09-13
chore: initial migration to new ESLint flat config format

emmenko:
> [!NOTE]
> This is an exploratory branch to attempt the migration of the ESLint configs to the new ESLint flat config format.
> This is not ready for review yet.

[ESLint has been working on a major refactoring of their config format](https://eslint.org/docs/latest/use/configure/configuration-files-new) which they call "flat" (or simply "new config").

I believe this will be the default format from version 9 onwards, however we can already opt into it.

I also found [this article](https://www.raulmelo.me/en/blog/migration-eslint-to-flat-config) doing a good job in explaining the differences and how you can migrate the config piece-by-piece.

If someone has time and wants to explore this, let's use this branch.

- [ ] Migrate repository ESLint config
- [ ] Migrate `eslint-config-mc-app` (will it be a breaking change or can we be backwards compatible?)


ByronDWall:
@emmenko you are correct that eslint v9+ requires this configuration format.  Updating any app that uses `'@commercetools-frontend/eslint-config-mc-app'` to eslint v9 results in errors such as these:
```
Test suite failed to run
  
      Invalid Options:
      - Unknown options: extensions, ignorePath, useEslintrc, reportUnusedDisableDirectives, resolvePluginsRelativeTo, rulePaths
      - 'extensions' has been removed.
      - 'resolvePluginsRelativeTo' has been removed.
      - 'ignorePath' has been removed.
      - 'rulePaths' has been removed. Please define your rules using plugins.
      - 'reportUnusedDisableDirectives' has been removed. Please use the 'overrideConfig.linterOptions.reportUnusedDisableDirectives' option instead.
  
        at processOptions (node_modules/eslint/lib/eslint/eslint-helpers.js:843:15)
        at new ESLint (node_modules/eslint/lib/eslint/eslint.js:449:34)
```

I'm assuming that updating the eslint config isn't something that is planned in the immediate future?

PR #3225, opened 2023-09-21
refactor(loggers): avoid using deprecated methods to parse client IP addresses

emmenko:
The `request.connection` has been [deprecated in Node since a long time](https://nodejs.org/api/http.html#requestconnection) in favor of `request.socket`.

When accessing `request.ip` or `request.ips` in Express, these methods attempt to resolve the `remoteAddress` using the deprecated `request.connection` (https://github.com/jshttp/forwarded/blob/af3830a175dbe316be3d943f505171c73853eb04/index.js#L46-L50).
Furthermore, the `request.socket` is not guaranteed to be defined as it can be `null` (see https://github.com/nodejs/node/issues/41011 and https://github.com/nodejs/undici/issues/1115).

This PR prevents using `request.ip` or `request.ips` to avoid the outdated implementation and instead attempts to parse these values manually, using the `x-forwarded-for` header.

### DIFF
--- a/.changeset/many-dryers-camp.md
+++ b/.changeset/many-dryers-camp.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-backend/loggers': patch
+---
+
+Avoid using deprecated methods to parse client IP addresses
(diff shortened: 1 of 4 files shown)


Rhotimee (approved):
👍🏽

tdeekens (approved):

PR #3237, opened 2023-09-27
Left navigation transition

ddouglasz:
#### Summary

Left navigation transition

### DIFF
--- a/packages/application-shell/src/components/application-shell/application-shell.spec.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.spec.js
@@ -126,7 +126,7 @@ const renderApp = (ui, options = {}) => {
     // Wait for the loading navbar to disappear. Instead of using `waitForElementToBeRemoved`,
     // which seems not stable enough, we wait to find the "Support" link, which is present
     // when the navbar is loaded.
-    await screen.findByText('Support');
+    await screen.findByRole('navigation');
   };
 
   return {
--- a/packages/application-shell/src/components/application-shell/application-shell.spec.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.spec.js
@@ -124,7 +124,7 @@ const renderApp = (ui, options = {}) => {
   const waitForLeftNavigationToBeLoaded = async () => {
     await findByLeftNavigation();
     // Wait for the loading navbar to disappear. Instead of using `waitForElementToBeRemoved`,
-    // which seems not stable enough, we wait to find the "Support" link, which is present
+    // which seems not stable enough, we wait to find the "navigation" role, which is present
     // when the navbar is loaded.
     await screen.findByRole('navigation');
   };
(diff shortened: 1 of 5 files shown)


ddouglasz:
@FilPob I faced some challenges with the transition for when the sidebar collapses. The transition works fine for the sidenav when it collapses but since the menu item title is basically made to disappear, there is hardly a way to make the transition happen before the item disappears, thus causing a small glitch in the collapse transition. We took a look at it and we concluded to ask you what you think about it. 

Currently:

https://github.com/commercetools/merchant-center-application-kit/assets/19975842/8d320bfc-19ce-451f-9fcb-a71dc37c9a05



Slowed down version sp you can see how the navbar transitions but the item title would not be captured:

https://github.com/commercetools/merchant-center-application-kit/assets/19975842/72dc41d0-1895-43ba-8e02-a25634a8f469

FilPob:
@ddouglasz thanks for the effort! yeah it is a little glitchy, but nothing too critical that should block the release. I think we can live with that in the phase 1

ddouglasz:
> @ddouglasz thanks for the effort! yeah it is a little glitchy, but nothing too critical that should block the release. I think we can live with that in the phase 1

Alright, great. My thought exactly, we could dig deeper in phase 2. 
Thank you.

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -126,7 +126,7 @@ const renderApp = (ui, options = {}) => {
     // Wait for the loading navbar to disappear. Instead of using `waitForElementToBeRemoved`,
     // which seems not stable enough, we wait to find the "Support" link, which is present
     // when the navbar is loaded.
-    await screen.findByText('Support');
+    await screen.findByRole('navigation');
   };
```
kark, line 130: Perhaps we could edit the comment as well.

packages/application-shell/src/components/navbar/navbar-new/navbar.mod.css
```diff
@@ -15,6 +15,24 @@
...
+}
+
+@keyframes hidden {
+  from {
+    opacity: 1;
+  }
+  to {
+    opacity: 0;
+  }
+}
```
kark, line 34: I don't think it's used anywhere

kark (approved):
Big thanks, Douglas! Great job. It looks like we're all set to proceed 🙌
Just a friendly reminder to make sure the changeset is incorporated

packages/application-shell/src/components/navbar/navbar-new/navbar.mod.css
ddouglasz: Thank you,
All cleaned up here: https://github.com/commercetools/merchant-center-application-kit/pull/3237/commits/8dc2680475702e90085f24a037efa476469e5290

chloe0592 (approved):
@ddouglasz thank you for your efforts! Looks good to me, we can work on the details in the phase 2. ☺️

PR #3238, opened 2023-09-27
fix: playground app in dev mode

kark:
#### Summary

I noticed some problems with running the playground app in dev mode.
Screenshot 2023-09-27 at 14 35 32
Screenshot 2023-09-27 at 14 35 42
Local production mode works fine.

A while back, we downgraded vite version https://github.com/commercetools/merchant-center-application-kit/pull/3152.  
It appears that the root cause (in `fsevents` package) has been resolved now, allowing us to upgrade Vite and, as a result, fix the playground app.


emmenko (approved):
Thanks

PR #3241, opened 2023-09-28
fix: disable route permission check for `<CustomViewDevHost>`

emmenko:
When running the Custom View template locally, the application renders as a Custom Application (using the `<ApplicationShell>`). However, we need to disable the route permissions check as the computed permissions are based on the `customViewId` rather than the `entryPointUriPath`.


CarlosCortizasCT (approved):
Nice catch! 

Thanks a lot 🙇

Rhotimee (approved):
👍🏽

PR #3242, opened 2023-09-28
New CustomViewsSelector component

CarlosCortizasCT:
#### Summary

New component to implement the Custom Views selection.

#### Description

The main goal of this PR is introducing the new `CustomViewsSelector` component which will be used from MC pages to allow users to trigger the Custom Views they have configured.

We added that new component and refactored all page components (`InfoMainPage`, `FormDetailPage`,...) so using the former it's just a matter of passing a new prop (**customViewLocatorCode**) to the page components.

The `CustomViewsSelector` component takes that locator code as a prop and it will send a request to the Settings API to fetch the available Custom Views for that locator (within the project and logged user constraints).
If there are any available, the component will render a new row element below the top bar with a selector so users can trigger them.

Demo example:
image

There are a lot more files updated in this PR but this is due to a new situation with this new component which required refactor and move some other components/hooks between packages.
This new component is the first one in the `application-components` package which needs to interact with the backend and most of the code to handle that was living in the `application-shell` package. The problem is that this package relies on `application-components` one (where the new component lives) but the opposite can't happen (circular dependencies). So we decided to move that logic to the `application-shell-connectors` package, which is then used by the other two.

Basic new dependencies schema:
image

This refactoring was addressed in [this PR](https://github.com/commercetools/merchant-center-application-kit/pull/3245) in case you want to review it in isolation.

The new component has been included in the Playground application.
Also, we have a [follow-up PR](https://github.com/commercetools/test-data/pull/384) in progress to include new visual regression tests to include it.


packages/application-components/package.json
```diff
@@ -30,6 +30,9 @@
     "@babel/runtime": "^7.22.15",
     "@babel/runtime-corejs3": "^7.22.15",
     "@commercetools-frontend/actions-global": "22.8.1",
+    "@commercetools-frontend/application-config": "22.8.1",
+    "@commercetools-frontend/application-shell": "22.8.1",
```
emmenko, line 34: ⚠️ This introduces a circular dependency. What is required from app shell in the components?

CarlosCortizasCT: Yes. This is something I already noted to discuss in our next sync meeting.

emmenko (approved):
Good to go from my side 🚀

packages/application-components/src/components/custom-views/custom-views-selector/custom-views-selector.tsx
```diff
@@ -0,0 +1,113 @@
...
+      <Container>
+        <Constraints.Horizontal max="scale">
+          <Spacings.Inline
+            scale="m"
+            justifyContent="flex-end"
+            alignItems="center"
+          >
+            <Spacings.Inline scale="xs" alignItems="center">
+              <SidebarCollapseIcon size="medium" color="neutral60" />
+              <Text.Detail tone="secondary">Custom Views:</Text.Detail>
```
Rhotimee, line 73: What do you think about using messages here and the other button labels?

Rhotimee (approved):
Looks great!! 

Just a minor feedback.

packages/application-components/src/components/custom-views/custom-views-selector/custom-views-selector.tsx
emmenko: Thanks! https://github.com/commercetools/merchant-center-application-kit/pull/3242/commits/ab9c6881d9e3fc3b9de4456d587995ae09f64882

PR #3245, opened 2023-10-02
Extract logic from `application-shell` to `application-shell-connector` package

CarlosCortizasCT:
#### Summary

Extract logic from `application-shell` to `application-shell-connector` package

#### Description

In the source PR (#3242) we included the new `CustomViewSelector` component in the `application-shell` package but this leads to a circular dependency between that one and the `application-components` package.
We are ok with the former depending on the latter, but not the other way around.

The problem is that this new component (`CustomViewSelector`) is going to be used from all page components (which live in the `application-components` packages) but it is the first one of this kind which also needs to interact with the backend API (fetch available Custom Views for an specific view locator). This means it needs access to Apollo Graphql machinery, which is implemented in the `application-shell` package.

To avoid this unwanted dependency (circular) flow, we propose to move most of the Apollo Graphql machinery to the `application-shell-connectors` packages, which is independent and can be used from both the other two (`application-shell` and `application-components`).

The initial idea was to move the Apollo client factory function and the custom hooks to interact with the queries and mutations (`useMcQuery` and `useMcMutation`), but we need to drag some other modules which those ones depend on.

So basically in this PR we are:
* Moving most of the stuff related to interacting with the backend using the Apollo client from the `application-shell` package to the `application-shell-connectors` one.
* Moving both the new `CustomViewSelector` component and the (already existing) `CustomViewSelector` (the former depends on this one from `application-shell` to the `application-components` package.

Here is a schema of the proposal implemented in this PR:

image


emmenko (approved):
Thanks for the refactor!

I pushed a couple of commits to move the constants into the constants package and keep the exported types from app shell for backwards compatiblity (and clean up the deps a bit).

CarlosCortizasCT:
> Thanks for the refactor!
> 
> I pushed a couple of commits to move the constants into the constants package and keep the exported types from app shell for backwards compatiblity (and clean up the deps a bit).

Thanks a lot for your help! 🙇

tdeekens (approved):
Nice refactor. Makes total sense. Thanks for the diagram in the description!

PR #3246, opened 2023-10-03
Include `CustomViewSelector` component in page components VRTs

CarlosCortizasCT:
#### Summary

Include `CustomViewSelector` component in page components VRTs


packages/application-shell/src/index.ts
```diff
@@ -37,6 +39,12 @@ export {
   entryPointUriPathToResourceAccesses,
 } from '@commercetools-frontend/application-config/ssr';
 
+// This is for internal purposes only.
+// Please do not rely on it as it might change at any time.
+export const _graphqlQueries = {
+  FetchCustomViewsQuery,
```
CarlosCortizasCT, line 45: Exporting this query so we don't need to duplicate it in the visual testing application.

PR #3247, opened 2023-10-04
fix(Navbar): performance issues

kark:
#### Summary
Create an instance of IntersectionObserver for each menu item only once (at first render) and reuse between re-renders.

We're comparing apples and oranges, but still the profiling results suggest it's at least making things a tad better.
I have no certainty it's the ultimate fix.
At least some garbage collection takes place over time and there are way less JS event listeners.
 
| Current prod (Welcome app) <br>Old Navbar <br> (that's our baseline) | Current staging (Welcome app) <br >New Navbar  |  Playground app in local prod mode (15 menu items) <br> New Navbar with changes from this PR |
|---|---|---|
|  Screenshot 2023-10-04 at 15 29 51 | Screenshot 2023-10-04 at 15 29 44 |  Screenshot 2023-10-04 at 15 29 36 |

### DIFF
--- a/packages/application-shell/src/components/navbar/navbar-new/navbar.tsx
+++ b/packages/application-shell/src/components/navbar/navbar-new/navbar.tsx
@@ -9,6 +9,7 @@ import {
 } from 'react';
 import { css } from '@emotion/react';
 import classnames from 'classnames';
+import snakeCase from 'lodash/snakeCase';
 import { FormattedMessage } from 'react-intl';
 import { matchPath, useLocation } from 'react-router-dom';
 import type { RouteComponentProps } from 'react-router-dom';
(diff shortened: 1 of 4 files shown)


packages/application-shell/src/components/navbar/navbar-new/navbar.tsx
```diff
@@ -103,57 +104,61 @@ const getIsSubmenuRouteActive = (
...
-    },
-    [menuItemBottom, menuItemTop, props.isMenuOpen]
-  );
   useLayoutEffect(() => {
-    const observer = new IntersectionObserver(callbackFn, {
+    observerRef.current = new IntersectionObserver(callbackFn, {
       rootMargin: '-100% 0px 0px 0px', // we want to observe if the submenu crosses the bottom line of the viewport - therefore we set the root element top margin to -100% of the viewport height
     });
+    return () => observerRef.current?.disconnect();
+  }, []);
```
kark, line 149: The main idea behind this PR is to cache the `IntersectionObserver` instance and create it only at the first render. This idea was brought up [here](https://github.com/commercetools/merchant-center-application-kit/pull/3204#discussion_r1328477809).

packages/application-shell/src/components/navbar/navbar-new/navbar.tsx
```diff
@@ -103,57 +104,61 @@ const getIsSubmenuRouteActive = (
...
-  const menuItemTop = menuItemBoundingClientRect?.top || 0;
-  const menuItemBottom = menuItemBoundingClientRect?.bottom || 0;
+  const menuItemIdentifier = snakeCase(props.menu.key);
+
+  const callbackFn: IntersectionObserverCallback = useCallback((entries) => {
+    const menuItemBoundingClientRect = document
+      .querySelector(`[data-menuitem="${menuItemIdentifier}"]`)
+      ?.getBoundingClientRect();
+    const menuItemTop = menuItemBoundingClientRect?.top || 0;
+    const menuItemBottom = menuItemBoundingClientRect?.bottom || 0;
```
kark, line 119: Instead of using a ref now we're using `querySelector` to find the relevant menu item element. This ensures the ref used in callback is not stale when the `IntersectionObserver` instance is cached.
Many thanks to @CarlosCortizasCT for the idea 🙏

emmenko (approved):
Thanks for looking into this and finding a viable solution! 💯

chloe0592 (approved):
Thanks so much @kark! 🙇‍♀️

tdeekens (approved):

PR #3248, opened 2023-10-04
Fix Navigation menu E2E tests

CarlosCortizasCT:
#### Summary

There were a couple of tests failing after enabling the new navigation menu.

#### Description

The trickiest one is was a test expecting the submenu links to be visible after hovering a navigation menu item (when collapsed) since we now control that behaviour with CSS and [Cypress does not support real `hover` events](https://docs.cypress.io/api/commands/hover#__docusaurus_skipToContent_fallback).

I've tries the alternatives suggested in their docs but had no luck so I decided to use [this package](https://github.com/dmtrKovalenko/cypress-real-events/) which adds a some custom commands for interacting with DOM elements using real events.

Since we will need to also update the E2E tests in consuming apps, I've added a couple of custom commands to our cypress exported package so we use a standard API to deal with this behaviour and it's under our control.

### DIFF
--- a/packages/cypress/src/add-commands/index.ts
+++ b/packages/cypress/src/add-commands/index.ts
@@ -1,3 +1,5 @@
+import { Matcher } from '@testing-library/dom';
+import 'cypress-real-events';
 import {
   loginByForm,
   loginByOidc,
@@ -32,4 +34,16 @@ Cypress.Commands.add('loginByOidc', (commandOptions: CommandLoginOptions) => {
   );
 
   loginByOidc(commandOptions);
+});
+
+Cypress.Commands.add('hover', { prevSubject: true }, (subject: JQuery) => {
+  cy.wrap(subject).realHover();
+});
+
+Cypress.Commands.add('navigationMenuHover', (textMatcher: Matcher) => {
+  cy.findByTestId('left-navigation')
+    .findByText(textMatcher)
+    .parents('[role="menuitem"]')
+    .first()
+    .realHover();
 });
--- a/cypress/e2e/playground/application-shell.cy.ts
+++ b/cypress/e2e/playground/application-shell.cy.ts
@@ -38,6 +38,7 @@ describe('when user is authenticated', () => {
 
 describe('navigation menu', () => {
   beforeEach(() => {
+    cy.viewport(1250, 800);
     cy.loginToMerchantCenter({
       entryPointUriPath: ENTRY_POINT_APP_KIT_PLAYGROUND,
       initialRoute: URL_APP_KIT_PLAYGROUND,
--- a/packages/cypress/package.json
+++ b/packages/cypress/package.json
@@ -41,6 +41,7 @@
     "@commercetools-frontend/application-shell": "22.8.2",
     "@manypkg/get-packages": "1.1.3",
     "@types/semver": "^7.5.1",
+    "cypress-real-events": "1.10.3",
     "semver": "7.5.2",
     "uuid": "9.0.0"
   },
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1899,6 +1899,9 @@ importers:
       '@types/semver':
         specifier: ^7.5.1
         version: 7.5.1
+      cypress-real-events:
+        specifier: 1.10.3
+        version: 1.10.3(cypress@12.12.0)
       semver:
         specifier: 7.5.2
         version: 7.5.2
@@ -6360,6 +6363,7 @@ packages:
   /@colors/colors@1.5.0:
     resolution: {integrity: sha512-ooWCrlZP11i8GImSjTHYHLkvFDP48nS4+204nGb1RiX/WXYHmJA2III9/e2DWVabCESdW7hBAEzHRqUn9OUVvQ==}
     engines: {node: '>=0.1.90'}
+    requiresBuild: true
 
   /@commercetools-community-kit/i18n@0.3.0:
     resolution: {integrity: sha512-EPxsJaNXUYwY6Q9gjyKOymzhlSNAhTur415yX25d6S/GA6sG0yWx+/WRKCr/AqSKOskv8SGpj8+tMyPFO6AeNQ==}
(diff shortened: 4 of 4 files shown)


packages/cypress/src/add-commands/index.ts
```diff
@@ -33,3 +35,15 @@ Cypress.Commands.add('loginByOidc', (commandOptions: CommandLoginOptions) => {
 
   loginByOidc(commandOptions);
 });
+
+Cypress.Commands.add('hover', { prevSubject: true }, (subject: JQuery) => {
```
CarlosCortizasCT, line 39: This allows to hover an arbitrary element.

Example:
```
cy.findByText('<text_matcher>').hover();
```

tdeekens (approved):
🤯

packages/cypress/src/add-commands/index.ts
```diff
@@ -1,3 +1,5 @@
+import { Matcher as TMatcher } from '@testing-library/dom';
+import 'cypress-real-events';
```
kark, line 2: I suppose all the type checking errors come from adding this here. I noticed though that type checking works fine when we add this in `cypress/support/commands.ts` as stated [in docs](https://github.com/dmtrKovalenko/cypress-real-events#installation). Obviously we need to import in the package but I'm not sure how.

packages/cypress/src/add-commands/index.ts
```diff
@@ -33,3 +35,15 @@ Cypress.Commands.add('loginByOidc', (commandOptions: CommandLoginOptions) => {
 
   loginByOidc(commandOptions);
 });
+
+Cypress.Commands.add('hover', { prevSubject: true }, (subject: JQuery) => {
+  cy.wrap(subject).realHover();
+});
+
+Cypress.Commands.add('navigationMenuHover', (textMatcher: Matcher) => {
```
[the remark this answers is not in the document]
CarlosCortizasCT: Not super sure about the name of this command.
Maybe something like `showNavigationMenuItemSublinks` would be better?

kark (approved):
Thank you for the fix 🎉

packages/cypress/src/add-commands/index.ts
emmenko: That's more for the consumers, here we are providing commands

packages/cypress/src/add-commands/index.ts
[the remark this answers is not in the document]
emmenko: Maybe `showNavigationSubmenuItems`?

emmenko (commented):
Don't forget the changeset ;)

packages/cypress/src/add-commands/index.ts
CarlosCortizasCT: I'm sorry but I don't understand what we need to do here.
We are using the `cypress-real-events` within this package (using the `realHover` command) so I think we need it.

> That's more for the consumers, here we are providing commands
Do you mean this should be a peer dependency?

emmenko:
@CarlosCortizasCT Do you need help with the types?

CarlosCortizasCT:
> Don't forget the changeset ;)

I was working on it.
Addded here: 79c3ceaa0d3a93744d1ef1db91c49eb4fff3f45b

packages/cypress/add-commands/index.d.ts
```diff
@@ -25,5 +25,27 @@ declare namespace Cypress {
...
+    hover(): Chainable<Subject>;
+
+    /**
+     * Triggers a `hover` (real) event on the navigation menu item found using the given text matcher.
+     * It will make the submenu links panel to appear.
+     *
+     * @example
+     *    cy.navigationMenuHover('navigation_item_menu_text_matcher')
+     *
+     * @deprecated Use the more generic {@link cy.loginToMerchantCenter} command as it automatically detects which login mechanism to use.
```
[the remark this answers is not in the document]
CarlosCortizasCT: Thanks!

Removed here: dadfc2d490eaf51fd344aa8579bfc781bfc117b0

chloe0592 (approved):
Thanks so much for your help @CarlosCortizasCT! 🙇‍♀️ 🙇‍♀️

packages/cypress/src/add-commands/index.ts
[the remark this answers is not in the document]
CarlosCortizasCT: I like it 👍 

Updated here: 476c89a239d7c68edb1a0fcfcf7d4a5b3adb82e8

CarlosCortizasCT:
> @CarlosCortizasCT Do you need help with the types?

Yes I do.

packages/cypress/src/add-commands/real-hover.ts
```diff
@@ -0,0 +1,275 @@
...
+ * https://github.com/dmtrKovalenko/cypress-real-events/blob/develop/src/commands/realHover.ts
+ *
+ * However, due to known issues with conflicting types between Cypress and Jest, importing the `cypress-real-events`
+ * package here will cause such issues with TypeScript as our `@commercetools-frontend/cypress` package
+ * is checked and built together with all other packages and not in isolation.
+ * See https://docs.cypress.io/guides/tooling/typescript-support#Clashing-types-with-Jest.
+ *
+ * Therefore, we are porting here the implementation of `realHover` to avoid importing it from the
+ * original package `cypress-real-events`.
+ */
```
emmenko, line 12: As explained here, we can't import the `cypress-real-events` package directly due to conflicting Cypress vs Jest types.

We basically have 2 options:
1. Port the implementation of the commands we need (e.g. `realHover`) as I did here.
2. Assume that the `realHover` command is "available" but it must be defined / imported into the Cypress setup by the consumers (so not part of the `@commercetools-frontend/cypress` package).

Option 1 will require us to maintain the implementation. Option 2 requires less maintenance but the consumer must import the package `cypress-real-events`.

@CarlosCortizasCT @tdeekens  Any preferences?

CarlosCortizasCT: I don't like us maintaining more code 😅 but I prefer that over giving more work to consumers in this case.
So I'd would keep the implementation locally.

_(Hopefully we won't need to touch it often)_

emmenko:
Hey @chloe0592 

I wanted to wait for @tdeekens 's feedback. Unless there is a need to release this asap, I would wait a bit.

Would that be ok for you?

chloe0592:
> Hey @chloe0592
> 
> I wanted to wait for @tdeekens 's feedback. Unless there is a need to release this asap, I would wait a bit.
> 
> Would that be ok for you?

Yeah, sure let's wait for Tobi's approval first.

packages/cypress/src/add-commands/real-hover.ts
tdeekens: I would like to co-locate things where it's lifecycle occurs. In this case the menu and the ability to test it and how to should be co-located and come from App Kit. I would assume that this code only changes if Cypress changes its APIs (e.g. breaking change) which they have historically not done often. If we do not add it here we have to add it to many repositories and maintain it there while keep cross-repo consumption and test-setup in sync has been costly I think having the "mess" here is our best option.

tdeekens (approved):

PR #3249, opened 2023-10-05
Menu item as a clickable link

chloe0592:
In previous work we disabled clicking on the menu item parent when it contained a submenu. 

This could lead to users reporting it as a bug, since it's not straightforward why some menu items are clickable and some not. 
But more importantly custom apps  can have different links from the parent menu item and the first submenu child, so it's a potentially a breaking change.

Therefore, we decided to have all menu items as links and clickable.

### DIFF
--- a/.changeset/big-mayflies-bake.md
+++ b/.changeset/big-mayflies-bake.md
@@ -2,4 +2,4 @@
 '@commercetools-frontend/application-shell': patch
 ---
 
-Navigation menu item is again clickable even if it has submenu.
+First level navigation menu items are clickable again even it they have submenu items.
--- a/.changeset/big-mayflies-bake.md
+++ b/.changeset/big-mayflies-bake.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': patch
+---
+
+Navigation menu item is again clickable even if it has submenu.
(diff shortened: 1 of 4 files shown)


emmenko (approved):
Thanks!

> But more importantly custom apps can have different links from the parent menu item and the first submenu child, so it's a potentially a breaking change.

Actually this is a bad behavior of our internal apps to have a duplicate link for the main item (for example `Product list`).
We discourage this kind of pattern (see note in the docs: https://docs.commercetools.com/custom-applications/api-reference/application-config#submenulinksuripath) and should eventually fix it also in our internal apps.

chloe0592:
> Actually this is a bad behavior of our internal apps to have a duplicate link for the main item (for example `Product list`). We discourage this kind of pattern (see note in the docs: https://docs.commercetools.com/custom-applications/api-reference/application-config#submenulinksuripath) and should eventually fix it also in our internal apps.

Oh, okay, thanks! That's really good to know and then it's certain we need to have all menu items as links regardless if being internal or custom apps.

kark (approved):
🎉

.changeset/big-mayflies-bake.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': patch
+---
+
+Navigation menu item is again clickable even if it has submenu.
```
CarlosCortizasCT, line 5: **nit**:
```suggestion
First level navigation menu items are clickable again even it they have submenu items.
```

CarlosCortizasCT (approved):
Looks good to me.

Thanks! 👍

PR #3256, opened 2023-10-05
fix(Navbar): wrong submenu position on expand button click

kark:
#### Summary

Apologies, In #3247 a bug was introduced 🙈.  We need to make lifecycle of the Navbar dependent on `props.isMenuOpen`. Otherwise the `IntersectionObserver` callback is stale.

You can see it [here](https://mc.europe-west1.gcp.escemo.com/almond-40/welcome) after clicking expand button.


CarlosCortizasCT (approved):
Thanks. 👍

tdeekens (approved):

PR #3258, opened 2023-10-06
fix(navbar): crashing Chrome tab with DevTools open

kark:
#### Summary

The issue was caused by [`aria-owns`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-owns) attribute value not referencing a valid id in DOM 🤯


emmenko (approved):
Wow, really?! 🤯

PR #3262, opened 2023-10-10
fix: fetch api polyfill

abreu-ct:
This PR adds `Request` and `Response` to the Fetch API polyfill. Unfortunately `jsdom` [does not provide a way](https://github.com/jsdom/jsdom/issues/1724) to use Node native Fetch API, I used the same approach for the `Headers` class and also exported `Request` and `Response` from `node-fetch`.

This PR will allow running jest tests with `jsdom` env that relies on the Fetch API fixing https://github.com/commercetools/merchant-center-frontend/pull/15461

#### Notes:
In [setup-test-framework.js#L2](https://github.com/commercetools/merchant-center-application-kit/blob/main/packages/jest-preset-mc-app/setup-test-framework.js#L2) we are using the `unfetch/polyfill` to export the Fetch function and relying on `node-fetch` to export `Headers`, `Request` and `Response`. I tried to change `unfetch/polyfill` polyfill but unfortunately some tests in `merchant-center-frontend` started to fail.

### DIFF
--- a/.changeset/nine-cows-add.md
+++ b/.changeset/nine-cows-add.md
@@ -1,0 +1,5 @@
+---
+"@commercetools-frontend/jest-preset-mc-app": patch
+---
+
+Adds `Request` and `Response` from `node-fetch` as globals to the Jest preset.
--- a/packages/jest-preset-mc-app/setup-tests.js
+++ b/packages/jest-preset-mc-app/setup-tests.js
@@ -8,6 +8,8 @@ global.window.app = {
 
 window.MutationObserver = MutationObserver;
 global.Headers = global.Headers || require('node-fetch').Headers;
+global.Request = global.Request || require('node-fetch').Request;
+global.Response = global.Response || require('node-fetch').Response;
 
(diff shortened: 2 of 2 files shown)


tdeekens (approved):

emmenko (approved):
Thanks!

harm-meijer:
@tdeekens @emmenko @abreu-ct  I understand why someone would add a polyfill for fetch since that is what a quick google would suggest but that polyfill isn't needed when you use node 18 and would not be needed in a browser for quite some time. The root cause of this issue is because the jsdom environment is [outdated](https://github.com/jsdom/jsdom/issues/1724) and a [custom profile](https://github.com/jsdom/jsdom/issues/1724#issuecomment-1446858041) representing a more up to date jsdom environment would have solved this better in my opinion.

emmenko:
Thanks for the feedback. That sounds indeed like a better solution.

I'll take a note for us to look into that as a follow up.

tdeekens:
Thanks make sense. Let's follow-up and first unblock the team :)

PR #3255, opened 2023-10-05
`CustomViewsSelctor` component RTL tests

CarlosCortizasCT:
Follow up of #3242 

Add basic tests for the `<CustomViewsSelctor>` component.

PR #3264, opened 2023-10-11
test(components): define apollo provider

emmenko:
Follow up of #3255 

Using the `@commercetools-frontend/application-shell` as a dev dependency still causes a warning for a circular dependency (app-shell <--> app-components), even though we only use the test utils.

So let's get rid of the dependency again. However, we need to ensure the correct test-utils setup in the application-components package is correct. At the moment we only need to add the `<ApolloProvider>` in order to handle the graphql requests.

Apologies for the back and forth.

### DIFF
--- a/packages/application-components/package.json
+++ b/packages/application-components/package.json
@@ -71,7 +71,6 @@
   },
   "devDependencies": {
     "@apollo/client": "3.7.14",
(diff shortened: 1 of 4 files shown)


CarlosCortizasCT (approved):
Thanks!

PR #3266, opened 2023-10-12
Fix `renovate.json` and order of `packageRules`

tdeekens:
#### Summary

As elsewhere applying similar fixes here:

1. The order matters for `packageRules`. You go from not important to important as they're evaluated in order and subsequent rules overwrite prior rules (Renovate docs)
2. `sourceUrlPrefixes` has not worked well elsewhere. We can use `matchSourceUrls` instead which works in all other repos
3. Isolate the update of our runtime in Node.js

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -31,6 +31,12 @@
       "groupName": "all svgr packages"
     },
     {
+      "groupName": "all Node.js updates",
+      "matchPackageNames": ["node"],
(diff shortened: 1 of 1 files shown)


emmenko (approved):

tdeekens:
👏🏼 finally works as intended:

CleanShot 2023-10-12 at 09 38 18@2x

CarlosCortizasCT (commented):
Many thanks!

PR #3273, opened 2023-10-17
feat: megamenu redesign docs-kit 22.8.2

gabriele-ct:

PR #3274, opened 2023-10-17
chore: linkinator fix

kark:
#### Summary

https://github.com/commercetools/merchant-center-application-kit/pull/3273#pullrequestreview-1681997065


CarlosCortizasCT (approved):
Many thanks! 🙇

kark:
> The config is correct (same as we use in the docs repo).
> 
> If the check fails is usually because there are indeed wrong links.

Hmm 🤔, I'm lost now.
I wonder why the `Documentation workflow` passes here with those links checked:
Screenshot 2023-10-17 at 13 04 20
https://github.com/commercetools/merchant-center-application-kit/actions/runs/6546016959/job/17775827479
as opposed to the original PR:
https://github.com/commercetools/merchant-center-application-kit/actions/runs/6545409731

I can close it, but I think the changes are correct

CarlosCortizasCT:
I thought it made sense the change since the name of the props seemed to be inverted:
```
urlRewriteReplace: <the_replacement_value>
urlRewriteSearch: <what_we_want_to_search_for_replace>

urlRewriteReplace: https://appkit-sha-${{ github.sha }}.commercetools.vercel.app
urlRewriteSearch: https://docs.commercetools.com
```

but it's weird is not working now so, after thinking again, now I believe this is due to new microsites being referenced in the new menu which we need to add to our [exclusions list](https://github.com/commercetools/merchant-center-application-kit/blob/main/.linkcheck/skip_list.txt#L14).

Apart from that, I find the name of the config props confusing.

.linkcheck/skip_list.txt
```diff
@@ -15,7 +15,11 @@ https://www.npmjs.com
 
 \/.+\/api$
 \/.+\/api\/
+\/.+\/checkout
+\/.+\/conntect
```
emmenko, line 19:
```suggestion
\/.+\/connect
```

emmenko (commented):
Thanks!

.linkcheck/skip_list.txt
kark: Thank you!

emmenko (approved):
Thanks!

kark:
Thank you @CarlosCortizasCT 🙌

CarlosCortizasCT:
So the problem is that we were applying the configuration in the wrong place.

The [skip_list.txt](https://github.com/commercetools/merchant-center-application-kit/blob/main/.linkcheck/skip_list.txt) file is not actually used. As we can see in the [CI configuration](https://github.com/commercetools/merchant-center-application-kit/blob/main/.github/workflows/docs.yml#L146), we're using [this one](https://github.com/commercetools/merchant-center-application-kit/blob/main/website/linkinator.config.json) instead.

I removed the `skip_list.txt` file so we don't get this confusion in the future.

.linkcheck/skip_list.txt
kark, line 1: I think we still use it when we run linkinator locally https://github.com/commercetools/merchant-center-application-kit/blob/57a9f3a38d292551ef172f7eeaeaf904633fe42a/website/package.json#L8

PR #3280, opened 2023-10-18
Fix Custom Views locator hook

CarlosCortizasCT:
#### Summary

Fix Custom Views locator hook

#### Description

There's an issue in the hook we use to calculate the Custom View locator for tab page components based on the URL.
It's configured to match URLs without being exact, which does not allow for using nested routes, which is something we use in MC frontend for tabs URLs.


emmenko (approved):
💯

emmenko:
Don't forget the changeset

PR #3287, opened 2023-11-02
feat(app-shell): allow to enable StrictMode through shell

### LINKED ISSUE #3286: Add support for `StrictMode` through the `ApplicationShell`

The App Kit so far does not natively support [StrictMode](https://react.dev/reference/react/StrictMode). Applications being built with it have to "opt-into" `StrictMode` within the application's code. Consumer would achieve this by wrapping the `ApplicationShell` within a `StrictMode` element as in:

```jsx
const EntryPoint = () => (
  <StrictMode>
    <ApplicationShell
      render={() => (
        <Application />
      )}
    />
  </StrictMode>
);
```

`StrictMode` itself can be more natively supported by the `ApplicationShell` using a `prop` which then conditionally wraps the entire tree into `<StrictMode/>`. This would allow App Kit to possibly perform other optimizations in `StrictMode` in the future.

```jsx
const EntryPoint = () => (
  <ApplicationShell
    enableStrictMode
    render={() => (
      <Application />
    )}
  />
);
```

This approach would also help to in the future enable StrictMode by default for Custom Applications by flipping the default of `enableStrictMode` to `true` while it's `false` initially. Then as another step one could remove `enableStrictMode` entirely and add a `disableStrictMode` for legacy applications.

**Describe alternatives you've considered**

`StrictMode` is designed to be applied to parts of an application. While you can not opt-out of it again if a parent opted-in. This means that applications not fully compatible with StrictMode can't use the prop but have to work their component tree upwards to support StrictMode everywhere.

**Additional context**

tdeekens:
#### Summary

Closes #3286

### DIFF
--- a/packages/application-shell/src/components/application-shell/application-shell.tsx
+++ b/packages/application-shell/src/components/application-shell/application-shell.tsx
@@ -58,9 +58,9 @@ type TStrictModeEnablementProps = {
 
 const StrictModeEnablement = (props: TStrictModeEnablementProps) => {
   if (props.enableReactStrictMode) {
-    return <StrictMode>{children}</StrictMode>;
+    return <StrictMode>{props.children}</StrictMode>;
   } else {
-    return <>{children}</>;
+    return <>{props.children}</>;
   }
 };
 
--- a/packages/application-shell/src/components/application-shell/application-shell.tsx
+++ b/packages/application-shell/src/components/application-shell/application-shell.tsx
@@ -1,4 +1,9 @@
-import { type ReactNode, type SyntheticEvent, useEffect } from 'react';
+import {
+  type ReactNode,
+  type SyntheticEvent,
+  useEffect,
+  StrictMode,
+} from 'react';
 import type { NormalizedCacheObject } from '@apollo/client';
 import { ApolloClient } from '@apollo/client';
 import type { TFlags } from '@flopflip/types';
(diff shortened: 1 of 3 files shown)


packages/application-shell/src/components/application-shell/application-shell.tsx
```diff
@@ -41,10 +46,24 @@ type TApplicationShellProps = {
   onRegisterErrorListeners?: (args: { dispatch: Dispatch }) => void;
   onMenuItemClick?: (event: SyntheticEvent<HTMLAnchorElement>) => void;
   disableRoutePermissionCheck?: boolean;
+  enableStrictMode?: boolean;
```
emmenko, line 49: Nit: wondering if we should be more explicit that is related to React.

```suggestion
  enableReactStrictMode?: boolean;
```

emmenko (commented):
Thanks! 💯

emmenko:
PS: we should probably add it to the `CustomViewShell` as well.

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
```diff
@@ -160,19 +170,23 @@ function CustomViewShell(props: TCustomViewShellProps) {
...
   if (process.env.NODE_ENV === 'development' && !props.disableDevHost) {
     return (
-      <Suspense fallback={<ApplicationLoader />}>
-        <CustomViewDevHost applicationMessages={props.applicationMessages}>
-          <CustomViewShell applicationMessages={props.applicationMessages}>
-            {props.children}
-          </CustomViewShell>
-        </CustomViewDevHost>
-      </Suspense>
+      <StrictModeEnablement enableReactStrictMode={props.enableReactStrictMode}>
```
tdeekens, line 173: /cc @CarlosCortizasCT hope I could the wrapping at the right level.

emmenko (approved):
💯

CarlosCortizasCT (approved):
Thanks! 🙇

PR #3288, opened 2023-11-06
App bar elements repositioning

chloe0592:
**Before the change:**

Screenshot 2023-11-06 at 09 52 16

**After the change:**

Screenshot 2023-11-06 at 09 51 59

### DIFF
--- a/packages/application-components/src/theming.ts
+++ b/packages/application-components/src/theming.ts
@@ -45,7 +45,6 @@ export const themesOverrides = {
     marginForPageContent: `${uiKitDesignTokens.spacing50} ${appKitSpacing55}`,
     marginTopForPageSubtitle: uiKitDesignTokens.spacing20,
     marginTopForTabControls: uiKitDesignTokens.spacing40,
-    paddingForAppbar: `0 ${appKitSpacing55}`,
     paddingForDetailPageHeader: `${uiKitDesignTokens.spacing50} ${appKitSpacing55} ${uiKitDesignTokens.spacing40}`,
     paddingForDialogContainer: `${uiKitDesignTokens.spacing20} ${uiKitDesignTokens.spacing30}`,
     paddingForDialogContent: `${uiKitDesignTokens.spacing40} 0 0`,
--- a/packages/application-components/src/theming.ts
+++ b/packages/application-components/src/theming.ts
@@ -45,6 +45,7 @@ export const themesOverrides = {
     marginForPageContent: `${uiKitDesignTokens.spacing50} ${appKitSpacing55}`,
     marginTopForPageSubtitle: uiKitDesignTokens.spacing20,
     marginTopForTabControls: uiKitDesignTokens.spacing40,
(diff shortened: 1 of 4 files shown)


tdeekens:
💯 nice to see these small adjustments these days 🤗. I don't want to derail the PR but am curious: did we think about adding an icon to the "project switcher" to provide some guidance and context same as we do with the language drop down?

chloe0592:
> 💯 nice to see these small adjustments these days 🤗. I don't want to derail the PR but am curious: did we think about adding an icon to the "project switcher" to provide some guidance and context same as we do with the language drop down?

@tdeekens thanks for the question! This is just the beginning of making the app bar more intuitive for the users and we do plan to make some adjustments in this phase for both drop downs. 
But I would say @FilPob is probably the best person to answer your question further. 😄

tdeekens (approved):

chloe0592:
I am unable to approve Percy changes when all looks as expected. Is that some new process we have in place? Or is it just me?

CarlosCortizasCT:
> I am unable to approve Percy changes when all looks as expected. Is that some new process we have in place? Or is it just me?

I can approve the changes.
Maybe it's a permissions problem and perhaps @emmenko has the rights to take a look at it.

packages/application-components/src/theming.ts
```diff
@@ -45,6 +45,7 @@ export const themesOverrides = {
     marginForPageContent: `${uiKitDesignTokens.spacing50} ${appKitSpacing55}`,
     marginTopForPageSubtitle: uiKitDesignTokens.spacing20,
     marginTopForTabControls: uiKitDesignTokens.spacing40,
+    paddingForAppbar: `0 ${appKitSpacing55}`,
```
CarlosCortizasCT, line 48: Since we already decided to remove these kind of design tokens, I think it would be better if we don't introduce a new one and just use this value in the `app-bar.tsx` file.

emmenko:
> > I am unable to approve Percy changes when all looks as expected. Is that some new process we have in place? Or is it just me?
> 
> I can approve the changes. Maybe it's a permissions problem and perhaps @emmenko has the rights to take a look at it.

Are you logged in? You do have access btw

chloe0592:
> > > I am unable to approve Percy changes when all looks as expected. Is that some new process we have in place? Or is it just me?
> > 
> > 
> > I can approve the changes. Maybe it's a permissions problem and perhaps @emmenko has the rights to take a look at it.
> 
> Are you logged in? You do have access btw

I was logged in with a different email. 🤦‍♀️ Thanks for the tip!

emmenko (approved):

packages/application-components/src/theming.ts
chloe0592: @CarlosCortizasCT thanks! Good point - [61753a4](https://github.com/commercetools/merchant-center-application-kit/pull/3288/commits/61753a4bcea2795f2c037df3440eeb32b75ff753)

kark (approved):

FilPob (approved):

PR #3292, opened 2023-11-07
Custom Views context hook adjustment

CarlosCortizasCT:
#### Summary

Custom Views context hook adjustment

#### Description

In this PR I propose refactoring the `useCustomView` hook to not only return the Custom Views context but also the internal Application context. This way, Custom Views developers requirement some data from any of the context can access to them from the same hook (dedicated to Custom Views) without them needing to know we manage two context internally.

### DIFF
--- a/.changeset/cuddly-clouds-pull.md
+++ b/.changeset/cuddly-clouds-pull.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-config': minor
+---
+
+Introducing the new `resolveCustomViewResourceAccesses` helper which allows calculating the resource accesses for a Custom View.
--- a/.changeset/cuddly-clouds-pull.md
+++ b/.changeset/cuddly-clouds-pull.md
@@ -2,4 +2,4 @@
 '@commercetools-frontend/application-config': minor
 ---
 
-Introducing the new `resolveCustomViewResourceAccesses` helper which allows calculating the resource accesses for a Custom View.
+Add new helper functions `computeCustomViewPermissionsKeys` and `computeCustomViewResourceAccesses` to compute the permission names for Custom Views.
--- a/packages/application-config/src/formatters.ts
+++ b/packages/application-config/src/formatters.ts
@@ -122,10 +122,19 @@ function entryPointUriPathToResourceAccesses<
   };
 }
 
-function resolveCustomViewResourceAccesses<PermissionGroupName extends string>(
+function computeCustomViewResourceAccesses<PermissionGroupName extends string>(
   permissionGroupNames?: PermissionGroupName[]
 ): TImplicitCustomViewResourceAccesses<PermissionGroupName> {
   return entryPointUriPathToResourceAccesses(
+    CUSTOM_VIEW_HOST_ENTRY_POINT_URI_PATH,
+    permissionGroupNames || []
+  );
+}
+
+function computeCustomViewPermissionsKeys<PermissionGroupName extends string>(
+  permissionGroupNames?: PermissionGroupName[]
+): TImplicitCustomApplicationPermissionKeys<PermissionGroupName> {
+  return entryPointUriPathToPermissionKeys(
     CUSTOM_VIEW_HOST_ENTRY_POINT_URI_PATH,
     permissionGroupNames || []
   );
@@ -162,5 +171,6 @@ export {
   entryPointUriPathToPermissionKeys,
   formatEntryPointUriPathToResourceAccessKey,
   formatPermissionGroupNameToResourceAccessKey,
-  resolveCustomViewResourceAccesses,
+  computeCustomViewResourceAccesses,
+  computeCustomViewPermissionsKeys,
 };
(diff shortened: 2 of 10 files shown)


packages/application-shell/src/test-utils/custom-views-utils.tsx
```diff
@@ -45,6 +49,10 @@ export const renderCustomView = (
       locale: props.locale,
       project: {
         key: props.projectKey,
+        allAppliedPermissions: props.projectAllAppliedPermissions || [],
```
CarlosCortizasCT, line 52: Writing the docs I found out we allow for custom permissions in Custom Application testing so I'm adding that possibility for Custom Views as well.

packages/application-shell/src/test-utils/custom-views-utils.tsx
```diff
@@ -45,6 +49,10 @@ export const renderCustomView = (
...
+      },
+      environment: {
+        entryPointUriPath: CUSTOM_VIEW_HOST_ENTRY_POINT_URI_PATH,
```
emmenko, line 55: 💯

packages/application-shell-connectors/src/components/custom-view-context/custom-view-context.tsx
```diff
@@ -1,11 +1,17 @@
 import { createContext, useContext, useMemo, type ReactNode } from 'react';
 import type { CustomViewData } from '@commercetools-frontend/constants';
+import {
+  type TApplicationContext,
+  useApplicationContext,
+} from '../application-context';
 
 export type TCustomViewContext = {
   hostUrl: string;
   config: CustomViewData;
```
emmenko, line 10: Nit: now that we have the data together, should we rename this to `customViewConfig`?

emmenko (approved):
Nice, thanks for following up on this! 🤗

PS: changeset

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
```diff
@@ -44,6 +45,7 @@ type THostEventData = {
 };
 
 type TCustomViewShellProps = {
+  apolloClient?: ApolloClient<NormalizedCacheObject>;
```
CarlosCortizasCT, line 48: Also allow consumers to provide their custom Apollo client (same as with Custom Applications)

packages/application-shell-connectors/src/components/custom-view-context/custom-view-context.tsx
CarlosCortizasCT: Yes, that makes sense. Thanks 👍 

Updated here: fa5fa67133a501c01858a1bde3737e9387e8574d

emmenko (approved):

packages/application-config/src/formatters.ts
```diff
@@ -118,6 +122,15 @@ function entryPointUriPathToResourceAccesses<
   };
 }
 
+function resolveCustomViewResourceAccesses<PermissionGroupName extends string>(
```
CarlosCortizasCT, line 125: When writing the docs for Custom Views I noticed if a developer needs to calculate the permissions he would need to use something like this:
```js
export const PERMISSIONS = entryPointUriPathToPermissionKeys(CUSTOM_VIEW_HOST_ENTRY_POINT_URI_PATH);
```

It seems confusing to me as the concept of `entry point` does not exist for Custom Views, so I decided to propose the inclusion of this new helper target for Custom Views where they can get both the basic resource accesses and also the one for the additional groups.

packages/application-shell/src/test-utils/custom-views-utils.tsx
```diff
@@ -23,8 +30,12 @@ const testCustomViewData: CustomViewData = {
 type TRenderCustomViewParams = {
   locale: string;
   projectKey?: string;
+  projectAllAppliedPermissions?: { name: string; value: boolean }[];
   customViewHostUrl?: string;
   customViewConfig?: Partial<CustomViewData>;
+  apolloClient?: ApolloClient<NormalizedCacheObject>;
+  environment?: Partial<TProviderProps<{}>['environment']>;
+  user?: Partial<TProviderProps<{}>['user']>;
```
CarlosCortizasCT, line 38: I'm also adding the `environment` and the `user` as they also seem quite helpful when testing Custom Views with different permissions.

packages/application-config/src/formatters.ts
CarlosCortizasCT: @emmenko please let me knot what you think about this 🙏

CarlosCortizasCT: Thanks for the suggestion.

I updated this part here: 96cc64e73e425db269f70c601169871eed7eff6f

.changeset/cuddly-clouds-pull.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-frontend/application-config': minor
+---
+
+Introducing the new `resolveCustomViewResourceAccesses` helper which allows calculating the resource accesses for a Custom View.
```
emmenko, line 5:
```suggestion
Add new helper functions `computeCustomViewPermissionsKeys` and `computeCustomViewResourceAccesses` to compute the permission names for Custom Views.
```

emmenko (approved):

PR #3294, opened 2023-11-08
chore: add drawer component

YahiaElTai:
#### Summary

This PR adds the new `Drawer` component which will implement a side panel container with similarities to modal pages.

image


packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.spec.tsx
```diff
@@ -70,9 +70,9 @@ describe('CustomViewLoader', () => {
       { projectKey }
     );
 
-    const iFrame = screen.getByTitle(
+    const iFrame = screen.getAllByTitle(
```
YahiaElTai, line 73: Because now the drawer has its own title (which the previous custom panel didn't) and the iframe has a title, we need to select the correct one.

packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
```diff
@@ -1,22 +1,41 @@
...
+  width: ${getContainerSize(_props.size)};
   display: flex;
   flex-direction: column;
-  background-color: ${customProperties.colorSurface};
+  background: linear-gradient(
+    270deg,
+    rgba(0, 0, 0, 0.05) 0%,
+    rgba(238, 238, 238, 0) 100%
+  );
+  background-blend-mode: darken;
```
YahiaElTai, line 38: These colors are in the Figma file provided by design team. I have applied these to the container which made the container a bit darker than the previous color which was much whiter. Now the container is dark but the header in the regular modal pages is still white which is looking pretty weird now

I'm assuming design team intended the dark colors of the container to be applied for the header as well?

packages/application-components/src/components/modal-pages/internals/modal-page.tsx
```diff
@@ -61,6 +61,10 @@ type MessageDescriptor = {
   defaultMessage?: string;
 };
 type Label = string | MessageDescriptor;
+
+// naming of `xlarge` should be discussed, I didn't know what to name it
+export type TModalPageSize = 'small' | 'large' | 'xlarge';
```
CarlosCortizasCT, line 66: We are moving towards always using numeric range scales for size props.
Can we maybe use a different set of values here?
```suggestion
export type TModalPageSize = 10 | 20 | 30;
```

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.spec.tsx
CarlosCortizasCT: No sure I understand this 🤔 
We did previously have a title on the Custom Panel, right?

image

packages/application-components/src/components/drawer/drawer.tsx
```diff
@@ -0,0 +1,178 @@
...
+      hideTopBar
+      onClose={props.onClose}
+      size={getModalPageSize(props.size)}
+      title={props.title}
+      afterOpenStyles={props.afterOpenStyles}
+      customViewLocatorCode={props.customViewLocatorCode}
+      getParentSelector={props.getParentSelector}
+      shouldDelayOnClose={props.shouldDelayOnClose}
+      topBarColor={props.topBarColor}
+      zIndex={props.zIndex} // should this be allowed to be set since drawers won't be nested inside each other?
```
CarlosCortizasCT, line 106: I don't think we needed now but maybe we can keep it just in case the `Drawer` would be used in a use case that needs it we can't imagine now.

packages/application-components/src/components/drawer/drawer.tsx
```diff
@@ -0,0 +1,178 @@
...
+    >
+      <div
+        css={css`
```
CarlosCortizasCT, line 109: Can we use a styled component here as you did for the `ContentWrapper`? 🙏

packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
CarlosCortizasCT: Maybe we can have a call about this.
I don't see this panel background colors nor the gradient in the designs, but I might be missing something.

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.spec.tsx
YahiaElTai: if you look at the code of the drawer, now the drawer has something like `<PageHeader>` which displays the title. the screenshot you have doesn't have that which is why we have 2 titles now.

packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
YahiaElTai: Apologizes that was my own confusion, just had a call with Filip and he explain to me that the dark colors he mentioned were for the box shadow and not for the actual background of the modal 😅  silly mistake from my side. all good here. These 2 comments were resolved by Filip.

YahiaElTai: After speaking with Filip, he said we need a new box shadow value for this one and he will speak with Carlos regarding that, once we have a decision there then we can continue with this PR.

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -130,10 +130,12 @@ function CustomViewLoader(props: TCustomViewLoaderProps) {
   ].join('/');
 
   return (
-    <CustomPanel
+    <Drawer
+      isOpen
       title={`Custom View: ${props.customView.defaultLabel}`}
       onClose={props.onClose}
-      size={panelSize}
+      // small is 10, but what is large? is it 20? or 30?
```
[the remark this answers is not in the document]
CarlosCortizasCT: Usually a property should have a set of values and the default one would be one of them. If we use `100%` here as the default value, it looks a bit weird to me as is not a value from the available set.

In other parts we use `scale` as the value for the cases when we need the `100%` space so maybe we can have something like this: `10`, `20`, `30`, `scale`.

I'd like to know what @chloe0592 thinks about it.

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.spec.tsx
CarlosCortizasCT: Thanks. Now I understand.

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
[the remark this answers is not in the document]
chloe0592: @YahiaElTai first of thanks for working on it! Great job! 👏 👏 
I would also agree that it makes sense that 100% is 'scale'. And we are using that all over UI-Kit and App-Kit when setting a size to 100%, so it will be consistent.
I do however, have a question on why are we changing the `ModalPage` sizes from 't-shirt' to 'numeric'?
I mean, there is no universal standard for representing sizes but it seems to me that Drawer component in other UI libraries usually have 't-shirt' sizes since it won't be changing often and it's much more user friendly.

packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
```diff
@@ -1,23 +1,39 @@
...
   top: 0;
   right: 0;
   height: 100%;
-  width: ${_props.size === 'small' ? '600px !important' : '100%'};
+  width: ${getContainerSize(_props.size)};
   display: flex;
   flex-direction: column;
   background-color: ${customProperties.colorSurface};
-  box-shadow: ${customProperties.shadow4}, ${customProperties.shadow6};
+  box-shadow: 0px 0px 40px 0px rgba(0, 0, 0, 0.1);
```
YahiaElTai, line 36: @CarlosCortizasCT  This is the box-shadow value that Filip wanted. I added it here.

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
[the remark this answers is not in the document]
YahiaElTai: I have added the new size `scale` and set it as default :)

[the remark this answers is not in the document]
CarlosCortizasCT: Thanks @chloe0592 for jumping in 👍 

Regarding this comment:
> I do however, have a question on why are we changing the ModalPage sizes from 't-shirt' to 'numeric'?
I mean, there is no universal standard for representing sizes but it seems to me that Drawer component in other UI libraries usually have 't-shirt' sizes since it won't be changing often and it's much more user friendly.

This is a due to what we discussed in our off-site about moving to numbered-range scales to be more flexible with size requirements changes.
Actually, this is something that had already started from time ago: if you take a look at dialog components (ex: [`InfoDialog`](https://github.com/commercetools/merchant-center-application-kit/blob/main/packages/application-components/src/components/dialogs/info-dialog/info-dialog.tsx#L9)) we were using t-shirt sizes in the beginning but introduced numbered-ranges later (we're keeping t-shirts for retro-compatibility for a while).

Does this make sense to you?

CarlosCortizasCT:
Thanks @YahiaElTai, this looks great! 👍 

Maybe just a couple of things more to play safe in the long term:
* Maybe we could add another VRT where both title and subtitle are so long they will use multiple lines: this would help to make sure those respect the close icon "column".
* Since the `Drawer` has intro and exit animations, it would be great to add an example to the playground application so we can validate it. [This](https://github.com/commercetools/merchant-center-application-kit/blob/main/playground/src/components/notifications-playground/notifications-playground.jsx) would be the file to update. Please let me know if you need some context about how the playground works.

Also for @FilPob, since you can't see the new component in a preview deployment, just wanted to share that the new overlay color seems too shallow to me and just wanted to double check it is intended 😅 

image

FilPob:
@CarlosCortizasCT thanks for the early update! Yes the light overlay is expected.

FilPob:
@emmenko great questions, I mentioned it somewhere but will leave it here as well to make sure:

- It should be not possible to open another drawer on top of a drawer
- Opening a info dialog from the Drawer should be allowed

emmenko:
Thanks for the feedback.

> It should be not possible to open another drawer on top of a drawer

I think this is more of a guideline (Dos / Donts) rather than something that we prevent in the code. It should be documented somewhere.

> Opening a info dialog from the Drawer should be allowed

@YahiaElTai Can you verify that in the playground? We can also have a VRT for that. Thanks!

YahiaElTai:
> Thanks for the feedback.
> 
> > It should be not possible to open another drawer on top of a drawer
> 
> I think this is more of a guideline (Dos / Donts) rather than something that we prevent in the code. It should be documented somewhere.
> 
> > Opening a info dialog from the Drawer should be allowed
> 
> @YahiaElTai Can you verify that in the playground? We can also have a VRT for that. Thanks!

Thanks for the questions and the help from all. I will take a look at that today :)

CarlosCortizasCT:
Now that we have the playground example, I noticed the animation we're using is the same one we currently have for modal pages but it seems (from the designs) we want to have a slightly different one for drawer. 
I actually think we would like to have even slightly different animations based on the drawer's width.

@FilPob @YahiaElTai I can schedule a call to discuss this point if it's ok with you.

[App-kit-playground - almond-40 - Merchant Center (6).webm](https://github.com/commercetools/merchant-center-application-kit/assets/97907068/ca8bc890-4bd1-42da-9447-1e45b2694ff9)

YahiaElTai:
> Now that we have the playground example, I noticed the animation we're using is the same one we currently have for modal pages but it seems (from the designs) we want to have a slightly different one for drawer. I actually think we would like to have even slightly different animations based on the drawer's width.
> 
> @FilPob @YahiaElTai I can schedule a call to discuss this point if it's ok with you.
> 
>  App-kit-playground.-.almond-40.-.Merchant.Center.6.webm

@CarlosCortizasCT Thanks and yes please schedule a call for us to discuss this further :)

YahiaElTai:
> Thanks for the feedback.
> 
> > It should be not possible to open another drawer on top of a drawer
> 
> I think this is more of a guideline (Dos / Donts) rather than something that we prevent in the code. It should be documented somewhere.
> 
> > Opening a info dialog from the Drawer should be allowed
> 
> @YahiaElTai Can you verify that in the playground? We can also have a VRT for that. Thanks!

I have added a VRT for that and also added it in the playground. The video below shows how the drawer behave in the playground.


Note: As @emmenko said we can't technically prevent drawer from opening from within a drawer, this is something to be added in the docs but the video shows the behavior of opening multi modal and multi modal all inside each other.



https://github.com/commercetools/merchant-center-application-kit/assets/30247059/20bf50b3-85eb-4af7-bac7-79f1a772ccc8

CarlosCortizasCT:
Thanks @YahiaElTai 

The video looks good to me, although I wouldn't add all the buttons to the drawer content but only the dialog trigger as that's the only we would like to allow.

YahiaElTai:
Apologies, I got confused with "info modal" and "info dialog" 😅 Please ignore the video above and watch this one which shows the behavior interaction between the info dialog and the drawer 

https://github.com/commercetools/merchant-center-application-kit/assets/30247059/90e35149-58c4-46c8-8ea5-36da45e50d11

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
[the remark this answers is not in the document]
chloe0592: Hey @CarlosCortizasCT yes, I know we talked about it, but I wasn't aware we made a decision yet. 😄 But please continue as planned. Thanks!

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -130,10 +130,11 @@ function CustomViewLoader(props: TCustomViewLoaderProps) {
   ].join('/');
 
   return (
-    <CustomPanel
+    <Drawer
+      isOpen
       title={`Custom View: ${props.customView.defaultLabel}`}
       onClose={props.onClose}
-      size={panelSize}
+      size={panelSize === 'small' ? 10 : 30}
```
YahiaElTai, line 137: FYI @CarlosCortizasCT  Since the drawer can only have 3 sizes and "Scale" is not one of them then this is the only option to make sure that the custom view loader is as big as possible if the panel size is not small.

packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
```diff
@@ -17,13 +18,27 @@ const getContainerSize = (size?: TModalPageSize) => {
...
-      return '100%';
     default:
       return '100%';
   }
 };
 
+const getTransformationSize = (size?: TModalPageSize) => {
+  switch (size) {
+    case 10:
+      return '600px';
```
CarlosCortizasCT, line 29: Since we are repeating these values, what do you think about having an internal mapper object we can reuse here? 

Something like this:
```js
const stylesBySize = {
  '10': {
    width: '600px',
    transitionTime: 200
  },
  '20': {
    width: '900',
    transitionTime: 200
  },
  '30': {
    width: '1200',
    transitionTime: 200
  },
 'scale': {
    width: '600',
    transitionTime: 300
  },
}
```

YahiaElTai: Great idea, pushed a commit for it.

packages/application-components/src/components/modal-pages/internals/modal-page.tsx
```diff
@@ -110,6 +114,9 @@ const ModalPage = (props: Props) => {
     };
   }, [props.isOpen]);
   const { onClose } = props;
+
+  const TRANSITION_DURATION = stylesBySize[props.size].transitionTime;
```
CarlosCortizasCT, line 118: Hey Yahia, do we really need to pull the transition duration value here?
I guess we're using it now just in case the `ModalPage` component changes its size once rendered, right? (don't know such a use case).
If so, maybe we can just add the `props.size` to the dependencies array.

What do you think? 🤔

packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
```diff
@@ -1,26 +1,53 @@
 import { css, type SerializedStyles } from '@emotion/react';
 import { customProperties } from '@commercetools-uikit/design-system';
+import type { TModalPageSize } from './modal-page';
 
-export const TRANSITION_DURATION = 200;
+const TRANSITION_DURATION = 200;
+const LONGER_TRANSITION_DURATION = 300;
```
CarlosCortizasCT, line 6: One of the reasons why I suggested to use the styles mapper object was to get rid of these constants, as they don't seem so useful now.

However, if you prefer to keep them, I would suggest renaming them as they're not very self explanatory right now.
Maybe something like: 
* `TRANSITION_DURATION_SHORT` and `TRANSITION_DURATION_LONG`
* `FULL_WIDTH_TRANSITION_DURATION` and `NARROWER_WIDTH_TRANSLATION_DURATION` 😨

packages/application-components/src/components/modal-pages/internals/modal-page.tsx
YahiaElTai: Actually no, the transition value of `200` was used in the modal page already before my PR so I just replaced that value which was coming from the constant to the value coming from the stylesBySize object but still it was used already in 2 places. mostly when closing the modal to handle the closing animation and the timeout of the closing.

packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
YahiaElTai: Removed them.

packages/application-components/src/components/modal-pages/internals/modal-page.tsx
CarlosCortizasCT: Oh! I see.

Thanks for the explanation. 👍

CarlosCortizasCT (commented):
Could you please move the PR out of draft mode?

packages/application-components/src/components/drawer/drawer.tsx
YahiaElTai: Actually this comment was removed already, if you go to "Files changes" in Github, you will see that it's not there.

YahiaElTai:
@CarlosCortizasCT Thanks for the comments and the help with this PR. The only thing left is to review and approve the percy snapshots. I will leave that to you 😄

packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
```diff
@@ -1,26 +1,50 @@
...
+  },
 };
 
 export const getContainerStyles = (_props: StyleProps): SerializedStyles => css`
   position: absolute;
   top: 0;
   right: 0;
   height: 100%;
-  width: ${_props.size === 'small' ? '600px !important' : '100%'};
+  width: ${_props.size !== 'scale'
```
emmenko, line 34: Nit: I know it was already like this but could you rename it to `props` instead of `_props`? In TS an underscore as a prefix is used to silent the compiler that a variable is unused.

Thanks

emmenko (commented):
Thanks a lot for the contribution! ❤️

Only a changeset is missing now 😉

emmenko (approved):
💯

PR #3295, opened 2023-11-13
New Custom Views Javascript template

CarlosCortizasCT:
#### Summary

New Custom Views Javascript template

#### Description

So far we only had a `Typescript` version of the template to bootstrap a new Custom View project.
We're adding a new Javascript based template.


CarlosCortizasCT:
@emmenko do you think we can already remove the experimental flag for creating Custom Views templates or should we keep it until the General Available stage?

```typescript
const throwIfApplicationTypeIsNotSupported = (
  applicationType: TApplicationType
) => {
  switch (applicationType) {
    case applicationTypes['custom-view']: {
      if (process.env.ENABLE_EXPERIMENTAL_CUSTOM_VIEWS !== 'true') {
        throw new Error(`Custom Views generation is not yet supported.`);
      }
      break;
    }
    case applicationTypes['custom-application']:
      break;
    default: {
      const applicationTypesList = Object.keys(applicationTypes).toString();
      throw new Error(
        `The provided application type "${applicationType}" does not exist. Available types are "${applicationTypesList}". Make sure you are also using the latest version of "@commercetools-frontend/create-mc-app".`
      );
    }
  }
};

```

CarlosCortizasCT:
@emmenko I also noticed we're [testing Custom Applications template](https://github.com/commercetools/merchant-center-application-kit/blob/main/.github/workflows/main.yml#L73) in CI but we're not doing it for Custom Views. Should I update the configuration to enable it?

emmenko:
> @emmenko do you think we can already remove the experimental flag for creating Custom Views templates or should we keep it until the General Available stage?

I think we can keep it until general availability considering that the docs won't be "publicly" visible yet and that the configuration UI is also enabled for early adopters.

CarlosCortizasCT:
> I think we can keep it until general availability considering that the docs won't be "publicly" visible yet and that the configuration UI is also enabled for early adopters.

Sorry I didn't give all the context in my first question.

I was thinking that now we will be having users starting to develop Custom Views, it would be easier for them if we don't ask them to set the env var.

If we keep the experimental flag, we need to also update the (private) docs to include over there the env variable consumers need to set in order for the `cli` to work.

Would it be ok to just update the example command?
```bash
shell> ENABLE_EXPERIMENTAL_CUSTOM_VIEWS="true" npx @commercetools-frontend/create-mc-app@latest \
  my-new-custom-view-project \
  --application-type custom-view \
  --template starter
```

package.json
```diff
@@ -49,6 +49,9 @@
     "template-starter-typescript:build": "pnpm --filter @commercetools-applications/merchant-center-template-starter-typescript run build",
     "template-starter-typescript:start": "pnpm --filter @commercetools-applications/merchant-center-template-starter-typescript run start",
     "template-starter-typescript:start:prod:local": "pnpm --filter @commercetools-applications/merchant-center-template-starter-typescript run start:prod:local",
+    "template-custom-view-starter-javascript:build": "pnpm --filter @commercetools-applications/merchant-center-custom-view-template-starter run build",
+    "template-custom-view-starter-javascript:start": "pnpm --filter @commercetools-applications/merchant-center-custom-view-template-starter run start",
+    "template-custom-view-starter-javascript:start:prod:local": "pnpm --filter @commercetools-applications/merchant-center-custom-view-template-starter run start:prod:local",
```
emmenko, line 54: Nit: names seem to be a bit inconsistent with the Custom Applications

```suggestion
    "template-custom-view-starter:build": "pnpm --filter @commercetools-applications/merchant-center-custom-view-template-starter run build",
    "template-custom-view-starter:start": "pnpm --filter @commercetools-applications/merchant-center-custom-view-template-starter run start",
    "template-custom-view-starter:start:prod:local": "pnpm --filter @commercetools-applications/merchant-center-custom-view-template-starter run start:prod:local",
```

scripts/install-template.mjs
```diff
@@ -70,3 +86,16 @@ console.log('==> Assert application version');
...
+  assert.strictEqual(
+    doesFileExist(path.join(applicationPath, 'custom-application-config.mjs')),
+    true
+  );
+} else if (applicationType === 'custom-view') {
+  assert.strictEqual(
+    doesFileExist(path.join(applicationPath, 'custom-view-config.mjs')),
+    true
+  );
+}
```
emmenko, line 101: 💯

scripts/install-template.mjs
```diff
@@ -70,3 +86,16 @@ console.log('==> Assert application version');
...
+
+console.log('==> Assert application config file exists');
+if (applicationType === 'custom-application') {
+  assert.strictEqual(
+    doesFileExist(path.join(applicationPath, 'custom-application-config.mjs')),
+    true
+  );
+} else if (applicationType === 'custom-view') {
+  assert.strictEqual(
+    doesFileExist(path.join(applicationPath, 'custom-view-config.mjs')),
```
emmenko, line 98: Nit: we can simply use `existsSync` (we use it many places in the repo), so there is no need to have a try/catch

```suggestion
    fs.existsSync(path.join(applicationPath, 'custom-application-config.mjs')),
    true
  );
} else if (applicationType === 'custom-view') {
  assert.strictEqual(
    fs.existsSync(path.join(applicationPath, 'custom-view-config.mjs')),
```

.github/actions/test-template-action/action.yml
```diff
@@ -0,0 +1,89 @@
...
+    required: true
+  cypress-project-key:
+    description: 'Cypress project key'
+    required: true
+
+runs:
+  using: 'composite'
+  steps:
+    # - name: Checkout
+    #   uses: actions/checkout@v4
```
emmenko, line 37: I guess this can be removed?

.github/actions/test-template-action/action.yml
```diff
@@ -0,0 +1,89 @@
+name: 'Test template'
+description: 'Tests MC Extension template'
+
+inputs:
+  template-name:
+    description: 'Name of the template to test'
```
emmenko, line 6: Nit: this is currently used as e.g. `template-starter-typescript` or `template-custom-view-starter-typescript`

When I think about the `create-mc-app` CLI options, the `--template` option is either `starter` or `starter-template`.

I'm wondering if maybe we could have a similar properties here like `application-type` and `template-name`.

Example usage:

```yaml
- name: Test Custom Application typescript starter template
  uses: ./.github/actions/test-template-action
  with:
    application-type: custom-application
    template-name: starter-typescript
```

With these two properties, we can then build the command names or whatever we need to. The main benefit I see with this is naming consistency and familiarity.

Just a suggestion anyway, happy to hear your thoughts 🤗

emmenko (commented):
Awesome job! 🙌

Great idea with the composite action to simplify the setup and to have a consistent commands 🎉

scripts/install-template.mjs
CarlosCortizasCT: I used this implementation as it's the one recommended in Nodejs docs since `fs.exists` is a [deprecated function](https://nodejs.org/docs/latest-v18.x/api/fs.html#fsexistspath-callback).

image

package.json
CarlosCortizasCT: I agree they are inconsistent right now as I didn't want to change anything related to Custom Applications just in case there were any unexpected impact.

From my point of view, Custom Application commands are the ones who lack a more meaningful name and what I think it would have more sense is something like this:
```
"template-custom-application-starter-javascript:build"
"template-custom-application-starter-javascript:start"
"template-custom-application-starter-javascript:start:prod:local"
"template-custom-application-starter-typescript:build"
"template-custom-application-starter-typescript:start"
"template-custom-application-starter-typescript:start:prod:local"

"template-custom-view-starter-javascript:build"
"template-custom-view-starter-javascript:start"
"template-custom-view-starter-javascript:start:prod:local"
"template-custom-view-starter-typescript:build"
"template-custom-view-starter-typescript:start"
"template-custom-view-starter-typescript:start:prod:local"
```

Does it make sense to you our do you prefer to infer javascript is the default implementation?

scripts/install-template.mjs
emmenko: 🤦‍♂️ My bad, sorry.

I guess we would need to refactor the other places then to stop using `fs.existsSync`.
I'll create a task.

package.json
emmenko: I see, thanks for the clarification.

From my experience using other tools and libraries, JS is normally the default implementation yes. However, I do see your point and I don't mind either way. For me it's more important to have it consistent.

Maybe you can run a quick poll in the team to get their feedback as well. Renaming the Custom Applications template then should be a follow up.
Thanks

.github/workflows/main.yml
```diff
@@ -161,57 +131,125 @@ jobs:
...
-          MC_API_URL: ${{ secrets.MC_API_URL }}
+          APPLICATION_TYPE: custom-application
+          TEMPLATE_NAME: starter-typescript
           CTP_INITIAL_PROJECT_KEY: ${{ secrets.CYPRESS_PROJECT_KEY }}
 
-      - name: Building Starter template application
-        run: pnpm template-starter-typescript:build
-        env:
-          CTP_INITIAL_PROJECT_KEY: ${{ secrets.CYPRESS_PROJECT_KEY }}
+  # TODO: Uncomment when we have Custom View e2e tests implemented
```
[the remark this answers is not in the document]
CarlosCortizasCT: Updated here: adc437d55708f9611433001be0dcff3bad79f6f8

custom-views-templates/starter/src/constants.js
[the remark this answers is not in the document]
CarlosCortizasCT: Removed here: 1b69e244f36629f9572e7b59a20941988604cb70

custom-views-templates/starter/src/test-utils/index.jsx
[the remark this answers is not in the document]
CarlosCortizasCT: Removed here: 1b69e244f36629f9572e7b59a20941988604cb70

I copied the Custom Application JS template.

package.json
CarlosCortizasCT: I updated to default names to Javascript here: adc437d55708f9611433001be0dcff3bad79f6f8

```
"template-custom-application-starter:build"
"template-custom-application-starter:start"
"template-custom-application-starter:start:prod:local"
"template-custom-application-starter-typescript:build"
"template-custom-application-starter-typescript:start"
"template-custom-application-starter-typescript:start:prod:local"

"template-custom-view-starter:build"
"template-custom-view-starter:start"
"template-custom-view-starter:start:prod:local"
"template-custom-view-starter-typescript:build"
"template-custom-view-starter-typescript:start"
"template-custom-view-starter-typescript:start:prod:local"
```

.github/actions/test-template-action/action.yml
CarlosCortizasCT: Removed here: adc437d55708f9611433001be0dcff3bad79f6f8

.github/actions/test-template-action/action.yml
CarlosCortizasCT: Yes, I agree it's easier to understand.

Updated here: adc437d55708f9611433001be0dcff3bad79f6f8

.github/workflows/main.yml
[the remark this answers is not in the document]
CarlosCortizasCT: Sorry I didn't understand the suggestion the first time I read it.

I updated after our meeting in this commit: b89ac73e80fa0e95bf53dee14ae6849fd9aacb78

emmenko (approved):
Great work, thanks again 🙌 🚀

packages/create-mc-app/src/validations.ts
emmenko, line 1: Nit: do we need a changeset for this change?

CarlosCortizasCT: Added here: d63c89311b2f4a222ee479f6ef8bcf5d874fb967

tdeekens (approved):

PR #3300, opened 2023-11-20
Add extra container in top nav for portal

jaikumar-tj:
## Summary

Creating a empty container in top-navigation panel for staffbar team switcher to display the `limited permission` stamp using portal.

Screenshot 2023-11-20 at 12 03 36


packages/application-shell/src/components/app-bar/app-bar.tsx
```diff
@@ -83,7 +83,12 @@ const AppBar = (props: Props) => {
           </Spacings.Inline>
         </Spacings.Inline>
       </div>
-      <div>
+      <div
+        css={css`
+          display: flex;
+        `}
+      >
+        <div id="extra-menu"></div>
```
jaikumar-tj, line 91: please suggest a better id or data-test name

jaikumar-tj:
FYI @chloe0592 I assume you were also trying to resolve this position issue. we discussed and decided that having an extra container would be better since the shield team can move it according to their new design independently.

packages/application-shell/src/components/app-bar/app-bar.tsx
jaikumar-tj: Also should it have a margin here or can it remain in mc-frontend instead where the portal is used?

tdeekens: 1. Naming

Given [here](https://github.com/commercetools/merchant-center-application-kit/pull/3066/files) on the prior PR we didn't introduce a `staff-bar--main` or something name dedicated to the usage we could maybe name this `left-of-profile`? Just as a suggestion?

2. Margin

I would not specify the margin here but define it in the MC FE. Maybe using spacer components.

Always found this a good article on the subject https://mxstbr.com/thoughts/margin/

emmenko, line 91: Nit: I would define the ID here: https://github.com/commercetools/merchant-center-application-kit/blob/db6e172ce4d679a66fc6030e4f67c4fb5661d065/packages/application-shell/src/constants.ts#L13-L15

emmenko: Agreed, no styles needed here.

emmenko (commented):
🤗

tdeekens (approved):

emmenko (approved):
Thanks!

PR #3302, opened 2023-11-21
Updating `pnpm-lock.yaml` file

chloe0592:
Some inconsistencies happened in `pnpm-lock.yaml` file on `main` while updating the dependencies. So I ran `pnpm install` in order to fix it.


CarlosCortizasCT (approved):
Thanks! 🙇

kark (approved):
Thank you!

PR #3306, opened 2023-11-23
refactor(Navbar): keyboard navigation

kark:
#### Summary

[Deployment link in the MC](https://mc-15795.mc-preview.europe-west1.gcp.escemo.com/adnan-nodejs-test-project/welcome)

To test out, navigate within the Navbar using `Tab` / `Shift` + `Tab`.

This PR introduces keyboard navigation to the Navbar component, adhering to the default browser focus ring color, as discussed with Filip.

~~Known limitation:
Primary limitation is in Firefox, details provided below 👇~~

### DIFF
--- a/packages/application-shell/src/components/navbar/navbar.mod.css
+++ b/packages/application-shell/src/components/navbar/navbar.mod.css
@@ -9,7 +9,6 @@
   --sublist-indentation-when-expanded: 248px;
   --sublist-width: 253px;
   --left-navigation-transition: all 150ms cubic-bezier(1, 0, 0.58, 1);
-  --focus-ring-color: -webkit-focus-ring-color;
 
   /* Left navigation */
   --width-left-navigation: 80px;
--- a/packages/application-shell/src/components/navbar/navbar.mod.css
+++ b/packages/application-shell/src/components/navbar/navbar.mod.css
@@ -9,6 +9,7 @@
   --sublist-indentation-when-expanded: 248px;
   --sublist-width: 253px;
   --left-navigation-transition: all 150ms cubic-bezier(1, 0, 0.58, 1);
+  --focus-ring-color: -webkit-focus-ring-color;
 
   /* Left navigation */
   --width-left-navigation: 80px;
@@ -537,10 +538,10 @@
 .list-item:not(.item_menu-collapsed):has(.text-link:focus),  /* An alternative to `.list-item:not(.item_menu-collapsed):focus-within`, with the added advantage that the main menu item doesn't receive focus outline when a submenu item is in focus. */ 
 /* Submenu item */
 .sublist-item:focus-within {
-  outline: auto 1px Highlight;
+  outline: auto 1px var(--focus-ring-color, Highlight);
   outline-offset: 1px;
 }
-/* TODO: Remove once Firefox version is adopted widely beyond v121 */
+/* TODO: Remove once Firefox version is adopted widely beyond v121 and `:has` pseudo-class can be used */
 @-moz-document url-prefix('') {
   .list-item:not(.item_menu-collapsed):focus-within,
   .sublist-item:focus-within {
--- a/packages/application-shell/src/components/navbar/navbar.mod.css
+++ b/packages/application-shell/src/components/navbar/navbar.mod.css
@@ -187,12 +187,6 @@
   border-radius: var(--border-radius-8);
 }
 
-.list-item:not(.item_menu__active):focus,
-.list-item:not(.item_menu-collapsed):focus-within {
-  outline: auto 1px Highlight;
-  outline-offset: 1px;
-}
-
 .item_menu-open {
   height: auto !important;
 }
@@ -353,11 +347,6 @@
     var(--spacing-30);
 }
 
-.sublist-item:focus-within {
-  outline: auto 1px Highlight;
-  outline-offset: 1px;
-}
-
 .navlink-clickable-content {
   height: 100%;
   width: 100%;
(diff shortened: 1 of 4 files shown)


chloe0592 (approved):
Thanks so much @kark for the great work! 👏  Looks all good from my side. 

@FilPob might also want to check if he hasn't already.

packages/application-shell/src/components/navbar/navbar.mod.css
```diff
@@ -514,3 +532,20 @@
...
+  outline-offset: 1px;
+}
+/* TODO: Remove once Firefox version is adopted widely beyond v121 and `:has` pseudo-class can be used */
+@-moz-document url-prefix('') {
+  .list-item:not(.item_menu-collapsed):focus-within,
+  .sublist-item:focus-within {
+    outline: auto 1px Highlight;
+    outline-offset: 1px;
+  }
+}
```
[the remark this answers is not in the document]
chloe0592: Thank you for the detailed explanation @kark ! I can see the challenges you faced with the focus ring and the use of the `:has()` pseudo-class. It seems to me like a pragmatic approach given the current browser support and the inconsistencies observed across different browsers. Well done on finding a solution!

packages/application-shell/src/components/navbar/navbar.mod.css
```diff
@@ -329,14 +334,21 @@
...
   padding: var(--spacing-25) var(--spacing-25) var(--spacing-25)
     var(--spacing-30);
+  transition: padding 150ms ease-out;
+}
+
+.sublist-item:not(.sublist-item__active):hover .text-link-sublist,
+.sublist-item:not(.sublist-item__active):focus-within .text-link-sublist {
+  /* additional left padding on hover and focus */
+  padding: var(--spacing-25) var(--spacing-25) var(--spacing-25)
+    calc(var(--spacing-30) + var(--spacing-20));
```
kark, line 351: I've revised the strategy in the last update. It appeared that the `'overflow: hidden'`s I removed might not be crucial (I couldn't detect differences, but reviewers may catch any regressions). As a result, I've shifted to adding the focus ring and extra padding directly to the `<a>` (`.text-link-sublist`) element instead of its parent.

This modification ensures compatibility across all browsers without the need for any workarounds.

For now, I'm holding off on the scheduled sync as there are no pending questions on my end.

Your feedback is most welcome!

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
```diff
@@ -71,6 +71,14 @@ const reducer = (state: State, action: Action): State => {
   }
 };
 
+const isFocusOutEventCalledBySubmenuItem = (event: FocusEvent) =>
+  // a case when a submenu item loses focus
+  event.type === 'focusout' &&
+  // element receiving focus
+  (event.relatedTarget as Element)?.matches('a[class*="__text-link__"]') &&
+  // element losing focus
+  (event.target as Element)?.matches('a[class*="__text-link-sublist__"]');
```
emmenko, line 80: I would suggest to use more "stable" selectors. We shouldn't rely on class names, as they might change depending on how we style things.

Can we use data attributes for example?

emmenko (commented):
Thanks @kark for working on this!

I'll do some testing locally, there is just one thing I noticed that I would like us to address.

emmenko:
I tested the <kbd>Tab</kbd> navigation. Works like a charm! 🙌

PS: navigating with arrows within the submenu doesn't work (I was expecting that first). Was that left out intentionally and will be tackled as a follow up?

kark:
> PS: navigating with arrows within the submenu doesn't work (I was expecting that first). Was that left out intentionally and will be tackled as a follow up?

Hmm, I think you're right. This goes back to draft; it doesn't meet the required standards.

emmenko:
Let's work on that as a follow up then, I think this PR is good to go in regards to tabs navigation (besides the [comment about the selector](https://github.com/commercetools/merchant-center-application-kit/pull/3306#discussion_r1425236314)).

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
kark: Thank you! Updated in 7b31dfb06b4db1c6c623a7b8c9589bcf3cb9cbcb. 
I will create a task for the follow-up work.

CarlosCortizasCT:
I've been testing the navigation with the Tab key and, although it works fine, I was wondering if we would like to refine the behaviour with arrow-keys navigation as Nicola mentioned (maybe as a follow-up).

A user who wants to access one of the applications at the bottom of the navbar must go over all the previous one and their respective sublinks: maybe we could make the Tab key to navigate over root navbar items and arrow keys to enter and select submenu links as we would be making the overall keyboard navigation more agile.

On a side note, when a CT user lands into the Merchant Center and starts pushing the Tab key, s/he needs to push 7 times before the focus gets to a visible element (the project selector) and I think this is due to the Staff bar, which should not receive focus while collapsed.
Maybe this is something we can raise to be addressed in that component as a follow-up.

kark:
Appreciate the feedback, @emmenko and @CarlosCortizasCT 💯. 
Following the suggestions, I've created a dedicated task for the follow-up work.

CarlosCortizasCT (approved):
Many thanks! 🙇 

_(don't forget to take a look at Percy 🙏 )_

kark:
> Many thanks! 🙇
> 
> _(don't forget to take a look at Percy 🙏 )_

Thank you 🙏
All the changes in the vrts are associated with the update in top padding for the `support-menu` item. This adjustment is crucial; otherwise, the focus ring would be cut off.
| before  | after  |
|---|---|
| Screenshot 2023-12-19 at 08 49 40  | Screenshot 2023-12-19 at 08 48 50  |

emmenko (approved):
Thanks! 💯

PR #3308, opened 2023-11-24
refactor(custom views selector): ui refinement

kark:
#### Summary

This PR changes the UI of the custom views selector component.
It doesn't cover the correct placement of the selector.

https://github.com/commercetools/merchant-center-application-kit/assets/49066275/f7342014-22ae-45fb-88ea-afdb0e5a43e4

The changes can be observed in visual regression tests output 👇.
Screenshot 2023-11-27 at 11 55 08

Open questions:

| designs  | PR  |
|---|---|
|  Screenshot 2023-11-27 at 10 50 32 | Screenshot 2023-11-27 at 10 43 59  |

- for now I haven't changed the icon and the title. 

Regarding the icon, it felt to me it is associated with an additional feature for hiding the selector upon clicking the icon. However, as of now, this functionality has not been implemented. 
As for the title, is it essential to remove the colon? Please note that if the colon is removed, all messages will need to be retranslated.

### DIFF
--- a/packages/application-components/src/components/custom-views/custom-views-selector/custom-views-selector.tsx
+++ b/packages/application-components/src/components/custom-views/custom-views-selector/custom-views-selector.tsx
@@ -91,14 +91,6 @@ const CustomViewSelectorItem = (props: TCustomViewSelectorItemProps) => {
     >
       <Text.Detail
         tone="inherit"
-        css={css`
-          color: ${designTokens.colorPrimary};
-          ${props.selected && hoverAndSelectedStateStyles}
-
-          :hover {
-            ${hoverAndSelectedStateStyles};
-          }
-        `}
         fontWeight={props.selected ? 'bold' : 'regular'}
       >
         {props.label}
--- a/packages/application-components/src/components/custom-views/custom-views-selector/custom-views-selector.tsx
+++ b/packages/application-components/src/components/custom-views/custom-views-selector/custom-views-selector.tsx
@@ -138,7 +138,15 @@ function CustomViewSelector(props: TCustomViewSelectorWithRequiredProps) {
           >
             <Spacings.Inline scale="xs" alignItems="center">
               <SidebarCollapseIcon size="medium" color="neutral60" />
-              <Text.Detail tone="secondary" intlMessage={messages.title} />
+              <div
+                css={css`
+                  div {
+                    font-weight: 300;
+                  }
+                `}
+              >
+                <Text.Detail tone="secondary" intlMessage={messages.title} />
+              </div>
             </Spacings.Inline>
             {customViews.map((customView, index) => {
               const isNotLastItem = index !== customViews.length - 1;
--- a/packages/application-components/src/components/custom-views/custom-views-selector/custom-views-selector.tsx
+++ b/packages/application-components/src/components/custom-views/custom-views-selector/custom-views-selector.tsx
@@ -12,7 +12,7 @@ import { reportErrorToSentry } from '@commercetools-frontend/sentry';
 import AccessibleButton from '@commercetools-uikit/accessible-button';
 import Constraints from '@commercetools-uikit/constraints';
 import { designTokens } from '@commercetools-uikit/design-system';
-import { SidebarCollapseIcon } from '@commercetools-uikit/icons';
+import { WindowEyeIcon } from '@commercetools-uikit/icons';
 import Spacings from '@commercetools-uikit/spacings';
 import Text from '@commercetools-uikit/text';
 import CustomViewLoader from '../custom-view-loader';
@@ -137,7 +137,7 @@ function CustomViewSelector(props: TCustomViewSelectorWithRequiredProps) {
             alignItems="center"
           >
             <Spacings.Inline scale="xs" alignItems="center">
-              <SidebarCollapseIcon size="medium" color="neutral60" />
+              <WindowEyeIcon size="medium" color="neutral60" />
               <div
                 css={css`
                   div {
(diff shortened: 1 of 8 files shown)


packages/application-components/src/components/custom-views/custom-views-selector/custom-views-selector.tsx
```diff
@@ -34,8 +41,68 @@ const Wrapper = styled.div<TWrapperProps>`
...
+      0deg,
+      ${designTokens.colorNeutral95},
+      ${designTokens.colorNeutral95}
+    ),
+    linear-gradient(
+      0deg,
+      ${designTokens.colorNeutral98},
+      ${designTokens.colorNeutral98}
+    );
+  padding: 7px ${designTokens.spacing20};
```
emmenko, line 54: Nit: why `7`? Can't we use the normal scale of multiple of 4s (4, 8, 12, 16, ...)?

packages/application-components/src/components/custom-views/custom-views-selector/custom-views-selector.tsx
```diff
@@ -34,8 +41,68 @@ const Wrapper = styled.div<TWrapperProps>`
...
+};
+
+const CustomViewSelectorItem = (props: TCustomViewSelectorItemProps) => {
+  return (
+    <AccessibleButton
+      css={css`
+        display: flex;
+        align-items: center;
+        padding: ${designTokens.spacing20} ${designTokens.spacing25};
+        height: 30px;
```
emmenko, line 80: Nit: also here, why do we need to make an exception for this?

Can't we use the design token for buttons (medium)?

https://github.com/commercetools/ui-kit/blob/dc24278e712b45a08872cbf3608a3d36dabe74d0/design-system/materials/internals/definition.yaml#L971-L972

packages/application-components/src/components/custom-views/custom-views-selector/custom-views-selector.tsx
```diff
@@ -34,8 +41,68 @@ const Wrapper = styled.div<TWrapperProps>`
...
+        {props.label}
+      </Text.Detail>
+    </AccessibleButton>
+  );
+};
+
+const Separator = styled.span`
+  width: 1px;
+  height: 18px;
+  background-color: #cccccc;
```
emmenko, line 105: Nit: these values shouldn't be hardcoded

emmenko (commented):
Nice work! 🤗

packages/application-components/src/components/custom-views/custom-views-selector/custom-views-selector.tsx
kark: Sounds good to me 👍, would it be ok with you @NataliaSclipet?
Screenshot 2023-11-28 at 10 06 50

NataliaSclipet:
Thank you for working on it @kark .
Regarding the title and the icon, could you implement the title and icon that are in the new designs and keep the colon, (don't remove it), please? 
Thank you for sharing your input regarding the icon, the decision of changing it comes from Simone and Sven.

packages/application-components/src/components/custom-views/custom-views-selector/custom-views-selector.tsx
kark: Sounds good, just double checking with @NataliaSclipet 
Screenshot 2023-11-28 at 10 10 03

NataliaSclipet: Can we keep this exception both for here and above please? I'm aware we are not using our design tokens, however these exceptions here has been defined and reviewed together with Sven and Simone.

kark: Thank you, that's fine with me 😊

packages/application-components/src/components/custom-views/custom-views-selector/custom-views-selector.tsx
emmenko: Don't we have an existing design token for this color? Why do we need to add a new one?

kark: If I'm not mistaken, `#cccccc` translates to `hsl(0, 0%, 80%)`. I don't think we have it available as a token.

CarlosCortizasCT:
Regarding the `Custom Views` label, please also be aware the design is using `font-weigth: 300` and we're not importing that value for the `Inter` font ([reference](https://github.com/commercetools/merchant-center-application-kit/blob/main/packages/mc-html-template/html-docs/application.html#L32)).

_(we would also need to include it in the [Vistual Testing app](https://github.com/commercetools/merchant-center-application-kit/blob/main/website-components-playground/index.html#L11))_

emmenko (approved):
Thanks! 🙌

As discussed in our sync let's keep these special changes as a (hopefully) temporary implementation.

CarlosCortizasCT:
@NataliaSclipet Here we are only changing the UI of the selector but we have [another PR](https://github.com/commercetools/merchant-center-application-kit/pull/3312) where we also change its placement on the different pages.

I would suggest you make your review in the other PR (after this one is merged and the other one gets this changes) instead so you get the full picture and, if there's any styling issue with the selector, I can adjust it over there.

What do you think?

kark:
> @NataliaSclipet Here we are only changing the UI of the selector but we have [another PR](https://github.com/commercetools/merchant-center-application-kit/pull/3312) where we also change its placement on the different pages.
> 
> I would suggest you make your review in the other PR (after this one is merged and the other one gets this changes) instead so you get the full picture and, if there's any styling issue with the selector, I can adjust it over there.
> 
> What do you think?

We have just discussed that with @NataliaSclipet to do what @CarlosCortizasCT suggested, so I'll merge this PR now 🙂

PR #3312, opened 2023-11-27
Change Custom Views selector positioning

CarlosCortizasCT:
#### Summary

As part of the UI updates requested, we need to modify where the Custom Views selector is located in the different page layout we export from `app-kit`.

#### Description

The main two rules are:
* For tabular pages, we want the selector to be placed below the tabs selector (outside the page header)
* For the rest of the pages, we want the selector to be placed below the title/subtitle (within the header)

**NOTE**: The PR is ready to review although I'm not merging it until this one #3308 is merged before and I can update this branch with those changes.


packages/application-components/src/components/custom-views/custom-views-selector/custom-views-selector.tsx
```diff
@@ -19,18 +19,22 @@ type TCustomViewSelectorBaseProps = {
...
+  margin?: string;
 };
 type TCustomViewSelectorWithRequiredProps = TCustomViewSelectorBaseProps & {
   customViewLocatorCode: string;
+  margin?: string;
 };
 
 type TWrapperProps = {
   shouldRender: boolean;
+  margin?: string;
```
emmenko, line 31: Nit: isn't this required?

packages/application-components/src/components/internals/page-header.tsx
```diff
@@ -1,37 +1,52 @@
...
-        align-items: flex-end;
-        justify-content: space-between;
         margin: ${appKitDesignTokens.marginForModalPageHeader};
         padding: ${appKitDesignTokens.paddingForModalPageHeader};
         border-bottom: 1px solid
-          ${appKitDesignTokens.borderColorForModalPageHeaderDivider};
-        & > * + * {
-          margin-left: ${appKitDesignTokens.marginLeftForModalPageHeaderControls};
-        }
+          ${appKitDesignTokens.borderColorForModalPageHeaderDivider};/
```
emmenko, line 25:
```suggestion
          ${appKitDesignTokens.borderColorForModalPageHeaderDivider};
```

packages/application-components/src/components/main-pages/custom-form-main-page/custom-form-main-page.tsx
```diff
@@ -70,31 +64,52 @@ const CustomFormMainPage = (props: CustomFormMainPageProps) => {
...
+          </div>
+        )}
+        {!props.hideDivider && (
+          <div
+            css={css`
+              margin-top: ${uiKitDesignTokens.spacing40};
+            `}
+          >
+            <Divider />
+          </div>
```
emmenko, line 97: Are all these custom styles necessary? Can't we just wrap everything with `<Spacings.Stack>` so that all siblings have the same spacing between them? 🤔

packages/application-components/src/components/detail-pages/tabular-detail-page/tabular-detail-page.tsx
```diff
@@ -121,6 +120,12 @@ const TabularDetailPage = (props: TTabularDetailPageProps) => {
           }
         />
       </TabularPageContainer>
+      <CustomViewsSelectorWrapper>
+        <CustomViewsSelector
+          margin={`${uiKitDesignTokens.spacing30} 0 0 0`}
```
emmenko, line 125: Is there a different approach to avoid passing margin explicitly? Shouldn't margins / spacings be determined by the parent container?

I might be missing something... 🤔

emmenko (commented):
Thanks, looks good overall!

I'm a bit worried about all the custom style changes related to spacing / margin. Is there a way to avoid that?

packages/application-components/src/components/detail-pages/tabular-detail-page/tabular-detail-page.tsx
CarlosCortizasCT: That's what I first tried but I couldn't make it work. The main problem is that the `CustomViewsSelector` does not not if the component will be rendered at all as it depends on the request it will send to the backend in order to get the configured Custom Views for the provided locator. If there are not Custom Views configured, the `CustomViewsSelector` component will not be rendered but the consumer margins would be applied so an extra white space would be rendered.

packages/application-components/src/components/main-pages/custom-form-main-page/custom-form-main-page.tsx
CarlosCortizasCT: I had to remove the `Stack` component as it was introducing extra margins when the `CustomViewsSelector` does not render anything (no Custom views configured for the provided locator).

packages/application-components/src/components/custom-views/custom-views-selector/custom-views-selector.tsx
CarlosCortizasCT: From the component's perspective, I don't think its needed as it would depend on the consumers needs.

packages/application-components/src/components/internals/page-header.tsx
CarlosCortizasCT: Thanks!

Removed extra character here: 9c9da629a2c7a6c1210b23fd29c41f2dd9ef978f

packages/application-components/src/components/detail-pages/tabular-detail-page/tabular-detail-page.tsx
emmenko: If the `CustomViewsSelector` returns `null`, there shouldn't be any extra margin right?

Let's maybe clarify this in our sync

CarlosCortizasCT: The problem is not about a the use case for a  `CustomViewsSelector` which does not receive a locator but the use case when it does receive one but does not have any Custom Views configured: since in case we do receive anything we need to render a `div` with zero height to be able to run the animation, the extra space coming from the `Spacing.Stack` is then applied.

packages/application-components/src/components/main-pages/custom-form-main-page/custom-form-main-page.tsx
```diff
@@ -70,31 +66,54 @@ const CustomFormMainPage = (props: CustomFormMainPageProps) => {
 
   return (
     <PageWrapper>
-      <CustomViewsSelector
-        customViewLocatorCode={props.customViewLocatorCode}
-      />
       <MainPageContainer>
-        <Spacings.Stack scale="l">
```
kark, line 77: Small note: I'm cool with the change, but I'm not entirely clear on what the issue is or what I might be overlooking. It seems to me that if we stick with `<Spacings.Stack scale="l">`, the margins should still be applied correctly without the need to set them individually for each part.

|  with the selector rendered | without the selector  |
|---|---|
| Screenshot 2023-11-30 at 10 32 03  |  Screenshot 2023-11-30 at 10 30 40 |

I don't anticipate any changes—I'm good with it. Once it's updated to use the new selector styles, it's good to me.

CarlosCortizasCT: If I understand correctly what you suggest I guess that's something I wanted to keep but could not.
The problem is that keeping the `Spacing.Stack` component will introduce an unexpected margin then there are not custom views available. In this situation, the selector is rendered anyway with zero height because that's needed for the potential animation in case there are Custom views available. As there is a "ghost" HTML element, the `Spacing.Stack` introduces that unwanted margin.

Maybe this explanation is not good enough so I would be open for a call where I can also demonstrate the issue if you want to.

kark: Ah, I see now. I overlooked the transition.

kark (approved):
Thank you for all the style updates 🙇‍♂️ it looks good to me 👍

packages/application-components/src/components/main-pages/custom-form-main-page/custom-form-main-page.tsx
```diff
@@ -70,31 +66,54 @@ const CustomFormMainPage = (props: CustomFormMainPageProps) => {
...
-          {!props.hideDivider && <Divider />}
-        </Spacings.Stack>
+          </div>
+        )}
+        {!props.hideDivider && (
+          <div
+            css={css`
+              & > hr {
+                margin: ${headerRowMargin};
+              }
```
emmenko, line 97: With Emotion it should be possible to [target another Emotion component](https://emotion.sh/docs/styled#targeting-another-emotion-component)

```suggestion
              ${Divider} {
                margin: ${headerRowMargin};
              }
```

emmenko (approved):
Thanks! 🤗

packages/application-components/src/components/main-pages/custom-form-main-page/custom-form-main-page.tsx
CarlosCortizasCT: Thanks for the tip 👍 

Updated here: 69a8b4192cdf1a77c4351dcbd8ea8e57aca35b81

PR #3313, opened 2023-11-27
refactor(l10n): update CLDR data, add `en-BE` locale

emmenko:
There was a request from productboard 

```
We are looking to implement a business case where there are locale specific values for the European English
ecom website for some product attributes (e.g. Shoes and Clothing sizes).
We were hoping on using en-eu [...], we are going with en-be (Belgium) there.
[...]
Create additional selection in language settings: EN-EU, EN-BE
```

There is no such locale as `en-EU` (as `EU` is not a country) but there is [`en-150` (English Europe)](https://www.localeplanet.com/icu/en-150/index.html).
However, according to our API [we only support two-digit country codes](https://docs.commercetools.com/api/types#countrycode) which means that `150` is not valid.

It appears that `en-150` is essentially the equivalent to `en-BE` and therefore we can add support for that one.

**Why wasn't `en-BE` already supported?**

In the CLDR data for countries, the `BE` territory is marked as "secondary". Until now we have only been including the primary territories, not the secondaries.

I propose to keep an exception list of secondary countries that we want to include instead of allowing all of them.

### DIFF
--- a/packages/l10n/scripts/generate-l10n-data.js
+++ b/packages/l10n/scripts/generate-l10n-data.js
@@ -3,7 +3,7 @@ process.env.NODE_ENV = 'production';
 /* eslint-disable no-console */
 const fs = require('node:fs');
 const os = require('node:os');
-const path = require('path');
+const path = require('node:path');
 const chalk = require('chalk');
 const cldr = require('cldr');
 const deepDiff = require('deep-diff');


packages/l10n/data/languages/en.json
```diff
@@ -322,6 +322,10 @@
     "language": "English",
     "country": "Barbados"
   },
+  "en-BE": {
+    "language": "English",
+    "country": "Belgium"
+  },
```
emmenko, line 328: This is the new locale

packages/l10n/scripts/generate-l10n-data.js
```diff
@@ -23,13 +24,18 @@ const L10N_KEYS = {
...
 // but our API does not allow them. After some investigation with the
 // data files of the cldr library there is nothing for identifying them
 // since they are valid codes in the ISO 3166 and its following updates
 // https://www.drupal.org/project/drupal/issues/2036219
 const excludedCountries = ['QO', 'UN', 'ZZ'];
 
+// These are secondary territories that we want to include for a specific language.
+const includedSecondaryTerritories = {
+  en: ['BE'],
+};
```
emmenko, line 37: The "allow" list

packages/l10n/scripts/generate-l10n-data.js
```diff
@@ -199,16 +205,23 @@ const extractLanguageDataForLocale = (locale) => {
             },
           });
         }
+        const allTerritories = [
+          ...(languages[language]?.territories ?? []),
+          ...(languages[language]?.secondary?.territories ?? []).filter(
+            (territory) =>
+              includedSecondaryTerritories[language]?.includes(territory)
+          ),
+        ].sort();
```
emmenko, line 214: Here we merge the primary territories with the secondaries

packages/l10n/scripts/generate-l10n-data.js
```diff
@@ -1,7 +1,8 @@
 process.env.NODE_ENV = 'production';
 
 /* eslint-disable no-console */
-const fs = require('fs');
+const fs = require('node:fs');
+const os = require('node:os');
 const path = require('path');
```
tdeekens, line 6: Nit
```suggestion
const path = require('node:path');
```

tdeekens (approved):

PR #3315, opened 2023-11-29
Updating Ui-Kit dependencies

chloe0592:
I needed to update the UI-Kit dependencies manually because even though we are using the caret (`^`) when defining the versions in App-Kit, it didn't pick up the latest updates of the UI-Kit that I needed.


tdeekens (approved):

chloe0592:
I'm not sure why `lint_and_test` is failing. Does anyone has any hint perhaps? 🤔

kark:
> I'm not sure why `lint_and_test` is failing. Does anyone has any hint perhaps? 🤔

The problem is with the different versions of `react-select` in ui-kit and app-kit. Updating to `5.8.0` here should sort it out.

kark (approved):
Thank you!

PR #3322, opened 2023-12-04
Fix Custom Views panel UI

CarlosCortizasCT:
#### Summary

Fix Custom Views panel UI

#### Description

We realized we broke the UI of the Custom Views panel container when working in the new `Drawer` component so, instead of having a clean container with no predefined header:

Screenshot 2023-12-04 at 18 32 55

we were including a non-wanted header:

Screenshot 2023-12-04 at 18 33 42

On top of that, I took the opportunity to update `@emotion/react` dependency to match the one used in `ui-kit`.


packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -130,21 +136,24 @@ function CustomViewLoader(props: TCustomViewLoaderProps) {
   ].join('/');
 
   return (
-    <Drawer
+    <ModalPage
```
CarlosCortizasCT, line 139: This is the file with the relevant changes.

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -130,21 +136,24 @@ function CustomViewLoader(props: TCustomViewLoaderProps) {
...
       isOpen
-      title={`Custom View: ${props.customView.defaultLabel}`}
       onClose={props.onClose}
       size={panelSize === 'small' ? 10 : 30}
+      title=""
```
emmenko, line 143: Shouldn't we provide a title here?

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -130,21 +136,24 @@ function CustomViewLoader(props: TCustomViewLoaderProps) {
   ].join('/');
 
   return (
-    <Drawer
```
emmenko, line 133: Ah so we can't use the Drawer component anymore?

emmenko (approved):
Thanks!

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
CarlosCortizasCT: The issue is that the title is required in the `ModalPage` component and I didn't want to change that but we don't want to display any title in this use case.

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
CarlosCortizasCT: Not in this case.
The `Drawer` component will always use a header which is something we don't want in this use case.

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
emmenko: The `title` of the `ModalPage` is only used for accessibility if I'm not mistaken, it's not a visible value.

CarlosCortizasCT: So are you suggesting we should make that property optional in the `ModalPage` component?

emmenko: No I'm suggesting to use it 😄

PR #3325, opened 2023-12-05
Update NodeJS preferred version for Custom Extensions starter templates

CarlosCortizasCT:
#### Summary

The Custom Applications and Custom Views starter templates prefer the NodeJS `16.20` version, which is deprecated.

#### Description

We include a `.nvmrc` file in all the Custom Applications and Custom Views templates with the recommended NodeJS version to use them (when working with them locally).

The current version we have defined is no longer supported so we want to update to the next lowest maintained version, in order to have a supported one but not _force_ consumers to update to the latest version available.

Currently we are proposing to set `v18` as the preferred one.

image


emmenko (approved):
Thanks!

PR #3331, opened 2023-12-06
chore: allow drawer to expand height with a scrollbar

YahiaElTai:
When I first added the drawer, I added it with a bug sadly. 

When the drawer has long content the scrollbar doesn't appear to allow the user to see the content in the bottom of the drawer. This PR fixes that by adopting the same behavior from the higher level modal components


emmenko (approved):
Thanks 💯

PR #3336, opened 2023-12-11
Fix Custom Views panel width behaviour

CarlosCortizasCT:
#### Summary

Custom Views panels are overlapping navigation bar in narrow viewports.

#### Description

Custom Views panels can have a `small` or `large` size which are defined as a fixed width for them.

The issue we were having is that, by using only fixed values, we get to situations where viewport width is not enough so we run out of space and overlap things:

[Customers - test-project-with-sample-data - Merchant Center.webm](https://github.com/commercetools/merchant-center-application-kit/assets/97907068/6f189232-c3b7-4c54-a556-a69112457199)

The expected behaviour is for the Custom Views panels to have that fixed width as long as there's enough viewport space. Whenever it's not, it should behave like the other modal components and get to a max width. If the viewport is narrower, the panel itself must shrink.

[App-kit-playground - almond-40 - Merchant Center (7).webm](https://github.com/commercetools/merchant-center-application-kit/assets/97907068/4334dd7c-523c-4e85-b4d8-829b43f54e9c)

### DIFF
--- a/packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
+++ b/packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
@@ -1,4 +1,5 @@
 import { css, type SerializedStyles } from '@emotion/react';
+import { PORTALS_CONTAINER_INDENTATION_SIZE } from '@commercetools-frontend/constants';
 import { customProperties } from '@commercetools-uikit/design-system';
 import type { TModalPageSize } from './modal-page';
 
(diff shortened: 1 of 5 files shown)


packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
```diff
@@ -32,7 +33,11 @@ export const getContainerStyles = (props: StyleProps): SerializedStyles => css`
   right: 0;
   height: 100%;
   width: ${props.size !== 'scale'
-    ? `${stylesBySize[props.size].width} !important`
+    ? // In case we're using a specific size, we want it to be used until there's no space left.
+      // In such scenario, we want the modal to be as wide as possible, but using the shared indentation width size.
+      `min(${
```
CarlosCortizasCT, line 38: We use the minimum value between the expected size for the modal and the available space in its container (subtracting the indentation gap we use for modals).

emmenko (approved):
Many thanks! ❤️

kark (approved):
🙌

chloe0592 (approved):
Thanks @CarlosCortizasCT! Looks good to me 🙏

PR #3339, opened 2023-12-12
test(custom-views): custom view templates e2e tests

kark:
#### Summary

This PR ensures consistent behavior of the custom view template in local production mode (used in e2e tests via `start-server-and-test`) with local development mode. Additionally, it introduces e2e tests to validate both custom view templates in our CI.
Screenshot 2023-12-13 at 09 40 21


packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
```diff
@@ -67,6 +67,10 @@ function StrictModeEnablement(props: TStrictModeEnablementProps) {
   }
 }
 
+const isLocalProdMode =
+  process.env.NODE_ENV === 'production' &&
+  window.location.origin.match(/http:\/\/(localhost|127.0.0.1)/g);
```
CarlosCortizasCT, line 72: I guess this is mainly so the template runs with the project code built for production but the development host is also available to be able to actually interact with the Custom View, right?

Instead of relying in the origin, maybe can add another env variable when building the template which can be used in this file to decide to render the mock host.

What do you think? 🤔

kark: Thank you 🙏, 

> I guess this is mainly so the template runs with the project code built for production but the development host is also available to be able to actually interact with the Custom View, right?

That's exactly it.

In terms of the suggested alternative, it seems theoretically valid and feasible.
It's worth noting that this approach would set a new precedent, as we typically do not inject environment variables directly into the JavaScript runtime, aside from `NODE_ENV`. However, I checked that and we can update Webpack and Vite production config to do that.

emmenko: There is already a variable for such use cases: `MC_APP_ENV`.

On runtime you can access that via `window.app.env`.

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
```diff
@@ -101,6 +101,9 @@ function StrictModeEnablement(props: TStrictModeEnablementProps) {
   }
 }
 
+const isLocalProdMode =
```
CarlosCortizasCT, line 104: **nit**: Maybe we can leave a comment here explaining why we need this constant.

CarlosCortizasCT (approved):
Thanks a lot 👏

cypress/e2e/custom-view-template-starter/channels.cy.ts
```diff
@@ -0,0 +1,22 @@
...
+  ENTRY_POINT_CUSTOM_VIEW_TEMPLATE_STARTER,
+  URL_CUSTOM_VIEW_TEMPLATE_STARTER,
+} from '../../support/urls';
+
+describe('Channels', () => {
+  beforeEach(() => {
+    cy.loginToMerchantCenter({
+      entryPointUriPath: ENTRY_POINT_CUSTOM_VIEW_TEMPLATE_STARTER,
+      initialRoute: URL_CUSTOM_VIEW_TEMPLATE_STARTER,
+    });
```
emmenko, line 11: Nit: when developing Custom Views, some of these options are not really relevant for users.
Instead of explicitly passing these options we could maybe have them as default within the command.

For that we might want to add a new command e.g. `loginToMerchantCenterForCustomView` or something. Then the `entryPointUriPath` does not need to be requested and the "base path" can be predefined.

Something to think about for a follow up.

emmenko (approved):
Nice work, thanks!

cypress/e2e/custom-view-template-starter/channels.cy.ts
kark: Sure, great point, I'll work on that.

emmenko: Thanks, you can look into that for a follow up. No need to do it in this PR

PR #3334, opened 2023-12-07
fix: svgo plugin to use collision resistant IDs

emmenko:
> [!NOTE]
> Original PR #3332 but somehow some CI jobs were failing there 🤷‍♂️

Fixes issue where `svgo`'s `cleanupIds` default preset would cause id collisions between different svg files.

For more context see the [svgo docs for cleanupIds](https://svgo.dev/docs/plugins/cleanup-ids/), [this svgo issue](https://github.com/svg/svgo/issues/1746), and [this mc-frontend PR](https://github.com/commercetools/merchant-center-frontend/pull/15854).

PR #3342, opened 2023-12-12
fix(svgo/r config): update handling of prefixIds plugin to prefix svg ID's with a string hash of their filename

ByronDWall:
#### Summary

Updates the `SVGO` config of `SVGR` in the dev/prod `webpack` config and `vite-plugin-svgr` located in `mc-scripts`.

#### Description
Implements a further fix for SVG ID collision issues, modifying the initial approach taken in [this PR](https://github.com/commercetools/merchant-center-application-kit/pull/3334).

##### Previous implementation, what it was, why it didn't work as expected

Initially, a naive counter was implemented based on this [github comment](https://github.com/svg/svgo/issues/1746#issuecomment-1803600573). While that implementation insured id’s that are unique, it did not work in cases where the id is both declared and referenced in the same file, because the reference gets a separate unique ID (a different number from the counter), as shown in this screenshot:
svg-counter-prefix

Note how both highlighted ID's end in `1b`, but are prefixed with different numbers.  This is because the counter prefix has incremented for the ID in both the `filter` and the `g url()` - when it should be stable between the 2.

The previous implementation results in SVG's that do not display `mask`, `gradient`, `filter`, and other properties correctly, as shown in this screenshot (note the lack of drop-shadows, etc):
Screenshot 2023-12-12 at 12 42 29 PM

##### Current implementation, how it differs from previous, and why it works

This PR's implementation is based on this [github comment](https://github.com/svg/svgo/issues/913#issuecomment-369373572).  It generates a hash of the filepath of the svg file, resulting in a prefix which is:
- Short,
- With characters valid for IDs,
- The same within a file,
- And different in different files.

The following screenshot is the DOM output for the same SVG as above, and shows that the ID declaration and reference refer to the same ID (with a suffix of `1b`):
svg-hashed-prefix

This implementation results in SVG's that are able to exist inline on the page without ID collisions, and _also_ have internally consistent ID's, resulting in SVG's displaying `mask`, `gradient`, `filter`, etc properties as-expected (note the drop-shadows, etc):
Screenshot 2023-12-12 at 12 44 18 PM

The webpack loader for this implementation was tested in `mc-frontend` by using a local build of `mc-scripts` linked to `mc-frontend` using pnpm's `file:` link syntax, as [shown here](https://stackoverflow.com/a/72106897)


stephsprinkle (approved):

tdeekens (commented):
Thanks for the in detail description. The general solution of a content addressable hash makes sense. The string-hash library however was published 7 years ago and uses a quite particular hashing algorithm. In those years a lot has changed in Node.js and maybe I missed it but can we avoid a dependency altogether and use crypto as a built in library?

```js
const crypto = require('crypto')

const str = 'This is the string I want to hash.'
const hash = crypto.createHash('sha1').update(str).digest('hex')
```

The string would eg be a readFileSync while we would split the hash to 0-6 for instance.

ByronDWall:
> can we avoid a dependency altogether and use crypto as a built in library?

@tdeekens thanks for the review, yes we sure can.  I've refactored to use `crypto`, and made sure that the webpack loader is working as-expected locally in `mc-frontend`, here is the screenshot:
Screenshot 2023-12-12 at 4 07 25 PM


Thanks for the review, I scheduled a message to you and Nicola about this to send tomorrow morning CET, but you've already reviewed so enjoy the stale message I guess.

tdeekens (approved):

CarlosCortizasCT (approved):
Thanks a lot for the fix and the detailed explanation 👍

emmenko (approved):
🤯

Thanks!

PR #3343, opened 2023-12-12
Fix Custom Views issues

CarlosCortizasCT:
#### Summary

In this PR we're tackling some last minute issues found around Custom Views rendering.

#### Description

These are the issues addressed in this PR:
* Unstyled Page Layout components when rendered within a Custom View
* Routing issues when navigating within a Custom View
* Custom View not including `PortalsContainer` component (required for modals, dialogs,...).

Regarding the last point, I included the component so it does not break the Custom View if a `Dialog`, `Modal` is used, but I don't think this is the solution we want.

Here's an example of how it looks like:

Screenshot 2023-12-12 at 20 51 46

I think it would be better for Custom Views to have a different mechanism for modals and dialogs that we should discuss.

### DIFF
--- a/.changeset/olive-pots-raise.md
+++ b/.changeset/olive-pots-raise.md
@@ -1,0 +1,6 @@
+---
+'@commercetools-frontend/application-components': patch
+'@commercetools-frontend/application-shell': patch
+---
+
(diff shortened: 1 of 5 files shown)


emmenko (approved):
Thanks, changes fix the issues as expected 💯

PR #3347, opened 2023-12-14
Fix Page Layout components scrolling issue

CarlosCortizasCT:
#### Summary

Fix Page Layout components scrolling issue

#### Description

This is a follow-up of #3343, where we updated some styles for page layout components to address a problem when rendering them inside a Custom View which makes them incorrectly manage scroll. Instead of the header to be fixed and the content scrollable, the whole page component is scrollable. It was spotted thanks to MC frontend e2e tests.

The right solution is to include some global styles (that we are including in the host applications) also in the Custom View content iframe.

### DIFF
--- a/packages/application-components/src/components/internals/page.styles.ts
+++ b/packages/application-components/src/components/internals/page.styles.ts
@@ -3,7 +3,7 @@ import { designTokens as appKitDesignTokens } from '../../theming';
 
 export const ContentWrapper = styled.div`
   flex: 1;
-  flex-basis: 0%;
+  flex-basis: 0;
   margin: ${appKitDesignTokens.marginForPageContent};
   overflow: auto;
 `;
--- a/packages/application-components/src/components/main-pages/internals/main-page.styles.ts
+++ b/packages/application-components/src/components/main-pages/internals/main-page.styles.ts
@@ -13,7 +13,7 @@ export const Divider = styled.hr`
 
 export const MainPageContent = styled.div`
   flex: 1;
-  flex-basis: 0%;
+  flex-basis: 0;
   overflow: auto;
   // NOTE: do not change to "padding" as this breaks sticky DataTable styles
   margin: ${appKitDesignTokens.marginForPageContent};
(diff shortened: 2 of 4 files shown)


packages/application-components/src/components/internals/page.styles.ts
```diff
@@ -3,7 +3,7 @@ import { designTokens as appKitDesignTokens } from '../../theming';
 
 export const ContentWrapper = styled.div`
   flex: 1;
-  flex-basis: 0%;
+  flex-basis: 0;
```
CarlosCortizasCT, line 6: Revert the change introduced in the previous PR.

packages/application-components/src/components/main-pages/internals/main-page.styles.ts
```diff
@@ -13,7 +13,7 @@ export const Divider = styled.hr`
 
 export const MainPageContent = styled.div`
   flex: 1;
-  flex-basis: 0%;
+  flex-basis: 0;
```
CarlosCortizasCT, line 16: Revert the change introduced in the previous PR.

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
```diff
@@ -184,58 +185,61 @@ function CustomViewShell(props: TCustomViewShellProps) {
...
-              hostUrl={hostUrl}
-              customViewConfig={hostContext.customViewConfig}
-            >
-              <CustomViewShellAuthenticated
-                dataLocale={hostContext.dataLocale}
-                environment={window.app}
-                messages={props.applicationMessages}
-                projectKey={hostContext.projectKey}
+    <>
+      <GlobalStyles />
```
CarlosCortizasCT, line 189: Basically we only include this `GlobalStyles` component which includes a 100% height for the main HTML document content.

emmenko: Ah good catch, thanks!

emmenko (approved):

kark (approved):
Thank you!

PR #3350, opened 2023-12-14
docs(custom views): e2e testing

kark:
**Previews:**

- https://merchant-center-application-7rxxyuary-commercetools.vercel.app/custom-applications/api-reference/commercetools-frontend-cypress
- https://appkit-cv-pr-3350.commercetools.vercel.app/custom-views/development/testing
- https://appkit-cv-pr-3350.commercetools.vercel.app/custom-views/development/api-reference/commercetools-frontend-cypress

#### Summary

Screenshot 2023-12-18 at 09 34 29


packages/cypress/src/task/index.ts
```diff
@@ -47,50 +45,54 @@ const loadEnvironmentVariables = (
...
+
         console.log(
-          `Found Custom Application config for ${packageInfo.packageJson.name}`
+          `Found Custom ${
+            isCustomViewConfig ? 'View' : 'Application'
+          } config for ${packageInfo.packageJson.name}`
         );
         return {
           ...allConfigs,
           [processedConfig.env.entryPointUriPath]: processedConfig.env,
```
kark, line 74: This function traverses the monorepo packages, loading all custom app/view configs. 
These configs are cached using the `entryPointUriPath` as the key, which implies that custom view configs may overwrite one another. Currently, both custom view packages  ([starter](https://github.com/commercetools/merchant-center-application-kit/blob/main/custom-views-templates/starter/custom-view-config.mjs) and [starter-typescript](https://github.com/commercetools/merchant-center-application-kit/blob/main/custom-views-templates/starter-typescript/custom-view-config.mjs)) have identical configurations. Consequently, there is no distinction between them (and it's not necessary at this point). I'm unsure about the best approach to address this issue and if anything should be done at this point

emmenko: Ah good point, we can't really use the `entryPointUriPath` here 🤔

What if we use the workspace path? It can also be optional as a way to narrow down the search in case there are multiple matches for the same entry point.

This might require some more refactoring though. I'm also not sure if Custom Applications will need it or not.

website-custom-views/src/content/development/testing.mdx
```diff
@@ -235,3 +235,84 @@ const renderView = (options = {}) => {
   });
 };
 ```
+
+# End-to-End
```
CarlosCortizasCT, line 239: Since we're adding docs here, maybe we should include @Anshuman71 as a reviewer?

packages/cypress/src/task/index.ts
CarlosCortizasCT: If I understand correctly the problem, this is something which only affects our internal e2e tests and not consumers of this package, right?

If that's the case, I would suggest to keep what we currently have and investing time only if we have several different templates with different configurations in the future.

kark: > If I understand correctly the problem, this is something which only affects our internal e2e tests and not consumers of this package, right?

Yes, for the time being, we are on the safe side. We won't be affected until we introduce e.g. in the `mc-frontend` more than one Internal Custom View that needs testing with Cypress.

emmenko: >  this is something which only affects our internal e2e tests and not consumers of this package, right?

Potentially anyone with a monorepo and multiple Custom Views will be affected.

I'm fine with looking into that as a follow up.

CarlosCortizasCT: I'm sorry but I don't get how consumers use this package so they would be affected. 

I was asking because it is internally using the [@manypkg/get-packages](https://www.npmjs.com/package/@manypkg/get-packages) which supports some monorepo tools, but not other (for instance, it does not seem to support [Nx](https://nx.dev/) or [Turborepo](https://turbo.build/)).

image

emmenko: If someone wants to develop multiple Custom Views within the same repository and they want to write E2E tests for it, they will use our Cypress task and they will bump into the issue that Kacper found.

kark: I've pushed a PoC for a proposal [here](https://github.com/commercetools/merchant-center-application-kit/pull/3356) and a discussion and any alterations are most welcome (I can resume working on that when I'm back). I'm sure there are alternative ways to tackle this. One drawback of the proposal is that the template directory must have the same name as the template itself. Your insights and suggestions are highly appreciated!

kark: This issue has been addressed in 398c5bd58461b50a381eefe4168982118e2ec3eb

website/src/content/development/testing.mdx
```diff
@@ -298,7 +298,7 @@ The `.env` file should be git ignored. On CI, you can define environment variabl
 In the `cypress/plugins/index.js` file, you need to configure the `task` for Custom Applications and to load the environment variables:
 
 ```js title="cypress/plugins/index.js"
-const customApplications = require('@commercetools-frontend/cypress/task');
+const { customApplicationConfig } = require('@commercetools-frontend/cypress/task');
```
Anshuman71, line 301: @kark will be a breaking change for users already using this API? If so, we should include a release note.

website-custom-views/src/content/api-reference/commercetools-frontend-cypress.mdx
```diff
@@ -73,37 +73,24 @@ import '@commercetools-frontend/cypress/add-commands';
...
 
 ### Options
 
 Available options are:
 
-* `entryPointUriPath` (**required**): User the constant for the entry point URI path: `CUSTOM_VIEW_HOST_ENTRY_POINT_URI_PATH`.
-* `dotfiles` (optional): A list of dotenv files to load when the `custom-view-config.mjs` is loaded (in case you're using an environment placeholder). By default the following dotenv files are loaded: `.env` and `.env.local`. You can also define the values using paths relative to the application folder.
-* `initialRoute` (optional): The route to open after login. If not defined, make sure to call `cy.visit` yourself.
+* `packageName` (optional): The package name as specified in the Custom View's `package.json`. If not provided, it defaults to `Cypress.env('PACKAGE_NAME').`
+* `dotfiles` (optional): A list of dotenv files to load when the `custom-view-config.mjs` is loaded (in case you're using an environment placeholder). By default the following dotenv files are loaded: `.env` and `.env.local`. You can also define the values using paths relative to the Custom View folder.
```
Anshuman71, line 93: @kark what do you mean by "environment placeholder" here? Can you give an example, currently it sounds ambiguous.

Anshuman71: Hi I found what you mean, I added the link (/api-reference/custom-view-config#using-variable-placeholders). let me know if that's correct.

website/src/content/development/testing.mdx
Anshuman71: Hi @kark, looks like the legacy export will solve the breaking change. Can you confirm this with a manual test? if it's good then we don't need a release note.

kark: Thanks, of course I tested it manually.

Steps to follow test it out:

1. Checkout the `SHIELD-993-custom-view-e2e-testing-docs` and apply changes:
Screenshot 2024-01-18 at 13 56 24

2. Build monorepo packages
3. Run `pnpm template-custom-application-starter:start:prod:local` in one terminal
4. Run `pnpm test:e2e:template-custom-application-starter` in another terminal

Error encountered without the legacy export:
Screenshot 2024-01-18 at 14 26 43

With the legacy export the tests work correctly.

Anshuman71: Yayy! Thanks @kark for going all the way 🙌🏼

website-custom-views/src/content/development/testing.mdx
```diff
@@ -235,3 +235,84 @@ const renderView = (options = {}) => {
...
+
+* [`@testing-library/cypress`](https://testing-library.com/docs/cypress-testing-library/intro/): provides commands to select elements using React/Dom Testing Library.
+* `@commercetools-frontend/cypress`: provides commands specific to Custom Views.
+
+Define a `.env` file in the `cypress` folder, containing some of the required environment variables:
+
+```bash title="cypress/.env"
+CYPRESS_LOGIN_USER=""
+CYPRESS_LOGIN_PASSWORD=""
+CYPRESS_PROJECT_KEY=""
```
CarlosCortizasCT, line 255: Maybe we can include here the `CYPRESS_PACKAGE_NAME` variable.

website-custom-views/src/content/development/testing.mdx
```diff
@@ -235,3 +235,84 @@ const renderView = (options = {}) => {
...
+    customViewConfig,
+  });
+
+  return {
+    ...cypressConfig,
+    env: {
+      ...cypressConfig.env,
+      LOGIN_USER: process.env.CYPRESS_LOGIN_USER,
+      LOGIN_PASSWORD: process.env.CYPRESS_LOGIN_PASSWORD,
+      PROJECT_KEY: process.env.CYPRESS_PROJECT_KEY,
```
CarlosCortizasCT, line 287: Maybe we can include here the `CYPRESS_PACKAGE_NAME` variable.

website/src/content/development/testing.mdx
Anshuman71, line 301: @kark do you think we can revert changes of this file now?

Anshuman71: Hi @kark, I thought keeping this would be fine. But I see your point about the mixed imports. And we want to recommend users to use named exports instead. Then, let's leave the change like this.

website-custom-views/src/content/development/testing.mdx
kark: Right, thanks for noticing 👍

Anshuman71 (approved):
Approving the changes. 
**Note**: The Cypress changes are compatible with Custom Apps and don't break existing user workflows. Hence, no release note has been added.

website-custom-views/src/content/api-reference/commercetools-frontend-cypress.mdx
```diff
@@ -73,37 +71,24 @@ import '@commercetools-frontend/cypress/add-commands';
...
 # Commands
 
-## cy.loginToMerchantCenter
+## cy.loginToMerchantCenterForCustomView
 
-This command performs the user login to the Merchant Center. It automatically detects whether the Custom View is running on localhost or production and chooses the appropriate login mechanism.<br/>
-
-- When it runs locally, the same mechanism used in the `cy.loginByOidc` is used.
-- When it runs on production, a normal login flow is used where the user enters the credentials into the login form.
+This command performs the user login to the Merchant Center to test a Custom View. Testing Custom Views is expected to occur on `localhost`, therefore the `cy.loginByOidc` mechanism is used.
```
CarlosCortizasCT, line 76: I'm wondering whether it makes sense to mention the `cy.loginByOidc` command here since we don't talk about it in the Custom Views docs (it's more like an internal detail to me).

PR #3351, opened 2023-12-15
refactor(constants): expose regex

emmenko:
To have a single source of truth for validation regex that we use in the MC apps.

### DIFF
--- a/packages/constants/src/constants.ts
+++ b/packages/constants/src/constants.ts
@@ -1,5 +1,24 @@
 import type { TCustomView } from './types/generated/settings';
 
+/**
+ * The entryPointUriPath may be between 2 and 36 characters and only contain alphabetic lowercase characters,
+ * non-consecutive underscores and hyphens. Leading and trailing underscore and hyphens are also not allowed.
+ */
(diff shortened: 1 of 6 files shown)


packages/constants/src/constants.ts
```diff
@@ -1,5 +1,24 @@
 import type { TCustomView } from './types/generated/settings';
 
+/**
+ * The entryPointUriPath may be between 2 and 36 characters and only contain alphabetic lowercase characters,
+ * non-consecutive underscores and hyphens. Leading and trailing underscore and hyphens are also not allowed.
+ */
+export const PROJECT_KEY_REGEX = /^[^-_#]([0-9a-z]|[-_](?![-_])){0,34}[^-_#]$/g;
```
emmenko, line 7: This is new and more strict that the one we currently have, as it allows several use cases that I believe should be allowed (see internal discussions).

It's based off on the entry point regex.

kark (approved):

CarlosCortizasCT (approved):
Thanks 👍

tdeekens (approved):

PR #3352, opened 2023-12-15
Move Custom Views panel paddings from loader component to shell component

CarlosCortizasCT:
#### Summary

Move Custom Views panel paddings from the `CustomViewLoader` component to the `CustomViewShell` component

#### Description

When trying to use `Dialog` components within a Custom View, we noticed a problem with the overlay not using the whole space of the Custom View panel:

Screenshot 2023-12-12 at 20 51 46

This is due to the fact that the actual Custom View components are rendered within the `CustomViewLoader` component which includes some paddings to make sure Custom Views contents have some white space around.
The problem is that the `iframe` where the Custom Views contents are rendered is inside those paddings, so when rendering a dialog overlay, which renders within the `iframe`, it has no way to use the whole Custom View panel space.

To address this situation we moved the paddings from the `CustomViewLoader` component to the `CustomViewShell` one. The latter is the root component of a Custom View contents rendered within the `iframe` so now the dialog, can use the whole space.

There's one challenge though regarding the close icon because it was rendered from the `CustomViewLoader` which is the one that controls the visibility of the panel. Now that the padding have been moved to the contents of the panel, we need to render that icon from within the `iframe`. 
Similar to what we did moving the paddings to the `CustomViewShell` component, we also moved the close icon, although now the action (happening within the `iframe`) can't directly close the panel, as it is controlled from a component in the host application.
To address, we implemented a new message from the `iframe` to the host application (using the communication channel we are already using for the other direction messages) to ask for the latter to close the panel.

Screenshot 2023-12-15 at 12 51 13

Even though that's working fine, I've found a little issue when using dialogs in `small` Custom Views panels: dialogs have a `size` property to control its width and, its smallest value, it's too wide for a `small` Custom View:

Screenshot 2023-12-15 at 12 51 36

There's also a value of `scale` which should adapt to the available horizontal space, but when using it, we're also using all vertical available space:
Screenshot 2023-12-15 at 12 52 18

I'm not sure if that was made on purpose and I'd like to ask about it.


packages/application-components/src/index.ts
```diff
@@ -13,6 +13,7 @@ export { default as InfoModalPage } from './components/modal-pages/info-modal-pa
 export { default as FormModalPage } from './components/modal-pages/form-modal-page';
 export { default as TabularModalPage } from './components/modal-pages/tabular-modal-page';
 export { default as CustomFormModalPage } from './components/modal-pages/custom-form-modal-page';
+export { default as ModalPageTopBar } from './components/modal-pages/internals/modal-page-top-bar';
```
CarlosCortizasCT, line 16: Now we need to use this component from the `CustomViewShell` component but it's weird to export an `internal` component.

What do you think about it?

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.spec.tsx
```diff
@@ -129,12 +129,12 @@ describe('CustomViewLoader', () => {
...
     const onCloseMock = jest.fn();
 
-    renderComponent(
+    const { baseElement } = renderComponent(
       <CustomViewLoader customView={TEST_CUSTOM_VIEW} onClose={onCloseMock} />
     );
 
-    const closeButton = screen.getByLabelText('Close Modal Page');
-    fireEvent.click(closeButton);
+    const overlay = baseElement.querySelector('[data-role="modal-overlay"]');
```
CarlosCortizasCT, line 136: Since the close icon is not rendered from this component anymore, now I'm using the overlay to trigger the `onClose` action.

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -58,11 +56,18 @@ function CustomViewLoader(props: TCustomViewLoaderProps) {
...
-    if (event.data.origin === window.location.origin) {
-      console.log('message received from iframe port: ', event);
-    }
-  }, []);
+  const messageFromIFrameHandler = useCallback(
+    (event: MessageEvent) => {
+      if (event.data.origin === window.location.origin) {
+        switch (event.data.eventName) {
+          case CUSTOM_VIEWS_EVENTS_NAMES.CUSTOM_VIEW_CLOSE:
+            props.onClose();
```
emmenko, line 64: 💯

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
```diff
@@ -94,6 +98,11 @@ function NotificationsContainer(props: TNotificationsContainerProps) {
   );
 }
 
+const ContentWrapper = styled.div`
+  height: 100%;
+  padding: ${designTokens.spacing40} 40px;
```
emmenko, line 103: Wondering if we should keep the shell container without margins / paddings.

Users can "easily" do that by using our page components or spacings components.

For instance, it would be tricky then for users to "remove" the double spacing rather than adding one.

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
```diff
@@ -218,7 +237,8 @@ function CustomViewShell(props: TCustomViewShellProps) {
                   <Route
                     path={`/custom-views/${hostContext.customViewConfig.id}/projects/${hostContext.projectKey}`}
                   >
-                    {props.children}
+                    <ModalPageTopBar onClose={handleClose} hidePathLabel />
+                    <ContentWrapper>{props.children}</ContentWrapper>
```
emmenko, line 241: I guess this results in the extra space at the top that we see here:

image

Wondering maybe if we should define the close button as a fixed/absolute position to not affect the page layout? 🤔

Then we don't need to expose `<ModalPageTopBar>`.

emmenko (commented):
Thanks for following up on this! 🙌

emmenko:
@CarlosCortizasCT 

> There's also a value of `scale` which should adapt to the available horizontal space, but when using it, we're also using all vertical available space:
> I'm not sure if that was made on purpose and I'd like to ask about it.

Yes that was on purpose.

Have you tried using the dialog size `7` (currently the smaller one)?

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
CarlosCortizasCT: That's an option, although it would force consumers to use one of our Page Layout components which I was trying to avoid so developers can decide what they want to render as the panel's content.

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
CarlosCortizasCT: That would be an issue for notifications, which should be rendered before the close icon and push it down.

CarlosCortizasCT:
> @CarlosCortizasCT
> 
> > There's also a value of `scale` which should adapt to the available horizontal space, but when using it, we're also using all vertical available space:
> > I'm not sure if that was made on purpose and I'd like to ask about it.
> 
> Yes that was on purpose.
> 
> Have you tried using the dialog size `7` (currently the smaller one)?

Yes, size `7` is the one I use for the screenshot in the description:

> Even though that's working fine, I've found a little issue when using dialogs in `small` Custom Views panels: dialogs have a `size` property to control its width and, its smallest value, it's too wide for a `small` Custom View:

Screenshot 2023-12-15 at 12 51 36

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
emmenko: Ok, let's maybe have a sync about it when Natalia is back and see what we can do about it.

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
emmenko: It's the same for Custom Applications, we don't use spacing in the main container.

The issue I see is that when people use the page components they will end up with double spacing, which is impossible for them to get rid of.
On the other hand, if they just render content they will see right away that spacing is missing and they can fix that by adding a container (either a page component or even spacing inset).

CarlosCortizasCT: Yes, you're right we do have the double spacing issue which we have created another task to address it. I was thinking in having some data attribute in the top of the document that page layout component can use to know which css spacing they need to use.

I understand your point although we would also have to think this taking the close icon and the notifications into account.

If this is something you think should block us here, then I'll keep this PR open for a couple of weeks, but we can also try to address one problem at a time, as we were trying to and maybe have a sync with Natalia when she comes back to see how we can improve things?

What do you think?

emmenko: No blocking, we can take it step by step yes.

Let's regroup then when Natalia is back. Thanks

packages/application-components/src/index.ts
emmenko: Are you still ok in releasing this change?

emmenko (approved):
As we agreed on looking at some of the points raised here as a follow up, the changes here are good to go from my side.

Thanks! 🤗

PS: add a changeset

kark (approved):
Thank you for the solution 🎉 
Everything looks good - just a tiny reminder about a changeset.

packages/application-components/src/index.ts
kark: `2` sounds good or `1` and moving the export down to `// Utilities`

emmenko: I'm fine with any option, up to you to decide. Thanks!

CarlosCortizasCT: I went for the second option.

Updated here: 906dfe4150dcd267dabbec87c1c50485754a5300

PR #3353, opened 2023-12-18
Refactor Page Layout components to render correctly in Custom Views panels

CarlosCortizasCT:
#### Summary

Refactor Page Layout components to render correctly in Custom Views panels

#### Description

When rendering this kind of component (eg: `InfoMainPage`, we noticed we were introducing a "double padding" effect since the Custom Views panel itself already contains some paddings to make sure, no matter what content is rendered inside, the contents have some white space around and these kind of components also render some surrounding paddings/margins.

To avoid this unwanted UI, we now add an attribute to the `body` element of the Custom View panel iFrame we render inside it and use it to adjust the CSS rules of those components when they are rendered in such context.

**BEFORE**:
image

**AFTER**:
image


packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
```diff
@@ -140,6 +142,13 @@ function CustomViewShell(props: TCustomViewShellProps) {
     []
   );
 
+  useLayoutEffect(() => {
```
CarlosCortizasCT, line 145: This is the simplest solution.

I think I would prefer to update how we generate the HTML document when running/building the application to already include this attribute but I was worried that end-users would need to make sure they update all `app-kit` dependencies at the same time, and last year we learnt that's not always the case.

Pelase, let me know waht you think 🙏

CarlosCortizasCT:
> We decided to block this PR as the proposed change is not agreed by the team and we need to have a discussion between engineers and designers in order to come to an alignment.

@emmenko and I have been discussing about the current approach and, although he's concern on the maintainability costs this implementation might have in the future, we agreed that the balance in terms of added complexity and developer experience is fair enough as we also don't see better approaches that do not imply either degrading DX or including some complexity anyway.

We discussed options like:
* remove default paddings for Custom Views panels: will force users to add them somehow and we have the problem with the positioning of the close icon and its integration with notifications or even modals
* remove the paddings only for large Custom Views panels: we would still have the problems as before and DX will be a little more degraded as developers would need to learn different approaches based on the Custom View panel size.

At the end of the day, we think adding this extra complexity it worth the DX os developers not needing to do anything special in terms of what components and paddings to use no matter the Custom View panel they's using (or if they are using page layout components in Custom Applications or Custom Views).

packages/application-components/src/components/internals/page-header.tsx
```diff
@@ -17,12 +18,18 @@ type Props = {
...
+    // Bear in mind the paddings are dependant on the context the component
+    // is rendered in. For example, when rendered in a custom view, its panel
+    // already contains default paddings, so we don't need to add them here.
     <div
       css={css`
         margin: ${appKitDesignTokens.marginForModalPageHeader};
         padding: ${appKitDesignTokens.paddingForModalPageHeader};
         border-bottom: 1px solid
           ${appKitDesignTokens.borderColorForModalPageHeaderDivider};
+        body[data-extension-type='${CUSTOM_EXTENSION_TYPES.CUSTOM_VIEW}'] & {
```
emmenko, line 30: This is a bit "surprising" to me that we can target global elements within a scoped style.

I would suggest to follow this approach (@ByronDWall actually proposed it 😅): https://github.com/emotion-js/emotion/issues/2836#issuecomment-1406917978

```tsx
css`
  * :where([data-extension-type="custom-view"]) & {
    margin: ${appKitDesignTokens.paddingForModalPageHeaderInCustomView};
  }
`
```

emmenko (commented):
Some suggestions, otherwise (as you mentioned) I'm ok with the approach. 🤗

packages/application-components/src/components/internals/page.styles.ts
```diff
@@ -1,11 +1,18 @@
...
 import { designTokens as appKitDesignTokens } from '../../theming';
 
 export const ContentWrapper = styled.div`
   flex: 1;
   flex-basis: 0;
-  margin: ${appKitDesignTokens.marginForPageContent};
   overflow: auto;
+  // Bear in mind the margins are dependant on the context the component
+  // is rendered in. For example, when rendered in a custom view, its panel
+  // already contains default paddings, so we don't need to add margins here.
```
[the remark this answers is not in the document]
CarlosCortizasCT: Updated here: cf1c6e521349c4a07621359d7b5c186cf97e4506

However, I checked the built bundles and I couldn't find those comments, although I could have missed something.

packages/application-components/src/components/internals/page-header.tsx
CarlosCortizasCT: Updated here: cf1c6e521349c4a07621359d7b5c186cf97e4506

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
CarlosCortizasCT: Thanks. That's a better approach for our current use case.

Updated here: 56946763ba1c249bf2c904492b35a4b00ffa2f22

packages/application-components/src/components/internals/tabular-page.tsx
```diff
@@ -58,7 +67,13 @@ const FormControlsContainer = styled.div`
 `;
 
 const CustomViewsSelectorWrapper = styled.div`
+  // Bear in mind the margins are dependant on the context the component
+  // is rendered in. For example, when rendered in a custom view, its panel
+  // already contains default paddings, so we don't need to add margins here.
```
emmenko, line 72: Move comments outsides of CSS?

packages/application-components/src/components/internals/tabular-page.tsx
```diff
@@ -58,7 +67,13 @@ const FormControlsContainer = styled.div`
...
   margin: ${appKitDesignTokens.marginForCustomViewsSelectorAsTabular};
+  body[data-extension-type='custom-view'] & {
+    margin: ${appKitDesignTokens.marginForCustomViewsSelectorAsTabularInCustomView};
+  }
```
emmenko, line 76: I guess this also needs to be adjusted

packages/application-components/src/components/main-pages/internals/main-page.styles.ts
```diff
@@ -15,6 +22,12 @@ export const MainPageContent = styled.div`
   flex: 1;
   flex-basis: 0;
   overflow: auto;
-  // NOTE: do not change to "padding" as this breaks sticky DataTable styles
+  // NOTE: do not change to "margin" as this breaks sticky DataTable styles
+  // Bear in mind the margins are dependant on the context the component
+  // is rendered in. For example, when rendered in a custom view, its panel
+  // already contains default paddings, so we don't need to add margins here.
```
emmenko, line 28: Here as well

packages/application-components/src/theming.ts
```diff
@@ -37,22 +37,28 @@ export const themesOverrides = {
...
     marginForCustomViewsSelectorAsTabular: `0 ${appKitSpacing55}`,
+    marginForCustomViewsSelectorAsTabularInCustomView: '0',
     marginForDialogContainerContents: `${uiKitDesignTokens.spacing30} 0 ${uiKitDesignTokens.spacing50} 0`,
     marginForModalPageHeader: `0 ${appKitSpacing55}`,
     marginForUserMenuItem: `${uiKitDesignTokens.spacing10} 0`,
     marginLeftForModalPageHeaderControls: uiKitDesignTokens.spacing50,
     marginRightForAppbar: appKitSpacing55,
     marginTopForDialogFooter: uiKitDesignTokens.spacing50,
     marginForPageContent: `${uiKitDesignTokens.spacing50} ${appKitSpacing55}`,
+    marginForPageContentInCustomView: `${uiKitDesignTokens.spacing50} 0`,
```
emmenko, line 48: As an alternative idea / approach, would it work if we override the css variables (theming) to change the margins instead of doing that explicitly in each component? 🤔

So basically in the CustomViewShell we inject styles in the in wrapper and override the css variables.

What do you think?

kark (approved):
Appreciate you providing insights into the decision-making process; everything looks good. Just a friendly reminder to include a changeset before merging. Thanks!

emmenko (commented):
Thanks for the improvements, glad that it worked! 🤗

packages/application-components/src/components/internals/tabular-page.tsx
```diff
@@ -58,7 +67,13 @@ const FormControlsContainer = styled.div`
 `;
 
 const CustomViewsSelectorWrapper = styled.div`
+  // Bear in mind the margins are dependant on the context the component
+  // is rendered in. For example, when rendered in a custom view, its panel
+  // already contains default paddings, so we don't need to add margins here.
   margin: ${appKitDesignTokens.marginForCustomViewsSelectorAsTabular};
+  body[data-extension-type='custom-view'] & {
```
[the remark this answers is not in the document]
CarlosCortizasCT: I forgot about that one. Sorry.

Updated here: 129d8b9fd8971c2518527446da20ee929eacaa27

packages/application-components/src/components/internals/tabular-page.tsx
CarlosCortizasCT: I forgot about that one. Sorry.

Updated here: 129d8b9fd8971c2518527446da20ee929eacaa27

packages/application-components/src/components/internals/tabular-page.tsx
CarlosCortizasCT: I forgot about that one. Sorry.

Updated here: 64f6bef6a0578c2b403344243f5b566d520f363b

packages/application-components/src/components/main-pages/internals/main-page.styles.ts
CarlosCortizasCT: I forgot about that one. Sorry.

Updated here: 64f6bef6a0578c2b403344243f5b566d520f363b

packages/application-components/src/theming.ts
CarlosCortizasCT: Thanks @emmenko. That's a great idea 🔝 

The only drawback that I can see is that maybe it's a little more complicated to spot where things are happening in terms of maintanability.
With the new approach, the Page Layout components have a different UI when they are rendered within a Custom View Panel but it is not pretty straightforward why as the code which makes that happen is not in those components but in a new file (`custom-view-shell.styles.ts`).

I've opened [this PR](https://github.com/commercetools/merchant-center-application-kit/pull/3378) with the refactoring so we can all take a look at how it would look like before applying those changes here.

@commercetools/shield-team-fe it would be great to read your thoughts 🙏

emmenko: Thanks Carlos. I left a review in the other PR 🤗

CarlosCortizasCT: I merged the other PR into this one so we have your idea implemented here.

packages/application-components/src/components/main-pages/internals/main-page.styles.ts
```diff
@@ -15,6 +15,5 @@ export const MainPageContent = styled.div`
   flex: 1;
   flex-basis: 0;
   overflow: auto;
-  // NOTE: do not change to "padding" as this breaks sticky DataTable styles
```
emmenko, line 18: Nit: could you keep and move this comment outside of the styles (L13)?

emmenko (approved):
Thanks for looking into different options and for the constructive discussions ❤️

I feel more confident now with this approach, so good to go from my side 🚀

packages/application-components/src/components/main-pages/internals/main-page.styles.ts
CarlosCortizasCT: Restored here: 82a180e3727b5946ad094cb809a14302e9c106f2

PR #3356, opened 2023-12-19
refactor(e2e): use package name for caching custom views configs

kark:
#### Summary
This PR addresses the issue related to [caching processed custom view configs](https://github.com/commercetools/merchant-center-application-kit/pull/3350#discussion_r1429811666).

This approach involves using `entryPointUriPath` for caching Custom App configs (not changing anything here for backwards compatibility) and `entryPointUriPath`-`templateName` combination as the cache key for Custom Views. The template name is expected to be passed as a Cypress environment variable.

### DIFF
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -32,6 +32,7 @@ export default defineConfig({
           LOGIN_USER: process.env.CYPRESS_LOGIN_USER,
           LOGIN_PASSWORD: process.env.CYPRESS_LOGIN_PASSWORD,
           PROJECT_KEY: process.env.CYPRESS_PROJECT_KEY,
+          PACKAGE_NAME: process.env.PACKAGE_NAME,
         }),
       });
     },
--- a/cypress.config.ts
+++ b/cypress.config.ts
@@ -32,7 +32,7 @@ export default defineConfig({
           LOGIN_USER: process.env.CYPRESS_LOGIN_USER,
           LOGIN_PASSWORD: process.env.CYPRESS_LOGIN_PASSWORD,
           PROJECT_KEY: process.env.CYPRESS_PROJECT_KEY,
-          PACKAGE_NAME: process.env.PACKAGE_NAME,
+          PACKAGE_NAME: process.env.CYPRESS_PACKAGE_NAME,
         }),
       });
     },
--- a/packages/cypress/src/add-commands/index.ts
+++ b/packages/cypress/src/add-commands/index.ts
@@ -10,6 +10,8 @@ import {
 import { realHover } from './real-hover';
 
 export type CommandLoginOptions = TCommandLoginOptions;
+export type LoginToMerchantCenterForCustomViewCommandLoginOptions =
+  TLoginToMerchantCenterForCustomViewCommandLoginOptions;
 export type Matcher = TMatcher;
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
--- a/packages/cypress/src/add-commands/index.ts
+++ b/packages/cypress/src/add-commands/index.ts
@@ -5,6 +5,7 @@ import {
   loginByOidc,
   isLocalhost,
   type CommandLoginOptions as TCommandLoginOptions,
+  type LoginToMerchantCenterForCustomViewCommandLoginOptions as TLoginToMerchantCenterForCustomViewCommandLoginOptions,
 } from './login';
 import { realHover } from './real-hover';
 
@@ -31,12 +32,7 @@ Cypress.Commands.add(
 
 Cypress.Commands.add(
   'loginToMerchantCenterForCustomView',
-  (
-    commandOptions: Omit<
-      CommandLoginOptions,
-      'entryPointUriPath' | 'initialRoute'
-    >
-  ) => {
+  (commandOptions: TLoginToMerchantCenterForCustomViewCommandLoginOptions) => {
     Cypress.log({ name: 'loginToMerchantCenterForCustomView' });
 
     const projectKey = Cypress.env('PROJECT_KEY');
--- a/packages/cypress/src/add-commands/login.ts
+++ b/packages/cypress/src/add-commands/login.ts
@@ -61,6 +61,10 @@ export type CommandLoginOptions = {
    * This is only relevant for Cypress version >= `10.9.0`.
    */
   disableCacheAcrossSpecs?: boolean;
+  /**
+   * The package name as listed in the tested Custom View's package.json
+   */
+  packageName?: string;
 };
 // Alias for backwards compatibility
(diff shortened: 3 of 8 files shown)


packages/cypress/src/task/index.ts
```diff
@@ -69,9 +70,16 @@ const loadAllCustomEntityConfigs = async (
             isCustomViewConfig ? 'View' : 'Application'
           } config for ${packageInfo.packageJson.name}`
         );
+
+        const customEntityConfigCacheKey = isCustomViewConfig
+          ? `${processedConfig.env.entryPointUriPath}-${path.basename(
+              packageInfo.dir
+            )}`
+          : processedConfig.env.entryPointUriPath;
```
kark, line 78: During monorepo traversal by `loadAllCustomEntityConfigs`, any discovered custom view config is cached using a key derived from the directory name (equivalent to the template name)

packages/cypress/src/task/index.ts
```diff
@@ -118,13 +126,17 @@ const customViewConfig = async (
   });
 
   const customViewConfig =
-    allCustomEntityConfigs[CUSTOM_VIEW_HOST_ENTRY_POINT_URI_PATH];
+    allCustomEntityConfigs[
+      `${CUSTOM_VIEW_HOST_ENTRY_POINT_URI_PATH}-${options.templateName}`
+    ];
```
kark, line 131: Later, the cached custom view config is retrieved based on the provided `templateName`

emmenko:
Thanks @kark for looking into that and proposing a solution!

I'm wondering though if this fixes the issue completely, as the template name does not guarantee the uniqueness of the cache key.

For example, I could have 2 or more Custom Views configured in my monorepository, all of them using the same template. Therefore, I would have all cache keys with the same value (as the entry point is also a fixed value):

```
# Custom View 1
@@custom-view-host@@-starter-typescript

# Custom View 2
@@custom-view-host@@-starter-typescript

# Custom View 3
@@custom-view-host@@-starter-typescript
```

If we use something unique like the workspace path we can ensure to have unique cache keys and thus a correct match for loading the correct config.

Assuming I have a monorepo with the following workspaces (all of them use the same template):

```
packages
  custom-view-1
    package.json
  custom-view-2
    package.json
  custom-view-3
    package.json
```

We can then leverage the workspace path as a discriminator:

```
./packages/custom-view-1
./packages/custom-view-2
./packages/custom-view-3
```

This would be an option to pass to the Cypress setup / commands for "identifying the correct Custom View".

Happy to have a call to clarify this 🤗

kark:
> For example, I could have 2 or more Custom Views configured in my monorepository, all of them using the same template. Therefore, I would have all cache keys with the same value (as the entry point is also a fixed value):

Absolutely, I appreciate the input 💯 
I'm considering the use of the package name as a discriminator, ensuring uniqueness within the monorepo (as in https://github.com/commercetools/merchant-center-application-kit/pull/3356/commits/88a5a06b35f1cb15b0eba0469f5113b2f3584179). Would that work for you?

> This would be an option to pass to the Cypress setup / commands for "identifying the correct Custom View".

I wonder if I fully grasp the recommendation. If it involves a more intricate refactor to enhance the user-friendliness of the API, I'd greatly appreciate some guidance and I'm more than willing to schedule a meeting 😊

cypress.config.ts
```diff
@@ -32,6 +32,7 @@ export default defineConfig({
           LOGIN_USER: process.env.CYPRESS_LOGIN_USER,
           LOGIN_PASSWORD: process.env.CYPRESS_LOGIN_PASSWORD,
           PROJECT_KEY: process.env.CYPRESS_PROJECT_KEY,
+          PACKAGE_NAME: process.env.CYPRESS_PACKAGE_NAME,
```
kark, line 35: As suggested by Carlos, I moved reading this env variable from the command:
```
pnpm start-server-and-test 'pnpm template-${{ inputs.application-type }}-${{ inputs.template-name }}:start:prod:local' http-get://127.0.0.1:3001 'pnpm ${{ inputs.e2e-tests-command }} --env PACKAGE_NAME=${{ inputs.package-name }}'
```
to our Cypress config

CarlosCortizasCT (approved):
Thanks a lot! 🙇

packages/cypress/src/add-commands/login.ts
```diff
@@ -61,6 +61,10 @@ export type CommandLoginOptions = {
    * This is only relevant for Cypress version >= `10.9.0`.
    */
   disableCacheAcrossSpecs?: boolean;
+  /**
+   * The package name as listed in the tested Custom View's package.json
```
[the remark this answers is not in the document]
emmenko: Also, given that this is only specific to Custom Views, we should only define this option for the `loginToMerchantCenterForCustomView` command

packages/cypress/src/add-commands/login.ts
```diff
@@ -166,11 +179,15 @@ function loginByForm(commandOptions: CommandLoginOptions) {
 const isCustomView = (commandOptions: CommandLoginOptions) =>
   commandOptions.entryPointUriPath === CUSTOM_VIEW_HOST_ENTRY_POINT_URI_PATH;
 
-function loginByOidc(commandOptions: CommandLoginOptions) {
+function loginByOidc(
+  commandOptions: CommandLoginOptions &
+    LoginToMerchantCenterForCustomViewCommandLoginOptions
+) {
```
kark, line 185: @emmenko Thank you for the feedback. All changes were applied in 153c06cc138d0f4ef9db5a9cf9adca5e3dba7237.
I wonder if this intersection type here works for you, as I didn't want to go for discriminated unions

packages/cypress/src/add-commands/index.ts
```diff
@@ -31,12 +32,7 @@ Cypress.Commands.add(
 
 Cypress.Commands.add(
   'loginToMerchantCenterForCustomView',
-  (
-    commandOptions: Omit<
-      CommandLoginOptions,
-      'entryPointUriPath' | 'initialRoute'
-    >
-  ) => {
+  (commandOptions: TLoginToMerchantCenterForCustomViewCommandLoginOptions) => {
```
emmenko, line 35: Thanks!

PS: you also need to update the type declarations here: https://github.com/commercetools/merchant-center-application-kit/blob/d437612a894be32eaa6c124c8e0d1b126148e049/packages/cypress/add-commands/index.d.ts#L22-L28

emmenko (commented):
Perfect, thanks! 🙌

packages/cypress/src/add-commands/index.ts
kark: Ah, I overlooked that. Thank you!

emmenko (approved):
Awesome, thanks!

PR #3362, opened 2023-12-22
FCT 615: suppress ResizeObserver error

ByronDWall:
When a page uses ResizeObserver API, it may occasionally throw "ResizeObserver loop completed with undelivered notifications" (Firefox) or "ResizeObserver loop limit exceeded" (Chrome). The error has weird semantics, it never gets printed to browser console, only gets caught by window.onerror listener (used by the error overlay)

This error is a known false alarm: https://github.com/w3c/csswg-drafts/issues/5488

This PR is to ignore the `ResizeObserver` loop error in sentry.  For more context see [this github issue](https://github.com/vercel/next.js/discussions/51551)

### DIFF
--- a/.changeset/rich-grapes-cheat.md
+++ b/.changeset/rich-grapes-cheat.md
@@ -1,0 +1,6 @@
+---
+'@commercetools-frontend/application-shell': patch
+'@commercetools-frontend/sentry': patch
+---
+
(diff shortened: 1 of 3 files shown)


tdeekens (approved):

tdeekens:
Don't forget a changeset :)

emmenko (approved):
Thanks a lot! 🙌

PR #3365, opened 2023-12-27
fix: local login side effects when running in strict mode

### LINKED ISSUE #3361: Local development OIDC callback yields invalid session when React strict mode is used
**Describe the issue**

When we enable React strict mode (via the `enableReactStrictMode` prop of `<ApplicationShell>`), login in local development doesn't work as expected and the "Invalid client session" screen is shown, even though the login was successful and the session is valid.

image

The issue seems to relate to the fact that the callback cannot be verified due to a failed validation of the nonce / session state.

Upon some investigation, it appears that this is related to React strict mode re-rendering (in development). The callback component does remove the nonce / session from session storage upon first render. When React re-renders, the state is gone and thus the validation fails.

According to [React docs](https://react.dev/learn/synchronizing-with-effects#how-to-handle-the-effect-firing-twice-in-development), these kind of behaviors should be solved with a proper effect implementation where a potential clean up is safely handled for subsequent renders. 

**To Reproduce**

Enable `enableReactStrictMode` in the `<ApplicationShell>` and attempt to log in.

**Expected behavior**

Login in local development should work even in strict mode.

**Additional context**

Similar issue in SO: https://stackoverflow.com/questions/75808338/react-18-strict-mode-impure-function-component-that-clears-local-storage

emmenko:
Fixes #3361

The actual fix is to move the logic with side effects (e.g. redirects, interacting with storage) within a React `useEffect` (_who would have thought of that_ 🤪). This solves the issue with the component re-rendering due to strict mode as the effect is run once.

Additionally, I found some minor things to adjust to some dev-related pages.

> [!NOTE]
> Also a useful note from the legacy docs:
> Once we upgrade to React 18, we can expect multiple (duplicate) logs in the devtools due to strict mode re-rendering.

image

### DIFF
--- a/.changeset/happy-trainers-reflect.md
+++ b/.changeset/happy-trainers-reflect.md
@@ -1,0 +1,8 @@
+---
+'@commercetools-applications/merchant-center-custom-view-template-starter-typescript': minor
+'@commercetools-applications/merchant-center-template-starter-typescript': minor
+'@commercetools-applications/merchant-center-custom-view-template-starter': minor
+'@commercetools-applications/merchant-center-template-starter': minor
+---
+
+Enable React strict mode by default
--- a/.changeset/large-rats-begin.md
+++ b/.changeset/large-rats-begin.md
@@ -1,0 +1,6 @@
+---
+'@commercetools-frontend/application-shell-connectors': patch
+'@commercetools-frontend/application-shell': patch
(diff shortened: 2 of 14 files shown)


ddouglasz (approved):
Thank you 🙏🏾

PR #3371, opened 2024-01-15
chore: upgrade to latest docs kit and enable AI assistant

gabriele-ct:

### DIFF
--- a/website-custom-views/gatsby-config.mjs
+++ b/website-custom-views/gatsby-config.mjs
@@ -1,5 +1,7 @@
 // eslint-disable-next-line import/extensions
 import colorPresets from '@commercetools-docs/gatsby-theme-docs/color-presets/index.mjs';
+const { isProd } = process.env;
+
 
 /**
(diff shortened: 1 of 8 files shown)


FFawzy (approved):
lgtm 👍

website-custom-views/gatsby-config.mjs
```diff
@@ -1,5 +1,7 @@
 // eslint-disable-next-line import/extensions
 import colorPresets from '@commercetools-docs/gatsby-theme-docs/color-presets/index.mjs';
+const { isProd } = process.env;
```
emmenko, line 3: We don't have this defined here in this repo.

emmenko (changes requested):
Thanks!

website-custom-views/gatsby-config.mjs
gabriele-ct: That's fine, I'll remove it and we can stick to the prod url
]

emmenko (approved):
Thanks 🤗

PR #3372, opened 2024-01-15
Add `isProductionProject` to project query

chloe0592:
We needed the new field `isProductionProject` so this PR adds it to the project query and generates new graphql types.


emmenko (approved):

tdeekens (approved):

CarlosCortizasCT:
I'm sorry but I've been trying to figure out how to fix the test that's failing but I don't find a solution. Also, I don't understand how the changes in this PR can cause that error.

@emmenko could you perhaps help us here? 🙏

emmenko:
Let me see...

PR #3378, opened 2024-01-17
[ALTERNATIVE] Integration of Page Layout components within Custom View panels

CarlosCortizasCT:
#### Summary

[ALTERNATIVE] Integration of Page Layout components within Custom View panels

#### Description

In this PR I've implemented @emmenko's idea about not using custom CSS rules to avoid double spacing around contents but to use app-kit's design tokens overriding.


packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
```diff
@@ -132,6 +139,13 @@ function CustomViewShell(props: TCustomViewShellProps) {
     notificationsGlobalRef,
     notificationsPageRef,
   });
+  const themeOverrides = useMemo(
+    () => ({
+      ...themesOverrides.default,
+      ...customViewsThemesOverrides.default,
+    }),
+    []
+  );
```
emmenko, line 148: Nit: we can also define this as a static variable outside the component, no need to use memo.

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
```diff
@@ -205,8 +219,14 @@ function CustomViewShell(props: TCustomViewShellProps) {
       : hostContext.hostUrl;
 
   return (
-    <div data-extension-type={CUSTOM_EXTENSION_TYPES.CUSTOM_VIEW}>
+    <div
+      css={css`
+        height: 100%;
+      `}
+      data-extension-type={CUSTOM_EXTENSION_TYPES.CUSTOM_VIEW}
+    >
```
emmenko, line 227: I guess this can also be removed now...?

emmenko (commented):
Oh wow, thanks a lot for giving it a try!

The benefit I see here is that the application components stay agnostic and we can override things from the consumer side (the custom view shell in this case).

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
CarlosCortizasCT: Yes. Sorry.

Updated here: 1b5f4702b075c058d749fef1287db758ab9d6088

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
CarlosCortizasCT: Totally.

Removed here: 1b5f4702b075c058d749fef1287db758ab9d6088

CarlosCortizasCT:
> Oh wow, thanks a lot for giving it a try!
> 
> The benefit I see here is that the application components stay agnostic and we can override things from the consumer side (the custom view shell in this case).

Yes, I totally agree with you.
This solution is also cleaner as the surface we touch is smaller the risk is lower and the updates would be also easier.

However, I think it's important for us to understand both the pros and the cons so we are aware of the compromises we make when choosing one option over the other.

As I mentioned in the original PR, we also have [a drawback](https://github.com/commercetools/merchant-center-application-kit/pull/3353#discussion_r1455700779) to be aware of. I'm not saying we should discard this alternative (I also like it better) but I would like for us to make sure we understand the compromises we take.

emmenko:
Yep, totally agree. Thanks for make sure we are aware of the tradeoffs!

PR #3383, opened 2024-01-22
fix(dialogs): make `scale` size prop take effect only on the horizontal constraints

kark:
#### Summary

|  Before | After  |
|---|---|
|  Screenshot 2024-01-23 at 08 45 45 |  Screenshot 2024-01-23 at 08 48 14 |


website-custom-views/src/content/api-reference/commercetools-frontend-application-components.mdx
```diff
@@ -1130,7 +1130,7 @@ const ChannelsInfo = () => {
...
 | `children`          | `node`     |    ✅    | -                 | -       | Content rendered within the dialog. If the content is long in height (depending on the screen size) a scrollbar will appear.                                                                                                                                                                                    |
-| `size`              | `string`   |    -     | `m`, `l`, `scale` | `l`     | Horizontal width of the dialog card. If `scale` is used, the dialog size expands to the viewport size, with some margin.                                                                                                                                                                                                                                                                            |
```
emmenko, line 1133: I think the comment is still valid regarding `scale`, we don't need to change that.

emmenko (commented):
Thanks for the fix! 🙌

website-custom-views/src/content/api-reference/commercetools-frontend-application-components.mdx
kark: Yes, I believe so, thanks

CarlosCortizasCT (approved):
Thanks a lot! 🙇

emmenko (approved):
💯

Rhotimee (approved):
💯

PR #3391, opened 2024-01-30
chore: remove Custom Application docs (keep components-playground)

emmenko:
The docs have been moved to our main docs repository, so we can remove the content from here.

We still need to keep the components playground website though.


CarlosCortizasCT (approved):
Thanks a lot 🙇

kark (approved):
Thanks!

PR #3395, opened 2024-02-05
refactor(main nav): use emotion for styling - follow up

kark:
#### Summary

[Deployment link](https://mc-15822.mc-preview.europe-west1.gcp.escemo.com/)

This PR resolves tech debt by migrating navbar styles from globally applied classname-based styles within `<Global>` to encapsulated component-level styles. It retains specific styling for '.body__menu-open' class within `<Global>`.

### DIFF
--- a/packages/application-shell/src/components/navbar/main-navbar.styles.ts
+++ b/packages/application-shell/src/components/navbar/main-navbar.styles.ts
@@ -1,6 +1,7 @@
 import { css, keyframes } from '@emotion/react';
 import styled from '@emotion/styled';
-import { designTokens } from '@commercetools-uikit/design-system';
+import { designTokens as appKitDesignTokens } from '@commercetools-frontend/application-components';
+import { designTokens as uiKitDesignTokens } from '@commercetools-uikit/design-system';
 import { NAVBAR } from '../../constants';
 import { MenuListItem, LeftNavigation } from './menu-items.styles';
 import { ItemIconText, Title } from './shared.styles';
@@ -26,7 +27,7 @@ const FixedMenu = styled.div`
 `;
 
 const ItemContent = styled.div`
-  color: var(--color-for-navbar-link);
+  color: ${appKitDesignTokens.colorForNavbarLink};
   width: ${NAVBAR.itemSize};
   position: relative;
   display: block;
@@ -36,12 +37,13 @@ const ScrollableMenu = styled.div`
   flex: 1 1 0;
   overflow-x: hidden;
   overflow-y: hidden;
-  padding: var(--spacing-30) var(--spacing-30) ${NAVBAR.itemSize};
+  padding: ${uiKitDesignTokens.spacing30} ${uiKitDesignTokens.spacing30}
+    ${NAVBAR.itemSize};
   width: ${NAVBAR.widthLeftNavigation};
   box-sizing: border-box;
   /* For Firefox */
   scrollbar-width: thin;
-  scrollbar-color: var(--color-primary-40) transparent;
+  scrollbar-color: ${uiKitDesignTokens.colorPrimary40} transparent;
 
   :hover {
     overflow-y: scroll;
@@ -57,8 +59,8 @@ const ScrollableMenu = styled.div`
   }
 
   ::-webkit-scrollbar-thumb {
-    background-color: var(--color-primary-40);
-    border-radius: var(--border-radius-8);
+    background-color: ${uiKitDesignTokens.colorPrimary40};
+    border-radius: ${uiKitDesignTokens.borderRadius8};
   }
 `;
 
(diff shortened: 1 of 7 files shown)


packages/application-shell/src/components/navbar/menu-items.tsx
```diff
@@ -270,22 +252,23 @@ type MenuItemProps = {
...
       onMouseLeave={props.onMouseLeave as MouseEventHandler<HTMLElement>}
       onFocus={props.onMouseEnter as FocusEventHandler<HTMLElement>}
       onBlur={props.onMouseLeave as FocusEventHandler<HTMLElement>}
       data-menuitem={props.identifier}
-      className={classnames('list-item', {
-        item__active: props.isActive,
-        item_menu__active: props.isMainMenuRouteActive ?? false,
-        'item_menu-collapsed': !props.isMenuOpen,
+      className={classnames({
+        active: props.isActive,
```
kark, line 264: classnames are retained here and in the  [`<MenuList>`](https://github.com/commercetools/merchant-center-application-kit/blob/a6ea9671704ccb3e40690b8df2c7ef7046985267/packages/application-shell/src/components/navbar/menu-items.tsx#L203-L220) rendered as part of  `<MenuGroup>` since some styling changes are applied from the outside of those components e.g. [here](https://github.com/commercetools/merchant-center-application-kit/blob/a6ea9671704ccb3e40690b8df2c7ef7046985267/packages/application-shell/src/components/navbar/main-navbar.styles.ts#L88) and [here](https://github.com/commercetools/merchant-center-application-kit/blob/a6ea9671704ccb3e40690b8df2c7ef7046985267/packages/application-shell/src/components/navbar/menu-items.styles.ts#L285-L317)

packages/application-shell/src/components/navbar/menu-items.styles.ts
```diff
@@ -110,14 +112,227 @@ const NavlinkClickableContent = styled.div`
...
+          ${getContainerPositionBasedOnMenuItemPosition(
+            props.isSubmenuAboveMenuItem,
+            props.isSublistActiveWhileIsMenuExpanded,
+            props.isSublistActiveWhileIsMenuCollapsed
+          )}
+        }
+      `,
+  ]}
+
+  .highlighted,
```
kark, line 183: used [here](https://github.com/commercetools/merchant-center-application-kit/blob/a6ea9671704ccb3e40690b8df2c7ef7046985267/packages/application-shell/src/components/navbar/menu-items.tsx#L306)

FilPob (approved):

packages/application-shell/src/components/navbar/menu-items.styles.ts
CarlosCortizasCT: It took me a while to understand this one so I leave a comment with some context for next reviewers.

The `MenuItems` component internally uses `NavLink` one to render the final menu links. The latter is a component coming from `react-router` and it allows for a class name to be provided in order to style the component.
Since the `MenuItems` component is its _parent_, it makes sense to define the class in this file.

I'm wondering if we could use a more specific selector to better indicate the relationship between both components.
Maybe something like this:
```
& .highlighted,
& .highlighted ${Title} {
```

packages/application-shell/src/components/navbar/menu-items.styles.ts
```diff
@@ -110,14 +112,227 @@ const NavlinkClickableContent = styled.div`
...
+  margin: 0;
+  padding: 0;
+  display: flex;
+  flex-direction: column;
+  min-height: 0;
+  flex: 1 1 0;
+`;
+
+const sublistStyles = css`
+  padding: var(--spacing-30);
```
CarlosCortizasCT, line 125: Since we're moving from the CSS modules world to the CSS-in-JS one, I think it would make sense to use the design tokens object instead of CSS vars: 
1. we would keep consistency with the rest of `app-kit` components
2. it would be easier to maintain as they are easily navigable in the editor and even more discoverable

```
padding: ${designTokens.spacing30}
```

kark: Great idea 🙌 Updated in f8e8dbc7693a881ea74c2ee71da5c721da11d49c

packages/application-shell/src/components/navbar/main-navbar.styles.ts
```diff
@@ -143,40 +145,41 @@ const TooltipContainer = styled.div<{
...
-  padding: var(--spacing-25);
+  padding: ${uiKitDesignTokens.spacing25};
   align-items: center;
   justify-content: center;
 `;
 
 const SupportMenu = styled.div`
-  padding: var(--spacing-10) var(--spacing-30) var(--spacing-20)
-    var(--spacing-30);
-  height: calc(var(--item-size) + var(--spacing-20));
```
kark, line 375: The differences in the visual regression tests come from this line. 
Before it was 
Screenshot 2024-02-07 at 15 15 02
(this css var doesn't exist)

Now it is
Screenshot 2024-02-07 at 15 13 28

CarlosCortizasCT (approved):
Thanks a lot for this awesome work, Kacper 👏 💯 💪

chloe0592 (approved):
Thanks Kacper! Awesome work 👏 ❤️

PR #3397, opened 2024-02-08
refactor: deploy components playground website to Vercel using github integration

emmenko:
Follow up of #3391 

Now that we have moved the docs to the docs repo, we can simplify the deployment of the components playground application as a standalone website.
For that, we can use the native Vercel / GitHub integration.

I create a new Vercel project to not affect the current (legacy Custom Applications) website. Once we switch the domain to render the components playground, we can remove the legacy project.


CarlosCortizasCT:
Thanks for working on this, Nicola.

I'd like to ask for some clarification as I don't understand this comment:
> I create a new Vercel project to not affect the current (legacy Custom Applications) website. Once we switch the domain to render the components playground, we can remove the legacy project.

I don't get where a new domain will be created and how it will be configured.
How we will be able to access the components playground.


Also, I was yesterday discussing with Klaudija how she can have designers to validate the changes she's working on related to the `AppBar` and I thought it would be easier if we would be also deploying the playground application so it would be accessible from the PRs and designers could preview some changes.
Do you think it would be a good idea to also deploy that one?

CarlosCortizasCT (approved):
Thanks! 👍

kark (approved):

PR #3409, opened 2024-02-19
Add integration with `jest-preview`

CarlosCortizasCT:
#### Summary

Add integration with `jest-preview`

#### Description

In this PR we introduce support for using the [jest-preview](https://www.jest-preview.com/) which can help us when writing and fixing integrations tests.

This tool allows to render in the browser the HTML of an specific test so it's easier to understand the UI state when validating requirements.


setup-tests.js
```diff
@@ -0,0 +1,5 @@
+import { jestPreviewConfigure } from 'jest-preview';
+
+if (!process.env.CI) {
```
CarlosCortizasCT, line 3: The idea is that we don't need to run the `autoPreview` mode in CI but just locally.

tdeekens (approved):

kark (approved):
Thank you, that's a great idea ❤️

PR #3412, opened 2024-02-20
Fix navigation sidebar menu icons alignment

CarlosCortizasCT:
#### Summary

Fix navigation sidebar menu icons alignment.

#### Description

We noticed a vertical misalignment in the main navigation sidebar between the icons and their corresponding label text:
image

This is how it looks after applying the fix:
image


kark (approved):
Thanks!

PR #3413, opened 2024-02-21
Allow React elements in Dialog components header titles

### LINKED ISSUE #1864: Allow ReactNode as a title prop on info-dialog component

Info-dialog component only accepts string type as a title. Throws type errors when we pass in a ReactNode as seen below which is a combination of some string and a stamp from ui-kit.

Screenshot 2020-11-11 at 10 00 11

We would like to change type of title from string to a ReactNode.

This would consequently affect the following components and respective props:

> DialogHeader :: title
> DialogContainer :: title
> Modal :: contentLabel
> ModalPortal :: contentLabel

**Additional context**

Here is how the info dialog looks like when we use a ReactNode for the title:

Screenshot 2020-11-11 at 09 56 52

CarlosCortizasCT:
#### Summary

Allow React elements in Dialog components header titles.

closes #1864

#### Description

In this PR we update Dialog components so consumers can provide React elements for its header's title.

Up until now, only `string` values where accepted because the title was not only used for the label rendered in the dialog's header but also for the dialog's accessible title so providing a React element would break the dialogs accessibility.

We're updating the allowed type for dialog's title so, instead of just `string` values, it will also allow for `ReactNode` values.
Also, we're introducing a new `string` prop named **ariaLabel**. This property is optional, but consumers are expected to use it only when they provide a `ReactNode` value in the `title` prop so the dialog components can have an accessible title in that case.

I also took the change to refactor the dialog components tests a bit, since all of them were quite similar.
I created some shared test utils that abstract most of the common logic from the actual components tests themselves.


packages/application-components/src/components/dialogs/form-dialog/form-dialog.spec.tsx
```diff
@@ -1,104 +1,95 @@
...
-      />
-      {props.children({ isOpen, setIsOpen })}
-    </div>
-  );
+const createFormDialogValidator = () => {
+  const onCancel = jest.fn();
+  const onConfirm = jest.fn();
+
+  return createDialogValidator({
+    component: FormDialog,
```
CarlosCortizasCT, line 14: I would appreciate some help with this TS error as I don't seem to be able to find the right configuration.

packages/application-components/src/components/dialogs/form-dialog/form-dialog.spec.tsx
```diff
@@ -1,104 +1,95 @@
...
+      extraChecks: () => {
+        screen.getByText('button icon');
+      },
+    }));
+});
 
-    fireEvent.click(screen.getByLabelText(/Open Form Dialog/i));
-    await screen.findByText(/Lorem ipsus/i);
-    expect(screen.getByText(/Hello/i)).toBeInTheDocument();
+describe('with custom title', () => {
```
CarlosCortizasCT, line 58: These tests were added to the three dialog components so we verify the new changes work as expected.

packages/application-components/src/components/dialogs/internals/dialog-container.tsx
```diff
@@ -63,70 +65,80 @@ const GridArea = styled.div<GridAreaProps>`
...
-          name="main"
-          css={css`
-            display: flex;
-            flex-direction: column;
-            align-items: center;
-            justify-content: center;
-            height: 100%;
-            overflow: hidden;
-          `}
+const DialogContainer = (props: Props) => {
```
CarlosCortizasCT, line 68: Changes here are mainly because of the indentation.
The relevant part is adding the warning for the accessible title.

emmenko (commented):
🙌

packages/application-components/src/components/dialogs/internals/dialog-container.tsx
```diff
@@ -63,70 +65,80 @@ const GridArea = styled.div<GridAreaProps>`
...
+  useWarning(
+    typeof props.title === 'string' ||
+      (typeof props.title !== 'string' && Boolean(props.ariaTitle)),
+    'app-kit/DialogHeader: "ariaTitle" prop is required when you provide a custom "title" element.'
+  );
```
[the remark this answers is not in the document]
CarlosCortizasCT: Updated here: fa5dbc94efd1178f36090400a432581235edc04e

packages/application-components/src/components/dialogs/internals/dialog-container.tsx
```diff
@@ -43,7 +44,8 @@ type Props = {
   onClose?: (event: SyntheticEvent) => void;
   size: 'm' | 'l' | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 16 | 'scale';
   zIndex?: number;
-  title: string;
+  title: ReactNode;
+  ariaTitle?: string;
```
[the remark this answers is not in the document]
CarlosCortizasCT: I changed the prop name to be `aria-label`.

Updated here: fa5dbc94efd1178f36090400a432581235edc04e

.changeset/chilled-eyes-greet.md
```diff
@@ -0,0 +1,33 @@
+---
+'@commercetools-frontend/application-components': minor
+---
+
+We've updated the `title` property type of dialog components (`InfoDialog`, `FormDialog` and `ConfirmationDialog`) so, instead of only supporting `string` values, it will support `ReactNode` as well.
+This change is intended to allow for consumers to provide more complex content, such as custom components, to the `title` property of these components.
```
emmenko, line 6:
```suggestion
We've updated the `title` property type of dialog components (`InfoDialog`, `FormDialog` and `ConfirmationDialog`) to be a `ReactNode` instead of `string` values.
This change is intended to enable consumers to define the title of the dialogs as a React component, thus allowing to support more use cases.
```

.changeset/chilled-eyes-greet.md
```diff
@@ -0,0 +1,33 @@
...
+
+Since the `title` property was also used as the accessible title of the dialogs and we can't extract the value from a `ReactNode` title, we're introducing a new `aria-label` property to these components.
+Consumers using a non-string value for the `title` property should also provide a string value for the `aria-label` property, which will be used as the accessible title of the dialog.
```
emmenko, line 9:
```suggestion
Furthermore, given that the `title` property was also used as the accessible title of the dialogs (`aria-label`) and a `ReactNode` cannot be used for that, we're introducing a new `aria-label` property to address the accessibility requirement.
The `aria-label` property is therefore required when the `title` property is not of type `string`.
```

packages/application-components/src/components/dialogs/form-dialog/form-dialog.spec.tsx
emmenko: Try typing the prop as `ElementType`

packages/application-components/src/components/dialogs/internals/dialog-test-utils.tsx
```diff
@@ -0,0 +1,115 @@
...
+      </Text.Headline>
+      <SecondaryButton label="Click me" tone="info" />
+    </Spacings.Inset>
+  );
+};
+
+export type TCreateDialogValidatorParams = {
+  component: ComponentType<
+    TConfirmationDialogProps | TFormDialogProps | TInfoDialogProps
+  >;
```
emmenko, line 47: Lets's keep it simple

```suggestion
  component: ElementType;
```

emmenko (commented):
Thanks, looks good to me! 🙌

packages/application-components/src/components/dialogs/internals/dialog-test-utils.tsx
CarlosCortizasCT: Thanks a lot for the suggestion 🙇 

Updated here: c31bc90c72b996181b5261a2fd4a9aee3d860509

packages/application-components/src/components/dialogs/form-dialog/form-dialog.spec.tsx
CarlosCortizasCT: Updated here: https://github.com/commercetools/merchant-center-application-kit/commit/c31bc90c72b996181b5261a2fd4a9aee3d860509

kark (approved):
🙌

packages/application-components/src/components/dialogs/confirmation-dialog/confirmation-dialog.tsx
```diff
@@ -62,5 +67,8 @@ ConfirmationDialog.defaultProps = defaultProps;
 // This is a convenience proxy export to expose pre-defined Intl messages defined in the `@commercetools-frontend/i18n` package.
 // The Intl messages can be used for button labels.
 ConfirmationDialog.Intl = sharedMessages;
+// Allow consumers who want to use a custom title to reuse the same title styles
+// as the default dialog title.
+ConfirmationDialog.TextTitle = TextTitle;
```
emmenko, line 72: Should we mention this in the changelog?

emmenko (approved):
All good, thanks a lot! 💯

PR #3421, opened 2024-02-27
Support feature flags in Custom Views

CarlosCortizasCT:
#### Summary

Support feature flags in Custom Views.

#### Description

When we first implemented Custom Views feature, we decided not to include feature flags support in the MVP as we thought that was something we wouldn't need, but now we have a use case where we need that support so we're adding it.

The idea to support feature flags in Custom Views is that, since they will always be rendered within a host application which already has the feature flags fetched by the time a Custom View in activated, we want the host to provide the feature flags to the Custom View in the bootstrap inter-application event.
This way, the Custom View can create a `flop-flip` instance with an in-memory adapter and fill it with the flags it will receive from the host application.


packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -1,5 +1,6 @@
 import { useCallback, useEffect, useRef } from 'react';
 import styled from '@emotion/styled';
+import { useAllFeatureToggles } from '@flopflip/react-broadcast';
```
CarlosCortizasCT, line 3: This is a new hook we've added to `flopflip`.
That's why we're also updating `flopflip` dependencies.

packages/application-shell/src/components/custom-view-flop-flip-provider/custom-view-flop-flip-provider.tsx
```diff
@@ -0,0 +1,37 @@
...
+  projectKey?: string;
+  user?: TFetchLoggedInUserQuery['user'];
+  flags?: TFlags;
+  children: ReactNode;
+};
+
+function CustomViewFlopFlipProvider(props: TCustomViewFlopFlipProviderProps) {
+  const adapterArgs = useMemo(
+    () => ({
+      user: {
```
CarlosCortizasCT, line 17: `user` is required in the adapter args.
We don't actually need anything here because they're use to evaluate the flags, but that's already done in the host application. However, I just wanted to add some basic information we already have available.

packages/application-shell/src/components/custom-view-shell/custom-view-shell.tsx
```diff
@@ -116,6 +118,20 @@ function StrictModeEnablement(props: TStrictModeEnablementProps) {
   }
 }
 
+function CustomViewThemeProvider() {
+  const theme = useFeatureToggle('mcRecolouring') ? 'recolouring' : 'default';
+
+  const customViewThemeOverrides = {
+    ...themesOverrides[theme],
+    // @ts-ignore
```
CarlosCortizasCT, line 126: This is temporary until the [recolouring PR](https://github.com/commercetools/merchant-center-application-kit/pull/3414) is merged.

kark (approved):
Thank you for discovering the solution upstream 🎉 it works great

packages/application-shell/src/components/custom-view-flop-flip-provider/custom-view-flop-flip-provider.tsx
```diff
@@ -0,0 +1,37 @@
...
+        kind: 'user',
+        key: props.user?.id,
```
tdeekens, line 19: I think the kind isn't needed. That's mainly LD adapter context related. If we want to be minimal we can pass the user and a key as a stable identifier. That to my memory is the basic requirement to match the TBaseUser in flopflip.

packages/application-shell/src/components/custom-view-flop-flip-provider/custom-view-flop-flip-provider.tsx
```diff
@@ -0,0 +1,37 @@
...
+        project: props.projectKey ?? '',
+      },
+    }),
+    [props.user?.id, props.projectKey]
+  );
+
+  return (
+    <ConfigureFlopFlip
```
tdeekens, line 27: Nice 💯!

packages/application-shell/src/components/custom-view-shell-authenticated/custom-view-shell-authenticated.tsx
```diff
@@ -95,9 +98,15 @@ function CustomViewShellAuthenticated(
                               projectDataLocale={props.dataLocale}
                               environment={props.environment}
                             >
-                              <CustomViewWithPermissionCheck>
-                                {props.children}
-                              </CustomViewWithPermissionCheck>
+                              <CustomViewFlopFlipProvider
```
tdeekens, line 101: I don't mind but the SetupFlipFlip is also camelcase on the Flip. Just to consistent on the casing here and the filename.

tdeekens (commented):
Nice work. Thanks for contributing to flopflip!

packages/application-shell/src/components/custom-view-flop-flip-provider/custom-view-flop-flip-provider.tsx
CarlosCortizasCT: Updated here: 1db66771d812b230663c074732580405a3cc8b58

packages/application-shell/src/components/custom-view-shell-authenticated/custom-view-shell-authenticated.tsx
CarlosCortizasCT: Sorry but I don't understand.

The current name of the component is already camel casing `FlopFlip`.
Could you maybe give an example of how the name should looks like?
Thanks.

tdeekens: Sorry for not making the point clear. I double checked myself which I couldn't yesterday.

I just wanted to make sure that `setup-flop-flip-provider.tsx` and this file follow the same `flop-flip` casing. The library itself isn't cased like it. Hence my "stumble" on the review.

tdeekens (approved):

PR #3425, opened 2024-03-05
refactor(i18n): change Transifex file format to Structured JSON

emmenko:
We want to use this to test out what happens in Transifex when we change an existing resource file format.

We will use the feedback to eventually plan the migration work for our main frontend repository.


CarlosCortizasCT:
Thanks!

PR #3426, opened 2024-03-05
fix(app-shell): to only events for flag variations

tdeekens:
#### Summary

This attempts to reduce the flag evaluation to occur for flags we don't request. This is the result of a LD support ticket.


emmenko (approved):
Thanks

tdeekens:
Thanks for the review. Will debug CI tomorrow.

CarlosCortizasCT (approved):
Thanks!

Just to better understand what this change does: is it about evaluation in the server side so the request time would potentially be shorter?

tdeekens:
> Thanks!
> 
> Just to better understand what this change does: is it about evaluation in the server side so the request time would potentially be shorter?

Not how I understood LaunchDarkly. There are some SDKs which have the legacy behaviour - to be non-breaking - that they send events to LaunchDarkly also for flags you don't requested. 

The docs say:

> Whether analytics events should be sent only when you call variation (true), or also when you call allFlags (false).
> By default, this is false (events will be sent in both cases).

They recommend to only send analytic events for enabled flags as otherwise you run into cases in which false or non used flags send analytics data and then the flags are considered to be evaluated. 

To be honest it's a recommendation they made. I will try out a local build to see what we get over the wire prior to merging this.

CarlosCortizasCT:
> > Thanks!
> > Just to better understand what this change does: is it about evaluation in the server side so the request time would potentially be shorter?
> 
> Not how I understood LaunchDarkly. There are some SDKs which have the legacy behaviour - to be non-breaking - that they send events to LaunchDarkly also for flags you don't requested.
> 
> The docs say:
> 
> > Whether analytics events should be sent only when you call variation (true), or also when you call allFlags (false).
> > By default, this is false (events will be sent in both cases).
> 
> They recommend to only send analytic events for enabled flags as otherwise you run into cases in which false or non used flags send analytics data and then the flags are considered to be evaluated.
> 
> To be honest it's a recommendation they made. I will try out a local build to see what we get over the wire prior to merging this.

Thanks for the explanation. I think I know understand better what we're trying to do. 👍

tdeekens:
It doesn't seem to solve the issue but it should reduce a lot of networking we don't need/want anyway at this time.

PR #3429, opened 2024-03-06
refactor(i18n): convert translation files to Structured JSON

emmenko:
Follow up of #3425 

As we can't simply change the file format, we tried then to reconfigure the github integration after manually removing the source file in transifex.

We now need to also configure the translation files to use the structured JSON format as otherwise transifex is not able to parse the data.
image


ddouglasz (approved):
Thank you

PR #3432, opened 2024-03-06
Shield 1129 support link tooltip

ddouglasz:
#### Summary

Show tooltip on support link when hovered
#### Description

When the navbar is collapsed, onhover, the support menu item should display a tooltip.

Screenshot:
image

### DIFF
--- a/.changeset/silver-buses-collect.md
+++ b/.changeset/silver-buses-collect.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': minor
+---
+
+When the navbar is collapsed, onhover, the support menu item should display a tooltip.
--- a/.changeset/silver-buses-collect.md
+++ b/.changeset/silver-buses-collect.md
@@ -2,4 +2,4 @@
 '@commercetools-frontend/application-shell': minor
 ---
 
-When the navbar is collapsed, onhover, the support menu item should display a tooltip.
+Updated the "Support" navigation menu item so it also has a tooltip when the menu is collapsed (same as the rest of the menu root items).
--- a/packages/application-shell/src/components/application-shell/application-shell.spec.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.spec.js
@@ -869,6 +869,8 @@ describe('when clicking on navbar menu toggle', () => {
     expect(
       navbarRendered.getByRole('tooltip', { hidden: true })
     ).toBeInTheDocument();
+    // Verify that only "SUpport" text from tooltip is available when the menu is collapsed
+    expect(navbarRendered.getAllByText('Support')).toHaveLength(1);
   });
 });
 
--- a/packages/application-shell/src/components/application-shell/application-shell.spec.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.spec.js
@@ -865,8 +865,10 @@ describe('when clicking on navbar menu toggle', () => {
       );
     });
 
-    // Verify that the "Support" link is not rendered when the menu is collapsed
-    expect(navbarRendered.queryByText('Support')).not.toBeVisible();
+    // Verify that the tooltip is available when the menu is collapsed
+    expect(
+      navbarRendered.getByRole('tooltip', { hidden: true })
+    ).toBeInTheDocument();
   });
 });
 
--- a/packages/application-shell/src/components/application-shell/application-shell.spec.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.spec.js
@@ -846,7 +846,9 @@ describe('when clicking on navbar menu toggle', () => {
     });
 
     // Check that the "Support" link is rendered when the menu is expanded
-    expect(navbarRendered.getByText('Support')).toBeInTheDocument();
+    expect(navbarRendered.queryByTestId('support-label').textContent).toBe(
+      'Support'
+    );
 
     // Collapse the menu and verify local storage changes
     fireEvent.click(button);
@@ -866,7 +868,7 @@ describe('when clicking on navbar menu toggle', () => {
     });
 
     // Verify that the "Support" link is not rendered when the menu is collapsed
-    expect(navbarRendered.queryByText('Support')).toBeNull();
+    expect(navbarRendered.queryByTestId('support-label')).toBeNull();
   });
 });
 
--- a/packages/application-shell/src/components/application-shell/application-shell.spec.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.spec.js
@@ -846,7 +846,9 @@ describe('when clicking on navbar menu toggle', () => {
     });
 
     // Check that the "Support" link is rendered when the menu is expanded
-    expect(navbarRendered.getAllByText('Support')).toHaveLength(1);
+    // Make sure the element we find is not the tooltip rendered when the menu is collapsed
+    const supportMenuItemLabel = navbarRendered.getByText('Support');
+    expect(supportMenuItemLabel.getAttribute('role')).not.toEqual('tooltip');
 
     // Collapse the menu and verify local storage changes
     fireEvent.click(button);
@@ -866,11 +868,9 @@ describe('when clicking on navbar menu toggle', () => {
     });
 
     // Verify that the tooltip is available when the menu is collapsed
-    expect(
-      navbarRendered.getByRole('tooltip', { hidden: true })
-    ).toBeInTheDocument();
-    // Verify that only "SUpport" text from tooltip is available when the menu is collapsed
-    expect(navbarRendered.getAllByText('Support')).toHaveLength(1);
+    const supportMenuItemTooltip = navbarRendered.getByText('Support');
+    expect(supportMenuItemTooltip.getAttribute('role')).toEqual('tooltip');
+    expect(supportMenuItemTooltip).not.toBeVisible();
   });
 });
 
--- a/packages/application-shell/src/components/application-shell/application-shell.spec.js
+++ b/packages/application-shell/src/components/application-shell/application-shell.spec.js
@@ -846,9 +846,7 @@ describe('when clicking on navbar menu toggle', () => {
     });
 
     // Check that the "Support" link is rendered when the menu is expanded
-    expect(navbarRendered.queryByTestId('support-label').textContent).toBe(
-      'Support'
-    );
+    expect(navbarRendered.getAllByText('Support')).toHaveLength(1);
 
     // Collapse the menu and verify local storage changes
     fireEvent.click(button);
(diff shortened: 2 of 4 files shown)


kark (commented):
Thanks Douglas 🙌
I've noticed 2 issues that we might want to address:
- the tooltip appears on hover but does not display when the support menu item receives focus, such as during keyboard navigation
- there's a fade-out effect present, which differs from the behaviour seen in all other menu items

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -866,7 +868,7 @@ describe('when clicking on navbar menu toggle', () => {
     });
 
     // Verify that the "Support" link is not rendered when the menu is collapsed
-    expect(navbarRendered.queryByText('Support')).toBeNull();
+    expect(navbarRendered.queryByTestId('support-label')).toBeNull();
```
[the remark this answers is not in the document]
ddouglasz: Updated here: https://github.com/commercetools/merchant-center-application-kit/pull/3432/commits/02f38c5d523f11b8d85fab44adf52dc0c4cc4561

Thank you

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -846,7 +846,9 @@ describe('when clicking on navbar menu toggle', () => {
     });
 
     // Check that the "Support" link is rendered when the menu is expanded
-    expect(navbarRendered.getByText('Support')).toBeInTheDocument();
+    expect(navbarRendered.queryByTestId('support-label').textContent).toBe(
```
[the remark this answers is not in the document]
ddouglasz: Updated here: https://github.com/commercetools/merchant-center-application-kit/pull/3432/commits/02f38c5d523f11b8d85fab44adf52dc0c4cc4561

Thank you

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -429,7 +437,7 @@ const NavBar = (props: TNavbarProps) => {
                     </Icon>
                   </IconWrapper>
                   {isMenuOpen ? (
-                    <Title>
+                    <Title data-testid="support-label">
```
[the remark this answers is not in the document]
ddouglasz: Updated here: https://github.com/commercetools/merchant-center-application-kit/pull/3432/commits/02f38c5d523f11b8d85fab44adf52dc0c4cc4561

Thank you

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -405,6 +406,13 @@ const NavBar = (props: TNavbarProps) => {
           {/* TODO: remove <Faded /> completely as part of the recolouring rollout cleanup process */}
           <Faded />
           <SupportMenu>
+            {!isMenuOpen && (
+              <SupportMenuTooltipContainer className="support">
```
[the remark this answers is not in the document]
ddouglasz: Updated here: https://github.com/commercetools/merchant-center-application-kit/pull/3432/commits/02f38c5d523f11b8d85fab44adf52dc0c4cc4561

Thank you

ddouglasz:
All feedback updated here: https://github.com/commercetools/merchant-center-application-kit/pull/3432/commits/02f38c5d523f11b8d85fab44adf52dc0c4cc4561

Thank you

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -846,7 +846,7 @@ describe('when clicking on navbar menu toggle', () => {
     });
 
     // Check that the "Support" link is rendered when the menu is expanded
-    expect(navbarRendered.getByText('Support')).toBeInTheDocument();
+    expect(navbarRendered.getAllByText('Support')).toHaveLength(1);
```
ddouglasz, line 849: Actually here, because we are checking when expanded, we have only 1 support element in the dom from the Support label and not from the tooltip.
`{!isMenuOpen && <SupportMenuTooltipContainer>...`

CarlosCortizasCT, line 849: I'm sorry but I don't understand this change. I thought we were updating the original assertion because the text were in the HTML twice, but if we're only having one, then there's no need to change the original assertion, right?

packages/application-shell/src/components/application-shell/application-shell.spec.js
```diff
@@ -866,7 +866,7 @@ describe('when clicking on navbar menu toggle', () => {
     });
 
     // Verify that the "Support" link is not rendered when the menu is collapsed
-    expect(navbarRendered.queryByText('Support')).toBeNull();
+    expect(navbarRendered.getAllByText('Support')).toHaveLength(1);
```
ddouglasz, line 869: Here, it is strange that the `display: none` property is supposed to take out the tooltip support from dom but it only just doesnt show it and its still in the dom

packages/application-shell/src/components/application-shell/application-shell.spec.js
ddouglasz: Yes, I just left a comment on it now that I think about it. Normally, there should be nothing in the dom at this point but I am investigating why `display: none` is not showing anything, as expected, but still shows the tooltip in the dom. 
image

packages/application-shell/src/components/application-shell/application-shell.spec.js
CarlosCortizasCT: Maybe there's a confusion about what `display: none` actually does.

It is not supposed to remove the element from the DOM but from the HTML rendering flow. This means it will not be included in the visible elements.
The main different between this and `visibility: hidden` is that this one hides the content but the content still uses its space (it's like making it transparent).

The only way for not including the element in the DOM is not to render it from the React component.

ddouglasz: This makes sense. Thank you 🙏🏾

Updated here: https://github.com/commercetools/merchant-center-application-kit/pull/3432/commits/955b9d531f31df690caa06d8ff239eed6b36f720

packages/application-shell/src/components/application-shell/application-shell.spec.js
CarlosCortizasCT: Ok, so I think I figured out what's happening: when we first render the menu expanded, the tooltip is not rendered, so there's only one "Support" text in the DOM with is the title of the menu item. When we collapse the menu, we don't render the menu item anymore, but we render the tooltip with the same text. So we always have one "Support" text.

Perhaps we can use the `tooltip` [role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role) for the tooltip so we can target the test selector using that role. _(for all menu items tooltips)_

ddouglasz: Yes, exactly what I am explaining here: https://github.com/commercetools/merchant-center-application-kit/pull/3432#discussion_r1515960589
>Actually here, because we are checking when expanded, we have only 1 support element in the dom from the Support label and not from the tooltip.
{!isMenuOpen && <SupportMenuTooltipContainer>...

ddouglasz: >Perhaps we can use the tooltip [role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role) for the tooltip so we can target the test selector using that role. (for all menu items tooltips)

@CarlosCortizasCT just to double-check what you mean here, do you mean have another test case to check for all menu item tooltips?

CarlosCortizasCT: What I mean is that, every time we us the Tooltip component, we should be passing the `role="tooltip"` attribute.
If we do so, then we can update the current failing test to better differentiate between the menu item label and its tooltip.

ddouglasz: Updated here: https://github.com/commercetools/merchant-center-application-kit/pull/3432/commits/4036a48c64968dd4531520268a00f2186803198f

Thank you

CarlosCortizasCT: I pushed a commit in order to improve tests assertions: 5fe19b4977bce8c436e1555de4032ec0a04731c4

.changeset/silver-buses-collect.md
```diff
@@ -0,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': minor
+---
+
+When the navbar is collapsed, onhover, the support menu item should display a tooltip.
```
CarlosCortizasCT, line 5:
```suggestion
Updated the "Support" navigation menu item so it also has a tooltip when the menu is collapsed (same as the rest of the menu root items).
```

CarlosCortizasCT (approved):
Looks good to me now 👍 

Thanks!

packages/application-shell/src/components/application-shell/application-shell.spec.js
ddouglasz: Thank you 🙏🏾

kark (commented):
Great job overall!
Just a minor observation: I'd be interested to hear @FilPob's thoughts on this. Unlike in other menu items, the tooltip here is added to a container larger than the menu item itself. In my view, this leads to some unexpected behavior. It might be worth addressing
image

ddouglasz:
Good catch! @kark I fixed it here: https://github.com/commercetools/merchant-center-application-kit/pull/3432/commits/6df9b796801d4e2b4683cbdd06e959a8e01a8bcb
Thank you 🙏🏾

packages/application-shell/src/components/navbar/main-navbar.styles.ts
```diff
@@ -26,11 +26,29 @@ const FixedMenu = styled.div`
...
+const SupportMenuTooltipContainer = styled.div`
+  position: fixed;
+  left: ${NAVBAR.sublistIndentationWhenCollapsed};
+  display: none;
+  align-items: center;
+  justify-content: flex-start;
+  height: ${NAVBAR.itemSize};
+  z-index: 20001;
+
+  ${MenuListItem} ${ItemIconText}
```
kark, line 38: I guess it's not used
```suggestion
```

kark (approved):
🙌

packages/application-shell/src/components/navbar/main-navbar.styles.ts
CarlosCortizasCT: removed here: 8cfa7c8042987160af5a9a3eb15758732848f874

PR #3439, opened 2024-03-12
Allow delaying rendering of ApplicationShell contents until all feature flags are fetched

CarlosCortizasCT:
#### Summary

Allow delaying rendering of ApplicationShell contents until all feature flags are fetched

#### Description

The idea implemented in this PR is that we want to have the opportunity to delay the applications rendering until the remote feature flags are resolved.
This is to avoid potential UI flickers when the initial rendering of the application is based on a feature flag.

Currently, since we don't wait for the flags, the application is first rendered with the default (hardcoded) values and later, when the remote flags are resolved, the application re-renders. So, if the default value and the remote value are different, users will see this flicker effect.

The new behaviour can be controlled by consumer applications using the new `shouldWaitForRemoteFlags` prop in the `ApplicationShell` component.
Its default value now is `true` so we will be including the behaviour right after the consumer applications update to the upcoming new `app-kit` release.

_(This behaviour is supposed to be temporary, although I think it might help to keep it, but change its default value to not delay the rendering by default.)_

### DIFF
--- a/packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
+++ b/packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
@@ -1,9 +1,9 @@
-import { ReactNode, useMemo } from 'react';
+import { ReactNode, useEffect, useMemo, useRef, useState } from 'react';
 import { useApolloClient } from '@apollo/client/react';
 import combineAdapters from '@flopflip/combine-adapters';
 import httpAdapter from '@flopflip/http-adapter';
 import ldAdapter from '@flopflip/launchdarkly-adapter';
-import { ConfigureFlopFlip } from '@flopflip/react-broadcast';
+import { ConfigureFlopFlip, useAdapterStatus } from '@flopflip/react-broadcast';
 import type { TFlags } from '@flopflip/types';
 import { useApplicationContext } from '@commercetools-frontend/application-shell-connectors';
 import {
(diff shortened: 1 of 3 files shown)


packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
```diff
@@ -17,6 +17,40 @@ import type {
 } from '../../types/generated/mc';
 import AllFeaturesQuery from './fetch-all-features.mc.graphql';
 
+const WAIT_FOR_REMOTE_FLAGS_TIMEOUT = 500; // milliseconds
```
CarlosCortizasCT, line 20: This is the maximum time we will wait for remote flags to be resolved. 
After that, we will render the application anyway (with the default flags values) and, if the flags resolution happens after this time threshold, we ignore it.

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
```diff
@@ -17,6 +17,40 @@ import type {
...
+  children: ReactNode;
+};
+const StatusAwareRenderer = (props: TStatusAwareRendererProps) => {
+  const [shouldRender, setShouldRender] = useState(
+    !props.shouldWaitForRemoteFlags
+  );
+  const { isReady } = useAdapterStatus();
+  const displayTimeout = useRef<NodeJS.Timeout>();
+
+  if (isReady && !shouldRender) {
```
CarlosCortizasCT, line 33: We take action after remote flags resolution only if the timeout have not happened so we haven't rendered the children yet.

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
```diff
@@ -17,6 +17,40 @@ import type {
...
+    setShouldRender(true);
+    clearTimeout(displayTimeout.current);
```
CarlosCortizasCT, line 35: In case we get the flags before the time threshold, we stop the timeout so we won't do anything after that time pases.

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
```diff
@@ -17,6 +17,40 @@ import type {
...
+
+    return () => {
+      clearTimeout(displayTimeout.current);
+    };
+  }, []);
+
+  return shouldRender ? <>{props.children}</> : null;
+};
+StatusAwareRenderer.defaultProps = {
+  shouldWaitForRemoteFlags: process.env.NODE_ENV === 'test' ? false : true,
```
CarlosCortizasCT, line 51: By default we will delay the rendering so consumer applications don't have to be updated.

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
```diff
@@ -25,6 +59,7 @@ type TSetupFlopFlipProviderProps = {
   ldClientSideId?: string;
   children: ReactNode;
   shouldDeferAdapterConfiguration?: boolean;
+  shouldWaitForRemoteFlags?: boolean;
```
CarlosCortizasCT, line 62: This prop is included so it will be easier to disable the render delaying.
Instead of removing this code, release a new `app-kit` version and update consumer applications, we will just need to update consumer apps with this prop.
Then we can work in rolling back these changes without rushing.

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
tdeekens, line 51: Given we will use the mainly internal we could force to pass the flag as true. Do you have an opinion on that?

tdeekens: Affirmative.

tdeekens (approved):

ddouglasz (approved):

mmaltsev-ct (approved):

PR #3451, opened 2024-03-18
Fix starter templates tests

### LINKED ISSUE #3443: Typescript starter tests are failing
**Describe the issue**

Hey there, we've set up the typescript starter with the following command last week:
```
npx @commercetools-frontend/create-mc-app@latest \
  custom-app-starter \
  --template starter-typescript
```
Everything works fine as described in the docs, however, the generated tests for the channel example are not passing. It seems like there is an error with the GraphQL mock.

**To Reproduce**

1. Set up a new typescript-starter as described on https://docs.commercetools.com/custom-applications/getting-started.
2. Execute `yarn test`

**Expected behavior**

All tests for the example app should be passing.

**Screenshots**

The following error is logged:
```
PASS test [src/components/welcome/welcome.spec.tsx](custom-app-starter/src/957503b5c6b79682ba3533abdc4c7e9d1713a22b/src/components/welcome/welcome.spec.tsx)
FAIL test src/components/channels/channels.spec.tsx (7.349 s)
  ● should render channels and paginate to second page
    Unable to find an element with the text: channel-key-0. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.
    Ignored nodes: comments, script, style
    <body>
      <div>
        <div
          style=""
        >
          <div
            class="css-m39x7a-baseStyles-getStyles"
            id="notifications-global"
          />
          <div
            class="css-177j1rc-baseStyles-getStyles"
            id="notifications-page"
          />
          <div
            class="css-88slgd-baseStyles-getStyles"
            id="notifications-side"
          />
          <div
            class="css-6ent15-Inset"
          >
            <div
              class="css-hig09w-ContentNotification"
            >
              <div
                class="css-1n4f74p-NotificationTypeIcon"
              >
                <svg
                  class="css-n970in-getIconStyles"
                  height="24"
                  role="img"
                  viewBox="0 0 24 24"
                  width="24"
                  xmlns="http://www.w3.org/2000/svg"
                >
                  <path
                    d="M12 21a8.8 8.8 0 0 1-3.511-.709 9 9 0 0 1-2.857-1.933 9.2 9.2 0 0 1-1.923-2.868A8.8 8.8 0 0 1 3 11.979q0-1.869.709-3.502a9.1 9.1 0 0 1 1.923-2.846 9.1 9.1 0 0 1 2.857-1.922A8.8 8.8 0 0 1 12 3q1.869 0 3.512.709a9.1 9.1 0 0 1 2.857 1.922 9.1 9.1 0 0 1 1.922 2.846A8.7 8.7 0 0 1 21 11.98a8.8 8.8 0 0 1-.709 3.511 9.2 9.2 0 0 1-1.922 2.868 9 9 0 0 1-2.857 1.933A8.8 8.8 0 0 1 12 21m0-2.277q2.814 0 4.768-1.965 1.956-1.966 1.955-4.78 0-1.052-.31-2.04a6.4 6.4 0 0 0-.936-1.847l-9.365 9.365q.816.645 1.826.955 1.01.312 2.062.312m-5.477-2.857 9.343-9.386a6.5 6.5 0 0 0-1.836-.902q-.977-.3-2.03-.301-2.814 0-4.768 1.943-1.955 1.945-1.955 4.759 0 1.052.31 2.04.313.989.936 1.847"
                  />
                </svg>
              </div>
              <div
                class="css-n4x9n6-ContentNotification"
              >
                <p
                  class="css-g23bwx-bodyStyles"
                >
                  Error message not found.
                </p>
              </div>
            </div>
          </div>
        </div>
      </div>
    </body>
      70 |
      71 |   // First page
    > 72 |   await screen.findByText('channel-key-0');
         |                ^
      73 |   expect(screen.queryByText('channel-key-22')).not.toBeInTheDocument();
      74 |
      75 |   // Go to second page
      at waitForWrapper (node_modules/.pnpm/@testing-library+dom@8.20.1/node_modules/@testing-library/dom/dist/wait-for.js:166:27)
      at node_modules/.pnpm/@testing-library+dom@8.20.1/node_modules/@testing-library/dom/dist/query-helpers.js:86:33
      at Object.findByText ([src/components/channels/channels.spec.tsx](custom-app-starter/src/957503b5c6b79682ba3533abdc4c7e9d1713a22b/src/components/channels/channels.spec.tsx):72:16)
```

**Additional context**

It seems like the mocked GQL request is failing and the app renders the error state:
```
 const { channelsPaginatedResult, error, loading } = useChannelsFetcher({
    page,
    perPage,
    tableSorting,
  });

  if (error) {
    return (
      <ContentNotification type="error">
        <Text.Body>{getErrorMessage(error)}</Text.Body>
      </ContentNotification>
    );
  }
```
This would also explain why the tests for the welcome page are working fine, as there are no GQL requests on that page.

The test code nor the page code was changed, we used everything as it was generated.

The same error is happening with the custom-view typescript starter as well.

CarlosCortizasCT:
#### Summary

Fix starter templates tests

fixes #3443

#### Description

**TL;DR** -> there'a s problem with a transitive dependency

Running starter templates tests fail due to GraphQL requests mocks not working. When inspecting the GraphQL error, we just see a generic message:
```
ApolloError: Error message not found.
```

After a lot of debugging, I found an inconsistency in dependencies declaration in the http mocking library we use (`msw`).
This library has a dependency on the `headers-polyfill` package set with this version range: `^3.1.0`, which means that any new minor or patch versions of that package can and will be used.

The main problem is that ` headers-polyfill` package removed some public APIs in a minor version ([v3.3.0](https://github.com/mswjs/headers-polyfill/releases/tag/v3.3.0)) and `msw` relies on one of those ([reference](https://github.com/mswjs/msw/blob/v0.49.3/src/node/SetupServerApi.ts#L77)).

When starting a new Custom Application or Custom View project, we download the template and install all the (latest) dependencies, which makes the `headers-polyfill` `v3.3.0` get installed in the `node_modules` with the missing helper methods the version of `msw` we use (`0.49.3`) relies on.

The problem does not happen when running the starter templates tests in this repository as we have an older version on the `headers-polyfill` package declared in the dependencies lock file ([reference](https://github.com/commercetools/merchant-center-application-kit/blob/main/pnpm-lock.yaml#L1939)).

Perhaps we can use this issue to start thinking about upgrading `msw` to its latest version (we are far behind).

### DIFF
--- a/.github/actions/test-template-action/action.yml
+++ b/.github/actions/test-template-action/action.yml
@@ -59,6 +59,10 @@ runs:
       shell: bash
       run: pnpm ${{ inputs.typecheck-command }}
 
+    - name: Running integration tests
+      shell: bash
+      run: pnpm test ${{ inputs.application-type == 'custom-application' && 'application-templates' || 'custom-views-templates'}}/${{ inputs.template-name }}/**/*.spec.js
+
     # https://github.com/bahmutov/cypress-gh-action-split-install/blob/ca3916d4e7240ebdc337825d2d78eb354855464b/.github/workflows/tests.yml#L23-L30
     # https://github.com/marketplace/actions/cypress-io#custom-install
     - name: Restoring Cypress cache
--- a/.github/actions/test-template-action/action.yml
+++ b/.github/actions/test-template-action/action.yml
@@ -61,7 +61,7 @@ runs:
 
     - name: Running integration tests
       shell: bash
-      run: pnpm test ${{ inputs.application-type == 'custom-application' && 'application-templates' || 'custom-views-templates'}}/${{ inputs.template-name }}/**/*.spec.js
+      run: pnpm test -- --testRegex="${{ inputs.application-type == 'custom-application' && 'application-templates' || 'custom-views-templates'}}/${{ inputs.template-name }}/.*\\.spec\\.(js\ts)"
 
     # https://github.com/bahmutov/cypress-gh-action-split-install/blob/ca3916d4e7240ebdc337825d2d78eb354855464b/.github/workflows/tests.yml#L23-L30
     # https://github.com/marketplace/actions/cypress-io#custom-install
--- a/.github/actions/test-template-action/action.yml
+++ b/.github/actions/test-template-action/action.yml
@@ -61,7 +61,7 @@ runs:
 
     - name: Running integration tests
       shell: bash
-      run: pnpm test -- --testRegex="${{ inputs.application-type == 'custom-application' && 'application-templates' || 'custom-views-templates'}}/${{ inputs.template-name }}/.*\\.spec\\.(js\ts)"
+      run: pnpm test -- --testRegex="${{ inputs.application-type == 'custom-application' && 'application-templates' || 'custom-views-templates'}}/${{ inputs.template-name }}/.*\\.spec\\.(js|ts)"
 
     # https://github.com/bahmutov/cypress-gh-action-split-install/blob/ca3916d4e7240ebdc337825d2d78eb354855464b/.github/workflows/tests.yml#L23-L30
     # https://github.com/marketplace/actions/cypress-io#custom-install
--- a/application-templates/starter-typescript/package.json
+++ b/application-templates/starter-typescript/package.json
@@ -99,6 +99,7 @@
     "@types/react": "<18",
     "@types/react-dom": "<18",
     "@types/react-router": "<6",
-    "@types/react-router-dom": "<6"
+    "@types/react-router-dom": "<6",
+    "headers-polyfill": "3.2.5"
   }
 }
--- a/application-templates/starter/package.json
+++ b/application-templates/starter/package.json
@@ -97,6 +97,7 @@
     "@types/react": "<18",
     "@types/react-dom": "<18",
     "@types/react-router": "<6",
-    "@types/react-router-dom": "<6"
+    "@types/react-router-dom": "<6",
+    "headers-polyfill": "3.2.5"
   }
 }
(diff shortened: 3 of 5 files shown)


.github/actions/test-template-action/action.yml
```diff
@@ -59,6 +59,10 @@ runs:
       shell: bash
       run: pnpm ${{ inputs.typecheck-command }}
 
+    - name: Running integration tests
```
CarlosCortizasCT, line 62: We are not currently running the starter templates integration tests and I wanted to include that step in the CI checks, first of all to verify the problem we have when we install the templates, however these tests do not fail because the `headers-polyfill` dependency we install in this repository is a _good_ on (we have it locked in the dependencies lock file), so I'm not sure if it's actually worthy to include this step.

application-templates/starter-typescript/package.json
```diff
@@ -99,6 +99,7 @@
     "@types/react": "<18",
     "@types/react-dom": "<18",
     "@types/react-router": "<6",
-    "@types/react-router-dom": "<6"
+    "@types/react-router-dom": "<6",
+    "headers-polyfill": "3.2.5"
```
CarlosCortizasCT, line 103: This is the newest version we can currently use.

emmenko:
> Perhaps we can use this issue to start thinking about upgrading msw to its latest version (we are far behind).

⬆️ Yes! 💯

emmenko (approved):
Thanks a lot! ❤️

Let's plan the upgrade of msw as a follow up then, I think we can also try using the codemod.

application-templates/starter-typescript/package.json
kark: I agree that this dependency is the root cause. From my observations in [msw `v1.3.1`](https://github.com/mswjs/msw/releases/tag/v1.3.1) the version of this dependency was pinned to `3.2.5` therefore updating msw to v>1.3.0 also resolves the issue.

kark (approved):
Thanks!

PR #3464, opened 2024-03-28
refactor: enable feature flag for Custom Views by default

emmenko:
We want to keep the flag as a long-lived flag (MC API), so that we can remove the one in LD.

### DIFF
--- a/packages/application-components/src/components/custom-views/custom-views-selector/use-custom-views-connector.ts
+++ b/packages/application-components/src/components/custom-views/custom-views-selector/use-custom-views-connector.ts
@@ -1,4 +1,4 @@
-import { useFeatureToggle } from '@flopflip/react-broadcast';
+import { useFlagVariation } from '@flopflip/react-broadcast';
 import { useMcQuery } from '@commercetools-frontend/application-shell-connectors';
 import {
   CustomViewData,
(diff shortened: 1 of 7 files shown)


packages/application-components/src/components/custom-views/custom-views-selector/use-custom-views-connector.ts
```diff
@@ -23,8 +23,10 @@ type TUseCustomViewsFetcher = (props: TUseCustomViewsFetcherParams) => {
 export const useCustomViewsConnector: TUseCustomViewsFetcher = ({
   customViewLocatorCode,
 }) => {
+  const enableCustomViews = useFlagVariation(featureFlags.CUSTOM_VIEWS);
   const areCustomViewsEnabled =
-    useFeatureToggle(featureFlags.CUSTOM_VIEWS) &&
+    // @ts-ignore In case it's coming from the MC API, it's an object { value: boolean }.
+    (enableCustomViews?.value ?? enableCustomViews) &&
```
emmenko, line 29: Just to evaluate it as a fallback.

packages/application-shell/src/components/setup-flop-flip-provider/setup-flop-flip-provider.tsx
```diff
@@ -152,7 +152,7 @@ export const SetupFlopFlipProvider = (props: TSetupFlopFlipProviderProps) => {
 
   const defaultFlags = useMemo(
     () => ({
-      ...featureFlags.FLAGS,
+      ...featureFlags.DEFAULT_FLAGS,
```
emmenko, line 155: This is the main fix. We only want to pass default flags here, not the short-lived flags.

packages/constants/src/feature-toggles.ts
```diff
@@ -21,6 +21,10 @@ export const CUSTOM_VIEWS = 'enableCustomViews';
...
 
 export const FLAGS = {
-  [CUSTOM_VIEWS]: false,
   [RECOLOURING]: false,
 };
+
+// Long-lived feature flags, defined in the MC API.
+export const DEFAULT_FLAGS = {
+  [CUSTOM_VIEWS]: { value: true },
+};
```
emmenko, line 30: Similarly to our internal setup, these are considered long-lived flags and they are defined as an object.

kark (approved):
💯
Thanks!

ddouglasz (approved):

PR #3476, opened 2024-04-04
chore: to update flopflip

tdeekens:
#### Summary

This updates flopflip and sets a cachePolicy to network-only for loading flags.

#### Description

This is needed to reload internal flags when we change the adapter context. It is needed to use internal entitlement flags more.


emmenko (approved):
Thanks!

PS: don't forget the changeset

kark (approved):

ddouglasz (approved):

PR #3478, opened 2024-04-04
Adjust documentation links used in cli

CarlosCortizasCT:
#### Summary

Adjust documentation links used in cli.

#### Description

In this PR we update the links to the docs we show in the terminal when a user bootstraps a new Custom Application or Custom View.

We've unified the link for both of them so we now link [here](https://docs.commercetools.com/merchant-center-customizations/).


emmenko (approved):
Thanks

PR #3480, opened 2024-04-04
fix(cypress): onBeforeLoad being called during login

tdeekens:
#### Summary

The `onBeforeLoad` never seems to be called which this intends to fix.

#### Description

In the `loginByOidc` we call the `onBeforeLoad` like:

```js
if (commandOptions.onBeforeLoad) {
  commandOptions.onBeforeLoad(win);
}
```

I noticed that we do not do this when logging in via the form. This causes issues internally when we use the general `loginToMerchantCenter` command which we often pass an `onBeforeLoad` to e.g. set local storage values to hide collapse the menu or hide certain models (like for rebranding). 

We could either invoke the `onBeforeLoad` in the `authCallback` or when navigating to the `initialRoute`. The latter seems unexpected as you then only invoke the callback when another parameter is passed.

### DIFF
--- a/packages/cypress/src/add-commands/login.ts
+++ b/packages/cypress/src/add-commands/login.ts
@@ -125,7 +125,7 @@ function loginByForm(commandOptions: CommandLoginOptions) {
     // const mcFrontendHostname = mcUrl.hostname.replace('mc-api', 'mc');
 
     function authCallback() {
-      cy.visit(url);
+      cy.visit(url, { onBeforeLoad: commandOptions.onBeforeLoad });
 
       // NOTE: using `cy.origin` is currently disabled as it does not seem to properly work.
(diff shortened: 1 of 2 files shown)


packages/cypress/src/add-commands/login.ts
```diff
@@ -125,7 +125,7 @@ function loginByForm(commandOptions: CommandLoginOptions) {
     // const mcFrontendHostname = mcUrl.hostname.replace('mc-api', 'mc');
 
     function authCallback() {
-      cy.visit(url);
+      cy.visit(url, { onBeforeLoad: commandOptions.onBeforeLoad });
```
tdeekens, line 128: Verified this locally with a patched package.

kark (approved):

PR #3482, opened 2024-04-05
feat(navbar keyboard navigation): optionally navigate to submenu items using the `Tab` and `Enter` keys.

ddouglasz:
Follow up to navbar keyboard navigation.

#### Summary

Optionally navigate to submenu items using the `Tab` and `Enter` keys. This way, we can choose to navigate through the submenu of a menu item that has a submenu or skip the submenu and move to the next menu item. 
Some context can be found in the github conversation [here](https://github.com/commercetools/merchant-center-application-kit/pull/3306#issuecomment-1855731612).

### DIFF
--- a/packages/application-shell/src/components/navbar/menu-items.tsx
+++ b/packages/application-shell/src/components/navbar/menu-items.tsx
@@ -252,7 +252,7 @@ type MenuItemProps = {
     | FocusEventHandler<HTMLElement>;
   children: ReactNode;
   identifier?: string;
-  isRightArrowPressed?: boolean;
+  isEnterKeyDown?: boolean;
   handleSubMenuKeyDown?: (e: React.KeyboardEvent<HTMLLIElement>) => void;
 };
 const MenuItem = (props: MenuItemProps) => {
@@ -271,7 +271,7 @@ const MenuItem = (props: MenuItemProps) => {
       isActive={props.isActive}
       isRouteActive={Boolean(props.isMainMenuRouteActive)}
       isCollapsed={!props.isMenuOpen}
-      isRightArrowPressed={props.isRightArrowPressed}
+      isEnterKeyDown={props.isEnterKeyDown}
       onKeyDown={props.handleSubMenuKeyDown}
     >
       <ItemContent>{props.children}</ItemContent>
--- a/packages/application-shell/src/components/navbar/menu-items.tsx
+++ b/packages/application-shell/src/components/navbar/menu-items.tsx
@@ -203,7 +203,6 @@ const MenuGroup = forwardRef<HTMLUListElement, MenuGroupProps>((props, ref) => {
         isSublistActiveWhileIsMenuCollapsed
       }
       onKeyDown={props.handleKeyDown}
-      tabIndex={props.level === 2 ? 0 : -1}
       className={classnames(
         {
           'sublist-expanded__active': isSublistActiveWhileIsMenuExpanded,
@@ -253,8 +252,8 @@ type MenuItemProps = {
     | FocusEventHandler<HTMLElement>;
   children: ReactNode;
   identifier?: string;
-  isSubmenOpen?: boolean;
   handleSubMenuKeyDown?: (e: React.KeyboardEvent<HTMLLIElement>) => void;
+  onKeyDown?: (e: React.KeyboardEvent<HTMLLIElement>) => void;
 };
 const MenuItem = (props: MenuItemProps) => {
   return (
@@ -265,6 +264,7 @@ const MenuItem = (props: MenuItemProps) => {
       onMouseLeave={props.onMouseLeave as MouseEventHandler<HTMLElement>}
       onFocus={props.onMouseEnter as FocusEventHandler<HTMLElement>}
       onBlur={props.onMouseLeave as FocusEventHandler<HTMLElement>}
+      onKeyDown={props.onKeyDown}
       data-menuitem={props.identifier}
       className={classnames({
         active: props.isActive,
@@ -272,8 +272,6 @@ const MenuItem = (props: MenuItemProps) => {
       isActive={props.isActive}
       isRouteActive={Boolean(props.isMainMenuRouteActive)}
       isCollapsed={!props.isMenuOpen}
-      isSubmenOpen={props.isSubmenOpen}
-      onKeyDown={props.handleSubMenuKeyDown}
     >
       <ItemContent>{props.children}</ItemContent>
     </MenuListItem>
@@ -289,6 +287,7 @@ export type MenuItemLinkProps = {
   onClick?: (event: SyntheticEvent<HTMLAnchorElement>) => void;
   useFullRedirectsForLinks?: boolean;
   isSubmenuLink?: boolean;
+  isSubmenuFocused?: boolean;
 };
 const menuItemLinkDefaultProps: Pick<MenuItemLinkProps, 'exactMatch'> = {
   exactMatch: false,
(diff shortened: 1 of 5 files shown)


ddouglasz:
The key features, as concluded on a call with @FilPob and @CarlosCortizasCT  should be:

- [x] - `Tab` key to focus on a menuItem
- [x] - If menu Item has subMenu, then `Enter` Key to navigate to the subMenuItems
- [x] - If no SubMenu, `Tab` key to Focus on next Available Item
- [x] - `Enter` Key to click on a SubMenu Item
- [x] - `Escape` Key to exit subMenu

kark (commented):
Big thanks for the improvements, it's a big step forward 🚀
Here's my feedback:
1. I'm uncertain about one aspect—when a menu item is focused, the submenu appears, but then closes upon the next <kbd>Tab</kbd> key press. I would expect that once the submenu is open we are able to navigate it with the next <kbd>Tab</kbd> key press. Currently, it feels a bit ambiguous whether pressing <kbd>Enter</kbd> allows entry (although if this was a deliberate decision, I'm ok with that)

2. Additionally, I've encountered what seems to be a bug. Pressing <kbd>Enter</kbd> on any menu item results in submenu items always being focusable.

ddouglasz:
> 1. I'm uncertain about one aspect—when a menu item is focused, the submenu appears, but then closes upon the next Tab key press. I would expect that once the submenu is open we are able to navigate it with the next Tab key press. Currently, it feels a bit ambiguous whether pressing Enter allows entry (although if this was a deliberate decision, I'm ok with that)

Thank you Kacper,
The agreed flows are:

**Scenario 1:**  When a menu item is focused, `Enter` key press should focus on the first item of it's submenu items, if there are any, and tab key to navigate within the submenu. 

**Scenario 2:** If the `Tab` key is pressed when a menu item is focused then the focus should move to the next menu item and not into the submenu item.

> 2. Additionally, I've encountered what seems to be a bug. Pressing Enter on any menu item results in submenu items always being focusable.
I will look into this, thank you! 👍🏾

kark:
Thanks Douglas 🙏, 
I'm not sure if this is the definite answer to my concern

> **Scenario 1:** When a menu item is focused, `Enter` key press should focus on the first item of it's submenu items, if there are any, and tab key to navigate within the submenu.
> 
> **Scenario 2:** If the `Tab` key is pressed when a menu item is focused then the focus should move to the next menu item and not into the submenu item.

Those scenarios are clear to me 👍 My question was: if a menu item gets focus, should the submenu open automatically or only after <kbd>Enter</kbd> press?

ddouglasz:
>  My question was: if a menu item gets focus, should the submenu open automatically or only after Enter press?
Thanks again @kark , so that was the initial implementation I had but later updated to show a preview of the submenu when a menu item gains focus. it was one of the requirements we initially discussed with @FilPob (although in our last call, he was okay with it implemented either ways)

CarlosCortizasCT:
To be honest, I don't like the behaviour of displaying the submenu on keyboard navigation focus, but that's a design request so we must evaluate if it's not too complicated to implement.

CarlosCortizasCT:
Thanks a lot for working on this, @ddouglasz 🙇 

From the functionality point of view, it is working as I think it was expected but I'd like for us to align on what we actually want to have as a feature.

Currently we have implemented the navigation using the `tab` and `enter` keys.
So when I get into the first root item of the navbar, I can get into their submenu items by using the enter key. If I then move around the submenu items and want to go back to the root items, I need to iterate over the previous submenu items. I don't have a quick access (eg: Escape key) to go back to the navbar root level.

Also, I think we said the tabular navigation was a first step but that we could build the arrows keys navigation also as an extra feature.

To me it's ok to merge the feature as it currently is, but I want us to be aligned in terms of the expectations of this PR and if we want to do more follow-up work.

cc/ @FilPob

FilPob:
totally agree. I think for the scope of this PR the behaviour is OK. For follow up work we should align first on the expectations

CarlosCortizasCT:
@ddouglasz could you please updated the title and description of the PR accordingly? 🙏

ddouglasz:
> @ddouglasz could you please updated the title and description of the PR accordingly? 🙏

Updated. Thank you

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -191,10 +191,37 @@ export const ApplicationMenu = (props: ApplicationMenuProps) => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [props.isMenuOpen]); // <-- (re)run this only when the all menu expands
 
+  // Observe submenu to ensure it is not showing on screen when focus is out of it's focus
+  useEffect(() => {
+    const currentRef = submenuRef.current;
+    const observer = new IntersectionObserver((entries) => {
```
CarlosCortizasCT, line 197: I'm sorry but I fail to understand this part so I would like to know if you would be up to a sync about it.

Also, bear in mind if we observe an element in the effect, we must unobserve it in the effect cleanup function to avoid memory leaks.

ddouglasz: Actually this was a leftover from the initial implementation workflow. It was supposed to ensure the open state of the submenu does not persist when we skip it to another element in the dom. 

Removed here: https://github.com/commercetools/merchant-center-application-kit/pull/3482/commits/083c49a0a00a8fb2f187f1db5db125770c866d58

Thank you

CarlosCortizasCT (approved):
Looks good to me.

Thanks 👍

Rhotimee (approved):
🙌🏽

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -195,6 +195,17 @@ export const ApplicationMenu = (props: ApplicationMenuProps) => {
...
+  const handleKeyDown = (e: React.KeyboardEvent<HTMLLIElement>) => {
+    const currentlyFocusedItem = submenuRef.current?.querySelector(':focus');
+
+    if (e.key === 'Enter') {
+      setIsSubmenuFocused(true);
+      if (!currentlyFocusedItem) {
+        submenuRef.current?.querySelector('a')?.focus();
+      }
+    }
+  };
```
kark, line 207: Just to double check regarding my previous comment:

> Additionally, I've encountered what seems to be a bug. Pressing <kbd>Enter</kbd> on any menu item results in submenu items always being focusable.

What we're doing in this handler - once the user presses <kbd>Enter</kbd> `isSubmenuFocused` is set to `true` for good and this state never changes again. Which means that once user presses <kbd>Enter</kbd> the navigation works as previously (so the user has to navigate through all the menu and submenu items).

Is this expected and are we ok with that?

packages/application-shell/src/components/navbar/menu-items.tsx
```diff
@@ -249,6 +252,8 @@ type MenuItemProps = {
     | FocusEventHandler<HTMLElement>;
   children: ReactNode;
   identifier?: string;
+  handleSubMenuKeyDown?: (e: React.KeyboardEvent<HTMLLIElement>) => void;
```
kark, line 255: This seems this prop is never used

kark (commented):
Thanks Douglas. Just a final comment from my end

packages/application-shell/src/components/navbar/navbar.tsx
ddouglasz: No it is not the expected behaviour, this is a bug. Sorry, I thought I addressed this in the initial comment you made. I understand you now. Will look into it. 
Thank you.

packages/application-shell/src/components/navbar/menu-items.tsx
ddouglasz: Updated here: https://github.com/commercetools/merchant-center-application-kit/pull/3482/commits/4e07032305fa1fd7a4aa2a4655e77c65b2a8ad5c
Thank you

packages/application-shell/src/components/navbar/navbar.tsx
ddouglasz: Fixed here: https://github.com/commercetools/merchant-center-application-kit/pull/3482/commits/4e07032305fa1fd7a4aa2a4655e77c65b2a8ad5c
Thank you

kark (approved):
Thanks again, works as expected 🙌

PR #3485, opened 2024-04-09
refactor(application-shell): to remove fetchPolicy again

tdeekens:
#### Summary

The context now changes as the projectKey is passed. Apollo will automatically refetch once the context changes to my understanding - due to the projectKey changes.

### DIFF
--- a/.changeset/tame-hats-glow.md
+++ b/.changeset/tame-hats-glow.md
@@ -1,0 +1,5 @@
+---
+"@commercetools-frontend/application-shell": patch
+---
+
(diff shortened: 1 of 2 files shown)


kark (approved):

ddouglasz (approved):

CarlosCortizasCT (approved):

tdeekens:
Turns out this won't work. The context on a query does not determine it being re-run. Only the variables would. We could add that but it requires work on the MC API. Plus it can cause a stale cache if the variable isn't used. In all as a result the network-only feels correct.

PR #3486, opened 2024-04-09
refactor: regex to define start of match

emmenko:
Fixes https://github.com/commercetools/merchant-center-application-kit/security/code-scanning/10


kark (approved):
Thanks for addressing this case!

CarlosCortizasCT (approved):

PR #3489, opened 2024-04-10
chore: update `@babel/traverse` to use security patch version

emmenko:
Fixes https://github.com/commercetools/merchant-center-application-kit/security/dependabot/162


CarlosCortizasCT:
@emmenko There are so many changes in the lock file.

Could you maybe share how did updating the file solves the issue? 🙏 
Is it `@babel/traverse` a transitive dependency with a range version declaration and in this PR we're just updating all range versions to the latests available?

CarlosCortizasCT (approved):

PR #3491, opened 2024-04-11
docs: changeset

emmenko:
Leftover of #3490


kark (approved):
Thanks! Sorry for overlooking that!

PR #3493, opened 2024-04-11
fix(TabularModalPage): styles in custom views

kark:
#### Summary

This PR fixes the styles of `TabularModalPage` rendered in a Custom View.
Unfortunately, I couldn't find way to apply styling through [Custom View specific css token overrides added to the theme provider](https://github.com/commercetools/merchant-center-application-kit/blob/c2bba1d065b6fd7882e6feb9162d91538962d85d/packages/application-shell/src/components/custom-view-shell/custom-view-shell.styles.ts#L11-L26), since `paddingForTabularPageHeader` used here is valid for `TabularDetailPage` and `TabularMainPage`. Therefore, the solution I thought was to introduce a wrapper that only shows up in the `TabularModalPage` and only when rendered in a Custom View and overwrites the styles of the `TabularPageContainer`.

Current state:
image

It does not change the current styling of `TabularModalPage` rendered in a Custom Application, and also does not change the way `TabularDetailPage` and `TabularMainPage` are rendered in Custom Views.

After changes:
|  TabularModalPage in Custom Application | TabularDetailPage in Custom View  |  TabularModalPage in Custom View |
|---|---|---|
| Screenshot 2024-04-11 at 15 01 24  |  Screenshot 2024-04-11 at 15 01 37 | Screenshot 2024-04-11 at 15 01 46 |

The PoC for the reviewers to play around was pushed to `SHIELD-1153-fix-tabularmodalpage-in-custom-views-poc` (to test out this PR locally in the playground application)

### DIFF
--- a/packages/application-components/src/components/internals/tabular-page.tsx
+++ b/packages/application-components/src/components/internals/tabular-page.tsx
@@ -1,8 +1,11 @@
-import type { ReactNode } from 'react';
+import { Fragment, type ReactNode } from 'react';
 import { css } from '@emotion/react';
 import styled from '@emotion/styled';
+import type { ApplicationWindow } from '@commercetools-frontend/constants';
 import { designTokens as uiKitDesignTokens } from '@commercetools-uikit/design-system';
 import { designTokens as appKitDesignTokens } from '../../theming';
+
+declare let window: ApplicationWindow;
 
 const TabControls = styled.div`
   margin-top: ${uiKitDesignTokens.spacingS};
@@ -59,9 +62,28 @@ const CustomViewsSelectorWrapper = styled.div`
   margin: ${appKitDesignTokens.marginForCustomViewsSelectorAsTabular};
 `;
 
+const CustomViewsTabularPageContainer = (props: { children: ReactNode }) => {
+  const isRenderedInCustomView = Boolean(window.app.customViewId);
+  // wrapper which only appears when rendered in a Custom View
+  const Wrapper = isRenderedInCustomView ? 'div' : Fragment;
+  return (
+    <Wrapper
+      css={css`
+        > div {
+          padding: ${uiKitDesignTokens.spacing50} ${uiKitDesignTokens.spacing55}
+            0 !important;
+        }
+      `}
+    >
+      {props.children}
+    </Wrapper>
+  );
+};
+
 export {
   ControlsContainter,
   TabularPageContainer,
   FormControlsContainer,
   CustomViewsSelectorWrapper,
+  CustomViewsTabularPageContainer,
 };
--- a/.changeset/lazy-vans-bow.md
+++ b/.changeset/lazy-vans-bow.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-components': patch
+---
+
+Styling adjustments for TabularModalPage to ensure proper rendering in Custom Views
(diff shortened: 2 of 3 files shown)


ddouglasz (approved):
Looks good to me. 
Thank you 🙏🏾

packages/application-components/src/components/internals/tabular-page.tsx
```diff
@@ -59,9 +62,28 @@ const CustomViewsSelectorWrapper = styled.div`
   margin: ${appKitDesignTokens.marginForCustomViewsSelectorAsTabular};
 `;
 
+const CustomViewsTabularPageContainer = (props: { children: ReactNode }) => {
+  const isRenderedInCustomView = Boolean(window.app.customViewId);
+  // wrapper which only appears when rendered in a Custom View
+  const Wrapper = isRenderedInCustomView ? 'div' : Fragment;
+  return (
+    <Wrapper
```
CarlosCortizasCT, line 70: Thanks for working on this fix.

I'm concerned about the solution though as we're overwriting children styles from a parent and also using the `!important` directive.

I haven't had the time to investigate an alternative so maybe we can merge this solution and I can give it a try as a follow-up.

emmenko: Can we define a new token that we can then override?

emmenko: Hmm but that's what the token is for, to allow overwriting.

Can you try to avoid using this wrapper container and define the token in the correct place?

kark: I'm not sure if I get this advice. Let me explain the current situation.
|  Page |  Header padding value needed in Custom/Internal Apps |  Header padding value needed Custom Views  | Comment |
|---|---|---|---|
| TabularMainPage |  `24px 40px 0` |  `0` |`paddingForTabularPageHeader` overridden in the custom views theme provider |
| TabularDetailPage  |  `24px 40px 0` | `0`  | `paddingForTabularPageHeader` overridden in the custom views theme provider |
| TabularModalPage  |  `24px 40px 0` | `24px 40px 0`  |`paddingForTabularModalPageHeader` token used from the main theme provider, nothing needs to be overridden in the custom views theme provider |

I'm not sure how to avoid using the wrapper and differentiating between TabularModalPage and TabularDetailPage/TabularMainPage tokens here: 
https://github.com/commercetools/merchant-center-application-kit/blob/d16ebd98c1b99a0c0c0e0f9a9df24cc8dcc0f542/packages/application-components/src/components/internals/tabular-page.tsx#L37-L46

kark: @emmenko As discussed, I applied the version referencing a styled component instead of `> div {`, all worked fine. 
Thanks for the suggestion 🙏

CarlosCortizasCT (approved):
Awesome 💯 

Thanks Kacper 🙇

packages/application-components/src/components/internals/tabular-page.tsx
```diff
@@ -59,9 +43,16 @@ const CustomViewsSelectorWrapper = styled.div`
   margin: ${appKitDesignTokens.marginForCustomViewsSelectorAsTabular};
 `;
 
+const TabularModalPageContainer = styled.div`
+  ${TabularPageContainer} {
```
emmenko, line 47: Awesome, thanks! 🙌

PR #3379, opened 2024-01-17
fix(public-page-layout): update logos used in MC authentication flows

tylermorrisford:
#### Summary

This PR updates all logos within `packages/assets/logos/` to use the new branding, using the existing names.

#### Description

| PublicPageLayout Before      | PublicPageLayout After |
| ----------- | ----------- |
| Screenshot 2024-01-25 at 11 25 16 AM |  Screenshot 2024-01-25 at 11 35 05 AM|
| Navbar Before      | Navbar After |
|Screenshot 2024-01-25 at 11 36 38 AM | Screenshot 2024-01-25 at 11 39 28 AM |

PR #3494, opened 2024-04-15
fix: to inject app env script as one liner

emmenko:
No idea why this started "failing" now but in the [graphql-explorer Custom Application](https://github.com/commercetools/merchant-center-graphql-explorer) (built with Vite) the deployment was broken as the CSP policy refused to execute the injected script `window.app = ` due to a mismatch with the SHA hash.

After some debugging and try/error, it appeared that the browser expected the hash to be created from a one-liner of the script.

In other words, instead of

```
<script>
  window.app = 
</script>
```

it should be

```
<script>window.app = </script>
```

Update: it looks like this was introduced due to Prettier in #3379 

image

### DIFF
--- a/packages/mc-html-template/html-docs/application.html
+++ b/packages/mc-html-template/html-docs/application.html
@@ -166,9 +166,7 @@
     __LOADING_SCREEN_JS__
 
     <!-- Application globals -->
-    <script>
-      window.app = __APPLICATION_ENVIRONMENT__;
-    </script>
+    __APPLICATION_ENVIRONMENT__
 
     <!-- Main application chunks -->
     __APPLICATION_SCRIPT_IMPORTS__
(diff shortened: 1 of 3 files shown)


packages/mc-html-template/html-docs/application.html
```diff
@@ -166,9 +166,7 @@
     __LOADING_SCREEN_JS__
 
     <!-- Application globals -->
-    <script>
-      window.app = __APPLICATION_ENVIRONMENT__;
-    </script>
+    __APPLICATION_ENVIRONMENT__
```
emmenko, line 169: This now ensures that prettier doesn't mess with the format

CarlosCortizasCT (approved):
It must have been quite difficult to figure this out.

Awesome work! 💯

emmenko:
Lots of debugging yes 😅

packages/mc-html-template/html-docs/application.html
tdeekens: Wow. That's quite a find.

PR #3495, opened 2024-04-19
refactor: refine some regex

emmenko:
As we are finalizing the stricter regex for project keys, I refined and added some tests for other regex as well.


CarlosCortizasCT (approved):
Thanks 👍

PR #3496, opened 2024-04-22
feat(message-loader): compile i18n messages during webpack build

dogayuksel:
#### Summary

Adds a webpack loader to compile translated messages during build.

#### Description

Introduces it behind an environment variable to release gradually. 

Loader as a module will be resolved by the client's process during build, as a result it should be somehow exposed/published. We add it as an entry point to reconstruct and refer to it as an npm module. 

#### Outstanding
- [x] sniff message contents to determine file type and compile accordingly (KEYVALUEJSON or STRUCTURED_JSON)
- [x] go over webpack loader related intricacies (https://webpack.js.org/api/loaders/)
  - cacheable (?) addDependency (?)
    - no relevant change for 'cacheable', loaders are set cacheable by default, and compile command of react-intl is deterministic, so should work out of the box.
    -  no relevant change for 'addDependency', resource we transform is in the dependency tree by default and we don't read something else from the file system or similar.
  - perhaps better error handling
    - any error thrown will cause the build process to fail, no need to catch and throw a new error, which would hide the actual error.


dogayuksel:
Resulting difference reported by webpack when building discounts app:

Screenshot 2024-04-23 at 14 39 51

packages/mc-scripts/src/config/create-webpack-config-for-production.ts
```diff
@@ -495,6 +495,14 @@ function createWebpackConfigForProduction(
           exclude: /node_modules/,
           use: [require.resolve('graphql-tag/loader')],
         },
+        process.env.ENABLE_MESSAGE_LOADER === 'true' && {
+          test: /i18n\/data\/.*\.json$/,
```
dogayuksel, line 499: Open question: Should this be configurable somehow? Is it possible that consumers of the app-kit have their translations located somewhere else?

packages/mc-scripts/package.json
```diff
@@ -119,6 +122,7 @@
     "@types/moment-locales-webpack-plugin": "^1.2.3",
     "@types/node-fetch": "2.6.11",
     "@types/serve-handler": "^6.1.1",
+    "memfs": "^4.8.2",
```
dogayuksel, line 125: tried using the `mock-fs` for testing first, but it seemed to intervene with the module resolution of webpack. Opted to use this file system mock library, as referenced by the webpack documentation https://webpack.js.org/contribute/writing-a-loader/#testing

packages/mc-scripts/src/config/create-webpack-config-for-production.ts
CarlosCortizasCT: Our [recommendation](https://docs.commercetools.com/merchant-center-customizations/development/translations#extracting-messages-for-translations) is to use the same path for Custom Applications authors, however they can do what they want.

I would say that we can keep the path static. The flow would work anyway and won't affect Custom Application authors which do not use that path.
Since we don't know how many there are, I would vote for keep it simple and revisit in case we learn about actual users require it.

@emmenko what do you think?

CarlosCortizasCT:
Thanks a lot for this contribution @dogayuksel 🙇 

One thing I was wondering about this sentence:
> Introduces it behind an environment variable to release gradually.

Is there anyone driving the rolling out of this feature?
I mean, is someone assigned to announce this feature and push for its adoption?

Thanks.

tdeekens:
> I mean, is someone assigned to announce this feature and push for its adoption?

Thanks for pointing that out. I think we can do this through the Chapter. Maybe we define a rollout plan there if that's alright with everybody.

CarlosCortizasCT:
> > I mean, is someone assigned to announce this feature and push for its adoption?
> 
> Thanks for pointing that out. I think we can do this through the Chapter. Maybe we define a rollout plan there if that's alright with everybody.

Thanks @tdeekens 👍 

I added a note in the [next FE-chapter meeting issue](https://github.com/commercetools/fe-chapter-notes/issues/511) so we don't forget.

dogayuksel:
> > I mean, is someone assigned to announce this feature and push for its adoption?
> 
> Thanks for pointing that out. I think we can do this through the Chapter. Maybe we define a rollout plan there if that's alright with everybody.

These are the next steps, we (priceless) intend to do ([some more detail here](https://commercetools.atlassian.net/browse/PRC-3394)):
- Implement the corresponding vite plugin
- Adjust integration tests to utilise the loader/plugin
  - e.g. when building the sample custom app 
- Make required documentation updates if any

(at this step we could consider removing the opt-in flag and make it the default behaviour)

Then we planned to do the following:
- Use the new loader on prices repo
- Use the new loader for mono-repo applications

packages/mc-scripts/src/config/create-webpack-config-for-production.ts
emmenko: I agree, let's go with this for now.

packages/mc-scripts/src/webpack-loaders/message-loader.spec.ts
```diff
@@ -0,0 +1,77 @@
+import path from 'path';
+import { createFsFromVolume, Volume } from 'memfs';
+import webpack, { type Stats } from 'webpack';
+
+const compiler = (fixture: string): Promise<Stats> => {
+  const compiler = webpack({
+    context: __dirname,
+    entry: `./${fixture}`,
```
emmenko, line 8: Nit: can we move the fixtures within a `fixtures` folder or something?

```suggestion
    entry: `./fixtures/${fixture}`,
```

packages/mc-scripts/src/webpack-loaders/message-loader.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+      { type: 0, value: expect.any(String) },
+    ],
+  };
+
+  describe('when messages are in KEYVALUEJSON format', () => {
+    it('should return compiled messages', async () => {
+      const stats = await compiler('example-keyvalue.json');
+      const output = stats.toJson({ source: true }).modules?.[0].source;
+
+      expect(JSON.parse(output!.toString())).toEqual(expectedOutput);
```
emmenko, line 56: Maybe we can use `toMatchInlineSnapshot` here?

```suggestion
      expect(JSON.parse(output!.toString())).toMatchInlineSnapshot();
```

packages/mc-scripts/src/config/create-webpack-config-for-production.ts
```diff
@@ -495,6 +495,14 @@ function createWebpackConfigForProduction(
           exclude: /node_modules/,
           use: [require.resolve('graphql-tag/loader')],
         },
+        process.env.ENABLE_MESSAGE_LOADER === 'true' && {
```
emmenko, line 498: Nit: let's name this a bit more specifically.

```suggestion
        process.env.ENABLE_WEBPACK_LOADER_I18N_MESSAGE_COMPILATION === 'true' && {
```

packages/mc-scripts/package.json
```diff
@@ -22,7 +22,8 @@
       "./application-runtime.ts",
       "./cli.ts",
       "./postcss.ts",
-      "./webpack.ts"
+      "./webpack.ts",
+      "./webpack-loaders/message-loader.ts"
```
emmenko, line 26:
```suggestion
      "./webpack-loaders/i18n-message-compilation-loader.ts"
```

packages/mc-scripts/src/webpack-loaders/message-loader.ts
emmenko, line 1: I would suggest to name this a bit more specifically like `i18n-message-compilation-loader.ts`

emmenko (commented):
Awesome work, thanks @dogayuksel ! ❤️

tdeekens (approved):
Awesome. Thanks for driving this!

ragafus (approved):
🚀

emmenko (commented):
Thanks, great work! 🙌

Can you also create a changeset? 🤗

emmenko (approved):
Thanks

PR #3504, opened 2024-04-30
chore(flopflip): increase pooling interval from 1 to 15 minutes

YahiaElTai:
#### Summary

This PR increase `flop-flip` polling interval from 1 minute to 15 minutes.

#### Background
I was working locally (and same thing happens on prod) when saw that we have requests for `AllFeatures` going every minute which seemed a bit wasteful since we don't update those values and when we do, we don't expect an instant change since it requires a PR and a CI run

### DIFF
--- a/.changeset/nine-goats-return.md
+++ b/.changeset/nine-goats-return.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': minor
+---
+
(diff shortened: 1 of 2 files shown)


tdeekens (approved):

tdeekens (commented):
Needs a changeset.

CarlosCortizasCT:
Cool! 😎

PR #3527, opened 2024-05-14
refactor(application-shell): testing some changes in the application …

CarlosCortizasCT:
…context

#### Summary

This is just a test PR to debug some CI errors


tdeekens (commented):
Thanks for debugging. 

If I read the PR checks right the fact of adding the field to the query causes the error to appear?

If so I can debug patching app-kit locally also maybe. I already tried a GraphQL query with the field using CURL which works.

CarlosCortizasCT:
> Thanks for debugging.
> 
> If I read the PR checks right the fact of adding the field to the query causes the error to appear?
> 
> If so I can debug patching app-kit locally also maybe. I already tried a GraphQL query with the field using CURL which works.

>  If I read the PR checks right the fact of adding the field to the query causes the error to appear?

That's what I'm trying to find out.

I've just added the field back to the query and the tests are still green so it seems it is not the root cause.

Next step is to check the generated code.

PR #3511, opened 2024-05-06
feat(message-loader): add a simple i18n message compilation plugin for vite

dogayuksel:
#### Summary

Implements the same functionality added to webpack build on https://github.com/commercetools/merchant-center-application-kit/pull/3496 for vite.

PR #3528, opened 2024-05-14
test(message-loader): cover message loader in integration tests

dogayuksel:
### Summary

Modifies build instructions on the pipeline to cover important paths related to recently added message loader functionality. (#3496, #3511)

#### Plugin for vite

We enable message loader on playground build. Playground application already has translations, so a successful build means that 'message loader plugin' has been used.

#### Loader for webpack

We enable the message loader on custom application template builds. Here, we don't have translations, so we add a simple script to simulate a translation workflow just before the build. 'Mock translate i18n messages for the build', copies `core.json` to `en.json`, which in return gets picked up by the message loader to be complied and bundled up.

So far message compilation has been done on the simple `KEYVALUE` format. ~In order to check for `STRUCTURED_JSON` format compilation, we change the message structure on `starter-typescript` template. Similar to normal `starter` template, translation workflow is simulated just before the build, in order to cover 'message-loader' compiling and bundling messages in `STRUCTURED_JSON` format.~

~This approach should have the added benefit of demonstrating app-kit's capabilities to work with different message formats, however if you believe this could confuse the users, we can adjust the mock translation script to do the conversion just before the build before the mock translation workflow.~

In order to check for `STRUCTURED_JSON` format compilation, we format and override the core.json file for the 'starter-typescript' ~application~ custom-view template in the mock translation script that runs just before the build. Afterwards, core.json is copied to en.json, which later on gets picked up by the message loader during the build to get compiled.

### Next Steps

With sufficient automated coverage through changes above, we hope to proceed to convert 'opt-in' behaviour to 'opt-out', renaming 'ENABLE_I18N_MESSAGE_COMPILATION' to 'DISABLE_I18N_MESSAGE_COMPILATION' and make necessary adjustments.


scripts/mock-translate-i18n-messages.mjs
```diff
@@ -0,0 +1,31 @@
...
+  ['cp', './src/i18n/data/core.json', './src/i18n/data/en.json'].join(' '),
+  { cwd: targetPackage.dir }
+);
+if (copyCoreJsonResult.code > 0) {
+  console.error(copyCoreJsonResult.stderr || copyCoreJsonResult.stdout);
+  throw new Error(
+    'Mock translation error: Copying core.json to en.json has failed.'
+  );
+}
+console.log(
```
mario-priceless, line 29: is this log needed ?

mario-priceless (approved):

scripts/mock-translate-i18n-messages.mjs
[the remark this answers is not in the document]
dogayuksel: I think having some visibility on the CI logs helps the observer to keep track of what is going on.

CarlosCortizasCT (commented):
Thanks a lot for working in this improvement @dogayuksel 🙇 

From my point of view, I would like for us to not change anything in the starter templates as it can introduce a little more complexity that might confuse developers coming to Custom Applications development.
I believe keeping the simple format in the starter templates should be considered as we don't expect developers needing the structured format from the get go (didn't have such a request so far).

What we can maybe do is to reformat an starter messages on the go as part of the test template CI action.
I can imagine a parameter we can use at that action level which would trigger the refactor of the message file only in the context of the test.

Something I'm also wondering is how we are covering the new functionality in the integration tests. Is the assumption that if the starter applications render the same text, the loader is working fine?

Thanks!

dogayuksel:
> What we can maybe do is to reformat an starter messages on the go as part of the test template CI action. I can imagine a parameter we can use at that action level which would trigger the refactor of the message file only in the context of the test. 

@CarlosCortizasCT Then I will add another branch to `scripts/mock-translate-i18n-messages.mjs` to override the `core.json` as `STRUCTURED_JSON` for when template is 'typescript-starter' before copying it to `en.json`. This way we build with `STRUCTURED_JSON` without modifying the template.

> Something I'm also wondering is how we are covering the new functionality in the integration tests. Is the assumption that if the starter applications render the same text, the loader is working fine?

Yes, given that `en.json` exists during build, it will be bundled up and will be using during e2e tests.
Here is the sanity check on this commit with the failing e2e tests, indeed 'translated' keys are used during e2e. https://github.com/commercetools/merchant-center-application-kit/pull/3528/commits/7b7f9c745e59365e28b451a9e243e5018dbf6170

application-templates/starter/.env
```diff
@@ -1,2 +1,3 @@
 ENABLE_NEW_JSX_TRANSFORM="true"
-FAST_REFRESH="true"
\ No newline at end of file
+ENABLE_I18N_MESSAGE_COMPILATION="true"
```
CarlosCortizasCT, line 2: We currently have 4 templates:
* Custom Applications - Javascript
* Custom Applications - Typescript
* Custom Views - Javascript
* Custom Views - Typescript

Perhaps it would be better to have the new compilation enabled for one template of Custom Applications and another one for Custom Views so we keep at least the other two with the default behaviour and we cover both cases in both extensions types.

What do you think?

dogayuksel: Wanted to go into the opt-out direction. I don't think there is a good reason to keep the messages in extracted object shape in the production bundle.
But there is no reason not to mix and match, I can move the `structured_json` case from custom app ts to custom view ts.

dogayuksel: @CarlosCortizasCT now we mix and match

CarlosCortizasCT (approved):
Thanks 💯

application-templates/starter/.env
dogayuksel: Maybe double negative got lost in the translation, I agree that having coverage on the custom views is a good idea. Hesitated to touch them, as a bit unfamiliar at this point 😅 

Yes this could be frustrating, will add the suggestion to use `DISABLE_I18N_MESSAGE_COMPILATION` on the error message of the compiler when doing the opt-out transition.

dogayuksel:
@CarlosCortizasCT Do we think we need a changset here too?

CarlosCortizasCT:
> @CarlosCortizasCT Do we think we need a changset here too?

I don't think we need it for this PR.

scripts/mock-translate-i18n-messages.mjs
```diff
@@ -0,0 +1,83 @@
...
+  const coreJson = JSON.parse(coreJsonRaw);
+
+  const coreJsonInStructuredJsonFormat = Object.keys(coreJson).reduce(
+    (allMessages, messageId) => ({
+      ...allMessages,
+      // return a simple STRUCTURED_JSON object
+      [messageId]: { string: coreJson[messageId] },
+    }),
+    {}
+  );
```
emmenko, line 57: Is it necessary to convert it to structured JSON or can we keep it as key/value?

scripts/mock-translate-i18n-messages.mjs
```diff
@@ -0,0 +1,83 @@
...
+    `Mock translations ==> Converted i18n/data/core.json to STRUCTURED_JSON format.`
+  );
+}
+
+// copy core.json to en.json
+// performed on custom-application/starter and custom-view/starter-typescript templates
+const copyCoreJsonResult = shelljs.exec(
+  ['cp', './src/i18n/data/core.json', './src/i18n/data/en.json'].join(' '),
+  { cwd: targetPackage.dir }
+);
```
emmenko, line 74: Nit: any particular reason not to use the Node.js API, for example `copyFile`?

application-templates/starter/.env
emmenko, line 2: If it's only for CI, maybe we can set the environment variable in the CI step instead of the dotenv file?

emmenko (commented):
Thanks @dogayuksel for improving this and having it properly tested ❤️

scripts/mock-translate-i18n-messages.mjs
dogayuksel: Not strictly necessary, I was including the conversion to structured JSON for a better coverage of possible cases. We also check for key/value.
 
| template | check |
| --- | --- |
| Custom Applications - Javascript | key/value |
| Custom Applications - Typescript | -- |
| Custom Views - Javascript | -- |
| Custom Views - Typescript | structured_json |

scripts/mock-translate-i18n-messages.mjs
dogayuksel: No good reason, thank you for the suggestion, converted it to `fs.copyFileSync`

application-templates/starter/.env
dogayuksel: I moved setting to the build step of CI.

I didn't write the expression to set it true when `app-type` and `template-type` matches 'x', 'y', as that might be unnecessarily complex. 

This essentially has the same effect, loader does not engage when translated files are not there.

emmenko (approved):
Nice, thanks! 💯

PR #3532, opened 2024-05-17
fix(loggers): catch errors when parsing loggers metadata

emmenko:


tdeekens (approved):

kark (approved):

Rhotimee (approved):
Makes sense. Thanks

packages-backend/loggers/src/middlewares/create-access-logger.ts
```diff
@@ -39,15 +39,20 @@ const createAccessLoggerMiddleware = (
...
-      const proxyIps = parseIps(req);
-      const [clientIp] = proxyIps;
-      return {
-        clientIp: clientIp ?? remoteAddress,
-        proxyIps,
-        hostname: req.hostname,
-        ...(remoteAddress ? { remoteAddress } : {}),
-      };
+      try {
+        const remoteAddress = req.socket?.remoteAddress;
```
namm2, line 43: Can we check the `req.socket` is not null first? I think the issue we got it because the socket isn't/wasn't success then there will be no remoteAddress as the result.

emmenko: Yes the `?.` does the null check

packages-backend/loggers/src/middlewares/create-access-logger.ts
```diff
@@ -39,15 +39,20 @@ const createAccessLoggerMiddleware = (
...
+        const proxyIps = parseIps(req);
+        const [clientIp] = proxyIps;
+        return {
+          clientIp: clientIp ?? remoteAddress,
+          proxyIps,
+          hostname: req.socket ? req.hostname : undefined,
```
emmenko, line 49: We also added here an extra check for `req.socket` before accessing `req.hostname`

PR #3536, opened 2024-05-23
Using Safe triangle to improve submenu navigation

ddouglasz:
#### Summary

Using the Safe triangle  pattern to improve submenu navigation

#### Description

Improve the submenu navigation by adding a safe area for users to smoothly navigate diagonally between the main menu and the submenu items, without awkwardly going out of focus.

Background: 
https://x.com/shadeed9/status/1745707615517032449?t=XnNt9gNwPHLyfrMfHEnFuA&s=03

Before: 

After (The visible triangle is only for the demo. It is not included in the PR):

### DIFF
--- a/packages/application-shell/src/components/navbar/menu-items.styles.ts
+++ b/packages/application-shell/src/components/navbar/menu-items.styles.ts
@@ -102,6 +102,7 @@ const listStyles = css`
   flex-direction: column;
   min-height: 0;
   flex: 1 1 0;
+  --safe-start: 0% 0%;
 `;
 
 const sublistStyles = css`
@@ -311,6 +312,15 @@ const MenuListItem = styled.li<{
   }
 `;
 
+const SafeAreaElement = styled.span`
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  right: 100%;
+  width: calc(100% - 16px);
+  clip-path: polygon(var(--safe-start), 100% 100%, 100% 0);
+`;
+
 export {
   Expander,
   ExpanderIcon,
@@ -320,4 +330,5 @@ export {
   SublistItem,
   TextLinkSublistWrapper,
   NavlinkClickableContent,
+  SafeAreaElement,
 };
--- a/packages/application-shell/src/components/navbar/menu-items.styles.ts
+++ b/packages/application-shell/src/components/navbar/menu-items.styles.ts
@@ -325,7 +325,10 @@ const SafeArea = styled.span`
   top: 0;
   bottom: 0;
   right: 100%;
-  width: calc(100% - 16px);
+  /** Ensure the full width of the safe triangle is 100% of the scrollable menu area, less of its left padding (16px)
+   * which is also the starting point of the safe triangle/menu item.
+   */
+  width: calc(100% - ${uiKitDesignTokens.spacing30});
   clip-path: polygon(var(--safe-start), 100% 100%, 100% 0);
 `;
 
--- a/packages/application-shell/src/components/navbar/menu-items.styles.ts
+++ b/packages/application-shell/src/components/navbar/menu-items.styles.ts
@@ -313,8 +313,6 @@ const MenuListItem = styled.li<{
 `;
 
 const SafeArea = styled.span`
-  background-color: gray;
-  opacity: 0.5;
   position: absolute;
   top: 0;
   bottom: 0;
--- a/packages/application-shell/src/components/navbar/menu-items.styles.ts
+++ b/packages/application-shell/src/components/navbar/menu-items.styles.ts
@@ -107,7 +107,6 @@ const listStyles = css`
   flex-direction: column;
   min-height: 0;
   flex: 1 1 0;
-  --safe-start: 100% 100%;
 `;
 
 const sublistStyles = css`
--- a/packages/application-shell/src/components/navbar/menu-items.styles.ts
+++ b/packages/application-shell/src/components/navbar/menu-items.styles.ts
@@ -34,7 +34,7 @@ const getContainerPositionBasedOnMenuItemPosition = (
     `,
 ];
 
-const bounce = keyframes`
+const fadeIn = keyframes`
 from {opacity: 0;}
   to { opacity: 1;}
 `;
@@ -286,7 +286,7 @@ const MenuListItem = styled.li<{
     ${MenuList}.sublist-collapsed__active__above,
     :focus-within
     ${MenuList}.sublist-expanded__active {
-    animation-name: ${bounce};
+    animation-name: ${fadeIn};
     animation-duration: 16ms;
     animation-delay: 100ms;
     animation-fill-mode: forwards;
--- a/packages/application-shell/src/components/navbar/menu-items.styles.ts
+++ b/packages/application-shell/src/components/navbar/menu-items.styles.ts
@@ -102,7 +102,7 @@ const listStyles = css`
   flex-direction: column;
   min-height: 0;
   flex: 1 1 0;
-  --safe-start: 0% 0%;
+  --safe-start: 100% 100%;
 `;
 
 const sublistStyles = css`
@@ -312,7 +312,9 @@ const MenuListItem = styled.li<{
   }
 `;
 
-const SafeAreaElement = styled.span`
+const SafeArea = styled.span`
+  background-color: gray;
+  opacity: 0.5;
   position: absolute;
   top: 0;
   bottom: 0;
@@ -330,5 +332,5 @@ export {
   SublistItem,
   TextLinkSublistWrapper,
   NavlinkClickableContent,
-  SafeAreaElement,
+  SafeArea,
 };
--- a/packages/application-shell/src/components/navbar/menu-items.styles.ts
+++ b/packages/application-shell/src/components/navbar/menu-items.styles.ts
@@ -1,4 +1,4 @@
-import { css } from '@emotion/react';
+import { css, keyframes } from '@emotion/react';
 import styled from '@emotion/styled';
 import { designTokens as appKitDesignTokens } from '@commercetools-frontend/application-components';
 import { designTokens as uiKitDesignTokens } from '@commercetools-uikit/design-system';
@@ -33,6 +33,11 @@ const getContainerPositionBasedOnMenuItemPosition = (
       top: 0;
     `,
 ];
+
+const bounce = keyframes`
+from {opacity: 0;}
+  to { opacity: 1;}
+`;
 
 const Expander = styled.li<{ isVisible: boolean }>`
   display: flex;
@@ -141,7 +146,7 @@ const MenuList = styled.ul<
       props.isSublistCollapsedAndActive ||
       props.isSublistCollapsedAndActiveAndAbove) &&
       css`
-        opacity: 1;
+        opacity: 0;
         display: none;
         text-align: left;
         background-color: ${uiKitDesignTokens.colorAccent20};
@@ -281,6 +286,10 @@ const MenuListItem = styled.li<{
     ${MenuList}.sublist-collapsed__active__above,
     :focus-within
     ${MenuList}.sublist-expanded__active {
+    animation-name: ${bounce};
+    animation-duration: 16ms;
+    animation-delay: 100ms;
+    animation-fill-mode: forwards;
     display: flex;
     flex-direction: column;
     align-items: flex-start;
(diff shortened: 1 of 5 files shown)


kark:
Thank you @ddouglasz for working on this ❤️
I've briefly checked it in the playground and for the most part it works fine, when the cursor move is consistent.
However, if I stop in the middle / change the pace of the movement, then the submenu hides.

I've noticed in the screencast from the description that the safe triangle only changes when cursor moves within the menu item boundaries. If it goes beyond the menu item, the triangle is set for good.
Screenshot 2024-05-23 at 14 41 01
In case of the implementation the triangle shrinks whenever cursor moves towards the submenu, so maybe this makes the behaviour not consistent

ddouglasz:
Thank you Kacper 🙏🏾
I will look into it and push a fix.

ddouglasz:
> I've noticed in the screencast from the description that the safe triangle only changes when cursor moves within the menu item boundaries. If it goes beyond the menu item, the triangle is set for good.

@kark thanks for your feedback again, after a couple of findings, I discovered a flaw with this implementation:
When a user begins to navigate diagonally and then stops halfway to move to the next item, vertically, they are blocked because the triangle is set for good until we close the submenu. 


Perhaps the current implementation that allows the the `safe-start` point of the triangle to move is what I see is consistent with most implementations:
If the users delays a in the diagonal navigation, then it closes the submenu and focuses on the next item the cursor is hovering on. 


Would be nice to know what @FilPob also thinks about this 🙏🏾

kark:
> > I've noticed in the screencast from the description that the safe triangle only changes when cursor moves within the menu item boundaries. If it goes beyond the menu item, the triangle is set for good.
> 
> @kark thanks for your feedback again, after a couple of findings, I discovered a flaw with this implementation: When a user begins to navigate diagonally and then stops halfway to move to the next item, vertically, they are blocked because the triangle is set for good until we close the submenu. 

Thanks! Perhaps we should consider adding a timeout for the triangle, similar to what we can see in your MacOS screencast. 🤔

ddouglasz:
> > > I've noticed in the screencast from the description that the safe triangle only changes when cursor moves within the menu item boundaries. If it goes beyond the menu item, the triangle is set for good.
> > 
> > 
> > @kark thanks for your feedback again, after a couple of findings, I discovered a flaw with this implementation: When a user begins to navigate diagonally and then stops halfway to move to the next item, vertically, they are blocked because the triangle is set for good until we close the submenu.
> 
> Thanks! Perhaps we should consider adding a timeout for the triangle, similar to what we can see in your MacOS screencast. 🤔

Updated here: https://github.com/commercetools/merchant-center-application-kit/pull/3536/commits/25c5b476da0f27e4acf695003b7eb8d09869fbb3

Thank you.

FilPob (approved):

packages/application-shell/src/components/navbar/menu-items.tsx
```diff
@@ -270,6 +271,7 @@ const MenuItem = (props: MenuItemProps) => {
       isActive={props.isActive}
       isRouteActive={Boolean(props.isMainMenuRouteActive)}
       isCollapsed={!props.isMenuOpen}
+      onMouseMove={props.onMouseMove as MouseEventHandler<HTMLLIElement>}
```
misama-ct, line 274: Is the casting really needed? I tried removing it locally and I am not getting any squiggly lines.

misama-ct (approved):
Thanks for being persistent and making it work 👏 .
I have only some suggestions, nothing that would block approval.

packages/application-shell/src/components/navbar/menu-items.tsx
ddouglasz: updated here: https://github.com/commercetools/merchant-center-application-kit/pull/3536/commits/4055a9cbc064d49ac4c90ec42d85b55cf0b284b4
Thank you.

packages/application-shell/src/components/navbar/menu-items.styles.ts
```diff
@@ -280,6 +286,10 @@ const MenuListItem = styled.li<{
     ${MenuList}.sublist-collapsed__active__above,
     :focus-within
     ${MenuList}.sublist-expanded__active {
+    animation-name: ${bounce};
```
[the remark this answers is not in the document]
ddouglasz: updated here: https://github.com/commercetools/merchant-center-application-kit/pull/3536/commits/4055a9cbc064d49ac4c90ec42d85b55cf0b284b4
Thank you.

packages/application-shell/src/components/navbar/menu-items.styles.ts
```diff
@@ -102,6 +107,7 @@ const listStyles = css`
   flex-direction: column;
   min-height: 0;
   flex: 1 1 0;
+  --safe-start: 100% 100%;
```
kark, line 110: I wonder if we need to define it here instead of relying on the initial state of `percentageX` and `percentageY`? And if that initial value is needed at all?

packages/application-shell/src/components/navbar/menu-items.styles.ts
```diff
@@ -311,6 +321,15 @@ const MenuListItem = styled.li<{
   }
 `;
 
+const SafeArea = styled.span`
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  right: 100%;
+  width: calc(100% - 16px);
```
kark, line 329: I wonder, where is this `16px` subtraction coming from?

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
```diff
@@ -34,22 +34,29 @@ type HookProps = {
   environment: TApplicationContext<{}>['environment'];
   project: TFetchProjectQuery['project'];
 };
+type TMousePosition = {
+  clientX: number;
+  clientY: number;
+};
```
kark, line 40: nit: I think we define this type twice in this PR

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
```diff
@@ -34,22 +34,29 @@ type HookProps = {
...
   isMenuOpen: boolean;
+  mousePosition: TMousePosition;
 };
 type Action =
   | { type: 'setActiveItemIndex'; payload: string }
   | { type: 'unsetActiveItemIndex' }
   | { type: 'setIsExpanderVisible' }
   | { type: 'toggleIsMenuOpen' }
   | { type: 'setIsMenuOpenAndMakeExpanderVisible'; payload: boolean }
+  | { type: 'setMousePosition'; payload: TMousePosition }
```
kark, line 53: nit: I suppose we read the mouse position, so maybe `getMousePosition` would be more suitable

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -111,8 +119,59 @@ export const ApplicationMenu = (props: ApplicationMenuProps) => {
...
+    safeAreaHeight,
+    safeAreaLeftPos,
+    safeAreaTopPos,
+    safeAreaWidth,
+  ]);
+
+  useLayoutEffect(() => {
+    submenuRef.current?.style.setProperty(
+      '--safe-start',
+      `${percentageX}% ${percentageY + 1}%`
```
kark, line 172: Can we also explain why adding 1 is needed?

kark (commented):
Thank you, Douglas! That's a great improvement 👏. I have to admit, I'm still trying to nail down where the mouse movement speed threshold is so that the safe triangle is applied, but overall, this looks great!

packages/application-shell/src/components/navbar/menu-items.styles.ts
ddouglasz: This 16 px will always be the padding-left of the navbar area, also the start of the navbar item, whether expanded or collapsed. That is where we want our safe-area to end. 
Screenshot 2024-08-26 at 12 45 15

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -111,8 +119,59 @@ export const ApplicationMenu = (props: ApplicationMenuProps) => {
...
+  const [percentageX, setPercentageX] = useState(0);
+  const [percentageY, setPercentageY] = useState(0);
   const observerRef = useRef<IntersectionObserver | null>(null);
   const submenuRef = useRef<HTMLUListElement>(null);
+  const submenuSafeAreaRef = useRef<HTMLElement>(null);
+
+  /* Getting the width and height of the menu*/
+  const subRefBoundingClientRect = submenuRef.current?.getBoundingClientRect();
+  const { width: menuItemWidth, height: menuItemHeight } =
+    subRefBoundingClientRect! ?? {};
```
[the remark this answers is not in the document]
ddouglasz: Thank you for helping out with this review again 🙏🏾
Updated here: https://github.com/commercetools/merchant-center-application-kit/pull/3536/commits/b662e8f9258af8523279fa2a301a35959e6a6b33

packages/application-shell/src/components/navbar/menu-items.styles.ts
ddouglasz: updated here: https://github.com/commercetools/merchant-center-application-kit/pull/3536/commits/b662e8f9258af8523279fa2a301a35959e6a6b33

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
ddouglasz: Updated here: https://github.com/commercetools/merchant-center-application-kit/pull/3536/commits/b662e8f9258af8523279fa2a301a35959e6a6b33

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
ddouglasz: Updated here: https://github.com/commercetools/merchant-center-application-kit/pull/3536/commits/b662e8f9258af8523279fa2a301a35959e6a6b33

packages/application-shell/src/components/navbar/navbar.tsx
ddouglasz: Updated here: https://github.com/commercetools/merchant-center-application-kit/pull/3536/commits/b662e8f9258af8523279fa2a301a35959e6a6b33

I left a comment just to explain it. We want the cursor so be slightly inside the safe area while we move it so we add an extra digit to keep it just inside the triangle at it's apex. Otherwise, it would be exactly at the end of the triangle and that would not shield the cursor from triggering an event we want to prevent.

packages/application-shell/src/components/navbar/menu-items.styles.ts
kark: Thanks got it 👍 Could we also add a comment here explaining that we account for the scrollable menu's padding and use the `spacing30` design token?

packages/application-shell/src/components/navbar/use-navbar-state-manager.ts
```diff
@@ -256,6 +266,18 @@ const useNavbarStateManager = (props: HookProps) => {
     [state.activeItemIndex]
   );
 
+  const handleMouseMove = useCallback(
```
kark, line 269: nit: just noticed that we have two functions named `handleMouseMove` that do different things. Could we rename this one to be more descriptive?

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -111,8 +116,60 @@ export const ApplicationMenu = (props: ApplicationMenuProps) => {
...
+
+  // /* We want to track the left, top, width, and height of the safe area */
+  const submenuSafeAreaRefBoundingClientRect =
+    submenuSafeAreaRef.current?.getBoundingClientRect();
+  const safeAreaLeftPos = submenuSafeAreaRefBoundingClientRect?.left || 0;
+  const safeAreaTopPos = submenuSafeAreaRefBoundingClientRect?.top || 0;
+  const safeAreaWidth = submenuSafeAreaRefBoundingClientRect?.width || 0;
+  const safeAreaHeight = submenuSafeAreaRefBoundingClientRect?.height || 0;
+
+  const handleMouseMove = useCallback(
```
kark, line 138: Actually, let's rename this one too. To e.g. `calculateSafeAreaStartPositon`

kark (approved):
🙌

PR #3541, opened 2024-05-27
New cli command to manage deployment previews configuration

CarlosCortizasCT:
#### Summary

Start work towards improving the Custom Applications deployment previews feature by adding a new command which will allow consumers to create/update deployments previews from the command line using our existing `mc-scripts` cli.

#### Description

There are internal teams with some challenges that could greatly benefit of having a better non-human interaction integration with Custom Applications deployment previews.
We do currently support this feature by manually adjusting some configuration in the Merchant Center UI, but it would be great if that configuration could be automatically updated from CI when a new feature-branch in created in a repository.

In order to achieve that we need to tackle different things but, as part of our X&X days, I wanted to start with this one which would allow for a logged-in user to create or update a new Custom Application deployment preview using the CLI:
```bash
# mc-scripts deployment-previews:set --alias my-alias --url https://www.nowhere.com
```
_(if the arguments are not provided, the command will prompt for them)_

This could allow consumers to automatically trigger from their CI the creation/update of a Custom Application deployment preview for any given created branch without human intervention.

This steps still required a CLI logged-in user to run this command, but it's a first step towards the end goal.


packages/mc-scripts/src/commands/deployments-push.ts
```diff
@@ -0,0 +1,253 @@
...
+  fetchCustomApplication,
+  updateCustomApplicationDeploymentPreview,
+  createCustomApplicationDeploymentPreview,
+} from '../utils/graphql-requests';
+
+const credentialsStorage = new CredentialsStorage();
+
+const deploymentPreviewAliasRegex = /^[^-#]([a-z]|[-](?![-])){0,62}[^-#]$/g;
+export const fqdnRegex =
+  /(?=^.{4,253}$)(^((?!-)[a-zA-Z0-9-]{0,62}[a-zA-Z0-9]\.)+[a-zA-Z]{2,63}$)/i;
```
CarlosCortizasCT, line 20: These ones were copied from the MC frontend.
I would suggest to move them to the constants package and use them from there both here and in the frontend.

On the other hand, I think we should adjust the RegExps or even think about what we want to validate as current ones deny some values I find legit:
```
// We're currently blocking aliases like these:
pr-1234
branch-1234
team-1234
v2-branch

// We're currently blocking URLs like these:
https://pr_1234.surge.sh

File: packages/mc-scripts/src/commands/deployment-previews-push.ts [hunk 2]
```diff
@@ -0,0 +1,266 @@
...
+  fetchCustomApplication,
+  updateCustomApplicationDeploymentPreview,
+  createCustomApplicationDeploymentPreview,
+} from '../utils/graphql-requests';
+
+const credentialsStorage = new CredentialsStorage();
+
+const deploymentPreviewAliasRegex = /^[^-#]([a-z]|[-](?![-])){0,62}[^-#]$/g;
+export const fqdnRegex =
+  /(?=^.{4,253}$)(^((?!-)[a-zA-Z0-9-]{0,62}[a-zA-Z0-9]\.)+[a-zA-Z]{2,63}$)/i;
```
emmenko, line 20: Wondering if we should move this validation in the API. In the MC we only check this client side. Just to avoid duplication.

Can be a follow up of course.

packages/mc-scripts/src/commands/deployment-previews-push.ts
```diff
@@ -0,0 +1,266 @@
...
+    throw new Error(
+      `You don't have a valid session. Please, run the "mc-scripts login" command to authenticate yourself.`
+    );
+  }
+
+  if (isCustomView) {
+    throw new Error(
+      'Deployments previews are not supported for Custom Views yet.'
+    );
+  }
```
emmenko, line 254: Do you envision deployment previews for Custom Views?

Maybe we can rephrase this just to avoid false expectations.

```suggestion
  if (isCustomView) {
    throw new Error(
      'Deployments previews are not supported for Custom Views.'
    );
  }
```

packages/mc-scripts/src/commands/deployment-previews-push.ts
```diff
@@ -0,0 +1,266 @@
...
+
+  const fetchedCustomApplication = await fetchCustomApplication({
+    mcApiUrl,
+    entryPointUriPath: localCustomEntityData.entryPointUriPath,
+    applicationIdentifier,
+  });
+
+  if (!fetchedCustomApplication) {
+    throw new Error(
+      `The Custom Application "${applicationIdentifier}" does not exist in the Merchant Center.`
```
emmenko, line 95:
```suggestion
      `The Custom Application "${applicationIdentifier}" does not exist in the Merchant Center. Please create a Custom Application first before configuring deployment previews.`
```

packages/mc-scripts/src/commands/deployment-previews-push.ts
```diff
@@ -0,0 +1,266 @@
...
+      message: [
+        `What's the alias for the deployment preview? (my-alias)`,
+        options.dryRun &&
+          chalk.gray(
+            'Using "--dry-run", no deployment preview will be created.'
+          ),
+      ]
+        .filter(Boolean)
+        .join('\n'),
+      validate: (value) => value && value.length > 2,
```
emmenko, line 114: Nit: should we suggest a default alias like `staging` or something?
I would also rephrase a bit the sentence:

```suggestion
    const { alias } = await prompts({
      type: 'text',
      name: 'alias',
      initial: 'my-alias',
      message: [
        `Enter the alias for the deployment preview`,
        options.dryRun &&
          chalk.gray(
            'Using "--dry-run", no deployment preview will be created.'
          ),
      ]
        .filter(Boolean)
        .join('\n'),
      validate: (value) => value && value.length > 2,
```

packages/mc-scripts/src/commands/deployment-previews-push.ts
```diff
@@ -0,0 +1,266 @@
...
+    });
+    deploymentAlias = alias as string;
+  }
+  if (!validateAlias(deploymentAlias)) {
```
emmenko, line 118: Should this be part of the `validate` function?

packages/mc-scripts/src/commands/deployment-previews-push.ts
```diff
@@ -0,0 +1,266 @@
...
+          deploymentAlias
+        )}" already exists. Do you want to proceed?`,
+        '(The URL you will provide will override the existing one.)',
+        options.dryRun &&
+          chalk.gray(
+            'Using "--dry-run", no deployment preview will be created.'
+          ),
+      ]
+        .filter(Boolean)
+        .join('\n'),
```
emmenko, line 144:
```suggestion
      message: [
        `The alias "${chalk.green(
          deploymentAlias
        )}" already exists. Do you want to proceed and update the URL?`,
        options.dryRun &&
          chalk.gray(
            'Using "--dry-run", no deployment preview will be created.'
          ),
      ]
        .filter(Boolean)
        .join('\n'),
```

packages/mc-scripts/src/commands/deployment-previews-push.ts
```diff
@@ -0,0 +1,266 @@
...
+    }
+  }
+
+  let deploymentUrl = options.url;
+  if (!deploymentUrl) {
+    const { url } = await prompts({
+      type: 'text',
+      name: 'url',
+      message: [
+        `What's the URL for the deployment preview?`,
```
emmenko, line 160:
```suggestion
        `Enter the URL for the deployment preview`,
```

packages/mc-scripts/src/cli.ts
```diff
@@ -197,6 +198,43 @@ async function run() {
       }
     );
 
+  // Command: deployment-previews:push
+  const usageDeploymentPreviewsPush =
+    'Creates or updates a deployment preview for the customization application.';
+  cli
+    .command('deployment-previews:push', usageDeploymentPreviewsPush)
```
emmenko, line 205: Thinking about the naming, I find `push` slightly confusing considering what it does.

I suggest something like `set` (since we're doing an upsert). Vercel for instance [uses that](https://vercel.com/docs/cli/alias#usage) as well `vercel alias set`.

```suggestion
    .command('deployment-previews:set', usageDeploymentPreviewsPush)
```

This can then scale with other commands:

```
deployment-previews:set
deployment-previews:remove
deployment-previews:list
```

emmenko (commented):
Great work, thanks a lot for picking this up and taking the first steps! ❤️

Looks good overall, I left some minor suggestions / comments.

packages/mc-scripts/src/cli.ts
```diff
@@ -197,6 +198,43 @@ async function run() {
       }
     );
 
+  // Command: deployment-previews:push
+  const usageDeploymentPreviewsPush =
+    'Creates or updates a deployment preview for the customization application.';
```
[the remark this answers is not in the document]
CarlosCortizasCT: I was intentionally general because I think we need to to support deployment previews in Custom Views as well.

That's something we had in that's feature plan but left out of the first iteration.

packages/mc-scripts/src/commands/deployment-previews-push.ts
CarlosCortizasCT: Could you check [this comment](https://github.com/commercetools/merchant-center-application-kit/pull/3541#discussion_r1616261551)?

It's about how to share the validations but also how to adjust them.

packages/mc-scripts/src/commands/deployment-previews-push.ts
CarlosCortizasCT: Yes I do.

That's something we considered as part of the Custom Views features but left out of the first iteration to keep the scope contained.

packages/mc-scripts/src/cli.ts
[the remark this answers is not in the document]
emmenko: Let's have a separate discussion about this as I'm not sure it's something we can easily support.

packages/mc-scripts/src/commands/deployment-previews-push.ts
emmenko: Yes I saw that. My point was mostly about having this validation additionally in the MC API.

packages/mc-scripts/src/cli.ts
CarlosCortizasCT: I personally prefer `push` as a more meaningful word (maybe only from my point of view) since that name does not imply how internally the logic will work but if you prefer a different naming, I'll just adjust things accordingly.

packages/mc-scripts/src/cli.ts
```diff
@@ -197,6 +198,43 @@ async function run() {
...
+    .command('deployment-previews:push', usageDeploymentPreviewsPush)
+    .usage(`\n\n  ${usageDeploymentPreviewsPush}`)
+    .option(
+      '--alias <deployment-preview-alias>',
+      "(optional) Alias to be used for the deployment preview. If you don't provide an alias, the script will prompt you for it."
+    )
+    .option(
+      '--url <deployment-preview-url>',
+      "(optional) URL to be used for the deployment preview. If you don't provide a URL, the script will prompt you for it."
+    )
```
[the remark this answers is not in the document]
CarlosCortizasCT: Updated here: 54c523145d410c538431c6c27eca10866e794a0b

packages/mc-scripts/src/commands/deployment-previews-push.ts
CarlosCortizasCT: Updated here: 54c523145d410c538431c6c27eca10866e794a0b

packages/mc-scripts/src/commands/deployment-previews-push.ts
CarlosCortizasCT: I updated the message here: 54c523145d410c538431c6c27eca10866e794a0b

Regarding the default value, I don't see the point as any consumer could use something different.

packages/mc-scripts/src/commands/deployment-previews-push.ts
CarlosCortizasCT: Updated here: 54c523145d410c538431c6c27eca10866e794a0b

packages/mc-scripts/src/commands/deployment-previews-push.ts
CarlosCortizasCT: Updated here: 54c523145d410c538431c6c27eca10866e794a0b

packages/mc-scripts/src/commands/deployment-previews-push.ts
CarlosCortizasCT: I agree we're missing the validation in the backend. We will need to create a follow-up task for that.

packages/mc-scripts/src/commands/deployment-previews-push.ts
CarlosCortizasCT: You're totally right.

Updated here: 60596062f40f0958f064165661c9e003363bed56

packages/mc-scripts/src/cli.ts
emmenko: I don't have a strong opinion, it was just my feeling as a consumer of the CLI.

Let's wait for the feedback from the others.

tdeekens: I can only share my mental model for `push` and `set`. I would prefer `set` if you `set` (update or create) a value singular entity. I would use `push` (or likely more `add`) if we add to a list of things. I would then exact that the addition to that bag of things as an effect or later I have to `mark` or `set` the added item to e.g. be the primary of something.

CarlosCortizasCT: Thanks for your feedback, @tdeekens 

I guess that in this context you would prefer `add` as the name of the command?

Bear in mind what we're actually doing is adding a new deployment preview link to the list of deployment preview links of a Custom Application. If the alias already exist, we update the current preview link configuration url.

tdeekens: Yes, I'd have chosen `add` then. Still, no strong opinions here. It's a suggestion, nothing more.

CarlosCortizasCT: Thanks!
We have several suggestions but we need to come to an agreement.

So far we have `push`, `set` and `add`.

How we should decide on what to use?

emmenko: @kark @Rhotimee Please can you also share your feedback here?

kark: Thank you for pinging me. First and foremost, this is a great improvement 💪 many thanks for the efforts! 🙇‍♂️

I believe that `push`, `add`, and `set` all work well in this context. However, I have a slight preference for `set`. To me, it clearly suggests intent of either creation or updating of a deployment preview config.

emmenko: Suggestion for a slight different option maybe: have "CRUD-like" commands like `create`, `update`, `remove`, etc.
Then it becomes clear and more explicit what the command does. This means of course that there won't be any upsert behavior.

Rhotimee: I also prefer `set` because it indicates an upsert operation.

CarlosCortizasCT: Ok, thank you all for the feedback.

We are two who prefer `add` but three who prefer `set` so I'll be changing the naming of the command.

packages/mc-scripts/src/commands/deployments-push.ts
CarlosCortizasCT: Simplified validations according to the agreement with @emmenko: e624165c7d4d0aaa07e2a0e0c8004d380171d750

packages/mc-scripts/src/cli.ts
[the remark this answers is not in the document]
CarlosCortizasCT: Message updated here: e624165c7d4d0aaa07e2a0e0c8004d380171d750

packages/mc-scripts/src/commands/deployment-previews-push.ts
CarlosCortizasCT: Message updated in this commit: aa8ac4684f43bce4c5332169e38bd039964e6212

packages/mc-scripts/src/cli.ts
CarlosCortizasCT: Command renamed in this commit: aa8ac4684f43bce4c5332169e38bd039964e6212

packages/mc-scripts/src/cli.ts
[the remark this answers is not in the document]
emmenko: This one hasn't been updated?

emmenko (commented):
Thanks, looks good to go from side 🙌

Please also provide a changeset 🙏

kark (approved):
Thank you! 🙇‍♂️ 
I pulled the changes and tested them locally. Both creating and updating a deployment preview using `deployment-previews:set` work perfectly.

packages/mc-scripts/src/cli.ts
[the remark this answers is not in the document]
CarlosCortizasCT: Sorry about it.

Updated here: adf3d479abcce1134a8b693d89ba1602c1c3ecaf

PR #3542, opened 2024-05-28
Add configuration for Playground application integration in Vercel

CarlosCortizasCT:
#### Summary

Add configuration for Playground application integration in Vercel

#### Description

We want to give it a try to configure deployments of the Playground Custom Application in Vercel and see if we can use them for having deployment previews in the Merchant Center.

In this PR we add the first basic configuration.


playground/vercel.json
emmenko, line 1: What do we need this for?

.github/workflows/playground-deployment-preview.yml
```diff
@@ -0,0 +1,23 @@
+name: Playground application deployment preview trigger
+
+on:
+  workflow_dispatch:
+  repository_dispatch:
+    types: [vercel_deployment]
```
emmenko, line 6: Just to understand: this workflow is expected to be triggered manually?

emmenko (commented):
Thanks 🚀

playground/vercel.json
CarlosCortizasCT: This is similar to what we have in the `website-components-playground` application ([reference](https://github.com/commercetools/merchant-center-application-kit/blob/main/website-components-playground/vercel.json)).

I first tried setting the configuration in Vercel's UI but it was not working so I checked how are we doing it with the other application and it started working in the playground once I provided this configuration file.

If you know a better way, please just let me know how I should do it 🙏

.github/workflows/playground-deployment-preview.yml
CarlosCortizasCT: Once the project is setup, we will set a webhook in Vercel to our repository which will be called whenever a deployment is done.

The idea is that we will use that webhook to trigger this gh action in order to create/update the deployment preview in the Merchant Center.

That's something we can't do now (because of the missing auth token) but I wanted to set it up and verify it works so it would be easier for us in the future to integrate it once the auth token is available.

image

emmenko: Thanks, this makes sense now.

playground/vercel.json
emmenko: Hmm yeah well the components playground is deployed automatically using GH integration.

The playground currently is not: https://github.com/commercetools/merchant-center-application-kit/blob/146cf672eb15a7b4d858c54d6a01f92d0437a86f/.github/workflows/release.yml#L89-L95

Should we maybe look into making the playground be deployed automatically? The setup might become easier then.

CarlosCortizasCT: I'm sorry but I don't understand. I already set up the playground to be deployed automatically in Vercel.

We just need this file to make it work correctly and the setup it's still easy from my point of view.

Do you think it would be better if we have a sync about this?

emmenko: Yes that might be better. Thanks 😅

.github/workflows/release.yml
```diff
@@ -85,11 +85,3 @@ jobs:
           MC_API_URL: ${{ secrets.MC_API_URL }}
           CTP_INITIAL_PROJECT_KEY: ${{ secrets.CYPRESS_PROJECT_KEY }}
           HOST_GCP_STAGING: ${{ secrets.HOST_GCP_STAGING }}
-
```
CarlosCortizasCT, line 88: Removing as the Vercel <-> GitHub integration will do the deployments for us.

emmenko (approved):
Thanks!

PR #3546, opened 2024-06-05
Fix Playground Custom Application deployment previews

CarlosCortizasCT:
#### Summary

Fix Playground Custom Application deployment previews

#### Description

We can't use a fixed value for the Custom Application configured URL but the one provided by Vercel for each deployment (`VERCEL_URL` env variable; [reference](https://vercel.com/docs/projects/environment-variables/system-environment-variables)).

### DIFF
--- a/playground/custom-application-config.mjs
+++ b/playground/custom-application-config.mjs
@@ -1,7 +1,9 @@
 import { PERMISSIONS, entryPointUriPath } from './src/constants';
 const name = 'AppKit Playground Application';
 
-const productionUrl = Boolean(process.env.VERCEL_URL) ? `https://${process.env.VERCEL_URL}` : process.env.APP_URL;
+const productionUrl = process.env.VERCEL_ENV !== 'production' && Boolean(process.env.VERCEL_URL) ?
+  `https://${process.env.VERCEL_URL}` :
+  process.env.APP_URL;
 
 /**
  * @type {import('@commercetools-frontend/application-config').ConfigOptionsForCustomApplication}
--- a/playground/custom-application-config.mjs
+++ b/playground/custom-application-config.mjs
@@ -1,6 +1,8 @@
 import { PERMISSIONS, entryPointUriPath } from './src/constants';
 const name = 'AppKit Playground Application';
 
+const productionUrl = Boolean(process.env.VERCEL_URL) ? `https://${process.env.VERCEL_URL}` : process.env.APP_URL;
+
 /**
  * @type {import('@commercetools-frontend/application-config').ConfigOptionsForCustomApplication}
  */
@@ -19,7 +21,7 @@ const config = {
     },
     production: {
       applicationId: '${env:APP_ID}',
-      url: '${env:APP_URL}',
+      url: productionUrl,
     },
   },
   additionalEnv: {
--- a/playground/.env
+++ b/playground/.env
@@ -1,5 +1,5 @@
 CLOUD_IDENTIFIER="gcp-eu"
-APP_URL="https://mc-app-kit-playgound.vercel.app"
+APP_URL="https://${VERCEL_URL}"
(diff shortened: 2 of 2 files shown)


playground/custom-application-config.mjs
```diff
@@ -1,6 +1,8 @@
 import { PERMISSIONS, entryPointUriPath } from './src/constants';
 const name = 'AppKit Playground Application';
 
+const productionUrl = Boolean(process.env.VERCEL_URL) ? `https://${process.env.VERCEL_URL}` : process.env.APP_URL;
```
emmenko, line 4: The main production URL has to be a static value, as this is what you configure in the Custom Application settings.

The dynamic URL is only required for preview deployments. I suggest to rely on the `VERCEL_BRANCH_URL` environment variable for that.

emmenko (commented):
Thanks @CarlosCortizasCT , there is an important catch though (see comment).

playground/custom-application-config.mjs
CarlosCortizasCT: I'm sorry but I think I don't understand.

These are the env variables available:

image

The way I understand those URLs, the `VERCEL_URL` one is per deployment but the `VERCEL_PROJECT_PRODUCTION_URL` is the static one:
* `VERCEL_URL`: `https://vercel.live/link/mc-app-kit-playground-4g7q8lt76-commercetools.vercel.app`
* `VERCEL_BRANCH_URL`: `https://vercel.live/link/mc-app-kit-playground-git-renovate-all-ui-151302-commercetools.vercel.app` 
* `VERCEL_PROJECT_PRODUCTION_URL`: `https://mc-app-kit-playgound.vercel.app/`

I don't see the real difference between `VERCEL_URL` and `VERCEL_BRANCH_URL` but they are both dynamic and point to a specific build.

The one we have configured in the Custom Application settings in the Merchant Center is the `VERCEL_PROJECT_PRODUCTION_URL` and that one will always be available and accessible.

Maybe what you mean is that, for the production deployment, we should not use `VERCEL_URL` nor `VERCEL_BRANCH_URL` because the application will fail to load and we should use `VERCEL_PROJECT_PRODUCTION_URL` for that.

I think that's something we can set up like this:
```js
const productionUrl = Boolean(process.env.VERCEL_URL) ? process.env.VERCEL_ENV === 'production' ? `https://${process.env. VERCEL_PROJECT_PRODUCTION_URL}` : `https://${process.env.VERCEL_URL}` : process.env.APP_URL;
```

If `VERCEL_PROJECT_PRODUCTION_URL` is not what we're looking for, we can use `process.env.APP_URL` instead.

I would like to be able to actually test this configuration to be sure how it works, but I would need approvals for merging the PR.

emmenko: Yes, the important thing is that for production the URL is static and known. The `VERCEL_PROJECT_PRODUCTION_URL` looks correct for that.

To detect if it's a preview or not, we can also rely on `VERCEL_ENV` maybe
image


PS: are the `VERCEL_URL` really starting with `https://vercel.live/link/`? Shouldn't use the actual alias (for example `https://mc-app-kit-playground-4g7q8lt76-commercetools.vercel.app`) instead? 🤔

CarlosCortizasCT: > To detect if it's a preview or not, we can also rely on VERCEL_ENV maybe
Yes. If you look at the code snippet I provided in my previous message, that's what I was using.

> are the VERCEL_URL really starting with https://vercel.live/link/? Shouldn't use the actual alias (for example https://mc-app-kit-playground-4g7q8lt76-commercetools.vercel.app) instead?
My mistake. I copied the browser link instead of the URL

CarlosCortizasCT: Updated here: 65bc4796e36bb6c0e3412d0c03a3cd1d4f0b8e3b

emmenko (approved):
Thanks!

PR #3548, opened 2024-06-07
refactor: change http-adapter to eager mode

tdeekens:
#### Summary

We've been to lazy and should be more eager.

#### Description

After a lot of communication and alignment we realized that being eager on the http-adapter helps drive user expectations.

The background is that we toggle functionality through the http-adapter which is generally available to users. Having a "delayed" appearing through a reload for these functionalities is not ideal. It's acceptable for anything which sits behind LaunchDarkly and there has also the benefit to avoid the flicker.

#### Followings

1. @CarlosCortizasCT will evaluate the delay waiting for flags would actually cost and if we can globally wait
2. @tdeekens will investigate if we can integrate our http-adapter query into a more global GraphQL query and hydrate the memory-adapter with it to avoid flash flickering

The general storyline is also to get flopflip ready for React Suspense finally when time permits.


CarlosCortizasCT (approved):
Thanks Tobi! 🙇

PR #3551, opened 2024-06-19
fix(customizations templates): include babel-preset package as a dependency

kark:
#### Summary

Creating and starting a new customization using yarn v3 is currently failing:
```
Error: Your application tried to access @commercetools-frontend/babel-preset-mc-app, but it isn't declared in your dependencies; this makes the require call ambiguous and unsound.

Make sure that all the Babel plugins and presets you are using
are defined as dependencies or devDependencies in your package.json
file. It's possible that the missing plugin is loaded by a preset
you are using that forgot to add the plugin to its dependencies: you
can workaround this problem by explicitly adding the missing package
to your top-level package.json.
```

Adding the babel preset dependency appears to solve the issue.

The alternative would be to only allow yarn classic.

### DIFF
--- a/.changeset/sharp-cups-think.md
+++ b/.changeset/sharp-cups-think.md
@@ -1,0 +1,8 @@
+---
+'@commercetools-applications/merchant-center-custom-view-template-starter-typescript': patch
+'@commercetools-applications/merchant-center-template-starter-typescript': patch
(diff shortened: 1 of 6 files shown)


CarlosCortizasCT (approved):
Thanks Kacper 🙇

Rhotimee (approved):
🙌🏽

PR #3556, opened 2024-06-20
chore(message-loader): convert to opt-out with fallback strategy

dogayuksel:
#### Summary

Sets message compilation as opt-out.

If for some reason compilation fails:
- build does not fail
- loader/plugin uses the original file as the fallback
- when it is not the CI environment, warning messages are printed (warning messages in CI behave as errors)

---

When the message file compilation throws an error it looks something like this (webpack):

Screenshot 2024-06-24 at 14 11 26

---

Notice `@formatjs/cli` warning printed to stdout. This does not get propagated through webpack stats, should not cause CI build to fail.

As a sanity check, I have introduced i18n syntax errors on commit: https://github.com/commercetools/merchant-center-application-kit/pull/3556/commits/cbbf361be2cb46ba20fa6bd1ff702852aef4f83c and ran the pipeline: https://github.com/commercetools/merchant-center-application-kit/actions/runs/9645464336
- **All builds were successful.**
- As expected, for cases where we copy core.json to en.json, tests have failed due to malformed strings.


packages/mc-scripts/src/commands/build.ts
```diff
@@ -44,16 +44,7 @@ async function run() {
...
-      console.log(
-        `\nSearch for the ${chalk.underline(
-          chalk.yellow('keywords')
-        )} to learn more about each warning.`
-      );
-      console.log(
-        `To ignore, add ${chalk.cyan(
-          '// eslint-disable-next-line'
-        )} to the line before.\n`
-      );
```
dogayuksel, line 56: we don't run eslint during our build, no need to special handle eslint warnings like this.

nima-ct (approved):

CarlosCortizasCT (approved):
Looks good to me 👌 

Thanks!

emmenko (approved):
Wonderful, thanks!

PR #3560, opened 2024-07-01
fix: ensure page content layout columns have the correct sizes

emmenko:
**Before**
Screenshot 2024-07-01 at 12 41 18

**After**
Screenshot 2024-07-01 at 12 42 36

### DIFF
--- a/.changeset/khaki-icons-smoke.md
+++ b/.changeset/khaki-icons-smoke.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-components': patch
+---
+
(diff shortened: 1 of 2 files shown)


CarlosCortizasCT:
I'm sorry but I fail to understand what was the error.
At least I would expect to see some snapshots to be updated with this fix.

Could you maybe provide a little more context about where we can see this error? 🙏

tdeekens (approved):

CarlosCortizasCT (approved):
Thanks 👏

PR #3578, opened 2024-08-05
refactor: to remove query to load menu feature toggles

tdeekens:
#### Summary

This removes the query to load all menu feature toggles.

#### Description

Internal applications can be hidden on demand. This can be due to them not being available across all clouds mostly. As applications pass flags of their interest to LD, so that we don't load all flags, we need to pass flags of other applications (not viewed) to LD too so that they're included in the response.

For this, each application (mostly internals) can declare a `featureToggle` property on their `menu.json`. This query (in a service) then walks over all internal applications and aggregates those feature toggles by name. These are then passed on to LaunchDarkly, so that LD includes them when building up its flag subscription.

When adopting feature flagging we first only relied on LD but soon after build more long-lived feature configuration flags into an internal API exposed using the `allFeatures` query. We then recommended to use this to enable/disable applications instead of using LD.

Currently, all teams use the allFeatures query to enable/disable applications across environments or projects (during rollout). No team uses, or intends to, use LD for this. In addition with using lazy caching with LD it becomes less interesting to use LD as flag rule changes don't affect running applications instantly any more.

As a result we can remove this query entirely. Saving us a query at boot time of the application. It also lowers the options to hide applications. Internally we will also adopt an `enableApplication*` pattern for features returned from allFeature to make clear if a feature targets an entire application.

Note, as a follow-up the query will be deprecated in the API and return an empty array.


emmenko (approved):
💯

PR #3587, opened 2024-08-08
Provide new prop to add content to left area of form dialog

ddouglasz:
#### Summary

Add a new property `leftAlignedFooterContent`to the form dialog. This provides an additional content positioned to the left area of the form dialog footer.

#### Screenshots

image

### DIFF
--- a/packages/application-components/src/components/dialogs/form-dialog/form-dialog.tsx
+++ b/packages/application-components/src/components/dialogs/form-dialog/form-dialog.tsx
@@ -31,6 +31,7 @@ export type TFormDialogProps = {
   dataAttributesPrimaryButton?: { [key: string]: string };
   getParentSelector?: () => HTMLElement;
   iconLeftSecondaryButton?: ReactElement;
+  leftAlignedFooterContent: ReactNode;
 };
 const defaultProps: Pick<TFormDialogProps, 'labelSecondary' | 'labelPrimary'> =
   {
@@ -59,6 +60,7 @@ const FormDialog = (props: TFormDialogProps) => (
       dataAttributesSecondaryButton={props.dataAttributesSecondaryButton}
       dataAttributesPrimaryButton={props.dataAttributesPrimaryButton}
       iconLeftSecondaryButton={props.iconLeftSecondaryButton}
+      leftAlignedFooterContent={props.leftAlignedFooterContent}
     />
   </DialogContainer>
 );
--- a/packages/application-components/src/components/dialogs/form-dialog/form-dialog.tsx
+++ b/packages/application-components/src/components/dialogs/form-dialog/form-dialog.tsx
@@ -31,7 +31,7 @@ export type TFormDialogProps = {
   dataAttributesPrimaryButton?: { [key: string]: string };
   getParentSelector?: () => HTMLElement;
   iconLeftSecondaryButton?: ReactElement;
-  leftAlignedFooterContent: ReactNode;
+  leftAlignedFooterContent?: ReactNode;
 };
 const defaultProps: Pick<TFormDialogProps, 'labelSecondary' | 'labelPrimary'> =
   {
--- a/packages/application-components/src/components/dialogs/form-dialog/form-dialog.tsx
+++ b/packages/application-components/src/components/dialogs/form-dialog/form-dialog.tsx
@@ -31,7 +31,7 @@ export type TFormDialogProps = {
   dataAttributesPrimaryButton?: { [key: string]: string };
   getParentSelector?: () => HTMLElement;
   iconLeftSecondaryButton?: ReactElement;
-  leftAlignedFooterContent?: ReactNode;
+  footerContent?: ReactNode;
 };
 const defaultProps: Pick<TFormDialogProps, 'labelSecondary' | 'labelPrimary'> =
   {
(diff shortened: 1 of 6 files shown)


FilPob:
@ddouglasz thanks, looks perfect on the screenshot, but if you look at the Percy screenshots it looks like the buttons are nudged to the left.

ddouglasz:
> @ddouglasz thanks, looks perfect on the screenshot, but if you look at the Percy screenshots it looks like the buttons are nudged to the left.

 @FilPob Thats true, I noticed it too, fixing it now. I will update it soon. Thanks! 
 Also, I think we need to have a follow-up task to update the documentation with the new prop and its display too.
 Somewhere here: https://docs.commercetools.com/merchant-center-customizations/api-reference/commercetools-frontend-application-components#formdialog

ddouglasz:
FIxed here: https://github.com/commercetools/merchant-center-application-kit/commit/9a98d8fd85b5615949825c35a2e3b67e612aeb45

FilPob:
Thanks! Looks good now. And Yes we should also add a short description of the prop to the docs

FilPob (approved):

misama-ct (approved):
👍  LGTM, thanks!

packages/application-components/src/components/dialogs/form-dialog/form-dialog.tsx
```diff
@@ -31,6 +31,7 @@ export type TFormDialogProps = {
   dataAttributesPrimaryButton?: { [key: string]: string };
   getParentSelector?: () => HTMLElement;
   iconLeftSecondaryButton?: ReactElement;
+  leftAlignedFooterContent?: ReactNode;
```
emmenko, line 34: Can we name this simply `footerContent`?

It will always be left aligned to the buttons. It also gives more flexibility in case the layout changes in the future.

packages/application-components/src/components/dialogs/internals/dialog-footer.tsx
```diff
@@ -49,19 +50,40 @@ const DialogFooter = (props: Props) => {
...
+            justifyContent="flex-start"
+          >
+            {props.leftAlignedFooterContent}
+          </Spacings.Inline>
+        )}
+        <div
+          css={css`
+            margin-left: auto;
+          `}
+        >
```
emmenko, line 67: I suggest to remove this custom CSS and render an empty `<div />` in the place of the footer content to allow flexbox to do its work (meaning that there are always 2 child elements).

emmenko (commented):
Thanks 🤗

packages/application-components/src/components/dialogs/form-dialog/form-dialog.tsx
tdeekens: I think it's due to the iconLeftSecondaryButton existing also to be consistent. I also would like footerElement more cause we're passing a React element.

packages/application-components/src/components/dialogs/internals/dialog-footer.tsx
ddouglasz: updated here: https://github.com/commercetools/merchant-center-application-kit/pull/3587/commits/e5dce4a0f9d506afde4f37109b791f1fb90c88d9

packages/application-components/src/components/dialogs/internals/dialog-footer.tsx
```diff
@@ -49,19 +50,32 @@ const DialogFooter = (props: Props) => {
...
-          {...filterDataAttributes(props.dataAttributesPrimaryButton)}
-        />
+      <Spacings.Inline alignItems="center" justifyContent="space-between">
+        <Spacings.Inline
+          scale="m"
+          alignItems="center"
+          justifyContent="flex-start"
+        >
+          <div>{props.footerContent}</div>
+        </Spacings.Inline>
```
emmenko, line 60: Unless there is a need for using `<Spacings.Inline>`, I would leave the layout style to consumers.

```suggestion
        <div>{props.footerContent}</div>
```

emmenko (commented):
Thanks!

packages/application-components/src/components/dialogs/form-dialog/form-dialog.tsx
ddouglasz: Yes @tdeekens that would also work. I think both names suggested cover the scope of the prop and its use case. 
I could change it if you want. 
Thanks!

emmenko: @ddouglasz Maybe check in App Kit and UI Kit if there are similar cases and see what kind of naming we are using there.

ddouglasz: Yes, thank you for the suggestion. We have cases like these mostly with "content". Like the MaintenancePageLayout with a prop of `bodyContent?: ReactNode;`

packages/application-components/src/components/dialogs/internals/dialog-footer.tsx
ddouglasz: updated here: https://github.com/commercetools/merchant-center-application-kit/pull/3587/commits/046d7bbe7500a883dd38c7f1cb39c94014e8e5ea
Thank you.

emmenko (approved):

PR #3590, opened 2024-08-09
chore(shell-connectors): add `workspaces` to staticUrlPathsInPositionOfProjectKey array

tylermorrisford:
#### Summary

Adds `workspaces` to the `staticUrlPathsInPositionOfProjectKey` array in `select-project-key-from-url.ts` updates the test in `select-project-key-from-url.spec.js`.

### DIFF
--- a/.changeset/heavy-crews-grow.md
+++ b/.changeset/heavy-crews-grow.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell-connectors': patch
+---
+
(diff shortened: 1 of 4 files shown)


ByronDWall (approved):
🪄✨

emmenko (changes requested):
Thanks but we should be able to get away without it (see https://github.com/commercetools/merchant-center-frontend/pull/17380).

Let's have a sync about it.

PR #3591, opened 2024-08-10
refactor(app-shell): export some internal components

emmenko:


tdeekens (approved):

packages/application-shell/src/index.ts
```diff
@@ -2,15 +2,20 @@ export * from './export-types';
...
 export { applyDefaultMiddlewares } from './configure-store';
 export { default as InjectReducers } from './components/inject-reducers';
+export { default as RedirectToLogin } from './components/redirect-to-login';
+export { default as RedirectToLogout } from './components/redirect-to-logout';
 export { default as RouteCatchAll } from './components/route-catch-all';
-export { default as setupGlobalErrorListener } from './utils/setup-global-error-listener';
 export { default as SetupFlopFlipProvider } from './components/setup-flop-flip-provider';
 export { default as ConfigureIntlProvider } from './components/configure-intl-provider';
 export { default as SuspendedRoute } from './components/suspended-route';
+export { default as setupGlobalErrorListener } from './utils/setup-global-error-listener';
```
ByronDWall, line 18:
```suggestion
export { default as getPreviousProjectKey } from './utils/get-previous-project-key`
export { default as setupGlobalErrorListener } from './utils/setup-global-error-listener';
```

Since we are making internal components available for spaces, would it be alright to also make `getPreviousProjectKey` available so that we can use it to route users to the correct project from the workspaces launcher?

emmenko: Sure. Feel free to push a commit. Thanks

PR #3592, opened 2024-08-16
feat(loggers): allow to ignore urls by regex

emmenko:
In the access logger the option `ignoreUrls` supports both strings and regular expressions.

For strings, the value is matched as-is. You can now use the regular expression to match a certain path structure, for example to ignore requests to load assets.

```ts
const access = createAccessLoggerMiddleware({
  level: 'info',
  json: true,
  ignoreUrls: ['/', '/health', /^\/static\/(.*)/],
});
```

### DIFF
--- a/packages-backend/loggers/src/helpers.ts
+++ b/packages-backend/loggers/src/helpers.ts
@@ -1,0 +1,51 @@
+import type { Request } from 'express';
+import type { TAccessLoggerOptions } from './types';
+
+const parseIps = (request: Request) => {
+  const forwardedFor = request.headers['x-forwarded-for'];
+
+  if (!forwardedFor) {
+    return [];
+  }
+
+  const remoteAddresses = Array.isArray(forwardedFor)
+    ? forwardedFor
+    : forwardedFor.split(',');
+
+  return remoteAddresses;
+};
+
+const createAccessLogSkipper =
+  (options: TAccessLoggerOptions) => (request: Request) => {
+    if (Boolean(options.silent)) {
+      return true;
+    }
+    const hasMatchingIgnoreUrl = (options.ignoreUrls ?? []).some(
+      (uriPathOrRegex) => {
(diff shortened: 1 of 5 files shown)


packages-backend/loggers/src/helpers.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  const remoteAddresses = Array.isArray(forwardedFor)
+    ? forwardedFor
+    : forwardedFor.split(',');
+
+  return remoteAddresses;
+};
+
+const createAccessLogSkipper =
+  (options: TAccessLoggerOptions) => (request: Request) => {
```
kark, line 19: 💯

kark (approved):
Thank you!

packages-backend/loggers/src/helpers.ts
```diff
@@ -0,0 +1,51 @@
...
+    return {
+      clientIp: clientIp ?? remoteAddress,
+      proxyIps,
+      hostname: request.socket ? request.hostname : undefined,
+      ...(remoteAddress ? { remoteAddress } : {}),
+    };
+  } catch (error) {
+    console.error(`Failed to parse request metadata`, error);
+    return {};
+  }
```
Rhotimee, line 48: The operations here are unlikely to throw exceptions. Is there a reason try/catch is being used here?

Rhotimee (approved):
Looks good! Thanks

packages-backend/loggers/src/helpers.ts
emmenko: We had an issue once where an error was thrown here (e.g. cannot access property of ...) and the server crashed.

Therefore, we decided to play it safe and wrap this with a try-catch. That's the main reason.

Rhotimee: oh, I see 👍🏽

PR #3596, opened 2024-08-26
[FCT-980] App kit background changes

jocelyn-dunkley:
#### Summary

This PR adds the gradient background and smaller commercetools logo behind the workspaces UI feature flag. These changes are for the public pages in the MC Auth app, but they use the public-page-layout component in app-kit, therefore these changes need to happen here. 

It was hard to actually simulate what it will look like but I just copied the code into the app kit simulation area and took screenshots. I hope that's enough. You can also refer to the Figma Identity Auth flow screens for what it will actually look like. 

https://www.figma.com/design/dNwDlmxA9AcApzJA8bVkkF/Authorization-Flows?node-id=1-654&t=ujvHJ2y2F0t9TO6I-0

Feature flag is on:
Screenshot 2024-08-26 at 16 28 27

Feature flag is off:
Screenshot 2024-08-26 at 16 28 47


packages/application-components/src/components/public-page-layout/public-page-layout.tsx
```diff
@@ -56,12 +719,29 @@ const PublicPageLayoutContent: FC<TProps> = (props) => {
...
 const PublicPageLayout: FC<TProps> = (props) => {
+  // JK: Do this in development
+  //const enableWorkspacesUi = { value: true };
+  // JK: Do this for the real thing
+  const enableWorkspacesUi = useFlagVariation(
+    featureFlags.ENABLE_WORKSPACES_UI
+  );
+  const isWorkspacesUiEnabled =
+    // @ts-ignore In case it's coming from the MC API, it's an object { value: boolean }.
+    enableWorkspacesUi?.value ?? enableWorkspacesUi;
```
kark, line 730: I believe we can omit the fallback value since we've defined this only as as an API flag, right?

kark (commented):
Hey 👋,
This branch has been deployed [here](https://mc-15303.mc-preview.europe-west1.gcp.escemo.com/login). It evaluates the real feature flag in the deployment.

We'll need a changeset to release these changes.

packages/application-components/src/components/public-page-layout/public-page-layout.tsx
jocelyn-dunkley: yes good point

.changeset/shy-pillows-confess.md
```diff
@@ -0,0 +1,6 @@
+---
+'@commercetools-frontend/application-components': minor
+'@commercetools-frontend/assets': minor
+---
+
+Added a gradient background and smaller commercetools logo behind the workspaces UI feature flag. These changes are for the public pages in the MC Auth app, but they use the public-page-layout component in app-kit.
```
kark, line 6:
```suggestion
Added a gradient background and smaller commercetools logo to the public-page-layout component as part of an ongoing feature effort
```

kark (commented):
Thanks! Just a quick note—shouldn’t the logo be centered?
Screenshot 2024-08-27 at 12 59 14
Just to confirm, this new gradient background will only be needed in the `PublicPageLayout`, right?

jocelyn-dunkley:
> Thanks! Just a quick note—shouldn’t the logo be centered? Screenshot 2024-08-27 at 12 59 14 Just to confirm, this new gradient background will only be needed in the `PublicPageLayout`, right?

Logo centered - I thought so too but all uses of PublicPageLayout use Spacings.Inline or something like that to center the whole content so I thought it would be redundant. 


**Just to confirm, this new gradient background will only be needed in the `PublicPageLayout`, right?**
Yes correct

packages/application-components/src/components/public-page-layout/public-page-layout.tsx
```diff
@@ -41,6 +44,666 @@ const Container = styled.div`
   background-size: cover;
   background-position: center;
 `;
+const NewContainer = styled.div`
```
ByronDWall, line 47: It might make sense to call this something like `GradientBackground`, since it won't always be a `NewContainer`?

packages/application-components/src/components/public-page-layout/public-page-layout.tsx
```diff
@@ -41,6 +44,666 @@ const Container = styled.div`
...
+    radial-gradient(
+      30% 90% at var(--1x) var(--1y),
+      var(--color-turquoise-90) 0% 5%,
+      transparent 80%
+    );
+  background-blend-mode: color;
+  animation: gradients 10s ease both infinite;
+}
+
+@keyframes gradients {
```
ByronDWall, line 86: These animations were originally intended to animate radial gradients of certain colors under specific elements in the workspaces launcher page, I'm wondering if there is a specific need for animation here? 

Not having to animate the gradients would probably make this whole component a little easier to understand / less verbose.

ByronDWall, line 86: ok, looked at the figma, we do need the animations - maybe it would make sense to put this in its own file?

jocelyn-dunkley: We discussed - based on app kit practices it seems all other styled components are all directly in the file so we'll just keep everything in the file here

Rhotimee: +1 on keeping the `GradientBackgroundContainer ` in a separate file to improve readability.

Rhotimee:
Nice work @jocelyn-dunkley.  The new logo is not centred. It would be nice to make it consistent with the Figma design. Thanks.

jocelyn-dunkley:
> I thought so too but all uses of PublicPageLayout use Spacings.Inline or something like that to center the whole content so I thought it would be redundant.

Hi @Rhotimee I mentioned this in another comment: "All uses of PublicPageLayout use Spacings.Inline or something like that to center the whole content so I thought it would be redundant."

mustafaasif2:
Hi @jocelyn-dunkley, just a quick thought: The `Spacings.Inline` component currently centers only its immediate children within the `PublicPageLayout` component, not the entire content (which also includes the logo). I believe it might be more effective to explicitly center the logo for better alignment. However, given that I might not have the full context, please feel free to correct me.

jocelyn-dunkley:
Ok I'll just center the logo - it's not a big deal if it's duplicated somewhere else as well.

packages/application-components/src/components/public-page-layout/public-page-layout.styles.ts
```diff
@@ -0,0 +1,680 @@
...
+    var(--color-turquoise-90) 0% 5%,
+    transparent 80%
+  ),
+  radial-gradient(
+    30% 90% at var(--1x) var(--1y),
+    var(--color-turquoise-90) 0% 5%,
+    transparent 80%
+  );
+background-blend-mode: color;
+animation: gradients 10s ease both infinite;
```
kark, line 57: It looks like the keyframes definition isn’t being recognized by the browser (see attached screenshot).
Screenshot 2024-08-30 at 09 31 12

Maybe we could try moving the `@keyframes gradient` definition higher up in the styled component to ensure it gets picked up correctly?

kark (commented):
Thanks, @jocelyn-dunkley, for all the improvements - it looks good to me. 
I’m just leaving one final note.

packages/application-components/src/components/public-page-layout/public-page-layout.styles.ts
jocelyn-dunkley: good catch, thank you Kacper!

kark (approved):

Rhotimee (approved):
🙌🏽

PR #3599, opened 2024-09-03
FEC-23: Add github action for preview release

jaikumar-tj:
### Summary

This is a just a trial PR to add a GitHub action for the preview release from specific branch. Currently, the releases (both canary & regular version) are triggered only after merging the PR to `main` branch. To ease up the migration workflow and also to prevent blocking the main branch workflow, I created this special `preview release` that could be used for testing purpose without affecting the main branch.
  
[Comment reference](https://commercetools.atlassian.net/wiki/spaces/FEC/pages/edit-v2/1056735357?focusedCommentId=1061814313)

### DIFF
--- a/.github/workflows/preview-release.yml
+++ b/.github/workflows/preview-release.yml
@@ -45,7 +45,7 @@ jobs:
       - name: Publishing preview releases to npm registry
         if: steps.changesets.outputs.published != 'true' && startsWith(github.ref, 'refs/heads/preview/')
         run: |
-          git checkout ${{ github.ref }}
+          git checkout ${{ github.head_ref }}
           pnpm changeset version --snapshot preview
           pnpm changeset publish --tag preview
         env:
--- a/.github/workflows/preview-release.yml
+++ b/.github/workflows/preview-release.yml
@@ -1,0 +1,53 @@
+name: Preview Release
+
+on:
+  push:
+    branches:
+      - 'preview/**' 
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+
+    timeout-minutes: 15
+
+    steps:
+      # Get GitHub token via the CT Changesets App
+      - name: Generate GitHub token (via CT Changesets App)
+        id: generate_github_token
+        uses: tibdex/github-app-token@v2.1.0
+        with:
+          app_id: ${{ secrets.CT_CHANGESETS_APP_ID }}
+          private_key: ${{ secrets.CT_CHANGESETS_APP_PEM }}
+
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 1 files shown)


jaikumar-tj:
This "preview release" github action approach is crucial for the proposed [react 18 migration plan](https://commercetools.atlassian.net/wiki/spaces/FEC/pages/1056735357/React+18+migration). I would like to hear your thoughts regarding this idea of publishing a `preview version` of UI-kit & App-kit from a specific branch (i.e, other than `main` branch). 

This could allow parallel workflow and teams can continue to merge their PR to `main` branch without being blocked by React 18 migration works([for reference](https://commercetools.atlassian.net/wiki/spaces/FEC/pages/1056735357/React+18+migration#Phase-1%3A-Preparation-of-Dependency-packages)). React 18 migration work can be done in a separate branch (eg: branch name starting with `preview-`) and packages can be released with tag "preview" eg: `ui-kit@1.0.0-preview-xyz`

is it feasible to achieve this or do you see any potential issues or challenges with this approach?

cc: @tdeekens , @misama-ct , @ddouglasz, @CarlosCortizasCT

CarlosCortizasCT:
> ve this or do you see any potential issues or challenges with this approa

I agree with the concept.

It seems currently we're releasing a new version every time we merge to main with this version number pattern:
```
0.0.0-canary-<ful-date>
```
We also have a `dist-tag` which is always pointing to the latest canary version released.

Screenshot 2024-09-05 at 12 41 39

Probably we can follow a similar pattern where a specific branch (eg: preview_r18) will trigger a release with a pattern like this:
```
0.0.0-preview_r18-<fulldate>
```
And we could also have a `dist-tag` ("preview_r18") pointing to the latest published version of that preview.

I'm not sure we need to do this abstract as it will be a one time thing and the Design System engineers might want to approach this in a different way in the future if they ever need it.

Another thing to consider is when to trigger that publications to NPM.
Best case scenario we wouldn't be publishing on every commit but perhaps we could have a way to trigger that action from the PR interface.

CarlosCortizasCT:
> @CarlosCortizasCT Since Douglas has implemented the preview deployment in UI-kit similar to how you explained above, will someone take care of implementing similar thing in App-kit based on this trial PR or what is the expectation here?

Hi @jaikumar-tj 

Thanks for bringing this up.
We're currently deciding who's going to continue with the React 18 migration proposal. Whenever we have a candidate s/he will continue the work.

I'll assign myself this PR in the meantime.

emmenko:
Please also consider using [Changeset's prerelease feature](https://github.com/changesets/changesets/blob/main/docs/prereleases.md). We already used it once or twice in the past.

emmenko:
Superseded by #3642

PR #3600, opened 2024-09-04
proposed Fix: issue with submenu vertical position when text wraps

ddouglasz:
#### Summary

Fix issue with submenu vertical position when text wraps

#### Description
When the submenu has a text link that wraps to the next line, the vertical position of the submenu is adjusted, causing the submenu submenu position to be recalculated but already out of visibility. This makes the submenu card to be hidden. This change fixes the issue by adding a 12-pixel buffer to the height calculation when determining if the submenu fits within the viewport below the menu item, this value accounts for the padding introduced on link hover. This ensures that the submenu is displayed correctly even when the text wraps to the next line.

Before: 

After: 

Shorter submenu items still fit when the buffer is added and it has enough vertical space:

### DIFF
--- a/packages/application-shell/src/components/navbar/navbar.tsx
+++ b/packages/application-shell/src/components/navbar/navbar.tsx
@@ -186,12 +186,18 @@ export const ApplicationMenu = (props: ApplicationMenuProps) => {
 
       const [entry] = entries;
 
-      /** Adding a 12 pixel (spacing-25) buffer to the height calculation when determining if the submenu fits within the viewport below the menu item,
-       * this value accounts for the padding introduced on link hover. We need this to account for the additional card height when a hovered submenu text link wraps to the next line*/
+      /**
+       * Adding a 12 pixel buffer to the height calculation when determining if the submenu fits within the viewport below the menu item.
+       * This buffer accounts for the padding added during link hover, accommodating one additional line height.
+       * We need this to account for the additional card height when a hovered submenu text link wraps to the next line,
+       * as otherwise the submenu would extend beyond the viewport's bottom edge and flip above the menu item.
+       */
+      const SINGLE_LINE_HEIGHT_BUFFER = 12;
+
       const doesSubmenuFitWithinViewportBelowMenuItem =
         entry.boundingClientRect.height +
           (props.isMenuOpen ? menuItemTop : menuItemBottom) +
-          12 >
+          SINGLE_LINE_HEIGHT_BUFFER >
         window.innerHeight;
       // if the submenu does not fit at the bottom of the viewport (below the menu item)
       if (doesSubmenuFitWithinViewportBelowMenuItem) {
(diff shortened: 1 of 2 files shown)


packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -186,9 +186,12 @@ export const ApplicationMenu = (props: ApplicationMenuProps) => {
 
       const [entry] = entries;
 
+      /** Adding a 12 pixel (spacing-25) buffer to the height calculation when determining if the submenu fits within the viewport below the menu item,
+       * this value accounts for the padding introduced on link hover. We need this to account for the additional card height when a hovered submenu text link wraps to the next line*/
```
kark, line 190:
```suggestion
      /**
       * Adding a 12 pixel buffer to the height calculation when determining if the submenu fits within the viewport below the menu item.
       * This buffer accounts for the padding added during link hover, accommodating one additional line height.
       * We need this to account for the additional card height when a hovered submenu text link wraps to the next line, 
       * as otherwise the submenu would extend beyond the viewport's bottom edge and flip above the menu item.
       */
```

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -186,9 +186,12 @@ export const ApplicationMenu = (props: ApplicationMenuProps) => {
...
       const doesSubmenuFitWithinViewportBelowMenuItem =
         entry.boundingClientRect.height +
-          (props.isMenuOpen ? menuItemTop : menuItemBottom) >
+          (props.isMenuOpen ? menuItemTop : menuItemBottom) +
+          12 >
```
kark, line 194: Maybe we could use here `const SINGLE_LINE_HEIGHT_BUFFER = 12;`

kark (commented):
Thank you, Douglas, for handling this issue and suggesting a fix! 🙌 I believe the drawback of this approach—having the menu flip 12px above the viewport's bottom—is something we can accept.

Please remember about the changeset

packages/application-shell/src/components/navbar/navbar.tsx
ddouglasz: updated here: https://github.com/commercetools/merchant-center-application-kit/pull/3600/commits/a5e97c3d1c08fd45cc8e8e31c76f2981029b129f

Thank you

packages/application-shell/src/components/navbar/navbar.tsx
ddouglasz: updated here: https://github.com/commercetools/merchant-center-application-kit/pull/3600/commits/a5e97c3d1c08fd45cc8e8e31c76f2981029b129f

Thank you

Rhotimee (approved):

ByronDWall (approved):

kark (approved):

PR #3607, opened 2024-09-19
fix(i18n): `parseChunkImport` helper function performance issue

ragafus:
#### Summary

When bootstrapping the MC Discount web app I noticed that 400ms from the TBT metric are coming from the load of the i18n messages. 
The i18n object loaded seems to be quite big and spreading the object on the loop (which is not needed) makes the load of the i18n messages 200x slower (I've tested it on my local env) for the Discounts app

#### Description
This is a trace from the load of the MC Discounts app.
Screenshot 2024-09-19 at 15 49 07

### DIFF
--- a/.changeset/khaki-dragons-stare.md
+++ b/.changeset/khaki-dragons-stare.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/i18n': patch
+---
+
+Fix performance issue on parseChunkImport helper function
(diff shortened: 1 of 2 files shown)


ragafus:
I was able to reproduce the issue on the MC Products app too.
Screenshot 2024-09-19 at 16 17 03

mario-priceless (approved):
🕵️

kark (approved):
Thanks!
That's a super nice improvement!
| Before (Discounts app)  | After  |
|---|---|
|  Screenshot 2024-09-20 at 08 41 42 | Screenshot 2024-09-20 at 09 03 44   |

PR #3609, opened 2024-09-23
fix(i18n): parallel loading of translations

ragafus:
#### Summary

While debugging https://github.com/commercetools/merchant-center-application-kit/pull/3607 I realized we have an opportunity to improve the `loadI18n` helper function by loading the translations in parallel.

It's a small improvement, but it could save ~50ms on each cold load.

#### Description
This is the network waterfall (Discounts app) that this PR avoids.
[
Screenshot 2024-09-23 at 10 59 50
](url)

### DIFF
--- a/.changeset/wicked-actors-smile.md
+++ b/.changeset/wicked-actors-smile.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/i18n': patch
+---
+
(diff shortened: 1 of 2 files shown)


kark (approved):
Thanks!

emmenko (approved):
Thanks, keep it going! 🙌

PR #3610, opened 2024-09-25
feat: extend the list of cloud identifiers to include Azure EU

mustafaasif2:
#### Summary

Add Azure EU as a new target in the MC configuration.

#### Description

Introduce a new cloudIdentifier configuration option for Custom Applications and Custom Views, allowing them to select Azure EU as the target environment.


kark (approved):
Thanks!
Looks good!

emmenko (approved):
Thanks

PR #3615, opened 2024-09-26
refactor(mc-scripts): manually group some vendor chunks

tylermorrisford:
#### Summary

This PR revives [another PR](https://github.com/commercetools/merchant-center-application-kit/pull/2998) created by @emmenko to manually group some vendor chunks in the `@commercetools-frontend/mc-scripts` package. Doing so gives us a few extra files in the bundle but a significantly smaller vendor chunk. 

#### Description

> To reduce the main bundle chunk even more, we can instruct the bundlers to group certain vendor packages in specific groups. As a result, the main bundle will have a smaller size. Users (via browsers) will then only need to download the smaller main bundle when it gets updated (caching) as the other vendor chunks didn't change (unless via dependency updates).
> 
> I found this article that explains well how and why it's useful like that: https://www.codemzy.com/blog/react-bundle-size-webpack-code-splitting

Using [this method](https://github.com/commercetools/test-data/?tab=readme-ov-file#local-development-with-linked-packages) to test, I rebuilt the account application before and after making these changes; this resulted in the main app chunk being reduced from ~660KB to ~412KB. Based on this I feel that all monorepo and satellite applications would benefit. 

**UPDATE**
We had to change a bit the packages we wanted to create independent bundles for.
We need to bear in mind that splitting is done at the application level, so what we can split are their direct dependencies so it's not possible to split internals of (for instance) the `application-shell` package (that's something we can do separately and internally to that package).
We decided to create manual bundles for:
* `application-shell` and `application-shell-connectors`
* `moment` and `moment-timezone`
* `@commercetools-uikit/icons`

Building the playground application, here is what biggest files originally looks like (**webpack**):
image

and this is how it looks like with the new manual bundles:
image

Also, this is the comparison when using **vite**.
BEFORE
image

AFTER
image

We're also adding new vite plugins that will help to analyze the built bundles using this env variables:
* `ANALYZE_BUNDLE`
* `ANALYZE_BUNDLE_TREE`

This is how they look like:
Screenshot 2024-10-08 at 18 31 02

Screenshot 2024-10-08 at 18 32 05


packages/mc-scripts/src/config/optimizations.ts
```diff
@@ -0,0 +1,40 @@
...
+    'react-dom',
+    'react-router',
+    'react-router-dom',
+    'react-intl',
+    '@reduxjs/toolkit',
+    'redux',
+    'react-redux',
+    'redux-logger',
+    'redux-thunk',
+  ],
```
tylermorrisford, line 22: @emmenko the only change I made is to lump redux and react together; since redux is fairly small this seems fine to do.

CarlosCortizasCT:
> Using [this method](https://github.com/commercetools/test-data/?tab=readme-ov-file#local-development-with-linked-packages) to test, I rebuilt the account application before and after making these changes; this resulted in the main app chunk being reduced from ~660KB to ~412KB. Based on this I feel that all monorepo and satellite applications would benefit.

@tylermorrisford by default the applications are built using webpack but we also have "experimental" support for Vite.
Could you please also run that comparison by building the application using Vite so we also get those numbers?
```
ENABLE_EXPERIMENTAL_VITE_BUNDLER_SOURCEMAP=true pnpm build
```

CarlosCortizasCT:
Building an application is failing for both `webpack` and `vite` bundlers.

To test it, you can get into the `playground` directory and run:
```
// Build with webpack
ENABLE_EXPERIMENTAL_VITE_BUNDLER=false pnpm build

// Build with Vite
ENABLE_EXPERIMENTAL_VITE_BUNDLER=true pnpm build
```

I think we need to investigate a little further what the updated configuration actually does.

tylermorrisford:
Two items I've discovered in testing the Vite build process by linking the my local version of app-kit with the monorepo and building the account application: 

1. Our Vite config seems to get tripped up on files wit `.js` extensions, updating those files to `.jsx` allowed the build to proceed.
2. There is an export in `packages-application/application-account/src/components/organizations/details/custom-extensions/custom-applications/configurations/_shared/form/default-raw-icon.js` using `!!raw-loader!` which causes the build to fail, unless the entire export is added to build.rollupOptions.external.

package.json
```diff
@@ -181,7 +181,8 @@
       "@typescript-eslint/parser": "^5.52.0"
     },
     "patchedDependencies": {
-      "babel-plugin-typescript-to-proptypes@1.4.2": "patches/babel-plugin-typescript-to-proptypes@1.4.2.patch"
+      "babel-plugin-typescript-to-proptypes@1.4.2": "patches/babel-plugin-typescript-to-proptypes@1.4.2.patch",
+      "vite-bundle-analyzer@0.12.1": "patches/vite-bundle-analyzer@0.12.1.patch"
```
CarlosCortizasCT, line 185: `vite-bundle-analyzer` has a `rollup` very flexible peer dependency (2.X -> 4.X) but it actually requires 3.X at least because of a typescript type usage.
Since we're using rollup 2.X in the repo (force by `rollup`), the alternative I took is to change the type usage (very simple change)

packages/mc-scripts/src/commands/build-vite.ts
```diff
@@ -77,6 +86,10 @@ async function run() {
       pluginSvgr(),
       pluginDynamicBaseAssetsGlobals(),
       pluginI18nMessageCompilation(),
+      process.env.ANALYZE_BUNDLE === 'true' &&
```
CarlosCortizasCT, line 89: This plugin create the bundles visualization size:

packages/mc-scripts/src/commands/build-vite.ts
```diff
@@ -77,6 +86,10 @@ async function run() {
...
+        analyzer({ defaultSizes: 'parsed', openAnalyzer: true }),
+      process.env.ANALYZE_BUNDLE_TREE === 'true' &&
```
CarlosCortizasCT, line 91: This plugin creates a bundles dependency visualization:
Screenshot 2024-10-08 at 18 32 05

packages/mc-scripts/src/config/optimizations.ts
```diff
@@ -0,0 +1,66 @@
...
+const manualChunks = {
+  'commercetools-uikit-icons': ['@commercetools-uikit/icons'],
+  moment: ['moment', 'moment-timezone'],
+  'app-shell': [
+    '@commercetools-frontend/application-shell',
+    '@commercetools-frontend/application-shell-connectors',
+  ],
+};
+
+const removeNonExistantDependencies = (
```
CarlosCortizasCT, line 15: This is just to make sure the packages we're defining in the `manualChunks` map actually exists in the application to build.
This is a safety measure so in case one of them are remove (eg: moment) the build does not fail but will only emit a warning about it.

CarlosCortizasCT (approved):

packages/mc-scripts/package.json
```diff
@@ -108,6 +109,7 @@
     "thread-loader": "3.0.4",
     "url": "^0.11.0",
     "vite": "~4.5.3",
+    "vite-bundle-analyzer": "0.12.1",
```
emmenko, line 112: I'm not sure if the patch will be used here when someone installs the `mc-scripts` package.

packages/mc-scripts/src/config/optimizations.ts
```diff
@@ -0,0 +1,66 @@
...
+  manualChunks: Record<string, string[]>,
+  appDependencies: string[]
+) => {
+  return Object.entries(manualChunks).reduce((acc, [chunkName, vendors]) => {
```
emmenko, line 19: Nit: let's use a more meaningful name

```suggestion
  return Object.entries(manualChunks).reduce((chunkGroups, [chunkName, vendors]) => {
```

packages/mc-scripts/src/config/optimizations.ts
```diff
@@ -0,0 +1,66 @@
...
+    const existingVendors = vendors.filter((vendor) =>
+      appDependencies.includes(vendor)
+    );
+    if (existingVendors.length > 0) {
```
emmenko, line 23:
```suggestion
    const hasExistingVendors = vendors.some((vendor) =>
      appDependencies.includes(vendor)
    );
    if (hasExistingVendors) {
```

emmenko (commented):
Great job, thanks for the contributions! 🙌

.changeset/grumpy-fireants-protect.md
```diff
@@ -0,0 +1,27 @@
...
+```bash
+ENABLE_EXPERIMENTAL_VITE_BUNDLER=true
+```
+
+Also, if you are using `vite`, we've also added a couple of plugins you can use to analyze the built bundle sizes:
+
+- [vite-bundle-analyzer](https://github.com/KusStar/vite-bundle-visualizer)
+- [rollup-plugin-visualizer](https://github.com/btd/rollup-plugin-visualizer) (tree visualization)
+
+You can use them by setting this environment variables"
```
tdeekens, line 20:
```suggestion
You can use them by setting this environment variables:
```

packages/mc-scripts/package.json
tdeekens: Fair question. Couldn't find documentation on it. I must say I assume it would be applied.

packages/mc-scripts/src/config/optimizations.ts
tdeekens: Thanks for the comment. I was fearful we'd remove something else but those we specify at first.

tdeekens (approved):
Thanks a lot for the effort on this work!

packages/mc-scripts/package.json
CarlosCortizasCT: I don't know either but it won't matter in this case: the change in the source code is about a Typescript type which is not generic in the `rollup` version we're using but the dependency code expects it to be.
So, the generated code is exactly the same (no typings go there).

packages/mc-scripts/src/config/optimizations.ts
CarlosCortizasCT: Updated here: f8f062ce27d6b580d54781cd3838f079cf43099d

.changeset/grumpy-fireants-protect.md
CarlosCortizasCT: Updated here: f8f062ce27d6b580d54781cd3838f079cf43099d

packages/mc-scripts/src/config/optimizations.ts
CarlosCortizasCT: Let's picture this scenario:
```ts
const manualChunks = {
  'chunk-A': ['dep-a', 'dep-b'];
}
```

With the current code, if the app does not have 'dep-b' anymore, the manual chunk would still be created with `dep-a`. This would not work if we apply your suggestion.

emmenko: Thanks, nevermind then. I misread the code 🙈

emmenko (approved):

CarlosCortizasCT (approved):

PR #3616, opened 2024-09-27
feat(sentry): send extra info for reported ErrorEvent

ragafus:
#### Summary
The goal of this PR is to add more info of any ErrorEvent reported to Sentry. 

The captured error message (see https://commercetools.sentry.io/issues/5880399228 as an example) is not enough to have a picture of the issue happening on production.

As an example: introducing the filename to the extra data, we will be able to understand if issues are caused by browser extensions.


Rhotimee (approved):
Thanks!

PR #3618, opened 2024-10-01
fix(l10n): explicitly include export file for supported locales

emmenko:
Otherwise TS gets confused and the types cannot be properly recognized.

image


Rhotimee (approved):

kark (approved):
Thanks!

PR #3622, opened 2024-10-09
Fct 1232 remove workspaces launcher button

tylermorrisford:
#### Summary

Removes the workspaces launcher button from the app-bar.


emmenko:
Looks like there are unwanted changes in the PR 🤔

PR #3630, opened 2024-10-24
[FEC-96] Migrate react-router redirects

CarlosCortizasCT:
#### Summary

Migrate react-router redirects

#### Description

As part of the preparations for the upcoming `react-router` v6 migration, there are some changes related to how client-side redirects are managed ([reference](https://reactrouter.com/en/main/upgrading/v5#remove-redirects-inside-switch)) that we can already apply to our codebase (supported in v5).

### DIFF
--- a/packages/application-components/src/components/detail-pages/tabular-detail-page/tabular-detail-page.spec.tsx
+++ b/packages/application-components/src/components/detail-pages/tabular-detail-page/tabular-detail-page.spec.tsx
@@ -19,9 +19,7 @@ jest.mock('@commercetools-uikit/utils', () => ({
 const Content = () => (
   <Spacings.Stack scale="m">
     <Switch>
-      <Route exact={true} path="/">
-        <Redirect to="/tab-one" />
-      </Route>
+      <Route exact={true} path="/" render={() => <Redirect to="/tab-one" />} />
       <Route path="/tab-one">
         <Text.Body>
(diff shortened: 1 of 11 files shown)


mario-priceless (approved):

tdeekens (approved):

kark (approved):
Thank you Carlos 🙌
Just a friendly reminder to add the changeset before merging.
Many thanks

Rhotimee (approved):
Nice!

CarlosCortizasCT:
> Thank you Carlos 🙌 Just a friendly reminder to add the changeset before merging. Many thanks

Thanks for the reminder, @kark 🙇 

Included here: 146330f102809a1145968ee10dabf5c0b152160e

packages/application-components/src/components/detail-pages/tabular-detail-page/tabular-detail-page.spec.tsx
```diff
@@ -19,9 +19,7 @@ jest.mock('@commercetools-uikit/utils', () => ({
 const Content = () => (
   <Spacings.Stack scale="m">
     <Switch>
-      <Route exact={true} path="/">
-        <Redirect to="/tab-one" />
-      </Route>
```
emmenko, line 24: Nit: this is technically correct.


The docs say 

> Normal `<Redirect>` elements that are not inside a `<Switch>` are ok to remain. 

> Remove any `<Redirect>` elements that are directly inside a `<Switch>`.

So only `<Switch><Redirect /></Switch>` should be migrated.

Using `render` or `children` is also equivalent, although if I remember back then they recommended to prefer `children`.

emmenko (approved):
Thanks

packages/application-components/src/components/detail-pages/tabular-detail-page/tabular-detail-page.spec.tsx
CarlosCortizasCT: Render `children` is discouraged in v6 as we will need to use the `element` property of the `Route` component so I thought it was easier to follow the prop recommendation.

emmenko: Yes, that makes sense. Thanks!

PR #3634, opened 2024-10-31
refactor(application.html): add meta tag to prevent search engine indexing

misama-ct:
## Problem
When a user Googles "Merchant Center commercetools", we don't want the various MC's to appear in the search result pages.

## Solution
We add a meta-tag to the `application.html` file to prevent search engines from indexing the apps and potentially following  links.

Breakdown of the meta-tag directive:
- `noindex`: Prevents search engines from indexing the page, so it won’t appear in search results.
- `nofollow`: Tells search engines not to follow the links on the page

### DIFF
--- a/.changeset/nasty-zoos-shop.md
+++ b/.changeset/nasty-zoos-shop.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/mc-html-template': patch
+---
+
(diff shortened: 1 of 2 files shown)


emmenko (commented):
Thanks! Please add a changeset to be able to release this.

tdeekens (approved):

emmenko (approved):

PR #3640, opened 2024-11-19
fix(navbar): prevent menu-items from growing when hovered

misama-ct:
#### Summary
When hovering an element in a navbar-submenu with a long wording, the wording may wrap into the next line when the item is hovered and cause sub-menus to close, essentially making some menu-items unnavigable.

#### Description
The problem is caused by the hovering element changing its size during the hover, triggering a layout shift of neighbouring elements.

After syncing on what is possible @FilPob and I agreed on keeping a fixed, but slightly increased width for submenus.
To preserve the animation of the text-indentation, the moving of the text is now done with relative positioning. This way, padding-values stay untouched and the DOM element won't change dimensions and can't cause layout shifts.

Additionally, the mechanism to truncate/clamp text after 2 consequent lines of text was fixed (currently it clamps, but also shows a third line of text).

### DIFF
--- a/packages/application-shell/src/components/navbar/menu-items.styles.ts
+++ b/packages/application-shell/src/components/navbar/menu-items.styles.ts
@@ -98,6 +98,16 @@ const TextLinkSublistWrapper = styled.div`
 const NavlinkClickableContent = styled.div`
   height: 100%;
   width: 100%;
+  display: block;
+  display: -webkit-box;
+  -webkit-line-clamp: 2;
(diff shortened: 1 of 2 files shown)


ddouglasz:
Feels like a good fix to me. @FilPob what do you think ?

tdeekens (approved):
Thanks!

kark (approved):
Thanks for the improvements 💯🎉

tylermorrisford (approved):

PR #3662, opened 2024-12-02
ci: define proper `.npmrc` file

emmenko:
Follow up of #3642

### DIFF
--- a/.github/workflows/preview-release-on-comment.yml
+++ b/.github/workflows/preview-release-on-comment.yml
@@ -25,11 +25,11 @@ jobs:
         with:
           app_id: ${{ secrets.CT_CHANGESETS_APP_ID }}
           private_key: ${{ secrets.CT_CHANGESETS_APP_PEM }}
-      
+
       - name: Get branch of PR
         uses: xt0rted/pull-request-comment-branch@v1
         id: comment-branch
-      
+
       - name: Ensure branch name matches 'preview/**'
         run: |
           BRANCH_NAME=${{ steps.comment-branch.outputs.head_ref }}
@@ -37,7 +37,7 @@ jobs:
             echo "Branch name does not match 'preview/**'. Exiting workflow."
             exit 1
           fi
-        
+
       - name: Set latest commit status as pending
         uses: myrotvorets/set-commit-status-action@master
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/preview-release-on-comment.yml
```diff
@@ -49,20 +49,21 @@ jobs:
         uses: actions/checkout@v4
         with:
           ref: ${{ steps.comment-branch.outputs.head_ref }}
-          token: ${{ steps.generate_github_token.outputs.token }}   
-
-      - name: Setup Node.js
-        uses: actions/setup-node@v3
-        with:
-          node-version: 18
```
emmenko, line 57: We don't want to hardcode it.

Installing node is already done in the `./.github/actions/ci` action

.github/workflows/preview-release-on-comment.yml
```diff
@@ -49,20 +49,21 @@ jobs:
...
 
+      - name: Creating .npmrc
+        run: |
+          cat << EOF > "$HOME/.npmrc"
+            provenance=true
+            email=npmjs@commercetools.com
+            //registry.npmjs.org/:_authToken=$NPM_TOKEN
+          EOF
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```
emmenko, line 65: This was missing, in particular for defining the `provenance=true` setting

.github/workflows/preview-release-on-comment.yml
```diff
@@ -25,44 +25,46 @@ jobs:
...
-        
+
       - name: Set latest commit status as pending
-        uses: myrotvorets/set-commit-status-action@master
+        uses: myrotvorets/set-commit-status-action@v2
         with:
           sha: ${{ steps.comment-branch.outputs.head_sha }}
           token: ${{ steps.generate_github_token.outputs.token }}
           status: pending
+          allowForks: false
```
emmenko, line 47: This is the default but I think it's better to be explicit about it

.github/workflows/preview-release-on-comment.yml
```diff
@@ -79,11 +81,12 @@ jobs:
...
+
       - name: Set latest commit status as ${{ job.status }}
-        uses: myrotvorets/set-commit-status-action@master
+        uses: myrotvorets/set-commit-status-action@v2
         if: always()
         with:
           sha: ${{ steps.comment-branch.outputs.head_sha }}
           token: ${{ steps.generate_github_token.outputs.token }}
           status: ${{ job.status }}
+          allowForks: false
```
CarlosCortizasCT, line 92: In case other reviewers don't know about this property (as me), here is its description:
> Whether to allow to set commit status if the request comes from a forked repository. Defaults to false because the GITHUB_TOKEN for a forked repository usually does not have the necessary permissions to update the commit status. This only affects pull requests.

CarlosCortizasCT (approved):
Thanks! 👍

.github/workflows/preview-release-on-comment.yml
Rhotimee: nice, thanks

PR #3663, opened 2024-12-02
Validate branch name in preview release workflow

CarlosCortizasCT:
#### Summary

Validate branch name in preview release workflow

#### Description

We detected some attempts to exploit the new workflow for preview releases so we want to include a quick and simple validation for the branch names we use in order to protect the flow.

We now only allow numbers, letters, underscores, and hyphens in the branch name.


emmenko (approved):
Thanks!

woodruffw:
FYI, this fix is incomplete: any amount of `${{ ... }}` expansion in a shell context with attacker-controlled data is vulnerable. 

I think the appropriate fix here is to replace `BRANCH_NAME=${{ steps.comment-branch.outputs.head_ref }}` with an `env:` block setting, and similarly for anywhere else that you use this kind of template-in-shell pattern.

I can send a fix PR for this in a bit.

PR #3668, opened 2024-12-06
feat(omlet): basic omlet.dev integration

misama-ct:
#### Summary

This PR integrates the possibility to scan the repository with [omlet.dev](https:/omlet.dev/) via their cli. A scan will collect information about component usage and dependencies.

#### Description
I added the omlet config file (`.omletrc`) to the root, configured it to the best of my knowledge and added an `omlet:scan` script-shortcut to the root `package.json`.

I also prepared a hook-file according to the omlet docs, so that we can easily add more meta-data to the component-scan results in the future (e.g. add the owning team to a component).

### DIFF
--- a/.omletrc
+++ b/.omletrc
@@ -1,0 +1,38 @@
+{
+  "ignore": [
+       "**/dist/**",
+       "**/visual-testing-app/**"
+  ],
+  "workspaces": {
+    "@commercetools-frontend/application-components": {
+      "exports": {
+        ".": "./src/index.ts"
+      }
+    },
+    "@commercetools-frontend/application-shell": {
+      "exports": {
+        ".": "./src/index.ts"
+      }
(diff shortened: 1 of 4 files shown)


emmenko (approved):
Thanks!

Is there also a GitHub <--> Omlet integration behind the scenes? Or it will be a follow up step?

misama-ct:
> Is there also a GitHub <--> Omlet integration behind the scenes? Or it will be a follow up step?

It'll be a follow-up step unfortunatly. Omlet needs `fetch-depth` set to `0` when checking out the code (to fetch all history for all branches and tags) and I didn't (and still don't) know what possible side-effects that could have. So there will be a bit of investigation needed and a follow-up.

ByronDWall (approved):

PR #3530, opened 2024-05-16
[Breaking] Drop support for Node.js `v16`

emmenko:
Drop support for Node.js `v16`, which [reached EOL](https://nodejs.org/en/about/previous-releases). The minimal required version is `v18` but we recommend to use `v20` or `v22` if possible as `v18` goes EOL mid 2025.

Remove support for `v16` unblocks also other upgrades, for example Vite `v6`.

PR #3669, opened 2024-12-06
refactor: upgrade `cosmiconfig` to `v9`

emmenko:
TL;DR: Upgrade `cosmiconfig` to the latest version `v9`.
However, there is more...

## Background

We currently run on Node.js `v18` but we are planning to upgrade to more recent and maintained versions and dropping `v16` (https://github.com/commercetools/merchant-center-application-kit/pull/3530).

As we are working on the Node upgrade, on Node `>=v20` we get some errors when loading the Custom Application config file, in particular when using an extension such as `.mjs`. The error  looks like this:

image

or this

image

These errors are happening now in Node 19+ as the `--experimental-specifier-resolution` was removed.

How can we fix that?

## Proposal

In our Merchant Center applications we mainly use file extensions `.js`, `.mjs` or `.ts` for the configuration file, while the application code is either `.js` or `.ts`.
Ideally, if we use `.mjs` (ESM) we would need to explicitly add the file extension to each import as expected by Nodejs.

However, for practical reasons and for backwards compatibility we would like to support different import combinations. For example, the config file using `.mjs` but importing a `.ts` or `.js` file.

The place to handle this logic would be around `cosmiconfig`, in particular the loaders. So far we have been using a "custom JS loader" which wasn't ideal in the first place and felt a bit hacky.

Looking around other tools I stumbled across [Commitlint that also uses cosmiconfig internally](https://github.com/conventional-changelog/commitlint/blob/master/%40commitlint/load/src/utils/load-config.ts) and noticed that they use the `cosmiconfig-typescript-loader` package.
So I gave that a try and it seems that it does solve the issues.


packages/application-config/src/load-config.ts
```diff
@@ -1,83 +1,86 @@
...
+
+// Is the given directory set up to use ESM (ECMAScript Modules)?
+// Copied from @commitlint/load/src/utils/load-config.ts
+export const isEsmModule = async (cwd: string) => {
+  const packagePath = path.join(cwd, 'package.json');
+
   try {
-    fs.accessSync(path);
-    return true;
+    await fs.promises.access(packagePath, fs.constants.F_OK);
```
CarlosCortizasCT, line 33: **nit**: do we actually need this `access` call?
It seems it will throw and error if user does not have access but the next line (read file) would also throw an error when no access is available and we're catching both in the same `catch` statement.

packages/application-config/src/load-config.ts
```diff
@@ -1,83 +1,86 @@
...
+    if (!tsLoaderInstance) {
+      tsLoaderInstance = TypeScriptLoader();
+    }
+    return tsLoaderInstance(...args);
+  };
 
-const createExplorerFor = (configFileName: string) => {
-  return cosmiconfigSync(configFileName, {
-    // Restrict the supported file formats / names
+  // If dynamic await is supported (Node >= v20.8.0) or directory uses ESM, support
```
CarlosCortizasCT, line 53: **nit**: I still struggle to understand why we need to use async loaders and how `defaultLoaders` seems to change its behaviour across node versions (we were already using it but not awaiting the search files).

CarlosCortizasCT (approved):
Thanks 👍

packages/application-config/src/load-config.ts
emmenko: True, we can omit the `access` call yes

packages/application-config/src/load-config.ts
emmenko: I just copied it from here: https://github.com/conventional-changelog/commitlint/blob/fed05e2ab12792ce488f34a580643669ab6ac831/%40commitlint/load/src/utils/load-config.ts#L33-L38

In any case, it refers to this part of the docs: https://github.com/cosmiconfig/cosmiconfig/blob/main/README.md#loading-js-modules
image

Rhotimee (approved):

packages/application-config/src/load-config.ts
emmenko: https://github.com/commercetools/merchant-center-application-kit/pull/3669/commits/fe0e58007bc6394dd54bee3e9de38674faad1cf9

PR #3670, opened 2024-12-06
workflows: fix some exploitable template injections

woodruffw:
This follows https://github.com/commercetools/merchant-center-application-kit/pull/3663, and more generally fixes these template injections.

I found these with [zizmor](https://github.com/woodruffw/zizmor).

**IMPORTANT**: I did *not* fix all of `zizmor's` findings, since there are 40+ of them on your repository. I strongly suggest going through each and evaluating them/your risk tolerance.


emmenko (approved):
Thanks a lot for the contribution and for helping out with the security fix!

We'll also check out the tool `zizmor` and see if we can adopt it 🙌

PR #3677, opened 2024-12-11
fix(AppShell): project switching

kark:
#### Summary

We recently updated how redirects are handled (see [PR #3630](https://github.com/commercetools/merchant-center-application-kit/pull/3630)). Unfortunately, this introduced a regression in project switching behavior. This fix addresses the issue by properly resolving route parameters during redirects.

image


Rhotimee (approved):
Thanks

emmenko (approved):
Thanks!

PR #3681, opened 2024-12-17
New codemod to help migrating React components "defaultProps" usage

CarlosCortizasCT:
## Summary

The way we are using React components `defaultProps` in our codebase is no longer supported in the new React 19 version ([reference](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-proptypes-and-defaultprops)).

This is the main concept:
```tsx
// BEFORE
type TMyComponentProps = {
  prop1: string;
  prop2: string;
  prop3: string;
};

function MyComponent(props: TMyComponentProps) {
  return (
    <ul>
    <li>Prop 1: {props.prop1}</li>
    <li>Prop 2: {props.prop2}</li>
    <li>Prop 3: {props.prop3}</li>
  </ul>
  );
}
MyComponent.defaultProps = {
  prop1: 'My default value',
};


// AFTER
type TMyComponentProps = {
  prop1?: string; // <--- Make sure this is marked as optional
  prop2: string;
  prop3: string;
};

function MyComponent({ prop1: 'My default value', ...props }: TMyComponentProps) {
  return (
    <ul>
    <li>Prop 1: {prop1}</li> {* <--- Direct access to destructured property *}
    <li>Prop 2: {props.prop2}</li>
    <li>Prop 3: {props.prop3}</li>
  </ul>
  );
}
// <--- No "defaultProps" component property
```

## Description

There are multiple adjustments that are needed and this codemod does not address 100% of them but it helps with most of the common use cases. However, a double check review is recommended after running it; at least checking the Typescript types and building the codebase.

These are the main steps the codemod goes through:
1. Search for React components that have a `defaultProps` property
```tsx
MyComponent.defaultProps = {
  prop1: 'My default value',
};
```
2. Extract the keys and values for the default props
3. Update the component function signature
```tsx
// BEFORE
function MyComponent(props: TMyComponentProps) { ... }

// AFTER
function MyComponent({ prop1, ...props }: TMyComponentProps) { ... }
```
4. Refactor the usages of the default props in the body of the component
```tsx
// BEFORE
function MyComponent(props: TMyComponentProps) {
  return (
    <ul>
      <li>Prop 1: {props.prop1}</li>
      <li>Prop 2: {props.prop2}</li>
      <li>Prop 3: {props.prop3}</li>
    </ul>
  );
}

// AFTER
// BEFORE
function MyComponent({ prop1, ...props }: TMyComponentProps) {
  return (
    <ul>
      <li>Prop 1: {prop1}</li>
      <li>Prop 2: {props.prop2}</li>
      <li>Prop 3: {props.prop3}</li>
    </ul>
  );
}

```
5. Update the component TS type definition so we make sure the default props are optional
```tsx
// BEFORE 
type TMyComponentProps = {
  prop1: string;
  prop2: string;
  prop3: string;
};

// AFTER
type TMyComponentProps = {
  prop1?: string;
  prop2: string;
  prop3: string;
};
```
6. Remove the `defaultProps` assignment from the component
```tsx
// Delete this code block
MyComponent.defaultProps = {
  prop1: 'My default value',
};
```

### NOTE
I tried to be very verbose with the inline comments of the codemod explaining what every block is for.

### DIFF
--- a/packages/codemod/src/transforms/react-default-props-migration.ts
+++ b/packages/codemod/src/transforms/react-default-props-migration.ts
@@ -3,10 +3,12 @@ import {
   AssignmentExpression,
   ASTPath,
   CallExpression,
+  Collection,
   FileInfo,
   JSCodeshift,
   MemberExpression,
   ObjectExpression,
+  PropertyPattern,
   TSTypeAliasDeclaration,
 } from 'jscodeshift';
 import prettier from 'prettier';
@@ -35,7 +37,7 @@ function replacePropsUsage({
 }: {
   j: JSCodeshift;
   defaultPropsKeys: string[];
-  scope: JSCodeshift;
+  scope: Collection;
 }) {
   /*
     Next code block replaces destructured props usage in the component body.
@@ -137,7 +139,7 @@ function updateComponentTypes({
   destructuredKeys,
 }: {
   j: JSCodeshift;
-  root: JSCodeshift;
+  root: Collection;
   typeName: string;
   destructuredKeys: string[];
 }) {
@@ -163,6 +165,66 @@ function updateComponentTypes({
     });
 }
 
+/*
+  This helper transforms the component function signature to use a destructured object
+  as first parameter, so we can append the default props to it.
+  Example:
+  ```
+  // BEFORE
+  const MyComponent = (props) => { ... }
+
+  // AFTER
+  const MyComponent = ({ prop1, ...props }) => { ... }
+  ```
+*/
+function transformComponentFunctionSignature({
+  functionPropsParam,
+  defaultPropsMap,
+  componentName,
+  j,
+}: {
+  functionPropsParam: PropertyPattern['pattern'];
+  defaultPropsMap: Record<string, unknown>;
+  componentName: string;
+  j: JSCodeshift;
+}): PropertyPattern['pattern'] {
+  let refactoredParameter: PropertyPattern['pattern'];
+  switch (functionPropsParam.type) {
+    // In this case, the component already has a destructured object as first parameter
+    // so we need to append the defaultProps to it
+    //   const MyComnponent = ({ prop1, ...props }) => { ... }
+    case 'ObjectPattern':
+      refactoredParameter = functionPropsParam;
+      refactoredParameter.properties = [
+        ...transformDefaultPropsToAST(defaultPropsMap, j),
+        ...functionPropsParam.properties,
+      ];
+      break;
+    // In this case, the component has a simple parameter as first parameter
+    // so we need to refactor it to a destructured object
+    //   const MyComnponent = (props) => { ... }
+    case 'Identifier':
+      refactoredParameter = j.objectPattern([
+        ...transformDefaultPropsToAST(defaultPropsMap, j),
+        j.spreadProperty(j.identifier('props')),
+      ]);
+
+      // Make sure the refactored parameter has the same type annotation
+      // as the original one
+      refactoredParameter.typeAnnotation = functionPropsParam.typeAnnotation;
+      break;
+    default:
+      console.warn(
+        `[WARNING]: Could not parse component function first parameter "${componentName}"`
+      );
+  }
+
+  return refactoredParameter!;
+}
+
+/*
+  This helper extracts the default props keys/values from the defaultProps object node
+*/
 function extractDefaultPropsFromNode(
   defaultPropsNode: ObjectExpression
 ): Record<string, unknown> {
@@ -179,6 +241,24 @@ function extractDefaultPropsFromNode(
     }
     return acc;
   }, {} as Record<string, unknown>);
+}
+
+/*
+  This helper transforms the default props keys/values to an AST representation
+  so we can easily append them to the component function signature
+*/
+function transformDefaultPropsToAST(
+  defaultPropsMap: Record<string, unknown>,
+  j: JSCodeshift
+) {
+  return Object.entries(defaultPropsMap).map(([key, value]) => {
+    const prop = j.objectProperty(
+      j.identifier(key),
+      j.assignmentPattern(j.identifier(key), j.literal(value as string))
+    );
+    prop.shorthand = true;
+    return prop;
+  });
 }
 
 async function reactDefaultPropsMigration(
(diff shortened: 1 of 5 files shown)


packages/codemod/src/transforms/react-default-props-migration.ts
```diff
@@ -0,0 +1,426 @@
...
+          return <div>{props.prop1}</div>;
+        }
+      }
+
+      // AFTER
+      const MyComponent = ({ prop1, ...props }) => {
+        return <div>{prop1}</div>;
+      }
+    ```
+  */
```
tdeekens, line 56: Thanks for the comments!

packages/codemod/src/transforms/react-default-props-migration.ts
```diff
@@ -0,0 +1,426 @@
...
+        const componentName = path.node.left.object.name;
+        const defaultPropsNode = path.node.right;
+        let defaultPropsMap: Record<string, unknown> = {};
+        let functionScope: Collection;
+
+        // 2. We now extract the default props values
+        // Default props can be defined inline or as a reference to another object
+        //  INLINE -- MyComponent.defaultProps: { prop1: 'value1', prop2: 'value2' }
+        //  REFERENCE -- MyComponent.defaultProps: defaultProps
+        if (defaultPropsNode.type === 'Identifier') {
```
tdeekens, line 300: Good point. Didn't think of this use case.

packages/codemod/src/transforms/react-default-props-migration.ts
```diff
@@ -0,0 +1,426 @@
...
+              // Get the function body scope so we only do the replacement
+              // within the component function we're currently processing
+              functionScope = j(functionFirstParamNode.init.body);
+            } else {
+              console.warn(
+                `[WARNING]: Could parse component function first parameter "${componentName}"`
+              );
+            }
+          }
+        }
```
tdeekens, line 389: I can follow along. Took me a while to realize - thanks to the comments - that we need to handle different ways in which React.FCs can be written.

packages/codemod/src/transforms/react-default-props-migration.ts
```diff
@@ -0,0 +1,426 @@
...
+        const typeAnnotation = typePath.node.typeAnnotation;
+
+        if (typeAnnotation.type === 'TSTypeLiteral') {
+          typeAnnotation.members.forEach((member) => {
+            if (
+              member.type === 'TSPropertySignature' &&
+              member.key.type === 'Identifier' &&
+              destructuredKeys.includes(member.key.name)
+            ) {
+              member.optional = true;
```
ddouglasz, line 160: This is where we make the type member of the prop optional when it is part of the prop right ?

packages/codemod/src/transforms/react-default-props-migration.ts
```diff
@@ -0,0 +1,426 @@
...
+          j,
+          defaultPropsKeys: Object.keys(defaultPropsMap),
+          scope: functionScope!,
+        });
+
+        // 5. Update the component TS type definition so we make sure the default props are optional
+        updateComponentTypes({
+          j,
+          root,
+          typeName: `${componentName}Props`,
```
ddouglasz, line 402: This is also something I noticed, if we build the typeName from this, sometimes, there are cases where the typeName is inconsistent and we end up having files with typeNames as just `${componentName}` without the Props suffix. 
This also would have to be updated manually for cases like these.

packages/codemod/src/transforms/react-default-props-migration.ts
```diff
@@ -0,0 +1,426 @@
...
+        ...transformDefaultPropsToAST(defaultPropsMap, j),
+        j.spreadProperty(j.identifier('props')),
+      ]);
+
+      // Make sure the refactored parameter has the same type annotation
+      // as the original one
+      refactoredParameter.typeAnnotation = functionPropsParam.typeAnnotation;
+      break;
+    default:
+      console.warn(
```
ddouglasz, line 217: 🙌🏽

ddouglasz (approved):
Thank you, especially for the step by step comments 🙌🏽

packages/codemod/src/transforms/react-default-props-migration.ts
CarlosCortizasCT: Yes. I need to add another comment.

packages/codemod/src/transforms/react-default-props-migration.ts
CarlosCortizasCT: Yes, this relies on us using a consistent naming.
I think I can improve it further as a follow-up.

emmenko (commented):
Thanks @CarlosCortizasCT ! 🙌

Could you write some tests for this new codemod? See [test folder](https://github.com/commercetools/merchant-center-application-kit/tree/main/packages/codemod/test).

packages/codemod/src/transforms/react-default-props-migration.ts
emmenko: We can remove the suffix `Props` from the existing name to make sure we can append it. If the component name doesn't end up with `Props`, the `replace` doesn't do anything and we can still append the suffix correctly.

```suggestion
          typeName: `${componentName.replace(/(Props)$/), ''}Props`,
```

CarlosCortizasCT:
> Thanks @CarlosCortizasCT ! 🙌
> 
> Could you write some tests for this new codemod? See [test folder](https://github.com/commercetools/merchant-center-application-kit/tree/main/packages/codemod/test).

@emmenko that's a legit request but the problem is that `jscodeshift` test utils are synchronous and the new codemod is asynchronous so it does not work.
I could not find a simple way to overcome this challenge and I don't think it makes sense to spend time writing a different set of utilities.

If you have any suggestion on how to easily implement the tests, please let me know.

packages/codemod/src/transforms/react-default-props-migration.ts
CarlosCortizasCT: To be honest, I run this codemod on the 100+ component in the `ui-kit` repository and didn't find this problem so I would prefer to actually have a use case for changing it.

CarlosCortizasCT: I've just realized I was wrong. I had to adjust some things manually in `ui-kit` but didn't think it was because of this.
However, I would like to do it as a follow-up in order to not expand the scope of this PR.

emmenko: > I would like to do it as a follow-up in order to not expand the scope of this PR

Do you mean adding `.replace(/(Props)$/, '')`? Or are you referring to other things?

emmenko:
Ah ok, sorry I didn't know that it's an issue in the test if we use `async`.

Then it's fine as long as you are confident that it works as expected 🤗

packages/codemod/src/transforms/react-default-props-migration.ts
CarlosCortizasCT: I had a more elaborated idea in mind but I already implemented while I was waiting for the review.

Please take a look at this commit: 9af02f73a9e11464c810ba5834ff9c7409ea18f3

CarlosCortizasCT:
> Ah ok, sorry I didn't know that it's an issue in the test if we use `async`.
> 
> Then it's fine as long as you are confident that it works as expected 🤗

So far I'm doing manual testing to verify everything works as expected so I'm pretty confident.

emmenko (approved):
💯

PR #3682, opened 2024-12-20
[FEC-138] Improve js components migration

CarlosCortizasCT:
#### Summary

Improve js components migration

#### Description

I discovered another use case we were not covering: when we have a React component in a Javascript file and it does not have already destructured properties in the function declaration.

```jsx
import * as PropTypes from 'prop-types';

function MyComponent(props) {
  return <div>{props.foo}</div>;
}
MyComponent.defaultProps = {
  foo: 'bar',
};
MyComponent.propTypes = {
  foo: PropTypes.string,
};

```


mmaltsev-ct (approved):
LGTM

PR #3688, opened 2025-01-09
refactor: deprecate STS header

emmenko:
https://commercetools.atlassian.net/browse/SHIELD-1580

This is a follow up of some improvements we are doing around the security headers.

There is no need to override the `Strict-Transport-Security` header, so we are deprecating the config option.

### DIFF
--- a/packages/application-config/custom-application.schema.json
+++ b/packages/application-config/custom-application.schema.json
@@ -14,31 +14,31 @@
   },
   "properties": {
     "name": {
-      "description": "See https://docs.commercetools.com/merchant-center-customizations/api-reference/custom-application-config#name",
+      "description": "See https://docs.commercetools.com/merchant-center-customizations/tooling-and-configuration/custom-application-config#name",
       "type": "string"
     },
     "description": {
-      "description": "See https://docs.commercetools.com/merchant-center-customizations/api-reference/custom-application-config#description",
+      "description": "See https://docs.commercetools.com/merchant-center-customizations/tooling-and-configuration/custom-application-config#description",
       "type": "string"
     },
     "entryPointUriPath": {
-      "description": "See https://docs.commercetools.com/merchant-center-customizations/api-reference/custom-application-config#entrypointuripath",
+      "description": "See https://docs.commercetools.com/merchant-center-customizations/tooling-and-configuration/custom-application-config#entrypointuripath",
       "type": "string"
     },
     "cloudIdentifier": {
-      "description": "See https://docs.commercetools.com/merchant-center-customizations/api-reference/custom-application-config#cloudidentifier",
+      "description": "See https://docs.commercetools.com/merchant-center-customizations/tooling-and-configuration/custom-application-config#cloudidentifier",
       "type": "string"
     },
     "mcApiUrl": {
-      "description": "See https://docs.commercetools.com/merchant-center-customizations/api-reference/custom-application-config#mcapiurl",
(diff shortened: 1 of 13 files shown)


packages/application-config/custom-application.schema.json
```diff
@@ -174,34 +174,34 @@
...
         },
         "strictTransportSecurity": {
-          "description": "See https://docs.commercetools.com/merchant-center-customizations/api-reference/custom-application-config#headersstricttransportsecurity",
           "type": "array",
           "items": {
             "enum": ["includeSubDomains", "preload"]
           },
-          "uniqueItems": true
+          "uniqueItems": true,
+          "deprecated": true
```
emmenko, line 186: New: marking the field `strictTransportSecurity` as `deprecated`

packages/application-config/package.json
```diff
@@ -57,7 +57,7 @@
   "devDependencies": {
     "@commercetools-frontend/assets": "workspace:*",
     "@types/jsdom": "^21.1.2",
-    "json-schema-to-typescript": "13.1.2"
+    "json-schema-to-typescript": "15.0.3"
```
emmenko, line 60: Took the chance to also update the library

packages/application-config/test/fixtures/custom-applications/config-full.json
```diff
@@ -38,9 +38,8 @@
       "connect-src": ["https://track.avengers.app"]
     },
     "permissionsPolicies": {
-      "microphone": "()"
-    },
-    "strictTransportSecurity": ["includeSubDomains"]
+      "microphone": "*"
```
emmenko, line 41: Just to have it more explicit that it is a different value than the default one.

packages/constants/src/constants.ts
```diff
@@ -304,11 +304,13 @@ export const STORAGE_KEYS = {
 } as const;
 
 export const HTTP_SECURITY_HEADERS = {
+  'Referrer-Policy': 'same-origin',
+  'Permissions-Policy':
+    'microphone=(), camera=(), payment=(), usb=(), geolocation=()',
```
emmenko, line 309: This is added now, so that when parsing the headers from the config we retain the default values.
It also makes it more transparent what are the default values being used.

packages/mc-html-template/src/process-headers.spec.ts
```diff
@@ -92,7 +92,7 @@ describe('csp', () => {
     expect(
       processedApplicationConfig['Content-Security-Policy']
     ).toMatchInlineSnapshot(
-      `"default-src 'none'; script-src 'self' 'sha256-RACUtlMzZqF+zLJFAGOjaIZxjIlawfiKwgX9HWbQ9yQ=' 'sha256-5kzN9QxmTvcO/1/aYsVEo7hnWU50vwMjy6wviGxCOA0=' 'sha256-K4tyBnwnqF68wrXckWx1ce5+E4534Hv/ZdQEZLf+Z7Y='; connect-src 'self' app.launchdarkly.com clientstream.launchdarkly.com events.launchdarkly.com app.getsentry.com *.sentry.io https://example.com; img-src * data:; style-src 'self' fonts.googleapis.com data: 'unsafe-inline'; font-src 'self' fonts.gstatic.com data:; frame-src 'self'; upgrade-insecure-requests "`
```
emmenko, line 95: Uh, this needs to be fixed...

tdeekens (approved):

PR #3690, opened 2025-01-10
refactor: evaluate `window.app` lazily for defining user agents

emmenko:
This is helpful in case the package is not used within the Merchant Center context, in particular for SSR.


tdeekens (approved):

tylermorrisford (approved):
Thank you so much for this!

PR #3691, opened 2025-01-10
refactor(components): lazy load custom view selector in modular tab components

emmenko:
Main goals:
* decouple loading of the Custom View selector when tabular component is used outside of MC (evaluate if `window.app` is defined)
* implies also a decoupling of react-router (used by the custom view selector)

### DIFF
--- a/packages/application-components/src/components/custom-views/custom-views-selector/custom-views-selector.tsx
+++ b/packages/application-components/src/components/custom-views/custom-views-selector/custom-views-selector.tsx
@@ -15,22 +15,18 @@ import { designTokens } from '@commercetools-uikit/design-system';
 import { WindowEyeIcon } from '@commercetools-uikit/icons';
 import Spacings from '@commercetools-uikit/spacings';
 import Text from '@commercetools-uikit/text';
+import useCustomViewLocatorSelector from '../../../hooks/use-custom-view-locator-selector';
 import CustomViewLoader from '../custom-view-loader';
 import messages from './messages';
+import type { TCustomViewSelectorProps } from './types';
 import { useCustomViewsConnector } from './use-custom-views-connector';
 
 const COMPONENT_HEIGHT = '52px';
 
-type TCustomViewSelectorBaseProps = {
-  onCustomViewsResolved?: (customViews: CustomViewData[]) => void;
-};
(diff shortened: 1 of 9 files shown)


packages/application-components/src/components/custom-views/custom-views-selector/custom-views-selector.tsx
```diff
@@ -177,13 +173,17 @@ function CustomViewSelector(props: TCustomViewSelectorWithRequiredProps) {
 CustomViewSelector.displayName = 'CustomViewSelector';
 
 const CustomViewSelectorOrNothing = (props: TCustomViewSelectorProps) => {
-  if (!props.customViewLocatorCode) {
+  const { currentCustomViewLocatorCode } = useCustomViewLocatorSelector(
+    props.customViewLocatorCodes
+  );
```
emmenko, line 178: We are now using the hook here to match the router location

packages/application-components/src/components/custom-views/custom-views-selector/types.ts
```diff
@@ -0,0 +1,15 @@
+import type { CustomViewData } from '@commercetools-frontend/constants';
+
+export type TCustomViewLocatorCodeLocationDescriptor = {
+  pathname?: string;
+  search?: string;
+};
```
emmenko, line 6: Instead of using the type `LocationDescriptor` from `history`, we define our own similar type.

tylermorrisford (approved):

packages/application-components/src/components/custom-views/custom-views-selector/index.ts
```diff
@@ -1 +1,15 @@
...
+
+declare let window: ApplicationWindow;
+
+const CustomViewSelector = lazy(() => {
+  if (typeof window !== 'undefined' && typeof window.app !== 'undefined') {
+    return import(
+      './custom-views-selector' /* webpackChunkName: "custom-views-selector" */
+    );
+  }
+  return Promise.resolve({ default: () => null });
```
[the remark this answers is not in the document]
Rhotimee: 💯

Rhotimee (approved):
Nice. Thanks!

PR #3696, opened 2025-01-20
refactor: refine security headers and ensure they are applied for local development

emmenko:
Minor improvements as follow up of recent support findings.

### DIFF
--- a/packages/constants/src/constants.ts
+++ b/packages/constants/src/constants.ts
@@ -303,14 +303,25 @@ export const STORAGE_KEYS = {
   LOGIN_STRATEGY: 'loginStrategy',
 } as const;
 
+export const HTTP_SECURITY_HEADER_KEYS = {
+  'Content-Security-Policy': 'Content-Security-Policy',
+  'Referrer-Policy': 'Referrer-Policy',
+  'Permissions-Policy': 'Permissions-Policy',
+  'Strict-Transport-Security': 'Strict-Transport-Security',
+  'X-XSS-Protection': 'X-XSS-Protection',
+  'X-Content-Type-Options': 'X-Content-Type-Options',
+  'X-Frame-Options': 'X-Frame-Options',
+} as const;
+export type THttpSecurityHeaders = keyof typeof HTTP_SECURITY_HEADER_KEYS;
 export const HTTP_SECURITY_HEADERS = {
-  'Referrer-Policy': 'same-origin',
-  'Permissions-Policy':
-    'microphone=(), camera=(), payment=(), usb=(), geolocation=()',
-  'Strict-Transport-Security': 'max-age=31536000; includeSubDomains; preload',
-  'X-XSS-Protection': '1; mode=block',
-  'X-Content-Type-Options': 'nosniff',
-  'X-Frame-Options': 'SAMEORIGIN',
+  [HTTP_SECURITY_HEADER_KEYS['Referrer-Policy']]: 'same-origin',
+  [HTTP_SECURITY_HEADER_KEYS['Permissions-Policy']]:
+    'microphone=(self), camera=(self), payment=(self), usb=(self), geolocation=(self)',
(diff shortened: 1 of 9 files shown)


packages/constants/src/constants.ts
```diff
@@ -303,14 +303,25 @@ export const STORAGE_KEYS = {
...
-  'Referrer-Policy': 'same-origin',
-  'Permissions-Policy':
-    'microphone=(), camera=(), payment=(), usb=(), geolocation=()',
-  'Strict-Transport-Security': 'max-age=31536000; includeSubDomains; preload',
-  'X-XSS-Protection': '1; mode=block',
-  'X-Content-Type-Options': 'nosniff',
-  'X-Frame-Options': 'SAMEORIGIN',
+  [HTTP_SECURITY_HEADER_KEYS['Referrer-Policy']]: 'same-origin',
+  [HTTP_SECURITY_HEADER_KEYS['Permissions-Policy']]:
+    'microphone=(self), camera=(self), payment=(self), usb=(self), geolocation=(self)',
```
emmenko, line 319: Changing the defaults to use `(self)`.

For reference: this is important to ensure that Custom Views (rendered within an `iframe`) can correctly use and override the `Permissions-Policy` header.

packages/mc-html-template/src/process-headers.ts
```diff
@@ -11,8 +15,8 @@ type TDirective = Record<string, TDirectiveValue>;
 const toArray = (value: TDirectiveValue) =>
   Array.isArray(value) ? value : [value];
 
-const mergeDirectives = (...directives: TDirective[]) =>
-  directives.reduce(
+const parseCSPDirectives = (...directives: TDirective[]) => {
```
emmenko, line 18: Just renaming

tdeekens (approved):

packages/mc-html-template/src/process-headers.ts
```diff
@@ -28,24 +32,33 @@ const mergeDirectives = (...directives: TDirective[]) =>
       ),
     {}
   );
-const toHeaderString = (directives: TDirective = {}) =>
```
emmenko, line 31: This was a bit confusing to have it as a separate function, as it's only used for the CSP case.

packages/mc-html-template/src/process-headers.ts
```diff
@@ -28,24 +32,33 @@ const mergeDirectives = (...directives: TDirective[]) =>
...
-const toStructuredHeaderString = (directives: TDirective = {}) =>
-  Object.entries(directives)
-    .map(
-      ([directive, value]) =>
-        `${directive}=${Array.isArray(value) ? value.join(' ') : value}`
-    )
-    .join(', ');
+};
+
+const parsePermissionsPolicyDirectives = (
```
emmenko, line 44: This is new: it now merges the `permissions-policy` default directives with the ones provided in the customization config.
This mimics the "merging behavior" that we have implemented in the MC Proxy.

packages/mc-scripts/src/commands/start-vite.ts
```diff
@@ -39,6 +44,7 @@ async function run() {
     },
     server: {
       port: DEFAULT_PORT,
+      headers: compiledHeaders,
```
emmenko, line 47: Ensures that Vite dev server includes the security headers.

packages/constants/src/constants.ts
CarlosCortizasCT: Perhaps we can add this as a code comment for future reference

CarlosCortizasCT (approved):
Thanks! 👍

packages/constants/src/constants.ts
emmenko: Good idea, thanks! https://github.com/commercetools/merchant-center-application-kit/pull/3696/commits/d062758b7436ada8d2a2141be501712041118f8a

Rhotimee (approved):

PR #3676, opened 2024-12-10
[FEC-139] Refactor React components "defaultProps" management

CarlosCortizasCT:
#### Summary

Update the way we manage default properties in React components.

#### Description

As part of the preparations for the upcoming update to the newest React version, we need to update how we manage components default properties as our current implementation will no longer be supported ([reference](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-proptypes-and-defaultprops)).

```ts
// BEFORE
type TMyComponentProps = {
  message: string;
  size: string;
}

function MyComponent(props: TMyComponentProps) {
 ...
}

MyComponent.defaultProps = {
  size: 'big'
}


// AFTER
type TMyComponentProps = {
  message: string;
  size?: string; // <--- Note this property is now defined as optional
}

function MyComponent({ size = 'big', ...props }: TMyComponentProps) {
 ...
}
```

PR #3700, opened 2025-01-21
fix(sdk): regression for default props

emmenko:
Regression from #3676

### DIFF
--- a/packages/sdk/src/components/sdk-get/sdk-get.tsx
+++ b/packages/sdk/src/components/sdk-get/sdk-get.tsx
@@ -28,8 +28,8 @@ type DispatchProps = {
 };
 type OwnProps = {
   actionCreator: (...args: TActionCreatorArgs) => TSdkAction;
-  actionCreatorArgs: TActionCreatorArgs;
-  shouldRefetch: (
+  actionCreatorArgs?: TActionCreatorArgs;
+  shouldRefetch?: (
     prevArgs: TActionCreatorArgs,
     nextArgs: TActionCreatorArgs
   ) => boolean;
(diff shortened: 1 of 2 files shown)


packages/sdk/src/components/sdk-get/sdk-get.tsx
```diff
@@ -28,8 +28,8 @@ type DispatchProps = {
 };
 type OwnProps = {
   actionCreator: (...args: TActionCreatorArgs) => TSdkAction;
-  actionCreatorArgs: TActionCreatorArgs;
-  shouldRefetch: (
+  actionCreatorArgs?: TActionCreatorArgs;
+  shouldRefetch?: (
```
emmenko, line 32: These props need to be optional

packages/sdk/src/components/sdk-get/sdk-get.tsx
```diff
@@ -54,11 +62,6 @@ export class SdkGet extends Component<Props, State> {
     requestsInFlight: 0,
     result: undefined,
     error: undefined,
-    actionCreatorArgs: [],
-    shouldRefetch: (
-      prevArgs: TActionCreatorArgs,
-      nextArgs: TActionCreatorArgs
-    ) => !deepEqual(prevArgs, nextArgs),
```
emmenko, line 61: This is the component's `state`, not the default props.

packages/sdk/src/components/sdk-get/sdk-get.tsx
```diff
@@ -40,6 +40,14 @@ type OwnProps = {
 export type Props = DispatchProps & OwnProps;
 type StaticErrorHandler = (error: TSdkError) => void;
 
+const defaultProps: Required<
+  Pick<Props, 'actionCreatorArgs' | 'shouldRefetch'>
+> = {
+  actionCreatorArgs: [],
+  shouldRefetch: (prevArgs: TActionCreatorArgs, nextArgs: TActionCreatorArgs) =>
+    !deepEqual(prevArgs, nextArgs),
+};
```
emmenko, line 49: Just creating an object with the default props (nothing to do with React's default props).

packages/sdk/src/components/sdk-get/sdk-get.tsx
```diff
@@ -89,19 +92,22 @@ export class SdkGet extends Component<Props, State> {
...
   componentDidUpdate(prevProps: Props) {
+    const shouldRefetch =
+      this.props.shouldRefetch ?? defaultProps.shouldRefetch;
     if (
-      this.props.shouldRefetch(
-        prevProps.actionCreatorArgs,
-        this.props.actionCreatorArgs
+      shouldRefetch(
+        prevProps.actionCreatorArgs ?? defaultProps.actionCreatorArgs,
+        this.props.actionCreatorArgs ?? defaultProps.actionCreatorArgs
```
emmenko, line 100: Simply falling back to the value defined in the `defaultProps` object.

PR #3705, opened 2025-01-22
FEC-220 - Update file output of build assets

valoriecarli:
As part of the migration from Webpack to Vite, we will be updating the Vite configuration to output all build assets directly under the `public` folder (as opposed to the default nested assets directory).

This approach ensures compatibility with our existing scripts and avoids requiring updates across all consuming applications


emmenko (approved):

CarlosCortizasCT (approved):
Thanks a lot 👍

PR #3672, opened 2024-12-09
feat: add `babel-plugin-formatjs` to `babel-preset-mc-app`

ragafus:
#### Summary

The goal of this PR is to add `babel-plugin-formatjs` to the `babel-preset-mc-app` and `mc-scripts` packages in order to avoid bloating bundles with useless data from `formatjs` messages (`description`, `defaultMessage` props).

#### Description
A couple of flags are added (see https://formatjs.github.io/docs/tooling/babel-plugin/#options):
* `i18nAst`: Pre-parse `defaultMessage` into AST for faster runtime perf
* `i18nRemoveDefaultMessage`: Remove `defaultMessage` field in generated js after extraction

PR #3707, opened 2025-01-24
fix(babel-preset): make formatjs plugin opt-in via env variable

emmenko:
As we are bumping into some unexpected (but also legit) issues due to the plugin `babel-plugin-formatjs` (introduced in #3672), we are now making the plugin opt-in and only explicitly enabled via env variable `ENABLE_BABEL_PLUGIN_FORMATJS`.


CarlosCortizasCT:
Perhaps I'm lacking some context but in the original PR I see the plugin is added to the `webpack` and `vite` commands directly so I was wondering if the current change is enough.

packages/mc-scripts/package.json
```diff
@@ -70,6 +72,7 @@
     "@vitejs/plugin-react-swc": "3.7.0",
     "autoprefixer": "^10.4.15",
     "babel-loader": "8.3.0",
+    "babel-plugin-formatjs": "^10.5.25",
```
emmenko, line 75: In Vite we are not using our custom babel preset but the `@vitejs/plugin-react`. These deps (used by the react babel plugin) should be listed explicitly.

Rhotimee (approved):

CarlosCortizasCT (approved):
LGTM 👍 

Thanks a lot for working on this so quickly. 🙇

PR #3710, opened 2025-01-28
refactor(application-components): use radix-ui components for modals and dialogs

kark:
[mc-fe branch preview](https://mc-15822.mc-preview.europe-west1.gcp.escemo.com/)

One of the vrts keeps failing, but I can't reproduce this locally:
Screenshot 2025-01-29 at 09 09 39
Here’s what I see locally:
screencapture-localhost-3000-info-modal-page-info-modal-nested-2025-01-28-12_57_11

This is the [html artifact](https://percy.io/api/v1/snapshots/2083774303/assets/head.html?) from the Percy build.
This confirms that all z-indexes of all overlays are applied correctly, which makes Percy’s output even more confusing.

### DIFF
--- a/packages/application-components/src/components/detail-pages/tabular-detail-page/tabular-detail-page.spec.tsx
+++ b/packages/application-components/src/components/detail-pages/tabular-detail-page/tabular-detail-page.spec.tsx
@@ -85,21 +85,12 @@ describe('rendering', () => {
     screen.getByText(/custom/i);
   });
   it('should warn if no title nor custom title row are provided', () => {
-    const warnSpy = jest.spyOn(console, 'warn').mockImplementation(() => {});
     renderTabularDetailPage();
 
     expect(warning).toHaveBeenCalledWith(
       false,
       'TabularDetailPage: one of either `title` or `customTitleRow` is required but both their values are `undefined`'
     );
-
-    expect(warnSpy).toHaveBeenCalledWith(
-      expect.stringContaining(
-        "Warning: Invalid prop 'title' supplied to 'TextHeadline'. Expected it to be nonempty string, but it was empty."
-      )
-    );
-
-    warnSpy.mockRestore();
   });
 });
 describe('navigation', () => {
--- a/packages/application-components/src/components/detail-pages/tabular-detail-page/tabular-detail-page.spec.tsx
+++ b/packages/application-components/src/components/detail-pages/tabular-detail-page/tabular-detail-page.spec.tsx
@@ -85,12 +85,21 @@ describe('rendering', () => {
     screen.getByText(/custom/i);
   });
   it('should warn if no title nor custom title row are provided', () => {
+    const warnSpy = jest.spyOn(console, 'warn').mockImplementation(() => {});
     renderTabularDetailPage();
 
     expect(warning).toHaveBeenCalledWith(
       false,
       'TabularDetailPage: one of either `title` or `customTitleRow` is required but both their values are `undefined`'
     );
+
+    expect(warnSpy).toHaveBeenCalledWith(
+      expect.stringContaining(
+        "Warning: Invalid prop 'title' supplied to 'TextHeadline'. Expected it to be nonempty string, but it was empty."
+      )
+    );
+
+    warnSpy.mockRestore();
   });
 });
 describe('navigation', () => {
--- a/packages/application-components/src/components/dialogs/internals/dialog-header.tsx
+++ b/packages/application-components/src/components/dialogs/internals/dialog-header.tsx
@@ -1,5 +1,6 @@
 import type { ReactNode, SyntheticEvent } from 'react';
 import { css } from '@emotion/react';
+import { Title as DialogTitle } from '@radix-ui/react-dialog';
 import { designTokens as uiKitDesignTokens } from '@commercetools-uikit/design-system';
 import { CloseIcon } from '@commercetools-uikit/icons';
 import SecondaryIconButton from '@commercetools-uikit/secondary-icon-button';
@@ -21,13 +22,28 @@ export const TextTitle = (props: TTextTitleProps) => (
   </Text.Headline>
 );
 
+const HiddenEmptyDialogTitle = () => (
+  <div aria-hidden={true} style={{ display: 'none' }}>
+    <DialogTitle />
+  </div>
+);
+
 type TitleProps = Pick<Props, 'title'>;
 const Title = (props: TitleProps) => {
-  if (typeof props.title === 'string') {
-    return <TextTitle title={props.title} />;
-  } else {
-    return <>{props.title}</>;
-  }
+  return (
+    <>
+      {typeof props.title === 'string' ? (
+        <TextTitle title={props.title} />
+      ) : (
+        props.title
+      )}
+      {/* FIXME: Temporary workaround for https://github.com/radix-ui/primitives/issues/2986
+        Radix UI's DialogContent requires rendering a DialogTitle, which renders as <h2>.
+        To meet this requirement and avoid rendering two heading elements with the title in the DOM (<TextTitle> renders as <h3>), we are hiding the DialogTitle.
+      */}
+      <HiddenEmptyDialogTitle />
+    </>
+  );
 };
 
 const DialogHeader = (props: Props) => (
@@ -56,6 +72,7 @@ const DialogHeader = (props: Props) => (
     </Spacings.Inline>
   </div>
 );
+
 DialogHeader.displayName = 'DialogHeader';
 
 export default DialogHeader;
--- a/packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
+++ b/packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
@@ -1,12 +1,9 @@
-import { css, type SerializedStyles } from '@emotion/react';
+import { keyframes, css } from '@emotion/react';
+import styled from '@emotion/styled';
+import { Content } from '@radix-ui/react-dialog';
 import { PORTALS_CONTAINER_INDENTATION_SIZE } from '@commercetools-frontend/constants';
-import { customProperties } from '@commercetools-uikit/design-system';
+import { designTokens } from '@commercetools-uikit/design-system';
 import type { TModalPageSize } from './modal-page';
-
-type StyleProps = {
-  zIndex?: number;
-  size: TModalPageSize;
-};
 
 export const stylesBySize = {
   '10': {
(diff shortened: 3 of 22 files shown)


kark:
[preview_deployment]

kark:
[preview_deployment]

packages/application-components/src/components/detail-pages/tabular-detail-page/tabular-detail-page.spec.tsx
```diff
@@ -85,12 +85,21 @@ describe('rendering', () => {
...
       'TabularDetailPage: one of either `title` or `customTitleRow` is required but both their values are `undefined`'
     );
+
+    expect(warnSpy).toHaveBeenCalledWith(
+      expect.stringContaining(
+        "Warning: Invalid prop 'title' supplied to 'TextHeadline'. Expected it to be nonempty string, but it was empty."
+      )
+    );
+
+    warnSpy.mockRestore();
```
kark, line 102: This console.warn call caused multiple tests to fail in CI. It comes from the UI Kit's `TextHeadline` component, which is expected since we're not passing a title. If you'd like me to suppress these warnings without asserting them, let me know.

packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
```diff
@@ -27,65 +24,94 @@ export const stylesBySize = {
...
+  flex-direction: column;
+  background-color: ${designTokens.colorSurface};
+  box-shadow: 0px 0px 40px 0px rgba(0, 0, 0, 0.1);
+  outline: 0;
+  z-index: inherit;
+
+  &[data-state='open'] {
+    animation: ${({ size }) => getContentShowAnimation(size)} 300ms ease
+      forwards;
+  }
```
kark, line 111: The `data-state` attribute is added to the `Dialog.Content` component rendered in the DOM. Animations are currently keyed off this attribute.

packages/application-components/src/components/modal-pages/internals/modal-page.tsx
```diff
@@ -125,57 +117,69 @@ const ModalPage = ({
...
+  const [portalContainer, setPortalContainer] = useState<HTMLElement | null>(
+    null
+  );
+
+  useEffect(() => {
+    const container = getParentSelector();
+    if (container) {
+      setPortalContainer(container);
+    }
+  }, []);
```
kark, line 130: I mentioned the other day that when the URL is refreshed, the order of modal rendering is reversed. This change was needed (the use of local state and useEffect to manage the portals container) to make it work

packages/application-components/src/components/portals-container/portals-container.tsx
```diff
@@ -162,41 +159,42 @@ const PortalsContainer = forwardRef<TLayoutRefs, TPortalsContainerProps>(
...
+
     return (
       <>
         <Global
-          // Apply some global styles, based on the `.ReactModal__Body--open` class.
           styles={[
-            css`
-              .ReactModal__Body--open
+            isModalOpen &&
+              css`
```
kark, line 169: Previously, the `.ReactModal__Body--open` class was added to the root element. Now, global styles are keyed off the stacking layers count.

website-components-playground/src/components/playground-controller.tsx
```diff
@@ -55,18 +52,6 @@ const PlaygroundController = (props: TPlaygroundControllerProps) => {
...
-                    // preview container instead of full screen.
-                    // This is only needed for the playground previews, therefore we
-                    // override the styles here.
-                    styles={css`
-                      .ReactModal__Body--open #${PORTALS_CONTAINER_ID} {
-                        position: absolute;
-                        overflow: auto;
-                      }
-                    `}
-                  />
```
kark, line 69: I’m removing this because I couldn’t figure out why it was needed and haven’t noticed any regressions.

packages/application-shell-connectors/src/components/application-context/application-context.tsx
```diff
@@ -250,6 +250,8 @@ function withApplicationContext<
           const mappedProps = mapApplicationContextToProps
             ? mapApplicationContextToProps(applicationContext)
             : { applicationContext };
+          // @ts-ignore FIXME: this is an issue with HOC typing while using Emotion's JSX namespace
+          // See: https://github.com/emotion-js/emotion/issues/3245, https://github.com/emotion-js/emotion/issues/3261
```
kark, line 254: This is unrelated but has started causing the type check to fail.
Had to add it in a bunch of places.

packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
```diff
@@ -27,65 +24,94 @@ export const stylesBySize = {
...
+  0% {
+    opacity: 0;
+  }
+  100% {
+    opacity: 1;
+  }
 `;
 
-export const getOverlayStyles = (props: StyleProps): SerializedStyles => css`
+// we can't use Overlay from Radix-ui because it doesn't appear in DOM when Dialog.Root is rendered with `modal={false}`
```
CarlosCortizasCT, line 36: Just trying to better understand the issue: why do we need to use `modal={false}`?

packages/application-components/src/components/detail-pages/tabular-detail-page/tabular-detail-page.spec.tsx
CarlosCortizasCT: Just wondering if we could provide a value for that prop 🤔

packages/application-components/src/components/dialogs/internals/dialog-header.tsx
```diff
@@ -21,13 +22,28 @@ export const TextTitle = (props: TTextTitleProps) => (
...
-  }
+  return (
+    <>
+      {typeof props.title === 'string' ? (
+        <TextTitle title={props.title} />
+      ) : (
+        props.title
+      )}
+      {/* FIXME: Temporary workaround for https://github.com/radix-ui/primitives/issues/2986
+        Radix UI's DialogContent requires rendering a DialogTitle, which renders as <h2>.
```
CarlosCortizasCT, line 41: What does it happen if we don't provide a `DialogTitle`? Does the component throw an error? a warning? is it only in development mode?

packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
kark: Thanks Carlos!
Otherwise, when the modal is open, `aria-hidden` is applied to all other elements on the page, preventing the use of the project selector, language selector, navbar, and profile context menu.

packages/application-components/src/components/detail-pages/tabular-detail-page/tabular-detail-page.spec.tsx
kark: But we're testing here `should warn if no title nor custom title row are provided`

packages/application-components/src/components/dialogs/internals/dialog-header.tsx
kark: Thanks, 
We're getting a console error in the browser and tests in CI fail with
image

> is it only in development mode?

Yes

packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
CarlosCortizasCT: Oh! I see. Thanks for the explanation 👍

packages/application-components/src/components/dialogs/internals/dialog-header.tsx
CarlosCortizasCT: What I was thinking is that, since this won't happen in production, we would be injecting an element in the DOM that we don't need and we're also adding this extra complexity to the code. Perhaps we can just [globally silence](https://github.com/commercetools/merchant-center-application-kit/blob/main/packages/jest-preset-mc-app/setup-test-framework.js#L87) that error in our tests as we know where it's coming from and that it's safe.

kark: Sure, we could suppress it here, that's a valid option.
But it would still fail in the mc-fe CI, right? So we'd need to silence it in multiple places.

Rhotimee (approved):
Nice Work @kark. Looks great!

kark:
[preview_deployment]

packages/application-components/src/components/dialogs/internals/dialog-header.tsx
kark: I pushed this change https://github.com/commercetools/merchant-center-application-kit/pull/3710/commits/789577efc86a5348fbfe0339a4ff41966d8f7224 (not 100% if this is what you meant), but it seems to me it's causing [fails in the CI downstream](https://app.circleci.com/pipelines/github/commercetools/merchant-center-frontend/99515/workflows/215b34f8-e4e1-49a5-bb48-686fff24ddde/jobs/2352459).
Please mind that we'll need to add additional setup for the app using vitest (it uses our dialogs too).

CarlosCortizasCT: Ok. I was just trying to know if we could avoid that extra HTML element but seems it's not that easy so I think it's better to keep it like you have it before the last commit.
(_Perhaps is something to look as a follow-up_)

CarlosCortizasCT (approved):
Thanks a lot for working on this one, Kacper 🙇

packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
```diff
@@ -27,65 +24,94 @@ export const stylesBySize = {
...
+})<{ size: TModalPageSize }>`
+  position: absolute;
+  top: 0;
+  right: 0;
+  height: 100%;
+  width: ${({ size }) =>
+    size !== 'scale'
+      ? // In case we're using a specific size, we want it to be used until there's no space left.
+        // In such scenario, we want the modal to be as wide as possible, but using the shared indentation width size.
+        `min(${stylesBySize[size].width}, calc(100% - ${PORTALS_CONTAINER_INDENTATION_SIZE})) !important`
```
emmenko, line 99: I think the `!mportant` is causing the width of the stacked layers to be incorrect (see bottom right of the screenshot)

image

Or we need to add an `!important` to the stacking layer override styles.

packages/application-components/src/components/modal-pages/internals/modal-page.tsx
```diff
@@ -125,57 +117,69 @@ const ModalPage = ({
...
+            onInteractOutside={(e) => {
+              // Prevent only the modal from closing, handle all other events
+              const overlay = document.querySelector(
+                '[data-role="modal-overlay-clickable"]'
+              );
+              if (!overlay?.contains(e.target as Node)) {
+                e.preventDefault();
+              }
+            }}
+            aria-labelledby={props.title}
```
emmenko, line 161: Nit: the `aria-labelledby` should reference another element. We should remove that, as we already have `aria-label`.

emmenko (commented):
Awesome work, many thanks! It's pretty cool that we can keep the same interface of the components and change the internals.

There is one issue that I noticed in the preview URL, but otherwise it looks good! 💯

packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
kark: Thanks, great catch!

kark: Actually, this `!important` is crucial to apply the correct width of the drawer component. 
I just checked, and since we're using a drawer there, this width is expected.
[staging](https://mc.europe-west1.gcp.escemo.com/adnan-nodejs-test-project/settings/project/taxes/new) vs [preview](https://mc-15822.mc-preview.europe-west1.gcp.escemo.com/adnan-nodejs-test-project/settings/project/taxes/new)

emmenko: OMG 🤦‍♂️

Thanks for double checking. It was unexpected for me that the `<Drawer>` is used there...nevermind then 🤗

packages/application-components/src/components/modal-pages/internals/modal-page.tsx
```diff
@@ -125,57 +117,69 @@ const ModalPage = ({
...
+            onInteractOutside={(e) => {
+              // Prevent only the modal from closing, handle all other events
+              const overlay = document.querySelector(
+                '[data-role="modal-overlay-clickable"]'
+              );
+              if (!overlay?.contains(e.target as Node)) {
+                e.preventDefault();
+              }
+            }}
+            aria-labelledby=""
```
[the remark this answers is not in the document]
kark: I left a comment https://github.com/commercetools/merchant-center-application-kit/pull/3710#discussion_r1940822006
If we remove it, `aria-labelledby` from the underlying radix-ui component takes precedence and our tests fail: 

> TestingLibraryElementError: Unable to find an accessible element with the role "dialog" and name "Custom aria title"
>  Here are the accessible roles:
    
    
Screenshot 2025-02-04 at 10 48 40

PR #3711, opened 2025-01-29
fix: updated public-page-layout to send props correctly

mustafaasif2:
Regression of #3676 

#### Summary

The login screen card was compressed and and moved to the left

#### Description

The props were not being passed in `PublicPageLayoutContent` due to which `ContainerColumn` was being applied instead of `ContainerColumnWide`. This PR fixes the issues by ensuring that `contentScale` is passed correctly

Existing layout: 

Screenshot 2025-01-29 at 16 51 46

### DIFF
--- a/.changeset/brown-singers-build.md
+++ b/.changeset/brown-singers-build.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-components': minor
+---
+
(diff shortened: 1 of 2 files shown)


tdeekens (approved):

emmenko (approved):

kark (approved):
💯

CarlosCortizasCT (approved):
Many thanks! 👍

PR #3712, opened 2025-01-29
Update preview releases NPM dist-tag every time we release a new RC version

CarlosCortizasCT:
#### Summary

Update preview releases NPM dist-tag every time we release a new RC version

#### Description

Out current [GH job](https://github.com/commercetools/merchant-center-application-kit/blob/main/.github/workflows/preview-release-on-comment.yml) to release preview RC candidates to NPM publishes release candidates versions to NPM and also creates a `dist-tag` the first time and links it to the very first released RC.

There's an issue though because, when we release a new RC, the `dist-tag` is not updated to point to the new RC version. This mean we can't get automatically the most recent version in the consumers, forcing us to go to each one and update the version manually.

In this PR we include a script that we use in the GH job so we make sure the `dist-tag` is updated every time we publish a new RC version. This means that consumers can link to it by `dist-tag` (eg: `fec-156-react19`) rather than by version (eg: `0.0.0-fec-156-react19-20250128154337`).

### DIFF
--- a/scripts/update-npm-tag.mjs
+++ b/scripts/update-npm-tag.mjs
@@ -40,8 +40,8 @@ async function run() {
 
   // Update the dist-tag of all the public packages to the latest release
   for (const packageName of packagesNames) {
-    // exec(`npm dist-tag add ${packageName}@${latestVersion} ${distTag}`);
-    const result = shell.exec(`sleep 1 && echo "Updating NPM dist: ${packageName}@${latestVersion} ${distTag}"`);
+    // const result = shell.exec(`sleep 1 && echo "Updating NPM dist: ${packageName}@${latestVersion} ${distTag}"`);
+    const result = shell.exec(`npm dist-tag add ${packageName}@${latestVersion} ${distTag}`);
     if (result.code !== 0) {
       throw new Error(`Error updating NPM dist-tag "${distTag}" for ${packageName} with version "${latestVersion}"`);
     }
--- a/scripts/update-npm-tag.mjs
+++ b/scripts/update-npm-tag.mjs
@@ -35,11 +35,12 @@ async function run() {
   // Resolve the public packages of the repository
   const packagesNames = getPublicPackagesNames();
 
-  // Fetch the latest release of the received dist-tag
-  const latestVersion = await getLatestReleaseCandidate(packagesNames[0], distTag);
-
   // Update the dist-tag of all the public packages to the latest release
   for (const packageName of packagesNames) {
+    // Fetch the latest release of the received dist-tag for this package
+    const latestVersion = await getLatestReleaseCandidate(packageName, distTag);
+
+    // Update the dist-tag of all the public packages to the latest release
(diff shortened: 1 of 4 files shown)


scripts/update-npm-tag.mjs
```diff
@@ -0,0 +1,53 @@
...
+    .filter((pkg) => !pkg.private)
+    .map((pkg) => pkg.name);
+}
+
+async function getLatestReleaseCandidate(packageName, distTag) {
+  const response = await fetch(`https://registry.npmjs.org/${packageName}`);
+  const packageDetails = await response.json();
+
+  const filteredVersions = Object.entries(packageDetails.time)
+    .filter(([version]) => version.includes(distTag))
```
tdeekens, line 16: Nit: includes or startsWith? Just wondering if being more precise would help 🤷🏼

scripts/update-npm-tag.mjs
```diff
@@ -0,0 +1,53 @@
+import shell from 'shelljs';
+import { getPackageInfos } from "workspace-tools";
+
+function getPublicPackagesNames() {
+  const packages = getPackageInfos(import.meta.resolve('..'));
```
tdeekens, line 5: Just for me to follow: what does this return? A list of packages in the repo?

scripts/update-npm-tag.mjs
```diff
@@ -0,0 +1,53 @@
...
+
+  if (!distTag) {
+    throw new Error("Please provide a NPM dist-tag as a parameter");
+  }
+
+  // Resolve the public packages of the repository
+  const packagesNames = getPublicPackagesNames();
+
+  // Fetch the latest release of the received dist-tag
+  const latestVersion = await getLatestReleaseCandidate(packagesNames[0], distTag);
```
tdeekens, line 39: This is just to know what the latest version of all packages is, right?

tdeekens (commented):
Nice. I think I can follow. Just some clarification questions 🤗 .

scripts/update-npm-tag.mjs
CarlosCortizasCT: Yes, this returns an array with the parsed `package.json` files of all this repository packages.

scripts/update-npm-tag.mjs
CarlosCortizasCT: Yes. The idea is that we have just released a new version this script is not aware of and we need to find about it.
This works because all packages share the same version.

scripts/update-npm-tag.mjs
tdeekens: Me neither. Just wanted to ask.

scripts/update-npm-tag.mjs
CarlosCortizasCT: @tdeekens Your comment got me thinking and I've discovered that not all repositories use the same version for all their public packages so I've updated the latest version resolution to be per package: e344e40e917d0b3711f0b31897d033088b7e11bd

ragafus (approved):

PR #3716, opened 2025-01-31
refactor(scripts): no need for extra dep

emmenko:
Follow up of #3712 

There is no need to install an extra dependency `workspace-tools` as we already have `@manypkg/get-packages`.

I also removed the `run` function wrapper as Node supports top-level `await`.


scripts/update-npm-tag.mjs
```diff
@@ -1,11 +1,11 @@
...
 import shell from 'shelljs';
-import { getPackageInfos } from "workspace-tools";
 
-function getPublicPackagesNames() {
-  const packages = getPackageInfos(import.meta.resolve('..'));
-  return Object.values(packages)
-    .filter((pkg) => !pkg.private)
-    .map((pkg) => pkg.name);
+async function getPublicPackagesNames() {
+  const { packages } = await getPackages(process.cwd());
```
CarlosCortizasCT, line 5: **nit**: This makes the script work only from the root folder of the repository. With the previous implementation, it can be used from the local directory of from the root one.

CarlosCortizasCT (approved):
Many thanks 🙇

scripts/update-npm-tag.mjs
emmenko: It works from any folder actually. Internally `getPackages` uses `find-root`, so it always starts looking for packages from the root of the repository.

tdeekens (approved):
Works for me. Thanks!

PR #3720, opened 2025-02-03
fix: pnpm v9 regression with linking workspace packages

emmenko:
Regression after upgrading to PNPM v9 (#3695).

The error occurs when bumping the version in one of the workspace packages which hasn't been published to the NPM registry.

image

In [v9](https://github.com/orgs/pnpm/discussions/7932) they changed the following setting:

image

Therefore, we are configuring the setting `link-workspace-packages=true` to restore the previous behavior.

Since PNPM config relies on `.npmrc` file, we also need to commit that. The release CI jobs have also been amended to "inject" the NPM auth token.

PS: using the `workspace` protocol is actually our intention as well. However, there were some issues with Changesets where packages would be published with the `workspace` protocol in the dependencies.
Maybe we need to revisit this setup and see if we can make it work again.

### DIFF
--- a/.github/workflows/preview-release-on-comment.yml
+++ b/.github/workflows/preview-release-on-comment.yml
@@ -48,13 +48,8 @@ jobs:
       - name: Installing dependencies and building packages
         uses: ./.github/actions/ci
 
-      - name: Creating .npmrc
-        run: |
-          cat << EOF > "$HOME/.npmrc"
-            provenance=true
-            email=npmjs@commercetools.com
-            //registry.npmjs.org/:_authToken=$NPM_TOKEN
-          EOF
+      - name: Setting up authorization to NPM registry (.npmrc)
+        run: echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > .npmrc
         env:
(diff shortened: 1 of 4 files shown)


kark (approved):

Rhotimee (approved):

.github/workflows/preview-release-on-comment.yml
```diff
@@ -48,13 +48,8 @@ jobs:
...
         uses: ./.github/actions/ci
 
-      - name: Creating .npmrc
-        run: |
-          cat << EOF > "$HOME/.npmrc"
-            provenance=true
-            email=npmjs@commercetools.com
-            //registry.npmjs.org/:_authToken=$NPM_TOKEN
-          EOF
+      - name: Setting up authorization to NPM registry (.npmrc)
```
tdeekens, line 51: That provenance us removed is intentional?

.npmrc
```diff
@@ -0,0 +1,6 @@
+# NPM configuration
+provenance=true
```
tdeekens, line 2: Ah. Disregard the above.

tdeekens (approved):

PR #3722, opened 2025-02-05
fix: compat with emotion `v11.14`

emmenko:
Same as https://github.com/commercetools/ui-kit/pull/3054

TL;DR: we are switching the `moduleResolution` to `bundler`.

### DIFF
--- a/packages-backend/express/src/middlewares/fixtures/jwt-token.ts
+++ b/packages-backend/express/src/middlewares/fixtures/jwt-token.ts
@@ -1,3 +1,4 @@
+// @ts-expect-error: migrate to jose v5
 import { JWT, JWK, JWKS } from 'jose';
 
 const keyRS256 = JWK.generateSync('RSA', 2048, { use: 'sig', alg: 'RS256' });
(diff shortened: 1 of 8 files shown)


tsconfig.json
```diff
@@ -10,5 +10,15 @@
...
+  // Old "moduleResolution": "Node" option required for Cypress
+  // https://github.com/cypress-io/cypress/issues/26308#issuecomment-1663592648
+  //
+  // TODO: Remove when upgrading Cypress to > v12.
+  "ts-node": {
+    "compilerOptions": {
+      "module": "ESNext",
+      "moduleResolution": "nodenext"
+    }
+  }
```
emmenko, line 23: See https://github.com/cypress-io/cypress/issues/26308

tdeekens (commented):
Good idea. It's not affecting and bundled files however, right? As there preconstruct is continuously used.

emmenko:
Yes preconstruct is used as before. I haven't noticed any issues with the bundles.

Rhotimee (approved):
🙌🏽

PR #3726, opened 2025-02-12
fix: modals and dialogs with default attributes added by radix-ui

kark:
#### Summary

Hoping to fix error happening in CI: 
image


emmenko (approved):
Thanks, let's give it a try!

Rhotimee (approved):

PR #3734, opened 2025-03-07
refactor: adjust auth routes for local dev

emmenko:
Identity-related changes


packages/mc-dev-authentication/src/create-mc-dev-authentication-middleware.ts
```diff
@@ -39,37 +30,21 @@ function createMcDevAuthenticationMiddleware(
...
-            next();
-          } else {
-            response.end(htmlLogin);
+          if (!isDevAuthenticationMiddlewareDisabled) {
+            // Redirect to the MC API to initiate the authorize flow.
+            const redirectTo = new URL(
+              request.originalUrl,
+              trimTrailingSlash(applicationConfig.env.mcApiUrl)
+            );
+            response.writeHead(301, { Location: redirectTo.toString() }).end();
```
emmenko, line 44: This is the main change, we simply redirect to the `/login/authorize` endpoint of the MC API.

packages/mc-dev-authentication/src/create-mc-dev-authentication-middleware.ts
```diff
@@ -39,37 +30,21 @@ function createMcDevAuthenticationMiddleware(
...
-      }
-      if (request.originalUrl === '/logout') {
-        logout(response);
-
-        if (isDevAuthenticationMiddlewareDisabled) {
-          next();
-        } else {
-          response.end(htmlLogout);
         }
-        return;
```
emmenko, line 72: I think this was the "legacy" login and can be removed.

packages/mc-dev-authentication/src/transformer-local.ts
```diff
@@ -1,36 +1,5 @@
...
-  const htmlLogout = pages.logoutPage;
-
-  fs.writeFileSync(path.join(paths.appBuild, 'login.html'), htmlLogin, 'utf8');
-  fs.writeFileSync(
-    path.join(paths.appBuild, 'logout.html'),
-    htmlLogout,
-    'utf8'
-  );
+const transformerLocal = () => {
+  // noop
```
emmenko, line 2: For backwards compatibility we keep the transformer module but keep it as a no-op.

packages/mc-scripts/src/commands/serve.ts
```diff
@@ -1,11 +1,27 @@
...
+      if (request.url?.startsWith('/login/authorize')) {
+        const redirectTo = new URL(request.url, applicationConfig.env.mcApiUrl);
+        response.writeHead(301, { Location: redirectTo.toString() }).end();
+        return;
+      }
+      if (request.url?.startsWith('/logout')) {
+        response.end('Please clear your session storage.');
+        return;
+      }
+    }
```
emmenko, line 23: Similarly, this is to handle the case of developing against local MC services when serving the production bundles.

packages/mc-scripts/src/commands/serve.ts
```diff
@@ -1,11 +1,27 @@
...
   const server = http.createServer((request, response) => {
+    // When developing against local APIs, handle the login|logout routes separately.
+    if (applicationConfig.env.mcApiUrl.startsWith('http://localhost')) {
+      if (request.url?.startsWith('/login/authorize')) {
+        const redirectTo = new URL(request.url, applicationConfig.env.mcApiUrl);
+        response.writeHead(301, { Location: redirectTo.toString() }).end();
+        return;
+      }
+      if (request.url?.startsWith('/logout')) {
+        response.end('Please clear your session storage.');
```
emmenko, line 20: To keep it simple, if someone wants to "log out" they simply need to clear the session storage. Again, given that this is only needed internally when developing MC services we don't need to overcomplicate the implementation.

kark (approved):

PR #3738, opened 2025-03-12
Update console messages handling in tests

CarlosCortizasCT:
#### Summary

Update console messages handling in tests.

#### Description

We've updated how `console` messages are handled in local environments: previously, `console.warn` and `console.error` messages did not make tests to fail locally but they did in CI; this behaviour led to several cases where we did not realize about an issue until some code reached CI whereas it could be detected locally.

With this change we expect some errors to be caught earlier in the development lifecycle and thus shorten the time to fix them.

### DIFF
--- a/.changeset/thirty-crabs-mix.md
+++ b/.changeset/thirty-crabs-mix.md
@@ -1,0 +1,7 @@
+---
+'@commercetools-frontend/jest-preset-mc-app': minor
+---
+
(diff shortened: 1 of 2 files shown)


emmenko (approved):
Thanks

CarlosCortizasCT:
@emmenko I can't merge this PR because of the flaky VRT in the playground.

Could you please merge it when you have some time? 🙏

PR #3740, opened 2025-03-12
refactor: simplify jwks validation with jose, no extra libs

emmenko:
* Uses `jose` (ESM version)
* Because of ESM, we need to instruct Jest to transpile it correctly
  * Uses `babel-jest` by default
  * Separated Jest config for "node" related tests

### DIFF
--- a/packages-backend/express/src/auth.ts
+++ b/packages-backend/express/src/auth.ts
@@ -1,8 +1,4 @@
-import {
-  expressjwt as expressJwtMiddleware,
-  type GetVerificationKey,
-} from 'express-jwt';
-import jwksRsa from 'jwks-rsa';
+import { createRemoteJWKSet, jwtVerify } from 'jose';
 import {
   CLOUD_IDENTIFIERS,
   MC_API_URLS,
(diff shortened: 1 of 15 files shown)


packages-backend/express/src/auth.ts
```diff
@@ -185,37 +189,25 @@ function createSessionAuthVerifier<Request extends TBaseRequest>(
...
+
+    const exchangeToken = authorizationHeader.replace(/^Bearer (.*)$/, '$1');
+    const jwksClient = getJwksClientByIssuer(issuer);
+    const verifiedToken = await jwtVerify<TDecodedJWT>(
+      exchangeToken,
+      jwksClient,
+      {
+        algorithms: ['RS256'],
         audience,
         issuer,
```
emmenko, line 205: This is the new part that replaces the old implementation.

packages-backend/express/src/auth.ts
```diff
@@ -139,6 +128,21 @@ export const getConfiguredAudience = <Request extends TBaseRequest>(
...
+  const client = jwksClientByIssuer.get(issuer);
+  if (client) {
+    return client;
+  }
+  const newClient = createRemoteJWKSet(
+    new URL(`/.well-known/jwks.json`, issuer)
+  );
+  jwksClientByIssuer.set(issuer, newClient);
+  return newClient;
+}
```
emmenko, line 144: This is needed as there might be different issuers being used, as the issuer can also be determined dynamically by the middleware.
Therefore, we "cache" each instance of `createRemoteJWKSet` for each issuer.

packages-backend/express/src/middlewares/fixtures/jwt-token.ts
```diff
@@ -1,20 +1,109 @@
-// @ts-expect-error: migrate to jose v5
```
emmenko, line 1: ✅

kark (approved):
This is an amazing improvement 💯
Much more direct, readable and maintainable

VictorAvelar (approved):

PR #3749, opened 2025-03-25
fix: parsing of CLI arguments

emmenko:
Leftovers of #3739

### DIFF
--- a/packages/mc-scripts/src/cli.ts
+++ b/packages/mc-scripts/src/cli.ts
@@ -18,7 +18,7 @@ program
   .description('CLI to develop and build Merchant Center customizations.')
   .version(pkgJson.version)
   .option(
-    '--env [path]',
+    '--env <path...>', // Variadic option. It allows multiple `--env` options to be provided.
     `(optional) Parses the file path as a dotenv file and adds the variables to the environment. Multiple flags are allowed.`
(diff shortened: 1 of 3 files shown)


packages/mc-scripts/src/cli.ts
```diff
@@ -18,7 +18,7 @@ program
   .description('CLI to develop and build Merchant Center customizations.')
   .version(pkgJson.version)
   .option(
-    '--env [path]',
+    '--env <path...>', // Variadic option. It allows multiple `--env` options to be provided.
```
emmenko, line 21: This is the actual fix (the `...`) was missing.

packages/mc-scripts/src/cli.ts
```diff
@@ -112,7 +112,7 @@ async function run() {
       'Compiles "index.html.template" file into a "index.html" with all the required runtime configuration. The security headers are also compiled and injected into the "index.html".'
     )
     .option(
-      '--transformer [path]',
+      '--transformer <path>',
```
emmenko, line 115: Minor change to being more strict as these option arguments are supposed to be required.

CarlosCortizasCT (approved):

kark (approved):

PR #3753, opened 2025-03-25
refactor(cli): add experimental auth flow for login command

emmenko:
As we are working on identity (internal project for now), we want to test out the new auth flow from the CLI.

```
ENABLE_EXPERIMENTAL_IDENTITY_AUTH_FLOW=true pnpm mc-scripts login
```


packages/mc-scripts/src/commands/login.ts
```diff
@@ -1,12 +1,57 @@
...
+    });
+    request.on('error', (error) => {
+      reject(error);
+    });
+  });
+};
 
 async function run() {
+  const shouldUseExperimentalIdentityAuthFlow =
+    process.env.ENABLE_EXPERIMENTAL_IDENTITY_AUTH_FLOW === 'true';
```
emmenko, line 53: Using an env variable for opting in.

packages/mc-scripts/src/commands/login.ts
```diff
@@ -18,27 +63,111 @@ async function run() {
...
-    name: 'password',
-    message: 'Password (hidden)',
-  });
+    const state = generateRandomHash();
+    const nonce = generateRandomHash();
 
-  if (!email || !password) {
-    throw new Error(`Missing email or password values. Aborting.`);
-  }
+    const authUrl = new URL('/login/authorize', mcApiUrl);
```
emmenko, line 72: Using the new auth endpoint in the MC API

packages/mc-scripts/src/commands/login.ts
```diff
@@ -18,27 +63,111 @@ async function run() {
...
+    authUrl.searchParams.set('response_type', 'id_token');
+    authUrl.searchParams.set('response_mode', 'form_post');
+    authUrl.searchParams.set('client_id', '__local:@@cli@@');
```
emmenko, line 75: Normally we use that to send the application identifier. The prefix `__local` is actually the important part, as we determine the callback URL (e.g. localhost vs mc hostname). This way we are "backwards compatible" with the OIDC flow from MC apps.

packages/mc-scripts/src/commands/login.ts
```diff
@@ -18,27 +63,111 @@ async function run() {
...
+    const server = http.createServer(async (request, response) => {
+      try {
+        if (
+          request.url?.includes('/oidc/callback') &&
+          request.method === 'POST' &&
+          request.headers['content-type'] ===
+            'application/x-www-form-urlencoded'
+        ) {
+          const body = await parseRequestBody(request);
+          const authParams = decode<AuthorizeCallbackFragments>(body);
```
emmenko, line 95: The request body contains the token (because of `response_mode: form_post`), so we parse it. The `decode` function simply transforms the data from url-form-encoded to JSON.

packages/mc-scripts/src/commands/login.ts
```diff
@@ -18,27 +63,111 @@ async function run() {
...
+              'Invalid authentication flow (missing sessionToken)'
+            );
+          }
+          const decodedSessionToken = jwtDecode<SessionToken>(sessionToken);
 
-  const credentials = await getAuthToken(mcApiUrl, { email, password });
-  credentialsStorage.setToken(mcApiUrl, credentials);
+          if (decodedSessionToken?.nonce !== nonce) {
+            throw new Error('Invalid authentication flow (nonce mismatch)');
+          }
```
emmenko, line 107: We decode the token to inspect the `nonce` and validate it.

packages/mc-scripts/src/commands/login.ts
```diff
@@ -18,27 +63,111 @@ async function run() {
...
+            token: sessionToken,
+            expiresAt: decodedSessionToken.exp,
+          });
+
+          console.log(chalk.green(`Login successful.`));
+          console.log();
+
+          response.writeHead(200, 'Success!', { 'content-type': 'text/html' });
+          response.end();
+          server.close();
```
emmenko, line 122: Important steps:
* Respond to the API request
* Close the callback server

packages/mc-scripts/src/utils/graphql-requests.ts
```diff
@@ -112,18 +115,26 @@ async function requestWithTokenRetry<Data, QueryVariables extends Variables>(
...
 ): Promise<Data> {
+  const shouldUseExperimentalIdentityAuthFlow =
+    process.env.ENABLE_EXPERIMENTAL_IDENTITY_AUTH_FLOW === 'true';
+
   const token = credentialsStorage.getToken(requestOptions.mcApiUrl);
 
+  const tokenHeader: Record<string, string | null> =
+    shouldUseExperimentalIdentityAuthFlow
+      ? { [SUPPORTED_HEADERS.AUTHORIZATION]: `Bearer ${token}` }
+      : { 'x-mc-cli-access-token': token };
```
emmenko, line 130: The session token now is treated as a "normal token" (like when developing MC apps locally), so we can send it via `authorization` header.

The special header `x-mc-cli-access-token` won't be needed anymore. This will simplify some parts of the MC API around that.
For the time being though, the logic will stay for backwards compatibility.

packages/mc-scripts/src/commands/login.ts
```diff
@@ -18,27 +63,111 @@ async function run() {
...
-  }
+    const authUrl = new URL('/login/authorize', mcApiUrl);
+    authUrl.searchParams.set('response_type', 'id_token');
+    authUrl.searchParams.set('response_mode', 'form_post');
+    authUrl.searchParams.set('client_id', '__local:@@cli@@');
+    authUrl.searchParams.set(
+      'scope',
+      [
+        'openid',
+        // 'project_key:??',
```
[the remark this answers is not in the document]
YahiaElTai: ah thanks a lot! 🚀

packages/mc-scripts/src/commands/login.ts
```diff
@@ -1,12 +1,39 @@
...
 import prompts from 'prompts';
 import { processConfig } from '@commercetools-frontend/application-config';
+import pkgJson from '../../package.json';
 import { getAuthToken } from '../utils/auth';
+import { createAuthCallbackServer } from '../utils/auth-callback';
 import CredentialsStorage from '../utils/credentials-storage';
 
 const credentialsStorage = new CredentialsStorage();
+const port = 3001;
+const clientIdentifier = `mc-scripts-${pkgJson.version}`;
```
emmenko, line 14: To have a more unique identifier of the request.

packages/mc-scripts/src/commands/login.ts
```diff
@@ -18,27 +45,67 @@ async function run() {
...
-  });
+    const state = generateRandomHash();
+    const nonce = generateRandomHash();
 
-  if (!email || !password) {
-    throw new Error(`Missing email or password values. Aborting.`);
-  }
+    const authUrl = new URL('/login/authorize', mcApiUrl);
+    authUrl.searchParams.set('response_type', 'id_token');
+    authUrl.searchParams.set('response_mode', 'query');
```
emmenko, line 56: Using the `query` response mode

kark (approved):
👍

PR #3764, opened 2025-05-02
feat: add an e2e test for notifications inside Custom Views

mustafaasif2:
#### Summary

Adds end-to-end test to verify notification functionality in Custom Views.

#### Description

There were no tests verifying whether notifications render correctly within Custom Views. This gap led to a bug reported by a client. This PR introduces an end-to-end test that ensures proper notification behaviour.

#### Approach

There were four possible solutions considered:

1. Unit Testing CustomViewLoader with useShowNotification

The useShowNotification hook is currently mocked in [custom-view-loader.spec.tsx](https://github.com/commercetools/merchant-center-application-kit/blob/03ef95a3347664717b4b8a51c0f034a552f928bf/packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.spec.tsx#L17C1-L20C5) file. Replacing the mock with the real hook would require setting up Redux in the test environment. While we have a `renderAppWithRedux` helper in the application-shell package, importing it in the application-components package (where CustomViewLoader resides) would cause a circular dependency. Reimplementing the Redux setup in application-components would be overly complex for simply testing notifications.

2. Modifying the Custom View Starter Template

Another option was to modify the starter template to ensure it renders notifications, but this would involve changing a base template used by all custom views, which we want to potentially avoid.

3. Testing Notifications in the Playground Using Jest and React Testing Library

Another idea was to test notifications by rendering them inside a Custom View within the API Playground and verifying their behaviour using Jest and React Testing Library. However, since Custom Views in the Playground are rendered within an `<iframe>`, this approach is not feasible. Jest does not provide full support for interacting with content inside iframes. As a result, it’s not possible to reliably access or assert on elements rendered within the iframe, making this setup unsuitable for unit or integration testing of notification behaviour.

4. Add End-to-End Tests in the Playground (Chosen Approach)

Ultimately, I opted for an end-to-end (e2e) test to verify that notifications appear correctly in Custom Views. The test renders two buttons - one to trigger a success notification and another to trigger an error notification. This approach avoids architectural complexity, circumvents the limitations of unit testing inside iframes, and does not require any changes to the starter templates.

Note: We can discuss whether maintaining this e2e test is worthwhile or if we prefer to close this PR without merging. 

Updated Demo Custom View in the Playground (One button could also have been enough) 

Screenshot 2025-05-02 at 12 44 07


cypress/support/urls.ts
```diff
@@ -5,7 +5,7 @@ export const URL_BASE = `/${projectKey}`;
 export const ENTRY_POINT_APP_KIT_PLAYGROUND = 'app-kit-playground';
 export const URL_APP_KIT_PLAYGROUND = `${URL_BASE}/${ENTRY_POINT_APP_KIT_PLAYGROUND}`;
 export const URL_APP_KIT_PLAYGROUND_NOTIFICATIONS = `${URL_APP_KIT_PLAYGROUND}/notifications`;
-export const URL_APP_KIT_PLAYGROUND_STATE_MACHINES_ID = `${URL_APP_KIT_PLAYGROUND}/12ad40eb-b33f-4e0e-ae91-bca373ccfd58`;
+export const URL_APP_KIT_PLAYGROUND_STATE_MACHINES_ID = `${URL_APP_KIT_PLAYGROUND}/3309e9d0-fb15-4056-afa7-dc12f31ea0ad`;
```
mustafaasif2, line 8: This is the updated ID for the 'Initial' key.

Screenshot 2025-05-02 at 12 41 44

emmenko (commented):
Thanks, I think this is a good approach! 🙌

cypress/e2e/playground/custom-view.cy.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    cy.get('iframe[title="Custom View: Demo Custom View"]')
+      .its('0.contentDocument.body')
+      .should('not.be.empty')
+      .then(cy.wrap)
+      .within(() => {
+        cy.findByRole('button', { name: /Show Success Notification/i }).click();
+        cy.findByText(/Operation completed successfully!/i).should('exist');
+      });
+  });
```
emmenko, line 29: Have you tried also taking a Percy snapshot `cy.percySnapshot();` at the end of the test?

emmenko (approved):
💯

tdeekens (approved):
💯 thanks!

kark (approved):
Awesome 🙌 
So happy to see it worked out!

cypress/e2e/playground/custom-view.cy.ts
mustafaasif2: Yes, good idea. I have updated it in this commit: 64e130dac10aa409bd266e41f082ce34c44198a3

PR #3770, opened 2025-05-13
Bw/svgr plugin fun

tylermorrisford:
#### Summary

Jira [reference](https://commercetools.atlassian.net/jira/software/c/projects/FCT/boards/362?selectedIssue=FCT-1473)

#### Description

🚨 🚧  Work and testing in progress 🚧 🚨


application-templates/starter/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "A starter example for the bare minimum setup to develop a Merchant Center Application",
   "private": true,
   "scripts": {
-    "build": "mc-scripts build",
+    "build": "ENABLE_EXPERIMENTAL_VITE_BUNDLER=true NODE_ENV=production mc-scripts build",
```
tdeekens, line 7: This will enable Vite by default for Custom Applications. I am not against it but Shield should share what they think. Plus this has differences in our built files are laid out in terms of folder structure to what I know which can be a breaking change for how Custom Application builders upload if I am not mistaken.

packages/mc-scripts/src/cli.ts
```diff
@@ -93,6 +93,7 @@ async function run() {
       if (shouldUseExperimentalBundler) {
         console.log('Experimental Vite bundler enabled! 🚀');
         console.log('');
+        console.log('super special edition');
```
tdeekens, line 96: 👋

packages/mc-scripts/src/commands/build-vite.ts
```diff
@@ -15,6 +15,7 @@ import pluginI18nMessageCompilation from '../vite-plugins/vite-plugin-i18n-messa
 import pluginSvgr from '../vite-plugins/vite-plugin-svgr';
 
 async function run() {
+  const CleanBuild = (await import('vite-plugin-clean-build')).default;
```
tdeekens, line 18: Why do we import this here?

packages/mc-scripts/src/vite-plugins/vite-plugin-svgr.ts
```diff
@@ -66,6 +117,47 @@ function vitePluginSvgr(): Plugin {
                         .digest('hex')}`,
                   },
                 },
+                // inline plugin to compress embedded pngs
```
tdeekens, line 120: ✌🏼

application-templates/starter/.env
```diff
@@ -1,2 +1,4 @@
 ENABLE_NEW_JSX_TRANSFORM="true"
 FAST_REFRESH="true"
+ENABLE_EXPERIMENTAL_VITE_BUNDLER="true"
+ENABLE_EXPERIMENTAL_VITE_BUNDLER_SOURCEMAP="true"
```
emmenko, line 4: We don't want those enabled by default in the starter templates. Please remove.

emmenko (commented):
Thanks for the contribution but I feel like I'm missing context on why we need to do that and why are we choosing this approach.

Maybe a sync would help to align and clarify the confusion.

Thanks!

application-templates/starter/package.json
tylermorrisford: I'll admit, I flailed a bit while learning how to test - this was part of that :) Removed this in [694c548](https://github.com/commercetools/merchant-center-application-kit/pull/3770/commits/694c5484231e6007e45b215813943f34312c801b)

packages/mc-scripts/src/vite-plugins/vite-plugin-svgr.ts
tylermorrisford: removed this in [694c548](https://github.com/commercetools/merchant-center-application-kit/pull/3770/commits/694c5484231e6007e45b215813943f34312c801b)

application-templates/starter-typescript/src/components/welcome/web-developer.svg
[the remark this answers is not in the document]
tylermorrisford: This was a file we used to test - removed with [694c548](https://github.com/commercetools/merchant-center-application-kit/pull/3770/commits/694c5484231e6007e45b215813943f34312c801b)

application-templates/starter/.env
tylermorrisford: Removed with [694c548](https://github.com/commercetools/merchant-center-application-kit/pull/3770/commits/694c5484231e6007e45b215813943f34312c801b)

packages/mc-scripts/src/vite-plugins/vite-plugin-svgr.ts
```diff
@@ -16,8 +16,59 @@ function vitePluginSvgr(): Plugin {
...
+
+          for (const match of matchesIterator) {
+            const originalBase64 = match[1];
+
+            if (originalBase64) {
+              // This length would indicate that the match is greater than ~1mb
+              if (originalBase64.length > 1000000) {
+                console.warn(
+                  '\n🚨 You have a large embedded png in your svg - consider using an image tag instead 🚨'
+                );
```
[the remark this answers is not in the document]
ByronDWall: This is in response to an incident.  For more context see [this jira ticket](https://commercetools.atlassian.net/browse/FCT-1473).  

Ideally we would not be embedding png's into svg, but figma does this automatically when exporting svgs that include a png as a child, so it can and has happened.  This is meant as an automatic guardrail to make sure that any embedded svgs do not cause large amounts of bloat in our package size, to insure that we do not trigger another incident like the one that prompted this work.  I know that audit log also has some svg's with embedded png, along with the fallback application.

packages/mc-scripts/src/commands/build-vite.ts
```diff
@@ -77,7 +78,9 @@ async function run() {
         return { runtime: `window.__toCdnUrl(${JSON.stringify(filename)})` };
       },
     },
+    clearScreen: false,
```
ByronDWall, line 81: This was helpful for debugging, but I think we should remove this for widespread release

packages/mc-scripts/src/vite-plugins/vite-plugin-svgr.ts
```diff
@@ -16,8 +16,59 @@ function vitePluginSvgr(): Plugin {
         const { transform } = await import('@svgr/core');
         const svgCode = await fs.promises.readFile(id, 'utf8');
 
+        const optimizeEmbeddedPngs = async (code: string) => {
+          const pngRegex =
+            /xlink:href="data:image\/png;base64,([a-zA-Z0-9+/]+=*)"/gs;
```
ByronDWall, line 21: modern svg's use `href` [directly](https://css-tricks.com/on-xlinkhref-being-deprecated-in-svg/) instead of `xlink:href`, so it might make sense to remove the `xlink` from the matcher

packages/mc-scripts/src/vite-plugins/vite-plugin-svgr.ts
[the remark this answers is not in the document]
emmenko: Thanks for the feedback. I'm just raising this concern because the implementation to "handle" this seems rather complex.

> figma does this automatically when exporting svgs that include a png as a child

The question is: do we need to have SVGs with embedded PNGs? Like, is this a "design requirement"?
If not, maybe we can focus on having the guardrail as a "CI job" or something.

packages/mc-scripts/src/commands/build-vite.ts
tylermorrisford: According to it's docs, one should be able to import it like a normal ESM but I was not able to locally, and encountered that a few times with this work.

packages/mc-scripts/src/commands/build-vite.ts
tylermorrisford: Good catch - removed in [8e389a2](https://github.com/commercetools/merchant-center-application-kit/pull/3770/commits/8e389a280384ad56dcd87cfffccc5ef27ab2a272)

packages/mc-scripts/src/vite-plugins/vite-plugin-svgr.ts
tylermorrisford: Updated ([8e389a2](https://github.com/commercetools/merchant-center-application-kit/pull/3770/commits/8e389a280384ad56dcd87cfffccc5ef27ab2a272)) - I imagine that would be backwards compatible by default to only match from `href` on.

packages/mc-scripts/src/vite-plugins/vite-plugin-svgr.ts
```diff
@@ -16,8 +16,59 @@ function vitePluginSvgr(): Plugin {
...
+          const matchesIterator = code.matchAll(pngRegex);
+
+          for (const match of matchesIterator) {
+            const originalBase64 = match[1];
+
+            if (originalBase64) {
+              // This length would indicate  that the match is greater than ~1mb
+              if (originalBase64.length > 1000000) {
+                console.warn(
+                  '\n🚨 You have a large embedded png in your svg - consider using an image tag instead 🚨'
```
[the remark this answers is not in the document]
tylermorrisford: Added this in [8e389a2](https://github.com/commercetools/merchant-center-application-kit/pull/3770/commits/8e389a280384ad56dcd87cfffccc5ef27ab2a272) - I trimmed the id to just show the file name instead of the entire path, maybe we *should* show the entire path? not sure.

packages/mc-scripts/src/vite-plugins/vite-plugin-svgr.ts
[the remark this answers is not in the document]
tylermorrisford: I wouldn't say that having PNGs embedded into SVGs is specifically a design requirement, But I do think that unless developers go line-by-line through any svg assets they take from figma(unlikely), another incident would likely occur for the same reason. Taking the Welcome page as an example, there are a lot of images that are used there between the carousel component and other onboarding elements.

application-templates/starter-typescript/src/components/welcome/web-developer.svg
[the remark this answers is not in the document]
tylermorrisford: Ahh - I've replaced it, it's needed.

tylermorrisford:
I'm unsure how to resolve the CI issues - It appears the test processes are timing out while trying to install the dependencies; I cherry-picked commits to a different [PR](https://github.com/commercetools/merchant-center-application-kit/pull/3773) that had the same issues. During both rebases I took the incoming changes for the lockfile. Any ideas?

.github/actions/ci/action.yml
```diff
@@ -23,6 +23,8 @@ runs:
         # https://github.com/bahmutov/cypress-gh-action-split-install/blob/ca3916d4e7240ebdc337825d2d78eb354855464b/.github/workflows/tests.yml#L14-L18
         # https://github.com/marketplace/actions/cypress-io#custom-install
         CYPRESS_INSTALL_BINARY: '0'
+        # https://github.com/puppeteer/puppeteer/issues/12833
+        PUPPETEER_DOWNLOAD_BASE_URL: 'https://storage.googleapis.com/chrome-for-testing-public'
```
emmenko, line 27: FYI: extracted this into a separate PR https://github.com/commercetools/merchant-center-application-kit/pull/3776

packages/mc-scripts/src/vite-plugins/vite-plugin-svgr.ts
```diff
@@ -16,8 +16,59 @@ function vitePluginSvgr(): Plugin {
...
+                    originalBase64
+                  );
+                }
+              }
+            }
+          }
+          return modifiedCode;
+        };
+
+        const optimizedSVGCode = await optimizeEmbeddedPngs(svgCode);
```
emmenko, line 69: For the sake of mitigating the issue from triggering another incident, I'm fine to proceed with this under the following prerequisites:

* This behavior is enabled via an opt-in flag (e.g. `ENABLE_EXPERIMENTAL_SVG_PNG_OPTIMIZATION`).
* We consider this a workaround and thus a tech debt. Reducing the size of the PNG does not guarantee that the issue won't happen again. A more sustainable solution / process / approach would need to be researched and proposed.

Thanks 🤗

tylermorrisford: @emmenko How is this [1018a28](https://github.com/commercetools/merchant-center-application-kit/pull/3770/commits/1018a287ca41dda2d1013c0d1e392006ec2f7bdb)? 
We could add a cmd in a package.json
`"build:optimize-svg": "mc-scripts build ENABLE_EXPERIMENTAL_SVG_PNG_OPTIMIZATION=true",`

emmenko:
Also provide a changeset please

packages/mc-scripts/src/commands/build-vite.ts
```diff
@@ -103,7 +105,13 @@ async function run() {
...
         },
       }),
-      pluginSvgr(),
+      // cleanbuild is needed to remove large svg assets
+      // that are not used after the build.
+      CleanBuild({
+        outputDir: paths.appBuild,
+        patterns: ['*.react-*.svg'],
+        verbose: true, // logs the files that are removed
+      }),
```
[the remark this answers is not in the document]
tylermorrisford: Sure - in the case we catch a gigantic file, CleanBuild removes the file from the resulting bundle, since it is un-used(the image would be compiled into the the destination, so we don't need to upload the original file).

packages/mc-scripts/src/vite-plugins/vite-plugin-svgr.ts
```diff
@@ -16,8 +17,17 @@ function vitePluginSvgr(): Plugin {
         const { transform } = await import('@svgr/core');
         const svgCode = await fs.promises.readFile(id, 'utf8');
 
+        let transformableSVGCode;
+        if (process.env.ENABLE_EXPERIMENTAL_SVG_PNG_OPTIMIZATION) {
```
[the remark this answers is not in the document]
tylermorrisford: TIL! fixed in [079593d](https://github.com/commercetools/merchant-center-application-kit/pull/3770/commits/079593da9e858b201a88f9332f044305d9afd4d7)

PR #3771, opened 2025-05-14
feat: update compatibility of mc-create-app to work with yarn 3 and 4

mustafaasif2:
#### Summary

Fix compatibility issues with Yarn 3 and Yarn 4 in App Kit starter templates by configuring `nodeLinker` to use `node-modules`.

#### Description

Custom Applications created using the App Kit starter templates currently only work reliably with Yarn 1. Users running Yarn 3 or Yarn 4 encounter various issues during local development, such as missing dependencies in IDEs after running yarn install and errors like “We could not find this Project” in the UI. This PR adds a `.yarnrc.yml` file with `nodeLinker: "node-modules"` to the starter templates. This restores the traditional `node_modules/` behavior, ensuring compatibility. This change ensures a smoother developer experience regardless of the globally installed Yarn version.

### DIFF
--- a/packages/create-mc-app/src/utils.ts
+++ b/packages/create-mc-app/src/utils.ts
@@ -26,6 +26,30 @@ const shouldUseYarn = () => {
   }
 };
 
+const getYarnVersion = () => {
+  try {
+    const result = execa.commandSync('yarn --version', { encoding: 'utf-8' });
+    return result.stdout.trim();
+  } catch (error) {
+    return null;
+  }
+};
+
+const configureYarn = async (projectDirectoryPath: string) => {
+  const yarnVersion = getYarnVersion();
+  if (!yarnVersion) return;
+
+  // Check if Yarn version is 3 or higher
+  const majorVersion = parseInt(yarnVersion.split('.')[0], 10);
+  if (majorVersion >= 3) {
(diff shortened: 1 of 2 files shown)


kark (approved):
Thanks Mustafa 🎉
I suppose this proposal maintains backward compatibility and I'm personally ok with not using PnP and falling back to the traditional `node_modules` structure.
I checked that it works with yarn 1, and the ` .yarnrc.yml` config worked in yarn 3 and yarn 4.9.1.
All good from my end!

packages/create-mc-app/src/utils.ts
```diff
@@ -26,6 +27,33 @@ const shouldUseYarn = () => {
...
+  }
+};
+
+const configureYarn = (projectDirectoryPath: string) => {
+  const yarnVersion = getYarnVersion();
+  if (!yarnVersion) return;
+
+  // Check if Yarn version is 2 or higher. Plug'n'Play was introduced in Yarn v2
+  const majorVersion = parseInt(yarnVersion.split('.')[0], 10);
+  if (majorVersion >= 2) {
```
emmenko, line 45: Nit: use `semver` package (already installed) to check for the version range?

emmenko (commented):
Thanks for looking into that! Please also provide a Changeset in order to release this.

PR #3781, opened 2025-05-30
fix: r19 migration

kark:


kark:
[preview_deployment]

kark:
[preview_deployment]

packages/application-shell/src/components/navbar/navbar.tsx
```diff
@@ -125,58 +125,87 @@ export const ApplicationMenu = (props: ApplicationMenuProps) => {
...
+      }
+
+      // Clamp Y to safe area bounds (0% to 100%)
+      nextY = Math.min(Math.max(nextY, 0), 100);
+      if (!Number.isFinite(nextY)) {
+        nextY = 0;
+      }
+
+      setPercentageX(nextX);
+      setPercentageY(nextY);
```
kark, line 189: This is to ensure we don't end up with values above 100% and evaluations like `Infinity % Infinity %` in case of division by 0

emmenko (approved):
Thanks!

I honestly have no idea what all these calculations do but I trust you tested that it works as expected!

kark:
> I honestly have no idea what all these calculations do but I trust you tested that it work as expected!

Sure, if I break it I'll fix it 😅

tdeekens (approved):
Tested visually and works as expected. Generally, I dislike the idea of hovering menus and that should eventually be revisted. I don't see anybody else (as in SaaS) doing this these days. Mostly using very cleanly nested sub menus which expand and collapse. 

Just to share some examples

Screenshot 2025-06-05 at 09 26 03@2x

Screenshot 2025-06-05 at 09 25 48@2x

I think we should revisit the decision on why we have submenus opening on hover with design at some point.

kark:
Yes, that's a fair conclusion.
> that should eventually be revisted

I'll make sure we have this as a backlog entry before handing out ownership

PR #3784, opened 2025-06-02
[WIP] Attempt to use polling to continuously attempt to initialise iFrame of Custom View

tdeekens:
⚠️ Please disregard unless contacting me. I couldn't stop thinking about this and wanted to explore an idea and I need CI to run ⚠️

### DIFF
--- a/packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
+++ b/packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
@@ -53,7 +53,7 @@ function useIFrameInitializationEffect({
   iFrameRef,
   onInitialized,
   maxAttempts = 10,
-  pollInterval = 100,
+  pollInterval = 500,
 }: {
   iFrameRef: React.RefObject<HTMLIFrameElement | null>;
   onInitialized: () => void;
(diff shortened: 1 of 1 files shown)


packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -49,6 +49,87 @@ const CustomPanelIframe = styled.iframe`
   border: none;
 `;
 
+function useIFrameInitializationEffect({
```
tdeekens, line 52: The idea was to use a polling mechanism to check for an element - in this case loading spinner - as a ready indicator for receiving messages inside the iframe. 

The polling would start upon onLoad and finish once the message was sent.

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -49,6 +49,87 @@ const CustomPanelIframe = styled.iframe`
...
+        );
+
+        return;
+      }
+
+      if (appLoaderElement) {
+        console.log(
+          '[CustomViewLoader] app-loader found, sending initialization messages.'
+        );
+        setTimeout(onInitialized, pollInterval * 5);
```
tdeekens, line 106: Apparently we still need to delay sending the message. 

Assumption being that while onLoad indicates JS execution it doesn't indicate rendering and component mounting. Basically meaning the code inside the iFrame isn't ready yet.

tdeekens (commented):
Might be too much complexity given benefit being low. The issue is that the iFrame should send a message up that it's ready or give a visual clue we could poll for. Given we can not change that it's hard to find better solutions than timeouts.

PR #3787, opened 2025-06-04
fix(application-shell): add Suspense wrapper for IconSwitcher to avoid blocking apps loading

ragafus:
#### Summary
MC app LCP has increased after React 19 upgrade. I found that when loading app we are hit by new Suspense throttling behavior described in https://github.com/facebook/react/issues/31819

#### Description

This is a screenshot from MC Discounts. When `ListWithSearchIcon` is loaded the whole app is blocked for 300ms. By adding a Suspense boundary when rendering `IconSwitcher` we can avoid blocking the whole app loading.

Screenshot 2025-06-04 at 11 08 28

### DIFF
--- a/.changeset/rotten-dots-care.md
+++ b/.changeset/rotten-dots-care.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': patch
+---
+
+Adds Suspense boundary when rendering IconSwitcher to avoid blocking the whole app loading. Minimizes LCP increase after React 19 upgrade due to https://github.com/facebook/react/issues/31819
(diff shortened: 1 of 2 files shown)


tdeekens:
Thanks for always writing thoughts down so well. That's my understanding too and I should have written what you did. 

What I wondered additionally if we can reduce that 300ms hit but I don't see that it's possible cause it can not be configured. Saying that cause 300ms for an icon now seems excessive but I am obviously wrong cause Sebastian Markbåge was involved.

ragafus:
If there is not disagreement, I'm going to close this PR in favor of further investigations.

PR #3789, opened 2025-06-04
Add backwards compatible bidirectional initialization of Custom Views

tdeekens:
#### Summary

This improves Custom View initialization by a backwards compatible bidirectional communication in which the Custom View (managed by `custom-view-shell`) informs the host application (managed by `custom-view-loader`).

#### Description

This is a follow-up from the React v19 migration in which we initially tried to implement a bidirectional approach but reverted it as we _can not assume_ Custom Views to be updated. As a result this communication needs to be backwards compatible. Hence, even if the Custom View does not send a "ready message" the host has to assume - in this case after 500ms - for the Custom View to be ready. 

##### Changes made

1. Add a ready message sent from Custom View to host
2. Adjust the host to be able to react to the ready message to send initialization message and payload
3. Host to remain backwards compatible to send initialization message after 500ms if Custom View has not sent ready message
4. Expand on host to keep track on whether initialization messages have been sent to avoid sending them multiple times
    - In StrictMode, which isn't supported yet, this may need to reset while also the channel can likely not be closed. This is not in scope of this change however. 
6. Refactor message channel generation to ensure message channel is only generated once using a singleton
7. Change host application to listen to messages earlier (to useEffect) as otherwise the Custom View can already send the ready message before the host started listening

##### Testing

This is verified by existing tests in my opinion and would be hard to test End-to-End due to the timing concerns. I've debugged this at length in the Playground and I can see that the Custom View _always_ send the ready message before the host times out to instruct initialization as a last resort.

### DIFF
--- a/packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
+++ b/packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
@@ -137,10 +137,6 @@ function CustomViewLoader(props: TCustomViewLoaderProps) {
       return;
     }
 
-    // Listen for messages from the iFrame
-    iFrameCommunicationChannel.current.port1.onmessage =
-      messageFromIFrameHandler;
-
     // This is for backwards compatibility with custom view shell older than v24.0.0
     // where the custom-view-shell does not send the CUSTOM_VIEW_READY message yet.
     setTimeout(() => {
@@ -154,7 +150,14 @@ function CustomViewLoader(props: TCustomViewLoaderProps) {
   useEffect(() => {
     // Close the channel when the component unmounts
     const communicationChannel = iFrameCommunicationChannel.current;
+
+    // Listen for messages from the iFrame
+    iFrameCommunicationChannel.current!.port1.onmessage =
+      messageFromIFrameHandler;
+    window.addEventListener('message', messageFromIFrameHandler);
+
     return () => {
+      window.removeEventListener('message', messageFromIFrameHandler);
       communicationChannel?.port1.close();
     };
   }, []);
--- a/packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
+++ b/packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
@@ -49,12 +49,23 @@ const CustomPanelIframe = styled.iframe`
   border: none;
 `;
 
+/**
+ * Provides a singleton MessageChannel instance for communication between Custom View iFrame and the host application.
+ */
+let messageChannelInstance: MessageChannel | null = null;
+function getMessageChannel(): MessageChannel {
+  if (!messageChannelInstance) {
+    messageChannelInstance = new MessageChannel();
+  }
+
+  return messageChannelInstance;
+}
 function CustomViewLoader(props: TCustomViewLoaderProps) {
   const iFrameElementRef = useRef<HTMLIFrameElement>(null);
   const dataLocale = useApplicationContext((context) => context.dataLocale);
   const projectKey = useApplicationContext((context) => context.project?.key);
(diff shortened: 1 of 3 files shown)


packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -49,16 +49,35 @@ const CustomPanelIframe = styled.iframe`
...
 function CustomViewLoader(props: TCustomViewLoaderProps) {
   const iFrameElementRef = useRef<HTMLIFrameElement>(null);
   const dataLocale = useApplicationContext((context) => context.dataLocale);
   const projectKey = useApplicationContext((context) => context.project?.key);
   const featureFlags = useAllFeatureToggles();
-  const iFrameCommunicationChannel = useRef(new MessageChannel());
+  const iFrameCommunicationChannel = useRef(getMessageChannel());
   const showNotification = useShowNotification();
   const intl = useIntl();
+  const hasSentInitializationMessages = useRef(false);
```
tdeekens, line 71: This is used to make sure we have a "only one winner comes first" in that we only either send the message as a result of a 500ms timeout or the Custom View reporting readiness.

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -110,14 +137,8 @@ function CustomViewLoader(props: TCustomViewLoaderProps) {
...
 
-    // Listen for messages from the iFrame
-    iFrameCommunicationChannel.current.port1.onmessage =
-      messageFromIFrameHandler;
-
-    // TODO: Locally (starter templates) we're seeing a little delay while rendering
-    // the iFrame. The CustomViewShell gets rendered, but the effect to start listening
-    // for messages is triggered after the iFrame is ready.
-    // This is a temporary fix to avoid the situation when running locally.
+    // This is for backwards compatibility with custom view shell older than v24.0.0
```
tdeekens, line 140: Just left some comments to ensure we keep track of it.

packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -129,7 +150,14 @@ function CustomViewLoader(props: TCustomViewLoaderProps) {
   useEffect(() => {
     // Close the channel when the component unmounts
     const communicationChannel = iFrameCommunicationChannel.current;
+
+    // Listen for messages from the iFrame
+    iFrameCommunicationChannel.current!.port1.onmessage =
```
tdeekens, line 155: We need to listen earlier. The Custom View will send the message in most cases before the iFrame is in ready state as the code already executed. If we do not move this earlier, then the host fails to react to the message.

tdeekens (commented):
Generally, the entire communication could also follow a send/ack approach where either side resends in the case of the receiver not ack'ing the message given a certain timeframe. I am not sure that's something we have to do however given we only need one message and that we will likely always have a fallback to send after 500ms which can be seen as a "non-ack'ed" resending.

kark:
> > Thank you for all the improvements! 🙏
> > I've noticed only one issue. A custom view can only be opened once.
> > After closing and re-opening (or opening some other custom view), we get the following error:
> > ```
> > 
> > DataCloneError: Failed to execute 'postMessage' on 'Window': Port at index 0 is already neutered.
> > ```
> 
> Thanks. Is this happening only on my fix PR here?

Yes, on `main` the demo custom view can be opened multiple times

mustafaasif2 (approved):
Thank you! LGTM 👍🏼

CarlosCortizasCT (approved):
Looks good to me.

I've also tested locally with the playground that the js custom view template (both in deb and prod modes).

Thanks for following up on this topic 🙇

PR #3792, opened 2025-06-05
Update `flopflip` to `v15` which supports only React `>=18`

tdeekens:
#### Summary

This major version is mainly an internal house keeping release which replaced e.g. ESLint with Biome, added Turborepo and replaced preconstruct with tsup.

The major version release is rooted in requiring React v18. Which is now supported across the repository.


emmenko (approved):
Nice!

PR #3795, opened 2025-06-10
Add components for protected ans suspended routing

tdeekens:
#### Summary

This adds two components which help with working with conditionally rendered routes which are suspended. 

These components add to the already existing `SuspendedRoute` why easing working with conditional routing such as permission based routes.

Using them you can more convienently built use cases such as:

```jsx
<SuspendedProtectedRoute
  path="/my-route/"
  condition={canViewDiscountGroups}
>
  <LazilyLoadedComponent />
</SuspendedProtectedRoute>
```

or just:

```jsx
<ProtectedRoute
  path="/my-route/"
  condition={canViewDiscountGroups}
>
  <Component />
</ProtectedRoute>
```

instead of:

```jsx
{condition && <Route
  path="/my-route/"
>
  <Suspense fallback={<LoadingSpinner />}>
    <LazilyLoadedComponent />
  </Suspense>
</Route> }
```

### DIFF
--- a/packages/application-shell/src/components/suspended-route/suspended-protected-route.tsx
+++ b/packages/application-shell/src/components/suspended-route/suspended-protected-route.tsx
@@ -2,14 +2,17 @@ import { Suspense } from 'react';
 import LoadingSpinner from '@commercetools-uikit/loading-spinner';
 import { ProtectedRoute, type TProtectedRouteProps } from './protected-route';
 
-type TSuspendedProtectedRouteProps = TProtectedRouteProps;
+type TSuspendedProtectedRouteProps = TProtectedRouteProps & {
+  fallback?: React.ReactNode;
+};
 
-const SuspendedProtectedRoute: React.FC<TSuspendedProtectedRouteProps> = (
-  props
-) => {
+const SuspendedProtectedRoute: React.FC<TSuspendedProtectedRouteProps> = ({
+  fallback = <LoadingSpinner />,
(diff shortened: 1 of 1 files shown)


packages/application-shell/src/components/suspended-route/suspended-protected-route.tsx
```diff
@@ -0,0 +1,17 @@
...
+import LoadingSpinner from '@commercetools-uikit/loading-spinner';
+import { ProtectedRoute, type TProtectedRouteProps } from './protected-route';
+
+type TSuspendedProtectedRouteProps = TProtectedRouteProps;
+
+const SuspendedProtectedRoute: React.FC<TSuspendedProtectedRouteProps> = (
+  props
+) => {
+  return (
+    <Suspense fallback={<LoadingSpinner />}>
```
ragafus, line 11: In some cases a route opens a drawer. In that scenario the LoadingSpinner would increase CLS. 
For those few scenarios we could provide a way of overriding the default value of the fallback

```
type TSuspendedProtectedRouteProps = TProtectedRouteProps & {
  fallback: React.ReactNode
};

const SuspendedProtectedRoute: React.FC<TSuspendedProtectedRouteProps> = (
  { fallback = <LoadingSpinner /> }
) => {
  return (
    <Suspense fallback={fallback>
          <ProtectedRoute {...props} />
    </Suspense>
  );
};
```

PR #3796, opened 2025-06-11
feat: secure preview release workflow by restricting access to collaborators

mustafaasif2:
#### Summary

Hardened the `preview-release-on-comment.yml` workflow by restricting its execution to trusted collaborators and explicitly scoping permissions on other workflows, mitigating the CodeQL security warning.

#### Description

The original `preview-release-on-comment.yml` workflow was flagged by CodeQL for a critical security vulnerability. It allowed untrusted pull request code to execute in a privileged context.

Changes made in this PR

- Added a check to ensure that only GitHub users who are members of commercetools organization can trigger the workflow via issue comments. 
- Reviewed and minimized the workflow’s GitHub permissions, ensuring only the necessary scopes (e.g., no blanket write-all) are granted to steps that require them.

#### Why not split the workflow or use label-based verification

While splitting the workflow (into build and publish phases) or using GitHub's recommended "safe to test" label are valid mitigations, they were not adopted due to the following:

- `MEMBER` verification is a robust and simpler solution for our use case, avoiding the overhead of managing artifacts and extra workflows.
- Label-based verification is vulnerable to race conditions, where malicious code could be pushed after the label is applied but before the workflow runs. GitHub notes this should only be used as a temporary workaround.

This approach effectively reduces the attack surface while keeping the workflow straightforward and maintainable. The changes were tested locally using the CodeQL CLI and confirmed to no longer trigger any security alerts. These changes remove **4 critical** and **2 medium** severity issues.

### DIFF
--- a/.github/workflows/preview-release-on-comment.yml
+++ b/.github/workflows/preview-release-on-comment.yml
@@ -15,7 +15,7 @@ jobs:
     if: |
       github.event.issue.pull_request && 
       startsWith(github.event.comment.body, '[preview_deployment]') &&
-      github.event.comment.author_association == 'COLLABORATOR' &&
+      github.event.comment.author_association == 'MEMBER' &&
       github.event.issue.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
 


.github/workflows/preview-release-on-comment.yml
```diff
@@ -74,4 +78,4 @@ jobs:
           gh issue comment ${{ github.event.issue.number }} \
             --body "Release workflow ${{ job.status == 'success' && 'succeeded ✅' || 'failed ❌' }}\nSee details: [Workflow Run](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})"
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
VictorAvelar, line 81: nitpick: for consistency, this file is missing an EOF line.

VictorAvelar (approved):

.github/workflows/preview-release-on-comment.yml
```diff
@@ -12,7 +12,11 @@ on:
 
 jobs:
   release:
-    if: github.event.issue.pull_request && startsWith(github.event.comment.body, '[preview_deployment]')
+    if: |
+      github.event.issue.pull_request && 
+      startsWith(github.event.comment.body, '[preview_deployment]') &&
+      github.event.comment.author_association == 'COLLABORATOR' &&
```
kark, line 18: nit: just wonder if this could be `MEMBER`? Otherwise FEC and others won't be able to create previews on their own?

kark (approved):
Awesome! Great work investigating this! 🙏
I'm happy to give this a try

.github/workflows/preview-release-on-comment.yml
mustafaasif2: With `COLLABORATOR`, other teams would be able to create previews too as they also have write access to the repository. However, since we do not give access to external collaborators (I think), it makes more sense to make add `MEMBER` instead of `COLLABORATOR`. I'll make the changes

mustafaasif2: Changed in this commit: a059917b2a607d119db9b86141e2e6acc65d45b7

PR #3800, opened 2025-06-17
Refactor `SuspendedProctedRoute` to support loader and fallback component

tdeekens:
#### Summary

This refactors the ProtectedRoute and SuspendedProtectedRoute to both support a `fallback` component which is not the Suspense boundary but rather the fallback rendered whenever the condition is not met.

### DIFF
--- a/packages/application-shell/src/components/suspended-route/protected-route.tsx
+++ b/packages/application-shell/src/components/suspended-route/protected-route.tsx
@@ -3,18 +3,16 @@ import { PageUnauthorized } from '@commercetools-frontend/application-components
 
 type TProtectedRouteProps = {
   condition: boolean;
+  fallback?: React.ReactNode;
 } & RouteProps;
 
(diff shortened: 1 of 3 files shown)


VictorAvelar (approved):

packages/application-shell/src/components/suspended-route/protected-route.tsx
```diff
@@ -3,18 +3,16 @@ import { PageUnauthorized } from '@commercetools-frontend/application-components
...
+  fallback = <PageUnauthorized />,
   children,
   ...routeProps
 }) => {
-  return condition ? (
-    <Route {...routeProps}>{children}</Route>
-  ) : (
-    <PageUnauthorized />
-  );
+  return condition ? <Route {...routeProps}>{children}</Route> : <>fallback</>;
```
CarlosCortizasCT, line 15: Sorry to jump in here late but if the `condition` is falsy we would be rendering the text "fallback" instead of rendering the component, righ?

PR #3801, opened 2025-06-19
Fix rendering fallback as string

tdeekens:
#### Summary

Fixes rendering the string fallback instead of the value from props.


CarlosCortizasCT (approved):
👍

tdeekens:
Thanks for catching @CarlosCortizasCT. Good reminder to not code from an S Bahn 😇

PR #3804, opened 2025-06-25
feat: add onCustomViewClose callback for data refresh

mustafaasif2:
#### Summary

Added an optional `onCustomViewClose` callback prop to the `TabularDetailPage` component that allows host applications to refresh their data when Custom Views are closed.

#### Description

This enhancement enables better data synchronization between Custom Views and their host applications. When a Custom View closes, the optional `onCustomViewClose` function is triggered, allowing the ability for the host applications to refresh their data after Custom View closure.

The implementation maintains full backward compatibility. Existing code continues to work without any changes. The callback is passed through the component hierarchy from TabularDetailPage → CustomViewsSelector → CustomViewLoader, ensuring it's triggered at the appropriate time when Custom Views close.

An [RFC](https://commercetools.atlassian.net/wiki/spaces/shield/pages/1570734207/RFC+Update+parent+page+data+on+Custom+View+changes) has been written that considered some other alternatives as well

### DIFF
--- a/.changeset/lucky-sheep-cough.md
+++ b/.changeset/lucky-sheep-cough.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-components': minor
+---
+
+Adds an optional `onCustomViewClose` callback prop to `TabularDetailPage`
--- a/packages/application-components/src/components/custom-views/custom-views-selector/custom-views-selector.tsx
+++ b/packages/application-components/src/components/custom-views/custom-views-selector/custom-views-selector.tsx
@@ -27,6 +27,7 @@ type TCustomViewSelectorWithRequiredProps = {
   customViewLocatorCode: string;
   margin?: string;
   onCustomViewsResolved?: TCustomViewSelectorProps['onCustomViewsResolved'];
+  onCustomViewClose?: TCustomViewSelectorProps['onCustomViewClose'];
 };
 
 type TWrapperProps = {
@@ -162,6 +163,9 @@ function CustomViewSelector(props: TCustomViewSelectorWithRequiredProps) {
               customView={selectedCustomView}
               onClose={() => {
                 setSelectedCustomView(null);
+                if (props.onCustomViewClose) {
+                  props.onCustomViewClose();
+                }
               }}
             />
           )}
--- a/packages/application-components/src/components/custom-views/custom-views-selector/types.ts
+++ b/packages/application-components/src/components/custom-views/custom-views-selector/types.ts
@@ -13,4 +13,5 @@ export type TCustomViewSelectorProps = {
   customViewLocatorCode?: string;
   customViewLocatorCodes?: Record<string, TCustomViewLocatorCode>;
   margin?: string;
+  onCustomViewClose?: () => void;
 };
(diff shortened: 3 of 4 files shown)


CarlosCortizasCT:
Sorry to jump in here but I'm curious about how is this supposed to work form the host application point of view.
I'm asking because the `TabularDetailPage` component usually don't know which specific tab is actually rendered so I don't understand how is the host application expected to refresh the data the user is seeing after the Custom View is closed.

emmenko:
Thanks @CarlosCortizasCT for the feedback. This is a very valid point.

The main use case we are looking to solve at the moment relates to product / order details. In that case, the tab itself doesn't matter much as the entire product data is supposed to be refetched.

However, there is definitely room for improvements here. For example, we can pass an object to the callback function with e.g. the `customViewLocatorCode` so that the consumer implementing the callback has more information about the locator that triggered it.

CarlosCortizasCT:
@emmenko @mustafaasif2 Thanks for sharing more context so I can better understand the goals.

I know you already discussed about this topic and you already have a RFC where you researched the topic and make a decision but I can't help also sharing the idea I had in mind some time ago when I worked in the Custom Views feature.

It seems the actual goal is to be able for a user to see changes done in a Custom View reflected in the data displayed in the host application view they're in.
The problem is that we fetch data from very different parts of the components tree in each application so it's no easy to have an abstract way to refresh data across different components. From my point of view, we would like to have the refresh logic as closer to the where the data is stored as possible. Data can be stored in the Apollo cache, a React context, a React state, etc. which is controlled by a specific component so I believe the component managing the data is the one which should have the logic to refresh it.

What I had in mind is a custom React hook any team could implement in any part of their application where a refresh would be needed. Instead of tying the solution to the `TabularDetailPage` which might limit the ability to refresh the actual piece of data that might be needed, the data owner component could implement not only the data fetching but also the data refresh.

A potential custom hook API which components could implement would look like this:
```ts
useCustomViewIntegration({
    locators: ['products.product_variant_details.general'],
    onRefreshDataRequested: (context) => {
      // Update relevant data
    },
  });
```

Let's say we want the inventories tab of a product variant to be updated when a Custom View changes something related to it. We could have something like this in the relevant component:
```ts
const {
    inventories,
    areInventoriesLoading,
    areAllChannelsAssigned,
    inventoriesTotal,
    refreshInventories,
  } = useInventories([props.originalSku], {
    limit: perPage.value,
    offset: (page.value - 1) * perPage.value,
    sort: `${tableSorting.value.key} ${tableSorting.value.order}`,
  });

useCustomViewIntegration({
    locators: ['products.product_variant_details.inventory'],
    onRefreshDataRequested: (context) => {
      refreshInventories();
    },
  });
```

This hook could use a simple Custom Event system internally to listen for events triggered from the Custom View: it could be based on the close event as you already decided or it could even be that Custom Views developers could also have a hook they could use to trigger an event as part of their implementation so the host update happens in the background and it's already there for the user when the Custom View is closed.

Of course this is your team's area so it's your decision but I just wanted to share the idea.

In case you are interested, I'm open to provide more details, meet or work on a PoC.

mustafaasif2:
@CarlosCortizasCT Thank you so much for the detailed explanation. I really like your idea and I think it is slightly similar to the option 2 I suggested. However your idea is more extensible and not just limited to the `TabularDetailPage` and could work for any event. I am assuming that `useCustomViewIntegration` would use a useEffect hook to register and register the event listener. 

The idea of giving the Custom View owners the ability to refresh it themselves was also discussed internally but eventually it was decided that lets not give them this ability and lets handle it ourselves instead. As the Host Application is unusable while the Custom View is open we decided to go the simpler route of only refreshing the Host Application once the Custom View closes. I think the exact event to trigger the refresh can be discussed. 

As my team is currently working on another initiative and we aren't code owners of this repo, maybe I can hand them over the work and then have a discussion with them about the possibility for the solution you suggested.

mustafaasif2:
Closing this PR. After discussion, it was decided that we will implement a more extensible solution that Carlos suggested.

PR #3810, opened 2025-07-17
feat: add custom event logic in Custom Views

mustafaasif2:
#### Summary

Implement `useCustomViewParentDataRefresher` hook to enable flexible and extensible data refresh for host applications when Custom Views close.

#### Description

This PR introduces a new React hook, `useCustomViewParentDataRefresher`, designed to be used by any component that needs to refresh its data after a Custom View closes.

This hook enables individual data-owning components (such as tabs) to register for refresh events. It listens for the Custom View closure event through a custom event system and invokes the provided refresh callback when the event occurs.

Example usage:

```
useCustomViewParentDataRefresher({
  locators: ['products.product_details.general'],
  onRefreshDataRequested: (context) => {
    console.log('[StateMachinesList] onRefreshDataRequested', context);
  },
});
```

More information can be found here: https://github.com/commercetools/merchant-center-application-kit/pull/3804

### DIFF
--- a/packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
+++ b/packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
@@ -26,6 +26,7 @@ type TCustomViewLoaderProps = {
   customView: CustomViewData;
   hostUrl?: string;
   onClose: () => void;
+  locatorCode: string;
 };
 
 const isIframeReady = (iFrameElementRef: HTMLIFrameElement) => {
@@ -102,7 +103,7 @@ function CustomViewLoader(props: TCustomViewLoaderProps) {
             window.dispatchEvent(
               new CustomEvent('custom-view-closed', {
                 detail: {
-                  locators: props.customView.locators,
+                  originLocatorCode: props.locatorCode,
                 },
               })
             );
--- a/packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
+++ b/packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
@@ -26,7 +26,7 @@ type TCustomViewLoaderProps = {
   customView: CustomViewData;
   hostUrl?: string;
   onClose: () => void;
-  locatorCode: string;
+  locatorCode?: string;
 };
 
 const isIframeReady = (iFrameElementRef: HTMLIFrameElement) => {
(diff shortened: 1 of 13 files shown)


packages/application-components/src/components/custom-views/custom-view-loader/custom-view-loader.tsx
```diff
@@ -99,6 +99,13 @@ function CustomViewLoader(props: TCustomViewLoaderProps) {
         switch (event.data.eventName) {
           case CUSTOM_VIEWS_EVENTS_NAMES.CUSTOM_VIEW_CLOSE:
             props.onClose();
+            window.dispatchEvent(
+              new CustomEvent('custom-view-closed', {
+                detail: {
+                  locators: props.customView.locators,
```
CarlosCortizasCT, line 105: The PR I shared previously was a basic implementation to showcase how this approach could be implemented. It was not a 100% solid solution, it was a PoC.

Something I deliberately left out of that example is how `locators` should be managed in the event.
In the current implementation we would be triggering any listener where this CustomView would be rendered as they can configured to be visible in different pages of different applications. In theory, this mean when we close a CustomView in a specific page of an application, not only the listener of that page would be executed but any listener that has been registered.

In theory this won't be a problem because most probably we would only have one listener registered for a page at a given time and I don't see how we could still have listeners for that CustomView registered from a different page (only if we have a memory reference hanging over there).
However, I think it would be better if we only trigger the listeners for this CustomView which are registered with the locator context this CustomView is executed from. That would probably require to pass the current locator from the `CustomViewsSelector` component to this one.

I know this comment might not be super clear so let me know what you think. 
I can also create another PR with an example of what I'm trying to explain.

packages/application-shell/src/components/custom-view-dev-host/custom-view-dev-host.tsx
```diff
@@ -81,6 +81,7 @@ const LocalCustomViewLauncher = (props: TLocalCustomViewLauncherProps) => {
           {shouldRenderCustomView && (
             <CustomViewLoader
               customView={customViewConfig}
+              locatorCode={''}
```
mustafaasif2, line 84: Is it fine to keep this empty?

packages/application-shell-connectors/src/hooks/custom-views/use-custom-view-integration.ts
```diff
@@ -0,0 +1,30 @@
...
+      const { originLocatorCode } = event.detail;
+      if (params.locators.includes(originLocatorCode)) {
+        params.onRefreshDataRequested(event.detail);
+      }
+    },
+    []
+  ) as EventListener;
+
+  useEffect(() => {
+    window.addEventListener('custom-view-closed', onCustomViewEventHandler);
```
emmenko, line 22: Nit: define the events in the `CUSTOM_VIEWS_EVENTS_NAMES` in the `constants` package

https://github.com/commercetools/merchant-center-application-kit/blob/e02bccd6aaac8e055a5f4c1376aafbfbcc12bb34/packages/constants/src/constants.ts#L335-L341

emmenko (commented):
Great job! 🙌

I left some comments to clarify a couple of things.

packages/application-shell-connectors/src/hooks/custom-views/use-custom-view-integration.ts
```diff
@@ -0,0 +1,30 @@
+import { useCallback, useEffect } from 'react';
+
+export type TCustomViewIntegrationParams = {
+  locators: string[];
+  onRefreshDataRequested: (context: Record<string, unknown>) => void;
```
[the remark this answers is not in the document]
tdeekens: +1 for making the naming more generic.

PR #3814, opened 2025-07-22
[PoC] React Router migration

CarlosCortizasCT:
## Description

This is a temporary branch just to test the React Router migration in the Playground application so we can learn how the process should look like when we want to do this in the MC applications.


CarlosCortizasCT:
Closed in favor of #3818

PR #3819, opened 2025-08-04
refactor: use cypress v14

emmenko:
We're using it already internally in the MC, let's use it also for the tests here in App Kit.


emmenko:
Correct, no new release.

PR #3824, opened 2025-08-06
fix: always redirect to account project creation when user has no projects

emmenko:
I can't remember why we decided to log users out in case they have no projects.
We should redirect them to the project creation page in the account app.

@tdeekens can you recall any specific logic or requirement around that?

### DIFF
--- a/packages/application-shell/src/components/project-container/project-container.tsx
+++ b/packages/application-shell/src/components/project-container/project-container.tsx
@@ -37,7 +37,7 @@ type QueryParams = {
 };
 type TProjectContainerProps = {
   user: TFetchLoggedInUserQuery['user'];
-  environment: TProviderProps<{ enableSignUp?: boolean }>['environment'];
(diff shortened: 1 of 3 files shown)


packages/application-shell/src/components/project-container/project-container.tsx
```diff
@@ -37,7 +37,7 @@ type QueryParams = {
 };
 type TProjectContainerProps = {
   user: TFetchLoggedInUserQuery['user'];
-  environment: TProviderProps<{ enableSignUp?: boolean }>['environment'];
```
emmenko, line 40: I checked in our internal apps and we don't seem to set this variable anymore in the app config. It should be safe to remove it.

tdeekens: It's a feature flag not an environment config now.

tdeekens (approved):
I can't recall a technical reason.

kark (approved):

PR #3830, opened 2025-08-11
feat(cypress): define configurable timeouts for login command interactions

emmenko:
By default the `cy.get` command waits for 4s. In Identity, the login is a 2-step flow and the loading of the second step might take a bit longer.

The PR adds a new command option `timeouts` to allow to configure timeouts for the login command interactions.

Options:
* `waitForEmailInput` (defaults to `4000`)
* `waitForPasswordInput` (defaults to `8000`)

Example:
```ts
cy.loginToMerchantCenter({
  // ...
  timeouts: {
    waitForPasswordInput: 10000,
  },
});
```

### DIFF
--- a/packages/cypress/src/add-commands/login.ts
+++ b/packages/cypress/src/add-commands/login.ts
@@ -260,20 +260,23 @@ function loginByForm(commandOptions: CommandLoginOptions) {
                   args: {
                     userCredentials,
                     identityUrl,
+                    timeouts: commandOptions.timeouts,
                   },
                 },
                 ({
                   userCredentials,
                   identityUrl,
+                  timeouts,
                 }: {
                   userCredentials: LoginCredentials;
                   identityUrl: string;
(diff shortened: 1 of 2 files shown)


packages/cypress/src/add-commands/login.ts
```diff
@@ -77,6 +94,11 @@ export type LoginToMerchantCenterForCustomViewCommandLoginOptions = Omit<
 // Alias for backwards compatibility
 export type CommandLoginByOidcOptions = CommandLoginOptions;
 
+const defaultTimeouts: LoginCommandTimeouts = {
+  waitForEmailInput: 4000,
+  waitForPasswordInput: 8000,
```
emmenko, line 99: Bumped by default to 8s instead of 4s.

kark (approved):
Thanks for the improvements 💯 I haven't noticed any regressions.

PR #3833, opened 2025-08-25
fix(cypress): login timeouts

kark:
https://commercetools.slack.com/archives/CHELEBDTK/p1756112353763879


packages/cypress/src/add-commands/login.ts
```diff
@@ -368,7 +390,15 @@ function loginByForm(commandOptions: CommandLoginOptions) {
             }
           } else {
             if (isGlobalIdentityEnabled) {
-              cy.url().should('include', `${identityUrl}/login`);
+              // Wait for redirect to Identity to complete
+              // eslint-disable-next-line cypress/no-unnecessary-waiting
+              cy.wait(8000);
```
kark, line 395: Added some defensive cy waits, which can probably be lowered down

tdeekens (approved):

PR #3985, opened 2026-04-27
feat(mc-scripts): add `--handle-auth-routes` flag to `serve` command

misama-ct:
## Summary

Adds one new CLI option to `mc-scripts serve`:

- **`--handle-auth-routes <enabled>`** (default `true`). When set to `false`, requests to `/login*` and `/logout*` fall through to the SPA fallback instead of being intercepted by the built-in localhost-redirect / clear-session-text / 404 handlers.

Default behavior is unchanged. Only the auth-route branches in the request handler are gated; the favicon rewrite and the `sirv` static + SPA fallback are unaffected.

6 new tests in `serve.spec.ts` covering the `handleAuthRoutes` opt-out under both localhost and non-localhost `mcApiUrl`. The existing 19-test suite already exercises non-default port handling via `run({ port: 0 })` and non-default `publicPath` (every test writes to a temp fixture dir).

## Motivation (FEC-820)

PR #3979 swapped `serve-handler` → `sirv` in `mc-scripts serve` while preserving the prior behavior of returning 404 (or applying inline localhost handlers) on `/login*` / `/logout*`. That behavior is wrong for applications that host their own React routes under those paths — they need plain SPA fallback there.

## Usage

**Default — every standard MC application.** Auth routes keep their existing intercept/404 contract:

```jsonc
// package.json
{
  "scripts": {
    "start:prod:local": "pnpm compile-html:local && mc-scripts serve"
  }
}
```

**Opt-out — applications that own `/login*` and `/logout*` themselves.** Required when migrating off the standalone `serve` npm package for an app like `application-authentication`, whose React routes live under those paths and would otherwise 404 against the default contract:

```jsonc
// packages-application/application-authentication/package.json
{
  "scripts": {
    // Was: `serve -s -l 3001 public`
    "start:prod:local": "pnpm compile-html:local && mc-scripts serve --handle-auth-routes false"
  },
  "devDependencies": {
    // The standalone `serve` npm package can be dropped — `mc-scripts serve`
    // covers this contract now: same default port (3001 = `developmentPort`)
    // and same default folder (`public` = `paths.appBuild`).
  }
}
```

## Test plan

- [x] `pnpm test packages/mc-scripts/src/commands/serve.spec.ts` — 25/25 pass (19 existing + 6 auth-route opt-out)
- [x] `pnpm typecheck` — clean
- [ ] Smoke-test locally with the opt-out: `mc-scripts serve --handle-auth-routes false`, hit a `/login/*` route, confirm the SPA shell loads (not 404)
- [ ] Smoke-test without the option, confirm `/login/authorize` still redirects under localhost `mcApiUrl` and bare `/login` still 404s

### DIFF
--- a/packages/mc-scripts/src/cli.ts
+++ b/packages/mc-scripts/src/cli.ts
@@ -148,6 +148,14 @@ async function run() {
       '(optional) If defined, requests to `/login*` and `/logout*` are passed through to the SPA fallback instead of being intercepted by the built-in login/logout handlers. Use this for applications that own the auth routes themselves (e.g. `application-authentication`).',
       false
     )
+    .addHelpText(
+      'after',
+      `
+Examples:
+  $ mc-scripts serve
+  $ mc-scripts serve --handle-auth-routes
+`
+    )
     .action(async (options: TCliCommandServeOptions) => {
       const globalOptions = program.opts<TCliGlobalOptions>();
 
--- a/packages/mc-scripts/src/cli.ts
+++ b/packages/mc-scripts/src/cli.ts
@@ -148,14 +148,6 @@ async function run() {
       '(optional) If defined, requests to `/login*` and `/logout*` are passed through to the SPA fallback instead of being intercepted by the built-in login/logout handlers. Use this for applications that own the auth routes themselves (e.g. `application-authentication`).',
       false
     )
-    .addHelpText(
-      'after',
-      `
-Examples:
-  $ mc-scripts serve
-  $ mc-scripts serve --handle-auth-routes
-`
-    )
     .action(async (options: TCliCommandServeOptions) => {
       const globalOptions = program.opts<TCliGlobalOptions>();
 
--- a/packages/mc-scripts/src/cli.ts
+++ b/packages/mc-scripts/src/cli.ts
@@ -144,9 +144,9 @@ async function run() {
       'Serves previously built and compiled application from the "public" folder.'
     )
     .option(
-      '--handle-auth-routes',
-      '(optional) If defined, requests to `/login*` and `/logout*` are passed through to the SPA fallback instead of being intercepted by the built-in login/logout handlers. Use this for applications that own the auth routes themselves (e.g. `application-authentication`).',
-      false
+      '--handle-auth-routes <enabled>',
+      '(optional) Whether `/login*` and `/logout*` are intercepted by the built-in login/logout handlers (`true`, the default) or passed through to the SPA fallback (`false`). Set to `false` for applications that own those routes themselves (e.g. `application-authentication`).',
+      'true'
     )
     .action(async (options: TCliCommandServeOptions) => {
       const globalOptions = program.opts<TCliGlobalOptions>();
@@ -160,7 +160,7 @@ async function run() {
 
       const serveCommand = await import('./commands/serve');
       const server = await serveCommand.default({
-        handleAuthRoutes: options.handleAuthRoutes,
+        handleAuthRoutes: options.handleAuthRoutes !== 'false',
       });
       const address = server.address();
       const boundPort =
--- a/packages/mc-scripts/src/cli.ts
+++ b/packages/mc-scripts/src/cli.ts
@@ -12,6 +12,7 @@ import type {
   TCliCommandConfigSyncCIOptions,
   TCliCommandSetDeploymentPreviewOptions,
   TCliCommandLoginOptions,
+  TCliCommandServeOptions,
 } from './types';
 import doesFileExist from './utils/does-file-exist';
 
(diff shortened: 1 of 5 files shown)


packages/mc-scripts/src/cli.ts
```diff
@@ -142,7 +143,21 @@ async function run() {
     .description(
       'Serves previously built and compiled application from the "public" folder.'
     )
-    .action(async () => {
+    .option(
+      '--port <port>',
+      '(optional) Port to bind the server to. Defaults to `3001`.',
+      '3001'
```
ByronDWall: Suggest dropping `--port` from this PR — it's broken-by-default outside the `--handle-auth-routes false` path.

`process-config.ts:40` hardcodes `developmentPort = 3001` (with a `// TODO: make it configurable.`), and that constant drives the OIDC `authorizeUrl` at `process-config.ts:96-98`. So `mc-scripts serve --port 4000` binds at 4000 but the runtime config still sends the browser back to `localhost:3001` after authorize → broken login on any non-default port that still uses the default auth-routes contract.

The 3001 coupling also reaches Cypress `baseUrl`, `frontendHost`/`cdnUrl` fixtures, and the playground custom-view URL — real non-3001 use needs a coordinated change, not just a bind override.

Recommend: land `--handle-auth-routes` here, then in a follow-up make `developmentPort` configurable *first* and add `--port` on top of that so the flag isn't shipped with a known footgun. If you keep it, the help string should call out the constraint — right now it reads safer than it is.

packages/mc-scripts/src/cli.ts
```diff
@@ -142,7 +143,21 @@ async function run() {
...
+    )
+    .option(
+      '--public-path <path>',
+      "(optional) Path to the folder to serve. Defaults to the application's `public` build folder."
```
ByronDWall: Same scope concern as `--port` — this isn't part of FEC-820 and would be easier to review on its own. Smaller blast radius than `--port` (no auth-flow coupling), but a few things worth nailing down before it lands:

- **Resolution semantics.** `paths.appBuild` is absolute; the option just hands the raw string to `sirv`. A relative `--public-path dist` resolves against `process.cwd()`, which is fine but undocumented — worth stating in the help string, or resolving against the app root explicitly so behavior doesn't depend on where the script is invoked.
- **Validation.** `--port` validates the integer range and throws a clear error; `--public-path` passes through unchecked. A missing/non-directory path surfaces as a `sirv` error at request time rather than at startup. A quick `fs.statSync` + `isDirectory()` check at CLI parse time would match the `--port` ergonomics.
- **Naming.** `--public-path` overloads a term that already means something specific in webpack/Rspack land (the URL prefix for emitted assets, not the on-disk folder). `--serve-dir` or `--build-dir` would be less ambiguous, especially since `mc-scripts` users see `publicPath` in webpack contexts too.

Happy to defer all of this to a follow-up if you'd rather keep this PR moving — just flagging while it's in front of us.

ByronDWall (commented):
The auth-routes flag is the real change here (and looks good). `--port` and `--public-path` are independent additions riding along — they review and revert independently, and `--port` in particular has a non-trivial footgun (see inline). Worth splitting so the FEC-820 fix can land clean and the serve-flag expansion gets its own discussion.

packages/mc-scripts/src/cli.ts
misama-ct: You're right about the `developmentPort` coupling — verified the chain (`process-config.ts:40` → `authorizeUrl` at :97; the same constant also drives `frontendHost` / dev `cdnUrl` via `developmentAppUrl`).

Not pretending to fix that here. It's pre-existing tech debt and a real fix needs coordinated changes across `process-config.ts` / `compile-html` / `serve` — outside FEC-820 scope. What this PR does is refuse the combination that breaks silently today:

- ba21d692c — `assertCompatibleServeOptions` throws at startup on `--port=<non-3001>` + default `--handle-auth-routes=true`. Error names the affected baked-in URLs and points the user at `--handle-auth-routes false`.
- 18cd3e6cb — tests covering the throw and its absence on supported combinations.

So `--port` ships intentionally narrow — only useful for apps that own their auth routes themselves. Not the end state, just enough to retire the standalone `serve` consumer without making the existing footgun more accessible.

emmenko: Why do we want to change the port? The port is also hardcoded / expected is some places of the MC API. It's not just a frontend change.

misama-ct: I do not want to change the port, I added this so that `mc-scripts serve` can act as a drop-in replacement for the `serve` package as it was used [here](https://github.com/commercetools/merchant-center-frontend/blob/main/packages-application/application-authentication/package.json#L21). Would you be more comfortable If I hardcode the port and adjust the `start:prod:local` command in the mc's auth app?

misama-ct:
@ByronDWall I don't want this to drag any longer, so I stripped it back to the bare essentials, which is being able to deactivate the auth-exception.

ByronDWall:
Thanks for paring it down, sorry I missed your reply. I agree that a follow up refactor/update pr would be the best place to handle adding the new args

ByronDWall (approved):

PR #4003, opened 2026-05-12
fix(mc-scripts): resolve URL relative to file when referenced in CSS files

emmenko:
Example build error:

```
[vite:css-post] { runtime: "window.__toCdnUrl("KaTeX_AMS-Regular-BQhdFMY1.woff2")" } is not supported for assets in css files: KaTeX_AMS-Regular-BQhdFMY1.woff2
```


tdeekens (approved):

ByronDWall (approved):
Thanks for the fix!

PR #4010, opened 2026-05-14
feat(logger): [CPO-861] Add log field removal

kittsville:
#### Summary

Adds another log formatter to natively support removing log fields

#### Description

Some bearer tokens were recently leaked to LogScale logs owing to a misconfiguration of the `rewriteFieldsFormatter`. This highlighted the need for an additional formatter dedicated to removing log fields, rather than having to use the hack of "moving it" from one field to the same field, and changing its value at the same time


emmenko:
Thanks for following up on this.

I think it's still valuable to see the field in the logs but with a redacted value. For instance, if we remove the field it might seem that the request didn't contain the field (e.g. auth header) at all which can lead to false assumptions when debugging issues.

kittsville:
I could replace the removal helper with a `redactFieldsFormatter` which replaces the value with `REDACTED`

ByronDWall:
@kittsville `redactFieldsFormatter` makes sense to me as well. Along with that update, would you mind adding a changeset to the PR?

ByronDWall (approved):
Thanks for making the update to redact instead of remove and adding a changeset. This looks good to me

PR #4017, opened 2026-05-18
fix(ci): use regex for jest testPathPattern in template tests

ByronDWall:
## Summary

- Fix the test pattern in the CI template action to use a proper regex instead of bash glob syntax
- The `{js,ts,tsx}` brace expansion inside quotes was passed literally to Jest's `--testPathPattern`, which expects a regex — causing "Invalid testPattern" warnings and Jest falling back to running **all** tests instead of just the template's tests

## Test plan

- [x] Verified locally that the regex pattern `.*\.spec\.(js|ts|tsx)$` correctly matches only template test files
- [x] CI template jobs should now show the correct number of matched test suites (not "Running all tests instead")

### DIFF
--- a/.github/actions/test-template-action/action.yml
+++ b/.github/actions/test-template-action/action.yml
@@ -71,7 +71,7 @@ runs:
 
     - name: Running integration tests
       shell: bash
-      run: pnpm test -- "${{ inputs.application-type == 'custom-application' && 'application-templates' || 'custom-views-templates'}}/${{ inputs.template-name }}/**/*.spec.{js,ts,tsx}"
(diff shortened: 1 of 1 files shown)


ByronDWall:
### Before / After

- **Before** (glob pattern, "Invalid testPattern", runs all tests): https://github.com/commercetools/merchant-center-application-kit/actions/runs/26062210694/job/76627480937#step:3:461
- **After** (regex pattern, matches only template tests): https://github.com/commercetools/merchant-center-application-kit/actions/runs/26063339429/job/76628468492?pr=4017#step:3:477

tylermorrisford (approved):

tdeekens (approved):

valoriecarli (approved):

PR #4018, opened 2026-05-20
ci: add package-shape baseline check (publint + attw)

misama-ct:
## Summary

Adds a CI gate that packs each publishable `@commercetools-frontend/*` and `@commercetools-backend/*` workspace with `pnpm pack` and runs `publint` + `@arethetypeswrong/cli` (attw) against the tarball. Findings are reported but **do not gate merges** (`REPORT_ONLY = true`) — the flip to blocking happens in the metadata cleanup sibling ([FEC-936 mirror](https://commercetools.atlassian.net/browse/FEC-936)) once the baseline reaches zero.

Ported from the Nimbus precedent ([nimbus#1498](https://github.com/commercetools/nimbus/pull/1498)).

Closes [FEC-951](https://commercetools.atlassian.net/browse/FEC-951).

## The baseline is a regression contract, not a punch list

**This PR does not fix any of the listed findings — that's not its job.** It freezes today's shape issues so the subsequent catalog adoption and metadata normalization work can be proven non-regressing. The expected outcome of a future PR is **"no new findings"**, not "all findings resolved." If a PR adds a row to the table below, that is the regression; leaving the existing rows unchanged means the baseline still holds. Findings get removed only when an explicit cleanup ticket resolves them.

## Baseline: 9 findings across 7 packages

| Package                                                                                         | Tool    | Finding                                                                                                                 |
| ----------------------------------------------------------------------------------------------- | ------- | ----------------------------------------------------------------------------------------------------------------------- |
| `@commercetools-frontend/application-shell`                                                     | attw    | `InternalResolutionError` on root + `/test-utils` (bare relative imports in emitted `.d.ts` fail node16 ESM resolution) |
| `@commercetools-frontend/browser-history`                                                       | attw    | `FalseExportDefault` on root                                                                                            |
| `@commercetools-frontend/cypress`                                                               | attw    | `FalseExportDefault` on `/task` subpath                                                                                 |
| `@commercetools-frontend/mc-html-template`                                                      | attw    | `FalseExportDefault` on `/webpack-html-template` subpath                                                                |
| `@commercetools-frontend/mc-scripts`                                                            | attw    | `FalseExportDefault` on `/webpack-loaders/i18n-message-compilation-loader` subpath                                      |
| `@commercetools-frontend/{assets,babel-preset-mc-app,jest-preset-mc-app,jest-stylelint-runner}` | publint | `pkg.module should be ESM, but the code is written in CJS` (4 preconstruct-excluded CJS packages with a `module` field) |

Two failure classes account for all 9 findings: attw `FalseExportDefault` on subpath exports, and publint declaring `module` for CJS code. `application-shell` is the structural outlier (relative-import rewriting needed, mirroring the Nimbus precedent).

## Heads-up: `fflate` pinned to 0.8.2

`attw@0.18.2` crashes on every tarball when its transitive dependency `fflate` resolves to `0.8.3` (a behavior change in `fflate`'s streaming API broke attw's tarball extraction). This PR adds a `pnpm.overrides` entry pinning `fflate` to `0.8.2`. The override is scoped to the attw dependency chain — nothing else in the repo uses `fflate` — so the blast radius is just this check. Drop the pin once attw ships a fix.

## What changed

- `scripts/check-package-shape.mjs` — new, ported from `maintained/nimbus/scripts/check-package-shape.mjs`. The package list is derived at runtime from `pnpm m ls` and filtered by `"private" !== true` — today, the 29 publishable workspaces under `@commercetools-frontend/*` and `@commercetools-backend/*`. Per-package work runs through a small worker pool sized by `os.availableParallelism()` (≈4 on standard GHA runners) so we don't spawn 29 simultaneous `pnpm pack` + attw + publint chains.
- `package.json` — adds `@arethetypeswrong/cli@^0.18.2` + `publint@^0.3.21` and the `check:package-shape` script. Adds `pnpm.overrides.fflate = "0.8.2"`.
- `.github/workflows/main.yml` — adds a `Check package shape` step in `lint_and_test` after typecheck. Sequential step (reuses the existing install+build), not a parallel job — matches the Nimbus precedent.
- `docs/package-shape-verification.md` — new, full overview + REPORT_ONLY toggle docs.

No changeset: this PR adds zero published-package source changes — no version bump or CHANGELOG entry is warranted.

## Test plan

- [ ] CI's new "Check package shape" step runs successfully and prints the 9 baseline findings.
- [ ] Job is non-blocking (exits 0 in REPORT_ONLY mode).
- [ ] Local repro: `pnpm install && pnpm build && pnpm check:package-shape` reproduces the same findings.

[FEC-951]: https://commercetools.atlassian.net/browse/FEC-951?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/check-package-shape.mjs
+++ b/scripts/check-package-shape.mjs
@@ -23,7 +23,7 @@
  */
 
 import { spawn } from 'node:child_process';
-import { mkdtempSync, rmSync, existsSync } from 'node:fs';
+import { mkdtempSync, rmSync } from 'node:fs';
 import { tmpdir } from 'node:os';
 import { join } from 'node:path';
 import { fileURLToPath } from 'node:url';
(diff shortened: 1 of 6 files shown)


scripts/check-package-shape.mjs
```diff
@@ -0,0 +1,300 @@
...
+// Surface findings without failing CI. Flip to `false` once the baseline
+// reaches zero so this gate starts blocking merges.
+const REPORT_ONLY = true;
+
+// Mirror of the publishable workspaces in the Changesets `fixed` group
+// (`.changeset/config.json`) — every `@commercetools-frontend/*` and
+// `@commercetools-backend/*` package the changeset bot ships. Private
+// workspaces (`@commercetools-applications/*`, `@commercetools-local/*`,
+// `@commercetools-website/*`) are excluded by design — they are not published.
+const PACKAGES = [
```
tdeekens: Is there a way to "parse" this by grepping package.json files in configured workspaces and excluding certain scopes? Would be easier to keep updated long-term.

tdeekens (approved):

ByronDWall (approved):
Looks good — clean script, well-scoped CI integration, and the REPORT_ONLY baseline strategy is the right approach. Approving.

ByronDWall:
**Minor: unbounded parallelism in `Promise.all`**

`scripts/check-package-shape.mjs:148` fires all ~29 packages through `pnpm pack` + two tool runs via `Promise.all` with no concurrency limit — that's potentially ~87 concurrent child processes. On resource-constrained CI runners this could cause contention or OOM.

A simple concurrency limiter (batches of 4–8) would be safer and likely faster in practice due to reduced context-switching. Not blocking — just something to consider hardening before flipping `REPORT_ONLY` to `false`.

misama-ct:
Addressed in 5ec7bad — `Promise.all` over all 29 packages replaced with a small worker pool sized by `os.availableParallelism()` (4 on standard GHA Ubuntu runners). Per-package output still buffered and printed in declared order. Thanks for the catch.

tdeekens (approved):

PR #4019, opened 2026-05-21
feat(deps): adopt pnpm catalogs + workspace:^ + constraint check (FEC-953)

misama-ct:
## 🎯 Summary

Adopt pnpm catalogs across `merchant-center-application-kit`, mirroring ui-kit's [FEC-938](https://commercetools.atlassian.net/browse/FEC-938) ([PR #3245](https://github.com/commercetools/ui-kit/pull/3245)).

- All internal `@commercetools-*` specifiers → `workspace:^` (155 rewrites, 27 files). `link-workspace-packages` stop-gap gone.
- `pnpm.overrides` moved into `pnpm-workspace.yaml` — single source of truth for pnpm config.
- 1 default catalog + 9 named (`apollo`, `build`, `flopflip`, `formatjs`, `lint`, `peer`, `react`, `repo`, `test`). 744+ specifier rewrites across 36 package.json files. `flopflip` and `formatjs` unsplit the cohorts that were stranded across default + `react`; `apollo` widened to absorb `apollo-link-*` (peer-pinned to @apollo/client). The "earns its name" rule for named catalogs is documented in the workspace YAML header.
- All 11 baseline drift cases resolved.
- `catalogMode: strict` + `cleanupUnusedCatalogs: true` enforce the catalog gate at `pnpm add` time.
- `scripts/check-workspace-constraints.js` enforces catalog adherence + cross-workspace drift (scans `dependencies`, `devDependencies`, `optionalDependencies`), wired into CI.
- `create-mc-app` flattens `workspace:^` and `catalog:*` across all dependency sections (`dependencies`, `devDependencies`, `peerDependencies`, `optionalDependencies`) when scaffolding from a template.
- Bumps `engines.pnpm` to `>=10` to match the actual `packageManager` pin.

Ticket: [FEC-953](https://commercetools.atlassian.net/browse/FEC-953).

## 🔄 Behavior change: `workspace:*` → `workspace:^`

Previously mixed `workspace:*` (apps + templates) and literal versions in publishable packages. Now unified on `workspace:^`. Consumer-visible deltas:

- **Published packages on npm:** cross-refs are now caret (`^27.6.2`) instead of pinned. With fixed versioning (every `@commercetools-*` bumps together) this is mostly moot — caret just lets transitive copies dedupe. Strictly safer than exact pinning.
- **Scaffolded user apps:** `create-mc-app` writes `^<releaseVersion>` (was: exact). A user running `pnpm install` later picks up the latest `27.x` — matches `pnpm add` convention.
- **Lockfile inside this monorepo:** unchanged.
- **Release flow:** verified — bumped one internal package and `pnpm install --offline` succeeded without a registry round-trip.

`update-package-json.ts` still rewrites `workspace:*` → exact for older template tags (pre-FEC-953 backwards-compat).

## ✅ Package-shape parity (acceptance criterion)

`pnpm check:package-shape` finding set on this branch is **identical** to the fresh baseline on `main` (9 findings, same set). Hands FEC-952 a clean baseline.

## 📦 Bundle byte-equivalence vs. `main`

Hashed every file under `packages/*/dist` + `packages-backend/*/dist` (716 files per side) and diffed. **77 files affected; all benign — zero runtime-bundle changes.**

| Package         | Files       | Why                                                                                                                                                                                                      |
| --------------- | ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `create-mc-app` | 3           | Intended: catalog/workspace resolver + yaml-library swap + inlined `package.json` reflecting catalog. **Adds runtime dep `yaml@2.8.1`** (zero transitives, 683 KB, scaffold-time only).                  |
| `mc-scripts`    | 72          | Preconstruct inlines `mc-scripts/package.json`; new `"catalog:..."` strings cascade chunk hashes. All renames — no same-path content diff. Runtime reads (`version`, `browserslist`) are byte-identical. |
| `sdk`           | 2 (`.d.ts`) | TypeScript reordered one property declaration. Same types.                                                                                                                                               |

## 🚫 Out of scope

- The 9 baseline package-shape findings → FEC-952.
- Strict hoisting (not set on app-kit; nothing to tighten).
- Cataloged dep major upgrades — adoption is intentionally a no-op for resolution.

## 🧪 Test plan

- [x] `pnpm install` clean
- [x] `pnpm build` / `typecheck` / `typecheck:cypress` / `typecheck:starters`
- [x] `pnpm test` (1061 frontend) + `pnpm test:node` (66 backend)
- [x] `pnpm lint`
- [x] `pnpm check:workspace-constraints` (new gate)
- [x] `pnpm check:package-shape` matches `main`
- [x] Release flow: `pnpm install --offline` succeeds after internal bump
- [x] `create-mc-app` resolver smoke-tested against actual templates locally
- [ ] CI green on this PR (incl. `test_*_starter_*_template_installation`)

[FEC-938]: https://commercetools.atlassian.net/browse/FEC-938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FEC-953]: https://commercetools.atlassian.net/browse/FEC-953?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


misama-ct:
@tdeekens @ByronDWall — want your take on the unification call before we merge this.

Repo previously had a mix of `workspace:*` (apps/templates) and literal versions (publishable packages). FEC-953 unifies everything on `workspace:^` to match the playbook + ui-kit's reference impl. For _published packages on npm_ this is essentially a no-op given our fixed-versioning model. But there's a real behavior change for **scaffolded customer apps**:

- **Before:** template's `workspace:*` got rewritten to exact `27.6.2` at scaffold time. A customer bootstraps an app, it works → it keeps working forever (pinned).
- **After:** template's `workspace:^` gets rewritten to `^27.6.2`. Customer bootstraps an app, it works at install time → but a `pnpm install` six months from now picks up the latest `27.x`. If we ship a regression in a future minor, the customer's previously-working app could break on their next install.

Yes, caret is the npm-convention default (`pnpm add foo` writes `^X.Y.Z`), and the playbook prescribes `workspace:^`. So in standards terms it's the _less_ surprising form. But I'm playing devil's advocate: we're trading a "works today → works forever (until they upgrade)" guarantee for a "works today → works until we mess up a patch release" floating range. That's a real shift in the contract we offer customers, and worth a deliberate yes/no rather than tagging along with the upstream migration by default.

Options if you don't like the change:

1. Keep `workspace:^` everywhere in the repo (per playbook), but have `create-mc-app` flatten `workspace:^` to **exact** `<releaseVersion>` (same as it does for `workspace:*`). Decouples internal repo convention from scaffolded-app guarantees.
2. Accept the caret (current PR state) and lean on changesets + the fixed-version cohort to make minor regressions rare.
3. Something else.

No strong preference from me — happy with whichever. Just want it to be an explicit decision.

packages/create-mc-app/src/tasks/update-package-json.ts
```diff
@@ -2,26 +2,169 @@ import fs from 'node:fs';
...
+  return out;
+}
+
+// `workspace:` specifier → rewriter that takes the release version and
+// returns the concrete spec. The supported set is data, not branching
+// logic — extend the table to handle new protocol additions.
+const WORKSPACE_REWRITERS: Record<string, (releaseVersion: string) => string> =
+  {
+    'workspace:*': (v) => v,
+    'workspace:^': (v) => `^${v}`,
```
misama-ct, line 42: @tdeekens @ByronDWall — flagging this rewriter specifically. Templates now declare `workspace:^` (was `workspace:*`), and the entry below expands it to `^x.y.z` in scaffolded apps (the `workspace:*` path above still resolves to an exact pin for older template tags).

Customer-facing contract genuinely shifts:

- **Before:** scaffold today → works today → works forever (pinned `27.6.2`).
- **After:** scaffold today → works today → next `pnpm install` floats inside `27.x`. If we ship a regression in a future minor, the previously-working app can break on reinstall.

Caret is the npm-convention default and matches the upstream playbook, so in standards terms it's the _less_ surprising form — but it is a real behavior change vs. the status quo. Long-form write-up + options in [#issuecomment-4543949949](https://github.com/commercetools/merchant-center-application-kit/pull/4019#issuecomment-4543949949). Just want this to be an explicit yes/no rather than tagging along by default.

tdeekens: Thanks. Makes sense to me. Can't recall us causing enough regressions in minors to warrant being careful.

tdeekens (approved):

ByronDWall (approved):

PR #4028, opened 2026-07-07
feat(agents): clip content to the modal's box

Sarah4VT:
#### Summary

Clip content to the modal's box so that when the container it is portaled into collapses (e.g. a resizable side panel steals all the width), the top bar's back/close buttons — which are laid out at a fixed left offset — cannot overflow and paint over the rest of the app.

#### Screenshots

Before  (X icon on top left)
Intake Agent Beta

After (X icon not visible)
image

### DIFF
--- a/.changeset/tired-bars-march.md
+++ b/.changeset/tired-bars-march.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-components': minor
+---
+
+clip modal content so when the container it is portaled into collapses, the top bar's back/close buttons cannot overflow
--- a/packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
+++ b/packages/application-components/src/components/modal-pages/internals/modal-page.styles.ts
@@ -104,6 +104,7 @@ export const ModalContent = styled(Content, {
   box-shadow: 0px 0px 40px 0px rgba(0, 0, 0, 0.1);
   outline: 0;
   z-index: inherit;
+  overflow: hidden;
 
   &[data-state='open'] {
     animation: ${({ size }) => getContentShowAnimation(size)} 300ms ease


mustafaasif2:
Thanks for the PR Sarah!

It seems there are a few more buttons that are showing on the agent panel when the width is small. I am not sure if they should be fixed in this PR or in another PR (or even if the changes should be implemented here or merchant-center-frontend)

It is visible here: https://mc.europe-west1.gcp.integration.commercetools.com/almond-40/settings/project/channels/a6e543d1-6150-4bd3-ba6c-8c2d17335633

Screenshot 2026-07-08 at 11 15 29

Sarah4VT:
> Thanks for the PR Sarah!
> 
> It seems there are a few more buttons that are showing on the agent panel when the width is small. I am not sure if they should be fixed in this PR or in another PR (or even if the changes should be implemented here or merchant-center-frontend)
> 
> It is visible here: https://mc.europe-west1.gcp.integration.commercetools.com/almond-40/settings/project/channels/a6e543d1-6150-4bd3-ba6c-8c2d17335633
> Screenshot 2026-07-08 at 11 15 29

When i test locally with my app kit changes, all the details pages are displaying correctly with the agent opened and “maximized”. Trying to figure out if there is a case that is not working. Since this is a fix in app kit and it hasn’t been merged, it wouldn’t be on staging yet. And even once it merges, then MC FE will need to pick it up. If you know of a page where it isn't working with the fix, I'm happy to investigate, but so far all the ones I've tested look good.

- intake agent from quotes page
- channels agent from Channels in project settings
- channels agent from Intervals in project settings
- promotions agent in Discount list pages

misama-ct:
This changed code is used by all 4 modal page variants in the MC, and those are instantiated in ~345 places.

`overflow:hidden` does not just prevent overflowing, it prevents scrolling as well and that can cause severe side-effects when applied to the wrong element so thank you both for being cautious and the thorough testing.

The good news is that scroll-issues most likely won't be a thing as there seems to be a separate wrapper element that takes care of scrolling. However, certain floating content (tooltips, popovers, dropdown content, etc.) triggered by DOM descendants and rendered inline will technically be clipped now. 

I am saying technically, because those page components usually seem to consume the whole available screen-estate which effectively means the modal clipping-edges are the natural browser-window boundaries. I don't know to which extend this changes with the agent open though. 

Do you have 15 minutes on Monday @Sarah4VT to show me the exact problem and how the agent stuff is integrated?

Sarah4VT:
> This changed code is used by all 4 modal page variants in the MC, and those are instantiated in ~345 places.
> 
> `overflow:hidden` does not just prevent overflowing, it prevents scrolling as well and that can cause severe side-effects when applied to the wrong element so thank you both for being cautious and the thorough testing.
> 
> The good news is that scroll-issues most likely won't be a thing as there seems to be a separate wrapper element that takes care of scrolling. However, certain floating content (tooltips, popovers, dropdown content, etc.) triggered by DOM descendants and rendered inline will technically be clipped now.
> 
> I am saying technically, because those page components usually seem to consume the whole available screen-estate which effectively means the modal clipping-edges are the natural browser-window boundaries. I don't know to which extend this changes with the agent open though.
> 
> Do you have 15 minutes on Monday @Sarah4VT to show me the exact problem and how the agent stuff is integrated?

Yes, I sure can!

If you want to take a peek beforehand:
- go to https://mc.europe-west1.gcp.integration.commercetools.com/almond-40/settings/project/channels
- Open the agent chat by clicking on the bottom right floating button
- Open the details page for a channel (outside the chat panel, but leave the chat panel open)
- Reduce the width of your browser and/or increase the width of the chat panel until the chat panel consumes the entire browser width.

You will see some of the content from the channel details modal appears overlapping the chat panel.

PR #4025, opened 2026-06-12
feat(application-shell): shell splitter + Region composition with cqw overlay scoping

ByronDWall:
## Summary

- Mount Nimbus `Splitter` in `ApplicationShellAuthenticated` with a `Region` target in the aside pane, lazy-loaded so apps without Nimbus get the existing layout unchanged
- Export `REGIONS.MC_RIGHT_PANEL` constant from `@commercetools-frontend/application-shell` for typed consumer access via `useRegion`
- Configure `NimbusProvider` with the shell's locale and react-router navigate function
- Add `@supports(width: 100cqw)` progressive enhancement to `PortalsContainer` so modals/overlays scope to the main pane when the aside is open
- Wire Nimbus bundler plugins (`tryLoadNimbusWebpackPlugin` / `tryLoadNimbusVitePlugin`) into all mc-scripts build configs with graceful MODULE_NOT_FOUND fallback
- Add `@commercetools/nimbus` as optional peerDep of `application-shell` and devDep for types
- Create dedicated `nimbus` named catalog in `pnpm-workspace.yaml` for the Nimbus ecosystem (nimbus, nimbus-icons, nimbus-tokens, @chakra-ui/react, slate family)

### Review hardening (e10ac62)

- **Replace Global CSS with `containerType` prop** on `Splitter.Main` — eliminates hardcoded ID selectors via `<Global>`, more resilient to Nimbus internal changes
- **Log errors in lazy load `.catch()`** — chunk failures (network, syntax) are now logged via `console.error` before falling back to the passthrough component
- **Simplify collapse state management** — route all collapse changes through the hook's `onCollapsedChange` option instead of dual-tracking with a manual ref override on `Splitter.Root`
- **Import `REGIONS` constant in VRT** — visual test now uses the exported constant instead of a hardcoded string, staying in sync if the value changes
- **Add unit tests** for the splitter component (containerType prop, hook integration, programmatic expand/collapse/toggle) and the lazy wrapper (error logging)

Closes FEC-1017

## Test plan
- [x] Build succeeds with Nimbus installed (splitter chunk loads)
- [x] Build succeeds without Nimbus installed (catch fallback renders children)
- [x] Splitter mounts collapsed at 0 width — no visual change from current layout
- [x] Consumer can fill the aside via `useRegion(REGIONS.MC_RIGHT_PANEL)` and expand/collapse it
- [x] Modals scope to the main pane via `cqw` when the aside is open
- [x] Existing overlay behavior is preserved when aside is collapsed
[...]

PR #4024, opened 2026-06-12
feat(mc-scripts): wire Nimbus bundler plugins into build configs

ByronDWall:
## Summary

- Add `@commercetools/nimbus/plugins/webpack` and `@commercetools/nimbus/plugins/vite` to all four mc-scripts build configurations (webpack dev/prod, Vite dev/prod)
- Plugins stub `@commercetools/nimbus` imports at build time when the package is not installed, enabling application-shell's lazy Nimbus chunks to resolve correctly
- Loaded via try/catch `require()` so builds succeed whether or not Nimbus is installed in the consuming app
- Add `@commercetools/nimbus` 3.1.0 as a root devDependency (build catalog) for development-time type access — no runtime dependency on any published package

Closes FEC-1016

## Test plan

- [x] Unit tests for `tryLoadNimbusWebpackPlugin` and `tryLoadNimbusVitePlugin` (6 tests covering loaded/not-loaded/options passthrough)
- [x] All mc-scripts tests pass (95 tests, 8 suites)
- [x] Typecheck passes
- [x] Workspace constraint check passes
- [ ] Verify mc-scripts webpack build succeeds with Nimbus installed
- [ ] Verify mc-scripts Vite build succeeds with Nimbus installed
- [ ] Verify mc-scripts builds succeed without Nimbus installed (external consumer)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4031, opened 2026-07-08
fix(mc-scripts): keep Nimbus optional when it is not installed (#4030)

### LINKED ISSUE #4030: Module not found: Error: Can't resolve '@commercetools/nimbus'
After creating a new Custom Application with the starter-typescript template, the build does not work.

Steps:

1. Run `npx @commercetools-frontend/create-mc-app@latest my-new-custom-application-project --template=starter-typescript`
2. `cd my-new-custom-application-project`
3. Run `yarn build`

Results in this error:

```
Creating an optimized production build...
Module not found: Error: Can't resolve '@commercetools/nimbus' in 'C:\tmp\test\my-new-custom-application-project\node_modules\@commercetools-frontend\application-shell\dist'


C:\tmp\test\my-new-custom-application-project\node_modules\@commercetools-frontend\mc-scripts\cli\dist\commercetools-frontend-mc-scripts-cli.cjs.dev.js:53
  throw err;
  ^

Error: Failed to compile
    at Object.run [as default] (C:\tmp\test\my-new-custom-application-project\node_modules\@commercetools-frontend\mc-scripts\dist\build-03da8462.cjs.dev.js:106:11)
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
    at async Command.<anonymous> (C:\tmp\test\my-new-custom-application-project\node_modules\@commercetools-frontend\mc-scripts\cli\dist\commercetools-frontend-mc-scripts-cli.cjs.dev.js:95:5)

Node.js v24.13.0
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

misama-ct:
Fixes #4030.

## Problem

A fresh Custom Application (`npx create-mc-app@latest my-app --template=starter-typescript`) fails `yarn build` with:

```
Module not found: Error: Can't resolve '@commercetools/nimbus'
  in '…/node_modules/@commercetools-frontend/application-shell/dist'
```

Reproduced end-to-end on the released **27.7.0** (`mc-scripts@27.7.0` + `application-shell@27.7.0`, Nimbus not installed).

## Root cause

27.7.0 shipped Nimbus as an *optional* dependency of the shell (nimbus #1599 bundler plugins + #4024 mc-scripts wiring + #4025 shell splitter). The intended contract — *stub Nimbus imports so the build succeeds, guard against `undefined` at runtime* — was broken in two places:

1. **Build fails.** `mc-scripts` loaded the stubbing plugin via `require('@commercetools/nimbus/plugins/webpack')`. That plugin (and the stub it redirects to) live **inside Nimbus**, so it cannot load precisely when Nimbus is absent — the one case it exists for. The loader returned `null`, nothing stubbed the import, and webpack could not resolve the splitter's static `@commercetools/nimbus` import. (PR #4024's checklist item *"Verify mc-scripts builds succeed without Nimbus installed (external consumer)"* was left unchecked; in the monorepo Nimbus is a hoisted devDep so it silently worked there.)
2. **Runtime guard missing.** The splitter's async wrapper only `.catch`-ed import *rejection*. An empty stub imports *successfully*, so once (1) is fixed with an empty stub the splitter would still mount and crash on `useResponsiveSplitterSizes(undefined)`. FEC-1017's planned availability guard had been simplified away.

## Fix (both packages in this repo; Nimbus untouched)

- **`@commercetools-frontend/mc-scripts`** — ships its own empty `nimbus-stub.js`. When `@commercetools/nimbus` is not installed, `loadNimbus{Webpack,Vite}Plugin` registers an mc-scripts-owned fallback that redirects Nimbus runtime imports to that stub (`NormalModuleReplacementPlugin` for webpack, a virtual module for Vite), mirroring Nimbus's own plugins but resolvable when Nimbus is absent. When Nimbus **is** installed it defers to Nimbus's own no-op plugin — unchanged.
- **`@commercetools-frontend/application-shell`** — the splitter exposes a `hasNimbus` flag; the async wrapper renders a passthrough when it is `false`, so no Nimbus code path is mounted without Nimbus.

## Verification

- **965 unit tests pass**, including new coverage for the fallback (webpack/Vite regex + stub target) and the `hasNimbus` gate.
- Reproduced the reporter's failure end-to-end on stock **27.7.0** (byte-for-byte the same error), then verified the fix by mocking the `npx` flow against this branch's real `pnpm pack` tarballs (`pnpm pack` resolves `catalog:`/`workspace:` like the release pipeline; `nimbus-stub.js` is confirmed in the tarball `files`).

### Build matrix — fresh `starter-typescript` scaffold, real tarballs of this branch

**Nimbus _absent_ (the scenario in #4030) — the fix:**

| | npm | pnpm | yarn |
|---|---|---|---|
| **webpack** | ✅ | ✅ | ✅ |
| **vite** | ✅ | ✅ | ✅ |

All 6 build cleanly (`Compiled successfully`, exit 0). webpack redirects Nimbus imports to the empty `nimbus-stub.js`; Vite inlines an equivalent virtual module (confirmed the Vite bundler engaged and Rollup raised no `@commercetools/nimbus` resolution error). The emitted `app-shell.js` carries the `hasNimbus ? splitter : Passthrough` guard, so the splitter is never mounted without Nimbus.

The webpack fallback resolves the stub to an **absolute path** from the app root (where `mc-scripts` is a declared dependency) rather than a bare specifier — a bare specifier would be re-resolved from `application-shell`'s context, which cannot see `mc-scripts` under **pnpm's strict `node_modules`**. This is why pnpm is green.

**Nimbus _present_ (Nimbus 3.2.0 installed) — regression / opt-in path:**

| | npm | pnpm | yarn |
|---|---|---|---|
| **webpack** | ✅ | ✅ | ⚠️¹ |
| **vite** | ✅ | ✅ | ⚠️¹ |

npm & pnpm bundle real Nimbus (splitter chunk ~1.3 MB, stub not used) — the happy path is intact.

¹ The two yarn cells fail with `Can't resolve '@chakra-ui/react/styled-system'` (then `slate-dom`). This is **not introduced by this PR**: Nimbus 3.2.0 declares `@chakra-ui/react`, `slate*`, `nimbus-icons`, `nimbus-tokens` as **peerDependencies**; npm and pnpm auto-install/wire peers, **yarn classic does not**, so Nimbus can't resolve its own deps. When Nimbus is present my fallback is never constructed (mc-scripts returns Nimbus's own plugin), and the replacement regex only matches `@commercetools/nimbus` — it cannot affect Chakra/Slate resolution. **Follow-up (separate from #4030):** yarn-classic Custom App users who adopt Nimbus need Nimbus's peer graph installed — worth template guidance or a Nimbus-side note.

### What is not covered

- **dev/serve** builds (only production `build` was run) and package managers beyond npm/pnpm/yarn (e.g. Yarn Berry/PnP).
- The literal `npx create-mc-app@latest … && yarn build` — it resolves `^27.7.0` from the `latest` dist-tag, so it only picks up this fix once a release bumps `latest`.

### Reproduce it yourself before merge

Because `create-mc-app@latest` scaffolds a template pinned to `^27.7.0` and `yarn` resolves that from the `latest` dist-tag, the vanilla `npx create-mc-app@latest … && yarn build` won't pick up this fix until a release bumps `latest`. To exercise the fix now, override the two packages to a build of this branch:

```bash
# from this branch's checkout
pnpm --filter @commercetools-frontend/mc-scripts --filter @commercetools-frontend/application-shell exec pnpm pack --pack-destination /tmp/preview

# fresh app, Nimbus intentionally NOT installed
npx --yes @commercetools-frontend/create-mc-app@latest app \
  --template starter-typescript --entry-point-uri-path myapp \
  --initial-project-key my-project-key --yes --skip-install
cd app
# pin the two deps + resolutions (application-shell is also a transitive dep) to the tarballs, then:
yarn install && yarn build   # expect: Compiled successfully
```

Alternatively, comment `[preview_deployment]` on this PR to publish snapshot versions to npm, then install those (`@<preview-tag>`) into a fresh scaffold — same result, hosted on npm.

### DIFF
--- a/.changeset/nimbus-optional-build-fallback.md
+++ b/.changeset/nimbus-optional-build-fallback.md
@@ -1,0 +1,17 @@
+---
+'@commercetools-frontend/mc-scripts': patch
+'@commercetools-frontend/application-shell': patch
+---
+
+Fix Custom Application builds failing with
+`Module not found: Can't resolve '@commercetools/nimbus'` when
+`@commercetools/nimbus` is not installed (e.g. a fresh
+`create-mc-app --template=starter-typescript` project).
+
+`@commercetools/nimbus` is an optional dependency of the application shell.
+Previously the build only stayed unaffected when the app already had Nimbus
+installed, because the stubbing plugin was loaded from Nimbus itself and could
+not run when Nimbus was absent. `mc-scripts` now ships its own build-time
+fallback that stubs Nimbus imports when the package is not installed, and the
+shell splitter degrades to a passthrough at runtime when Nimbus is absent — so
+Nimbus stays genuinely optional for both webpack and Vite builds.
--- a/packages/application-shell/src/components/application-shell-splitter/application-shell-splitter-wrapper.spec.tsx
+++ b/packages/application-shell/src/components/application-shell-splitter/application-shell-splitter-wrapper.spec.tsx
@@ -24,5 +24,46 @@ describe('ApplicationShellSplitterWrapper', () => {
       expect(source).toContain('.catch(');
       expect(source).toContain('children');
     });
+
+    it('source gates the splitter on `hasNimbus`, falling back to passthrough', () => {
+      const fs = require('fs');
+      const path = require('path');
+      const source = fs.readFileSync(
+        path.join(__dirname, 'application-shell-splitter.async.tsx'),
+        'utf8'
+      );
+
+      expect(source).toContain('hasNimbus');
+      expect(source).toContain('Passthrough');
+    });
+  });
+
+  describe('hasNimbus availability flag', () => {
+    afterEach(() => {
+      jest.resetModules();
+    });
+
+    it('is true when @commercetools/nimbus resolves', () => {
+      jest.isolateModules(() => {
+        jest.doMock('@commercetools/nimbus', () => ({
+          NimbusProvider: () => null,
+          Splitter: { Root: () => null },
+          Region: () => null,
+          useResponsiveSplitterSizes: () => ({ rootProps: {} }),
+        }));
+        const { hasNimbus } = require('./application-shell-splitter');
+        expect(hasNimbus).toBe(true);
+      });
+    });
+
+    it('is false when @commercetools/nimbus is stubbed to an empty module', () => {
+      jest.isolateModules(() => {
+        // Mirrors what the mc-scripts bundler fallback produces when Nimbus is
+        // not installed: an empty module, so every named import is `undefined`.
+        jest.doMock('@commercetools/nimbus', () => ({}));
+        const { hasNimbus } = require('./application-shell-splitter');
+        expect(hasNimbus).toBe(false);
+      });
+    });
   });
 });
--- a/packages/application-shell/src/components/application-shell-splitter/application-shell-splitter.async.tsx
+++ b/packages/application-shell/src/components/application-shell-splitter/application-shell-splitter.async.tsx
@@ -1,12 +1,21 @@
 import { type ReactNode, lazy, Suspense } from 'react';
 import { useHistory } from 'react-router-dom';
 
+const Passthrough = ({ children }: { children: ReactNode }) => <>{children}</>;
+
 const LazyApplicationShellSplitter = lazy(() =>
   import(
     './application-shell-splitter' /* webpackChunkName: "application-shell-splitter" */
-  ).catch(() => ({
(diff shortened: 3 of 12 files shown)


misama-ct:
[preview_deployment]

tdeekens (approved):

valoriecarli (approved):

ByronDWall (approved):

PR #4037, opened 2026-07-16
Add main container portal target for save toolbar

ByronDWall:
## Why

When the agent side panel is open, fixed-position components like the SaveToolbar span the full viewport width, rendering underneath the panel. The current approach in [commerce-agents#570](https://github.com/commercetools/commerce-agents/pull/570) uses a CSS custom property (`--mc-right-panel-width`) published via a ResizeObserver, requiring cross-repo coordination between commerce-agents and mc-fe. Discussion on that PR converged on the idea that the application shell should provide a portal target instead, since it owns the splitter layout.

## Summary

- Add `MC_MAIN_CONTAINER_PORTAL_ID` constant to `@commercetools-frontend/constants`
- Add a portal target `<div>` inside `MainContainer` (last child, `position: sticky; bottom: 0; pointer-events: none`) so components can portal into the MC content area and automatically constrain their width to the main pane
- Add visual test specs demonstrating the portal-based toolbar in both collapsed and expanded splitter states

The consuming side (SaveToolbar using `createPortal`) is in the companion mc-fe PR.

> To see the splitter demo with the toolbar, you need to pull down this branch and run `pnpm visual-testing-app:start` locally, then navigate to `localhost:5173/shell-splitter`

## Test plan

- [x] Visual test specs render correctly at `/shell-splitter` (toolbar constrained to main pane when aside is open)
- [x] Existing splitter unit tests pass
- [x] No regressions in modal/dialog behavior (PortalsContainer unchanged)

### DIFF
--- a/.changeset/save-toolbar-portal-target.md
+++ b/.changeset/save-toolbar-portal-target.md
@@ -1,0 +1,6 @@
+---
+'@commercetools-frontend/constants': minor
+'@commercetools-frontend/application-shell': minor
+---
+
+Add a portal target (`mc-main-container-portal`) inside `MainContainer` so that fixed-position components like `SaveToolbar` can portal into the MC content area and automatically constrain their width to the main pane, excluding the agent side panel.
--- a/packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
+++ b/packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
@@ -18,7 +18,11 @@ import {
   selectUserLanguageFromStorage,
   type TApplicationContext,
 } from '@commercetools-frontend/application-shell-connectors';
-import { DOMAINS, LOGOUT_REASONS } from '@commercetools-frontend/constants';
+import {
+  DOMAINS,
+  LOGOUT_REASONS,
+  MC_MAIN_CONTAINER_PORTAL_ID,
+} from '@commercetools-frontend/constants';
 import type { TAsyncLocaleDataProps } from '@commercetools-frontend/i18n';
 import { AsyncLocaleData } from '@commercetools-frontend/i18n';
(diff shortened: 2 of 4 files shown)


tylermorrisford:
**Note — not a blocker, and likely not app-kit's responsibility to solve, but worth flagging for the consumer side:**

When the agent panel is **collapsed**, there's no aside pane taking width, so `MainContainer` spans the full content area and a `width: 100%` toolbar portaled into the target runs all the way to the right edge — which is where the floating agent-panel trigger button sits (bottom-right). Since the SaveToolbar right-aligns its primary action, the **Save button can end up underneath the trigger** in the collapsed state.

The CSS-var approach in commercetools/commerce-agents#570 handled this explicitly — its mock reserved `RIGHT_INSET = 96px` with the note that it "must clear the floating trigger button (24px from the edge + 48px wide = 72px footprint) plus a comfortable margin."

Since app-kit doesn't (and arguably shouldn't) know about the trigger, the right-inset most naturally belongs on the consumer side — the mc-fe `SaveToolbar`, or the trigger reserving its own footprint — rather than being baked into the portal target here. The portal target itself looks correct; just noting this so it doesn't fall through the cracks, since the splitter demo here has no trigger button and so can't surface it.

ByronDWall:
> **Note — not a blocker, and likely not app-kit's responsibility to solve, but worth flagging for the consumer side:**

Yeah, this PR does not even have the real save toolbar available. Handling designs with multiple interaction patterns is not app-kit or nimbus' concern, this is meant to give consumers the tools to build the view not to have the view pre-built for the consumer. The FAB should be handled by whoevers implementing it in either mc-fe or commerce-agents

mustafaasif2 (approved):
Thank you! 💯

tylermorrisford (approved):
I like avoiding the resizeObserver!

PR #4041, opened 2026-07-28
fix: use accurate prop

tylermorrisford:
The "secondary" variant was removed from Nimbus Button; both usages in shell-splitter.visualroute.tsx are updated to "subtle" to unblock the typecheck CI failure


ByronDWall (approved):
_subtle_

PR #4044, opened 2026-08-03
fix(supply-chain): adopt strict minimumReleaseAge posture

tylermorrisford:
Flip the pnpm age-gate from the soft fall-back to the strict posture, aligning with nimbus (#1859):

  - minimumReleaseAgeStrict: false -> true — no silent fall-back (or auto-generated exclude) when no version satisfies a range within the 24h gate
  - minimumReleaseAgeIgnoreMissingTime: true -> false — enforce the gate even when a package's registry metadata omits a publish time

Refs: FEC-1199 (epic FEC-964 — Supply Chain Hardening)

### DIFF
--- a/package.json
+++ b/package.json
@@ -182,7 +182,7 @@
     "yaml": "catalog:"
   },
   "engines": {
-    "node": ">=22",
+    "node": ">=24",
     "npm": ">=6",
     "pnpm": ">=11"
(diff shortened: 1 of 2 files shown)


tylermorrisford:
closing this for now to avoid potential issues

PR #4046, opened 2026-08-04
FEC 1249 + friends - Stand up Storybook & Chromatic VRT for `application-components`

valoriecarli:
Covers FEC-1249 & FEC-1259 through FEC-1268.

app-kit's component VRTs run on Percy: `visual-testing-app` renders each component at its own URL & a Puppeteer script screenshots it. 29 route/spec pairs, 64 snapshots.

Chromatic screenshots **Storybook stories** instead. app-kit had no Storybook, so this PR builds one, wires up Chromatic & converts all 29 routes.

Chromatic [Builds](https://www.chromatic.com/builds?appId=6a7214131a71921f118f4b58) & [Library](https://www.chromatic.com/library?appId=6a7214131a71921f118f4b58)

**Percy stays running & untouched. Nothing is deleted rn**

## Structure

```mermaid
graph LR
  subgraph pub["packages/* · published to npm"]
    C["components"]
  end
  subgraph priv["storybook/ · private, never published"]
    S["stories + fixtures"]
    H["decorators · helpers · stubs"]
  end
  C -->|"public entrypoints"| S
  H --> S
  S --> B["storybook build"]
  B --> X["Chromatic · app-kit-components"]
```

## What's in it

New private `storybook/` workspace: Storybook config, two global decorators, the `VisualSpec` / `VisualSpecGroup` layout helpers, 27 story files & the fixtures they share.

`.github/workflows/chromatic.yml` with TurboSnap & a changed-files gate, reusing the existing `./.github/actions/ci` composite for setup.

**All 29 routes, 64 of 64 Percy snapshots**, across four packages: 26 in `application-components`, plus `notifications` (`react-notifications`), `shell-splitter` (`application-shell`) & `application-icons` (`assets`).

## Decisions

**Stories live in `storybook/`, not next to components.** Colocated stories need the `VisualSpec` helpers, which live in the private workspace, so published packages ended up importing an alias only Storybook defines. Moving them out keeps published packages free of test scaffolding & points the dependency one way.

**They import public entrypoints** (`@commercetools-frontend/*`), never deep paths, so the suite pins the same API consumers use.

**Decorators are global.** Every MC view assumes Apollo, FlopFlip, react-intl, a router & the theme provider. Wiring that per story is a failure mode; wiring it once isn't.

**Capture by default, no viewport pinned.** Every story exists to be snapshotted, so there's nothing to opt into & Chromatic's default width applies everywhere.

**Four shims** (`window.app`, `process.env`, msw, l10n's locale list) stand in for things the MC runtime or the repo's babel pipeline provides & Vite doesn't. Each fails silently when absent, so each carries a comment.

## Parity

Reconciled against Percy build #12615 on `main`: **64 snapshots, all 64 mapped to a story**, state counts matching per component.

67 stories total. The extra 3 are deliberate — `page-content-container-wide` gains two sub-routes the route defined but the spec never visited (so they never had a baseline), & `public-page-layout` splits one Percy snapshot into two stories. A few states were also corrected where the source route was wrong.

Sprinkled in some objective flair to make it visually appealing. 

## Scope

Nothing gates on Chromatic yet. Parity sign-off & cleanup is FEC-1269; the Cypress playground & Percy's removal are FEC-1270.


.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,96 @@
...
+          (steps.changed-files.outputs.any_changed != 'true' || github.head_ref
+          == 'changeset-release/main') && github.event_name !=
+          'workflow_dispatch'
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          echo "::notice::No UI-affecting changes detected. Skipping the Chromatic visual regression build."
+          gh api --method POST "repos/${{ github.repository }}/statuses/${{ github.event.pull_request.head.sha || github.sha }}" \
+            -f state=success \
+            -f context='UI Tests' \
```
ByronDWall: Small thing worth double checking: this posts the skip status under the context `UI Tests`, but the real Chromatic run reports as `UI Tests: app-kit-components` (that's the check name showing on this PR's own run). If main's branch protection ever adds this as a required check, it would most likely be added under the real context string, and this skip path would not satisfy it. Might be worth matching the two contexts now while it's cheap to change.

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,96 @@
...
+        if: (steps.changed-files.outputs.any_changed == 'true' ||
+          github.event_name == 'workflow_dispatch') && github.head_ref !=
+          'changeset-release/main'
+        uses: ./.github/actions/ci
+
+      - name: Run Chromatic
+        if: (steps.changed-files.outputs.any_changed == 'true' ||
+          github.event_name == 'workflow_dispatch') && github.head_ref !=
+          'changeset-release/main'
+        uses: chromaui/action@94713c544284a14195de3b50ef24301579f1877e # v18
```
ByronDWall: Nit: this pin resolves to `v18.0.1`, but the floating `v18` tag has since moved on to `v18.1.0`, so the `# v18` comment reads as slightly ahead of what's actually pinned. Might be nice to spell out the exact patch version here, the same way the other two pins in this file are annotated.

storybook/AGENTS.md
```diff
@@ -0,0 +1,65 @@
...
+- **`VisualSpec` and `VisualSpecGroup`** (`src/helpers/`) wrap each captured state. Picking the wrong one is the most common way a story renders wrong.
+- Chromatic runs against the **`app-kit-components`** project. The Cypress playground suite is a separate Chromatic project and does not involve this workspace.
+
+## How To Work Here
+
+| Task          | Command                                                      |
+| ------------- | ------------------------------------------------------------ |
+| Dev server    | `pnpm storybook:start` — port 6006                           |
+| Build         | `pnpm storybook:build` — this is what Chromatic builds in CI |
+| Browse hosted | TODO: paste the Chromatic permalink for `main`               |
```
ByronDWall: Just flagging this TODO so it doesn't get lost, might be worth filling in the Chromatic permalink for main before or shortly after merge.

storybook/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "@emotion/styled": "catalog:build",
+    "@flopflip/react-broadcast": "catalog:flopflip",
+    "@rollup/plugin-graphql": "catalog:build",
+    "@storybook/react-vite": "catalog:",
+    "@types/node": "catalog:",
+    "@types/react": "catalog:react",
+    "@types/react-dom": "catalog:react",
+    "@types/react-router-dom": "catalog:react",
+    "@vitejs/plugin-react": "catalog:build",
+    "chromatic": "catalog:",
```
ByronDWall: Curious about this one: `chromatic` is listed as a workspace dependency here (pinned to `^17.7.2` via the catalog), but CI drives the actual build through `chromaui/action` pinned to `v18.0.1`, and the docs explicitly say not to run Chromatic locally. Wondering if this dependency is still needed anywhere, or if it's a leftover from an earlier iteration of the setup. If it is needed for something, a short comment on what would help future readers.

storybook/.storybook/main.ts
```diff
@@ -0,0 +1,65 @@
...
+         Without \`app\`, \`custom-views-selector/index.ts\` lazy-resolves to a null component;
+         without \`process\`, \`use-custom-views-connector\` throws reading \`process.env\`. -->
+    <script>
+      window.app = { applicationName: 'storybook', mcApiUrl: 'http://localhost:8080' };
+      window.process = window.process || { env: {} };
+    </script>
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Inter:wght@300;400;500;600;700&family=Open+Sans:ital,wght@0,300;0,400;0,700;1,400;1,700&display=swap"
```
ByronDWall, line 31: Small flakiness concern here: these font links pull from Google Fonts over the network at snapshot time, with nothing that explicitly waits for the fonts to finish loading before Chromatic captures. That can be a source of intermittent diffs if the fonts load slowly or inconsistently in CI. Probably fine to leave as is for now, but worth keeping in mind if flaky snapshots show up, and self hosting the font files could be a fallback if that happens.

ByronDWall (commented):
This is a genuinely well built and documented VRT setup, thanks for putting it together. The three tier AGENTS.md docs make the reasoning behind each decision easy to follow, and the CI gating logic (draft skip, TurboSnap, the self posted skip status) all held up when I traced through it.

I left a handful of small suggestions inline, mostly polish items rather than blockers. The one thing worth resolving before merge regardless is that the PR currently shows as CONFLICTING against main, with the drift concentrated in pnpm-lock.yaml, so it will need a rebase and lockfile regen either way.

.github/workflows/chromatic.yml
valoriecarli: good catch, i thought about this while working the other package. updated in https://github.com/commercetools/merchant-center-application-kit/pull/4046/commits/8d146a9a02f67b6e38131654601756b66047f3a2

.github/workflows/chromatic.yml
valoriecarli: updated in https://github.com/commercetools/merchant-center-application-kit/pull/4046/commits/8d146a9a02f67b6e38131654601756b66047f3a2

storybook/AGENTS.md
valoriecarli: updated in https://github.com/commercetools/merchant-center-application-kit/pull/4046/commits/8d146a9a02f67b6e38131654601756b66047f3a2

storybook/package.json
valoriecarli: oh noice. i did forget to remove this once i laid down the law. updated in https://github.com/commercetools/merchant-center-application-kit/pull/4046/commits/8d146a9a02f67b6e38131654601756b66047f3a2

storybook/.storybook/main.ts
valoriecarli: interesting, i see chromatic mentions this possibility as well. I have added this to my notes for potential future enhancements. 

also note that i went ahead and added `docs/chromatic-components.md` in https://github.com/commercetools/merchant-center-application-kit/commit/8d146a9a02f67b6e38131654601756b66047f3a2 i couldnt take it.

storybook/src/apollo-client.ts
```diff
@@ -0,0 +1,71 @@
+import { ApolloClient, InMemoryCache } from '@apollo/client';
+import { v4 as uuidv4 } from 'uuid';
+import FetchCustomViewsQuery from '../../packages/application-components/src/components/custom-views/custom-views-selector/fetch-custom-views-by-locator.settings.graphql';
```
tylermorrisford, line 5: Nit: this is the one place in the workspace that reaches into a package's internals instead of its public entrypoint. Makes sense — there's no way to export a raw `.graphql` file from the package index — but if that component's directory ever gets reorganized, this breaks with no warning. A comment here explaining the constraint would help whoever trips over it later.

storybook/src/globals.css
```diff
@@ -0,0 +1,13 @@
...
+@import '@commercetools-uikit/design-system/materials/custom-properties.css';
+@import '@commercetools-uikit/design-system/materials/resets.css';
+
+html,
+body {
+  font-size: var(--font-size-30);
+}
+
+/* Modals autofocus their close button, and the ring would show up in every snapshot. */
+:focus-visible {
```
tylermorrisford, line 11: This suppresses focus rings everywhere, not just on the modal close button. Totally understand why — autofocus on open means a ring shows up in every modal snapshot. But as written, we can never VRT a focused state if we want to down the road. Would it be feasible to scope this more narrowly, or is there a Chromatic-side way to handle the autofocus? Happy to leave it if it's genuinely not worth the complexity.

storybook/src/stubs/supported-locales.ts
```diff
@@ -0,0 +1,5 @@
+// `packages/l10n` reads this through `babel-plugin-preval`, which Vite doesn't run,
+// so the import would land on a CJS file with no default export. Synced by hand.
+const supportedLocales = ['en', 'de', 'es', 'fr-FR', 'pt-BR'];
```
tylermorrisford, line 3: "Synced by hand" is the thing to watch — if someone adds a locale to `packages/l10n` and misses this file, the locale selector renders with the wrong options in Chromatic and nothing fails loudly. Worth a comment here listing the source of truth so it's obvious what to check when updating.

tylermorrisford (approved):
que bonita! a couple nits inline, but this is yuge

storybook/src/apollo-client.ts
valoriecarli: thanks! added in https://github.com/commercetools/merchant-center-application-kit/pull/4046/commits/cfb329ea93229f762910e286a7ea658d7626ae20

storybook/src/stubs/supported-locales.ts
valoriecarli: used my words here: https://github.com/commercetools/merchant-center-application-kit/pull/4046/commits/cfb329ea93229f762910e286a7ea658d7626ae20

storybook/src/globals.css
valoriecarli: thanks, as discussed, this is intentional due to the scope of this migration.

PR #4066, opened 2026-08-07
FEC-1278 - `e2e-playground` Chromatic snapshots

valoriecarli:
## Why

Migrating visual regression testing off Percy. This covers the Cy e2e playground;
`application-components` is a separate project (#4046).

Percy is untouched & keeps running, with both tools snapshotting in the same Cy pass. 
Percy comes out of the monorepo in a later sweep.

Chromatic [Builds](https://www.chromatic.com/builds?appId=6a72232dda895583557c8a27) & [Library](https://www.chromatic.com/builds?appId=6a72232dda895583557c8a27)

*****

## What this adds

**Dependencies**

- `chromatic` & `@chromatic-com/cypress`
- `cypress` 14.5.4 to **15.20.0**, required by `@chromatic-com/cypress@0.13.x`, which reads
  its config through `Cypress.expose()`. 

**Wiring**

- `installPlugin()` in `cypress.config.ts` & the support import in `cypress/support/e2e.ts`, both gated on `CHROMATIC_VRT`. The root config is shared by all 3 Cy suites & only the playground archives; the two gates must stay in sync, or the starter-template suites fail on an unregistered task.
- Viewports pinned in Cy config, since Chromatic re-renders server-side & ignores
  `cy.viewport()`. 
  
**Coverage: 11 snapshots**

Two tests opt out via `expose: { disableAutoSnapshot: true }`; both still run & assert:

- `should log out with reason "user"`, which asserts an `href` & nothing visual.
- The Custom View iframe test. Chromatic archives only the top-level document, so the
  iframe comes out blank. 

**CI (`test_playground`)**

- Upload step runs `chromatic --cypress --exit-zero-on-changes`, so a visual diff fails the  `UI Tests: app-kit-e2e-playground` check rather than the job & the coverage steps still run.
- Path gate on `packages/`, `playground/`, `cypress/`, `cypress.config.ts` & `pnpm-lock.yaml`, plus a changesets-release bypass. When skipped, the workflow posts the passing status itself, since a required check that never reports would wedge the PR.
- `fetch-depth` 2 to 0, so the gate can diff against the merge base.

Runbook: `docs/chromatic-e2e-playground.md`.

## Follow-up

add `UI Tests: app-kit-e2e-playground` to the required status checks on `main`, next to the existing `percy/app-kit-e2e-playground`.


tylermorrisford (approved):
i like it

PR #4120, opened 2026-08-27
Prefer CSS to JS DOM trolling

tylermorrisford:
This pull request updates the `@commercetools-frontend/application-shell` package to improve how the main container is hidden when a dialog is open. Instead of using JavaScript to detect dialogs, it now leverages the CSS `:has()` selector for a cleaner and more efficient approach.

**CSS/DOM Handling Improvements:**

* Replaced JavaScript DOM sniffing with the CSS `:has()` selector to control the visibility of the main container when a dialog is present, resulting in simpler and more maintainable code. (`.changeset/hot-ducks-obey.md`, [.changeset/hot-ducks-obey.mdR1-R5](diffhunk://#diff-604bde3a5d81d9f39c43293292355f2fd7780e58f452a767cf0a79d45592cfd8R1-R5))
* Added a `Global` style using Emotion in `application-shell-authenticated.tsx` to automatically hide the main container (`#${MC_MAIN_CONTAINER_PORTAL_ID}`) when a dialog (`[role='dialog']`) is open.

**Dependency Updates:**

* Updated Emotion imports in `application-shell-authenticated.tsx` to include `Global` for injecting the new global style.

PR #4132, opened 2026-08-31
fix(application-shell): move portal target above splitter

ByronDWall:
## Problem

PR #4120 placed `#mc-main-container-portal` inside the grid (inside `<ApplicationShellSplitter>`). This causes two issues:

1. **z-index occlusion**: the portal target's `z-index: 9999` is below the modal portals container's `z-index: 10000`, so the save toolbar is hidden behind any active modal page.
2. **splitter clipping**: the portal target is inside the splitter's main pane, so `position: fixed` content portaled into it doesn't span across the agent panel aside.

## Fix

- Move `#mc-main-container-portal` outside `<ApplicationShellSplitter>` so it is not constrained by the splitter's layout or `overflow: hidden`.
- Switch from grid positioning to `position: fixed; bottom: 0; left: 0; right: 0` so the toolbar spans the full viewport width (across both splitter panes).
- Bump `z-index` from `9999` to `10001` so content portaled here renders above the modal portals container (`z-index: 10000`).

## Sibling PR

- mc-fe: https://github.com/commercetools/merchant-center-frontend/pull/20986 (portals SaveToolbar into this target via `createPortal`)

### DIFF
--- a/.changeset/save-toolbar-above-splitter.md
+++ b/.changeset/save-toolbar-above-splitter.md
@@ -1,0 +1,5 @@
+---
+'@commercetools-frontend/application-shell': patch
+---
+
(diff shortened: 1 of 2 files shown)


ByronDWall:
Superseded by #4136, which places the portal target in the splitter async wrapper instead of outside the splitter (avoids the IntlProvider crash from sibling-of-splitter rendering).

PR #4138, opened 2026-09-01
feat(mc-html-template): HTML skeleton, non-blocking resources, and view transitions

ismaelocaramelo:
#### Summary

Merchant Center apps show a spinner on a blank page while React boots — around 900ms warm, a few seconds cold. This replaces that with a skeleton of the real app chrome (header, dark sidebar, content area) drawn straight from the HTML document, so something structural is on screen almost immediately.

It also makes both webfonts and the app's own CSS non-blocking, and opts the document in to cross-document view transitions so moving between MC apps crossfades instead of flashing white.

Jira: [FEC-1298](https://commercetools.atlassian.net/browse/FEC-1298) · Plan: [`perceived-loading-performance/plan.md` §B](https://github.com/commercetools/mc-foundation-planning/blob/main/perceived-loading-performance/plan.md#b-html-skeleton--non-blocking-resources--view-transitions)

#### Description

**Why the CSS handling looks unusual.** The ticket suggested the usual `onload="this.rel='stylesheet'"` preload trick. That cannot work here: production `script-src` is `'self'` plus hashes of three known inline scripts, and CSP hashes do not cover inline event-handler attributes — the preloads would never upgrade and the app would render completely unstyled, which local dev would not show you because dev gets `'unsafe-inline'`. So the upgrade happens inside `loading-screen.js` instead, whose hash `process-headers.ts` derives from the file itself. No security header changes.

**The skeleton is gated on three things** so it does not fight what React renders a moment later: a cached session (`localStorage.isAuthenticated`), the route (`/login`, `/logout` and `/account` render no navbar; Custom Views get the old spinner), and viewport width (a pinned-open menu collapses at ≤1200px, matching `useNavbarStateManager`). Sizes and colours come from the same constants the real chrome uses.

**On the fonts.** Both Open Sans and Inter are now preloads with `<noscript>` fallbacks. Inter needed care because `NimbusProvider` renders a byte-identical stylesheet and React 19 dedupes by matching `link[rel="stylesheet"][href]`, which a preload does not match — so `data-nimbus-fonts` is kept on the link as the marker Nimbus uses to recognise a host-owned font, and hosts that mount the provider themselves should pass `loadFonts={false}`. Leaving it on the default turns out to be harmless: the URL is identical, so the extra link resolves from the already-warm preload (`transferSize` 0, no second request) and first paint is not blocked either way. `application-shell` needed no change — its only provider already passes `loadFonts={false}`.

The knock-on effect is the bigger win. `loading-screen.js` is a parser-inserted classic script, so it was deferred behind Inter's blocking stylesheet before the skeleton could even be revealed. With both fonts non-blocking that deferral is gone: with **both Google Fonts requests aborted outright**, the skeleton still paints and `mc:skeleton-visible` fires at **11ms**.

**One thing from the ticket is deliberately not implemented.** `view-transition-name` is omitted, because a name has to exist on both the outgoing and incoming document to morph an element, and the skeleton is long gone by the time anyone navigates. Naming the React chrome alongside the skeleton is scoped to ticket C.

Both bundler paths get non-blocking app CSS: MC's own applications through the Vite plugin, starter templates and customer-built Custom Applications through `webpack-html-template.ts`.

#### Two acceptance criteria this does not meet

Flagging these so the ticket is not closed as fully accepted:

- **AC3's timing budget** — the mechanism is there (the skeleton mark fires at 11ms locally, even with fonts blocked), but AC3 asks for a p50 across runs on the integration environment and there is no performance harness in the repo yet to produce that.
- **AC5** — `useFullRedirectsForLinks` is `true` in `merchant-center-frontend`, and `menu-items.tsx:305` implements it as `location.replace()`. Programmatic navigation does not trigger a view transition, so the rule is inert until that override is retired.

AC4 (both fonts and the app CSS non-render-blocking) is met on both bundler paths.

#### Worth knowing if you consume these packages

- **`window.onAppLoaded()` is no longer always synchronous.** While the app's own stylesheets are still loading it defers removing the loader, bounded by a 2s deadline after which the app is revealed regardless. It used to remove unconditionally.
- **`data-mc-css`** is a new build-time marker emitted by `mc-scripts` and read by `mc-html-template`. The repo versions them in lockstep; keep them aligned if you pin independently.
- **`__LOADING_SCREEN_CSS__` moved into `<head>`.** This is required, not tidying: `onAppLoaded()` removes `#app-loader`, and the `@view-transition` opt-in has to still be in the document when the user navigates or the transition silently never fires.
- **Performance marks.** Only `mc:skeleton-visible` is emitted here, since the skeleton is the only thing that knows when it became visible; the rest of the `mc:*` set belongs in `application-shell`. Its measure is named `mc:skeleton-visible:duration` so `getEntriesByName('mc:skeleton-visible')` returns one entry rather than two of different `entryType`. The mark fires on the reveal call, not a confirmed paint.

#### Follow-ups

- Retire `useFullRedirectsForLinks` so view transitions actually fire in production
- Optionally pass `loadFonts={false}` at the `NimbusProvider` mounts in the app repos, to drop the now-redundant second Inter link
- `view-transition-name` on the real `NavBar`/`AppBar` for genuine chrome persistence
- Re-measure once a performance harness exists; this ships unattributed otherwise
- Fix the upstream plan doc (wrong dimensions, the CSP-invalid snippet, font URLs that do not exist, and a first-ever-visitor savings row the auth gate makes impossible)

#### Notes for reviewers

- **Expect a Chromatic diff.** `demo-custom-view.jsx` is one of the snapshot-gated playground views. Its `.loading-screen` removal stopped collapsing the loader once a 100vh skeleton sibling existed, hence the one-line change to `window.onAppLoaded()`.
- **The content area is intentionally empty**, per the ticket. On warm cache that means a motionless screen from ~100ms until React chrome arrives. Worth a look at whether it reads as loading or as broken.
- `stylelint.config.js` is relaxed for `@view-transition` and friends because stylelint is pinned at 14.16.1, which predates the API.
- This is the loading screen's first test coverage — there was none before.

[FEC-1298]: https://commercetools.atlassian.net/browse/FEC-1298?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/mc-html-template/html-scripts/loading-screen.js
+++ b/packages/mc-html-template/html-scripts/loading-screen.js
@@ -200,9 +200,14 @@
       window.performance.mark('mc:skeleton-visible');
 
       if (window.performance.measure) {
-        // Omitting `end`/`duration` measures from the time origin
+        // Distinct name on purpose: sharing it with the mark would make
+        // `getEntriesByName('mc:skeleton-visible')` return two entries of
+        // different entryType, and a consumer indexing [0] would read whichever
+        // sorted first. Omitting `end`/`duration` measures from the time origin
         // (navigationStart) to now.
-        window.performance.measure('mc:skeleton-visible', { start: 0 });
+        window.performance.measure('mc:skeleton-visible:duration', {
+          start: 0,
+        });
       }
     } catch (error) {
       // Instrumentation is best-effort.
@@ -233,7 +238,20 @@
   // invisible: before this gate existed `onAppLoaded` removed the loader
   // unconditionally, so an unbounded wait would be a new availability
   // regression. Degrade to visible-but-unstyled instead.
+  //
+  // Upgrading the stragglers first matters: a link left as `rel="preload"` would
+  // never apply its CSS even if the response arrives later, so zeroing the
+  // counter alone would trade a temporarily unstyled app for a permanently
+  // unstyled one. `false` skips the decrement, since the counter is zeroed next.
   setTimeout(function releaseStylesheetGate() {
+    const pendingEls = document.querySelectorAll(
+      'link[rel="preload"][as="style"]'
+    );
+
+    for (let index = 0; index < pendingEls.length; index += 1) {
+      upgradeStylesheetPreload(pendingEls[index], false);
+    }
+
     window.__CSS_REMAINING__ = 0;
     removeAppLoaderWhenReady();
   }, LONG_LOADING_DELAY);
--- a/packages/mc-html-template/html-scripts/loading-screen.js
+++ b/packages/mc-html-template/html-scripts/loading-screen.js
@@ -1,18 +1,12 @@
-// Drives the loading screen shown before React mounts:
-// - reveals either the authenticated app-chrome skeleton or the bare spinner
-// - upgrades non-blocking stylesheet preloads once they have loaded
-// - exposes `window.onAppLoaded`, which ConfigureIntlProvider calls when React
-//   is ready, and which removes the loader
-//
 // Needs to be compatible with all browsers supported without transpilation:
 // this file is inlined into the HTML document verbatim (only minified).
 (function initLoadingScreen() {
   const SPINNER_SHOW_DELAY = 250;
   const LONG_LOADING_DELAY = 2000;
-  // Mirrors WINDOW_SIZES.WIDE in packages/application-shell/src/constants.ts.
+  // Mirrors WINDOW_SIZES.WIDE in application-shell/src/constants.ts.
   const WIDE_VIEWPORT = 1200;
-  // Mirrors `staticUrlPathsInPositionOfProjectKey` in
-  // packages/application-shell-connectors: the shell renders no NavBar here.
+  // Mirrors `staticUrlPathsInPositionOfProjectKey`: the shell renders no
+  // NavBar on these routes.
   const NAVBARLESS_ROUTES = ['login', 'logout', 'account'];
 
   function readStorage(key) {
@@ -25,13 +19,9 @@
     }
   }
 
-  /* ---------------------------------------------------------------- *
-   * Loader removal, gated on our own stylesheets                     *
-   * ---------------------------------------------------------------- */
-
-  // Count of app stylesheets (marked `data-mc-css` by the build) that have not
-  // finished loading. React must not be revealed while any is outstanding or
-  // the app flashes unstyled.
+  // Count of app stylesheets (marked `data-mc-css` by the build) still
+  // pending. The app must not be revealed while this is above zero, or it
+  // flashes unstyled.
   window.__CSS_REMAINING__ = 0;
 
   let isAppReady = false;
@@ -50,16 +40,11 @@
     }
   }
 
-  // Assigning global callback used by ConfigureIntlProvider to remove the
-  // loading screen.
+  // Called by ConfigureIntlProvider once React is ready.
   window.onAppLoaded = function onAppLoaded() {
     isAppReady = true;
     removeAppLoaderWhenReady();
   };
-
-  /* ---------------------------------------------------------------- *
-   * Non-blocking stylesheet preloads                                 *
-   * ---------------------------------------------------------------- */
 
   function upgradeStylesheetPreload(linkEl, isCounted) {
     if (linkEl.getAttribute('data-mc-upgraded') !== null) {
@@ -76,15 +61,10 @@
   }
 
   function hasAlreadyLoaded(linkEl) {
-    // `linkEl.sheet` is always null on a preload - a preload has no associated
-    // stylesheet - so completion is detected through Resource Timing, which is
-    // populated for cross-origin responses without Timing-Allow-Origin.
-    //
-    // This matters more than it looks: this script is a parser-inserted classic
-    // script, so it is deferred while a script-blocking stylesheet is
-    // outstanding (Inter still is). By the time it runs, same-origin preloads
-    // started at parse time have usually already fired `load`, so relying on
-    // the event alone would leave them unupgraded forever.
+    // `linkEl.sheet` is always null on a preload, so completion is detected via
+    // Resource Timing instead. This matters: this script can run late enough
+    // (deferred behind another blocking stylesheet) that the `load` event on a
+    // fast preload has already fired and would otherwise be missed forever.
     if (!window.performance || !window.performance.getEntriesByName) {
       return false;
     }
@@ -101,30 +81,27 @@
   }
 
   function upgradeStylesheetPreloads() {
-    // Scoped to `as="style"` on purpose: other preload types share this
-    // document. FEC-1301 adds `as="fetch"` prefetch hints and Vite already
-    // emits `rel="modulepreload"`; rewriting either would break them.
+    // Scoped to `as="style"`: other preload types (e.g. `as="fetch"` prefetch
+    // hints, `rel="modulepreload"`) share this document and must not be touched.
     const linkEls = document.querySelectorAll(
       'link[rel="preload"][as="style"]'
     );
 
     for (let index = 0; index < linkEls.length; index += 1) {
       const linkEl = linkEls[index];
-      // Only app CSS gates the reveal. Fonts use `display=swap`, so gating on a
-      // third-party font request would delay the app for no visual benefit.
+      // Only app CSS gates the reveal; fonts already use `display=swap`.
       const isCounted = linkEl.getAttribute('data-mc-css') !== null;
 
       if (isCounted) {
         window.__CSS_REMAINING__ += 1;
       }
 
-      // Listeners are attached before the already-loaded check so a response
-      // arriving mid-loop cannot be missed.
+      // Attached before the already-loaded check, so a response arriving
+      // mid-loop cannot be missed.
       linkEl.addEventListener('load', function onStylesheetLoad() {
         upgradeStylesheetPreload(linkEl, isCounted);
       });
       linkEl.addEventListener('error', function onStylesheetError() {
-        // A failed stylesheet must never hold the app hostage.
         upgradeStylesheetPreload(linkEl, isCounted);
       });
 
@@ -133,10 +110,6 @@
       }
     }
   }
-
-  /* ---------------------------------------------------------------- *
-   * Variant selection                                                *
-   * ---------------------------------------------------------------- */
 
   function getFirstPathSegment() {
     return (window.location.pathname.split('/')[1] || '').toLowerCase();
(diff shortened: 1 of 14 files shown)


ByronDWall:
Thanks for the thorough corrections — the CSP finding and the NimbusProvider dedup issue are exactly the kind of things that only surface from verifying against the real production environment. Updated the plan and Jira based on what this PR established:

**New ticket B2 ([FEC-1319](https://commercetools.atlassian.net/browse/FEC-1319)): Non-blocking Inter font.** We own Nimbus, so no external coordination needed. The approach: NimbusProvider checks `document.querySelector('link[data-nimbus-fonts]')` before injecting its font link — if the marker exists (already in the HTML template from this PR), it skips injection. Then the template converts Inter to the same preload pattern you built for Open Sans here. This unblocks both the cold-cache 250ms skeleton target and the `loading-screen.js` parser-blocking issue you flagged.

**Named view transitions added to ticket C ([FEC-1299](https://commercetools.atlassian.net/browse/FEC-1299)).** Your correction #4 is right that naming skeleton elements alone doesn't work — the skeleton is gone by the time the user navigates away. But the fix is naming the **React chrome** on the outgoing page and the **skeleton** on the incoming page. Same screen positions, browser morphs between them. Concrete changes scoped to C:

- `navbar.tsx`: `view-transition-name: mc-sidebar`
- `app-bar.tsx`: `view-transition-name: mc-header`
- `loading-screen.css`: matching names on `.loading-skeleton__sidebar` / `.loading-skeleton__header`

This upgrades the whole-page crossfade from this PR into actual chrome persistence. C is the natural home since it already restructures those components.

**What stays as-is from this PR:** The `@view-transition { navigation: auto; }` crossfade ships here as the incremental first step — named transitions layer on top in C. The Inter link stays blocking until B2. The `loading-screen.js` CSP-safe upgrade pattern is the foundation B2 reuses.

Plan commit: [`f7e052a`](https://github.com/commercetools/mc-foundation-planning/commit/f7e052a)

[FEC-1319]: https://commercetools.atlassian.net/browse/FEC-1319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FEC-1299]: https://commercetools.atlassian.net/browse/FEC-1299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

ismaelocaramelo:
> Thanks for the thorough corrections — the CSP finding and the NimbusProvider dedup issue are exactly the kind of things that only surface from verifying against the real production environment. Updated the plan and Jira based on what this PR established:
> 
> **New ticket B2 ([FEC-1319](https://commercetools.atlassian.net/browse/FEC-1319)): Non-blocking Inter font.** We own Nimbus, so no external coordination needed. The approach: NimbusProvider checks `document.querySelector('link[data-nimbus-fonts]')` before injecting its font link — if the marker exists (already in the HTML template from this PR), it skips injection. Then the template converts Inter to the same preload pattern you built for Open Sans here. This unblocks both the cold-cache 250ms skeleton target and the `loading-screen.js` parser-blocking issue you flagged.
> 
> **Named view transitions added to ticket C ([FEC-1299](https://commercetools.atlassian.net/browse/FEC-1299)).** Your correction #4 is right that naming skeleton elements alone doesn't work — the skeleton is gone by the time the user navigates away. But the fix is naming the **React chrome** on the outgoing page and the **skeleton** on the incoming page. Same screen positions, browser morphs between them. Concrete changes scoped to C:
> 
> * `navbar.tsx`: `view-transition-name: mc-sidebar`
> * `app-bar.tsx`: `view-transition-name: mc-header`
> * `loading-screen.css`: matching names on `.loading-skeleton__sidebar` / `.loading-skeleton__header`
> 
> This upgrades the whole-page crossfade from this PR into actual chrome persistence. C is the natural home since it already restructures those components.
> 
> **What stays as-is from this PR:** The `@view-transition { navigation: auto; }` crossfade ships here as the incremental first step — named transitions layer on top in C. The Inter link stays blocking until B2. The `loading-screen.js` CSP-safe upgrade pattern is the foundation B2 reuses.
> 
> Plan commit: [`f7e052a`](https://github.com/commercetools/mc-foundation-planning/commit/f7e052a)

Quick update since this comment. Inter is in now, following the plan change that folded it into ticket B, so both fonts are preloads with <noscript> fallbacks.

application-shell didn't need changing after all. its only NimbusProvider already passes loadFonts={false}, and for hosts that mount the provider themselves and leave loadFonts on the default, the second Inter link turns out to be free: same href, so it resolves from the preload with no extra request, and first paint isn't blocked either way.

the bigger effect is that loading-screen.js was parser-blocked behind Inter's stylesheet. that's gone now. with both font requests aborted outright the skeleton still paints, and mc:skeleton-visible fires at around 11ms.

so AC4 is met on both bundler paths. AC3 and AC5 are still open

[FEC-1319]: https://commercetools.atlassian.net/browse/FEC-1319?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FEC-1299]: https://commercetools.atlassian.net/browse/FEC-1299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

packages/mc-html-template/html-scripts/loading-screen.js
```diff
@@ -1,26 +1,236 @@
...
+
+    const segment = getFirstPathSegment();
+
+    // Custom Views mount their own shell inside a host application and never
+    // render the app chrome, so the skeleton would be wrong there.
+    if (segment === 'custom-views') {
+      return null;
+    }
+
+    return { hasNavbar: NAVBARLESS_ROUTES.indexOf(segment) === -1 };
```
kterry1: This appears to be a regression. This mirrors the route list, but the shell gates the NavBar on the result of `selectProjectKeyFromUrl`, and the two disagree on /.

There, `getFirstPathSegment()` returns '', which is not in `NAVBARLESS_ROUTES`, so the skeleton paints the sidebar. But selectProjectKeyFromUrl('/') returns '', so `application-shell-authenticated.tsx:284` hits if (!projectKeyFromUrl) return null and renders none. The sidebar disappears on mount and content shifts by the navbar width. That regresses main (bare spinner, no shift) on the entry point users reach by typing the bare domain, and fails the ticket's CLS < 0.1 criterion.

Deriving hasNavbar from the shell's own condition also fixes a second divergence: .toLowerCase() on line 115 makes /Account/profile suppress the sidebar, while selectProjectKeyFromUrl returns the unlowercased 'Account' and the shell renders one.

Adding / to the it.each at `loading-screen.spec.ts:214` could catch this.

playground/src/components/custom-views/demo-custom-view.jsx
```diff
@@ -100,7 +100,10 @@ function DemoCustomView() {
     window.app.customViewId = CUSTOM_VIEW_ID;
     window.app.__DEVELOPMENT__.customViewConfig = DEMO_CUSTOM_VIEW;
     window.app.__DEVELOPMENT__.customViewHostUrl = window.location.href;
-    document.querySelector('.loading-screen').remove();
+    // Removes the whole #app-loader: it also contains a 100vh skeleton
+    // sibling now, so removing only .loading-screen would leave this view
+    // pushed below the fold.
```
kterry1: Non-blocking, but the comment above promises what the function no longer guarantees. onAppLoaded now removes #app-loader only once `window.__CSS_REMAINING__ `drains. This playground builds via mc-scripts build, which emits every stylesheet as a data-mc-css preload, so if any is still in flight at mount, `removeAppLoaderWhenReady()` bails while setIsLoading(false) runs anyway, and the view renders below the 100vh loader until the CSS resolves or the 2s deadline fires. Narrow in practice, and bounded, but either remove the loader directly here or soften the comment.

Also unguarded, unlike `configure-intl-provider.tsx:30` (if (window.onAppLoaded)), so it throws inside the effect wherever the inline script did not run.

packages/mc-html-template/html-scripts/loading-screen.js
```diff
@@ -1,26 +1,236 @@
...
+        return;
+      }
+
+      window.performance.mark('mc:skeleton-visible');
+
+      if (window.performance.measure) {
+        // Named separately from the mark, or getEntriesByName would return two
+        // entries of different entryType. Omitting end/duration measures from
+        // the time origin to now.
+        window.performance.measure('mc:skeleton-visible:duration', {
```
kterry1: Non-blocking, but worth settling before this and my PR [FEC-1297-part-2](https://github.com/commercetools/merchant-center-application-kit/pull/4144) both land. They merge cleanly and are clearly designed to fit: performance-marks.ts defines PERFORMANCE_MARKS.SKELETON_VISIBLE = 'mc:skeleton-visible' and excludes it from TShellPerformanceMark on the assumption this script emits it. But the measure names diverge, ${mark}:from-nav there against mc:skeleton-visible:duration here, so a query reading `<mark>:from-nav` across `PERFORMANCE_MARKS` gets nothing for the one mark this feature exists to report.

The mark string is also hardcoded here, necessarily, but nothing asserts it still matches `PERFORMANCE_MARKS.SKELETON_VISIBLE`. A rename either side breaks the pairing silently. Cheaper to fix now than once a published name is in a dashboard.

kterry1 (changes requested):
Just one request that needs changing before an approval and a few nits. Otherwise looks great!

packages/mc-html-template/html-scripts/loading-screen.js
ismaelocaramelo: Fixed in [5bae5c4f0](https://github.com/commercetools/merchant-center-application-kit/pull/4138/commits/5bae5c4f08253bbf5187eb85ee2fbda8579883e8)

playground/src/components/custom-views/demo-custom-view.jsx
ismaelocaramelo: Corrected the comment 👍

packages/mc-html-template/html-scripts/loading-screen.js
ismaelocaramelo: Now mc:skeleton-visible:from-nav 👍

ismaelocaramelo:
All three fixed in 5bae5c4f0, thanks @kterry1 @ByronDWall

kterry1 (approved):
Thanks for making the changes!

PR #4144, opened 2026-09-04
[FEC-1297(Part 2)] emit performance marks for mc loading

kterry1:
#### Summary

Emits the six canonical `mc:*` performance marks for the Merchant Center loading sequence, and exports `PERFORMANCE_MARKS` so the FEC-1297 harness can read the names rather than hardcoding them. Groundwork for [`FEC-1297`](https://commercetools.atlassian.net/browse/FEC-1297).

Five marks ship here. `mc:skeleton-visible` is emitted by FEC-1298 from `mc-html-template`, which is untranspiled ES5 with no module system.

<details>

**Where each mark is emitted**

| Mark | Location |
| --- | --- |
| `mc:intl-ready` | `application-shell-authenticated.tsx`, first child of `ConfigureIntlProvider` |
| `mc:shell-chrome-mounted` | Inside `ApplicationShellSplitter` |
| `mc:content-rendered` | The else-branch `MainContainer` |
| `mc:hydration-user` | `fetch-user.tsx`, on the query resolving |
| `mc:hydration-project` | `fetch-project.tsx`, on the query resolving |

**Why the marks live at the composition site, not in the components**

`ConfigureIntlProvider` renders in six places and is public API: the
unauthenticated branch of `application-shell-provider`, the OIDC callback error
page, both Custom View shells, the user-fetch error path, and the intended path.
Emitting from inside it meant a Custom View load recorded `mc:intl-ready` while
none of the other marks fired, and any custom application rendering the exported
component did the same. The two hydration marks are acceptable in
`FetchUser`/`FetchProject` because those only ever fetch the MC user and project.

**The Suspense double-mount**

`application-shell-splitter.async.tsx` uses `props.children` as its Suspense
fallback, so the whole chrome subtree mounts once while the splitter chunk is in
flight and again when it resolves. `markOnce` keeps the first write, which is
when the user actually saw the chrome; the second mount only reflects chunk-load
latency. `mc:intl-ready` sits outside the splitter and mounts once.

**API surface**

`markOnce` is deliberately internal. `TShellPerformanceMark` excludes
`mc:skeleton-visible`, so the shell cannot emit a mark that FEC-1298 already
emits from a separate module scope where this module's dedupe cannot see it, and
`performance.mark` appends rather than overwrites.

**Defensive behaviour**

`performance.measure`'s options-object form is newer than `mark` (Safari 14.1+,
Chrome 78+) and is a DOM API, so core-js does not polyfill it. Both are guarded,
and the calls are wrapped in `try/catch` because callers invoke `markOnce` from
`useEffect`, where a throw surfaces as an uncaught React error. All three
behaviours are pinned by tests: reverting the guard, rethrowing from the catch,
or moving `alreadyMarked.add` each fails at least one.

**Tests**

`performance-marks.spec.ts` covers `markOnce` directly, re-requiring the module
per test since the dedupe `Set` is module-scoped. `performance-mark.spec.tsx`
mocks `markOnce` and asserts only the component's contract. A new case in
`application-shell.spec.js` renders the authenticated shell and asserts all five
marks fire exactly once, which is the only test that would catch a refactor
dropping one from the shell.

</details>

#### Manual testing

Not required; the shell integration test covers the wiring. If you want to see it live, start a playground app and run in the console:

```
performance.getEntriesByType('measure').filter((e) => e.name.startsWith('mc:'))
```

Expect five entries with monotonically increasing durations. Note `mc:intl-ready` and `mc:shell-chrome-mounted` land in the same commit, so they will be very close together.

### DIFF
--- a/packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
+++ b/packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
@@ -28,7 +28,7 @@ import {
 } from '@commercetools-frontend/sentry';
 import { DIMENSIONS, NAVBAR } from '../../constants';
 import { TFetchLoggedInUserQuery } from '../../types/generated/mc';
-import { getPreviousProjectKey } from '../../utils';
+import { getPreviousProjectKey, PERFORMANCE_MARKS } from '../../utils';
 import AppBar from '../app-bar';
 import ApplicationLoader from '../application-loader';
 import { getBrowserLocale } from '../application-shell-provider/utils';
@@ -38,6 +38,7 @@ import ErrorApologizer from '../error-apologizer';
 import FetchProject from '../fetch-project';
 import FetchUser from '../fetch-user';
 import NavBar from '../navbar';
+import PerformanceMark from '../performance-mark';
 import ProjectContainer from '../project-container';
 import RedirectToLogout from '../redirect-to-logout';
 import RedirectToProjectCreate from '../redirect-to-project-create';
@@ -213,6 +214,9 @@ export const ApplicationShellAuthenticated = (
                   // is not loaded.
                   {...(isLoadingLocaleData ? {} : { locale, messages })}
                 >
+                  {/* Not inside `ConfigureIntlProvider`, which also renders on
+                  unauthenticated, Custom View and error surfaces. */}
+                  <PerformanceMark mark={PERFORMANCE_MARKS.INTL_READY} />
                   <SetupFlopFlipProvider
                     user={normalizedUser}
                     projectKey={projectKeyFromUrl}
@@ -221,6 +225,9 @@ export const ApplicationShellAuthenticated = (
                     defaultFlags={props.defaultFeatureFlags}
                   >
                     <ApplicationShellSplitter locale={locale ?? 'en'}>
+                      <PerformanceMark
+                        mark={PERFORMANCE_MARKS.SHELL_CHROME_MOUNTED}
+                      />
                       <ThemeSwitcher />
                       {/* NOTE: the requests in flight loader will render a loading
                       spinner into the AppBar. */}
@@ -318,6 +325,9 @@ export const ApplicationShellAuthenticated = (
                           </MainContainer>
                         ) : (
                           <MainContainer role="main">
+                            <PerformanceMark
+                              mark={PERFORMANCE_MARKS.CONTENT_RENDERED}
+                            />
                             <div ref={notificationsPageRef}>
                               <NotificationsList domain={DOMAINS.PAGE} />
                             </div>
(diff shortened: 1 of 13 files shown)


packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
```diff
@@ -213,6 +214,9 @@ export const ApplicationShellAuthenticated = (
                   // is not loaded.
                   {...(isLoadingLocaleData ? {} : { locale, messages })}
                 >
+                  {/* Not inside `ConfigureIntlProvider`, which also renders on
+                  unauthenticated, Custom View and error surfaces. */}
+                  <PerformanceMark mark={PERFORMANCE_MARKS.INTL_READY} />
```
ismaelocaramelo, line 220: Small wording thing, and I might be misreading it: the comment says "Not inside ConfigureIntlProvider" but the element is nested inside <ConfigureIntlProvider>. I take it you mean "not inside the component's own file, because that component also renders on unauthenticated / Custom View / error surfaces" — is that right? If so, maybe worth making that explicit, I read it twice before it clicked.

packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
```diff
@@ -221,6 +225,9 @@ export const ApplicationShellAuthenticated = (
                     defaultFlags={props.defaultFeatureFlags}
                   >
                     <ApplicationShellSplitter locale={locale ?? 'en'}>
```
ismaelocaramelo, line 228: Question rather than a concern: ApplicationShellSplitter comes from .async, so mc:shell-chrome-mounted waits for that chunk. On the Vite build it's the biggest one (~1.2MB), so the intl-ready → shell-chrome-mounted gap will be mostly its download. Is that the reading you want from the mark, or would a separate mark right after the intl provider be more useful to separate "shell ready" from "splitter downloaded"?

kterry1: Good question, but the mark isn't gated on the chunk. `ApplicationShellSplitter` renders `<Suspense fallback={props.children}>`, so the children (including the mark) render immediately on the fallback pass while the chunk is still downloading, and `markOnce` is first-write-wins so the fallback value is the one kept. That's exactly the double-mount case called out in the changeset: without first-write-wins the mark would record chunk-load latency instead of when the chrome appeared.

I've added a short comment at the mark site so the next reader doesn't have to trace through the `.async` file.

packages/application-shell/src/components/application-shell-authenticated/application-shell-authenticated.tsx
kterry1: Fixed

ByronDWall (approved):

PR #4146, opened 2026-09-08
FEC-1289: fix bottom save toolbar overlap with AI chat panel

nima-ct:
#### Summary

This fixes the Cancel/Save bar (SaveToolbar / SaveToolbarSteps) staying window-wide and overlapping the AI chat in the MC.

#### Description
`#mc-main-container-portal` now lives inside `Splitter.Main`, so `SaveToolbar` and `SaveToolbarSteps` from `shared-ui` stay as wide as the main column (nav + page) and do not sit on the AI chat.

If the splitter is not mounted, the same id is still rendered and pinned to the bottom of the window.

Before:
https://github.com/user-attachments/assets/dab48c90-c544-481f-91bd-dac1b5532d03
Screenshot 2026-09-08 at 14 47 36

Now
https://github.com/user-attachments/assets/c3659419-3e99-47ad-9efc-999de45f625f
Screenshot 2026-09-08 at 14 51 41

### DIFF
--- a/packages/application-shell/src/components/application-shell-splitter/application-shell-splitter.spec.tsx
+++ b/packages/application-shell/src/components/application-shell-splitter/application-shell-splitter.spec.tsx
@@ -1,4 +1,5 @@
 import { render, act } from '@testing-library/react';
+import { MC_MAIN_CONTAINER_PORTAL_ID } from '@commercetools-frontend/constants';
 import ApplicationShellSplitter from './application-shell-splitter';
 
 const mockCapturedProps: Record<string, Record<string, unknown>> = {};
@@ -83,8 +84,29 @@ describe('ApplicationShellSplitter', () => {
       );
 
       expect(mockCapturedProps['Splitter.Main']).toEqual(
-        expect.objectContaining({ containerType: 'inline-size' })
+        expect.objectContaining({
+          containerType: 'inline-size',
+          style: expect.objectContaining({
+            position: 'relative',
+            isolation: 'isolate',
+          }),
+        })
       );
+    });
+  });
+
+  describe('SaveToolbar portal target', () => {
+    it('renders #mc-main-container-portal inside Splitter.Main', () => {
+      const { getByTestId } = render(
+        <ApplicationShellSplitter {...defaultProps}>
+          <div>content</div>
+        </ApplicationShellSplitter>
+      );
+
+      const portal = getByTestId(MC_MAIN_CONTAINER_PORTAL_ID);
+      expect(getByTestId('splitter-main')).toContainElement(portal);
+      expect(getByTestId('splitter-root')).toContainElement(portal);
+      expect(getByTestId('splitter-aside')).not.toContainElement(portal);
     });
   });
 
--- a/packages/application-shell/src/components/application-shell-splitter/application-shell-splitter.spec.tsx
+++ b/packages/application-shell/src/components/application-shell-splitter/application-shell-splitter.spec.tsx
@@ -9,6 +9,15 @@ let mockHookOptions: Record<string, unknown> = {};
 jest.mock('@commercetools/nimbus', () => {
   return {
     NimbusProvider: ({ children }: { children: unknown }) => <>{children}</>,
+    Box: (p: { children?: unknown }) => {
+      const { children, ...rest } = p as Record<string, unknown>;
+      mockCapturedProps['Box'] = rest;
+      return (
+        <div data-testid={rest['data-testid'] as string}>
+          {children as never}
+        </div>
+      );
+    },
     Splitter: {
       Root: (p: { children: unknown }) => {
         const { children, ...rest } = p as Record<string, unknown>;
@@ -86,11 +95,13 @@ describe('ApplicationShellSplitter', () => {
       expect(mockCapturedProps['Splitter.Main']).toEqual(
         expect.objectContaining({
           containerType: 'inline-size',
-          style: expect.objectContaining({
-            position: 'relative',
-            isolation: 'isolate',
-          }),
+          position: 'relative',
+          overflow: 'hidden',
         })
+      );
+      expect(mockCapturedProps['Splitter.Main']).not.toHaveProperty('style');
+      expect(mockCapturedProps['Splitter.Main']).not.toHaveProperty(
+        'isolation'
       );
     });
   });
@@ -107,6 +118,27 @@ describe('ApplicationShellSplitter', () => {
       expect(getByTestId('splitter-main')).toContainElement(portal);
       expect(getByTestId('splitter-root')).toContainElement(portal);
       expect(getByTestId('splitter-aside')).not.toContainElement(portal);
+    });
+
+    it('sizes the portal as a 0-height strip with a fixed containing block', () => {
+      render(
+        <ApplicationShellSplitter {...defaultProps}>
+          <div>content</div>
+        </ApplicationShellSplitter>
+      );
+
+      expect(mockCapturedProps['Box']).toEqual(
+        expect.objectContaining({
+          id: MC_MAIN_CONTAINER_PORTAL_ID,
+          position: 'absolute',
+          left: 0,
+          right: 0,
+          bottom: 0,
+          height: 0,
+          zIndex: 10001,
+          transform: 'translateZ(0)',
+        })
+      );
     });
   });
 
--- a/packages/application-shell/src/components/application-shell-splitter/application-shell-splitter.tsx
+++ b/packages/application-shell/src/components/application-shell-splitter/application-shell-splitter.tsx
@@ -1,7 +1,7 @@
 import { type ReactNode, useCallback, useMemo, useRef, useState } from 'react';
-import { css } from '@emotion/react';
 import type { NimbusRouterConfig } from '@commercetools/nimbus';
 import {
+  Box,
   NimbusProvider,
   Splitter,
   Region,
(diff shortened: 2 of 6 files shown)


packages/application-shell/src/components/application-shell-splitter/application-shell-splitter.tsx
```diff
@@ -1,11 +1,13 @@
 import { type ReactNode, useCallback, useMemo, useRef, useState } from 'react';
+import { css } from '@emotion/react';
```
ByronDWall: There's no need for the `css` prop here, since this view has nimbus imported, examples below

packages/application-shell/src/components/application-shell-splitter/application-shell-splitter.tsx
```diff
@@ -1,11 +1,13 @@
...
 import type { NimbusRouterConfig } from '@commercetools/nimbus';
 import {
   NimbusProvider,
```
ByronDWall, line 5:
```suggestion
import {
Box,
NimbusProvider,
```

packages/application-shell/src/components/application-shell-splitter/application-shell-splitter.tsx
```diff
@@ -89,8 +113,17 @@ const ApplicationShellSplitter = (props: TApplicationShellSplitterProps) => {
         collapsedSize={0}
         collapsed={!open}
       >
-        <Splitter.Main id={SPLITTER_MAIN_ID} containerType="inline-size">
+        <Splitter.Main
+          id={SPLITTER_MAIN_ID}
+          containerType="inline-size"
+          style={splitterMainStyle}
+        >
```
ByronDWall, line 99:
```suggestion
        <Splitter.Main
          id={SPLITTER_MAIN_ID}
          containerType="inline-size"
          position="relative"
          isolation="isolate"
        >
```

packages/application-shell/src/components/application-shell-splitter/application-shell-splitter.tsx
```diff
@@ -89,8 +113,17 @@ const ApplicationShellSplitter = (props: TApplicationShellSplitterProps) => {
...
           {props.children}
+          <div
+            id={MC_MAIN_CONTAINER_PORTAL_ID}
+            data-testid={MC_MAIN_CONTAINER_PORTAL_ID}
+            css={portalInMainPaneCss}
+          />
```
ByronDWall:
```suggestion
          <Box
            id={MC_MAIN_CONTAINER_PORTAL_ID}
            data-testid={MC_MAIN_CONTAINER_PORTAL_ID}
            position="absolute"
            inset="0"
            overflow="hidden"
            pointer-events="none"
            z-index="10001"
            transform="translateZ(0)"
          />
```

packages/application-shell/src/components/application-shell-splitter/application-shell-splitter.tsx
```diff
@@ -41,6 +43,28 @@ const SPLITTER_ROOT_ID = 'mc-shell-splitter';
...
+// `z-index: 10001` is the same as the no-splitter fallback and sits above `#portals-container` (10000).
+const portalInMainPaneCss = css`
+  position: absolute;
+  inset: 0;
+  overflow: hidden;
+  pointer-events: none;
+  z-index: 10001;
+  transform: translateZ(0);
+`;
+
```
ByronDWall:
```suggestion

```

ByronDWall (commented):
Things look good overall, really more of a "how to style things using nimbus" nit.  Out of curiosity, have you tested this out to make sure it will render over the MC's "modal" pages?

nima-ct:
> Things look good overall, really more of a "how to style things using nimbus" nit. Out of curiosity, have you tested this out to make sure it will render over the MC's "modal" pages?

@ByronDWall Thx for the review. Can you provide an example of a modal page? Not sure what you are referring to

ByronDWall:
@nima-ct looking closer at your screenshot, the discount detail page should be a modal page, and if you open change history it puts 2 on the page

ByronDWall (commented):
Overall looks good. There are some relatively complex CSS rendering issues to consider here and this handles them well. I have a few suggestions and concerns, especially regarding the inset styling on the portal, but this is a really solid change given the constraints.

packages/application-shell/src/components/application-shell-splitter/application-shell-splitter.tsx
ByronDWall: on second thought, I don't think `style={splitterMainStyle}` is necessary at all, they're are already covered by the existing CSS. 

```suggestion
        <Splitter.Main
          id={SPLITTER_MAIN_ID}                            
          containerType="inline-size"
        >
```
The Nimbus pane recipe applies `isolation: isolate` to all panes, and`containerType='inline-size'` implies `contain: layout` which establishes the containing block.

packages/application-shell/src/components/application-shell-splitter/application-shell-splitter.tsx
```diff
@@ -41,6 +43,28 @@ const SPLITTER_ROOT_ID = 'mc-shell-splitter';
...
+// matches that space, not the window or the chat.
+// `inset: 0` covers the page so `pointer-events: none` lets clicks reach what is under it.
+// `z-index: 10001` is the same as the no-splitter fallback and sits above `#portals-container` (10000).
+const portalInMainPaneCss = css`
+  position: absolute;
+  inset: 0;
+  overflow: hidden;
+  pointer-events: none;
+  z-index: 10001;
+  transform: translateZ(0);
```
ByronDWall: This is a blocker. `inset: 0` renders a container that covers all of it's parent container's content. Which means that users will not be able to interact accessibly with any content on the page whenever this portal is mounted to the DOM. `pointer-events: none` disables pointer events. So nothing would be clickable in the main application window.

The portal div only needs to establish stacking order and trap fixed positioning:
```suggestion
  position: absolute;
  z-index: 10001;
  transform: translateZ(0);
```
Collapses to 0×0, sits above `#portals-container` (`z-index: 10000`), and `transform` traps `position: fixed` children (unlike `contain: layout`, which doesn't reliably do so across browsers). No overlay at all.

packages/application-shell/src/components/application-shell-splitter/application-shell-splitter.tsx
ByronDWall: As noted above, this should actually be
```suggestion
          <Box
            id={MC_MAIN_CONTAINER_PORTAL_ID}
            data-testid={MC_MAIN_CONTAINER_PORTAL_ID}
            position="absolute"
            z-index="10001"
            transform="translateZ(0)"
          />
```

ByronDWall (commented):
OK, actually took some time to sit down with this, because it's haunting me. Thanks for handling it!

packages/application-shell/src/components/application-shell-splitter/application-shell-splitter.tsx
nima-ct: `isolation="isolate"` is not needed but removing `position="relative"` reintroduces the overlap. Because the portal is absolute, it sizes against the wider window instead of Splitter.Main and the AI panel renders above the SaveToolbar.

packages/application-shell/src/components/application-shell-splitter/application-shell-splitter.tsx
nima-ct: Good catch. Made the changes [here](https://github.com/commercetools/merchant-center-application-kit/pull/4146/changes/3b11539ee88aded57fbd7086d7f41065c6c1e457)

nima-ct:
@ByronDWall Thank you for taking a closer look at my PR!
Ready for re-review,
this in particular: https://github.com/commercetools/merchant-center-application-kit/pull/4146#discussion_r3967378193

packages/application-shell/src/components/application-shell-splitter/application-shell-splitter.spec.tsx
```diff
@@ -9,6 +9,15 @@ let mockHookOptions: Record<string, unknown> = {};
 jest.mock('@commercetools/nimbus', () => {
```
ByronDWall, line 9: would `mockActual` be better than `mock` here?

packages/application-shell/src/components/application-shell-splitter/application-shell-splitter.tsx
ismaelocaramelo: Dropped. The portal is a Nimbus Box now.

packages/application-shell/src/components/application-shell-splitter/application-shell-splitter.tsx
ismaelocaramelo: Applied in 3b11539

packages/application-shell/src/components/application-shell-splitter/application-shell-splitter.tsx
ismaelocaramelo: Checked again. With position: static the bar runs into the side panel. `container-type: inline-size` was set both times. Keeping `position="relative"`. No isolation.

packages/application-shell/src/components/application-shell-splitter/application-shell-splitter.tsx
ismaelocaramelo: The overlay is gone. Same strip as above, no inset, no pointer-events.

packages/application-shell/src/components/application-shell-splitter/application-shell-splitter.spec.tsx
ismaelocaramelo: Kept jest.mock. The spec checks props, not layout. requireActual would load the pane recipe, and jsdom still would not show width or overlap....

ismaelocaramelo:
Checked the modal question in the `application-kit playground`, not in Merchant Center. The shell and `FormModalPage` are real. The bar is a local fixture portaled into `#mc-main-container-portal`, with the same fixed positioning as .container.

With the modal open, `Save` on the bar fired and the modal's own button fired. A button behind the modal did not. With the demo side panel open, the bar stays in the main column and does not overlap it.

It does sit short of the panel. That comes from the old width math in frontend (save-toolbar.module.css and save-toolbar-steps.module.css): 100% is now the pane, but the formula still subtracts the nav and  X px. This PR does not change that. Want me to open a frontend PR to drop that extra inset? @ByronDWall

PR #4147, opened 2026-09-08
refactor: simplify portal styling, add container-type to shell grid

ByronDWall:
Builds on #4146 with three changes:

1. **Remove overlay styling from portal div**: `inset: 0` renders a container covering all of `Splitter.Main`'s content, blocking accessible interaction with the page ([details](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events)). Simplified to `position: absolute; z-index: 10001` (0×0, no overlay).

2. **Remove redundant `splitterMainStyle`**: The Nimbus [pane recipe](https://github.com/commercetools/nimbus/blob/main/packages/nimbus/src/components/splitter/splitter.recipe.ts#L46-L54) already applies `isolation: isolate`, and `container-type: inline-size` implies `contain: layout`.

3. **Add `container-type: inline-size` to the shell grid div**: Gives `100cqw` as a stable width reference in both splitter and fallback paths, so the toolbar and [`PortalsContainer`](https://github.com/commercetools/merchant-center-application-kit/blob/main/packages/application-components/src/components/portals-container/portals-container.tsx#L161-L170) (which already uses `cqw`) can size with `calc(100cqw - offsets)`.

### DIFF
--- a/.changeset/save-toolbar-cqw.md
+++ b/.changeset/save-toolbar-cqw.md
@@ -1,0 +1,6 @@
+---
+'@commercetools-frontend/application-shell': patch
+'@commercetools-frontend/constants': patch
+---
+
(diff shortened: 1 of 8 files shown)


ByronDWall:
[preview_deployment]

ByronDWall:
[publish_preview]

ByronDWall:
[preview_deployment]

ByronDWall:
[preview_deployment]

ByronDWall:
[preview_deployment]

PR #4153, opened 2026-09-15
[FEC-1297(Part 3)] attach mc:* performance marks to pageload transactions

kterry1:
#### Summary

Attaches the `mc:*` loading performance marks to sampled `pageload` transactions as Sentry measurements, so they can be charted and alerted on. Part 3 of [`FEC-1297`](https://commercetools.atlassian.net/browse/FEC-1297). Self-contained in `packages/sentry`; no changes to `application-shell`.

Measurement keys swap colons for dots (`mc.intl-ready`), since Sentry rejects colons in measurement names. Navigation transactions are untouched and the sample rate is unchanged at 5%.

<details>

**Why `beforeSendTransaction` rather than reporting at the mark sites**

The pageload span is an idle span that auto-ends 1s after its last child
activity (`@sentry/core@8.55.2`, `tracing/idleSpan.js`: `idleTimeout: 1000`,
`finalTimeout: 30000`). Calling `getActiveSpan()` where a mark fires therefore
races span closure, and when it loses, `setAttribute` writes to nothing and the
sample is lost with no error.

`beforeSendTransaction` runs once per sampled transaction at send time, after
the span has ended. By then every mark that will fire has fired: the child spans
keeping the pageload span alive are the same GraphQL requests whose resolution
triggers the marks.

It also keeps the dependency direction clean. `sentry` must not import from
`application-shell` (the dependency runs the other way), and with this design it
doesn't need to: the handler filters measures by the `mc:` prefix and
`:from-nav` suffix.

**Measurement name constraint**

Sentry measurement names may not contain colons. `mc:intl-ready:from-nav` is
mapped to `mc.intl-ready` by dropping the suffix and replacing `:` with `.`.
The mapping is pinned by a test. Raw names and durations are also mirrored onto
`contexts.trace.data`, where colons are legal, which satisfies the ticket's
"span attributes" wording; measurements are what dashboards can chart.

**Behaviour**

| Event | Result |
| --- | --- |
| `pageload` transaction | `mc.*` measurements attached; raw names mirrored to trace data |
| `navigation` transaction | Untouched. The document's measures describe the original load, not the route change |
| No trace context / no Performance API / API throws | Event returned unchanged; a throw here would lose the whole transaction |
| Existing measurement keys (web vitals) | Never overwritten |

**Known accepted edge:** a query resolving more than ~1s after all other
activity can miss the transaction. That measurement is then absent for the
sample, never wrong, and the mark stays in the browser timeline where the
Part 4 harness reads it independently.

`mc:skeleton-visible` needs no special handling: FEC-1298's inline script marks
it before Sentry initialises, and the handler reads the timeline at send time.

**Tests:** eight behaviour cases on the handler plus a `boot()` wiring
assertion. Mutation-checked: removing the pageload guard fails 2 tests,
rethrowing from the catch fails 1, removing the overwrite guard fails 1, and
breaking the key mapping fails 3.

</details>


packages/sentry/src/report-performance-marks.ts
```diff
@@ -0,0 +1,47 @@
+import type { TransactionEvent } from '@sentry/types';
+
+const MEASURE_SUFFIX = ':from-nav';
```
ismaelocaramelo, line 3: I ran into the same shape in #4149 with the CSP hash (two hand-written literals that had to agree, and dev hid the divergence), so I may be pattern-matching too eagerly here. Would it be worth exporting the prefix and suffix from @commercetools-frontend/constants so all three read the same value? Or do you think the risk is low enough that a rename would obviously touch all three anyway?

packages/sentry/src/report-performance-marks.ts
```diff
@@ -0,0 +1,47 @@
...
+const toMeasurementKey = (measureName: string) =>
+  measureName.slice(0, -MEASURE_SUFFIX.length).replace(/:/g, '.');
+
+export function addPerformanceMeasurementsToTransaction(
+  event: TransactionEvent
+): TransactionEvent {
+  // The document's `mc:*` measures describe the original page load. Navigation
+  // transactions belong to SPA route changes and must not repeat them.
+  if (event.contexts?.trace?.op !== 'pageload') return event;
+  if (!performance?.getEntriesByType) return event;
```
ismaelocaramelo: Small one, and possibly not worth acting on: the catch below says a throw here would lose the whole transaction, which I agree with,, but this guard sits above the try. If performance is ever an undeclared global rather than an assigned-but-undefined one, performance?. throws a ReferenceError outside the protection. In a browser that can't happen, so this may be purely theoretical. Worth moving inside the try, or switching to typeof performance === 'undefined', or is it fine as is?

ismaelocaramelo (commented):
Great one 🚀 I've added some comments :)

Rombelirk (approved):

packages/sentry/src/report-performance-marks.ts
kterry1: Thanks for the call out. `loading-screen.js` can't import a shared const, it gets read as plain text and pasted into the HTML so we should probably leave it as is.

On the silent drift risk: each spot is already pinned by its own test
(loading-screen.spec.ts, performance-marks.spec.ts,
report-performance-marks.spec.ts), so a rename that misses one fails there.

packages/sentry/src/report-performance-marks.ts
kterry1: Fixed thanks! Likely would've never been an issue but didn't hurt to add.

PR #4154, opened 2026-09-16
fix(publish-release):  keep built package output during canary publish

kterry1:
#### Summary

The canary step runs `pnpm changeset version --snapshot canary`, which re-runs the install and triggers postinstall's `preconstruct dev`. That replaces the `pnpm build` output with dev stubs pointing at `src`, so the subsequent playground build loaded `mc-scripts` through the Babel require hook and failed with a missing `NODE_ENV`.

This PR sets `SKIP_POSTINSTALL_DEV_SETUP` to `true`, matching the guard already present on the changesets step.


ByronDWall (approved):
I could have SWORN I set this in a previous PR. Thanks for catching this!

PR #4156, opened 2026-09-18
feat: support routing for new top-level route

emmenko:
This is a proposal as part of internal work (BOTS-2437).


emmenko:
[preview_deployment]

kark (approved):
💯

